PR #95, opened 2014-09-18
Add Array.prototype.includes tests

domenic:
Passing in my local build of V8. Let's get 'em merged and then we can update V8's test262 tests and then we can rock and roll.

### DIFF
--- a/harness/testTypedArrays.js
+++ b/harness/testTypedArrays.js
@@ -1,0 +1,21 @@
+// Copyright (C) 2014 Domenic Denicola. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+function testTypedArrays(callback) {
+    [
+        Uint8Array,
+        Int8Array,
+        Uint16Array,
+        Int16Array,
+        Uint32Array,
+        Int32Array,
+        Uint8ClampedArray,
+        Float32Array,
+        Float64Array
+    ]
+    .forEach(callback);
+}
+
+testTypedArrays.floatOnly = function (callback) {
+    [Float32Array, Float64Array].forEach(callback);
+};
--- a/test/built-ins/Array/prototype/includes/adding-getter.js
+++ b/test/built-ins/Array/prototype/includes/adding-getter.js
@@ -1,0 +1,19 @@
+// Copyright (C) 2014 Domenic Denicola. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Array.prototype.includes sees a new element added by a getter that is hit during iteration
+author: Domenic Denicola
+---*/
+
+var arrayLike = {
+    length: 5,
+    0: 'a',
+    get 1() {
+        this[2] = 'c';
+        return 'b';
+    }
+};
+
+assert(Array.prototype.includes.call(arrayLike, 'c'),
+    'Expected array-like to contain "c", which was added by the getter for the 1st element');
--- a/test/built-ins/Array/prototype/includes/from-index-default.js
+++ b/test/built-ins/Array/prototype/includes/from-index-default.js
@@ -1,0 +1,24 @@
+// Copyright (C) 2014 Robert Kowalski. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: >
+    Array.prototype.includes should search the whole array, as the optional second argument fromIndex defaults to 0
+author: Robert Kowalski
+---*/
+
+assert([10, 11].includes(10), 'Expected that the whole array was searched');
+assert([10, 11].includes(11), 'Expected that the whole array was searched');
+
+var arrayLike = {
+    length: 2,
+    get 0() {
+        return '1';
+    },
+    get 1() {
+        return '2';
+    }
+};
+
+assert(Array.prototype.includes.call(arrayLike, '1'), 'Expected that the whole array-like was searched');
+assert(Array.prototype.includes.call(arrayLike, '2'), 'Expected that the whole array-like was searched');
--- a/test/built-ins/Array/prototype/Symbol.unscopables/value.js
+++ b/test/built-ins/Array/prototype/Symbol.unscopables/value.js
@@ -48,6 +48,11 @@ verifyEnumerable(unscopables, 'findIndex');
 verifyWritable(unscopables, 'findIndex');
 verifyConfigurable(unscopables, 'findIndex');
 
+assert.sameValue(unscopables.includes, true, '`includes` property value');
+verifyEnumerable(unscopables, 'includes');
+verifyWritable(unscopables, 'includes');
+verifyConfigurable(unscopables, 'includes');
+
 assert.sameValue(unscopables.keys, true, '`keys` property value');
 verifyEnumerable(unscopables, 'keys');
 verifyWritable(unscopables, 'keys');
--- a/test/built-ins/Array/prototype/includes/array-like.js
+++ b/test/built-ins/Array/prototype/includes/array-like.js
@@ -1,0 +1,17 @@
+// Copyright (C) 2014 Domenic Denicola. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Array.prototype.includes works on array-like objects
+author: Domenic Denicola
+---*/
+
+var arrayLike1 = { length: 5, 0: "a", 1: "b" };
+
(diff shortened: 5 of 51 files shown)


domenic:
I need to amend this PR a bit (there's some invalid YAML in the typed array tests apparently). But, my question is, will you accept it now that A.p.includes is stage 3? Or should I wait until later?

domenic:
I guess stage 3 is the right time to accept, since passing acceptance tests are required for stage 4.

domenic:
Tests updated and ready to pull including unscopables change

leobalter:
It would be great to set the es6id references too.

test/built-ins/Array/prototype/includes/from-index-default.js
```diff
@@ -0,0 +1,24 @@
+// Copyright (C) 2014 Robert Kowalski. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: >
+    Array.prototype.includes should search the whole array, as the optional second argument fromIndex defaults to 0
+author: Robert Kowalski
```
leobalter: is this info correct? (also the same name is on the first line)

domenic: Yes; why would it not be?

domenic:
> It would be great to set the es6id references too.

I disagree; maintaining fragile section numbers is busywork.

test/built-ins/Array/prototype/includes/from-index-default.js
leobalter: just a safe check. We've had other PRs with wrong authorship info.

jugglinmike:
Actually, as a feature proposal for the *next* version of ECMAScript,
`Array.prototype.includes` has not been assigned an identifier. I think this
trivially resolves the issue of whether or not to include an `es6id` in these
tests, but it forces a more fundamental question I've had about the order of
operations in the standardization process.

Although `Array.prototype.includes` has reached stage 4 (nice work to @domenic
and @rwaldron on that), until the publication of ES2016, it is still
non-standard. Strictly speaking, this means that ECMA has not formally
committed to the feature. From that perspective, including tests for it in the
official language test suite is somewhat contradictory.

I'm not arguing consistency for consistency's sake, though. I'm thinking of
implementors who are targeting conformance. If these tests land today,
implementors will begin failing even if they are 100% in-line with the most
recent standard.

One practical solution would be (you guessed it) additional metadata--some way
to denote that these tests describe almost-but-not-quite finalized language
features. This is effectively equivalent to `es5id`/`es6id` in that it allows
implementors to filter tests according to their expectations for their
runtimes.

@goyakin @domenic @bterlson I'm interested to hear your thoughts about this.

domenic:
The agreement was stage 3 and up proposals go in test262.

jugglinmike:
Thanks Domenic! From [September 24th](https://github.com/rwaldron/tc39-notes/blob/dd8bb263aef908509c91f58e4912482a75c1c74d/es7/2015-09/sept-24.md#8-test262-update):

> _ general agreement that test262 tests should be stage 3 _

I still think there's value in a dedicated tag, though, especially considering the "run in a browser" use case that was discussed in that session.

caitp:
it's okay for test262 tests to fail when run in a browser (they always have, that's not anything new).

jugglinmike:
What I have in mind is the "marketing feature" that Brian mentions in those minutes. You're right that it's not really an option now, but metadata like that would be a good step in that direction.

rwaldron:
@jugglinmike I agree that there needs to be a dedicated tag, even if it looks like: 

```
es7id: pending
```
(Or whatever, I'm not interested in bike shedding there)

The value is that it will make it easier to identify the tests that need to be updated with their new section numbers, when the time comes to do so.

test/built-ins/Array/prototype/includes/from-index-default.js
rwaldron: > Yes; why would it not be?

Because the name shown is not the name of the author of this commit—that's not an unreasonable inquiry. 

@bterlson Do we have confirmation that the contributing author has signed http://tc39.github.io/test262-cla/ ?

domenic:
I disagree with the division of tests into version numbers. The spec text will be in master soon. And that's what matters; spec version numbers are bulls hit. 

But anyway, feel free to modify these in any way you want while merging. I'm done here unless there are substantive technical comments.

rwaldron:
>  spec version numbers are bulls hit.

This has nothing to do with spec version numbers. There needs to be a way to correlate any given test file with the actual _thing_ that's being tested within. The number part of "esNid" was necessary to indicate where to look in which version of the spec to understand what was being tested and why. This is critical for maintaining this large of a test suite. The request is not unreasonable.

> But anyway, feel free to modify these in any way you want while merging. I'm done here unless there are substantive technical comments.

This response is really surprising and lacks any sense of respect for the committee, community, maintainers and consumers of this test suite. Surely that's not your intention? I would appreciate it if you were more willing to cooperate, I think we agree that this endeavor will be more productive when everyone is cooperating to their fullest capability. Thank in advance.

domenic:
My intention is to communicate I am done with this pull request until or unless substantive technical feedback is provided, nothing more.

leobalter:
@bterlson, I have a suggestion. This can be merged into a `pending` branch instead of `master`. As soon as we're able to have references for the tests, we can update and merge it to master through new commits / PRs.

rwaldron:
Thanks @domenic 

These tests can't be accepted in their current form because they are missing requisite frontmatter, as described in https://github.com/tc39/test262/blob/master/CONTRIBUTING.md. 

Per [Frontmatter](https://github.com/tc39/test262/blob/master/CONTRIBUTING.md#frontmatter):

> There must be exactly one Frontmatter per test.

And of the supported frontmatter tags, [**description** is required](https://github.com/tc39/test262/blob/master/CONTRIBUTING.md#description)

harness/testTypedArrays.js
```diff
@@ -0,0 +1,21 @@
+// Copyright (C) 2014 Domenic Denicola. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
```
rwaldron: Missing required frontmatter `description` tag.

test/built-ins/Array/prototype/includes/adding-getter.js
```diff
@@ -0,0 +1,19 @@
+// Copyright (C) 2014 Domenic Denicola. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Array.prototype.includes sees a new element added by a getter that is hit during iteration
+author: Domenic Denicola
+---*/
```
rwaldron: :+1:

harness/testTypedArrays.js
domenic: This is not a test.

test/built-ins/Array/prototype/includes/from-index-default.js
rwaldron: Update: I reviewed the CLA responses and there is no "Robert Kowalski". Please have the author sign http://tc39.github.io/test262-cla/

harness/testTypedArrays.js
rwaldron: Indeed, sorry for the confusion there, I was scanning through all of the files quickly. I'll update my main comment accordingly.

test/built-ins/Array/prototype/includes/from-index-default.js
domenic: @robertkowalski

rwaldron:
@robertkowalski thanks for the contribution to @domenic's Array.prototype.incudes test suite! If you would kindly sign the [Ecma International, TC39 ECMAScript Conformance Test Suites Software Contribution Form](http://tc39.github.io/test262-cla/), your tests will be :+1: for inclusion!

bterlson:
To clarify, I am not saying your tests are a mess (they are overall very good!), just pointing out that it's important for champions to write tests acceptable to the test262 project contributors. I this makes sense.

test/built-ins/Array/prototype/includes/from-index-default.js
robertkowalski: submitted! :)

rwaldron: @robertkowalski confirmed—thank you very much!

jugglinmike:
Superseded by gh-643.

PR #96, opened 2014-09-23
Exponentiation Operator draft tests

rwaldron:
This is not expected to land any time soon. It's really just for show.


Yaffle:
Should the result be exact, if both operands are integers and that result can be exactly represented ?

Yaffle:
Math.pow(3, 3) === 26.999999999999993 in some versions of Opera Classic

rwaldron:
@yaffle I'm going to bring this up at the next TC39 meeting

Yaffle:
```
 * Accuracy:
 *	pow(x,y) returns x**y nearly rounded. In particular
 *			pow(integer,integer)
 *	always returns the correct integer provided it is 
 *	representable.
```
source : http://www.netlib.org/fdlibm/e_pow.c

Firefox and Chrome uses "exponentiation by squaring" when second arg is a "small" integer, so integer**integer == <safe integer> should be exact

rwaldron:
@bterlson updated with latest tests

test/language/expressions/exponentiation/exponentiation-operator-accessor-invocation.js
```diff
@@ -0,0 +1,29 @@
...
+  get p() {
+    calls++;
+    return q;
+  }
+};
+
+o.p.q **= 2;
+
+assert.sameValue(calls, 1, "Expected o.p.q accessor to be called once");
+assert.sameValue(o.p.q, 9, "Expected (o.p.q **= 2) to be 9");
```
leobalter: this o.p.q test seems more appropriated in a syntax test file, where it gets the value from the lefthandexpression.

goyakin:
Can we add tests for edge cases in the [Applying the ** operator](https://rwaldron.github.io/exponentiation-operator/#applying-the-exp-operator) section of the proposal?

test/language/expressions/exponentiation/exponentiation-operator.js
```diff
@@ -0,0 +1,20 @@
...
+
+
+assert.sameValue(2 ** 3, 8, "Expected (3 * 2 ** 3) to be 8");
+assert.sameValue(3 * 2 ** 3, 24, "Expected (3 * 2 ** 3) to be 24");
+
+var x = 2;
+assert.sameValue(2 ** ++x, 8, "Expected (2 ** ++x) to be 8");
+assert.sameValue(2 ** -1 * 2, 512, "Expected (2 ** -1 * 2) to be 1");
+assert.sameValue(2 ** (3 ** 2), 512, "Expected (2 ** (3 ** 2)) to be 512");
+assert.sameValue(2 ** 3 ** 2, 512, "Expected (2 ** 3 ** 2) to be 512");
```
goyakin: Since we're testing operator precedence here, shouldn't these be in `test/language/expressions/exponentiation/operator-precedence.js` instead?

test/language/expressions/exponentiation/exponentiation-operator-accessor-invocation.js
```diff
@@ -0,0 +1,29 @@
...
+var o = {
+  get p() {
+    calls++;
+    return q;
+  }
+};
+
+o.p.q **= 2;
+
+assert.sameValue(calls, 1, "Expected o.p.q accessor to be called once");
```
goyakin: Do we care how many calls are made for `q` property in var `q`?

rwaldron:
@goyakin "Applying the ** Operator" is just Math.pow definition, which is tested here: https://github.com/tc39/test262/tree/master/test/built-ins/Math/pow

test/language/expressions/exponentiation/exponentiation-operator-accessor-invocation.js
rwaldron: No, that's a typo, will fix.

domenic:
I think you need to actually test that implementations follow the same rules. The spec states they should, but the whole point of test262 is to make sure people are following the spec.

This is similar to how you can't just test a given abstract operation once and assume everywhere that is specced to call it, calls it in implementations. You have to test each call site.

test/language/expressions/exponentiation/exponentiation-operator-accessor-invocation.js
rwaldron: On further look, that's a totally bogus test, so it's getting removed.

rwaldron:
> I think you need to actually test that implementations follow the same rules. The spec states they should, but the whole point of test262 is to make sure people are following the spec.

Are you saying that the same tests should just be duplicated? The spec normatively defines that both `**` and `Math.pow` do the same thing—there's only one definition of the exponentiation operation, and that's defined in the section titled "Applying the ** Operator"

domenic:
Yeah, I think they definitely need to be duplicated. If we had a test-gen framework we would be in better shape. But not duplicating them would be like only testing destructuring assignment, not destructuring binding. The spec says they use the same rules, but we clearly need tests for both.

leobalter:
> @domenic:  If we had a test-gen framework we would be in better shape. 

This is what we want on #467, it's already a work in progress.

rwaldron:
Updated to include tests for "Applying the ** Operator". Here's the output of running SpiderMonkey's latest "js" binary:

```
language/expressions/exponentiation/applying-the-exp-operator_A1 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A10 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A11 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A12 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A13 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A14 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A15 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A16 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A17 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A18 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A19 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A2 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A20 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A21 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A22 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A23 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A24 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A3 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A4 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A5 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A6 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A7 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A8 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A9 passed in non-strict mode
language/expressions/exponentiation/exponentiation-assignment-operator passed in non-strict mode
language/expressions/exponentiation/exponentiation-operator passed in non-strict mode
=== language/expressions/exponentiation/negative-base-requires-grouping was expected to fail in non-strict mode, but didn't ===
--- expected error: SyntaxError ---

===
language/expressions/exponentiation/operator-precedence passed in non-strict mode

```

rwaldron:
Looks like my latest push did not arrive...?

rwaldron:
After a rebase, updates again. Here's the same run, new output: 

```
language/expressions/exponentiation/applying-the-exp-operator_A1 passed in strict mode
language/expressions/exponentiation/applying-the-exp-operator_A1 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A10 passed in strict mode
language/expressions/exponentiation/applying-the-exp-operator_A10 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A11 passed in strict mode
language/expressions/exponentiation/applying-the-exp-operator_A11 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A12 passed in strict mode
language/expressions/exponentiation/applying-the-exp-operator_A12 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A13 passed in strict mode
language/expressions/exponentiation/applying-the-exp-operator_A13 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A14 passed in strict mode
language/expressions/exponentiation/applying-the-exp-operator_A14 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A15 passed in strict mode
language/expressions/exponentiation/applying-the-exp-operator_A15 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A16 passed in strict mode
language/expressions/exponentiation/applying-the-exp-operator_A16 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A17 passed in strict mode
language/expressions/exponentiation/applying-the-exp-operator_A17 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A18 passed in strict mode
language/expressions/exponentiation/applying-the-exp-operator_A18 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A19 passed in strict mode
language/expressions/exponentiation/applying-the-exp-operator_A19 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A2 passed in strict mode
language/expressions/exponentiation/applying-the-exp-operator_A2 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A20 passed in strict mode
language/expressions/exponentiation/applying-the-exp-operator_A20 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A21 passed in strict mode
language/expressions/exponentiation/applying-the-exp-operator_A21 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A22 passed in strict mode
language/expressions/exponentiation/applying-the-exp-operator_A22 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A23 passed in strict mode
language/expressions/exponentiation/applying-the-exp-operator_A23 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A24 passed in strict mode
language/expressions/exponentiation/applying-the-exp-operator_A24 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A3 passed in strict mode
language/expressions/exponentiation/applying-the-exp-operator_A3 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A4 passed in strict mode
language/expressions/exponentiation/applying-the-exp-operator_A4 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A5 passed in strict mode
language/expressions/exponentiation/applying-the-exp-operator_A5 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A6 passed in strict mode
language/expressions/exponentiation/applying-the-exp-operator_A6 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A7 passed in strict mode
language/expressions/exponentiation/applying-the-exp-operator_A7 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A8 passed in strict mode
language/expressions/exponentiation/applying-the-exp-operator_A8 passed in non-strict mode
language/expressions/exponentiation/applying-the-exp-operator_A9 passed in strict mode
language/expressions/exponentiation/applying-the-exp-operator_A9 passed in non-strict mode
language/expressions/exponentiation/exponentiation-assignment-operator passed in strict mode
language/expressions/exponentiation/exponentiation-assignment-operator passed in non-strict mode
language/expressions/exponentiation/exponentiation-operator passed in strict mode
language/expressions/exponentiation/exponentiation-operator passed in non-strict mode
=== language/expressions/exponentiation/negative-base-requires-grouping was expected to fail in strict mode, but didn't ===
--- expected error: SyntaxError ---

===
=== language/expressions/exponentiation/negative-base-requires-grouping was expected to fail in non-strict mode, but didn't ===
--- expected error: SyntaxError ---

===
language/expressions/exponentiation/operator-precedence passed in strict mode
language/expressions/exponentiation/operator-precedence passed in non-strict mode
```

goyakin:
Can we change the `applying-the-exp-operator` prefix in the `Math.pow` filenames to something else? They don't test the operator.

rwaldron:
> Math.pow -> ** in the frontmatter. Can you also remove the es5id and change es6id to es7id?

Yes, update to follow

> Can we change the applying-the-exp-operator prefix in the Math.pow filenames to something else? They don't test the operator.

Math.pow is now defined as: 

> Return the result of Applying the ** operator with base and exponent as specified in 12.7.3.4.

But I agree with changing for clarity sake, so how about this: `math-pow-applying-the-exp-operator...`?

goyakin:
I missed the reference in the spec update from `Math.pow` to the `Applying the ** operator` section. If the spec stays that way, I think the filenames are okay the way they are now.

rwaldron:
> I missed the reference in the spec update from Math.pow to the Applying the ** operator section. If the spec stays that way, I think the filenames are okay the way they are now.

It's not going to change :)

test/language/expressions/exponentiation/exp-operator-precedence-unary-expression-semantics.js
```diff
@@ -0,0 +1,63 @@
...
+assert.sameValue(2 ** void 1, NaN);
+
+// typeof 1 -> "number" -> ToNumber("number") -> NaN
+assert.sameValue(2 ** typeof 1, NaN);
+
+var s = "2";
+var n = 2;
+
+// +s -> +"2" -> 2 -> ToNumber(2) -> 2
+assert.sameValue(2 ** +s, 4);
```
caitp: This makes use of the ToNumber() from https://tc39.github.io/ecma262/#sec-unary-plus-operator-runtime-semantics-evaluation rather than the ToNumber() in http://rwaldron.github.io/exponentiation-operator/#exp-operator-runtime-semantics-evaluation --- is that intentional?

rwaldron: > http://rwaldron.github.io/exponentiation-operator/#exp-operator-runtime-semantics-evaluation

Is wrong! The latest, correct spec is in the current spec draft. I actually just pushed up a new test a bit earlier, it's based on the bug I introduced there—trying to optimize the spec language—which @anba spotted and [I corrected](https://github.com/tc39/ecma262/commit/28bdf0798662bcb8efb11c4070f8a981a496ccc1#diff-3540caefa502006d8a33cb1385720803R12301) : https://github.com/rwaldron/test262/blob/exponentiation-operator/test/language/expressions/exponentiation/exp-operator-evaluation-order.js 

Slightly modify to make it easier to run in console: 

```js
var leftValue = { valueOf() { console.log("leftValue"); return 3; }};
var rightValue = { valueOf() { console.log("rightValue"); return 2; }};

(console.log("left"), leftValue) - +(console.log("right"), rightValue);
// left
// right
// rightValue
// leftValue

(console.log("left"), leftValue) * +(console.log("right"), rightValue);
// left
// right
// rightValue
// leftValue


(console.log("left"), leftValue) / +(console.log("right"), rightValue);
// left
// right
// rightValue
// leftValue


(console.log("left"), leftValue) / void (console.log("right"), rightValue);
// left
// right
// leftValue
// NaN

(console.log("left"), leftValue) * delete (console.log("right"), rightValue);
// left
// right
// leftValue
// 3
```


So anyway, that's the long-ass version of "yes, that's expected" :D

rwaldron:
Rebased and updated (es7id -> id)

test/language/expressions/exponentiation/applying-the-exp-operator_A1.js
```diff
@@ -0,0 +1,26 @@
+// Copyright 2009 the Sputnik authors.  All rights reserved.
```
leobalter: these are new files for a feature previously not available.

I believe the year and author should be changed. The license file is already referenced, regarding the BSD demands.

test/language/expressions/exponentiation/exp-operator-precedence-unary-expression-semantics.js
```diff
@@ -0,0 +1,78 @@
+// Copyright (C) 2015 Rick Waldron. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+author: Rick Waldron
+email: waldron.rick@gmail.com
```
leobalter: I'm afraid the email tag [is not supported on our frontmatter](https://github.com/tc39/test262/blob/master/CONTRIBUTING.md#frontmatter).

test/language/expressions/exponentiation/applying-the-exp-operator_A1.js
rwaldron: Updated

test/language/expressions/exponentiation/exp-operator-precedence-unary-expression-semantics.js
rwaldron: Updated

rwaldron:
Rebased on 5cb97c293b4a44ee1c31961a07663993c521fb5f

test/language/expressions/exponentiation/applying-the-exp-operator_A13.js
```diff
@@ -0,0 +1,22 @@
+// Copyright 2016 Rick Waldron.  All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+id: sec-applying-the-exp-operator
+description: >
+    Checking if (base ** exponent) equals to -Infinity, where base is
+    -Infinity and exponent>0
```
goyakin: Can you mention this is when `exponent` is **odd**?

test/language/expressions/exponentiation/applying-the-exp-operator_A14.js
```diff
@@ -0,0 +1,24 @@
+// Copyright 2016 Rick Waldron.  All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+id: sec-applying-the-exp-operator
+description: >
+    Checking if (base ** exponent) equals to +Infinity, where base is
+    -Infinity and exponent>0
```
goyakin: ... when `exponent` is **even**?

test/language/expressions/exponentiation/applying-the-exp-operator_A15.js
```diff
@@ -0,0 +1,22 @@
+// Copyright 2016 Rick Waldron.  All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+id: sec-applying-the-exp-operator
+description: >
+    Checking if (base ** exponent) equals to -0, where base is -Infinity and
+    exponent<0
```
goyakin: ... when `exponent` **is** an odd integer?

test/language/expressions/exponentiation/applying-the-exp-operator_A16.js
```diff
@@ -0,0 +1,24 @@
+// Copyright 2016 Rick Waldron.  All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+id: sec-applying-the-exp-operator
+description: >
+    Checking if (base ** exponent) equals to +0, where base is -Infinity and
+    exponent<0
```
goyakin: ... when `exponent` **is not** an odd integer?

test/language/expressions/exponentiation/applying-the-exp-operator_A19.js
```diff
@@ -0,0 +1,20 @@
+// Copyright 2016 Rick Waldron.  All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+id: sec-applying-the-exp-operator
+description: Checking if (base ** exponent) equals to -0, where base is -0 and exponent>0
```
goyakin: ... when `exponent` **is** an odd integer?

test/language/expressions/exponentiation/applying-the-exp-operator_A21.js
```diff
@@ -0,0 +1,22 @@
+// Copyright 2016 Rick Waldron.  All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+id: sec-applying-the-exp-operator
+description: >
+    Checking if (base ** exponent) equals to -Infinity, where base is -0 and exponent
+    is an odd integer
```
goyakin: ... when `exponent < 0`

test/language/expressions/exponentiation/applying-the-exp-operator_A23.js
```diff
@@ -0,0 +1,35 @@
...
+bases[3] = -0.000000000000001;
+
+exponents[0] = -Math.PI;
+exponents[1] = -Math.E;
+exponents[2] = -1.000000000000001;
+exponents[3] = -0.000000000000001;
+exponents[4] = 0.000000000000001;
+exponents[5] = 1.000000000000001;
+exponents[6] = Math.E;
+exponents[7] = Math.PI;
```
goyakin: What do you think about adding `1.7976931348623157E308` as an exponent?

test/language/expressions/exponentiation/applying-the-exp-operator_A3.js
```diff
@@ -0,0 +1,28 @@
...
+
+/*---
+id: sec-applying-the-exp-operator
+description: >
+    Checking if (base ** exponent) equals to 1, where base is number or NaN
+    and exponent is -0
+---*/
+
+
+var exponent = +0;
```
goyakin: `+0` -> `-0`

test/language/expressions/exponentiation/exp-operator-evaluation-order.js
```diff
@@ -0,0 +1,31 @@
...
+var rightValue = { valueOf() { capture.push("rightValue"); return 2; }};
+
+(capture.push("left"), leftValue) ** (capture.push("right"), rightValue);
+
+// Expected per operator evaluation order: "left", "right", "leftValue", "rightValue"
+
+assert.sameValue(capture[0], "left");
+assert.sameValue(capture[1], "right");
+assert.sameValue(capture[2], "leftValue");
+assert.sameValue(capture[3], "rightValue");
```
goyakin: These need assertion messages.

test/language/expressions/exponentiation/exp-operator-precedence-unary-expression-semantics.js
```diff
@@ -0,0 +1,77 @@
...
+var rightValue = { valueOf() { capture.push("rightValue"); return 2; }};
+
+(capture.push("left"), leftValue) ** +(capture.push("right"), rightValue);
+
+// Expected per operator evaluation order: "left", "right", "leftValue", "rightValue"
+
+assert.sameValue(capture[0], "left");
+assert.sameValue(capture[1], "right");
+assert.sameValue(capture[2], "leftValue");
+assert.sameValue(capture[3], "rightValue");
```
goyakin: These need assertion messages.

test/language/expressions/exponentiation/exp-operator-precedence-unary-expression-semantics.js
```diff
@@ -0,0 +1,77 @@
...
+assert.sameValue(2 ** ~n, 0.125);
+
+// !s -> !"2" -> false -> ToNumber(false) -> 0
+assert.sameValue(2 ** !s, 1);
+// !s -> !2 -> false -> ToNumber(false) -> 0
+assert.sameValue(2 ** !n, 1);
+
+
+
+var capture = [];
```
goyakin: We can remove this section since we already have it in `exp-operator-evaluation-order.js`.

test/language/expressions/exponentiation/exp-operator-precedence-unary-expression-semantics.js
```diff
@@ -0,0 +1,77 @@
...
+
+// ~s -> ~"2" -> -3 -> ToNumber(-3) -> -3
+assert.sameValue(2 ** ~s, 0.125);
+// ~s -> ~2 -> -3 -> ToNumber(-3) -> -3
+assert.sameValue(2 ** ~n, 0.125);
+
+// !s -> !"2" -> false -> ToNumber(false) -> 0
+assert.sameValue(2 ** !s, 1);
+// !s -> !2 -> false -> ToNumber(false) -> 0
+assert.sameValue(2 ** !n, 1);
```
goyakin: These need assertion messages.

test/language/expressions/exponentiation/exp-operator-precedence-update-expression-semantics.js
```diff
@@ -0,0 +1,61 @@
...
+var base = 4;
+assert.sameValue(--base ** 2, 9, "(--base ** 2) === 9");
+assert.sameValue(++base ** 2, 16, "(++base ** 2) === 16");
+assert.sameValue(base++ ** 2, 16, "(base++ ** 2) === 16");
+assert.sameValue(base-- ** 2, 25, "(base-- ** 2) === 25");
+
+assert.sameValue(--base ** 2, 9, "(--base ** 2) === 9");
+assert.sameValue(++base ** 2, 16, "(++base ** 2) === 16");
+assert.sameValue(base++ ** 2, 16, "(base++ ** 2) === 16");
+assert.sameValue(base-- ** 2, 25, "(base-- ** 2) === 25");
```
goyakin: These are duplicate.

test/language/expressions/exponentiation/exp-operator.js
```diff
@@ -0,0 +1,21 @@
...
+var exponent = 2;
+assert.sameValue(2 ** 3, 8, "(2 ** 3) === 8");
+assert.sameValue(3 * 2 ** 3, 24, "(3 * 2 ** 3) === 24");
+assert.sameValue(2 ** ++exponent, 8, "(2 ** ++exponent) === 8");
+assert.sameValue(2 ** -1 * 2, 1, "(2 ** -1 * 2) === 1");
+assert.sameValue(2 ** 2 * 4, 16, "(2 ** 2 * 4) === 16");
+assert.sameValue(2 ** 2 / 2, 2, "(2 ** 2 / 2) === 2");
+assert.sameValue(2 ** (3 ** 2), 512, "(2 ** (3 ** 2)) === 512");
+assert.sameValue(2 ** 3 ** 2, 512, "(2 ** 3 ** 2) === 512");
+assert.sameValue(2 * 3 ** 2, 18, "(2 * 3 ** 2) === 18");
```
goyakin: Since we already have `3 * 2 ** 3` (the second assertion), any reason we would need this?

test/language/expressions/exponentiation/applying-the-exp-operator_A3.js
goyakin: Looks like `test/built-ins/Math/pow/applying-the-exp-operator_A3.js` needs to be updated too.

goyakin:
@rwaldron, things LGTM except a few minor comments I left inline.

test/language/expressions/exponentiation/exp-operator.js
rwaldron: Ha! Nope, not at all.

test/language/expressions/exponentiation/exp-operator-precedence-update-expression-semantics.js
rwaldron: Removed.

test/language/expressions/exponentiation/exp-operator-precedence-unary-expression-semantics.js
rwaldron: Removed, but kept the second part because it tests something different.

test/language/expressions/exponentiation/exp-operator-evaluation-order.js
rwaldron: Added

test/language/expressions/exponentiation/applying-the-exp-operator_A3.js
rwaldron: Wow, that's been overlooked for many years—nice catch!

test/language/expressions/exponentiation/applying-the-exp-operator_A23.js
rwaldron: Added

test/language/expressions/exponentiation/applying-the-exp-operator_A21.js
rwaldron: Ugh. This was copied from the existing Math.pow tests :|

test/language/expressions/exponentiation/applying-the-exp-operator_A19.js
rwaldron: ¯\_(ツ)_/¯ I didn't write it—this came from the existing Math.pow tests. I just ran a mechanical transform to update most of the content.

test/language/expressions/exponentiation/applying-the-exp-operator_A23.js
rwaldron: Removed:

```
Number.isInteger(Number.MAX_VALUE); // true
```

But it's not a "safe" integer

rwaldron:
@goyakin I'm pushing a bunch of updates

- matched all the descriptions to the text found in the spec, for both Math.pow and Exp Op
- made updates according to your feedback points
- added a new test for `Math.pow` (a property verification)

rwaldron:
I guess I need to update "id" to "esid"...?

leobalter:
#516 will confirm this, the answer is probably yes.

goyakin:
@rwaldron FYI, I just merged #516.

rwaldron:
@leobalter @goyakin updated and rebased

test/built-ins/Math/pow/name.js
```diff
@@ -2,11 +2,9 @@
 // This code is governed by the BSD license found in the LICENSE file.
 
 /*---
-es6id: 20.2.2.26
+esid: sec-math.pow
 description: >
   Math.pow.name is "pow".
-info: >
-  Math.pow ( x, y )
```
leobalter: You should keep the `info` tag here.

test/built-ins/Math/pow/length.js
```diff
@@ -2,18 +2,16 @@
 // This code is governed by the BSD license found in the LICENSE file.
 
 /*---
-es6id: 20.2.2.26
+esid: sec-math.pow
 description: >
   Math.pow.length is 2.
-info: >
-  Math.pow ( x, y )
```
leobalter: keep the info tag. 

the `description` tag is designed to be a single line, where info holds any extra information.

test/language/expressions/exponentiation/exp-assignment-operator.js
```diff
@@ -0,0 +1,24 @@
...
+      LeftHandSideExpression AssignmentOperator AssignmentExpression
+
+    1. Let lref be the result of evaluating LeftHandSideExpression.
+    2. Let lval be ? GetValue(lref).
+    3. Let rref be the result of evaluating AssignmentExpression.
+    4. Let rval be ? GetValue(rref).
+    5. Let op be the @ where AssignmentOperator is @=.
+    6. Let r be the result of applying op to lval and rval as if evaluating the expression lval op rval.
+    7. Perform ? PutValue(lref, r).
+    8. Return r.
```
leobalter: Can we move this to `info` and have a short, single-lined, description?

leobalter: also in:
- language/expressions/exponentiation/exp-operator-evaluation-order.js
- language/expressions/exponentiation/exp-operator-precedence-unary-expression-semantics.js
- language/expressions/exponentiation/exp-operator-precedence-update-expression-semantics.js
- ...

rwaldron: Done

leobalter:
LGTM :shipit:

test/language/expressions/exponentiation/exp-operator-precedence-unary-expression-semantics.js
```diff
@@ -0,0 +1,64 @@
...
+assert.sameValue(2 ** ~s, 0.125, "~s -> ~'2' -> -3 -> ToNumber(-3) -> -3");
+assert.sameValue(2 ** ~n, 0.125, "~s -> ~2 -> -3 -> ToNumber(-3) -> -3");
+
+assert.sameValue(2 ** !s, 1, "!s -> !'2' -> false -> ToNumber(false) -> 0");
+assert.sameValue(2 ** !n, 1, "!s -> !2 -> false -> ToNumber(false) -> 0");
+
+
+var capture = [];
+
+(capture.push("left"), leftValue) ** +(capture.push("right"), rightValue);
```
leobalter: oops, missing quotes on leftValue and rightValue

leobalter:
as I mentioned, it LGTM, there's just a minor thing missing (see comment above), preventing the tests to pass.

test/language/expressions/exponentiation/exp-operator-precedence-unary-expression-semantics.js
rwaldron: No, I accidentally deleted the definitions of those objects. Check the test just before this one

rwaldron:
@leobalter all set, I fixed that missing object definition

leobalter:
:shipit:

rwaldron:
@goyakin your review points have been addressed. This is all set to go whenever you're ready

test/language/expressions/exponentiation/exp-assignment-operator.js
goyakin: Looks like we have more of these. @rwaldron, could you update the rest too?

rwaldron: Ok... well I think I got them all. TBH, I didn't bother looking at anything but the ones @leobalter mentioned...

goyakin:
@rwaldron Just FYI, I converted `info` back to `description` in `test/language/expressions/exponentiation/exp-operator.js` before merging this PR.

PR #149, opened 2015-02-11
First v8 tests

jugglinmike:
This is the first of a series of test imports from the Google V8 project. The commit message body for this patch contains information relating to the provenance of the tests.

Some modifications were made to retrofit those tests to the Test262 harness. Since each change has the potential to introduce a new bug, I have opted to include the V8 "MJSUnit" assertion suite as a helper utility in Test262. It may seem like overkill for the small set of tests here, but it may be re-used by future contributions from V8.

### DIFF
--- a/test/annexB/B.2.3.10.js
+++ b/test/annexB/B.2.3.10.js
@@ -1,0 +1,27 @@
+// Copyright (C) 2014 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+// Tests taken from:
+// http://mathias.html5.org/tests/javascript/string/
+
+/*---
+ description: >
+     String.prototype.link returns a string of HTML describing a single HTML
+     anchor element. The element's content is the `this` value of the function
+     invocation, coerced to a string. If specified, the first argument will be
+     coerced to a string, escaped, and set as the element's `href` attribute.
+ es6id: B.2.3.10
+ ---*/
+
+assert.sameValue('_'.link('b'), '<a href="b">_</a>');
+assert.sameValue('<'.link('<'), '<a href="<"><</a>');
+assert.sameValue('_'.link(0x2A), '<a href="42">_</a>');
+assert.sameValue('_'.link('\x22'), '<a href="&quot;">_</a>');
+assert.sameValue(String.prototype.link.call(0x2A, 0x2A), '<a href="42">42</a>');
+assert.throws(TypeError, function() {
+  String.prototype.link.call(undefined);
+});
+assert.throws(TypeError, function() {
+  String.prototype.link.call(null);
+});
+assert.sameValue(String.prototype.link.length, 1);
--- a/test/annexB/B.2.3.11.js
+++ b/test/annexB/B.2.3.11.js
@@ -1,0 +1,24 @@
+// Copyright (C) 2014 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+// Tests taken from:
+// http://mathias.html5.org/tests/javascript/string/
+
+/*---
+ description: >
+     String.prototype.small returns a string of HTML describing a single HTML
+     small print element. The element's content is the `this` value of the
+     function invocation, coerced to a string.
+ es6id: B.2.3.11
+ ---*/
+
+assert.sameValue('_'.small(), '<small>_</small>');
+assert.sameValue('<'.small(), '<small><</small>');
+assert.sameValue(String.prototype.small.call(0x2A), '<small>42</small>');
+assert.throws(TypeError, function() {
+  String.prototype.small.call(undefined);
+});
+assert.throws(TypeError, function() {
+  String.prototype.small.call(null);
+});
+assert.sameValue(String.prototype.small.length, 0);
--- a/test/annexB/B.2.3.12.js
+++ b/test/annexB/B.2.3.12.js
@@ -1,0 +1,24 @@
+// Copyright (C) 2014 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+// Tests taken from:
+// http://mathias.html5.org/tests/javascript/string/
+
+/*---
+ description: >
+     String.prototype.strike returns a string of HTML describing a single HTML
+     strikethrough element. The element's content is the `this` value of the
+     function invocation, coerced to a string.
+ es6id: B.2.3.12
+ ---*/
+
+assert.sameValue('_'.strike(), '<strike>_</strike>');
+assert.sameValue('<'.strike(), '<strike><</strike>');
+assert.sameValue(String.prototype.strike.call(0x2A), '<strike>42</strike>');
+assert.throws(TypeError, function() {
+  String.prototype.strike.call(undefined);
+});
+assert.throws(TypeError, function() {
+  String.prototype.strike.call(null);
+});
+assert.sameValue(String.prototype.strike.length, 0);
--- a/test/annexB/B.2.3.13.js
+++ b/test/annexB/B.2.3.13.js
@@ -1,0 +1,24 @@
+// Copyright (C) 2014 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+// Tests taken from:
+// http://mathias.html5.org/tests/javascript/string/
+
+/*---
+ description: >
+     String.prototype.sub returns a string of HTML describing a single HTML
+     subscript element. The element's content is the `this` value of the
+     function invocation, coerced to a string.
+ es6id: B.2.3.13
+ ---*/
+
+assert.sameValue('_'.sub(), '<sub>_</sub>');
+assert.sameValue('<'.sub(), '<sub><</sub>');
+assert.sameValue(String.prototype.sub.call(0x2A), '<sub>42</sub>');
+assert.throws(TypeError, function() {
+  String.prototype.sub.call(undefined);
+});
+assert.throws(TypeError, function() {
+  String.prototype.sub.call(null);
+});
+assert.sameValue(String.prototype.sub.length, 0);
--- a/test/annexB/B.2.3.14.js
+++ b/test/annexB/B.2.3.14.js
@@ -1,0 +1,24 @@
+// Copyright (C) 2014 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+// Tests taken from:
+// http://mathias.html5.org/tests/javascript/string/
+
+/*---
+ description: >
+     String.prototype.sup returns a string of HTML describing a single HTML
+     superscript element. The element's content is the `this` value of the
+     function invocation, coerced to a string.
+ es6id: B.2.3.14
+ ---*/
+
+assert.sameValue('_'.sup(), '<sup>_</sup>');
+assert.sameValue('<'.sup(), '<sup><</sup>');
+assert.sameValue(String.prototype.sup.call(0x2A), '<sup>42</sup>');
+assert.throws(TypeError, function() {
+  String.prototype.sup.call(undefined);
+});
+assert.throws(TypeError, function() {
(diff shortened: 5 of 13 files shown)


domenic:
I think in general it would be better to use the test262 assertion library than to import the entire V8 assertion library...

bterlson:
Looks good! But, strongly agree with domenic - we don't need two assertion libraries. It should be simple find-and-replace to use what test262 already has I think?

domenic:
Also, do we want `description` fields for each test?

bterlson:
Ahh yep! Good catch, we definitely want description. It's what harnesses display to identify tests.

rwaldron:
Passive displeasure ellipsis aside.

This PR is to act as a staging grounds for all involved to establish exactly what needs to be changed about the V8 tests. Ideally we will determine the broadstroke changes here and can apply them to all that follow.

bterlson:
I'm not sure what you're responding to @rwaldron. I think all I'm saying one of the things that needs to be changed is the assertion library. Another is adding the frontmatter including description and es6id.

smikes:
> Looks good! But, strongly agree with domenic - we don't need two assertion libraries. It should be simple find-and-replace to use what test262 already has I think?

No.  MJS uses deep equality testing by default, while our little `assert` is only shallow.

I vote we take the hit and accept a redundant assertion library, at least for now.  The value of having potentially thousands of well-tested tests from v8 coming into test262 is enormous.

bterlson:
If there were a risk that v8 wouldn't contribute the tests if we didn't take their assertion library as-is I'd agree @smikes, but AFAIK no one has suggested this is the case. Even if there were a risk, I would not be too happy considering we can't just keep accumulating assertion libraries when different vendors want to contribute their tests. So I continue to advocate for the ideal of no assertion library duplication.

Deep equals vs. shallow equals is a good point, though none of the tests here depend on deep equals. We can definitely add a deepEquals assertion helper. I am not against adding new helpers with similiar or identical semantics to existing v8 assertions. It should just be an otherwise useful addition that doesn't make the lives of test authors harder.

smikes:
> no assertion library duplication.

Well, since we don't already have a deep-equality assertion, there's no duplication, right? ;-)  As we get more contributed tests they will be less likely to introduce new assertion code, so the earliest contributors will "win" in terms of having to do the least work to conform to test262

> Deep equals vs. shallow equals is a good point, though none of the tests here depend on deep equals. 

Emphasis on **here**. Since v8's default equality test is a deep one, there will certainly be tests that use deep equality.  In general deciding on the appropriate equality operator is a difficult problem, and if we do not provide a clean mapping of `v8-mjsunit` assertions to test262 assertions, importing their tests will become a process that requires human intervention: slower and more error-prone than .

If it's desirable, I'd be willing to write forwarding functions where possible, so
that `v8-mjsunit` doesn't contain implementation of anything that's already present in `assert`.

rwaldron:
> . I think all I'm saying one of the things that needs to be changed is the assertion library. Another is adding the frontmatter including description and es6id.

Yep, that's the info we're looking for—thanks.

jugglinmike:
Here's a comparison of the assertions available in each test harness. It's a long table, and the formatting is a bit wonky, but at the end I've included some ideas for extensions to the Test262 harness. If that sounds good, I'll update this pull request accordingly and we can continue review from there--let me know!

### Test262-MJSUnit Assertion Equivalency

MSJUnit [1] | MJSUnit Description [1] | Test262 equivalent [2][3]
--------|-------------|-------------------
`assertSame(expected, found, name_opt)` | Expected and found values the same objects, or the same primitive values. For known primitive values, please use assertEquals. | `assert.sameValue(found, expected, name_opt)`
`assertEquals(expected, found, name_opt)` | Expected and found values are identical primitive values or functions or similarly structured objects (checking internal properties of, e.g., Number and Date objects, the elements of arrays and the properties of non-Array objects). | (none)
`assertEqualsDelta(expected, found, delta, name_opt)` | The difference between expected and found value is within certain tolerance. | `assert(found > expected - delta); assert(found < expected + delta)` (may be shorted with `Math.abs` if it's okay to use the STL within assertions)
`assertArrayEquals(expected, found, name_opt)` | The found object is an Array with the same length and elements as the expected object. The expected object doesn't need to be an Array, as long as it's "array-ish". | `assert(compareArray(expected, found), name_opt)`
`assertPropertiesEqual(expected, found, name_opt)` | The found object must have the same enumerable properties as the expected object. The type of object isn't checked. | (none)
`assertToStringEquals(expected, found, name_opt)` | Assert that the string conversion of the found value is equal to the expected string. Only kept for backwards compatability, please check the real structure of the found value. | `assert.sameValue(String(found), expected)` (although we should take the advice of MJSUnit and change the test to assert structure)
`assertTrue(value, name_opt)` | Checks that the found value is true. Use with boolean expressions for tests that doesn't have their own assertXXX function. | `assert(value, name_opt)`
`assertFalse(value, name_opt)` | Checks that the found value is false. | `assert.sameValue(value, false)`
`assertNull(value, name_opt)` | Checks that the found value is null. Kept for historical compatibility, please just use assertEquals(null, expected). | `assert.sameValue(value, null, name_opt))`
`assertNotNull(value, name_opt)` | Checks that the found value is *not* null. | `assert(value !== null, name_opt)`
`assertThrows(fn, ErrCtor)` | Assert that the passed function or eval code throws an exception. The optional second argument is an exception constructor that the thrown exception is checked against with "instanceof". The optional third argument is a message type string that is compared to the type property on the thrown exception. | `assert.throws(ErrCtor, fn)` (note switched argument order)
`assertDoesNotThrow(fn, ErrCtor)` | Assert that the passed function or eval code does not throw an exception. | `assert.throws($ERROR, function() { assert.throws(ErrCtor, fn); })` (relatively concise, but probably still warrants a helper)
`assertInstanceof` | Asserts that the found value is an instance of the constructor passed as the second argument. | (none--MJSUnit does more than a simple `instanceof` check)
`assertUnreachable(name_opt)` | Assert that this code is never executed (i.e., always fails if executed). | `assert(false, name_opt)`
`assertOptimized` | Assert that the function code is (not) optimized.  If "no sync" is passed as second argument, we do not wait for the concurrent optimization thread to finish when polling for optimization status. Only works with --allow-natives-syntax. | n/a -- remove
`assertUnoptimized` | | n/a -- remove

I recommend implemeting the following:

- `assert.isInstance` - for `assertInstanceof`; renamed to avoid confusion with the semantics of the
   native JavaScript operator
- `assert.withinDelta` for `assertEqualsDelta`; this one is a bit of a luxury,
   but it shows up 41 times in V8's ES6 tests.
- `asset.deepEquals` - for `assertEquals`

We can probably ignore:

- `assertDoesNotThrow` - it's not used in V8's ES6 tests
- `assertPropertiesEqual` - we can expand it using `assert.deepEquals`, or we
  can make comparisons manually because of its limited use [4]

[1] https://github.com/v8/v8-git-mirror/blob/8be79b005df2e17a1431f282344e71ec24fc3ff3/test/mjsunit/mjsunit.js
[2] https://github.com/tc39/test262/blob/d075338699cc8eaf123d5f5491f0d68116ee224a/harness/assert.js
[3] https://github.com/tc39/test262/blob/d075338699cc8eaf123d5f5491f0d68116ee224a/harness/compareArray.js
[4] 3 times in the files we're interested in, it looks like:

    $ grep assertProperties test/mjsunit/{harmony,es6} -r
    test/mjsunit/harmony/debug-blockscopes.js:      assertPropertiesEqual(global_object, scope.scopeObject().value());
    test/mjsunit/es6/generators-debug-scopes.js:  assertPropertiesEqual(scope1.scopeObject().value(), scope2.scopeObject().value());
    test/mjsunit/es6/generators-debug-scopes.js:      assertPropertiesEqual(this, scope.scopeObject().value());

bterlson:
@jugglinmike I like your recommendations. Unless anyone else (@domenic?) has reservations I'm alright adding isInstance, withinDelta, and deepEquals.

I have a slight concern with withinDelta, which is that [historically](*https://code.google.com/p/v8/issues/detail?id=3006) v8 has had poor precision for trig functions. If there are cases where the delta specified could be tighter, we should take the opportunity to tighten it assuming it passes on the major implementations. The ECMAScript spec says nothing about how precise Math functions have to be but clearly there are issues with lower precision so it makes sense to me to test for the most precision we can get away with.

domenic:
I think it's worth only adding those as-needed and making sure they're really needed each time. For example if isInstance is really doing something magical, maybe those tests need to be less complicated. And deepEquals can often be replaced by one-level strict-equals comparison instead of V8's sloppy and multi-level algorithm (if I recall how it works correctly).

Also `assertDoesNotThrow` should not be translated that way---just leave the code there, and if it throws the test will fail.

Also I am pretty sure we have `assert.notSameValue` which can make some of those translations nicer.

jugglinmike:
Thanks for the feedback! I've translated the MJSUnit assertions to their Test262 equivalents (no new assertions are required by this set). I've also added descriptions. Having no high-level description to copy from the spec, I had to get a little creative here, and I'm not sure if what I came up with is appropriate or not; let me know!

rwaldron:
```
$ test262 --command "v8" annexB
annexB/B.2.1 passed in non-strict mode
annexB/B.2.1.propertyCheck passed in non-strict mode
annexB/B.2.2 passed in non-strict mode
annexB/B.2.2.propertyCheck passed in non-strict mode
annexB/B.2.3.10 passed in non-strict mode
annexB/B.2.3.11 passed in non-strict mode
annexB/B.2.3.12 passed in non-strict mode
annexB/B.2.3.13 passed in non-strict mode
annexB/B.2.3.14 passed in non-strict mode
annexB/B.2.3.2 passed in non-strict mode
annexB/B.2.3.3 passed in non-strict mode
annexB/B.2.3.4 passed in non-strict mode
annexB/B.2.3.5 passed in non-strict mode
annexB/B.2.3.6 passed in non-strict mode
annexB/B.2.3.7 passed in non-strict mode
annexB/B.2.3.8 passed in non-strict mode
annexB/B.2.3.9 passed in non-strict mode
annexB/B.2.3 passed in non-strict mode
annexB/B.2.4 passed in non-strict mode
annexB/B.2.4.propertyCheck passed in non-strict mode
annexB/B.2.5 passed in non-strict mode
annexB/B.2.5.propertyCheck passed in non-strict mode
annexB/B.2.6 passed in non-strict mode
annexB/B.2.6.propertyCheck passed in non-strict mode
annexB/B.RegExp.prototype.compile passed in non-strict mode
```

rwaldron:
> I think it's worth only adding those as-needed and making sure they're really needed each time.

We'll be submitting `assert.deepEquals` as its own PR, that will block a forthcoming set of tests that rely on a deep equality comparison.

jugglinmike:
> > I think it's worth only adding those as-needed and making sure they're really needed each time.
>
> We'll be submitting `assert.deepEquals` as its own PR, that will block a forthcoming set of tests that rely on a deep equality comparison.


See gh-153

bterlson:
I think this one is good to go right?

But do we need the .v8 license file? It's BSD, so the same license as test262 itself. The copyright owner is indicated in the copyright header of the tests.

bterlson:
Also generally I'd prefer squashed PRs for cleanliness (although @domenic will laugh at me at this point because merge commits still abound regardless of how clean the PR is :-P)

rwaldron:
@bterlson we also prefer squashed commits, but the squash comes after the final approval :) 

For this and all future PRs, once you give the "LGTM", we'll squash and I can assist you with landing.

jugglinmike:
> But do we need the .v8 license file? It's BSD, so the same license as test262 itself. The copyright owner is indicated in the copyright header of the tests.

I was a little unsure about the exact terms of these contributions, but it looks like it's unnecessary. I've removed the V8 license file and updated the attribution boilerplate as appropriate. Since this is another change, I'd like to hold off on squashing until I get a final approval from you

bterlson:
LGTM!

jugglinmike:
Thanks Brian--I've rebased, squashed, and updated the commit message accordingly.

Issue #150, opened 2015-02-13
"Test Case Names" in CONTRIBUTING.md appears out of date


rwaldron:
https://github.com/tc39/test262/blob/master/CONTRIBUTING.md

This is the current content, but I don't yet feel comfortable re-writing this:

```
Test Case Names

These are guidelines. There is a substantial amount of variation in existing test names.

Test cases are often named by the portion of the spec or draft spec they reference. A test for ECMAScript 5.1 section 8.7.2 PutValue should be placed in the directory test262/suite/ch08/8.7/8.7.2/ and have a filename which contains the substring 8.7.2. A test for ECMAScript 6.0 draft section 25.4.4.1 Promise.all should be placed in directory test262/suite/es6/ch25/25.4/25.4.4/25.4.4.1 and have a filename that contains the substring 25.4.4.1.

One common convention is to break down a section into a number of testable assertions (A) and to test each assertion with one or more test cases (T). For example, section 25.4.4.1 of ES6 could be broken down like this:

A1 Promise.all is a function of one argument
A1.1 Promise.all is callable
A1.2 Promise.all expects 1 argument
A2 Promise.all([]) is a Promise immediately resolved with []
A2.1 Promise.all([]) returns a Promise
A2.2 Promise.all([]) is resolved immediately
A2.3 Promise.all([]) is resolved with a new empty array
with test case S25.4.4.1_A1.1_T1.js being the first test for assertion A1.1, and S25.4.4.1_A2.3_T3.js being the third test for assertion A2.3.
```

smikes:
You are correct, this portion of the doc outdated.

rwaldron:
I can take a crack at re-wording

Issue #136, opened 2014-12-10
CONTRIBUTING.md


mathiasbynens:
<https://github.com/tc39/test262/blob/master/CONTRIBUTING.md#test-case-names> still says:

> Test cases are often named by the portion of the spec or draft spec they reference.  A test for ECMAScript 5.1 section [8.7.2 PutValue](http://www.ecma-international.org/ecma-262/5.1/#sec-8.7.2) should be placed in the directory `test262/suite/ch08/8.7/8.7.2/` and have a filename which contains the substring `8.7.2`.  A test for ECMAScript 6.0 draft section [25.4.4.1 Promise.all](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-promise.all) should be placed in directory `test262/suite/es6/ch25/25.4/25.4.4/25.4.4.1` and have a filename that contains the substring `25.4.4.1`.

This doesn’t seem to be true anymore. Can the contribution guide be updated?

smikes:
This is fixed in pending PR #151 -- @mathiasbynens , is that right?

mathiasbynens:
Yep.

PR #151, opened 2015-02-13
Contributing: updates the Test Case Names section. Closes gh-150

rwaldron:
Initial draft of test case naming re-write

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,21 +2,16 @@
 
 ## Test Case Names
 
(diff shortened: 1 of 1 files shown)


smikes:
Also closes #136, which dropped off my radar.

rwaldron:
I missed that, but will update the commit message to include it.

rwaldron:
Updated. @bterlson if this is good to go, I will land

bterlson:
Good to go, thanks! I'd press the merge button now but it sounds like you'd like to do that ;)

PR #152, opened 2015-02-14
fix undefined var

smikes:
Ran into this today, probably just an omission when the assert function's variable was renamed.


rwaldron:
Looks pretty safe to land this ;)

PR #153, opened 2015-02-16
Add `assert.deepEquals`

jugglinmike:
Source: v8/test/mjsunit/mjsunit.js

### DIFF
--- a/harness/assert.js
+++ b/harness/assert.js
@@ -88,7 +88,7 @@ assert.throws = function (expectedErrorConstructor, func) {
   }
 
   function deepEquals(a, b) {
-    if (asset._isSameValue(a, b)) {
+    if (assert._isSameValue(a, b)) {
       return true;
     }
     if (typeof a !== typeof b) return false;
--- a/harness/assert.js
+++ b/harness/assert.js
@@ -118,6 +118,7 @@ assert.throws = function (expectedErrorConstructor, func) {
     }
     return deepObjectEquals(a, b);
   }
+
   assert.deepEquals = function (actual, expected, message) {
     assert(deepEquals(actual, expected), message);
   };
--- a/harness/assert.js
+++ b/harness/assert.js
@@ -88,13 +88,10 @@ assert.throws = function (expectedErrorConstructor, func) {
   }
 
   function deepEquals(a, b) {
-    if (a === b) {
-      // Check for -0.
-      if (a === 0) return (1 / a) === (1 / b);
+    if (asset._isSameValue(a, b)) {
       return true;
     }
     if (typeof a != typeof b) return false;
-    if (typeof a == "number") return isNaN(a) && isNaN(b);
     if (typeof a !== "object" && typeof a !== "function") return false;
     // Neither a nor b is primitive.
     var objectClass = classOf(a);
--- a/harness/assert.js
+++ b/harness/assert.js
@@ -91,7 +91,7 @@ assert.throws = function (expectedErrorConstructor, func) {
     if (asset._isSameValue(a, b)) {
       return true;
     }
-    if (typeof a != typeof b) return false;
+    if (typeof a !== typeof b) return false;
     if (typeof a !== "object" && typeof a !== "function") return false;
     // Neither a nor b is primitive.
     var objectClass = classOf(a);
@@ -104,7 +104,7 @@ assert.throws = function (expectedErrorConstructor, func) {
     if (objectClass === "Function") return false;
     if (objectClass === "Array") {
       var elementCount = 0;
-      if (a.length != b.length) {
+      if (a.length !== b.length) {
         return false;
       }
       for (var i = 0; i < a.length; i++) {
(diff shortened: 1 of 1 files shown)


harness/assert.js
```diff
@@ -62,3 +62,66 @@ assert.throws = function (expectedErrorConstructor, func) {
...
+    return true;
+  }
+
+  function deepEquals(a, b) {
+    if (a === b) {
+      // Check for -0.
+      if (a === 0) return (1 / a) === (1 / b);
+      return true;
+    }
+    if (typeof a != typeof b) return false;
```
bterlson: Strict equality checks preferred throughout

bterlson:
It seems like, for primitive types (perhaps `if typeof a !== 'object' && typeof a !== 'function'`) you could call the _isSameValue helper? There is some duplication here (notably the handling of -0). Thoughts?

jugglinmike:
@bterlson That makes a lot of sense. I've incorporated your suggestions.

harness/assert.js
```diff
@@ -62,3 +62,64 @@ assert.throws = function (expectedErrorConstructor, func) {
...
+(function() {
+  function classOf(object) {
+    // Argument must not be null or undefined.
+    var string = Object.prototype.toString.call(object);
+    // String has format [object <ClassName>].
+    return string.substring(8, string.length - 1);
+  }
+
+  function deepObjectEquals(a, b) {
+    var aProps = Object.keys(a);
```
jugglinmike: @bterlson I think it would be better to explicitly build this array (and the one following it) with a for..in loop and "own" property check. Do you agree?

bterlson: Hmm yeah, good call I think. Object.keys is an ES5 method we shouldn't depend on unless absolutely necessary. Doesn't seem necessary here, as gross as for-in enumeration is.

domenic:
This is a *lot* of complexity to put in a language-level assertion framework. Are we sure we can't find simpler primitives like (sort-less) `compareArray`?

If you look at all the language semantics this depends on, I count:

- `Object.prototype.toString` / `@@toStringTag`
- `String.prototype.substring`
- `Object.hasOwnProperty`
- Since the above was `Object.hasOwnProperty` and not `Object.prototype.hasOwnProperty`, implicitly the link between `Object.prototype` and functions like `Object`.
- `Function.prototype.call`
- `Array.prototype.sort`
- `Array.prototype.push`
- `typeof`
- `RegExp.prototype.toString`
- `{String, Number, Boolean, Date}.prototype.valueOf`
- Plus of course basics like array indexing, property access, function calls, etc. (which are mostly unavoidable I'd imagine)

This is basically an integration test of many built-in methods wrapped up into one.

Notably we'd have to be sure that this function is never used in testing any of the above things.

rwaldron:
@domenic if this is truly a problem, then why is V8 relying on such a mechanism for its own "language-level assertion framework"? Right now I'm completely blocked by this because V8 implementors abuse the "deep equal" behaviour of "assertEquals" quite freely. As much as we can, we're replacing `assertEquals` with `assert.sameValue` wherever appropriate, but the `array-*` tests that I've moved in other PRs have many "deep equals" tests—and I'm not sure "rewriting everything from scratch" falls under "moving mjsunit tests".

domenic:
I think test262 and V8 generally have different standards for how rigorous their tests are.

bterlson:
@rwaldron I think it's worth getting more details on just what the conversion process would be like if we had a simpler assertion library (which I agree with @domenic would be vastly preferrable). It seems like it might not be too difficult to replace assert.equals with multiple assertions in the test itself to verify the various things. It wouldnt' be a rewrite, just replacing one assertion with many assertions. Can we see more details on whether something like this is workable? I honestly have no idea, but I'm curious to see more details and am surprised the only alternative is to "rewrite everything from scratch".

rwaldron:
@bterlson 

> replace assert.equals with multiple assertions

Is actually what I was thinking of when I wrote: 

> rewrite everything from scratch

I'll get back to you when we've had more time to discuss what to do now given that any notion of automation is out the window.

bterlson:
Why is automation out the window, though? That's the part I'm confused about. It seems like we could apply the same transformation to each assert.equals usage, in which case automation could still be used...

rwaldron:
> It seems like we could apply the same transformation to each assert.equals usage

I'm completely lost now. This is literally what we want to do, but are being told that `assert.deepEquals` is too complex. What transformation are you talking about? 

Right now we want to do this: `assertEquals` => `assert.deepEquals` (or `sameValue` in cases that are appropriate, but we'll figure that out separately), because that matches the existing V8 tests.

bterlson:
assert.deepEquals, as it is implemented, currently tests many things and requires much of the language to be implemented to begin running the tests. What I'm suggesting is that, assuming the semantics the V8 helper validates are useful to validate (which I think is safe to assume in some cases, though perhaps @domenic disagrees), I wouldn't have a problem explicitly testing them in the test body rather than depending on a helper. Basically, explode one assertEquals call into multiple assert.* calls against the arguments passed to assertEquals. Does this make more sense? I'm not saying I'm sure it's possible, but it seems worth considering :)

smikes:
For reference, I did something like this (expanding a fn call to multiple asserts) to kill the `dataPropertyAttributesAreCorrect` helpers (see https://github.com/tc39/test262/pull/116 ).  

I was able to apply a lot of automation, mainly with some ugly perl scripts (I think that's redundant though...)

rwaldron:
>  Basically, explode one assertEquals call into multiple assert.* calls against the arguments passed to assertEquals. Does this make more sense? I'm not saying I'm sure it's possible, but it seems worth considering :)

I need more time to think about this, but I'm worried that changes like this will lose something in-flight.

jugglinmike:
Sounds like this assertion is too complex for what is ultimately a convenience. Thanks for the feedback, everyone!

PR #154, opened 2015-02-16
Import tests from Google V8 (Array isConcatSpreadable)

rwaldron:
project:

	test/mjsunit/harmony/array-concat.js

Blocked by https://github.com/tc39/test262/pull/153

### DIFF
--- a/test/built-ins/Array/prototype/concat/Array.prototype.concat_array-like-string-length.js
+++ b/test/built-ins/Array/prototype/concat/Array.prototype.concat_array-like-string-length.js
@@ -1,0 +1,28 @@
+// Copyright (c) 2014 the V8 project authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+
+/*---
+es6id: 22.1.3.1_3
+description: Array.prototype.concat array like string length
+includes: [compareArray.js]
+---*/
+var obj = {
+  "length": "6",
+  "1": "A",
+  "3": "B",
+  "5": "C"
+};
+obj[Symbol.isConcatSpreadable] = true;
+var obj2 = { length: 3, "0": "0", "1": "1", "2": "2" };
+var arr = ["X", "Y", "Z"];
+
+var expected = [
+  void 0, "A", void 0, "B", void 0, "C",
+  obj2,
+  "X", "Y", "Z"
+];
+var actual = Array.prototype.concat.call(obj, obj2, arr);
+
+assert(compareArray(actual, expected));
--- a/test/built-ins/Array/prototype/concat/Array.prototype.concat_array-like.js
+++ b/test/built-ins/Array/prototype/concat/Array.prototype.concat_array-like.js
@@ -1,0 +1,28 @@
+// Copyright (c) 2014 the V8 project authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+
+/*---
+es6id: 22.1.3.1_3
+description: Array.prototype.concat array like
+includes: [compareArray.js]
+---*/
+var obj = {
+  "length": 6,
+  "1": "A",
+  "3": "B",
+  "5": "C"
+};
+obj[Symbol.isConcatSpreadable] = true;
+var obj2 = { length: 3, "0": "0", "1": "1", "2": "2" };
+var arr = ["X", "Y", "Z"];
+
+var expected = [
+  void 0, "A", void 0, "B", void 0, "C",
+  obj2,
+  "X", "Y", "Z"
+];
+var actual = Array.prototype.concat.call(obj, obj2, arr);
+
+assert(compareArray(actual, expected));
--- a/test/built-ins/Array/prototype/concat/Array.prototype.concat_arity.js
+++ b/test/built-ins/Array/prototype/concat/Array.prototype.concat_arity.js
@@ -1,0 +1,9 @@
+// Copyright (c) 2014 the V8 project authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+/*---
+es6id: 22.1.3.1_3
+description: Array.prototype.concat arity
+---*/
+assert.sameValue(Array.prototype.concat.length, 1);
--- a/test/built-ins/Array/prototype/concat/Array.prototype.concat_array-like-length-to-string-throws.js
+++ b/test/built-ins/Array/prototype/concat/Array.prototype.concat_array-like-length-to-string-throws.js
@@ -1,0 +1,22 @@
+// Copyright (c) 2014 the V8 project authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+/*---
+es6id: 22.1.3.1_3
+description: Array.prototype.concat array like length to string throws
+---*/
+function MyError() {}
+var obj = {
+  "length": { toString: function() {
+      throw new MyError();
+    }, valueOf: null
+  },
+  "1": "A",
+  "3": "B",
+  "5": "C"
+};
+obj[Symbol.isConcatSpreadable] = true;
+assert.throws(MyError, function() {
+  [].concat(obj);
+});
--- a/test/built-ins/Array/prototype/concat/Array.prototype.concat_array-like-length-value-of-throws.js
+++ b/test/built-ins/Array/prototype/concat/Array.prototype.concat_array-like-length-value-of-throws.js
@@ -1,0 +1,25 @@
+// Copyright (c) 2014 the V8 project authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+
+/*---
+es6id: 22.1.3.1_3
+description: Array.prototype.concat array like length valueOf throws
+---*/
+function MyError() {}
+var obj = {
+  "length": {
+    valueOf: function() {
+      throw new MyError();
+    },
+    toString: null
+  },
+  "1": "A",
+  "3": "B",
+  "5": "C"
+};
+obj[Symbol.isConcatSpreadable] = true;
+assert.throws(MyError, function() {
+  [].concat(obj);
+});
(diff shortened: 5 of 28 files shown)


test/built-ins/Array/prototype/concat/concat-array-likes-1.js
```diff
@@ -0,0 +1,677 @@
...
+
+/*---
+es6id: 22.1.3.1_3
+description: >
+    Array.prototype.concat: Symbol.isConcatSpreadable & array-likes
+---*/
+
+(function testArrayConcatArity() {
+  "use strict";
+  assert.deepEquals(1, Array.prototype.concat.length);
```
domenic: Why does this use deepEquals? Same for many other places in this file.

Also, it looks like expected and actual are switched throughout the file.

test/built-ins/Array/prototype/concat/concat-array-likes-1.js
```diff
@@ -0,0 +1,677 @@
...
+})();
+
+
+(function testArrayConcatNoPrototype() {
```
domenic: Up to @bterlson but to me each of these IIFEs look like a separate test (i.e. separate test file) that can fail or succeed independently.

test/built-ins/Array/prototype/concat/concat-array-likes-1.js
```diff
@@ -0,0 +1,677 @@
...
+  "use strict";
+  assert.deepEquals(void 0, Array.prototype.concat.prototype);
+})();
+
+
+(function testArrayConcatDescriptor() {
+  "use strict";
```
domenic: Enforcing "use strict" here seems bad since the test runner is usually used to signal that a test can run in both strict and non-strict. I doubt many of these depend on strict mode.

test/built-ins/Array/prototype/concat/concat-array-likes-1.js
```diff
@@ -0,0 +1,677 @@
...
+
+  assert.throws(MyError, function() {
+    Array.prototype.concat.call(obj, 1, 2, 3);
+  });
+})();
+
+
+(function testConcatArraySubclass() {
+  "use strict";
+  // TODO(caitp): when concat is called on instances of classes which extend
```
domenic: Probably didn't mean to port this here?

test/built-ins/Array/prototype/concat/concat-array-likes-1.js
```diff
@@ -0,0 +1,677 @@
...
+    Float32Array,
+    Float64Array
+  ].forEach(function(ctor, i) {
+    testConcatTypedArray(ctor, 1, max[i]);
+  });
+})();
+
+
+(function testConcatLargeTypedArray() {
+  var max = [2^8, 2^16, 2^32, false, false];
```
domenic: Are these tests correct? `2^32 === 34`. I think they probably wanted `4294967296`?

test/built-ins/Array/prototype/concat/concat-array-likes-1.js
rwaldron: Only because I missed it when I had to manually transform all of the uses of `assertEquals`, since it's used so liberally.

rwaldron: Actually, this was from last week, before we discovered all of the unnecessary and ambiguous uses of `assertEquals` and had a transformer that automatically changed all `assertEquals` to `assert.deepEquals` (as illustrated in the table [here](https://github.com/tc39/test262/pull/149#issuecomment-73975846)).

test/built-ins/Array/prototype/concat/concat-array-likes-1.js
rwaldron: Seems like something that should be done on the V8 side.

test/built-ins/Array/prototype/concat/concat-array-likes-1.js
rwaldron: https://code.google.com/p/v8/issues/entry

test/built-ins/Array/prototype/concat/concat-array-likes-1.js
rwaldron: There is no way to efficiently (ie. not manually reviewing, line-by-line) and unambiguously (even if done manually??) remove comments from every file.

test/built-ins/Array/prototype/concat/concat-array-likes-1.js
bterlson: Let's remove use strict. If strictness is required then set the frontmatter. Can perhaps do a blanked removal of all prologues during the conversion process.

bterlson:
Can we split the 666-line test into individual tests? Every IIFE should be its own file. The experience of using these tests with a new implementation will be extremely poor as only one assertion will be reported per file...

domenic:
Expected and actual are switched in all assertions.

test/built-ins/Array/prototype/concat/concat-array-likes-1.js
caitp: We were using explicit strict-mode in order to be able to test concat-spreadable Arguments objects, and in V8 there are three flavours of Arguments objects.

caitp: you probably don't need to test "three flavours" in test 262, but for V8 I think we want to make sure it works correctly with the "slow" sloppy arguments object

bterlson: For test262, tests that don't specify onlyStrict or noStrict are expected to work in both modes. Not sure what the other flavor is but strict and non-strict arguments should be tested if you simply omit the "use strict" prologue and let the harness control strictness of the test.

caitp: > Not sure what the other flavor is

It basically just means "make sure it works correctly when there are duplicate formal parameters in sloppy mode", although I don't think there's any script-observable difference between sloppy + duplicates and sloppy without duplicates.

rwaldron:
TODO: `compareArray()` => `assert(compareArray())`

test/built-ins/Array/prototype/concat/array-concat-arity.js
```diff
@@ -0,0 +1,11 @@
+// Copyright 2014 the V8 project authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+// Flags: --harmony-arrays --harmony-classes
```
domenic: Does test262 want to keep these flags comments?

test/built-ins/Array/prototype/concat/array-concat-arity.js
```diff
@@ -0,0 +1,11 @@
+// Copyright 2014 the V8 project authors. All rights reserved.
```
domenic: Slightly different copyright than found in https://github.com/tc39/test262/blob/master/CONTRIBUTING.md#test-case-style

test/built-ins/Array/prototype/concat/array-concat-arity.js
smikes: Or should they be converted to "features" front matter, or are we dropping that idea @bterlson ?

rwaldron:
Squashed & Rebased.

test/built-ins/Array/prototype/concat/Array.prototype.concat_array-like-string-length.js
```diff
@@ -0,0 +1,29 @@
...
+  "3": "B",
+  "5": "C"
+};
+obj[Symbol.isConcatSpreadable] = true;
+var obj2 = { length: 3, "0": "0", "1": "1", "2": "2" };
+var arr = ["X", "Y", "Z"];
+
+var expected = [
+  void 0, "A", void 0, "B", void 0, "C",
+  { "length": 3, "0": "0", "1": "1", "2": "2" },
```
anba: This line needs to be replaced with `obj2`.

test/built-ins/Array/prototype/concat/Array.prototype.concat_array-like.js
```diff
@@ -0,0 +1,29 @@
...
+  "3": "B",
+  "5": "C"
+};
+obj[Symbol.isConcatSpreadable] = true;
+var obj2 = { length: 3, "0": "0", "1": "1", "2": "2" };
+var arr = ["X", "Y", "Z"];
+
+var expected = [
+  void 0, "A", void 0, "B", void 0, "C",
+  { "length": 3, "0": "0", "1": "1", "2": "2" },
```
anba: This line needs to be replaced with `obj`.

test/built-ins/Array/prototype/concat/Array.prototype.concat_es5.js
```diff
@@ -0,0 +1,236 @@
...
+includes: [compareArray.js]
+---*/
+var poses;
+var pos;
+
+poses = [140, 4000000000];
+while (pos = poses.shift()) {
+  var a = new Array(pos);
+  var array_proto = [];
+  a.__proto__ = array_proto;
```
anba: `__proto__` is AnnexB

test/built-ins/Array/prototype/concat/Array.prototype.concat_array-like-string-length.js
rwaldron: https://github.com/v8/v8-git-mirror/blob/c11c0bdbeda573640400cd01ff9f9cbcf33390df/test/mjsunit/harmony/array-concat.js#L43-L57

test/built-ins/Array/prototype/concat/Array.prototype.concat_array-like.js
rwaldron: https://github.com/v8/v8-git-mirror/blob/c11c0bdbeda573640400cd01ff9f9cbcf33390df/test/mjsunit/harmony/array-concat.js#L26-L40

test/built-ins/Array/prototype/concat/Array.prototype.concat_array-like-string-length.js
caitp: there's no reason it couldn't be replaced with obj2, tbh --- it should be a reference to the same object. Don't ask me why I didn't write it that way o_O

rwaldron: okey doke.

PR #165, opened 2015-02-17
Add Array Iteration Tests

jugglinmike:
project:

test/mjsunit/es6/array-iterator.js

### DIFF
--- a/test/built-ins/Array/prototype/Symbol.iterator/property-descriptor.js
+++ b/test/built-ins/Array/prototype/Symbol.iterator/property-descriptor.js
@@ -1,0 +1,14 @@
+// Copyright (C) 2013 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+  description: >
+      The method should exist on the Array prototype, and it should be writable
+      and configurable, but not enumerable.
+  includes: [propertyHelper.js]
+  es6id: 22.1.3.4
+ ---*/
+
+verifyNotEnumerable(Array.prototype, Symbol.iterator);
+verifyWritable(Array.prototype, Symbol.iterator);
+verifyConfigurable(Array.prototype, Symbol.iterator);
--- a/test/built-ins/Array/prototype/Symbol.iterator/prototype/Symbol.toStringTag/property-descriptor.js
+++ b/test/built-ins/Array/prototype/Symbol.iterator/prototype/Symbol.toStringTag/property-descriptor.js
@@ -1,0 +1,18 @@
+// Copyright (C) 2013 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+  description: >
+      `Object.prototype.getOwnPropertyDescriptor` should reflect the value and
+      writability of the @@toStringTag attribute.
+  includes: [propertyHelper.js]
+  es6id: 22.1.5.2.2
+ ---*/
+
+var ArrayIteratorProto = Object.getPrototypeOf([][Symbol.iterator]());
+
+assert.sameValue("Array Iterator", ArrayIteratorProto[Symbol.toStringTag]);
+
+verifyNotEnumerable(ArrayIteratorProto, Symbol.toStringTag);
+verifyNotWritable(ArrayIteratorProto, Symbol.toStringTag);
+verifyConfigurable(ArrayIteratorProto, Symbol.toStringTag);
--- a/test/built-ins/Array/prototype/Symbol.iterator/prototype/Symbol.toStringTag/value-direct.js
+++ b/test/built-ins/Array/prototype/Symbol.iterator/prototype/Symbol.toStringTag/value-direct.js
@@ -1,0 +1,12 @@
+// Copyright (C) 2013 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+  description: >
+      The @@toStringTag attribute should be defined directly on the prototype.
+  es6id: 22.1.5.2.2
+ ---*/
+
+var ArrayIteratorProto = Object.getPrototypeOf([][Symbol.iterator]());
+
+assert.sameValue("Array Iterator", ArrayIteratorProto[Symbol.toStringTag]);
--- a/test/built-ins/Array/prototype/Symbol.iterator/prototype/Symbol.toStringTag/value-from-to-string.js
+++ b/test/built-ins/Array/prototype/Symbol.iterator/prototype/Symbol.toStringTag/value-from-to-string.js
@@ -1,0 +1,13 @@
+// Copyright (C) 2013 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+  description: >
+      `Object.prototype.toString` should honor the value of the @@toStringTag
+      attribute.
+  es6id: 22.1.5.2.2
+ ---*/
+
+var iter = [][Symbol.iterator]();
+
+assert.sameValue("[object Array Iterator]", Object.prototype.toString.call(iter));
--- a/test/built-ins/Array/prototype/Symbol.iterator/prototype/next/iteration-mutable.js
+++ b/test/built-ins/Array/prototype/Symbol.iterator/prototype/next/iteration-mutable.js
@@ -1,0 +1,32 @@
+// Copyright (C) 2013 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+  description: >
+      When an item is added to the array after the iterator is created but
+      before the iterator is "done" (as defined by 22.1.5.2.1), the new item
+      should be accessible via iteration. When an item is added to the array
+      after the iterator is is "done", the new item should not be accessible
+      via iteration.
+  es6id: 22.1.3.30
+ ---*/
+
+var array = [];
+var iterator = array[Symbol.iterator]();
+var result;
+
+array.push('a');
+
+result = iterator.next();
+assert.sameValue(result.done, false, 'First result `done` flag');
+assert.sameValue(result.value, 'a', 'First result `value`');
+
+result = iterator.next();
+assert.sameValue(result.done, true, 'Exhausted result `done` flag');
+assert.sameValue(result.value, undefined, 'Exhausted result `value`');
+
+array.push('b');
+
+result = iterator.next();
+assert.sameValue(result.done, true, 'Exhausted result `done` flag (after push)');
+assert.sameValue(result.value, undefined, 'Exhausted result `value (after push)');
--- a/test/built-ins/Array/prototype/Symbol.iterator/prototype/next/iteration.js
+++ b/test/built-ins/Array/prototype/Symbol.iterator/prototype/next/iteration.js
@@ -1,0 +1,29 @@
+// Copyright (C) 2013 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+  description: >
+      The method should return a valid iterator with the context as the
+      IteratedObject.
+  es6id: 22.1.3.30
+ ---*/
+
+var array = ['a', 'b', 'c'];
+var iterator = array[Symbol.iterator]();
+var result;
+
+result = iterator.next();
+assert.sameValue(result.value, 'a', 'First result `value`');
+assert.sameValue(result.done, false, 'First result `done` flag');
+
+result = iterator.next();
+assert.sameValue(result.value, 'b', 'Second result `value`');
+assert.sameValue(result.done, false, 'Second result `done` flag');
+
+result = iterator.next();
+assert.sameValue(result.value, 'c', 'Third result `value`');
+assert.sameValue(result.done, false, 'Third result `done` flag`');
+
+result = iterator.next();
+assert.sameValue(result.value, undefined, 'Exhausted result `value`');
+assert.sameValue(result.done, true, 'Exhausted result `done` flag');
--- a/test/built-ins/Array/prototype/Symbol.iterator/prototype/next/non-own-slots.js
+++ b/test/built-ins/Array/prototype/Symbol.iterator/prototype/next/non-own-slots.js
@@ -1,0 +1,17 @@
+// Copyright (C) 2013 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+  description: >
+      If the `this` value does not have all of the internal slots of an Array
+      Iterator Instance (22.1.5.3), throw a TypeError exception.
+  es6id: 22.1.5.2.1
+ ---*/
+
+var array = [0];
+var iterator = array[Symbol.iterator]();
+var object = Object.create(iterator);
+
+assert.throws(TypeError, function() {
+  object.next();
+});
--- a/test/built-ins/Array/prototype/Symbol.iterator/prototype/next/property-descriptor.js
+++ b/test/built-ins/Array/prototype/Symbol.iterator/prototype/next/property-descriptor.js
@@ -1,0 +1,16 @@
+// Copyright (C) 2013 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+  description: >
+      The method should exist on the ArrayIterator prototype, and it should be
+      writable and configurable, but not enumerable.
+  includes: [propertyHelper.js]
+  es6id: 17
+ ---*/
+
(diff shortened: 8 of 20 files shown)


rwaldron:
```
$ for name in $(git diff master --name-only | sed "s/^.*\///" | tr '\n' ' '); do test262 --command "v8 --harmony-tostring" $name; done

built-ins/Array/prototype/Symbol.iterator/iteration-mutable passed in non-strict mode
built-ins/Array/prototype/entries/iteration-mutable passed in non-strict mode
built-ins/Array/prototype/keys/iteration-mutable passed in non-strict mode

built-ins/Array/prototype/Symbol.iterator/iteration passed in non-strict mode
built-ins/Array/prototype/entries/iteration passed in non-strict mode
built-ins/Array/prototype/keys/iteration passed in non-strict mode

built-ins/Array/prototype/Symbol.iterator/non-enumerable passed in non-strict mode
built-ins/Array/prototype/entries/non-enumerable passed in non-strict mode
built-ins/Array/prototype/keys/non-enumerable passed in non-strict mode
built-ins/ArrayIterator/prototype/next/non-enumerable passed in non-strict mode

built-ins/Array/prototype/Symbol.iterator/iteration-mutable passed in non-strict mode
built-ins/Array/prototype/entries/iteration-mutable passed in non-strict mode
built-ins/Array/prototype/keys/iteration-mutable passed in non-strict mode

built-ins/Array/prototype/Symbol.iterator/iteration passed in non-strict mode
built-ins/Array/prototype/entries/iteration passed in non-strict mode
built-ins/Array/prototype/keys/iteration passed in non-strict mode

built-ins/Array/prototype/Symbol.iterator/non-enumerable passed in non-strict mode
built-ins/Array/prototype/entries/non-enumerable passed in non-strict mode
built-ins/Array/prototype/keys/non-enumerable passed in non-strict mode
built-ins/ArrayIterator/prototype/next/non-enumerable passed in non-strict mode

built-ins/Array/prototype/entries/returns-iterator passed in non-strict mode
built-ins/Array/prototype/keys/returns-iterator passed in non-strict mode

built-ins/Array/prototype/Symbol.iterator/iteration-mutable passed in non-strict mode
built-ins/Array/prototype/entries/iteration-mutable passed in non-strict mode
built-ins/Array/prototype/keys/iteration-mutable passed in non-strict mode

built-ins/Array/prototype/Symbol.iterator/iteration passed in non-strict mode
built-ins/Array/prototype/entries/iteration passed in non-strict mode
built-ins/Array/prototype/keys/iteration passed in non-strict mode

built-ins/Array/prototype/Symbol.iterator/non-enumerable passed in non-strict mode
built-ins/Array/prototype/entries/non-enumerable passed in non-strict mode
built-ins/Array/prototype/keys/non-enumerable passed in non-strict mode
built-ins/ArrayIterator/prototype/next/non-enumerable passed in non-strict mode

built-ins/Array/prototype/entries/returns-iterator passed in non-strict mode
built-ins/Array/prototype/keys/returns-iterator passed in non-strict mode

built-ins/ArrayIterator/prototype/Symbol.iterator passed in non-strict mode
language/for-of/Array.prototype.Symbol.iterator passed in non-strict mode

built-ins/ArrayIterator/prototype/Symbol.toStringTag/property-descriptor passed in non-strict mode

built-ins/ArrayIterator/prototype/Symbol.toStringTag/value-direct passed in non-strict mode

built-ins/ArrayIterator/prototype/Symbol.toStringTag/value-from-to-string passed in non-strict mode

built-ins/Array/prototype/Symbol.iterator/non-enumerable passed in non-strict mode
built-ins/Array/prototype/entries/non-enumerable passed in non-strict mode
built-ins/Array/prototype/keys/non-enumerable passed in non-strict mode
built-ins/ArrayIterator/prototype/next/non-enumerable passed in non-strict mode

built-ins/ArrayIterator/prototype/next/non-own-slots passed in non-strict mode

built-ins/ArrayIterator/prototype/prototype passed in non-strict mode

language/for-of/Array.prototype.Symbol.iterator passed in non-strict mode

language/for-of/Array.prototype.entries passed in non-strict mode

language/for-of/Array.prototype.keys passed in non-strict mode

language/for-of/array passed in non-strict mode
```

anba:
Please change tests with `__proto__` to use either `Object.getPrototypeOf` or `Object.create`, so the main test262 suite does not depend on Annex B extensions.

And `test/built-ins/ArrayIterator/prototype/Symbol.iterator.js` is not correct. It should be:
```javascript
var ArrayIteratorProto = Object.getPrototypeOf([][Symbol.iterator]());
assert.sameValue(ArrayIteratorProto.hasOwnProperty(Symbol.iterator), false);

var desc = Object.getOwnPropertyDescriptor(ArrayIteratorProto, Symbol.iterator);
assert.sameValue(desc, void 0);
```

jugglinmike:
@anba Thanks for the feedback! Regarding `test/built-ins/ArrayIterator/prototype/Symbol.iterator.js`: I've removed it entirely because the correct form only asserts the absense of a property. This would be useful as a regression test in the context of V8's test suite, but for Test262, there isn't any context to motivate that particular assertion. Does that seem reasonable to you?

anba:
Property absence resp. property inheritance is tested sometimes, for example the "valueOf" property for `Array` and `Function`: 
`built-ins/Array/prototype/S15.4.4_A2.1_T1.js` and `built-ins/Function/prototype/S15.3.4_A4.js`

jugglinmike:
In that case, I'm happy to add it back. I did have one question about the form you suggested. In your example, you used `void 0` in place of `undefined`. My intuition would be to use the value directly instead of relying on the behavior of an unrelated operator. Can you explain why `void 0` is preferable? I know that `undefined` can be shadowed with a local variable, but can't tell from your example if that is your motivation (or if it is a case we should even be concerned with).

This is likely to come up in many future tests, so I'd like to learn about this sooner rather than later :)

anba:
I've used `void 0` just out of habit, `undefined` is also ok. Or even preferable for the reason outlined by you (=> unrelated operator).

jugglinmike:
Great, thanks! I've corrected the assertions and re-introduced the test.

bterlson:
This overall seems great to me!

We do have a propertyHelper helper that should be used for validating descriptors - include propertyHelper.js, and then you can call verifyConfigurable(object, prop). The benefit here is two-fold: you get better error messages, and it also validates the descriptor is respected (historically some implementations have had issues where the descriptor says something isn't writable but it is).

That said, looking at the implementation of these helpers, it doesn't seem to be checking the descriptor at all. @Smikes do you recall if this is intentional?

smikes:
`verifyConfigurable` and friends were introduced as direct ports of the behavior of `accessorPropertyAttributesAreCorrect` which were pretty terrible large functions that returned a boolean.

It would be reasonable for `verifyConfigurable` (etc.) to check the property as well.  When they were introduced I just wanted to land the patch as a refactor with minimal changes to functionality.

One caveat with these functions is that `verifyConfigurable`/`verifyNotConfigurable` should be called last, because `isConfigurable` will delete the property if it is in fact configurable, so all your other tests of that property & its attributes should be completed by then.

bterlson:
I made #170 to track adding descriptor validation. In the meantime, this PR should be updated to use these helpers (it will not be validating the descriptor for now, but will once #170 is fixed).

bterlson:
Well #170 is already closed, so I think using the existing property helpers will just work now (and give you more rigorous validation to boot!)

test/built-ins/Array/prototype/Symbol.iterator/iteration.js
```diff
@@ -0,0 +1,29 @@
+// Copyright (C) 2013 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+  description: >
+      The method should return a valid iterator with the context as the
```
arv: This description does not seem to explain what this test does.

test/built-ins/Array/prototype/entries/non-enumerable.js
```diff
@@ -0,0 +1,19 @@
...
+/*---
+  description: >
+      The method should exist on the Array prototype, and it should be writable
+      and configurable, but not enumerable.
+  es6id: 22.1.3.4
+ ---*/
+
+var proto = Array.prototype;
+
+assert(proto.hasOwnProperty('entries'));
```
arv: Is this needed? We already test that we get an own descriptor below? I guess it ensures that a VM does not have bugs related to hasOwnProperty?

rwaldron: This one like is just irrelevant to this particular file's test case, but looking at the mjsunit/es6/array-iterator.js file, I can see how this got here

jugglinmike:
> In the meantime, this PR should be updated to use these helpers

I began making this change, but it looks as though `verifyWritable` is also destructive. I think we can update it to make it non-destructive, but I still generally feel uneasy with this approach. For many other language features, Test262 maintains a group of tests asserting their behavior and assumes their proper functioning everywhere else. For example, [`compareArray` uses a `for` loop internally](https://github.com/tc39/test262/blob/2eca2c71e8632ad3093e95bdc2ba499920362465/harness/compareArray.js), and this is generally considered "safe" because [the `for` statement is tested exhaustively in isolation](https://github.com/tc39/test262/tree/2eca2c71e8632ad3093e95bdc2ba499920362465/test/language/statements/for). The functions in `propertyHelper.js` definitely seem useful for the purposes of testing the implementation of property descriptors. That said, re-verifying the behavior of meta-properties (as opposed to the value of property descriptors) across the test suite seems like an exception to the "test it and forget it" approach.

You've mentioned implementations behaving inconsistently in the past:

> (historically some implementations have had issues where the descriptor says something isn't writable but it is)

and while I can appreciate that danger, I'm wondering how I can weigh the danger of inconsistencies in the run time against tractability problems in using a language to test itself.

test/built-ins/Array/prototype/Symbol.iterator/iteration.js
jugglinmike: The description relies on an implicit understanding of the term "valid iterator". Do you think it would be better to describe the behavior of the iterator in the description?

test/built-ins/Array/prototype/entries/non-enumerable.js
jugglinmike: Definitely unnecessary. I've removed that assertion from this file and the similar assertions from files for the property descriptors of the other Array prototype methods.

jugglinmike:
@bterlson I'm working on porting additional iterator tests for typed arrays, and it's made me re-think the file structure here. As it stands, this patch introduces a new directory at `test/built-ins/ArrayIterator/`. Even though a directory name like "built-in" does not necessarily mean "global identifier", this would be the only subdirectory that cannot be referenced from the global scope directly.

I think it would be better (if a little more cumbersome) to move that directory to `test/built-ins/Array/prototype/Symbol.iterator/` . Given that this patch is mid-review, I wanted to get your input before making that change.

jugglinmike:
Alternatively: `test/build-ins/Array/prototype/@@iterator/` . This is a little unconventional but should be fine across platforms. I don't mean to bikeshed here, just trying to find a logical scheme that scales while it's still easy to change.

rwaldron:
> `test/built-ins/Array/prototype/Symbol.iterator/`

+ 1

`@@iterator` is a spec shorthand

bterlson:
Descriptors on objects are not treated equally (at least in some implementations) so it is not ideal IMO to only test that descriptors are respected in one place. The historical issue here was with host objects - some host objects would misreport their descriptor values (or not respect them).

I don't think there is a "test it and forget it" approach to Test262 :) I'd say for loop is fine to use in assertion because it's reasonable to expect implementations have a functioning for loop before running test262. Not so for many more advanced features (eg. everything in ES6).

Can you help me understand the concern better, though? Since test262 tests are supposed to test only one thing and each test is run in isolation I don't see the destructiveness being too much of a hindrance. Thoughts?

jugglinmike:
@bterlson I'm mostly trying to understand how you think about these issues
generally. I've written a fair number of tests before, but nothing quite like
this! It's useful to get your perspective.

Specifically here, the problem I see is that exposing destructive assertions
places an extra burden on the test author (and maintainer during review). I'll
defer to you on the history of the project and implementation bugs--that's not
something I'm able to accurately weigh against the tradeoffs.

Anyway, I've opened gh-191 with an idea of how to make the writability check
non-destructive. This will make it safer for the same test to assert both
writability and configurability (the latter of which cannot be made generally
non-destructive... as far as I know).

rwaldron:
> The historical issue here was with host objects - some host objects would misreport their descriptor values (or not respect them).


Why are host objects relevant to to Test262?

bterlson:
Global object is a host object (or can be).

@jugglinmike I see your point about the destructiveness being gross. I am definitely ok with repairing the object. Are there any downsides to doing so? I would be fine with removing the destructive validation completely, although it seems like good coverage to have if it's free.

jugglinmike:
We can make the writable check generally non-destructive, but it's not completely free because it depends on additional language features (`delete`, `else`, `Object.prototype.hasOwnProperty`, and `Function.prototype.call`) within the assertion. Not quite as demanding as the rejected `assert.deepEqual` but still worth mentioning.

The benefit is that few caveats put a smaller burden on the test author--they don't have to be aware of as many edge cases. I'll defer to you whether that is worth it or not.

domenic:
I am still sympathetic to @bterlson's point that

> Since test262 tests are supposed to test only one thing and each test is run in isolation I don't see the destructiveness being too much of a hindrance.

The test author shouldn't need to worry about the side-effects of the destructive test, if the test is appropriately small.

jugglinmike:
@domenic The property descriptor assertions are usually going to be made in groups, and my concern is about side-effects within a set of assertions for the same test. Obviously assertions against the property value will fail after making a destructive test of its writability, but test authors will probably catch those problems immediately. But there's at least one slightly less obvious interaction that would be good to avoid:

```js
var obj = {};
isEnumerable(obj, 'foo'); // false
isWriteable(obj, 'foo');  // true

obj = {};
isWriteable(obj, 'foo');  // true
isEnumerable(obj, 'foo'); // true
```

caitp:
here's an idea: what if you had a constant set of flags, representing property attributes, and you passed a set of them to a single test function, which worked through them in a specific order, so that the destructiveness definitely didn't affect other property attribute tests?

jugglinmike:
@caitp Totally! I started writing a function just like that, but it started to feel a little heavy. Then again, it's probably not much heavier than the existing functions combined (just a little more logic to parse the "expected descriptor values")

caitp:
you could also do something fancier like `hasAttributes(O, P, "readonly, configurable, !enumerable")` or something --- so that you could easily have negated flags --- then all you need to do is translate the string into a set of tests to run in a particular order (with destructive tests running last).

jugglinmike:
I have a feeling that such a nice API would be rejected on [the same grounds as `assert.deepEquals`](https://github.com/tc39/test262/pull/153).

The simplest approach I can think of accepts an object literal to describe expectations and assumes that all three (enumerability, writability, and configurability) should be verified for every call.

```js
function verifyProperty(obj, name, expected) {
  if (expected.enumerable) {
    verifyEnumerable(obj, name);
  } else {
    verifyNotEnumerable(obj, name);
  }

  if (expected.writable) {
    verifyWritable(obj, name);
  } else {
    verifyNotWritable(obj, name);
  }

  if (expected.configurable) {
    verifyConfigurable(obj, name);
  } else {
    verifyNotConfigurable(obj, name);
  }
}
```

smikes:
We had something like that before and it was a readability nightmare. I think it was assertPropertyDescriptorsAreCorrect()

jugglinmike:
Ah yeah, looks like that function was removed last year: 49abae4f5e6b38931d86dc9b9d34ae30d9ced2ca . This version isn't victim to the boolean trap (avoiding readability problems), but I'm starting to worry that this pull request is getting a bit derailed.

I'll update my patch to use the property helpers directly, and I'll just take special care about the sequence. I'll update gh-191 with the additional context/options so that the discussion about safety/usability in these helpers can continue independently.

jugglinmike:
@bterlson I've pushed some additional commits to use propertyHelper.js and rename a few files. Let me know what you think!

bterlson:
LGTM (esp the commit messages)! Squash and I'll pull?

jugglinmike:
Sure thing! I can copy the commit messages into the squashed commit, but I don't think they will make much sense after that. Did you mean to say "esp the assertion messages"?

jugglinmike:
In the interest of avoiding more round-trips, I'm going to assume that was a typo. I've force-pushed a squashed commit *without* the extra commit messages, but I have a local copy of the branch in a pre-squash state and would be happy to append the extra commit messages at your request.

bterlson:
Sorry, esp === especially! Please add them back and I'll pull.

PR #167, opened 2015-02-18
initial set of generator tests

smikes:
batch of generator tests
fix info, desc

### DIFF
--- a/test/language/generators/generator.expression.no-name.js
+++ b/test/language/generators/generator.expression.no-name.js
@@ -1,0 +1,20 @@
+// Copyright 2015 Cubane Canada, Inc.  All rights reserved.
+// See LICENSE for details.
+
+/*---
+info: >
+    Generator can be declared with GeneratorExpression syntax
+es6id: 14.4
+author: Sam Mikes
+description: can create generator function expressions (no name)
+---*/
+
+var a = [function *(a) { yield a+1; return; }];
+var f = a[0];
+
+assert.sameValue(f.name, '');
+
+var g = f(3);
+
+assert.sameValue(g.next().value, 4);
+assert.sameValue(g.next().done, true);
--- a/test/language/generators/generator.declaration.js
+++ b/test/language/generators/generator.declaration.js
@@ -1,0 +1,17 @@
+// Copyright 2015 Cubane Canada, Inc.  All rights reserved.
+// See LICENSE for details.
+
+/*---
+info: >
+    Generator can be declared with GeneratorDeclaration syntax
+es6id: 14.4
+author: Sam Mikes
+description: can declare generator functions
+---*/
+
+function *foo(a) { yield a+1; return; }
+
+var g = foo(3);
+
+assert.sameValue(g.next().value, 4);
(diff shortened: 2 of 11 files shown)


test/language/generators/generator.declareMethod.shadow-parameter-catch.js
```diff
@@ -0,0 +1,22 @@
...
+description: GeneratorMethod error with lexical shadowing
+negative: SyntaxError
+---*/
+
+var obj = {
+    *foo(a) {
+        try {
+        }
+        catch (e) {
+            let a = 3;
```
bterlson: Should this really be an error? I think you can shadow formals, but you can't redeclare them (eg. let a = 3 in the generator body would be an error).

smikes: I'll remove this test for now.  Originally the parameter to `catch` was named `a`, but I am unclear on whether that's legal or not, and v8 is not a suitable reference implementation for this because their let/const handling is broken.

test/language/generators/generator.expression.no-name.js
```diff
@@ -0,0 +1,19 @@
...
+info: >
+    Generator can be declared with GeneratorExpression syntax
+es6id: 14.4
+author: Sam Mikes
+description: can create generator function expressions (no name)
+---*/
+
+var f = function *(a) { yield a+1; return; };
+
+assert.sameValue(f.name, '');
```
bterlson: I believe the name of this generator should be 'f', cf.

https://people.mozilla.org/~jorendorff/es6-draft.html#sec-variable-statement-runtime-semantics-evaluation
https://people.mozilla.org/~jorendorff/es6-draft.html#sec-isanonymousfunctiondefinition
https://people.mozilla.org/~jorendorff/es6-draft.html#sec-generator-function-definitions-static-semantics-isfunctiondefinition

test/language/generators/generator.declareMethod.shadow-parameter-catch.js
bterlson: I think if you rename let a = 3 to let e = 3 it should error, but this is not a generator test but a try/catch test IMO. Fine with removal :)

test/language/generators/generator.expression.no-name.js
smikes: Huh.  Okay, I will fix it.

To test for no name, I guess I need something like:

```
var a = [ function () {} ];
assert.sameValue(a[0].name, '')
```

Will that work?

bterlson: I believe that test is correct.

bterlson:
Can you squash?

smikes:
yup

test/language/generators/generator.expression.no-name.js
anba: Using an array works, alternatively you can use a comma expression `var a = (0, function(){})` or if you like to confuse readers `var a; (a) = function(){};`. :wink:

PR #168, opened 2015-02-18
Import tests from Google V8 (Arrow Functions)

rwaldron:
project:

	test/mjsunit/harmony/arrow-functions.js

```
test262 --command "v8 --harmony-arrow-functions" Arrow-Function;
test262 --command "js" Arrow-Function;
```

### DIFF
--- a/test/language/arrow-function/Arrow-Function_rules-for-prototype.js
+++ b/test/language/arrow-function/Arrow-Function_rules-for-prototype.js
@@ -1,0 +1,13 @@
+// Copyright (C) Copyright 2014 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 14.2
+description: >
+    Arrow functions are like functions, except they throw when using the
+    "new" operator on them.
+---*/
+
+assert.sameValue(typeof (() => {}), "function");
+assert.sameValue(Object.getPrototypeOf(() => {}), Function.prototype);
+assert.throws(TypeError, function() { new (() => {}); });
+assert.sameValue("prototype" in (() => {}), false);
--- a/test/language/arrow-function/Arrow-Function_semantics.js
+++ b/test/language/arrow-function/Arrow-Function_semantics.js
@@ -1,0 +1,35 @@
+// Copyright (C) Copyright 2014 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 14.2
+description: >
+    Semantics associated with specific Arrow Function syntactic forms.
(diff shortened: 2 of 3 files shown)


test/language/arrow-function/Arrow-Function_semantics.js
```diff
@@ -0,0 +1,35 @@
+// Copyright (C) Copyright 2014 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 14.2
+description: >
+    Semantics associated with specific syntactic forms.
```
domenic: specific *arrow function* syntactic forms :)

test/language/arrow-function/Arrow-Function_rules-for-prototype.js
```diff
@@ -0,0 +1,13 @@
+// Copyright (C) Copyright 2014 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 14.2
+description: >
+    Arrow functions are like functions, except they throw when using the
```
domenic: This test seems a bit broader than this sentence implies. Not sure what a good description would be though...

test/language/arrow-function/Arrow-Function_semantics.js
```diff
@@ -0,0 +1,35 @@
...
+var square = x => x * x;
+assert.sameValue(square(3), 9);
+
+// Parenthesize the body to return an object literal expression
+var key_maker = val => ({key: val});
+assert.sameValue(key_maker(empty).key, empty);
+
+// Expression Body implicit return
+var evens = [0, 2, 4, 6, 8];
+assert(compareArray(evens.map(v => v + 1), [1, 3, 5, 7, 9]));
```
domenic: We should add `assert.arrayLikeEqual` maybe to encompass this case, before this pattern gets too widespread.

rwaldron: I will do that after this and the two array-specific batches land. That way I can do a single pass to update them all.

domenic:
Overall LGTM, nice smoke tests. One could make a case for splitting up into more files, but probably not worth it IMO.

test/language/arrow-function/Arrow-Function_semantics.js
domenic: :+1:

rwaldron:
>  One could make a case for splitting up into more files,

In cases like this, where there are relatively few tests, I'd like to hold off on big splits until phase 2 when we're actively writing new tests.

bterlson:
Well I'd prefer it to be split but it's probably fine for now.

Issue #173, opened 2015-02-20
Rationale for `assert()` expecting literal `true` instead of any truthy value?


smikes:
This bit me last night, because I was doing

```
assert(Symbol.species, 'Symbol.species exists');
```

which failed because `assert` expects a true value.  I switched it to

```
assert(Symbol.species !== undefined, 'Symbol.species exists');
```

which works fine.  Just wondering what the rationale was, since this seems to have been a deliberate choice, and I certainly won't be the last person who's bitten by it.

domenic:
I think @arv talked me into it. In a language-level suite, you shouldn't use concepts like truthiness. You should be more precise in a given assertion.

For example, your second test will pass with `Symbol.species === 0`, whereas your first will fail.

smikes:
Fair enough.  It certainly shaped my assertion choices, I found myself preferring `assert.sameValue()` for later assertions, so I think it helped me be less sloppy overall.

bterlson:
It forces you to be explicit. Many values are truthy but you should be testing for one specific value not any value that happens to be truthy.

PR #175, opened 2015-02-20
Import tests from Google V8 (Class syntax & semantics)

rwaldron:
project:

- test/mjsunit/harmony/class-computed-property-names-super.js
- test/mjsunit/harmony/class-property-name-eval-arguments.js
- test/mjsunit/harmony/classes-experimental.js
- test/mjsunit/harmony/classes-lazy-parsing.js
- test/mjsunit/harmony/classes.js

I'd like to recommend adding the following to the harness: 

```js
assert.methodDescriptor = function(object, name) {
  var descr = Object.getOwnPropertyDescriptor(object, name);
  assert.sameValue(descr.configurable, true);
  assert.sameValue(descr.enumerable, false);
  assert.sameValue(descr.writable, true);
  assert.sameValue(typeof descr.value, 'function');
  assert.sameValue('prototype' in descr.value, false);
};

assert.getterDescriptor = function(object, name) {
  var descr = Object.getOwnPropertyDescriptor(object, name);
  assert.sameValue(descr.configurable, true);
  assert.sameValue(descr.enumerable, false);
  assert.sameValue(typeof descr.get, 'function');
  assert.sameValue('prototype' in descr.get, false);
  assert.sameValue(descr.set, undefined);
};

assert.accessorDescriptor = function(object, name) {
  var descr = Object.getOwnPropertyDescriptor(object, name);
  assert.sameValue(descr.configurable, true);
  assert.sameValue(descr.enumerable, false);
  assert.sameValue(typeof descr.get, 'function');
  assert.sameValue(typeof descr.set, 'function');
  assert.sameValue('prototype' in descr.get, false);
  assert.sameValue('prototype' in descr.set, false);
};

assert.setterDescriptor = function(object, name) {
  var descr = Object.getOwnPropertyDescriptor(object, name);
  assert.sameValue(descr.configurable, true);
  assert.sameValue(descr.enumerable, false);
  assert.sameValue(descr.get, undefined);
  assert.sameValue(typeof descr.set, 'function');
  assert.sameValue('prototype' in descr.set, false);
};

```

I'm confident that there are plenty of places these could be used in the existing tests, as well as future tests. There is already some duplication among the tests here, but that could be eliminated by porting the V8 versions to test262 (as shown above)

# Notes: 

The frontmatter description field values are just the name of the file, which came from the IIFE they were originally wrapped in. I sincerely hope that I don't have to write 51 specific descriptions. The frontmatter es6id field values all point to the class section of the spec and no deeper. Locating the specific deep sections and writing descriptions should be written up as a new issue and offered as a "good first bug" to a community contributor.

### DIFF
--- a/test/language/class/arguments/access.js
+++ b/test/language/class/arguments/access.js
@@ -1,0 +1,54 @@
+// Copyright (C) Copyright 2014 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 14.5
+description: >
+    class arguments access
+---*/
+var constructCounts = {
+  base: 0,
+  subclass: 0,
+  subclass2: 0
+};
+
+class Base {
+  constructor() {
+    constructCounts.base++;
+    assert.sameValue(arguments.length, 2, "The value of `arguments.length` is `2`");
+    assert.sameValue(arguments[0], 1, "The value of `arguments[0]` is `1`");
+    assert.sameValue(arguments[1], 2, "The value of `arguments[1]` is `2`");
+  }
+}
+
+var b = new Base(1, 2);
+
+class Subclass extends Base {
+  constructor() {
+    constructCounts.subclass++;
+    assert.sameValue(arguments.length, 2, "The value of `arguments.length` is `2`");
+    assert.sameValue(arguments[0], 3, "The value of `arguments[0]` is `3`");
+    assert.sameValue(arguments[1], 4, "The value of `arguments[1]` is `4`");
+    super(1, 2);
+  }
+}
+
+var s = new Subclass(3, 4);
+assert.sameValue(Subclass.length, 0, "The value of `Subclass.length` is `0`, because there are 0 formal parameters");
+
+class Subclass2 extends Base {
+  constructor(x, y) {
+    constructCounts.subclass2++;
+    assert.sameValue(arguments.length, 2, "The value of `arguments.length` is `2`");
+    assert.sameValue(arguments[0], 3, "The value of `arguments[0]` is `3`");
+    assert.sameValue(arguments[1], 4, "The value of `arguments[1]` is `4`");
+    super(1, 2);
+  }
+}
+
+var s2 = new Subclass2(3, 4);
+assert.sameValue(Subclass2.length, 2, "The value of `Subclass2.length` is `2`, because there are 2 formal parameters");
+
+
+assert.sameValue(constructCounts.base, 3, "The value of `constructCounts.base` is `3`");
+assert.sameValue(constructCounts.subclass, 1, "The value of `constructCounts.subclass` is `1`");
+assert.sameValue(constructCounts.subclass2, 1, "The value of `constructCounts.subclass2` is `1`");
--- a/test/language/class/arguments/default-constructor.js
+++ b/test/language/class/arguments/default-constructor.js
@@ -1,0 +1,30 @@
+// Copyright (C) Copyright 2014 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 14.5
+description: >
+    class default constructor arguments
+---*/
+var args, that;
+class Base {
+  constructor() {
+    that = this;
+    args = arguments;
+  }
+}
+class Derived extends Base {}
+
+new Derived;
+assert.sameValue(args.length, 0, "The value of `args.length` is `0`");
+
+new Derived(0, 1, 2);
+assert.sameValue(args.length, 3, "The value of `args.length` is `3`");
+assert.sameValue(
+  that instanceof Derived,
+  true,
+  "The result of `that instanceof Derived` is `true`"
+);
+
+var arr = new Array(100);
+var obj = {};
+assert.throws(TypeError, function() {Derived.apply(obj, arr);});
--- a/test/language/class/definition/accessors.js
+++ b/test/language/class/definition/accessors.js
@@ -1,0 +1,52 @@
+// Copyright (C) Copyright 2014 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 14.5
+description: >
+    class accessors
+---*/
+
+function assertAccessorDescriptor(object, name) {
+  var desc = Object.getOwnPropertyDescriptor(object, name);
+  assert.sameValue(desc.configurable, true, "The value of `desc.configurable` is `true`");
+  assert.sameValue(desc.enumerable, false, "The value of `desc.enumerable` is `false`");
+  assert.sameValue(typeof desc.get, 'function', "`typeof desc.get` is `'function'`");
+  assert.sameValue(typeof desc.set, 'function', "`typeof desc.set` is `'function'`");
+  assert.sameValue(
+    'prototype' in desc.get,
+    false,
+    "The result of `'prototype' in desc.get` is `false`"
+  );
+  assert.sameValue(
+    'prototype' in desc.set,
+    false,
+    "The result of `'prototype' in desc.set` is `false`"
+  );
+}
+
+
+class C {
+  constructor(x) {
+    this._x = x;
+  }
+
+  get x() { return this._x; }
+  set x(v) { this._x = v; }
+
+  static get staticX() { return this._x; }
+  static set staticX(v) { this._x = v; }
+}
+
+assertAccessorDescriptor(C.prototype, 'x');
+assertAccessorDescriptor(C, 'staticX');
+
+var c = new C(1);
+c._x = 1;
+assert.sameValue(c.x, 1, "The value of `c.x` is `1`, after executing `c._x = 1;`");
+c.x = 2;
+assert.sameValue(c._x, 2, "The value of `c._x` is `2`, after executing `c.x = 2;`");
+
+C._x = 3;
+assert.sameValue(C.staticX, 3, "The value of `C.staticX` is `3`, after executing `C._x = 3;`");
+C._x = 4;
+assert.sameValue(C.staticX, 4, "The value of `C.staticX` is `4`, after executing `C._x = 4;`");
--- a/test/language/class/definition/basics.js
+++ b/test/language/class/definition/basics.js
@@ -1,0 +1,43 @@
+// Copyright (C) Copyright 2014 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 14.5
+description: >
+    class basics
+---*/
+var C = class C {}
+assert.sameValue(typeof C, 'function', "`typeof C` is `'function'`");
+assert.sameValue(
+    Object.getPrototypeOf(C.prototype),
+    Object.prototype,
+    "`Object.getPrototypeOf(C.prototype)` returns `Object.prototype`"
+);
+assert.sameValue(
+    Object.getPrototypeOf(C),
+    Function.prototype,
+    "`Object.getPrototypeOf(C)` returns `Function.prototype`"
+);
+assert.sameValue(C.name, 'C', "The value of `C.name` is `'C'`");
+
+class D {}
+assert.sameValue(typeof D, 'function', "`typeof D` is `'function'`");
+assert.sameValue(
+    Object.getPrototypeOf(D.prototype),
+    Object.prototype,
+    "`Object.getPrototypeOf(D.prototype)` returns `Object.prototype`"
+);
+assert.sameValue(
+    Object.getPrototypeOf(D),
+    Function.prototype,
+    "`Object.getPrototypeOf(D)` returns `Function.prototype`"
+);
+assert.sameValue(D.name, 'D', "The value of `D.name` is `'D'`");
+
+class D2 { constructor() {} }
+assert.sameValue(D2.name, 'D2', "The value of `D2.name` is `'D2'`");
+
+var E = class {}
+assert.sameValue(E.name, 'E', "The value of `E.name` is `'E'`");
+
+var F = class { constructor() {} };
+assert.sameValue(F.name, 'F', "The value of `F.name` is `'F'`");
--- a/test/language/class/definition/constructable-but-no-prototype.js
+++ b/test/language/class/definition/constructable-but-no-prototype.js
@@ -1,0 +1,11 @@
+// Copyright (C) Copyright 2014 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 14.5
+description: >
+    class constructable but no prototype
+---*/
+var Base = function() {}.bind();
+assert.throws(TypeError, function() {
+  class C extends Base {}
+});
--- a/test/language/class/definition/constructor-property.js
+++ b/test/language/class/definition/constructor-property.js
@@ -1,0 +1,17 @@
+// Copyright (C) Copyright 2014 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 14.5
+description: >
+    class constructor property
+---*/
+class C {}
+assert.sameValue(
+    C,
+    C.prototype.constructor,
+    "The value of `C` is `C.prototype.constructor`"
+);
+var desc = Object.getOwnPropertyDescriptor(C.prototype, 'constructor');
+assert.sameValue(desc.configurable, true, "The value of `desc.configurable` is `true`, after executing `var desc = Object.getOwnPropertyDescriptor(C.prototype, 'constructor');`");
+assert.sameValue(desc.enumerable, false, "The value of `desc.enumerable` is `false`, after executing `var desc = Object.getOwnPropertyDescriptor(C.prototype, 'constructor');`");
+assert.sameValue(desc.writable, true, "The value of `desc.writable` is `true`, after executing `var desc = Object.getOwnPropertyDescriptor(C.prototype, 'constructor');`");
--- a/test/language/class/definition/constructor-strict-by-default.js
+++ b/test/language/class/definition/constructor-strict-by-default.js
@@ -1,0 +1,15 @@
+// Copyright (C) Copyright 2014 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 14.5
+description: >
+    class constructor strict
+---*/
+class C {
+  constructor() {
+    assert.throws(ReferenceError, function() {
+      nonExistingBinding = 42;
+    });
+  }
+}
+new C();
--- a/test/language/class/definition/constructor.js
+++ b/test/language/class/definition/constructor.js
@@ -1,0 +1,35 @@
+// Copyright (C) Copyright 2014 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 14.5
+description: >
+    class constructor
+---*/
+var count = 0;
+class C {
+  constructor() {
+    assert.sameValue(
+      Object.getPrototypeOf(this),
+      C.prototype,
+      "`Object.getPrototypeOf(this)` returns `C.prototype`"
+    );
+    count++;
+  }
+}
+assert.sameValue(
+  C,
+  C.prototype.constructor,
+  "The value of `C` is `C.prototype.constructor`"
+);
+var desc = Object.getOwnPropertyDescriptor(C.prototype, 'constructor');
+assert.sameValue(desc.configurable, true, "The value of `desc.configurable` is `true`, after executing `var desc = Object.getOwnPropertyDescriptor(C.prototype, 'constructor');`");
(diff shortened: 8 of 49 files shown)


domenic:
+1 to helpers, although they might belong in a more specific helpers file like descriptorHelpers.js or something (not sure). Also they should possibly use the verify* helpers instead of just testing the descriptor.

smikes:
If we use the verifyConfigurable / verifyNotConfigurable helper, it'll be destructive if the property is in fact configurable. (I'm a broken record about this, but I think it's a big footgun.)

test/language/class/class-basics-extends.js
```diff
@@ -0,0 +1,16 @@
+// Copyright (C) Copyright 2014 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 14.5
+description: >
+    class basics extends
+---*/
+class C extends null {}
```
arv: This one is not valid with the new syntax. If there is an extends expression then the default constructor is:

```js
constructor(...args) {
  super(..args);
}
```

which throws because `Function.prototype` is not constructable.

rwaldron: Will remove.

test/language/class/class-basics.js
```diff
@@ -0,0 +1,29 @@
...
+assert.sameValue(Object.getPrototypeOf(D), Function.prototype);
+assert.sameValue(D.name, 'D');
+
+class D2 { constructor() {} }
+assert.sameValue(D2.name, 'D2');
+
+// TODO(arv): The logic for the name of anonymous functions in ES6 requires
+// the below to be 'E';
+var E = class {}
+assert.sameValue(E.name, '');  // Should be 'E'.
```
arv: Remove this invalid test line?

test/language/class/class-basics.js
```diff
@@ -0,0 +1,29 @@
...
+
+var F = class { constructor() {} };
+assert.sameValue(F.name, '');  // Should be 'F'.
```
arv: ditto

test/language/class/class-basics.js
rwaldron: Instead I will update them to test for the correct values.

test/language/class/class-basics.js
rwaldron: Fixed

test/language/class/class-default-constructor-2.js
```diff
@@ -0,0 +1,63 @@
...
+assert.sameValue(s2prime.x, 3);
+assert.sameValue(s2prime.y, 4);
+
+
+var obj = {};
+class Base3 {
+  constructor() {
+    return obj;
+  }
+  wtf() {}
```
arv: this line can be removed

test/language/class/class-to-string.js
```diff
@@ -0,0 +1,15 @@
...
+/*---
+es6id: 14.5
+description: >
+    class to string
+---*/
+class C {}
+assert.sameValue(C.toString(), 'class C {}');
+
+class D { constructor() { 42; } }
+assert.sameValue(D.toString(), 'class D { constructor() { 42; } }');
```
jugglinmike: I think these tests are overly specific for Test262. From [19.2.3.5 Function.prototype.toString ( )](http://people.mozilla.org/~jorendorff/es6-draft.html#sec-function.prototype.tostring):

> The use and placement of white space, line terminators, and semicolons within the representation String is implementation-dependent.

test/language/class/class-subclass-binding.js
```diff
@@ -0,0 +1,40 @@
...
+    this.x = x;
+    this.y = y;
+  }
+}
+
+var obj = {};
+class Subclass extends Base {
+  constructor(x, y) {
+    super(x,y);
+    assert.sameValue(this !== obj, true);
```
arv: or just `assert(this !== obj)`

test/language/class/class-default-constructor-2.js
rwaldron: Thats hilarious. I added that while I was trying to figure out some unrelated non-sense with https://github.com/Benvie/JavaScriptNext.tmLanguage

test/language/class/class-super-in-static-setter.js
```diff
@@ -0,0 +1,23 @@
...
+  }
+  static get x() {
+    return 2;
+  }
+}
+class C extends B {
+  static set x(v) {
+    assert.sameValue(v, 3);
+    assert.sameValue(super.x, 2);
+    assert.sameValue(super.method(), 1);
```
jugglinmike: Do you think it's worth added a flag to avoid false positives (i.e. if the setter is never called)?

jugglinmike: Reading further, it looks like if we wanted to do this, the assertion would need to be repeated in a lot of these tests.

test/language/class/class-subclass-binding.js
rwaldron: This was transformed from `assertTrue`, which implies a level of specificity. It's balanced by `assertFalse` transforming to `assert.sameValue(expr, false);`. We could transform `assertTrue(Binary Expression)` into `assert(Binary Expression)`...?

jugglinmike: Based on [recent conversation](https://github.com/tc39/test262/issues/173#issuecomment-75316654), it sounds like we can simply transform `assertTrue` to `assert`, regardless of arguments.

test/language/class/class-strict-mode.js
```diff
@@ -0,0 +1,25 @@
...
+    class strict mode
+---*/
+class C {}
+
+with ({a: 1}) {
+  assert.sameValue(a, 1);
+}
+
+assert.throws(SyntaxError,
+             'class C extends function B() { with ({}); return B; }() {}');
```
jugglinmike: It doesn't look like Test262's implementation of `assert.throws` accepts a string: https://github.com/tc39/test262/blob/ba3be4a575b453bd9792ecbfc5e4e844b2422434/harness/assert.js#L44-L64

test/language/class/class-subclass-binding.js
rwaldron: It looks messy and confusing when there are assertions mixed together, eg: 

```js
assert(descr.configurable);
assert.sameValue(descr.enumerable, false);
assert(descr.writable);
assert.sameValue(typeof descr.value, 'function');
assert.sameValue('prototype' in descr.value, false);
```

When transforming it to: 

```js
assert.sameValue(descr.configurable, true);
assert.sameValue(descr.enumerable, false);
assert.sameValue(descr.writable, true);
assert.sameValue(typeof descr.value, 'function');
assert.sameValue('prototype' in descr.value, false);
```

Has a readable uniformity to it.

rwaldron: Note: I'm not going to manually re-order the assertions in the first example :P

test/language/class/class-strict-mode.js
rwaldron: Yeah, good catch. Funny thing though: 


```js
assert.throws(SyntaxError, function() {
  class C extends (function B() { with ({}); return B; }()) {}
});
```

Fails with: 

```
$ test262 --command "v8 --harmony-classes --harmony-computed-property-names --harmony-sloppy" class-strict-mode
=== language/class/class-strict-mode failed in non-strict mode ===
--- output ---
 /var/folders/jl/3l91m2l97xjbh10v99_xmwmh0000gn/T/test262-hMYUkc.js:130: SyntaxError: Strict mode code may not include a with statement
  class C extends (function B() { with ({}); return B; }()) {}
                                  ^^^^
SyntaxError: Strict mode code may not include a with statement===
```

How is that not caught?

jugglinmike: Is the error being thrown prior to execution? As in, statically? That would explain why the V8 tests eval'd a string, but I'm not sure it really makes any sense.

jugglinmike: This is probably why Test262 defines [the `negative` flag](https://github.com/tc39/test262/blob/master/CONTRIBUTING.md#negative): so you can write expectations for SyntaxErrors without relying on `eval` (and assuming its correct functioning)

test/language/class/class-name-binding.js
```diff
@@ -0,0 +1,36 @@
+// Copyright (C) Copyright 2014 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 14.5
+description: >
+    class name binding
+---*/
+var C2;
+class C {
```
jugglinmike: It seems like this test is only interesting if the class is defined with a ClassExpression instead of a ClassDeclaration. As it stands, it's just asserting that class methods have access to the outer environment record.

Instead, what about

```js
var C;
var OuterName = class InnerName {
  constructor() {
    C = InnerName
  }
// etc...
```

jugglinmike: Gah, [there's already a test for that](https://github.com/bocoup/test262/blob/harmony-classes/test/language/class/class-name-binding-expression.js). Sorry for the noise; I'll try to be a little more patient!

test/language/class/class-default-constructor-arguments.js
```diff
@@ -0,0 +1,26 @@
+// Copyright (C) Copyright 2014 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 14.5
+description: >
+    class default constructor arguments
+---*/
+var args, self;
```
jugglinmike: This probably cannot pass in browser environments. How about `that` or `instance` instead of `self`?

jugglinmike: Alternatively, you could place the assertions inside the constructor and then use a flag to ensure that codepath is hit (that might be slightly more consistent with the other tests)

test/language/class/class-strict-mode.js
```diff
@@ -0,0 +1,25 @@
+// Copyright (C) Copyright 2014 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 14.5
+description: >
+    class strict mode
```
domenic: Based on the name, this test file should probably be tagged onlyStrict?

domenic: Nevermind, wrong interpretation. In fact maybe it should be tagged noStrict, given that it has a top-level `with`.

test/language/class/class-strict-mode.js
domenic: Yes, this would best be tested with the negative flag (in its own file) I think.

rwaldron: Fixed.

test/language/class/class-accessors.js
```diff
@@ -0,0 +1,44 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 14.5
+description: >
+    class accessors
+---*/
+
+function assertAccessorDescriptor(object, name) {
+  var descr = Object.getOwnPropertyDescriptor(object, name);
+  assert.sameValue(descr.configurable, true);
```
bterlson: I'd prefer just `assert(descr.configurable)` (and `assert(!descry.enumerable)`) personally, but it's not a big deal...

bterlson: Also an informative message would be appreciated. Failures are difficult to debug without them :(

test/language/class/class-default-constructor-2.js
```diff
@@ -0,0 +1,62 @@
...
+    return obj;
+  }
+}
+
+class Subclass3 extends Base3 {};
+
+var s3 = new Subclass3();
+assert.sameValue(s3, obj);
+
+class ExtendedUint8Array extends Uint8Array { }
```
bterlson: This test should be pulled out into a separate file as its an integration test with another feature (that may not be present).

test/language/class/class-default-constructor-no-crash.js
```diff
@@ -0,0 +1,13 @@
+// Copyright (C) Copyright 2014 the V8 project authors. All rights reserved.
```
bterlson: I'm not sure this is interesting from a test262 perspective. Would be ok removing this test.

test/language/class/class-name-binding-in-extends-expression.js
```diff
@@ -0,0 +1,18 @@
...
+assert.throws(ReferenceError, function() {
+  class x extends x {}
+});
+
+assert.throws(ReferenceError, function() {
+  (class x extends x {});
+});
+
+assert.throws(ReferenceError, function() {
+  var x = (class x extends x {});
```
bterlson: Per https://people.mozilla.org/~jorendorff/es6-draft.html#sec-runtime-semantics-classdefinitionevaluation 6.f this should be a TypeError not a reference error. X is declared due to var hoisting, but its value is undefined failing the IsConstructor check.

test/language/class/class-strict-mode.js
```diff
@@ -0,0 +1,17 @@
...
+flags: [noStrict]
```
bterlson: I guess this flag can be removed now that with is split out.

test/language/class/class-name-binding-in-extends-expression.js
anba: It's a `ReferenceError`, `x` is still in TDZ, cf. step 4 and step 6.a

test/language/class/class-strict-mode-with.js
```diff
@@ -0,0 +1,15 @@
...
+description: >
+    class strict mode, with
+flags: [noStrict]
+negative: SyntaxError
+---*/
+with ({a: 1}) {
+  assert.sameValue(a, 1);
+}
+
+class C extends (function B() { with ({}); return B; }()) {}
```
bterlson: I'm not sure what the syntax error should be here. Possibly include it in the description.

test/language/class/class-name-binding-in-extends-expression.js
bterlson: Ahh yes, good call!

test/language/class/class-strict-mode.js
arv: The top level with was used to ensure that the outer context of the class was sloppy.

test/language/class/class-strict-mode.js
arv: Yup. The test should pass in both modes.

test/language/class/class-strict-mode-with.js
arv: with is not allowed in strict mode.

http://people.mozilla.org/~jorendorff/es6-draft.html#sec-strict-mode-code

 * All parts of a ClassDeclaration or a ClassExpression are strict mode code.

@bterlson If the test does not have a [noStrict] does it run the test in strict and sloppy mode? If so, then the noStrict flag can be removed here too.

test/language/class/class-accessors.js
domenic: I disagree on assert usage actually. You could imagine horrible implementation bugs that make it return `undefined` instead of `false` or ... something ... instead of `true`.

bterlson: But our assert checks specifically for the true value!

test/language/class/class-strict-mode-with.js
bterlson: Yes, the default without flags is to run it in both modes (although so many tests fail this that we can't yet do a clean run in both modes :()

test/language/class/class-accessors.js
smikes: { writable: true, enumerable: "bunny rabbit" }

assert (!desc.enumerable)

Passes

bterlson: Bah. Why even have an assert function then?

test/language/class/class-to-string.js
bterlson: Agree, I'm not sure we can get away with this. Suggest removal, unless there is major pushback. I'm not sure impls will agree precisely on toString output (and toString output of class constructors especially seems open to interpretation)

bterlson:
Once open feedback is addressed this is good to go from my perspective.

Overall I'm really missing descriptions that actually discuss the semantics under test and assert messages that helpfully inform about what a failure might mean. But I won't push on it if I'm alone in thinking it's important :)

smikes:
What I'd rather have (short term) rather than descriptions is a sense of the roadmap Bocoup is using. Like, could you do the generators tests sooner, how long will the mjsunit porting phase last.  Etc. Even gross metrics in terms of lines of code would be cool. 

I can sign up for documenting the semantics under test for at least some of the imported tests. But we will wind up doing that in a review while checking for coverage so overall I'm not fussed about comment/front matter stuff for now.

rwaldron:
> What I'd rather have (short term) rather than descriptions is a sense of the roadmap Bocoup is using.

Alphabetical order, migrating from these directories: 

- https://chromium.googlesource.com/v8/v8.git/+/master/test/mjsunit/es6/
- https://chromium.googlesource.com/v8/v8.git/+/master/test/mjsunit/harmony/

>  how long will the mjsunit porting phase last.

Initially, 5 weeks, but my guess is that we'll continue to work on migration in Phase 2.

test/language/class/class-accessors.js
rwaldron: > Also an informative message would be appreciated. Failures are difficult to debug without them :(

Added informative messages

test/language/class/class-default-constructor-2.js
rwaldron: Done (class-extends-typed-array.js)

test/language/class/class-default-constructor-no-crash.js
rwaldron: Removed.

test/language/class/class-name-binding-const.js
```diff
@@ -0,0 +1,15 @@
...
+    class name binding const
+---*/
+assert.throws(TypeError, 'class C { constructor() { C = 42; } }; new C();');
+assert.throws(TypeError, 'new (class C { constructor() { C = 42; } })');
+assert.throws(TypeError, 'class C { m() { C = 42; } }; new C().m()');
+assert.throws(TypeError, 'new (class C { m() { C = 42; } }).m()');
+assert.throws(TypeError, 'class C { get x() { C = 42; } }; new C().x');
+assert.throws(TypeError, '(new (class C { get x() { C = 42; } })).x');
+assert.throws(TypeError, 'class C { set x(_) { C = 42; } }; new C().x = 15;');
+assert.throws(TypeError, '(new (class C { set x(_) { C = 42; } })).x = 15;');
```
rwaldron: These have been converted to:

```js
  class C { constructor() { C = 42; } }; new C();
});
  new (class C { constructor() { C = 42; } })
});
  class C { m() { C = 42; } }; new C().m()
});
  new (class C { m() { C = 42; } }).m()
});
  class C { get x() { C = 42; } }; new C().x
});
  (new (class C { get x() { C = 42; } })).x
});
  class C { set x(_) { C = 42; } }; new C().x = 15;
});
  (new (class C { set x(_) { C = 42; } })).x = 15;
});
```

test/language/class/class-name-binding-in-extends-expression.js
rwaldron: No change

test/language/class/class-strict-mode-with.js
rwaldron: The outer `with` can be removed, making this viable in either mode. `class C extends (function B() { with ({}); return B; }()) {}` is a syntax error in all modes

test/language/class/class-strict-mode.js
rwaldron: Removed

test/language/class/class-to-string.js
rwaldron: Removed

smikes:
Thanks!   Alphabetical order seems arbitrary, so can I persuade you to do generators next?  That would free me up to work on additional generator tests without worrying about duplicating work.

rwaldron:
@bterlson 

https://github.com/bocoup/test262/commit/d63ce029a6682bc54a3fab48df7d8a8059a5db4f


> Overall I'm really missing descriptions that actually discuss the semantics under test 

I mentioned this in the initial pull request message—I think this should be filed as an issue and marked "good first bug".

> and assert messages that helpfully inform about what a failure might mean. But I won't push on it if I'm alone in thinking it's important :)

This migration contains 351 assertions in 49 files—I think it's important as well, but I have to balance resource time.

rwaldron:
@smikes I believe @jugglinmike is actually working on that as I type this :)

jugglinmike:
> @smikes I believe @jugglinmike is actually working on that as I type this :)

[I'm in the process of porting some iteration tests](https://github.com/bocoup/test262/tree/iteration-semantics) that touch generators, but they are largely focused on `for..of`. I'm on another project this week, though. I suspect there are plenty of other generator tests to work on, but if I'm blocking, then just let me know.

smikes:
No worries, I am kind of swamped with paying work and after that, Symbol.species is next on my list. If you just pick your next file from the "g"s then we probably won't block each other.

rwaldron:
Squashed & Rebased.

rwaldron:
Re-org & assertion messages

bterlson:
Looks like this is good enough to pull in. Thanks everyone!

PR #183, opened 2015-03-02
Import tests from Google V8 (Block Scoping)

rwaldron:
project:

	test/mjsunit/harmony/block-conflicts.js
	test/mjsunit/harmony/block-for.js
	test/mjsunit/harmony/block-leave.js
	test/mjsunit/harmony/block-let-declaration.js
	test/mjsunit/harmony/block-let-semantics.js
	test/mjsunit/harmony/block-scoping.js

### DIFF
--- a/test/language/block-scope/syntax/for-in/acquire-properties-from-object.js
+++ b/test/language/block-scope/syntax/for-in/acquire-properties-from-object.js
@@ -1,0 +1,21 @@
+// Copyright (C) Copyright 2011 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 13.1
+description: >
+    for-in to acquire properties from object
+includes: [compareArray.js]
+---*/
+function props(x) {
+  var array = [];
+  for (let p in x) array.push(p);
+  return array.sort();
+}
+
+assert.sameValue(props({}).length, 0);
+assert.sameValue(props({x:1}).length, 1);
+assert.sameValue(props({x:1, y:2}).length, 2);
+
+assert(compareArray(props({x:1}), ["x"]));
+assert(compareArray(props({x:1, y:2}), ["x", "y"]));
+assert(compareArray(props({x:1, y:2, zoom:3}), ["x", "y", "zoom"]));
--- a/test/language/block-scope/syntax/for-in/disallow-initialization-assignment.js
+++ b/test/language/block-scope/syntax/for-in/disallow-initialization-assignment.js
@@ -1,0 +1,11 @@
+// Copyright (C) Copyright 2011 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 13.1
+description: >
+    for declaration:
+    disallow initialization assignment
+negative: SyntaxError
+---*/
+for (let x = 3 in {}) { }
+
--- a/test/language/block-scope/syntax/for-in/disallow-multiple-lexical-bindings-with-and-without-initializer.js
+++ b/test/language/block-scope/syntax/for-in/disallow-multiple-lexical-bindings-with-and-without-initializer.js
@@ -1,0 +1,11 @@
+// Copyright (C) Copyright 2011 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 13.1
+description: >
+    for declaration:
+    disallow multiple lexical bindings, with and without initializer
+negative: SyntaxError
+---*/
+for (let x = 3, y in {}) { }
+
--- a/test/language/block-scope/syntax/for-in/disallow-multiple-lexical-bindings-without-and-with-initializer.js
+++ b/test/language/block-scope/syntax/for-in/disallow-multiple-lexical-bindings-without-and-with-initializer.js
@@ -1,0 +1,11 @@
+// Copyright (C) Copyright 2011 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 13.1
+description: >
+    for declaration:
+    disallow multiple lexical bindings, without and with initializer
+negative: SyntaxError
+---*/
+for (let x, y = 4 in {}) { }
+
--- a/test/language/block-scope/syntax/for-in/disallow-multiple-lexical-bindings.js
+++ b/test/language/block-scope/syntax/for-in/disallow-multiple-lexical-bindings.js
@@ -1,0 +1,11 @@
+// Copyright (C) Copyright 2011 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 13.1
+description: >
+    for declaration:
+    disallow multiple lexical bindings
+negative: SyntaxError
+---*/
+for (let x, y in {}) { }
+
--- a/test/language/block-scope/syntax/for-in/missing-identifier-let-disallowed-as-bound-name.js
+++ b/test/language/block-scope/syntax/for-in/missing-identifier-let-disallowed-as-bound-name.js
@@ -1,0 +1,11 @@
+// Copyright (C) Copyright 2011 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 13.1
+description: >
+    for declaration:
+    missing identifier, "let" disallowed as bound name
+negative: SyntaxError
+---*/
+for (let in {}) { }
+
--- a/test/language/block-scope/syntax/function-declarations/in-statement-position-label-statement.js
+++ b/test/language/block-scope/syntax/function-declarations/in-statement-position-label-statement.js
@@ -1,0 +1,10 @@
+// Copyright (C) Copyright 2011 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 13.1
+description: >
+    function declarations in statement position in strict mode:
+    label: Statement
+---*/
+label: function g() {}
+
--- a/test/language/block-scope/syntax/let-declarations/without-initialisers-in-statement-positions-case-expression-statement-list.js
+++ b/test/language/block-scope/syntax/let-declarations/without-initialisers-in-statement-positions-case-expression-statement-list.js
@@ -1,0 +1,10 @@
+// Copyright (C) Copyright 2011 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 13.1
+description: >
+    let declarations without initialisers in statement positions: 
+    case Expression : StatementList
+negative: SyntaxError
+---*/
+switch (true) { case true: let x; }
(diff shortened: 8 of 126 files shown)


smikes:
That's all I have time for right now, I will return to this later today.

test/language/block-scope/syntax/leave/verify-context-in-finally-block-1.js
```diff
@@ -0,0 +1,21 @@
...
+  (function(x) {
+    try {
+      let x = 'inner';
+      throw 0;
+    } finally {
+      f();
+      assert.sameValue(x, 'outer');
+    }
+  })('outer');
+} catch (e) {}
```
jugglinmike: I think you need to re-throw here--otherwise, assertion errors will not be reported.

test/language/block-scope/syntax/leave/try-block-let-declaration-only-shadows-outer-parameter-value-1.js
```diff
@@ -0,0 +1,16 @@
+// Copyright (C) Copyright 2011 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 13.1
+description: >
+    try block let declaration only shadows outer parameter value 1
+---*/
+(function(x) {
```
jugglinmike: For these tests, do you think it matters that the outermost `x` declaration is made via an arguments list (`var` semantics) and not a `let` statement? I guess they're identical in this context, but I'm wondering if using arguments here could hide weird implementation bugs.

test/language/block-scope/syntax/leave/verify-context-in-finally-block-1.js
jugglinmike: Maybe to avoid the extra logic of conditionally re-throwing, it would be better in these cases to simply capture the value inside the `try` block and make the assertion afterward:


```diff
+var result;
 try {
  (function(x) {
    try {
      let x = 'inner';
      throw 0;
    } finally {
      f();
-     assert.sameValue(x, 'outer');
+     result = x;
     }
   })('outer');
 } catch (e) {}
+assert.sameValue(result, 'outer');
```

...but then again, since this test concerns variable scoping directly, it might be dangerous to use the closure like this.

test/language/block-scope/syntax/for-loop/iteration-variable-is-freshly-allocated-for-each-iteration-multi-let-binding.js
```diff
@@ -0,0 +1,16 @@
...
+    In a normal for statement the iteration variable is freshly allocated for each iteration. Multi let binding
+---*/
+let a = [], b = [];
+for (let i = 0, j = 10; i < 5; ++i, ++j) {
+  a.push(function () { return i; });
+  b.push(function () { return j; });
+}
+for (let k = 0; k < 5; ++k) {
+  console.log(k, a[k]());
+  console.log(k + 10, b[k]());
```
jugglinmike: `console.log` -> `assert.sameValue`

rwaldron: Derp.

test/language/block-scope/syntax/for-in/acquire-properties-from-object.js
```diff
@@ -0,0 +1,21 @@
+// Copyright (C) Copyright 2011 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 13.1
+description: >
+    for-in to acquire properties from object
+includes: [compareArray.js]
+---*/
+function props(x) {
```
jugglinmike: You can get away without `Array.prototype.sort` and `compareArray.js` if you use an object literal:

```diff
 function props(x) {
-  var array = [];
+  var obj = {};
+  var count = 0;
-  for (let p in x) array.push(p);
+  for (let p in x) {
+    obj[p] = true;
+    count++;
+  }
-  return array.sort();
+  return { obj: obj, count: count }
 }
```

...but it's hard to tell if that's really worth it. Sometimes this feels more like code golf, so I won't mind if you tell me it is :)

test/language/block-scope/syntax/for-in/disallow-initialization-assignment.js
```diff
@@ -0,0 +1,11 @@
+// Copyright (C) Copyright 2011 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 13.1
+description: >
+    for declaration:
+    disallow initialization assignment
+negative: SyntaxError
+---*/
+for (let x = 3 in {}) { } }
```
jugglinmike: The trailing `}` means that this file will always yield a syntax error, even in runtimes that ought to fail (i.e. those that allow initialization assignments as the left hand side of a for..in statement)

test/language/block-scope/syntax/for-in/disallow-multiple-lexical-bindings-with-and-without-initializer.js
```diff
@@ -0,0 +1,11 @@
+// Copyright (C) Copyright 2011 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 13.1
+description: >
+    for declaration:
+    disallow multiple lexical bindings, with and without initializer
+negative: SyntaxError
+---*/
+for (let x = 3, y in {}) { } }
```
jugglinmike: Same as above

test/language/block-scope/syntax/for-in/missing-identifier-let-disallowed-as-bound-name.js
```diff
@@ -0,0 +1,11 @@
+// Copyright (C) Copyright 2011 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 13.1
+description: >
+    for declaration:
+    missing identifier, "let" disallowed as bound name
+negative: SyntaxError
+---*/
+for (let in {}) { } }
```
jugglinmike: this one too

test/language/block-scope/syntax/for-in/disallow-multiple-lexical-bindings.js
```diff
@@ -0,0 +1,11 @@
+// Copyright (C) Copyright 2011 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 13.1
+description: >
+    for declaration:
+    disallow multiple lexical bindings
+negative: SyntaxError
+---*/
+for (let x, y in {}) { } }
```
jugglinmike: You asked for me to be explicit :P

test/language/block-scope/syntax/for-in/disallow-multiple-lexical-bindings-without-and-with-initializer.js
```diff
@@ -0,0 +1,11 @@
+// Copyright (C) Copyright 2011 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 13.1
+description: >
+    for declaration:
+    disallow multiple lexical bindings, without and with initializer
+negative: SyntaxError
+---*/
+for (let x, y = 4 in {}) { } }
```
jugglinmike: one more

jugglinmike:
Done reviewing. Nice work, Rick!

arv:
@rossberg-chromium FYI

rwaldron:
Squashed & Rebased.

test/language/block-scope/syntax/redeclaration-in-block/attempt-to-redeclare-function-declaration-with-function-declaration.js
```diff
@@ -0,0 +1,11 @@
+// Copyright (C) Copyright 2011 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: B.3.3
+description: >
+    redeclaration within block:
+    attempt to redeclare function declaration with function declaration
+negative: SyntaxError
```
bterlson: I can see this being true in strict mode but is it also true in sloppy? Fairly surprising.

bterlson:
Awesome tests (modulo lack of assertion failure messages).

Presumably most/all of the let/const cases would also apply to class decls, so perhaps this PR should include those as well?

test/language/block-scope/syntax/redeclaration-in-block/attempt-to-redeclare-let-binding-with-var.js
```diff
@@ -0,0 +1,11 @@
+// Copyright (C) Copyright 2011 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: B.3.3
+description: >
+    redeclaration within block:
+    attempt to redeclare let binding with var
+negative: SyntaxError
+---*/
+{ let f; var f; }
```
bterlson: I think this test is wrong? The var decl is hoisted out so there should be no error afaict?

anba: Early error per 13.1.1, second item

bterlson: :ok_hand:

test/language/block-scope/syntax/let-declarations/without-initialisers-in-statement-positions-default-statement-list.js
```diff
@@ -0,0 +1,10 @@
+// Copyright (C) Copyright 2011 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 13.1
+description: >
+    let declarations without initialisers in statement positions: 
+    default : StatementList
+negative: SyntaxError
+---*/
+switch (true) { default: let x; }
```
anba: This declaration is allowed.

test/language/block-scope/syntax/let-declarations/without-initialisers-in-statement-positions-case-expression-statement-list.js
```diff
@@ -0,0 +1,10 @@
+// Copyright (C) Copyright 2011 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 13.1
+description: >
+    let declarations without initialisers in statement positions: 
+    case Expression : StatementList
+negative: SyntaxError
+---*/
+switch (true) { case true: let x; }
```
anba: This declaration is allowed.

test/language/block-scope/syntax/for-in/missing-identifier-let-disallowed-as-bound-name.js
```diff
@@ -0,0 +1,11 @@
+// Copyright (C) Copyright 2011 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 13.1
+description: >
+    for declaration:
+    missing identifier, "let" disallowed as bound name
+negative: SyntaxError
+---*/
+for (let in {}) { }
```
anba: This is also allowed.

test/language/block-scope/syntax/function-declarations/in-statement-position-label-statement.js
```diff
@@ -0,0 +1,10 @@
+// Copyright (C) Copyright 2011 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 13.1
+description: >
+    function declarations in statement position in strict mode:
+    label: Statement
+---*/
+label: function g() {}
```
anba: Only allowed for non-strict with AnnexB.

PR #184, opened 2015-03-02
Import tests from Google V8 (iteration semantics)

jugglinmike:
project:

test/mjsunit/es6/iteration-semantics.js

### DIFF
--- a/test/language/for-of/break-from-try.js
+++ b/test/language/for-of/break-from-try.js
@@ -1,0 +1,28 @@
+// Copyright (C) Copyright 2013 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 13.6.4.13 S5.n
+description: >
+    Control flow during body evaluation should honor `break` statements within
+    `try` blocks.
+---*/
+
+function* values() {
+  yield 1;
+  yield 1;
+}
+var iterable = values();
+var i = 0;
+
+for (var x of iterable) {
+  i++;
+
+  try {
+    break;
+    $ERROR('This code is unreachable.');
+  } catch (err) {}
+
+  $ERROR('This code is unreachable.');
+}
+
+assert.sameValue(i, 1);
--- a/test/language/for-of/iterator-as-proxy.js
+++ b/test/language/for-of/iterator-as-proxy.js
@@ -1,0 +1,35 @@
+// Copyright (C) Copyright 2013 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 13.6.4.13
+description: >
+    Iterators that are implemented as proxies should behave identically to
+    non-proxy versions.
+---*/
+
+var iterable = {};
+var nextResult = { value: 23, done: false };
+var lastResult = { value: null, done: true };
+var i;
+
+var iterator = {
+  next: function() {
+    var result = nextResult;
+    nextResult = lastResult;
+    return result;
+  }
+};
+var proxiedIterator = new Proxy(iterator, {
+  get: function(target, name) {
+    return target[name];
+  }
+});
+iterable[Symbol.iterator] = function() { return proxiedIterator; };
+
+i = 0;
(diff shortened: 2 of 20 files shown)


smikes:
Looks good overall.  Some of the later tests which test multiple cases (all the non-Objects) might be better split off into separate files, but you should wait for @anba or @bterlson to deliver a definitive opinion.

test/language/for-of/break-from-try.js
```diff
@@ -0,0 +1,28 @@
...
+}
+var iterable = values();
+var i = 0;
+
+for (var x of iterable) {
+  i++;
+
+  try {
+    break;
+    assert(false, 'This code is unreachable.');
```
arv: Maybe we should add `assert.unreachable()`?

test/language/for-of/iterator-next-result-type.js
```diff
@@ -0,0 +1,103 @@
...
+for (var x of iterable) {}
+
+firstIterResult = Proxy.create({
+  get: function(receiver, name) {
+    if (name === 'done') {
+      return true;
+    }
+    if (name === 'value') {
+      return null;
+    }
```
arv: assert unreachable?

arv:
Very nice.

test/language/for-of/iterator-next-result-type.js
```diff
@@ -0,0 +1,103 @@
...
+  for (var x of iterable) {}
+});
+
+firstIterResult = /regexp/;
+for (var x of iterable) {}
+
+firstIterResult = {};
+for (var x of iterable) {}
+
+firstIterResult = Proxy.create({
```
anba: `Proxy.create` is the old Proxy API, the test needs to be adjusted to use direct proxies.

test/language/for-of/iterator-next-result-type.js
```diff
@@ -0,0 +1,103 @@
...
+    if (name === 'value') {
+      return null;
+    }
+  }
+});
+for (var x of iterable) {
+  assert(false, 'This code is unreachable.');
+}
+
+firstIterResult = Proxy.create({
```
anba: Needs to use direct proxy API.

test/language/for-of/iterator-as-proxy.js
```diff
@@ -0,0 +1,35 @@
...
+var i;
+
+var iterator = {
+  next: function() {
+    var result = nextResult;
+    nextResult = lastResult;
+    return result;
+  }
+};
+var proxiedIterator = Proxy.create({
```
anba: Needs to use direct proxy API.

test/language/for-of/iterator-next-result-type.js
jugglinmike: Roger that

jugglinmike:
@arv re:

> Maybe we should add `assert.unreachable()`?

[I made a branch with this assertion](https://github.com/tc39/test262/compare/master...bocoup:assert-unreachable), but then I did some research into the existing tests. [It looks like the idiomatic way of expressing this is through invoking the global `$ERROR` function directly.](https://github.com/tc39/test262/search?utf8=%E2%9C%93&q=%24ERROR&type=Code) While it isn't as consistent as a new `assert.unreachable`, it's definitely better than `assert(false)`. I've updated my tests to use `$ERROR`, but I'll be happy to follow up with that patch for `assert.unreachable` if @bterlson thinks it's worthwhile.

test/language/for-of/iterator-next-result-type.js
jugglinmike: Couldn't hurt!

bterlson:
I'm not seeing what assert.unreachable gives us so I'm fine with $ERROR("should not execute this code") or somesuch.

jugglinmike:
I think I've addressed all the feedback so far. @bterlson if you're happy with the changeset, let me know and I'll squash it down for you.

bterlson:
Squash away! :+1:

jugglinmike:
All squashed

PR #189, opened 2015-03-04
Import tests from Google V8 (Computed Property Names)

rwaldron:
project:

        test/mjsunit/harmony/computed-property-names-classes.js
        test/mjsunit/harmony/computed-property-names-object-literals-methods.js
        test/mjsunit/harmony/computed-property-names-super.js
        test/mjsunit/harmony/computed-property-names.js


test/language/computed-property-names/computed-property-names-static-class-method-symbol.js
```diff
@@ -0,0 +1,38 @@
...
+//
+// Since there is no guarantee of order:
+//
+//    - Assert only that the symbol is present
+//    - Assert that the length is correct
+//
+var symbols = Object.getOwnPropertySymbols(C);
+
+assert(symbols.indexOf(sym1) !== -1);
+assert(symbols.indexOf(sym2) !== -1);
```
jugglinmike: Again, this might be golf, but we *could* avoid `indexOf` without too much hassle:

```diff
-assert(symbols.indexOf(sym1) !== -1);
-assert(symbols.indexOf(sym2) !== -1);
+assert(symbols[0] === sym1 || symbols[1] === sym1);
+assert(symbols[0] === sym2 || symbols[1] === sym2);
```

test/language/computed-property-names/computed-property-names-setter-object.js
```diff
@@ -0,0 +1,55 @@
...
+    calls++;
+  }
+};
+object.b = 'B';
+assert.sameValue(calls, 1);
+
+calls = 0;
+object = {
+  set c(_) {
+    assert(false)
```
jugglinmike: Keep an eye on [this discussion](https://github.com/tc39/test262/pull/184#issuecomment-76986716); depending on its resolution, we may want to switch these to `$ERROR`

test/language/computed-property-names/computed-property-names-generator-computed-name-class.js
```diff
@@ -0,0 +1,16 @@
...
+    computed property names can be used as the name of a generator method in a class
+includes: [compareArray.js]
+---*/
+class C {
+  *['a']() {
+    yield 1;
+    yield 2;
+  }
+}
+assert(compareArray(Object.keys(C.prototype), []));
```
jugglinmike: It's probably enough to just `assertEqual(Object.keys(C.prototype).length, 0)

test/language/computed-property-names/computed-property-names-double-name.js
```diff
@@ -0,0 +1,23 @@
...
+  [-Infinity]: 'F',
+  [NaN]: 'G',
+};
+assert.sameValue(object['1.2'], 'A');
+assert.sameValue(object['1e+55'], 'B');
+assert.sameValue(object['0.000001'], 'C');
+assert.sameValue(object[0], 'D');
+assert.sameValue(object[Infinity], 'E');
+assert.sameValue(object[-Infinity], 'F');
+assert.sameValue(object[NaN], 'G');
```
jugglinmike: Maybe write these last four in string form, too? I know it's not strictly necessary, but that way this test will only rely on implicit string coercion in one place.

jugglinmike:
All done. Looking good

test/language/computed-property-names/computed-property-names-prototype.js
```diff
@@ -0,0 +1,30 @@
+// Copyright (C) Copyright 2014 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 12.2.5
+description: >
+    computed property names can be "prototype"
+---*/
+class C {
+  static ['prototype']() {
```
anba: static methods cannot be named "prototype", because the .prototype property is non-writable + non-configurable, cf. 14.5.14 step 15, 9.2.9 step 8.

js> (class C { static ['prototype']() { } })
uncaught exception: TypeError: cannot create property "prototype"

test/language/computed-property-names/computed-property-names-static-class-method-symbol.js
```diff
@@ -0,0 +1,38 @@
...
+  static [sym1]() { return 'B'; }
+  static c() { return 'C'; }
+  static [sym2]() { return 'D'; }
+}
+assert.sameValue(C.a(), 'A');
+assert.sameValue(C[sym1](), 'B');
+assert.sameValue(C.c(), 'C');
+assert.sameValue(C[sym2](), 'D');
+assert(compareArray(Object.keys(C), []));
+assert(compareArray(Object.getOwnPropertyNames(C), ['length', 'name', 'arguments', 'caller', 'prototype', 'a', 'c']));
```
anba: Missing call to `Array.prototype.sort`.

test/language/computed-property-names/computed-property-names-static-class-method-string.js
```diff
@@ -0,0 +1,20 @@
...
+  static ['b']() { return 'B'; }
+  static c() { return 'C'; }
+  static ['d']() { return 'D'; }
+}
+assert.sameValue(C.a(), 'A');
+assert.sameValue(C.b(), 'B');
+assert.sameValue(C.c(), 'C');
+assert.sameValue(C.d(), 'D');
+assert(compareArray(Object.keys(C), []));
+assert(compareArray(Object.getOwnPropertyNames(C), ['length', 'name', 'arguments', 'caller', 'prototype', 'a', 'b', 'c', 'd']));
```
anba: Missing call to `sort()`.

test/language/computed-property-names/computed-property-names-static-class-method-number.js
```diff
@@ -0,0 +1,20 @@
...
+  static [1]() { return 'B'; }
+  static c() { return 'C'; }
+  static [2]() { return 'D'; }
+}
+assert.sameValue(C.a(), 'A');
+assert.sameValue(C[1](), 'B');
+assert.sameValue(C.c(), 'C');
+assert.sameValue(C[2](), 'D');
+assert(compareArray(Object.keys(C), []));
+assert(compareArray(Object.getOwnPropertyNames(C), ['1', '2', 'length', 'name', 'arguments', 'caller', 'prototype', 'a', 'c']));
```
anba: And here.

test/language/computed-property-names/computed-property-names-static-class-method-symbol.js
rwaldron: Yep. I forgot that this happened last summer. I will file a V8 bug as well

rwaldron: https://code.google.com/p/v8/issues/detail?id=3944&thanks=3944&ts=1425519421

test/language/computed-property-names/computed-property-names-double-name.js
rwaldron: Can we add that to the phase 2 work list?

test/language/computed-property-names/computed-property-names-prototype.js
rwaldron: Yeah, seemed backward to me as well, but... https://github.com/v8/v8-git-mirror/blob/bd21d72d507625e5caf70f31ee11c8c21ead1651/test/mjsunit/harmony/computed-property-names-classes.js#L314-L390

caitp: This might be correct (???) --- The error is specified [here](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-class-definitions-static-semantics-early-errors), but for a computed property name you can't really have an early error like that. The spec doesn't say anything about it being a runtime error, afaik.

test/language/computed-property-names/computed-property-names-static-class-method-symbol.js
anba: Hmm, issue 3944 has been deleted.

test/language/computed-property-names/computed-property-names-prototype.js
anba: Some mjsunit tests are not (or no longer) correct. https://github.com/anba/es6draft/blob/master/src/test/resources/mjsunit.list contains more infos (but I cannot guarantee that list is up-to-date because it requires manual updates :unamused: ).

arv: My bad... https://code.google.com/p/v8/issues/detail?id=3945

test/language/computed-property-names/computed-property-names-setter-class.js
```diff
@@ -0,0 +1,55 @@
+// Copyright (C) Copyright 2014 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 12.2.5
+description: >
+    In a class, duplicate computed property setter names produce only a single property of
```
arv: The name of this file is misleading. Should probably be renamed to duplicate-property-names-setter-class.js

test/language/computed-property-names/computed-property-names-setter-object.js
```diff
@@ -0,0 +1,55 @@
+// Copyright (C) Copyright 2014 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 12.2.5
+description: >
+    In an object, duplicate computed property getter names produce only a single property of
```
arv: Same here re file name

arv: Or not... this one actually uses computed prop names... deleting my comments...

test/language/computed-property-names/computed-property-names-to-name-side-effects-object.js
```diff
@@ -0,0 +1,33 @@
+// Copyright (C) Copyright 2014 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 12.2.5
+description: >
+    to name, accessor side effects 2
```
arv: to name, accessor side effects object literal

and the other one

to name, accessor side effects class literal

rwaldron: Fixed in next push

test/language/computed-property-names/computed-property-names-prototype.js
rwaldron: Fixed in next push (broken into separate syntax error tests)

test/language/computed-property-names/computed-property-names-setter-class.js
rwaldron: Fixed in next push. (There are now two files each, for getter and setter, one that has a single case in it and one for duplicate keys)

bterlson:
This seems good to go, so pulling it in. My only weak concern is that we're not allowing any extra properties on these objects which, I believe, is allowed per spec. But, in the meantime, these tests have the effect of failing anyone holding on to arguments and caller properties, so I'm happy to pull this in anyway :)

PR #190, opened 2015-03-04
Import tests from Google V8 (Object Literal Concise Method and Property Initialization)

rwaldron:
These tests are derived from the following files within the Google V8 project:

        test/mjsunit/harmony/method-name-eval-arguments.js
        test/mjsunit/harmony/object-literals-method.js
        test/mjsunit/harmony/object-literals-property-shorthand.js
        test/mjsunit/harmony/object-literals-super.js

### DIFF
--- a/test/language/object-literal/concise-generator.js
+++ b/test/language/object-literal/concise-generator.js
@@ -1,0 +1,26 @@
+// Copyright (C) Copyright 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 12.2.5
+description: >
+    super method calls in object literal concise generator
+---*/
+var proto = {
+  method() {
+    return 42;
+  }
+};
+
+var object = {
+  *g() {
+    yield super.method();
+  }
+};
+
+Object.setPrototypeOf(object, proto);
+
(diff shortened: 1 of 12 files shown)


test/language/object-literal/object-literal-let-non-strict.js
```diff
@@ -0,0 +1,12 @@
+// Copyright (C) Copyright 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 12.2.5
+description: >
+    In non-strict mode, let is a valid Identifier.
+flags: [noStrict]
+---*/
+var let = 1;
+var object = {let};
```
jugglinmike: The file name and test description are a little vague. If I'm reading the grammar correctly (and if I understand the intent of the test), a better name might be "property-definition-identifier-reference-let" since that would account for some future test named "property-definition-property-name-let":

```js
var object = {
  let: 1
};
```

jugglinmike: Or is it `property-definition-literal-property-name-let`? Sorry, I'm having a little trouble with the grammar

bterlson:
Seems good to go to me.

PR #191, opened 2015-03-04
Make writability helper function non-destructive

jugglinmike:
After checking the writability of a given property, restore the property
value to its original state.

### DIFF
--- a/harness/propertyHelper.js
+++ b/harness/propertyHelper.js
@@ -29,6 +29,9 @@ function isEqualTo(obj, name, expectedValue) {
 
 function isWritable(obj, name, verifyProp, value) {
     var newValue = value || "unlikelyValue";
+    var hadValue = Object.prototype.hasOwnProperty.call(obj, name);
+    var oldValue = obj[name];
+    var result;
 
     try {
         obj[name] = newValue;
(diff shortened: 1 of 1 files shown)


jugglinmike:
In gh-165, @caitp [pointed out](https://github.com/tc39/test262/pull/165#issuecomment-77430460) that we could define a new function that combines all the property descriptor assertions in a "safe" (though still destructive) way. For example, the following function avoids false negatives that can arise from using destructive tests out of order:

```js
function verifyProperty(obj, name, expected) {
  if (expected.enumerable) {
    verifyEnumerable(obj, name);
  } else {
    verifyNotEnumerable(obj, name);
  }

  if (expected.writable) {
    verifyWritable(obj, name);
  } else {
    verifyNotWritable(obj, name);
  }

  if (expected.configurable) {
    verifyConfigurable(obj, name);
  } else {
    verifyNotConfigurable(obj, name);
  }
}
```

In the same issue, @smikes [mentioned](https://github.com/tc39/test262/pull/165#issuecomment-77439347) that [Test262 previously offered a similar function](https://github.com/tc39/test262/commit/49abae4f5e6b38931d86dc9b9d34ae30d9ced2ca), but that it suffered from readability problems.

The above implementation differs in that it accepts a single object describing the expected values instead of a number of positional arguments, so it should be easier to read and use.

bterlson:
I think this is good to go. I would also support @caitp's suggestion. The previous API was removed mostly because it used mysterious Boolean parameters that I always had to look up the order of each time I reviewed a test. The helper proposed above wouldn't suffer a similar problem so I'd be ok with it :)

PR #197, opened 2015-03-18
Import tests from Google V8 (Symbol)

rwaldron:
These tests are derived from the following files within the Google V8 project:

	test/mjsunit/harmony/private.js


rwaldron:
This isn't quite ready...

- Where should the `Object.*` tests go?
- Should the non-API surface stuff be moved into dir of it's own? eg. `test/built-ins/Symbol/semantics/*`?

domenic:
Curious what @bterlson thinks, but my guess is that Object.* tests go in their own subdirectories of test/built-ins/Object/, and other tests will go in test/language/ as appropriate, e.g. test/language/expressions/typeof/ or test/language/expressions/strict-equals and similar

rwaldron:
> and other tests will go in test/language/ as appropriate, e.g. test/language/expressions/typeof/ or test/language/expressions/strict-equals 

Spreading them out makes it hard to run _just_ these Symbol-centric tests as a group; despite that, I agree with this organization.

rwaldron:
Some updates pushed to the organization. Equality tests haven't been moved yet, but work in progress.

bterlson:
Sorry for the late reply. Organizationally speaking, I believe the following should be changed. Normal caveats apply here of course :)

* Coercion tests should be relocated to folders for their respective operators (see test/language/expressions/logical-not/S9.2*.js for the logical not tests). Adding "features: symbol" to the frontmatter would be appropriate I suppose.
* Collection tests should move to the appropriate collection node (eg. no symbol as weakmap key test should move to test/built-ins/WeakMap/prototype/set I suspect)
* Symbol.species tests should be located along with the feature that should implement the species symbol (so Promise[Symbol.species] should be tested as part of the promise tests under test/built-ins/Promise)

Test content wise, I'm having trouble finding why get-set expects, for example, sym.toString = 1 to throw. My read is auto-boxing occurs, and since Symbol.prototype.toString is writable, toString is set on the object wrapper

rwaldron:
@bterlson thanks, I'll get it updated this week

rwaldron:
- [x] Coercion tests should be relocated
- [ ] Collection tests should move to the appropriate collection node
- [ ] Symbol.species tests should be located along with the feature that should implement the species

rwaldron:
| Attribute Name | Default Value | 
| -------------- | ------------- |
| [[Value]] | undefined |
| [[Get]] | undefined |
| [[Set]] | undefined |
| [[Writable]] |  false |
| [[Enumerable]] |  false |
| [[Configurable]] |  false |

rwaldron:
Whoops, I hit that on accident

anba:
> Test content wise, I'm having trouble finding why get-set expects, for example, sym.toString = 1 to throw. My read is auto-boxing occurs, and since Symbol.prototype.toString is writable, toString is set on the object wrapper

6.2.3.2 PutValue, step 6.b calls 9.1.9 \[[Set]]. 9.1.9 \[[Set]], step 5.b returns false, 6.2.3.2 PutValue, step 6.d throws a TypeError.

rwaldron:
Easier to view latest changes with: https://github.com/tc39/test262/pull/197/files?w=1

test/built-ins/Object/seal/symbol_object-contains-symbol-properties.js
```diff
@@ -0,0 +1,19 @@
...
+description: >
+    Object.seal(obj) where obj contains symbol properties.
+---*/
+var symA = Symbol("A");
+var symB = Symbol("B");
+var obj = {};
+obj[symA] = 1;
+Object.seal(obj);
+obj[symA] = 2;
+obj[symB] = 1;
```
arv: why doesn't this throw in strict mode?

test/language/computed-property-names/class/static/method-symbol.js
```diff
@@ -22,6 +22,7 @@ assert(
   compareArray(Object.keys(C), []),
   "`compareArray(Object.keys(C), [])` returns `true`"
 );
+print(Object.getOwnPropertyNames(C));
```
arv: remove?

test/built-ins/Object/seal/symbol_object-contains-symbol-properties.js
rwaldron: Ah, it will. I should make two files from this for strict and non-strict.

test/language/computed-property-names/class/static/method-symbol.js
rwaldron: Yep. Thanks!

bterlson:
@rwaldron, the default attributes are defined in clause 17 for subsequent clauses as writable true configurable true (enumerable false).

@anba, I think I see your reasoning now. It should only throw for properties that exist somewhere in the prototype chain, correct? New properties will not throw? If so, two questions:
1) Should we have tests for the latter? I think so.
2) This behavior should apply to all primitives correct? I don't think any browser is currently throwing on, for example, `1..toString = 1;`

rwaldron:
> the default attributes are defined in clause 17 for subsequent clauses as writable true configurable true (enumerable false).

Yes, they certainly are 0_o

arv:
> 2) This behavior should apply to all primitives correct? I don't think any browser is currently throwing on, for example, 1..toString = 1;

Both V8 and JSC do:

```
(function() { 'use strict';  1 .b = 42; })()
TypeError: Cannot assign to read only property 'b' of 1
```

rwaldron:
@arv yep, a few months ago someone @bocoup encountered and we found  https://code.google.com/p/v8/issues/detail?id=3725 already filed

bterlson:
@arv yeah you're right, Canary is half right per my read. If I read the spec right, assigments to properties on primitives should do the following:

* In sloppy mode, no errors.
* In strict mode:
    * Assignment to non-existent property results in no error - primitive is boxed, primitive gets a new slot, primitive is collected (9.1.9 returns true).
    * Assignment to existing property on a primitive results in an error (9.1.9 returns false).

caitp:
>Assignment to non-existent property results in no error - primitive is boxed, primitive gets a new slot, primitive is collected (9.1.9 returns true).

1. primitive `O` is boxed
2. prototype chain is walked in search of the `P` name (9.1.9.4)
3. end of prototype chain is reached, `P` still not found. New data descriptor `ownDesc` is created (9.1.9.4.d)
4. ownDesc.[[Writable]] is true, continue (9.1.9.5.a)
5. Type(Receiver) is not Object so return false. (9.1.9.5.b)
6. Throw a TypeError (6.2.3.2.6.d)

bterlson:
Heh, for some reason I was reading step 5 as an else if branch :( Good call @caitp, I agree.

Issue #198, opened 2015-03-18
Invalid tests in "computed-property-names"


anba:
The following tests are invalid.

test/language/computed-property-names/class/method/constructor-cannot-be-generator.js
test/language/computed-property-names/class/method/constructor-cannot-be-getter.js
test/language/computed-property-names/class/method/constructor-cannot-be-setter.js
test/language/computed-property-names/class/method/constructor-no-duplicate-1.js
test/language/computed-property-names/class/method/constructor-no-duplicate-2.js
test/language/computed-property-names/class/method/constructor-no-duplicate-3.js

- Expected: Valid to create a property named "constructor" if the property name is a computed property name

test/language/computed-property-names/class/static/generator-constructor.js
test/language/computed-property-names/class/static/getter-constructor.js
test/language/computed-property-names/class/static/method-constructor.js
test/language/computed-property-names/class/static/setter-constructor.js

- Expected: Valid to create a static property named "constructor"

test/language/computed-property-names/class/static/generator-prototype.js
test/language/computed-property-names/class/static/getter-prototype.js
test/language/computed-property-names/class/static/method-prototype.js
test/language/computed-property-names/class/static/setter-prototype.js

- Expected: `TypeError` at runtime.

rwaldron:
I can clean this up.

rwaldron:
I'm not sure about this: 


> test/language/computed-property-names/class/method/constructor-cannot-be-generator.js
> test/language/computed-property-names/class/method/constructor-cannot-be-getter.js
> test/language/computed-property-names/class/method/constructor-cannot-be-setter.js
> test/language/computed-property-names/class/method/constructor-no-duplicate-1.js
> test/language/computed-property-names/class/method/constructor-no-duplicate-2.js
> test/language/computed-property-names/class/method/constructor-no-duplicate-3.js
>
> - Expected: Valid to create a property named "constructor" if the property name is a computed property name

https://people.mozilla.org/~jorendorff/es6-draft.html#sec-static-semantics-constructormethod specifically, the non-normative note: 

"Early Error rules ensure that there is only one method definition named "constructor" and that it is not an accessor property or generator definition."


https://people.mozilla.org/~jorendorff/es6-draft.html#sec-class-definitions-static-semantics-early-errors

Have I misinterpreted this?

rwaldron:
> test/language/computed-property-names/class/static/generator-constructor.js
> test/language/computed-property-names/class/static/getter-constructor.js
[...]

PR #200, opened 2015-03-18
computed-property-names: corrections. Fixes gh-198

rwaldron:


rwaldron:
@bterlson that patch wasn't quite done, but this is the rest.

bterlson:
Guess I should be more patient. Thanks :)

Issue #201, opened 2015-03-19
Invalid tests in test/language/block-scope


anba:
test/language/block-scope/syntax/for-in/missing-identifier-let-disallowed-as-bound-name.js

- Expected: `let` is allowed as IdentifierReference in `for-in` (lookahead restriction only applies to `let [`).

test/language/block-scope/syntax/let-declarations/without-initialisers-in-statement-positions-case-expression-statement-list.js
test/language/block-scope/syntax/let-declarations/without-initialisers-in-statement-positions-default-statement-list.js

- Expected: `LexicalDeclaration` is allowed (DefaultClause -> StatementList -> Declaration -> LexicalDeclaration), and `Initializer` expression is not mandatory.
 
test/language/block-scope/syntax/function-declarations/in-statement-position-label-statement.js

- Expected: Labelled function declarations are only allowed per B.3.2, but the test case is not placed in the "annexB" directory.

rwaldron:
What would I do without you? :)

rwaldron:
> test/language/block-scope/syntax/for-in/missing-identifier-let-disallowed-as-bound-name.js

> - Expected: let is allowed as IdentifierReference in for-in (lookahead restriction only applies to let [).

In https://people.mozilla.org/~jorendorff/es6-draft.html#sec-for-in-and-for-of-statements-static-semantics-early-errors : 

IterationStatement :
  for ( ForDeclaration in Expression ) Statement
  for ( ForDeclaration of AssignmentExpression ) Statement

- It is a Syntax Error if the BoundNames of ForDeclaration contains "let".


> Expected: LexicalDeclaration is allowed (DefaultClause -> StatementList -> Declaration -> LexicalDeclaration), and Initializer expression is not mandatory.

Fixed

> Expected: Labelled function declarations are only allowed per B.3.2, but the test case is not placed in the "annexB" directory.

Fixed.

anba:
`for (let in {}) { }` should parse as
`for (LeftHandSideExpression in Expression) Statement`
where `LeftHandSideExpression` is the `IdentifierReference` "let". 

The early error restriction only applies to `for (let let in {}) { }`.

rwaldron:
@anba thanks for the clarification.

For reference: 

```
ForDeclaration : 
  LetOrConst ForBinding

LetOrConst :  
  let 
  const 

ForBinding : BindingIdentifier
  
```

I will file bugs for V8 and SpiderMonkey

PR #203, opened 2015-03-19
block-scope: corrections. Fixes gh-201

rwaldron:


arv:
LGTM

test/language/block-scope/syntax/for-in/identifier-let-allowed-as-lefthandside-expression.js
```diff
@@ -1,11 +1,10 @@
...
 // This code is governed by the BSD license found in the LICENSE file.
 /*---
-es6id: 13.1
+es6id: 13.6.0.1
 description: >
     for declaration:
-    missing identifier, "let" disallowed as bound name
-negative: SyntaxError
+    identifier "let" allowed as lefthandside expression
 ---*/
```
bterlson: I think you need "noStrict" for this one, though I think you could declare "var let" earlier in the test to make it work in strict mode too.

rwaldron: > I think you need "noStrict" for this one

Yep. Update coming. 


> I think you could declare "var let" 

Should be another test.

bterlson:
Landed this.

PR #204, opened 2015-03-19
Import tests from Google V8 (generator syntax)

jugglinmike:
project:

    test/mjsunit/es6/generators-parsing.js

I have deliberately omitted one assertion from the original test file because I believe it to be invalid:

```js
assertThrows("function* yield() { \"use strict\"; (yield 3) + (yield 4); }",
             SyntaxError);
```

(source: https://github.com/v8/v8-git-mirror/blob/3a79259c0d4ae2d314bb57ad2d8dcde8f2478974/test/mjsunit/es6/generators-parsing.js#L96-L97)

[I've reported this to the V8 team by filing a bug against that project](https://code.google.com/p/v8/issues/detail?id=3975)

### DIFF
--- a/test/language/class/method-definition/generator-no-yield.js
+++ b/test/language/class/method-definition/generator-no-yield.js
@@ -1,0 +1,20 @@
+// Copyright (C) 2013 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+  description: >
+      Generators declared with GeneratorMethod syntax do not require a
+      `yield` expression.
+  features: [generators]
+  es6id: 14.4
+ ---*/
+
+var result;
+class A {
+  *foo(a) {}
+}
+
+result = A.prototype.foo(3).next();
+
+assert.sameValue(result.value, undefined);
+assert.sameValue(result.done, true);
--- a/test/language/class/method-definition/generator-return.js
+++ b/test/language/class/method-definition/generator-return.js
@@ -1,0 +1,23 @@
+// Copyright (C) 2013 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+  description: >
+      `return` is a valid statement within generator function bodies.
+  features: [generators]
+  es6id: 14.4
+ ---*/
+
+var result;
+class A {
+  *g1() { return; }
+  *g2() { return 1; }
+}
+
+result = A.prototype.g1().next();
+assert.sameValue(result.value, undefined);
+assert.sameValue(result.done, true);
+
+result = A.prototype.g2().next();
+assert.sameValue(result.value, 1);
+assert.sameValue(result.done, true);
--- a/test/language/class/method-definition/yield-as-binding-identifier.js
+++ b/test/language/class/method-definition/yield-as-binding-identifier.js
@@ -1,0 +1,17 @@
+// Copyright (C) 2013 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+  description: >
+      `yield` is a reserved keyword within generator function bodies and may
+      not be used as a binding identifier.
+  features: [generators]
+  es6id: 12.1.1
+  negative: SyntaxError
+ ---*/
+
+class A {
+  *g() {
+    yield = 1;
+  }
+}
--- a/test/language/class/method-definition/yield-as-expression-with-rhs.js
+++ b/test/language/class/method-definition/yield-as-expression-with-rhs.js
@@ -1,0 +1,118 @@
+// Copyright (C) 2013 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+  description: >
+      `yield` is a valid expression within generator function bodies.
+  features: [generators]
+  es6id: 14.4
+ ---*/
+
+var iter, result;
+class A {
+  *g1() { (yield 1) }
+  *g2() { [yield 1] }
+  *g3() { {yield 1} }
+  *g4() { yield 1, yield 2; }
+  *g5() { (yield 1) ? yield 2 : yield 3; }
+}
+
+iter = A.prototype.g1();
+result = iter.next();
+assert.sameValue(result.value, 1, 'Within grouping operator: result `value`');
+assert.sameValue(
+  result.done, false, 'Within grouping operator: result `done` flag'
+);
+result = iter.next();
+assert.sameValue(
+  result.value, undefined, 'Following grouping operator: result `value`'
+);
+assert.sameValue(
+  result.done, true, 'Following grouping operator: result `done` flag'
+);
+
+iter = A.prototype.g2();
+result = iter.next();
+assert.sameValue(result.value, 1, 'Within array literal: result `value`');
+assert.sameValue(
+  result.done, false, 'Within array literal: result `done` flag'
+);
+result = iter.next();
+assert.sameValue(
+  result.value, undefined, 'Following array literal: result `value`'
+);
+assert.sameValue(
+  result.done, true, 'Following array literal: result `done` flag'
+);
+
+iter = A.prototype.g3();
+result = iter.next();
+assert.sameValue(result.value, 1, 'Within object literal: result `value`');
+assert.sameValue(
+  result.done, false, 'Within object literal: result `done` flag'
+);
+result = iter.next();
+assert.sameValue(
+  result.value, undefined, 'Following object literal: result `value`'
+);
+assert.sameValue(
+  result.done, true, 'Following object literal: result `done` flag'
+);
+
+iter = A.prototype.g4();
+result = iter.next();
+assert.sameValue(
+  result.value, 1, 'First expression in comma expression: result `value`'
+);
+assert.sameValue(
+  result.done,
+  false,
+  'First expression in comma expression: result `done` flag'
+);
+result = iter.next();
+assert.sameValue(
+  result.value, 2, 'Second expression in comma expression: result `value`'
+);
+assert.sameValue(
+  result.done,
+  false,
+  'Second expression in comma expression: result `done` flag'
+);
+result = iter.next();
+assert.sameValue(
+  result.value, undefined, 'Following comma expression: result `value`'
+);
+assert.sameValue(
+  result.done, true, 'Following comma expression: result `done` flag'
+);
+
+iter = A.prototype.g5();
+result = iter.next();
+assert.sameValue(
+  result.value,
+  1,
+  'Conditional expression in conditional operator: result `value`'
+);
+assert.sameValue(
+  result.done,
+  false,
+  'Conditional expression in conditional operator: result `done` flag'
+);
+result = iter.next();
+assert.sameValue(
+  result.value,
+  3,
+  'Branch in conditional operator: result `value`'
+);
+assert.sameValue(
+  result.done,
+  false,
+  'Branch in conditional operator: result `done` flag'
+);
+result = iter.next();
+assert.sameValue(
+  result.value, undefined, 'Following conditional operator: result `value`'
+);
+assert.sameValue(
+  result.done, true, 'Following conditional operator: result `done` flag'
+);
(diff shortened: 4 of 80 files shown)


test/language/generators/syntax/return.js
```diff
@@ -0,0 +1,18 @@
...
+
+/*---
+  description: >
+      `return` is a valid statement within generator function bodies.
+  es6id: 14.4
+ ---*/
+
+function* g() { return; }
+function* g() { return 1; }
+function* g() {
```
arv: Not clear why this is repeated?

test/language/generators/syntax/yield-as-binding-identifier.js
```diff
@@ -0,0 +1,14 @@
+// Copyright (C) 2013 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+  description: >
+      `yield` is a reserved keyword within generator function bodies and may
+      not be used as a binding identifier.
```
arv: This description is not correct. In this case `yield` is an `IdentitfierExpression`.

test/language/generators/syntax/yield-as-computed-property-name.js
```diff
@@ -0,0 +1,13 @@
+// Copyright (C) 2013 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+  description: >
+      `yield` may be used as a computed property name in an object literal
```
arv: s/computed property name/literal property name/

test/language/generators/syntax/yield-as-expression-with-rhs.js
```diff
@@ -0,0 +1,19 @@
...
+      `yield` is a valid expression within generator function bodies.
+  es6id: 14.4
+ ---*/
+
+function* g() { (yield 1) }
+function* g() { [yield 1] }
+function* g() { {yield 1} }
+function* g() { yield 1, yield 1; }
+function* g() { (yield 1) ? yield 1 : yield 1 }
+function* g() {
```
arv: this is another repeat

test/language/generators/syntax/yield-as-statement.js
```diff
@@ -0,0 +1,18 @@
...
+
+/*---
+  description: >
+      `yield` is a valid statement within generator function bodies.
+  es6id: 14.4
+ ---*/
+
+function* g() { yield; }
+function* g() { yield 1; }
+function* g() {
```
arv: repeat?

test/language/generators/syntax/yield-as-yield-operand.js
```diff
@@ -0,0 +1,13 @@
+// Copyright (C) 2013 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+  description: >
+      `yield` expressions may be used as the right-hand-side of other `yeild`
```
arv: typo

s/yeild/yield/

test/language/generators/syntax/yield-as-binding-identifier.js
jugglinmike: I'm not seeing `IdentifierExpression` in the spec. Was this maybe a typo?

test/language/generators/syntax/return.js
jugglinmike: I suppose to account for an edge case where a newline would trigger different parsing behavior. That seems more like a regression test, though. If we were to try to catch all problems like this exhaustively, we'd be here all day! I'll remove all the duplicates you've identified.

test/language/generators/syntax/yield-as-computed-property-name.js
jugglinmike: Thanks--renamed the file to match.

test/language/generators/syntax/yield-as-yield-operand.js
jugglinmike: Thanks

test/language/generators/syntax/yield-as-function-expression-binding-identifier.js
```diff
@@ -0,0 +1,14 @@
...
+/*---
+  description: >
+      `yield` is a reserved keyword within generator function bodies and may
+      not be used as the binding identifier of a function expression.
+  es6id: 12.1.1
+  negative: SyntaxError
+ ---*/
+
+function* g() {
+  (function yield() {})
```
anba: This is actually allowed: 
> FunctionExpression :
> &nbsp;&nbsp;function BindingIdentifier<sub>opt</sub> ( FormalParameters ) { FunctionBody }

BindingIdentifier has not [Yield] production parameter, so `yield` is a valid identifier in this context.

jugglinmike: And here I thought I was getting good at reading the spec,,, Nice catch!

I've written up a bug report for V8. Since you identified the problem, I thought I'd check with you first: would you like to report it instead?

anba: > would you like to report it instead?

No, I don't mind if you report this.

jugglinmike: Got it! You can follow along here if you like: [Issue 3983: Incorrect restriction on binding identifiers of function expressions](https://code.google.com/p/v8/issues/detail?id=3983)

bterlson:
Please remove the syntax subdirectory many of these are under.

jugglinmike:
The "syntax" subdirectory may seem extraneous in the context of this pull request, but I have organized the tests like this to better accommodate the semantics-related tests defined in these files:

- [v8/test/mjsunit/es6/generators-iteration.js](https://github.com/v8/v8-git-mirror/blob/8e4e83119ca187e8c0e2a3b518cac2dc3def3fa3/test/mjsunit/es6/generators-iteration.js)
- [v8/test/mjsunit/es6/generators-objects.js](https://github.com/v8/v8-git-mirror/blob/8e4e83119ca187e8c0e2a3b518cac2dc3def3fa3/test/mjsunit/es6/generators-objects.js)
- [v8/test/mjsunit/es6/generators-runtime.js](https://github.com/v8/v8-git-mirror/blob/8e4e83119ca187e8c0e2a3b518cac2dc3def3fa3/test/mjsunit/es6/generators-runtime.js)
- [v8/test/mjsunit/es6/generators-states.js](https://github.com/v8/v8-git-mirror/blob/8e4e83119ca187e8c0e2a3b518cac2dc3def3fa3/test/mjsunit/es6/generators-states.js)

Of course, I'm happy to use whatever organization you like; I just wanted to make sure my intentions were clear before updating the patch.

bterlson:
Hmm, thanks for the clarification. I see what you're doing... here's my thoughts, but I have just now begun thinking about this so it could be dumb :)

generators-iteration and generators-states are exploring various usages of yield. I think this belongs in language/generators (or potentially language/yield, although I don't really think that distinction is necessary or desirable).

Many tests in generators-objects and generators-runtime are testing shapes of built-ins so I would place these under built-ins/Generator and built-ins/Generator/prototype as appropriate.

One oddity is that I guess is that tests for the shape of the object returned from calling a generator should be under language/generators.

Thoughts?

The reason I want to make the organization clean is that the organization is also basically a test tagging scheme so I want to be very clear and deliberate with the organizational scheme. Eventually it will be reflected on our new website so the simpler it is the easier the mapping will be in the future.

bterlson:
Test content looks good. The only other comment I have is to suggest that language/generators be renamed to language/statements/generator, and have every test duplicated for a generator expression (put those under language/expressions/generator).

jugglinmike:
@bterlson sounds good to me! I've re-named the tests as you've requested.

I've added the GeneratorExpression analogs along with copies for GeneratorMethods declared within object literals and within class bodies. I plan on preserving those additions as separate commits in the final version of this pull request (since they did not come from V8 directly).

One thing that occurs to me now is that most of the "positive syntax" tests don't actually assert anything. For an example of one file that *does* include an assertion, see [the proposed `/language/expressions/generators/no-yield.js`](https://github.com/bocoup/test262/blob/6e5fdbebff436e41a49999090aaf2015ac933879/test/language/expressions/generators/no-yield.js). I'd like to add a basic assertion along those lines to each of the positive syntax tests so they provide some minimal assurance that the runtime actually executes the code that it parsed. Does that sound good to you?

bterlson:
> One thing that occurs to me now is that most of the "positive syntax" tests don't actually assert anything. 

I thought about this too. I could go either way but I think they're fine as-is. test262-harness verifies that the test runs to completion by appending `;$DONE();` to the end of the test file. I suppose you could do more in-depth validation to ensure the syntax results in a value you expect, but this seems of questionable utility to me. So, IOW, I think the tests are fine as is, but if you want to do more in depth validations I certainly won't object :)

jugglinmike:
`yield` is not a valid identifier in generator function bodies, regardless of whether the code is strict mode or not. This means that the following tests are invalid in their current form:

- test/language/expressions/generators/yield-as-generator-expression-binding-identifier.js
- test/language/method-definition/class/generators/yield-as-generator-method-binding-identifier.js 
- test/language/method-definition/object/generators/yield-as-generator-method-binding-identifier.js 
- test/language/statements/generators/yield-as-generator-declaration-binding-identifier.js

I will remove the `strictOnly` flag and add `negative: SyntaxError` in all cases

jugglinmike:
After a bit more time studying the subscripting notation (and with thanks to @caitp and @arv), I've realized that only the first test I referenced above (yield-as-generator-expression-binding-identifier.js) is invalid.

From [14.4 Generator Function Definitions](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-generator-function-definitions-static-semantics-early-errors):

> *GeneratorExpression* :
>    **function** * *BindingIdentifier*<sub>[Yield]opt</sub> *FormalParameters*<sub>[Yield,GeneratorParameter]</sub> ) { *GeneratorBody* }

...and from [12.1 Identifiers](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-identifiers):

> *BindingIdentifier*<sub>[Yield]</sub> :
>     Identifier
>     [~Yield] **yield**

...and [5.1.5 Grammar Notation](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-grammar-notation):

> If a right-hand side alternative is prefixed with “[+parameter]” that
> alternative is only available if the named parameter was used in
> referencing the production’s nonterminal symbol. If a right-hand side
> alternative is prefixed with “[~parameter]” that alternative is only
> available if the named parameter was not used in referencing the
> production’s nonterminal symbol.

The `[Yield]` subscript modifies the definition of `BindingIdentifier` by restricting it to the set defined by `Identifier`, so `yield` is never a valid *GeneratorExpression* name.

The three other tests I referenced concern *GeneratorMethod*s and *GeneratorDeclaration*s. Section 14.4 defines each of these with "*BindingIdentifier*<sub>[?Yield]</sub>", meaning `yield` *is* a valid binding identifier in those cases.

jugglinmike:
Okay @bterlson, I've pushed up a few more commits to fix the bugs and add basic assertions.

Prior to landing, I intend to squash these into three commits:

1. Import tests from Google V8 (generator syntax)
  - fixup! Remove repeated assertions
  - fixup! Correct typos
  - fixup! Remove repeated assertion
  - fixup! Correct faulty test
  - fixup! Rename tests
 - fixup! Limit Generator statement to non-strict
2. Add asertions for Generator Statement syntax tests
3. Add tests for additional generator function forms
  - Add tests for generator expressions
  - Add tests for generator method definitions
  - fixup! Limit tests to non-strict
  - fixup! Correct invalid test
  - fixup! Add basic tests for positive syntax

Let me know what you think!

jugglinmike:
As per our discussion in IRC, I've made the following changes:

- Moved `language/identifiers/syntax/*` -> `language/identifiers/`
- Moved `language/method-definition/class/generators/*` -> `language/statements/class/method-definition/*` (renaming `return.js` to `generator-return.js` and `no-yield.js` to `generator-no-yield.js`)
- Moved `language/method-definition/object/generators/*` -> `language/expressions/object/method-definition/*` (renaming `return.js` to `generator-return.js` and `no-yield.js` to `generator-no-yield.js`)
- Added  `features: [generators]` to the frontmatter of the "method-definition" tests

bterlson:
Great work landing this!!

jugglinmike:
Thanks for your help :)

PR #205, opened 2015-03-23
Add a test for String.prototype.concat() with single Object argument

caitp:
v8 would fail this test, so it would be nice to have it


rwaldron:
@bterlson this is :+1:

PR #206, opened 2015-03-23
Make test/language/statements compatible for strict mode

anba:
It should now be possible to run the tests in test/language/statements with strict mode enabled. For the most part it was only necessary to add `noStrict` flags or variable declarations. If further changes were required, I've added explicit notes in the commit messages.


test/language/statements/for/S12.6.3_A10.1_T2.js
```diff
@@ -0,0 +1,113 @@
+// Copyright 2009 the Sputnik authors.  All rights reserved.
```
rwaldron: Where did these come from?

anba: S12.6.3_A10.1_T2 is a copy of S12.6.3_A10.1_T1, therefore I've retained the original copyright notice.

bterlson:
Landed this in 15cc639.

PR #207, opened 2015-03-23
Changes for currency data

### LINKED ISSUE #98: Update expected minor unit value for CLF
intl402/ch11/11.1/11.1.1_20_c.js needs to be updated to change the expected minor unit value for CLF (Unidad de Fomento) from 0 to 4.

The URL in the test file should also be updated, from "http://www.currency-iso.org/dl_iso_table_a1.xml" to new the location at "http://www.currency-iso.org/dam/downloads/table_a1.xml".

See: http://www.currency-iso.org/dam/isocy/downloads/dl_currency_iso_amendment_157.pdf
See Also: https://bugzilla.mozilla.org/show_bug.cgi?id=946528

anba:
- Update minor unit for CLF
- Remove withdrawn currencies (LVL, LTL, USS)
- Update link to new location

Fixes #98

### DIFF
--- a/test/intl402/11.1.1_20_c.js
+++ b/test/intl402/11.1.1_20_c.js
@@ -10,7 +10,7 @@ description: >
 author: Norbert Lindenberg
 ---*/
 
-// data from http://www.currency-iso.org/dl_iso_table_a1.xml, 2013-02-25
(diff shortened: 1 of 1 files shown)


rwaldron:
@bterlson  :+1:

PR #209, opened 2015-03-23
Don't use Annex B extensions in main test suite

### LINKED ISSUE #185: numeric literal in 15.2.3.6-2-17-1
test/built-ins/Object/defineProperty/15.2.3.6-2-17-1.js
has the line:
     Object.defineProperty(obj, 000001, {});
but 000001 is not a valid token according to the main body of the spec.
(I think it's valid according to Annex B.)

### LINKED ISSUE #186: string literal in S15.5.4.11_A5_T1
test/built-ins/String/prototype/replace/S15.5.4.11_A5_T1.js
has the line:

    $ERROR('.... /^(a+)\1*,\1+$/ ....'+...);

but the occurrences of `\1` within a string literal are not valid according to the main body of the spec.
(Looks like it's a LegacyOctalEscapeSequence in Annex B.)

anba:
Fixes #185
Fixes #186

### DIFF
--- a/test/built-ins/Object/defineProperty/15.2.3.6-2-17-1.js
+++ b/test/built-ins/Object/defineProperty/15.2.3.6-2-17-1.js
@@ -8,13 +8,13 @@
 es5id: 15.2.3.6-2-17-1
 description: >
     Object.defineProperty - argument 'P' is a number that converts to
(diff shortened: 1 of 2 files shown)


rwaldron:
Confirmed

bterlson:
Good catch.

PR #210, opened 2015-03-26
Import tests from Google V8 (native iterators)

jugglinmike:
project:

    test/mjsunit/es6/typed-array-iterator.js
    test/mjsunit/es6/arguments-iterator.js
    test/mjsunit/es6/string-iterator.js
    test/mjsunit/es6/collection-iterator.js


test/built-ins/Map/prototype/Symbol.iterator/property-descriptor.js
```diff
@@ -0,0 +1,14 @@
+// Copyright (C) 2014 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+  description: >
+      The method should exist on the Array prototype, and it should be writable
```
arv: s/Array/Map/

test/built-ins/Set/prototype/Symbol.iterator/prototype/next/iteration-mutable.js
```diff
@@ -0,0 +1,47 @@
+// Copyright (C) 2014 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+  description: >
+      When an item is added to the set after the iterator is created but
+      before the iterator is "done" (as defined by 23.2.5.2.1), the new item
+      should be accessible via iteration. When an item is added to the set
+      after the iterator is is "done", the new item should not be accessible
```
arv: s/is is/is/

test/built-ins/Set/prototype/values/iteration-mutable.js
```diff
@@ -0,0 +1,47 @@
+// Copyright (C) 2014 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+  description: >
+      When an item is added to the set after the iterator is created but before
+      the iterator is "done" (as defined by 23.2.5.2.1), the new item should be
+      accessible via iteration. When an item is added to the array after the
+      iterator is is "done", the new item should not be accessible via
```
arv: same typo here... maybe in more places?

test/built-ins/String/prototype/Symbol.iterator/prototype/ancestry.js
```diff
@@ -0,0 +1,15 @@
...
+    The [[Prototype]] internal slot ofthe %StringIteratorPrototype% is the
+    %IteratorPrototype% intrinsic object (25.1.2).
+---*/
+
+var strItrProto = Object.getPrototypeOf(String.prototype[Symbol.iterator]());
+var itrProto = Object.getPrototypeOf(
+  Object.getPrototypeOf([][Symbol.iterator]())
+);
+
+assert.sameValue(Object.getPrototypeOf(strItrProto), itrProto);
```
arv: This test will pass even if there is no intermediate %IteratorPrototype%. Maybe you can check the next [[Prototype]] too and make sure it is %ObjectPrototype%?

test/built-ins/String/prototype/Symbol.iterator/prototype/next/missing-internal-slots.js
```diff
@@ -0,0 +1,15 @@
...
+description: >
+    If the `this` value does not have all of the internal slots of an String
+    Iterator Instance (21.1.5.3), throw a `TypeError` exception.
+---*/
+
+var iterator = ''[Symbol.iterator]();
+var object = Object.create(iterator);
+
+assert.throws(TypeError, function() {
+  object.next();
```
arv: This is a valuable.

The other similar tests didn't have this case (where a valid instance is in the prototype chain).

test/built-ins/String/prototype/Symbol.iterator/prototype/ancestry.js
jugglinmike: Good point. That's really an assertion against the %IteratorPrototype% intrinsic, though. I'd like to include it in a file explicitly labeled for the iterator prototype. The alternative would be to re-write the assertion for every %___IteratorPrototype%, but I think that would be overkill. What do you think?

test/built-ins/String/prototype/Symbol.iterator/prototype/next/missing-internal-slots.js
jugglinmike: Totally. Since these tests are submitted under the banner of "Import from V8", I'd like to hold off on including brand new tests for now. I've made the following note to come back to this:
> - Native iterators
>   - `next` methods should throw a `TypeError` if the context does not have all the internal slots of the appropriate iterator interface (e.g. [22.1.5.2.1 %ArrayIteratorPrototype%.next( )](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-%arrayiteratorprototype%.next), step 3). Specifically test this is the case even when such an object is in the context's prototype chain. See [this comment](https://github.com/tc39/test262/pull/210#discussion_r27394897) for context.

test/built-ins/Map/prototype/Symbol.iterator/property-descriptor.js
jugglinmike: Got it

test/built-ins/Set/prototype/values/iteration-mutable.js
jugglinmike: Yes, I repeated this typo a number of times. [The changeset](https://github.com/bocoup/test262/commit/13fe84559330e6b4b09c3c6f94c39cb5ed8255f0) is a little noisy because the correction required re-wrapping the text in some cases.

jugglinmike:
Thanks, @arv!

test/built-ins/Map/prototype/Symbol.iterator/prototype/Symbol.toStringTag/property-descriptor.js
```diff
@@ -0,0 +1,18 @@
...
+
+/*---
+  description: >
+      `Object.prototype.getOwnPropertyDescriptor` should reflect the value and
+      writability of the @@toStringTag attribute.
+  includes: [propertyHelper.js]
+  es6id: 23.1.3.13
+ ---*/
+
+var SetIteratorProto = Object.getPrototypeOf(new Map()[Symbol.iterator]());
```
bterlson: MapIteratorProto :)

test/language/for-of/arguments-object-aliasing.js
```diff
@@ -0,0 +1,25 @@
+// Copyright (C) Copyright 2014 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 13.6.4
+description: >
+    Arguments objects should be able to be traversed using a `for..of` loop,
+    and dynamic changes to their contents should be reflected in the iterated
+    values.
+---*/
```
bterlson: I think this needs noStrict since it depends on sloppy arguments.

bterlson:
Organizationally, I'm having second thoughts about the depth of folders under the Symbol.iterator path. I think it would be okay to not have any subdirectories under feature\prototype\Symbol.iterator. This would help those of us on windows who have to worry about such things as max path lengths ;)

Other than the couple minor issues above, this is good to squash and check in.

test/language/for-of/arguments-object-aliasing.js
arv: Yes. Needs `noStrict` since only sloppy mode has mapped arguments object

jugglinmike:
> Organizationally, I'm having second thoughts about the depth of folders under the Symbol.iterator path. I think it would be okay to not have any subdirectories under feature\prototype\Symbol.iterator. This would help those of us on windows who have to worry about such things as max path lengths ;)

I was tempted to flatten the organization by moving all files to the respective "Symbol.iterator" directory and naively re-naming each to include the information previously encoded in the path. eg

    from:
    test/language/arguments-object/mapped/Symbol.iterator/prototype/next/expansion-after-exhaustion.js
    to:
    test/language/arguments-object/mapped/Symbol.iterator/prototype-next-expansion-after-exhaustion.js

This would seem to satisfy your expectations for general file organization, but it does nothing to address your concern about path length limitations. All that information seems necessary to me, though. I suppose substituting "proto" for "prototype" would help a little, but I don't know if that's sufficient. What do you think?

domenic:
I think Windows users like myself just need to suck it up and use tools that can cope.

bterlson:
@jugglinmike While I agree with Domenic generally, I'm not sure why it's super important that the file system encode all of the information you want. It should be in the test description or otherwise encoded into the frontmatter. But I like the shallower proposal with more info encoded into the file name rather than parent directories. If you want to go that route I'm fine with it - I'll suck it up (and maybe rename my `C:\test262\` to `C:\t` to save a few characters ;))

jugglinmike:
> I'm not sure why it's super important that the file system encode all of the information you want.

Sorry, I don't mean to come across as stubborn. I think maybe I'm just not being imaginative enough. The only naming scheme I can think of besides "describe test content" is "sequential numbering". This seems arbitrary and potentially confusing because the names will strongly suggest an ordering which is irrelevant. That said, I don't mean to make a production over this. I'm happy to use the convention that you prefer; I just want to make sure I understand it, first.

bterlson:
I guess the principle I'm operating under is shallow folder structure (no nested folders under a feature location unless really necessary) and a shorter name that conveys what's inside a test is better than a longer one that does the same. IMO 'expansion-after-exhaustion' or 'next-expansion-after-exhaustion' is better than 'prototype-next-expansion-after-exhaustion' but I am fine with whatever you think is best!

(It's impossible to be seen as stubborn working on Test262 IMO. Pedantry comes part-and-parcel with the job :-D)

jugglinmike:
Alright--files renamed!

test/built-ins/String/prototype/Symbol.iterator/Symbol.toStringTag.js
```diff
@@ -0,0 +1,20 @@
...
+es6id: 21.1.5.2.1
+description: >
+    The initial value of the @@toStringTag property is the string value "String
+    Iterator". This property has the attributes { [[Writable]]: false,
+    [[Enumerable]]: false, [[Configurable]]: true }.
+includes: [propertyHelper.js]
+---*/
+
+var StringIteratorProto = Object.getPrototypeOf(
+  String.prototype[Symbol.iterator]()
```
anba: `new String()[Symbol.iterator]()` or `""[Symbol.iterator]()` to avoid a `TypeError` with the current (rev37) semantics. (`String.prototype[Symbol.iterator]()` calls `String.prototype.toString()` which throws an error because `String.prototype` is not a string object.)

test/built-ins/String/prototype/Symbol.iterator/ancestry.js
```diff
@@ -0,0 +1,15 @@
+// Copyright (C) Copyright 2014 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 21.1.5.2
+description: >
+    The [[Prototype]] internal slot ofthe %StringIteratorPrototype% is the
+    %IteratorPrototype% intrinsic object (25.1.2).
+---*/
+
+var strItrProto = Object.getPrototypeOf(String.prototype[Symbol.iterator]());
```
anba: Same issue here.

test/built-ins/String/prototype/Symbol.iterator/Symbol.toStringTag.js
jugglinmike: I'm surprised I missed this, given how much I've been looking at Annex E lately. Thank you!

bterlson:
I think this is good after a squash.

jugglinmike:
Squashed!

Issue #215, opened 2015-03-30
YAML issues with indentated front matter?


arv:
The following files have indented the front matter:

built-ins/Array/prototype/Symbol.iterator/property-descriptor.js
built-ins/Array/prototype/Symbol.iterator/prototype/Symbol.toStringTag/property-descriptor.js
built-ins/Array/prototype/Symbol.iterator/prototype/next/property-descriptor.js
built-ins/Array/prototype/entries/property-descriptor.js
built-ins/Array/prototype/keys/property-descriptor.js

This is causing issues on one of our (v8) bots. It is probably picking up a YAML parser that does not work as expected.

@smikes

smikes:
Thanks!  I will take a look at it. If it is using the fallback parser (monkeyYaml) then indeed, it's vulnerable to indentation.

smikes:
Candidate fix in PR #216 ; unfortunately my environment has the 'real' YAML parser so the easiest way to test is to throw it to the wolves...

PR #216, opened 2015-03-30
candidate fix for #215

smikes:
This addresses the issue with property-descriptor.js; testing others now

EDIT: to be specific, the issue with "built-ins/Array/prototype/keys/property-descriptor.js".

### DIFF
--- a/tools/packaging/monkeyYaml.py
+++ b/tools/packaging/monkeyYaml.py
@@ -90,7 +90,7 @@ def myFlowList(value):
 def myMultiline(lines, value):
     # assume no explcit indentor (otherwise have to parse value)
     value = []
-    indent = 1
+    indent = myLeadingSpaces(lines[0])
     while lines:
         line = lines.pop(0)
         if myIsAllSpaces(line):
--- a/tools/packaging/test/test_monkeyYaml.py
+++ b/tools/packaging/test/test_monkeyYaml.py
@@ -58,13 +58,28 @@ class TestMonkeyYAMLParsing(unittest.TestCase):
         self.assertEqual(lines, [])
         self.assertEqual(value, yaml.load(y))
 
-    def test_Multiline_1(self):
+    def test_Multiline_2(self):
         lines = [" foo", " bar"]
         value = ">"
         y = "\n".join([value] + lines)
         (lines, value) = monkeyYaml.myMultiline(lines, value)
         self.assertEqual(lines, [])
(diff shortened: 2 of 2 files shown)


tools/packaging/test/test_monkeyYaml.py
```diff
@@ -66,6 +66,14 @@ def test_Multiline_1(self):
         self.assertEqual(lines, [])
         self.assertEqual(value, yaml.load(y))
 
+    def test_Multiline_2(self):
```
arv: This should be 3 and the previous one should be 2 (right now there are 2 test_Multiline_1)

tools/packaging/test/test_monkeyYaml.py
```diff
@@ -66,6 +66,14 @@ def test_Multiline_1(self):
...
+        lines = ["  foo", "  bar"]
+        value = ">"
+        y = "\n".join([value] + lines)
+        (lines, value) = monkeyYaml.myMultiline(lines, value)
+        self.assertEqual(lines, [])
+        self.assertEqual(value, yaml.load(y))
```
arv: I just got this working too :-)

Here is my test which tests the broken case:

```py
    def test_Multiline_3(self):
        lines = ["    foo", "    bar", "  other: 42"]
        value = ">"
        (lines, value) = monkeyYaml.myMultiline(lines, value)
        self.assertEqual(lines, ["  other: 42"])
        self.assertEqual(value, "foo bar")
```

tools/packaging/monkeyYaml.py
```diff
@@ -90,7 +90,7 @@ def myFlowList(value):
 def myMultiline(lines, value):
     # assume no explcit indentor (otherwise have to parse value)
     value = []
-    indent = 1
+    indent = -1
```
arv: Can this be?

```py
indent = myLeadingSpaces(lines[0])
```

or is there a case where lines is empty?

smikes: It's legal for lines to be empty.

smikes: Nope, I'm wrong.  Switching it.

smikes:
Incorporated your test, fixed the flow there (thanks!), squashed and rebased.

arv:
LGTM

Thank you for the fast fix.

smikes:
I'm just happy that the (very bad) python that I wrote was readable enough that you were willing to dive in and fix it too.

PR #218, opened 2015-04-02
Import tests from Google V8 (templates)

jugglinmike:
project:

    test/mjsunit/es6/templates.js

Some of these tests include non-printable characters, causing git to
infer that they are binary files. Introduce a `.gitattributes` file to
configure git to consistently display the source text of all JavaScript
files.


test/built-ins/String/raw/special-characters.js
```diff
@@ -0,0 +1,21 @@
...
+  'Unicode escape sequences'
+);
+assert.sameValue(
+  String.raw`\
\
+\
+`,
+  '\\\n\\\n\\\n',
+  'Literal characters'
+);
```
arv: Missing `);`

arv: nm, the diff was stupid

test/language/templates/expressions/function-invocation.js
```diff
@@ -0,0 +1,11 @@
+// Copyright (C) Copyright 2014 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 12.2.8
+description: >
+    Expressions should be evaluated and converted to Stings accordingto the
```
arv: s/Stings/Strings/

arv: s/accordingto/according to/

test/language/templates/expressions/legacy-octal-escape-sequence-non-strict.js
```diff
@@ -0,0 +1,13 @@
...
+es6id: 12.2.8
+description: >
+    The expression within the template should be evaluated according to the
+    semantics of the surrounding context.
+    The SV of EscapeSequence :: HexEscapeSequence is the SV of the
+    HexEscapeSequence.
+flags: [noStrict]
+---*/
+
+assert.sameValue(`${'\07'}`, '');
```
arv: This is Annex B stuff? How do we handle that in general?

test/language/templates/expressions/method-invocation.js
```diff
@@ -0,0 +1,13 @@
+// Copyright (C) Copyright 2014 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 12.2.8
+description: >
+    Expressions should be evaluated and converted to Stings accordingto the
```
arv: same typo

test/language/templates/tagged/call-expression-argument-list-evaluation.js
```diff
@@ -0,0 +1,38 @@
...
+calls = 0;
+(function(site, n, s, o, f, r) {
+  calls++;
+  assert.sameValue(arguments.length, 6);
+  assert.sameValue(n, number);
+  assert.sameValue(s, string);
+  assert.sameValue(o, object);
+  assert.sameValue(f, fn);
+  assert.sameValue(r, 'result');
+  assert.sameValue(arguments.length, 6, 'TemplateHead arguments length');
```
arv: repeat of line 30

test/language/templates/tagged/template-object.js
```diff
@@ -0,0 +1,47 @@
...
+    defined by the GetTemplateObject abstract operation.
+includes: [propertyHelper.js]
+---*/
+var templateObject;
+
+(function(parameter) {
+  templateObject = parameter;
+})`${1}`;
+
+templateObject.test262Prop = true;
```
arv: this will throw in strict mode

jugglinmike: Good point. To limit duplication, I've removed just the assertions for "frozen-ness" and created two new files for them: one runs only in strict mode and expects an error, the other runs only in non-strict mode and simply asserts that the assignment is not honored.

test/language/templates/tagged/call-expression-argument-list-evaluation.js
jugglinmike: I removed line 30 since it lacks an assertion message

test/language/templates/expressions/method-invocation.js
jugglinmike: Grepping the changeset, this shows up 6 times. I've corrected all occurrences.

test/language/templates/expressions/function-invocation.js
jugglinmike: Grepping the changeset, this shows up 6 times. I've corrected all occurrences.

test/built-ins/String/raw/special-characters.js
jugglinmike: Yeah, I've run into a few instances where non-printable characters break GitHub's UI while writing these tests!

test/language/templates/expressions/legacy-octal-escape-sequence-non-strict.js
jugglinmike: I don't know that. Do you have any thoughts on this, @bterlson?

caitp: annex b stuff belongs in test/annexB, I think :> --- no reason template-related stuff couldn't go in there?

jugglinmike: Definitely. I should have known this because I've already added some tests there (for some reason, I was thinking it was Annex E...)

anba:
`language/templates/template-values/non-escape-character.js` should be removed for now because it's not covered by the spec (see also https://bugs.ecmascript.org/show_bug.cgi?id=3477).

bterlson:
templates/expressions --> expressions/template
templates/tagged --> expressions/template (or expressions/tagged-template, although I think it's fine to club them together)

test/language/templates/template-values/line-continuation.js
```diff
@@ -0,0 +1,19 @@
...
+var calls;
+
+calls = 0;
+(function(cs) {
+  calls++;
+  assert.sameValue(cs.raw[0], '\u005C\n\u005C\n\u005C\n');
+})`\
+\
+\
`
```
bterlson: I think this template literal should also test <LS>, <PS>, and <CR><LF> (similar to line-terminator-sequence?)

test/language/templates/template-values/line-terminator-sequence.js
```diff
@@ -0,0 +1,20 @@
+// Copyright (C) Copyright 2014 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 11.8.6
+description: >
+    The TRV of LineTerminatorSequence :: <LF> is the code unit value 0x000A.
+    The TRV of LineTerminatorSequence :: <CR> is the code unit value 0x000A.
+    The TRV of LineTerminatorSequence :: <CR><LF> is the sequence consisting of
+    the code unit value 0x000A.
```
bterlson: Add <LS> and <PS>?

jugglinmike: Sorry, I don't understand. Can you rephrase?

bterlson: See the antepenultimate and penultimate bullets of [11.8.6.1](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-static-semantics-tv-s-and-trv-s)

jugglinmike: Ah, now I understand. (I think GitHub was initially having some trouble with the `<` and `>` characters in your comment.) This batch of tests is ported from V8 directly, but I can certainly include more assertions. I'll just introduce them with a standalone commit to avoid confusion.

jugglinmike:
I've addressed all the feedback I received. Thanks for the review!

test/annexB/templates/legacy-octal-escape-sequence-non-strict.js
```diff
@@ -0,0 +1,13 @@
...
+es6id: 12.2.8
+description: >
+    The expression within the template should be evaluated according to the
+    semantics of the surrounding context.
+    The SV of EscapeSequence :: HexEscapeSequence is the SV of the
+    HexEscapeSequence.
+flags: [noStrict]
+---*/
+
+assert.sameValue(`${'\07'}`, '');
```
caitp: I'm still kind of keen on using hex escapes or unicode escapes for the expected values here --- I know it means using another language feature to test this one, but it's a lot easier to see what's going on on github, and in many text editors, too.

bterlson:
Merged without squash. Oh well :-P

test/annexB/templates/legacy-octal-escape-sequence-non-strict.js
bterlson: These tests are really hard to debug. I would be ok with using escapes for expected values.

PR #223, opened 2015-04-08
Re-structure generator tests

jugglinmike:
@bterlson This is how I'm interpreting [your advice for generator test re-structuring](https://github.com/tc39/test262/pull/204#issuecomment-90769610) for the existing generators tests. (I'm proposing it in a separate pull request to help manage the process of landing new tests vs. updating existing tests.)

I've gone out on a limb with the tests for generator method definition. These are neither statements nor expressions, so I'm suggesting a new directory `language/method-definition/`, and (further) `language/method-definition/object/` to account for future tests for method definitions that occur within class bodies (as the semantics are different).


bterlson:
Hmm, didn't realize I pulled in some of these generator tests already.

I think your limb is solid WRT method-definition. I would probably be ok with putting method definition inside the object literal expr directory, but they should be easy to move from your proposal later if we want so I say go ahead :)

PR #224, opened 2015-04-08
14.5 Class Definitions

rwaldron:
- 14.5.1 Static Semantics: Early Errors
- 14.5.3 Static Semantics: ConstructorMethod
- 14.5.14 Runtime Semantics: ClassDefinitionEvaluation


test/language/class/syntax/default-class-declaration.js
```diff
@@ -0,0 +1,9 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 14.5
+description: >
+    ClassDeclaration:
+      [+Default] class ClassTail
+---*/
+export default class {}
```
arv: I don't think the test harness is set up to handle Module yet.

rwaldron: It should be able to parse this and do nothing with it.

arv: But you need to tell the VM that this is a Module production and not a Script.

rwaldron: Ah, yes you're right. I overlooked that aspect. Thanks.

rwaldron: Removed for now

test/language/class/subclass/class-definition-superclass-generator.js
```diff
@@ -0,0 +1,21 @@
...
+/*---
+es6id: 14.5.14_S6.g.i
+description: >
+    Runtime Semantics: ClassDefinitionEvaluation
+
+    If superclass has a [[FunctionKind]] internal slot whose value is "generator", throw a TypeError exception.
+---*/
+function * G() {}
+
+class A extends G {
```
anba: The `TypeError` is already thrown during 14.5.14 ClassDefinitionEvaluation, that means the complete class declaration needs to be moved into the `assert.throws` function. (And `new A();` can be removed.)

rwaldron: Fixed

test/language/class/syntax/class-declaration-binding-identifier-class-element-list.js
```diff
@@ -0,0 +1,35 @@
...
+      MethodDefinition[?Yield]
+      static MethodDefinition[?Yield]
+      ;
+
+
+---*/
+class A {
+  method() {}
+  static method() {}
+  ;
```
domenic: This is really allowed!? O_O I guess the grammar above does imply it...

arv: What do you find weird about it?

domenic: There's no need for semicolons in the class body, so why allow them? They don't delimit statements or anything.

domenic: It seems as weird as allowing backticks or commas or forward slashes or any other random punctuation mark.

arv: We talked about it. One motivation was that once you have class properties it makes sense to allow semicolons everywhere.

rwaldron: Right, it's also roughly the same as this: 

```
EmptyStatement :
;
```

ie. 

```js
;
```

Except that it's inside the `ClassBody`. 

> We talked about it. One motivation was that once you have class properties it makes sense to allow semicolons everywhere.

- https://github.com/zenparsing/es-private-fields
- https://gist.github.com/jeffmo/054df782c05639da2adb

domenic: Why can't we just add this weirdness when/if those proposals ever get accepted?

bterlson:
gah I thought I commented on this already but I guess I didn't.

Can we remove the block-scope "feature" and move those tests to language/class?

Otherwise collateral looks good!

(I'll probably reorganize this section into statements/class eventually)

rwaldron:
> Can we remove the block-scope "feature" and move those tests to language/class?

Yes

> (I'll probably reorganize this section into statements/class eventually)

I'll just move it now

test/language/class/subclass/null.js
```diff
@@ -1,18 +0,0 @@
...
-
-class Foo extends N {
-  constructor(x, y) {
-    assert.sameValue(x, 1, "The value of `x` is `1`");
-    assert.sameValue(y, 2, "The value of `y` is `2`");
-    return {};
-  }
-}
-
-new Foo(1,2);
```
rwaldron: Removed because this is handled in more specific tests now

rwaldron:
@bterlson all set.

bterlson:
rwaldron: alright, looks good, thanks :) I was going to do the move because there's a bunch of other class tests in language/class (and presumably some of them are class expression tests), but this is fine for now too!

PR #226, opened 2015-04-13
Invalid language tests

anba:
- in-statement-position-label-statement.js is an Annex-B test (B.3.2 Labelled Function Declarations), moved to annexB directory
- identifier-let-allowed-as-lefthandside-expression-strict.js needs to check for a SyntaxError (ES6, 12.1.1)
- method-*.js: compareArray() no longer sorts its input, added Array.prototype.sort() calls

### DIFF
--- a/test/annexB/labelled-function-declaration.js
+++ b/test/annexB/labelled-function-declaration.js
@@ -3,8 +3,9 @@
 /*---
 es6id: 13.1
 description: >
-    function declarations in statement position in strict mode:
+    function declarations in statement position in non-strict mode:
     label: Statement
+flags: [noStrict]
 ---*/
 label: function g() {}
 
--- a/test/built-ins/Object/getOwnPropertyNames/15.2.3.4-4-2.js
+++ b/test/built-ins/Object/getOwnPropertyNames/15.2.3.4-4-2.js
@@ -15,7 +15,6 @@ includes:
 function testcase() {
   var result = Object.getOwnPropertyNames(Object);
   var expResult = ["getPrototypeOf", "getOwnPropertyDescriptor", "getOwnPropertyNames", "create", "defineProperty", "defineProperties", "seal", "freeze", "preventExtensions", "isSealed", "isFrozen", "isExtensible", "keys", "prototype", "length"];
-  var found;
 
   return arrayContains(result, expResult);
  }
--- a/test/built-ins/Object/getOwnPropertyNames/15.2.3.4-4-44.js
+++ b/test/built-ins/Object/getOwnPropertyNames/15.2.3.4-4-44.js
@@ -19,9 +19,9 @@ function testcase() {
   var str = new String("abc");
   str[5] = "de";
 
-  var expected = ["0", "1", "2", "length", "5"];
+  var expected = ["0", "1", "2", "5", "length"];
   var actual = Object.getOwnPropertyNames(str);
 
-  return compareArray(actual.sort(), expected.sort());
+  return compareArray(actual, expected);
 }
 runTestCase(testcase);
--- a/test/built-ins/Object/getOwnPropertyNames/15.2.3.4-4-49.js
+++ b/test/built-ins/Object/getOwnPropertyNames/15.2.3.4-4-49.js
@@ -19,6 +19,6 @@ function testcase() {
   var expected = ["0", "1", "2", "length"];
   var actual = Object.getOwnPropertyNames(arr);
 
-  return compareArray(actual.sort(), expected.sort());
+  return compareArray(actual, expected);
 }
 runTestCase(testcase);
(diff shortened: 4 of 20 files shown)


anba:
I think adding `sort()` calls in test/language/computed-property-names/class/static/method-*.js makes more sense than checking for the property order as defined in the spec. WDYT? 

For example `Object.getOwnPropertyNames(class C { static a(){} static [1]() {} })` is supposed to return `[ "1", "length", "prototype", "a", "name" ]` according to the current draft. Reason:
- "1" as the first entry because it's an integer indexed property (9.1.12, step 2)
- followed by "length", "prototype", "a", "name" (string properties in creation order, 9.1.12, step 3)
  - "length" added when the function is initialized (9.2.4 FunctionInitialize, step 3)
  - "prototype" added when the function is made a constructor (9.2.8 MakeConstructor, step 6)
  - "a" added in 14.5.14 Runtime Semantics: ClassDefinitionEvaluation, step 21
  - "name" added in 14.5.16 Runtime Semantics: Evaluation, step 5.c.i

bterlson:
I agree with your read on the ordering of properties. I think we need a strong reason not to test it, especially for new language constructs that shouldn't have much of a legacy compat issue. How would you defend continuing to allow implementations leeway on enumeration order?

rwaldron:
> I think adding sort() calls in test/language/computed-property-names/class/static/method-*.js makes more sense than checking for the property order as defined in the spec. WDYT?

I agree. 

Also, I think these ended up breaking as a result of the Array tests having that change in limbo for so long. Sorry I didn't catch that before all the PRs landed.

bterlson:
> I agree. 

Why? Test262 needs really strong reasoning to give implementations a pass...

caitp:
isn't the spec pretty clear on the order of Object.getOwnPropertyNames? (and the order computed property keys are added to objects, and...)

bterlson:
@caitp I believe so (@anba spelled out his reasoning above). ES6 I believe should be completely standardized WRT enumeration order of OWN properties. I think we should test for it if we can get away with it.

rwaldron:
> Why? Test262 needs really strong reasoning to give implementations a pass...

I have zero explanation or excuse for why I completely ignored looking at [[OwnPropertyKeys]] when I looked at https://people.mozilla.org/~jorendorff/es6-draft.html#sec-getownpropertykeys . I take it back, I do not agree. I also apologize for causing two people to have to spend time correcting something I should've looked at further. I can only say: long day and the spec is wearing out my brain.

bterlson:
Ok, so I think we should remove the sort(). @anba, what do you think?

anba:
> Ok, so I think we should remove the sort(). @anba, what do you think?

First of all, I didn't expect that the strict interpretation of the specification w.r.t. property creation order has so many supporters. I'm really surprised! :open_mouth:

Apart from that, I'd prefer not to mix tests for property order and support for computed property names in a single file.

allenwb:
Differing property ordering among implementations would be a potential source of non-determinist behavior when  running on different implementations. The specification's intent is that implementations must follow the (usually implicitly) specified property creation ordering. This is something that  should be explicitly tested.

There is one big hole in this regard.  I don't believe we specified the creation order for the properties of built-in objects. And it isn't valid to assume that it is the order they are defined in the spec. as that ordering changed from ES5 to ES6.  This is probably something that needs to be addressed in the ES7 spec.

caitp:
That is a bit of a hole --- but luckily it shouldn't affect these O.gOPN tests

bterlson:
Property enumeration order tests of various sorts are pervasive enough that I think we should not attempt to avoid them for practical reasons. Also this PR removes what I think is useful coverage and requires sort to be implemented properly (which seems like a similar problem to testing enumeration order - we're testing that sort works properly).

I'm pretty bullish on enumeration order tests because this is an area of historical deviations (and extremely difficult to find bugs!) and I want to make sure it doesn't happen again :) I say keep the enumeration order validation!

anba:
I'm going to add the following three files to this PR:

[yield-as-identifier-in-nested-function.js](https://github.com/tc39/test262/blob/b9e4a43f5e4e0d3d0d6a0972bd6a7ece0efb7c20/test/language/class/method-definition/yield-as-identifier-in-nested-function.js#L17): Class code is implicitly strict, so nested functions are also strict and `yield` is not a valid identifier in strict mode.

[yield-as-function-expression-binding-identifier.js](https://github.com/tc39/test262/blob/b9e4a43f5e4e0d3d0d6a0972bd6a7ece0efb7c20/test/language/class/method-definition/yield-as-function-expression-binding-identifier.js#L16): Class code is implicitly strict, so nested functions are also strict and `yield` is not a valid binding identifier in strict mode.

[yield-as-function-expression-binding-identifier.js](https://github.com/tc39/test262/blob/b9e4a43f5e4e0d3d0d6a0972bd6a7ece0efb7c20/test/language/expressions/object/method-definition/yield-as-function-expression-binding-identifier.js#L20): Typo `A.prototype` -> `obj`.

anba:
Rebased and updated the commit. New changes:
- Removed sort() calls in various tests
- Fixed some more invalid tests

bterlson:
going to pull this in, I think you've addressed all the things you said you would :) (this is awesome btw!)

PR #228, opened 2015-04-15
14.2 Arrow Function Definitions

rwaldron:


test/language/arrow-function/lexical-bindings-overriden-by-formal-parameters-non-strict.js
```diff
@@ -0,0 +1,14 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 14.2.16
+description: >
+    Runtime Semantics: Evaluation
+
+flags: [noStrict]
```
arv: Why the no strict?

caitp: it would be a syntax error in strict mode (parameter named "arguments")

arv: Sorry, I completely misread what this was testing. I'll pay more attention from now on.

arv:
Wow. LGTM. Very extensive set of tests.

I don't see a test that ensures that the arrow function does not have an own "prototype" property. Maybe that already exists?

test/language/arrow-function/lexical-bindings-overriden-by-formal-parameters-non-strict.js
rwaldron: :+1:

rwaldron:
> Maybe that already exists?

@arv that's correct, they are here: https://github.com/tc39/test262/blob/master/test/language/arrow-function/Arrow-Function_rules-for-prototype.js

test/language/arrow-function/Arrow-Function_semantics.js
```diff
@@ -1,35 +0,0 @@
...
-// Expression Body implicit return
-var evens = [0, 2, 4, 6, 8];
-assert(compareArray(evens.map(v => v + 1), [1, 3, 5, 7, 9]));
-
-// Statement body needs braces, must use 'return' explicitly if not void
-var fives = [];
-[1, 2, 3, 4, 5, 6, 7, 8, 9, 10].forEach(v => {
-  if (v % 5 === 0) fives.push(v);
-});
-assert(compareArray(fives, [5, 10]));
```
rwaldron: This file was eliminated because the semantics tested here are now spread out over several files.

bterlson:
Let's move to language/exprs/arrow-function, but otherwise these tests are amazing and good to go I think!

rwaldron:
@bterlson done!

PR #231, opened 2015-04-17
19.1.2.10 Object.is

rwaldron:
7.2.9 SameValue(x, y)

### DIFF
--- a/test/built-ins/Object/is/not-same-value-x-y-boolean.js
+++ b/test/built-ins/Object/is/not-same-value-x-y-boolean.js
@@ -1,0 +1,30 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 19.1.2.10
+description: >
+    Object.is ( value1, value2 )
+
+    7.2.9 SameValue(x, y)
+
+    ...
+    8. If Type(x) is Boolean, then
+      a. If x and y are both true or both false,
+          return true; otherwise, return false.
+---*/
+
+assert.sameValue(Object.is(true, false), false, "`Object.is(true, false)` returns `false`");
+assert.sameValue(Object.is(false, true), false, "`Object.is(false, true)` returns `false`");
+assert.sameValue(Object.is(true, 1), false, "`Object.is(true, 1)` returns `false`");
+assert.sameValue(Object.is(false, 0), false, "`Object.is(false, 0)` returns `false`");
+assert.sameValue(Object.is(true, {}), false, "`Object.is(true, {})` returns `false`");
+assert.sameValue(Object.is(true, undefined), false, "`Object.is(true, undefined)` returns `false`");
+assert.sameValue(Object.is(false, undefined), false, "`Object.is(false, undefined)` returns `false`");
+assert.sameValue(Object.is(true, null), false, "`Object.is(true, null)` returns `false`");
+assert.sameValue(Object.is(false, null), false, "`Object.is(false, null)` returns `false`");
+assert.sameValue(Object.is(true, NaN), false, "`Object.is(true, NaN)` returns `false`");
+assert.sameValue(Object.is(false, NaN), false, "`Object.is(false, NaN)` returns `false`");
+assert.sameValue(Object.is(true, ''), false, "`Object.is(true, '')` returns `false`");
+assert.sameValue(Object.is(false, ''), false, "`Object.is(false, '')` returns `false`");
+assert.sameValue(Object.is(true, []), false, "`Object.is(true, [])` returns `false`");
+assert.sameValue(Object.is(false, []), false, "`Object.is(false, [])` returns `false`");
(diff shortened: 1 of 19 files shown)


test/built-ins/Object/is/same-value-x-y-boolean.js
```diff
@@ -0,0 +1,28 @@
...
+    8. If Type(x) is Boolean, then
+      a. If x and y are both true or both false, 
+          return true; otherwise, return false.
+---*/
+
+assert.sameValue(Object.is(true, true), true, "`Object.is(true, true)` returns `true`");
+assert.sameValue(Object.is(false, false), true, "`Object.is(false, false)` returns `true`");
+
+assert.sameValue(
+    Object.is(Boolean(true), Boolean(true)),
```
arv: Not sure this adds anything... it just tests that `Boolean(true) === true` and `Boolean(false) === false`.

test/built-ins/Object/is/same-value-x-y-number.js
```diff
@@ -0,0 +1,21 @@
...
+    ...
+    6. If Type(x) is Number, then
+      a. If x is NaN and y is NaN, return true.
+      b. If x is +0 and y is -0, return false.
+      c. If x is -0 and y is +0, return false.
+      d. If x is the same Number value as y, return true.
+      e. Return false.    
+    ...
+---*/
+
```
arv: Maybe add tests for Infinity and -Infinity as well?

arv:
LGTM

test/built-ins/Object/is/same-value-x-y-boolean.js
rwaldron: Removed.

test/built-ins/Object/is/same-value-x-y-number.js
rwaldron: Will do

test/built-ins/Object/is/not-same-value-x-y-number.js
```diff
@@ -0,0 +1,21 @@
...
+      c. If x is -0 and y is +0, return false.
+      d. If x is the same Number value as y, return true.
+      e. Return false.
+    ...
+---*/
+
+assert.sameValue(Object.is(+0, -0), false, "`Object.is(+0, -0)` returns `false`");
+assert.sameValue(Object.is(-0, +0), false, "`Object.is(-0, +0)` returns `false`");
+assert.sameValue(Object.is(0), false, "`Object.is(0)` returns `false`");
+assert.sameValue(Object.is(Infinity, -Infinity), true, "`Object.is(Infinity, -Infinity)` returns `false`");
```
bterlson: I would expect false here (-Inf and +Inf seem like distinct number values to me). Thoughts?

test/built-ins/Object/is/not-same-value-x-y-boolean.js
```diff
@@ -0,0 +1,26 @@
...
+assert.sameValue(Object.is(false, true), false, "`Object.is(false, true)` returns `false`");
+assert.sameValue(Object.is(true, 1), false, "`Object.is(true, 1)` returns `false`");
+assert.sameValue(Object.is(false, 0), false, "`Object.is(false, 0)` returns `false`");
+assert.sameValue(Object.is(true, {}), false, "`Object.is(true, {})` returns `false`");
+assert.sameValue(Object.is(true, undefined), false, "`Object.is(true, undefined)` returns `false`");
+assert.sameValue(Object.is(false, undefined), false, "`Object.is(false, undefined)` returns `false`");
+assert.sameValue(Object.is(true, null), false, "`Object.is(true, null)` returns `false`");
+assert.sameValue(Object.is(false, null), false, "`Object.is(false, null)` returns `false`");
+assert.sameValue(Object.is(true, NaN), false, "`Object.is(true, NaN)` returns `false`");
+assert.sameValue(Object.is(false, NaN), false, "`Object.is(false, NaN)` returns `false`");
```
bterlson: Probably good to check all falsey values just to be sure - maybe add empty string and empty array?

test/built-ins/Object/is/not-same-value-x-y-number.js
rwaldron: That's a typo (note the assertion message)

test/built-ins/Object/is/not-same-value-x-y-boolean.js
rwaldron: Good call, adding.

bterlson:
Ideas for additional coverage:
* Object.is length and descriptor properties
* Object.is behavior when called with 0 and 1 parameters (same as passing undefined I guess?)

bterlson:
Otherwise looks good.

rwaldron:
> Object.is behavior when called with 0 and 1 parameters (same as passing undefined I guess?)

- 0 parameters: 
  - https://github.com/tc39/test262/pull/231/files#diff-5cefd17720bf43572502b937753b0938
- 1 parameter: 
  - https://github.com/tc39/test262/pull/231/files#diff-5a135a02445c3aba3c810807505172b0
  - https://github.com/tc39/test262/pull/231/files#diff-82e3b7127fbc67d065a90b7feeb66732

I can more calls with 1 parameter where the first argument is something that might be misunderstood by a bad implementation?

bterlson:
I think what's there now is good!

PR #232, opened 2015-04-19
remove non-printable characters from quoted strings in tests

caitp:


caitp:
https://github.com/tc39/test262/pull/218/files#r28427330 for context --- `grep '[^[:print:]]' -r test/` finds a lot of other instances of this though, which can't all be removed

PR #234, opened 2015-04-21
Fix bug in `isWritable` utility function

jugglinmike:
Only attempt to re-set the property value in cases where it was
successfully modified as part of the function's execution. This avoids
errors when the underlying value is not writable. Rename the internal
result-tracking variable to make this more clear.


jugglinmike:
This URL renders a slightly more readable diff: https://github.com/tc39/test262/pull/234/files?w=1

rwaldron:
LGTM

bterlson:
LGTM after squash :)

jugglinmike:
All set

smikes:
Thanks, I just ran into this and almost opened a new PR but then I got your change. ;-)

PR #235, opened 2015-04-21
Correct error messages

jugglinmike:
The `testBuiltInObject` function verifies the presence of the specified
properties for *all* object types (not just function objects). Update
the error messages to reflect this.


arv:
LGTM

PR #236, opened 2015-04-21
Add Tests for Harness Helper Functions

jugglinmike:
Introduce tests for code in the `harness/` directory. This is limited to what I thought were the most critical files in that directory (`assert.js`, `compareArray.js`, `testBuiltInObject.js`, and `propertyHelper.js`), but I'd be happy to add tests for any others.

For the `verifyEnumerable` and `verifyConfigurable` assertions, I've created distinct files for strict mode and non-strict mode (because that distinction matters for the internal execution). It might be better to combine these and just use a more robust test runner (i.e. one that can run tests whose mode is unspecified in *both* strict and non-strict mode).

There are a few specific details that I couldn't find a way to validate. If anyone has any suggestions about these, I'm all ears!

- [testBuiltInObject.js](https://github.com/tc39/test262/blob/6d945e2a1e83c716cf86ab200b51958c197f5ff3/harness/testBuiltInObject.js)
  - function objects' prototypes (as in `(function() {}).__proto__`, not `(function() {}).prototype`) ([untested code](https://github.com/tc39/test262/blob/6d945e2a1e83c716cf86ab200b51958c197f5ff3/harness/testBuiltInObject.js#L40-L42))
  - plain objects' prototypes (again, `({}).__proto__`) ([untested code](https://github.com/tc39/test262/blob/6d945e2a1e83c716cf86ab200b51958c197f5ff3/harness/testBuiltInObject.js#L44-L46))
  - function lengths--incorrect property descriptors ([untested code](https://github.com/tc39/test262/blob/6d945e2a1e83c716cf86ab200b51958c197f5ff3/harness/testBuiltInObject.js#L65-L74))
  - function lengths--non-integer values ([untested code](https://github.com/tc39/test262/blob/6d945e2a1e83c716cf86ab200b51958c197f5ff3/harness/testBuiltInObject.js#L56-L58))
  - function objects without `prototype` properties ([untested code](https://github.com/tc39/test262/blob/6d945e2a1e83c716cf86ab200b51958c197f5ff3/harness/testBuiltInObject.js#L117-L119))
- [propertyHelper.js](https://github.com/tc39/test262/blob/6d945e2a1e83c716cf86ab200b51958c197f5ff3/harness/propertyHelper.js)
  - non-`TypeError` thrown in response to deleting a configurable property ([untested code](https://github.com/tc39/test262/blob/6d945e2a1e83c716cf86ab200b51958c197f5ff3/harness/propertyHelper.js#L6-L8))
  - non-`TypeError` thrown in response to setting a writable property ([untested code](https://github.com/tc39/test262/blob/6d945e2a1e83c716cf86ab200b51958c197f5ff3/harness/propertyHelper.js#L39-L41))

### DIFF
--- a/test/harness/assert-throws-native.js
+++ b/test/harness/assert-throws-native.js
@@ -1,0 +1,36 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: >
+    Functions that throw instances of the specified native Error constructor
+    satisfy the assertion.
+---*/
+
+assert.throws(Error, function() {
+  throw new Error();
+});
+
+assert.throws(EvalError, function() {
+  throw new EvalError();
+});
+
+assert.throws(RangeError, function() {
+  throw new RangeError();
+});
+
+assert.throws(ReferenceError, function() {
+  throw new ReferenceError();
+});
+
+assert.throws(SyntaxError, function() {
+  throw new SyntaxError();
+});
+
+assert.throws(TypeError, function() {
+  throw new TypeError();
+});
+
+assert.throws(URIError, function() {
+  throw new URIError();
+});
--- a/test/harness/compare-array-empty.js
+++ b/test/harness/compare-array-empty.js
@@ -1,0 +1,12 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: >
+    Empty arrays of are equivalent.
+includes: [compareArray.js]
+---*/
+
+if (compareArray([], []) !== true) {
+  $ERROR('Empty arrays are equivalent.');
+}
--- a/test/harness/assert-false.js
+++ b/test/harness/assert-false.js
@@ -1,0 +1,25 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: >
+    `false` does not satisfy the assertion.
+---*/
+
+var threw = false;
+
+try {
+  assert(false);
+} catch(err) {
+  threw = true;
+  if (err.constructor !== Test262Error) {
+    $ERROR(
+      'Expected a Test262Error, but a "' + err.constructor.name +
+      '" was thrown.'
+    );
+  }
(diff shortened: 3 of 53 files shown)


test/harness/assert-throws-native.js
```diff
@@ -0,0 +1,36 @@
...
+  throw new SyntaxError();
+});
+
+assert.throws(TypeError, function() {
+  throw new TypeError();
+});
+
+assert.throws(URIError, function() {
+  throw new URIError();
+});
```
arv: What about?

```js
assert.throws(Error, function() {
  throw new TypeError();
});
```

I hope this would fail since we should not use instanceof in these kind of tests.

test/harness/compare-array-empty.js
```diff
@@ -0,0 +1,12 @@
...
+
+/*---
+description: >
+    Empty arrays of are equivalent.
+includes: [compareArray.js]
+---*/
+
+if (compareArray([], []) !== true) {
+  $ERROR('Empty arrays are equivalent.');
+}
```
arv: What about holes?

bterlson:
LGTM!!

test/harness/assert-throws-native.js
jugglinmike: Agreed; this is already covered by https://github.com/bocoup/test262/blob/30ad8abdb5fea908c5f825f3b269752e873cb58d/test/harness/assert-throws-incorrect-ctor.js

```js
/*---
description: >
Functions that throw values whose constructor does not match the specified
constructor do not satisfy the assertion.
negative: Test262Error
---*/
assert.throws(Error, function() {
  throw new TypeError();
});
```

test/harness/compare-array-empty.js
jugglinmike: Good idea! https://github.com/bocoup/test262/commit/668368ef4d2ef014f7125d55d69297940744e776

jugglinmike:
@bterlson as discussed in IRC, I've re-factored the tests to use explicit `try..catch` statements instead of the `negative` frontmatter (which we hope to reserve for early errors).

In addition, I've added a test for the global `$ERROR` function--how did I miss that one?

If this all looks good to you, I'll squash it all into one commit

bterlson:
Squash away!

jugglinmike:
Ready to go :)

PR #237, opened 2015-04-23
Generators semantics

jugglinmike:
@bterlson I'm submitting this as two commits to separate the tests that are derived directly from V8 from the analogous tests I added to fill out the coverage

### DIFF
--- a/test/built-ins/GeneratorPrototype/next/context-constructor-invocation.js
+++ b/test/built-ins/GeneratorPrototype/next/context-constructor-invocation.js
@@ -1,0 +1,16 @@
+// Copyright (C) 2013 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 25.2
+description: >
+    If the generator was invoked using [[Construct]], the this bind is not
+    initialized and any references to `this` within the FunctionBody will
+    produce a ReferenceError exception.
+---*/
+
+function* g() { this; }
+var iter = new g();
+
+assert.throws(ReferenceError, function() {
+  iter.next();
+});
--- a/test/built-ins/GeneratorPrototype/next/incorrect-context.js
+++ b/test/built-ins/GeneratorPrototype/next/incorrect-context.js
@@ -1,0 +1,22 @@
+// Copyright (C) 2013 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 25.3.1.2
+description: >
+    A TypeError should be thrown from GeneratorValidate (25.3.3.2) if the
+    context of `next` does not defined the [[GeneratorState]] internal slot.
+---*/
+
+function* g() {}
+var GeneratorPrototype = Object.getPrototypeOf(g).prototype;
+
+assert.throws(TypeError, function() { GeneratorPrototype.next.call(1); });
+assert.throws(TypeError, function() { GeneratorPrototype.next.call(1, 1); });
+assert.throws(TypeError, function() { GeneratorPrototype.next.call({}); });
+assert.throws(TypeError, function() { GeneratorPrototype.next.call({}, 1); });
+assert.throws(TypeError, function() { GeneratorPrototype.next.call(function() {}); });
+assert.throws(TypeError, function() { GeneratorPrototype.next.call(function() {}, 1); });
+assert.throws(TypeError, function() { GeneratorPrototype.next.call(g); });
+assert.throws(TypeError, function() { GeneratorPrototype.next.call(g, 1); });
+assert.throws(TypeError, function() { GeneratorPrototype.next.call(g.prototype); });
+assert.throws(TypeError, function() { GeneratorPrototype.next.call(g.prototype, 1); });
--- a/test/built-ins/GeneratorPrototype/throw/incorrect-context.js
+++ b/test/built-ins/GeneratorPrototype/throw/incorrect-context.js
@@ -1,0 +1,17 @@
+// Copyright (C) 2013 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 25.3.1.4
+description: >
+    A TypeError should be thrown from GeneratorValidate (25.3.3.2) if the
+    context of `throw` does not defined the [[GeneratorState]] internal slot.
+---*/
+
+function* g() {}
+var GeneratorPrototype = Object.getPrototypeOf(g).prototype;
+
+assert.throws(TypeError, function() { GeneratorPrototype.throw.call(1); });
+assert.throws(TypeError, function() { GeneratorPrototype.throw.call({}); });
+assert.throws(TypeError, function() { GeneratorPrototype.throw.call(function() {}); });
+assert.throws(TypeError, function() { GeneratorPrototype.throw.call(g); });
+assert.throws(TypeError, function() { GeneratorPrototype.throw.call(g.prototype); });
--- a/test/built-ins/GeneratorPrototype/throw/try-catch-before-try.js
+++ b/test/built-ins/GeneratorPrototype/throw/try-catch-before-try.js
@@ -1,0 +1,35 @@
+// Copyright (C) 2013 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 25.3.1.4
+description: >
+    When a generator is paused before a `try..catch` statement, `throw` should
+    interrupt control flow as if a `throw` statement had appeared at that
+    location in the function body.
+---*/
+
+function* g() {
+  yield 1;
+  try {
+    yield 2;
+  } catch (e) {
+    yield e;
+  }
+  yield 3;
+}
+var iter, result, exception;
+
+iter = g();
+
+result = iter.next();
+assert.sameValue(result.value, 1, 'First result `value`');
+assert.sameValue(result.done, false, 'First result `done` flag');
+assert.throws(Test262Error, function() { iter.throw(new Test262Error()); });
+
+result = iter.next();
+assert.sameValue(result.value,
+  undefined, 'Result `value` is undefined when done'
+);
+assert.sameValue(result.done, true, 'Result `done` flag is `true` when done');
+
+iter.next();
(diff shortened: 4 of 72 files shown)


test/built-ins/GeneratorPrototype/next/incorrect-context.js
```diff
@@ -0,0 +1,22 @@
...
+es6id: 25.3.1.2
+description: >
+    A TypeError should be thrown from GeneratorValidate (25.3.3.2) if the
+    context of `next` does not defined the [[GeneratorState]] internal slot.
+---*/
+
+function* g() {}
+var iter = g();
+
+assert.throws(TypeError, function() { iter.next.call(1); });
```
arv: Would it be better to test this with `%IteratorPrototype%.next`?

arv: Sorry, `%GeneratorPrototype%.

test/built-ins/GeneratorPrototype/throw/incorrect-context.js
```diff
@@ -0,0 +1,17 @@
...
+es6id: 25.3.1.4
+description: >
+    A TypeError should be thrown from GeneratorValidate (25.3.3.2) if the
+    context of `throw` does not defined the [[GeneratorState]] internal slot.
+---*/
+
+function* g() {}
+var iter = g();
+
+assert.throws(TypeError, function() { iter.throw.call(1); });
```
arv: same here?

test/built-ins/GeneratorPrototype/throw/try-catch-before-try.js
```diff
@@ -0,0 +1,35 @@
...
+  yield 3;
+}
+var iter, result, exception;
+
+iter = g();
+
+result = iter.next();
+assert.sameValue(result.value, 1, 'First result `value`');
+assert.sameValue(result.done, false, 'First result `done` flag');
+assert.throws(Test262Error, function() { iter.throw(new Test262Error()); });
```
arv: Use a different constructor function here

test/built-ins/GeneratorPrototype/throw/try-catch-following-catch.js
```diff
@@ -0,0 +1,43 @@
...
+  try {
+    yield 2;
+  } catch (e) {
+    yield e;
+  }
+  yield 3;
+}
+var iter, result, exception;
+
+exception = new Test262Error();
```
arv: It seems bad to use Test262Error for these kind of tests.

test/language/expressions/generators/length-property-descriptor.js
```diff
@@ -0,0 +1,15 @@
...
+    Generator objects should define a `length` property.
+includes: [propertyHelper.js]
+es6id: 25.2.4
+---*/
+
+var g = function*() {};
+
+verifyNotEnumerable(g, 'length');
+verifyNotWritable(g, 'length');
+verifyConfigurable(g, 'length');
```
arv: Verify that the length is 0?

test/language/expressions/generators/name-property-descriptor.js
```diff
@@ -0,0 +1,15 @@
...
+    Generator functions should define a `name` property.
+includes: [propertyHelper.js]
+es6id: 25.2.4
+---*/
+
+var g = function*() {};
+
+verifyNotEnumerable(g, 'name');
+verifyNotWritable(g, 'name');
+verifyConfigurable(g, 'name');
```
arv: Verify that the name is 'g'?

test/language/statements/generators/length-property-descriptor.js
```diff
@@ -0,0 +1,15 @@
...
+    Generator functions should define a `length` property.
+includes: [propertyHelper.js]
+es6id: 25.2.4
+---*/
+
+function* g() {}
+
+verifyNotEnumerable(g, 'length');
+verifyNotWritable(g, 'length');
+verifyConfigurable(g, 'length');
```
arv: same

test/language/statements/generators/name-property-descriptor.js
```diff
@@ -0,0 +1,15 @@
...
+    Generator functions should define a `name` property.
+includes: [propertyHelper.js]
+es6id: 25.2.4
+---*/
+
+function* g() {}
+
+verifyNotEnumerable(g, 'name');
+verifyNotWritable(g, 'name');
+verifyConfigurable(g, 'name');
```
arv: same

test/built-ins/GeneratorPrototype/throw/try-catch-following-catch.js
bterlson: The cat is already out of the bag. The python harness (somehow) always injects a definition of Test262Error into the test environment, and many of the recent tests use it. I just had to update test262-harness to work with this yesterday :)

arv: My concern is that there might be some false positives (or negatives) when using these but maybe I'm worrying too much?

test/built-ins/GeneratorPrototype/next/incorrect-context.js
jugglinmike: That would simplify things a bit (no implicit prototype resolution and no need to create an iterable). It would introduce a dependency on `Object.getPrototypeOf`, but that's probably still better overall. I'm pushing up a commit to address this now!

test/language/expressions/generators/name-property-descriptor.js
jugglinmike: There's a separate test for the value of the `name` property in `master` already: https://github.com/tc39/test262/blob/c49e2bd6b902aa699bedeabcf6e84adf99688a99/test/language/expressions/generators/name.js

@bterlson Do you think it's worthwhile to keep these assertions separate? Or should I merge the existing file into this one?

test/language/expressions/generators/length-property-descriptor.js
jugglinmike: Unlike `name`, there aren't any tests for the value of the length property. I agree it's worth adding a test now, but I'll wait on Brian's response to https://github.com/tc39/test262/pull/237#discussion_r29007374 before acting on this (since I want to include the new assertions in a way that is consistent with the others).

test/built-ins/GeneratorPrototype/throw/try-catch-following-catch.js
bterlson: I can't think of a case where you'd get false behavior. It's equivalent in risk to making your own error constructor I think.

test/language/expressions/generators/name-property-descriptor.js
bterlson: Seems ok to be separate. But generally, a descriptor contains a value property, so it makes some sense to combine them - you're testing all the descriptor properties.

jugglinmike:
I've addressed all your feedback, @arv. Thanks for the review!

test/built-ins/GeneratorPrototype/next/context-constructor-invocation.js
```diff
@@ -0,0 +1,16 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 25.2
+description: >
+    When a generator function is invoked as a constructor, its context is the
+    newly-created iterable object.
+---*/
+
+var context;
+function* g() { context = this; }
```
anba: Access to `this` in new'ed generators throws a ReferenceError. (Generators are created as derived constructor functions.)

test/language/expressions/yield/star-iterable.js
```diff
@@ -0,0 +1,43 @@
+// Copyright (C) 2013 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 25.2
+description: >
+    When an iterator is the operand of a `yield *` expression, the generator
+    should produce an iterator that visits each iterated item.
+---*/
+
+var results = [{ value: 1 }, 13, 'foo', { value: 34, done: true }];
```
anba: Non-object types throw a TypeError, cf. 7.4.2 IteratorNext step 4.

test/language/expressions/generators/name-property-descriptor.js
```diff
@@ -0,0 +1,16 @@
...
+/*---
+description: >
+    Generator functions should define a `name` property.
+includes: [propertyHelper.js]
+es6id: 25.2.4
+---*/
+
+var g = function*() {};
+
+assert.sameValue(g.name, '');
```
anba: `g.name` is "g".

test/built-ins/GeneratorPrototype/next/context-constructor-invocation.js
jugglinmike: This caught me off guard. Doing my due diligence, I verified in the spec:

> **9.2.3 FunctionAllocate (functionPrototype, strict [,functionKind] )**
> ...
> 9. If needsConstruct is true, then
>    a. Set F’s [[Construct]] internal method to the definition specified in
>       9.2.2.
>    b. If functionKind is "generator", set the [[ConstructorKind]] internal
>       slot of F to "derived".
>    c. Else, set the [[ConstructorKind]] internal slot of F to "base".
>    d. NOTE Generator functions are tagged as "derived" constructors to
>       prevent [[Construct]] from preallocating a generator instance.
>       Generator instance objects are allocated when EvaluateBody is applied
>       to the GeneratorBody of a generator function.

and:

> **9.2.2 `[[Construct]]` ( argumentsList, newTarget)**
> ...
> 5. If kind is "base", then
>    a. Let thisArgument be OrdinaryCreateFromConstructor(newTarget,
>       "%ObjectPrototype%").
>    b. ReturnIfAbrupt(thisArgument).

Thanks!

test/language/expressions/yield/star-iterable.js
jugglinmike: Right you are. I think that behavior warrants a dedicated test, so I think I will simplify this to return a valid object for each iteration.

jugglinmike:
@bterlson I've squashed this patch into two distinct commits to maintain the distinction between the tests inspired by the V8 test suite directly, and the tests I added to fill out coverage a bit.

PR #239, opened 2015-04-26
negative tests always fail when ran by test262.py

rniwa:
Don't fail when a negative test finishes with a pass state since gs.js returns a pass state.

### DIFF
--- a/tools/packaging/test262.py
+++ b/tools/packaging/test262.py
@@ -218,7 +218,7 @@ class TestResult(object):
     if self.case.IsAsyncTest():
        return self.AsyncHasFailed() or self.HasFailed()
     elif self.case.IsNegative():
-       return not (self.HasFailed() and self.case.NegativeMatch(self.GetErrorOutput()))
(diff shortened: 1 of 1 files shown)


tools/packaging/test262.py
```diff
@@ -218,7 +218,7 @@ def HasUnexpectedOutcome(self):
     if self.case.IsAsyncTest():
        return self.AsyncHasFailed() or self.HasFailed()
     elif self.case.IsNegative():
-       return not (self.HasFailed() and self.case.NegativeMatch(self.GetErrorOutput()))
+       return not (not self.HasFailed() and self.case.NegativeMatch(self.GetErrorOutput()))
```
rniwa: We could, instead, do:
return self.HasFailed() or not self.case.NegativeMatch(self.GetErrorOutput())

rniwa:
Oh, never mind. This is a problem with run-jsc not passing non-zero exit code around.

bterlson:
There is also https://github.com/bterlson/test262-harness in case it is helpful to you.

PR #240, opened 2015-04-27
Update remaining tests for ES2015 compatibility (rev38)

anba:
- Duplicate properties are allowed in object literals: test/language/expressions/object/*.js
- RegExpBuiltinExec was changed to use ToLength(lastIndex): test/built-ins/RegExp/prototype/exec/S15.10.6.2_A5_T3.js
- Non-undefined flags arguments allowed in RegExp constructor call: test/built-ins/RegExp/*.js
- Array.prototype.push throws TypeError if new length exceeds Number.MAX_SAFE_INTEGER: test/built-ins/Array/prototype/push/S15.4.4.7_A2_T2.js
- .length property of bound functions is configurable: test/built-ins/Function/prototype/bind/15.3.4.5-15-2.js
- Array.prototype changed back to exotic Array object: test/built-ins/Array/prototype/*.js, test/built-ins/Array/isArray/15.4.3.2-0-5.js

### DIFF
--- a/test/built-ins/Array/isArray/15.4.3.2-0-5.js
+++ b/test/built-ins/Array/isArray/15.4.3.2-0-5.js
@@ -14,7 +14,7 @@ includes: [runTestCase.js]
 
 function testcase() {
   var b = Array.isArray(Array.prototype);
-  if (b === false) {
+  if (b === true) {
     return true;
   }
  }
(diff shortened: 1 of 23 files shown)


bterlson:
Thanks!!

PR #241, opened 2015-04-28
23.2 Set Objects

rwaldron:


anba:
set-iterator-close-after-add-failure.js is a duplicate of set-close-iterator-after-failure.js

test/built-ins/Set/prototype/Symbol.iterator/toStringTag-property-descriptor.js
```diff
@@ -11,7 +11,11 @@
 
 var SetIteratorProto = Object.getPrototypeOf(new Set()[Symbol.iterator]());
 
-assert.sameValue('Set Iterator', SetIteratorProto[Symbol.toStringTag]);
+assert.sameValue(
+ 'Set Iterator',
```
arv: strange indentation

test/built-ins/Set/prototype/add/add.js
```diff
@@ -0,0 +1,21 @@
...
+es6id: 23.2.3.1
+description: >
+    Set.prototype.add ( value )
+
+    17 ECMAScript Standard Built-in Objects
+
+includes: [propertyHelper.js]
+---*/
+
+assert.sameValue(
```
arv: Maybe we should have helper for validating standard methods.

We should also validate the `name` and `length` properties and that they do not have a `prototype` property.

```js
verifyBuiltinMethod(Set.prototype, 'add', 1)
```

test/built-ins/Set/set-iterator-close-after-add-failure.js
```diff
@@ -0,0 +1,40 @@
...
+---*/
+
+var count = 0;
+var iterable = {};
+iterable[Symbol.iterator] = function() {
+  return {
+    next: function() {
+      return { value: null, done: false };
+    },
+    'return': function() {
```
arv: this is Sparta^d^d^d^d^d^dES6 and we do not have to limit ourselves to ES3 silliness.

arv:
It is too bad we cannot use an control flow structures. Having to repeat all the tests for all 6 primitive types gets verbose.

test/built-ins/Set/prototype/Symbol.iterator/toStringTag-property-descriptor.js
rwaldron: Fixed. I can't tell if this was caused by recast or not :\

test/built-ins/Set/prototype/add/add.js
rwaldron: Sure, but in a different issue

rwaldron:
Is this blocked?

bterlson:
In the sense that I need to make sense of the results yet. I will hopefully get to this later today!

rwaldron:
Ok, that's reasonable

bterlson:
LGTM, squash away!

Issue #35, opened 2014-06-04
Need to run tests in both strict mode and non-strict mode


bterlson:
If a test doesn't indicate that it's specifically for sloppy or strict mode it should be run in both. This change should be made at least for the web harness, but also perhaps for the python harness.

erights:
What's the status of this?

anba:
> What's the status of this?

The numbers from [my earlier comment](https://github.com/tc39/test262/issues/35#issuecomment-66597897) still apply, so ~1200 failing tests.

bterlson:
@anba, any comments on how much further to go after your latest PRs?

anba:
> @anba, any comments on how much further to go after your latest PRs?

- ~240 files in built-ins/Object: All remaining Object tests require changes to use propertyHelper functions. 
- 72 files in language: I still need to decide how to handle tests using `eval`. I think I'll convert `eval("var x = 0;"); if (x !== 0) { ... }` to `var result = eval("var x = 0; x); if (result !== 0) { ... }`, but I haven't yet made a final decision. (See language/line-terminators and language/white-space for examples where this is necessary.)

smikes:
I can take on the Object tests , if that would help.

anba:
> I can take on the Object tests , if that would help.

Sure. Only a handful of test cases were changed in #249, [the rest](https://gist.github.com/anba/81787fc6c6d30c88a69d) still needs some work.

smikes:
Working on that this morning.

bterlson:
What's left? :-D

anba:
> What's left? :-D

All tests work now in both modes! :smile:  

The only thing left is updating the python runner to use the new default. 

```diff
diff --git a/tools/packaging/test262.py b/tools/packaging/test262.py
index b9172b0..ed8939e 100755
--- a/tools/packaging/test262.py
+++ b/tools/packaging/test262.py
@@ -66,9 +66,7 @@ def BuildOptions():
                     help="Test only strict mode")
   result.add_option("--non_strict_only", default=False, action="store_true",
                     help="Test only non-strict mode")
-  # TODO: Once enough tests are made strict compat, change the default
-  # to "both"
-  result.add_option("--unmarked_default", default="non_strict",
+  result.add_option("--unmarked_default", default="both",
                     help="default mode for tests of unspecified strictness")
   result.add_option("--logname", help="Filename to save stdout to")
   result.add_option("--junitname", help="Filename to save test results in JUnit XML format")
```

[...]

PR #242, opened 2015-04-29
Fix strict mode errors in built-ins/Array

anba:
- Move code which requires non-strict semantics from Array.prototype.find_this-arg.js to Array.prototype.find_this-arg-receiver-primitive.js
- Remove duplicate code from Array.prototype.findIndex_this-arg.js, already present in Array.prototype.findIndex_this-arg-receiver-primitive.js
- Remove test code from Array.of_assignment-to-new-object-length.js, does not work in strict mode

Part of issue #35


bterlson:
Thanks so much @anba, this PR looks great. I'll slowly work my way through the others!

PR #244, opened 2015-04-29
Fix strict mode errors in built-ins/Date

anba:

Part of issue #35.

### DIFF
--- a/test/built-ins/Date/UTC/S15.9.4.3_A1_T1.js
+++ b/test/built-ins/Date/UTC/S15.9.4.3_A1_T1.js
@@ -7,7 +7,7 @@ es5id: 15.9.4.3_A1_T1
 description: Checking absence of ReadOnly attribute
 ---*/
 
-x = Date.UTC;
+var x = Date.UTC;
 if(x === 1)
   Date.UTC = 2;
 else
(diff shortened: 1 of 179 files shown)


bterlson:
@anba, a lot of the noStrict tests are a result of checking attributes - do you think it's possible to update those tests to use the attribute helpers (eg. verifyWritable, etc.)? This would allow the test to run in both modes... I understand if this isn't something you want to do as part of these PRs, but what do yo think?

anba:
> I understand if this isn't something you want to do as part of these PRs, but what do yo think?

Yes, I prefer a separate issue to track this. While working on the last PRs, I've noticed a few noStrict flags which are no longer valid. My plan is to check all test cases for invalid noStrict flags and update them so they work in both modes. 

And it's a pre-existing issue:

https://github.com/tc39/test262/blob/5d6f33d242ca8ad902a3aa545190575dd2be846b/test/built-ins/Object/prototype/isPrototypeOf/S15.2.4.6_A10.js#L12
https://github.com/tc39/test262/blob/5d6f33d242ca8ad902a3aa545190575dd2be846b/test/built-ins/Object/prototype/propertyIsEnumerable/S15.2.4.7_A10.js#L12

bterlson:
I am happy tracking this as a separate issue!

PR #245, opened 2015-04-29
Fix strict mode errors in built-ins/Error

anba:

Part of issue #35.


bterlson:
Don't forget the typos you fixed too ;)

PR #249, opened 2015-04-29
Fix some strict mode errors in built-ins/Object

anba:
- Remove invalid noStrict flags
- ietestcenter tests need fixed, require updates to use propertyHelper.js

Part of issue #35.


arv:
LGTM

PR #251, opened 2015-04-29
Fix strict mode errors in built-ins/String

anba:
- Remove with-statements (minor change in behaviour - primitive vs wrapper lookup, but does not seem too important for these tests)

Part of issue #35.


test/built-ins/String/prototype/indexOf/S15.5.4.7_A4_T1.js
```diff
@@ -12,21 +12,17 @@ includes: [$FAIL.js]
 
 var __obj = {toString:function(){return "\u0041B";}}
 var __obj2 = {valueOf:function(){throw "intointeger";}}
-var __str = {str__:"ABB\u0041BABAB"};
+var __str = "ABB\u0041BABAB";
 
 //////////////////////////////////////////////////////////////////////////////
 //CHECK#1
-with(__str){
-    with(str__){
```
arv: What were they thinking?

arv:
LGTM

test/built-ins/String/prototype/indexOf/S15.5.4.7_A4_T1.js
bterlson: I feel like all those with statements must have been for something but I can't imagine what....

arv: Maybe their codegen used them because it was simpler?

Anyway. Good riddance!

anba: It seems to be some case of: Let's use additional language features to increase the test surface. :confused:

PR #253, opened 2015-04-29
Fix strict mode errors in {decode, encode}URI[Component]

anba:

Part of issue #35.


arv:
LGTM

PR #254, opened 2015-05-02
make Object / property tests work with testStrict=true

smikes:
First 34 completed (see PR).  working on rest

PR #255, opened 2015-05-02
rewrite tests to work in strict mode, use helpers

smikes:
use helper fn; need to remove writeable test for non-own properties

Here are the rest of them.  The PR is messed up because I squashed it on top of the other one, #254 .  I don't know what's the best way to proceed with this - is there an easy way to drop the doubly-applied ones?

### DIFF
--- a/test/built-ins/Object/defineProperty/15.2.3.6-4-217.js
+++ b/test/built-ins/Object/defineProperty/15.2.3.6-4-217.js
@@ -13,29 +13,15 @@ description: >
 includes: [propertyHelper.js]
 ---*/
 
-        var arrObj = [];
+var arrObj = [];
 
-        Object.defineProperty(arrObj, "0", { value: NaN });
+Object.defineProperty(arrObj, "0", { value: NaN });
 
-        Object.defineProperty(arrObj, "0", { value: NaN });
+Object.defineProperty(arrObj, "0", { value: NaN });
 
-        var hasProperty = arrObj.hasOwnProperty("0");
-        var verifyValue = (arrObj[0] !== arrObj[0]);
+assert(arrObj.hasOwnProperty("0"));
+assert(arrObj[0] !== arrObj[0]);
 
-        var verifyWritable = false;
-        arrObj[0] = 1001;
-        verifyWritable = arrObj[0] !== 1001 && arrObj[0] !== arrObj[0];
-
-        var verifyEnumerable = false;
-        for (var p in arrObj) {
-            if (p === "0") {
-                verifyEnumerable = true;
-            }
-        }
-
-        var verifyConfigurable = false;
-        delete arrObj[0];
-        verifyConfigurable = arrObj.hasOwnProperty("0");
-
-        return hasProperty && verifyValue && verifyWritable && !verifyEnumerable && verifyConfigurable;
-    }
+verifyNotWritable(arrObj, "0");
+verifyNotEnumerable(arrObj, "0");
+verifyNotWritable(arrObj, "0");
--- a/test/built-ins/Object/defineProperty/15.2.3.6-4-256.js
+++ b/test/built-ins/Object/defineProperty/15.2.3.6-4-256.js
@@ -16,40 +16,31 @@ description: >
 includes: [propertyHelper.js]
 ---*/
 
-        var arrObj = [];
-        function getFunc() {
-            return 12;
-        }
+var arrObj = [];
+function getFunc() {
+    return 12;
+}
 
-        Object.defineProperty(arrObj, "1", {
-            get: getFunc
-        });
+Object.defineProperty(arrObj, "1", {
+    get: getFunc
+});
 
-        try {
-            Object.defineProperty(arrObj, "1", {
-                get: undefined
-            });
-            return false;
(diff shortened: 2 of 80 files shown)


bterlson:
Hmm, you can try doing a rebase against master and it might be smart enough to merge. If not I can play around with it and see, not sure off-hand.

smikes:
Will try that. Worst case I can manually pick to a new branch

smikes:
I think it's fixed now, review welcome.

smikes:
This commit effectively does two things: 

 - fixes errors introduced by early merge of #254 
 - finishes making built-ins/Object tests that deal with properties strict-compliant

Separate PR #257 does only part one of this.  Ideally, #257 should **not** be merged, unless we need to get back to a "clean" state (pre-254) on the tests which got broken.

bterlson:
Ok so I don't mess this up again - you want this PR merged, which will fix the tests I accidentally broke by pulling in #254 and make further progress for strict-clean of O.dP and O.dPs?

Also to be clear there is still work left for O.dP and O.dPs correct? I am seeing some strict-mode specific failures yet. Just making sure we're on the same page :)

smikes:
Yes, I think this PR is the right one.

 I am available today to make further changes to fix this.  Sorry for the mixups!

smikes:
I believe that is true, yes.

bterlson:
Do you want me to pull this now and fix the remaining O.dP/O.dPs failures in a later PR or should I wait?

Also feel free to close #257 :)

smikes:
Okay, I will kill #257.

I only have one remaining failure, `test/built-ins/Object/defineProperty/15.2.3.6-4-354-8.js` which I believe is a v8 bug.  What problems are you seeing?

bterlson:
before I look at the others, here's one example I found - I am seeing built-ins/Object/defineProperty/15.2.3.6-4-85.js as not using the property helpers and failing in strict mode?

smikes:
I agree that both of those claims are true.  However, I was only attacking the files listed in this gist file ( https://gist.github.com/anba/81787fc6c6d30c88a69d )

bterlson:
Ahhhh ok. So future PR then and this one is good to go?

smikes:
I think so.  Working on a list of the remaining 30 or so broken tests for that next PR.

bterlson:
Alrighty, pulling this in then!

test/built-ins/Object/defineProperty/15.2.3.6-4-217.js
```diff
@@ -13,29 +13,15 @@ description: >
...
-
-        var verifyConfigurable = false;
-        delete arrObj[0];
-        verifyConfigurable = arrObj.hasOwnProperty("0");
-
-        return hasProperty && verifyValue && verifyWritable && !verifyEnumerable && verifyConfigurable;
-    }
+verifyNotWritable(arrObj, "0");
+verifyNotEnumerable(arrObj, "0");
+verifyNotWritable(arrObj, "0");
```
anba: Typo: verifyNotWritable -> verifyNotConfigurable

test/built-ins/Object/defineProperty/15.2.3.6-4-256.js
```diff
@@ -16,40 +16,31 @@ description: >
...
+    assert(arrObj.hasOwnProperty("1"));
+
+    var desc = Object.getOwnPropertyDescriptor(arrObj, "1");
+
+    assert(arrObj[1] === getFunc());
+    assert(desc.hasOwnProperty("set") && typeof desc.set === "undefined");
+
+    verifyNotWritable(arrObj, "1");
+
+    verifyNotWritable(arrObj, "1");
```
anba: -> verifyNotConfigurable

test/built-ins/Object/defineProperty/15.2.3.6-4-470.js
```diff
@@ -13,24 +13,26 @@ description: >
...
-        var propertyDefineCorrect = obj.hasOwnProperty("prop");
-        var desc = Object.getOwnPropertyDescriptor(obj, "prop");
+assert(obj.hasOwnProperty("prop"));
 
-        delete obj.prop;
+var desc = Object.getOwnPropertyDescriptor(obj, "prop");
 
-        return propertyDefineCorrect && desc.configurable === false && obj.hasOwnProperty("prop");
-    }
+verifyNotWritable(obj, "prop");
```
anba: verifyNotConfigurable ?

test/built-ins/Object/defineProperty/15.2.3.6-4-501.js
```diff
@@ -14,25 +14,29 @@ description: >
...
+Object.defineProperty(obj, "prop", {
+    configurable: false
+});
+
+var desc2 = Object.getOwnPropertyDescriptor(obj, "prop");
+
+assert.sameValue(desc1.configurable, true);
+assert.sameValue(desc2.configurable, false);
+
+verifyNotWritable(obj, "prop");
```
anba: verifyNotConfigurable ?

smikes:
Thank you - new PR #263 to address those.

PR #256, opened 2015-05-05
New tests for built-ins and language features

suwc:
Add new tests for built-ins and language features


suwc:
Breaking down into smaller pull requests.

rwaldron:
I understand that this was probably compiled mechanically, but there is no way to reasonably review 1694 files. I don't think this should land as-is. 

Other questions: 

- How much of this duplicates existing tests?
- What level of coverage should be expected?
- Why are the files named in the old convention?
- Will the proper `es6id` front matter be added?

I'll try to spend more time with this tomorrow.

Issue #258, opened 2015-05-06
Remaining tests broken in strict mode (built-ins/Object)


smikes:
Here is a gist with the remaining files (post #255)

https://gist.github.com/smikes/066f80ef651b47a0a289

smikes:
test/built-ins/Object/defineProperties/15.2.3.7-6-a-24.js has the known global-proxy-properties bug

smikes:
Changes merged in.

PR #259, opened 2015-05-06
fix tests in strict mode - Object/defineProperty

smikes:
Fixes additional strict mode issues noted in #258

### DIFF
--- a/test/built-ins/Object/defineProperty/15.2.3.6-4-591.js
+++ b/test/built-ins/Object/defineProperty/15.2.3.6-4-591.js
@@ -9,51 +9,61 @@ es5id: 15.2.3.6-4-591
 description: >
     ES5 Attributes - Fail to update value of property of
     [[Proptotype]] internal property (Object.create)
-includes: [runTestCase.js]
+includes: [propertyHelper.js]
 ---*/
 
-function testcase() {
-        var appointment = {};
+var appointment = {};
 
-        var data1 = 1001;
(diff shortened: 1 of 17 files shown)


bterlson:
Looks great!

test/built-ins/Object/defineProperty/15.2.3.6-4-591.js
```diff
@@ -9,51 +9,61 @@ es5id: 15.2.3.6-4-591
...
+
+var teamMeeting = Object.create(meeting);
+
+verifyNotWritable(teamMeeting, "name", "nocheck");
+verifyNotWritable(teamMeeting, "startTime", "nocheck");
+verifyNotWritable(teamMeeting, "conferenceCall", "nocheck");
+
+try {
+    teamMeeting.name = "IE Team Meeting";
+} catch (e) {}
```
anba: The empty catch blocks are kind of odd. And shouldn't the verifyNotWritable calls test this case?

smikes: I was a bit confused by this one, but I think it's a test that grew from a real user experience  (hence the unusually real-looking data).  

I added verifyNotWritable() because it is strict/nonstrict safe.  But I also chose to preserve the assignments (eg

```
teamMeeting.name = "IE Team Meeting";
```

I then had to wrap each of those in a `try` for strict mode.  I could add an assertion to the catch block to make it nonempty...

PR #262, opened 2015-05-07
Candidate fix of empty catch{}

smikes:
This is a candidate fix to the line note on #259
https://github.com/tc39/test262/pull/259#discussion_r29866745


bterlson:
Seems fine to me.

PR #267, opened 2015-05-11
Add tests for destructuring assignment

jugglinmike:
@bterlson This does not include coverage for destructuring binding patterns. Those tests will have so much in common with the tests for destructuring assignment that I thought it best to get feedback on these before starting.

### DIFF
--- a/test/language/expressions/assignment/destructuring/array-elem-init-order.js
+++ b/test/language/expressions/assignment/destructuring/array-elem-init-order.js
@@ -1,0 +1,19 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: >
+    Initializer values should be assigned in left-to-right order.
+es6id: 12.14.5.3
+---*/
+
+var value = [];
+var x = 0;
+var a, b, result;
+
+result = [ a = x += 1, b = x *= 2 ] = value;
+
+assert.sameValue(result, value);
+assert.sameValue(a, 1);
+assert.sameValue(b, 2);
+assert.sameValue(x, 2);
--- a/test/language/expressions/assignment/destructuring/array-elem-nested-array-yield-ident-valid.js
+++ b/test/language/expressions/assignment/destructuring/array-elem-nested-array-yield-ident-valid.js
@@ -1,0 +1,21 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: >
+    When a `yield` token appears within the DestructuringAssignmentTarget of a
+    nested destructuring assignment outside of strict mode, it behaves as an
+    IdentifierReference.
+es6id: 12.14.5.3
+flags: [noStrict]
+---*/
+
+var value = [[22]];
+var yield = 'prop';
+var x = {};
+var result;
+
+result = [[x[yield]]] = value;
+
+assert.sameValue(result, value);
+assert.sameValue(x.prop, 22);
--- a/test/language/expressions/assignment/destructuring/array-iteration.js
+++ b/test/language/expressions/assignment/destructuring/array-iteration.js
@@ -1,0 +1,53 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: >
+    Value iteration only proceeds for the number of elements in the
+    ArrayAssignmentPattern.
+es6id: 12.14.5.3
+features: [generators]
+---*/
+
+var count, iter, result;
+var g = function*() {
+  count += 1;
+  yield;
+  count += 1;
+  yield;
+  count += 1;
+}
+
+count = 0;
+iter = g();
+result = [] = iter;
+
+assert.sameValue(result, iter);
+assert.sameValue(count, 0);
+
+iter = g();
+result = [,] = iter;
+
+assert.sameValue(result, iter);
+assert.sameValue(count, 1);
+
+count = 0;
+iter = g();
+result = [,,] = iter;
+
+assert.sameValue(result, iter);
+assert.sameValue(count, 2);
+
+count = 0;
+iter = g();
+result = [,,,] = iter;
+
+assert.sameValue(result, iter);
+assert.sameValue(count, 3);
+
+count = 0;
+iter = g();
+result = [,,,,] = iter;
+
+assert.sameValue(result, iter);
+assert.sameValue(count, 3);
--- a/test/language/expressions/assignment/destructuring/array-rest-iteration.js
+++ b/test/language/expressions/assignment/destructuring/array-rest-iteration.js
@@ -1,0 +1,26 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: >
+    In the presense of an AssignmentRestElement, value iteration exhausts the
+    iterable value;
+es6id: 12.14.5.3
+features: [generators]
+---*/
+
+var count = 0;
+var g = function*() {
+  count += 1;
+  yield;
+  count += 1;
+  yield;
+  count += 1;
+}
+var iter, result, x;
+
+iter = g();
+result = [...x] = iter;
+
+assert.sameValue(result, iter);
+assert.sameValue(count, 3);
--- a/test/language/expressions/assignment/destructuring/array-rest-nested-obj-undefined.js
+++ b/test/language/expressions/assignment/destructuring/array-rest-nested-obj-undefined.js
@@ -1,0 +1,19 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: >
+    When DestructuringAssignmentTarget is an obect literal and the iterable is
+    emits no values, an empty array should be used as the value of the nested
+    DestructuringAssignment.
+es6id: 12.14.5.3
+---*/
+
+var value = [];
+var result, x, length;
+
+result = [...{ 0: x, length }] = value;
+
+assert.sameValue(result, value);
+assert.sameValue(x, undefined);
+assert.sameValue(length, 0);
--- a/test/language/expressions/assignment/destructuring/obj-id-init-order.js
+++ b/test/language/expressions/assignment/destructuring/obj-id-init-order.js
@@ -1,0 +1,19 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: >
+    Initializer values should be assigned in left-to-right order.
+es6id: 12.14.5.4
+---*/
+
+var value = {};
+var x = 0;
+var a, b, result;
+
+result = { a = x += 1, b = x *= 2 } = value;
+
+assert.sameValue(result, value);
+assert.sameValue(a, 1);
+assert.sameValue(b, 2);
+assert.sameValue(x, 2);
--- a/test/language/expressions/assignment/destructuring/obj-id-init-yield-expr.js
+++ b/test/language/expressions/assignment/destructuring/obj-id-init-yield-expr.js
@@ -1,0 +1,32 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: >
+    When a `yield` token appears within the Initializer of an
+    AssignmentProperty and within a generator function body, it should behave
+    as a YieldExpression.
+es6id: 12.14.5
+features: [generators]
+---*/
+
+var value = {};
+var assignmentResult, iterationResult, x, iter;
+
+iter = (function*() {
+  assignmentResult = { x = yield } = value;
+}());
+
+iterationResult = iter.next();
+
+assert.sameValue(iterationResult.value, undefined);
+assert.sameValue(iterationResult.done, false);
+assert.sameValue(x, undefined);
+assert.sameValue(assignmentResult, undefined);
+
+iterationResult = iter.next(3);
+
+assert.sameValue(iterationResult.value, undefined);
+assert.sameValue(iterationResult.done, true);
+assert.sameValue(x, 3);
+assert.sameValue(assignmentResult, value);
--- a/test/language/expressions/assignment/destructuring/obj-prop-nested-array-null.js
+++ b/test/language/expressions/assignment/destructuring/obj-prop-nested-array-null.js
@@ -1,0 +1,15 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: >
+    When DestructuringAssignmentTarget is an array literal and the value is
+    `null`, a TypeError should be thrown.
+es6id: 12.14.5.4
+---*/
+
+var x;
+
+assert.throws(TypeError, function() {
+  ({ x: [ x ] } = { x: null });
+});
--- a/test/language/expressions/assignment/destructuring/obj-prop-nested-array-undefined-own.js
+++ b/test/language/expressions/assignment/destructuring/obj-prop-nested-array-undefined-own.js
@@ -1,0 +1,15 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: >
+    When DestructuringAssignmentTarget is an array literal and the value is
+    `undefined`, a TypeError should be thrown.
+es6id: 12.14.5.4
+---*/
+
+var x;
+
+assert.throws(TypeError, function() {
+  ({ x: [ x ] } = { x: undefined });
+});
--- a/test/language/expressions/assignment/destructuring/obj-prop-nested-array-undefined.js
+++ b/test/language/expressions/assignment/destructuring/obj-prop-nested-array-undefined.js
@@ -1,0 +1,15 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: >
+    When DestructuringAssignmentTarget is an array literal and the value is not
+    defined, a TypeError should be thrown.
+es6id: 12.14.5.4
+---*/
+
+var x;
+
+assert.throws(TypeError, function() {
+  ({ x: [ x ] } = {});
+});
--- a/test/language/expressions/assignment/destructuring/obj-prop-nested-array-yield-expr.js
+++ b/test/language/expressions/assignment/destructuring/obj-prop-nested-array-yield-expr.js
@@ -1,0 +1,32 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: >
+    When a `yield` token appears within the Initializer of a nested
+    destructuring assignment and within a generator function body, it should
+    behave as a YieldExpression.
+es6id: 12.14.5.4
+features: [generators]
+---*/
+
+var value = { x: [] };
+var assignmentResult, iterationResult, iter, x;
+
+iter = (function*() {
+  assignmentResult = { x: [x = yield] } = value;
+}());
+
+iterationResult = iter.next();
+
+assert.sameValue(assignmentResult = undefined);
+assert.sameValue(iterationResult.value, undefined);
+assert.sameValue(iterationResult.done, false);
+assert.sameValue(x, undefined);
+
+iterationResult = iter.next(24601);
+
+assert.sameValue(assignmentResult, value);
+assert.sameValue(iterationResult.value, undefined);
+assert.sameValue(iterationResult.done, true);
+assert.sameValue(x, 24601);
--- a/test/language/expressions/assignment/destructuring/obj-prop-nested-obj-null.js
+++ b/test/language/expressions/assignment/destructuring/obj-prop-nested-obj-null.js
@@ -1,0 +1,15 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: >
+    When DestructuringAssignmentTarget is an object literal and the value is
+    `null`, a TypeError should be thrown.
+es6id: 12.14.5.4
+---*/
+
+var x;
+
+assert.throws(TypeError, function() {
+  ({ x: { x } } = { x: null });
+});
--- a/test/language/expressions/assignment/destructuring/obj-prop-nested-obj-undefined-own.js
+++ b/test/language/expressions/assignment/destructuring/obj-prop-nested-obj-undefined-own.js
@@ -1,0 +1,15 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: >
+    When DestructuringAssignmentTarget is an object literal and the value is
+    `undefined`, a TypeError should be thrown.
+es6id: 12.14.5.4
+---*/
+
+var x;
+
+assert.throws(TypeError, function() {
+  ({ x: { x } } = { x: undefined });
+});
--- a/test/language/expressions/assignment/destructuring/obj-prop-nested-obj-undefined.js
+++ b/test/language/expressions/assignment/destructuring/obj-prop-nested-obj-undefined.js
@@ -1,0 +1,15 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: >
+    When DestructuringAssignmentTarget is an object literal and the value is
+    not defined, a TypeError should be thrown.
+es6id: 12.14.5.4
+---*/
+
+var x;
+
+assert.throws(TypeError, function() {
+  ({ x: { x } } = {});
+});
--- a/test/language/expressions/assignment/destructuring/obj-prop-put-prop-ref-user-err.js
+++ b/test/language/expressions/assignment/destructuring/obj-prop-put-prop-ref-user-err.js
@@ -1,0 +1,20 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: >
+    Any error raised as a result of setting the value should be forwarded to
+    the runtime.
+es6id: 12.14.5.4
+---*/
+
+var value = { a: 23 };
+var x = {
+  set y(val) {
+    throw new Test262Error();
+  }
+};
+
+assert.throws(Test262Error, function() {
+  ({ a: x.y } = value);
+});
--- a/test/language/expressions/assignment/destructuring/array-elem-elision.js
+++ b/test/language/expressions/assignment/destructuring/array-elem-elision.js
@@ -1,0 +1,19 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: >
+    ArrayAssignmentPattern may include elisions at any position in a
+    AssignmentElementList that does not contain an AssignmentRestElement.
+es6id: 12.14.5
+---*/
+
+var value = [1, 2, 3, 4, 5, 6];
+var x, y;
+var result;
+
+result = [, x, , y, ,] = value;
+
+assert.sameValue(result, value);
+assert.sameValue(x, 2);
+assert.sameValue(y, 4);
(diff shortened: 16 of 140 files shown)


test/language/expressions/assignment/destructuring/obj-prop-put-prop-ref-user-err.js
```diff
@@ -0,0 +1,20 @@
...
+
+var value = { a: 23 };
+var x = {
+  set y(val) {
+    throw new Test262Error();
+  }
+};
+
+assert.throws(Test262Error, function() {
+  ({ a: x.y } = value);
```
caitp: Some bonus test cases: 12.14.5.1 for `DestructuringAssignmentTarget : LeftHandSideExpression`'s early errors, I don't see these tested but might be missing something

test/language/expressions/assignment/destructuring/obj-prop-nested-array-yield-expr.js
```diff
@@ -0,0 +1,32 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: >
+    When a `yield` token appears within the Initializer of a nested
+    destructuring assignment and within a generator function body, it should
+    behave as a YieldExpression.
+es6id: 12.14.5.4
```
caitp: I'm not sure where this description is coming from --- It's a syntax error trying to parse `[x = yield]` as an `AssignmentPattern[Yield]`

caitp: Nevermind, I take it back --- this is totally correct. Weird though!

test/language/expressions/assignment/destructuring/array-elem-init-order.js
```diff
@@ -0,0 +1,19 @@
...
+description: >
+    Initializer values should be assigned in left-to-right order.
+es6id: 12.14.5.3
+---*/
+
+var value = [];
+var x = 0;
+var result;
+
+result = [ a = x += 1, b = x *= 2 ] = value;
```
anba: Please add `var a, b;` for strict-mode compatibility (or mark test as noStrict).

test/language/expressions/assignment/destructuring/array-elem-nested-array-yield-ident-valid.js
```diff
@@ -0,0 +1,21 @@
...
+---*/
+
+var value = [[22]];
+var yield = 'prop';
+var x = {};
+var result;
+
+result = [[x[yield]]] = value;
+
+assert.sameValue(assignmentResult, value);
```
anba: Please add `assignmentResult;` for strict-mode compatibility.

test/language/expressions/assignment/destructuring/array-elem-put-constant-no-strict.js
```diff
@@ -0,0 +1,19 @@
...
+es6id: 12.14.5.3
+features: [const]
+flags: [noStrict]
+---*/
+
+var value = [1];
+const c = null;
+var result;
+
+result = [ c ] = value;
```
anba: Assignment to immutable strict binding throws a TypeError in non-strict (and strict) mode (8.1.1.1.5 SetMutableBinding, step 6). (`const` is always created as immutable strict binding, cf. 15.1.8 Runtime Semantics: GlobalDeclarationInstantiation, step 15.b.i.1).

test/language/expressions/assignment/destructuring/array-iteration.js
```diff
@@ -0,0 +1,53 @@
...
+  count += 1;
+  yield;
+  count += 1;
+  yield;
+  count += 1;
+}
+
+count = 0;
+iter = g();
+result = [] = iter;
```
anba: Please add `var result;` for strict-mode compatibility.

test/language/expressions/assignment/destructuring/array-rest-iteration.js
```diff
@@ -0,0 +1,26 @@
...
+  count += 1;
+  yield;
+  count += 1;
+  yield;
+  count += 1;
+}
+var iter;
+
+iter = g();
+result = [...x] = iter;
```
anba: Please add `var iter, x;` for strict-mode compatibility.

test/language/expressions/assignment/destructuring/array-rest-nested-obj-undefined.js
```diff
@@ -0,0 +1,19 @@
...
+es6id: 12.14.5.3
+---*/
+
+var value = [];
+var result, x, length;
+
+result = [...{ 0: x, length }] = value;
+
+assert.sameValue(result, value);
+assert.sameValue(x, null);
```
anba: Typo? `null` -> `undefined`

test/language/expressions/assignment/destructuring/array-rest-put-constant-no-strict.js
```diff
@@ -0,0 +1,19 @@
...
+es6id: 12.14.5.3
+features: [const]
+flags: [noStrict]
+---*/
+
+var value = [1];
+const c = null;
+var result;
+
+result = [ ...c ] = value;
```
anba: Assignment to immutable strict binding throws a TypeError in non-strict (and strict) mode (8.1.1.1.5 SetMutableBinding, step 6). (`const` is always created as immutable strict binding, cf. 15.1.8 Runtime Semantics: GlobalDeclarationInstantiation, step 15.b.i.1).

test/language/expressions/assignment/destructuring/obj-id-init-order.js
```diff
@@ -0,0 +1,19 @@
...
+description: >
+    Initializer values should be assigned in left-to-right order.
+es6id: 12.14.5.4
+---*/
+
+var value = {};
+var x = 0;
+var result;
+
+result = { a = x += 1, b = x *= 2 } = value;
```
anba: Please add `var a, b;` for strict-mode compatibility.

test/language/expressions/assignment/destructuring/obj-id-init-yield-expr.js
```diff
@@ -0,0 +1,32 @@
...
+---*/
+
+var value = {};
+var assignmentResult, iterationResult, x, iter;
+
+iter = (function*() {
+  assignmentResult = { x = yield } = value;
+}());
+
+result = iter.next();
```
anba: Please add `var result` for strict-mode compatibility.

test/language/expressions/assignment/destructuring/obj-id-put-constant-no-strict.js
```diff
@@ -0,0 +1,16 @@
...
+description: >
+    The assignment target should obey `const` semantics outside of strict mode.
+es6id: 12.14.5.4
+features: [const]
+flags: [noStrict]
+---*/
+
+const c = null;
+
+({ c } = { c: 1 });
```
anba: Assignment to immutable strict binding throws a TypeError in non-strict (and strict) mode (8.1.1.1.5 SetMutableBinding, step 6). (`const` is always created as immutable strict binding, cf. 15.1.8 Runtime Semantics: GlobalDeclarationInstantiation, step 15.b.i.1).

test/language/expressions/assignment/destructuring/obj-prop-nested-array-null.js
```diff
@@ -0,0 +1,15 @@
...
+description: >
+    When DestructuringAssignmentTarget is an array literal and the value is
+    `null`, a TypeError should be thrown.
+es6id: 12.14.5.4
+---*/
+
+var x;
+
+assert.throws(TypeError, function() {
+  ({ x: [ x ] }) = { x: null };
```
anba: Assignment expression needs to be parenthesised instead of only left-hand side expression.

test/language/expressions/assignment/destructuring/obj-prop-nested-array-undefined-own.js
```diff
@@ -0,0 +1,15 @@
...
+description: >
+    When DestructuringAssignmentTarget is an array literal and the value is
+    `undefined`, a TypeError should be thrown.
+es6id: 12.14.5.4
+---*/
+
+var x;
+
+assert.throws(TypeError, function() {
+  ({ x: [ x ] }) = { x: undefined };
```
anba: Assignment expression needs to be parenthesised instead of only left-hand side expression.

test/language/expressions/assignment/destructuring/obj-prop-nested-array-undefined.js
```diff
@@ -0,0 +1,15 @@
...
+description: >
+    When DestructuringAssignmentTarget is an array literal and the value is not
+    defined, a TypeError should be thrown.
+es6id: 12.14.5.4
+---*/
+
+var x;
+
+assert.throws(TypeError, function() {
+  ({ x: [ x ] }) = {};
```
anba: Assignment expression needs to be parenthesised instead of only left-hand side expression.

test/language/expressions/assignment/destructuring/obj-prop-nested-obj-null.js
```diff
@@ -0,0 +1,15 @@
...
+description: >
+    When DestructuringAssignmentTarget is an object literal and the value is
+    `null`, a TypeError should be thrown.
+es6id: 12.14.5.4
+---*/
+
+var x;
+
+assert.throws(TypeError, function() {
+  ({ x: { x } }) = { x: null };
```
anba: Assignment expression needs to be parenthesised instead of only left-hand side expression.

test/language/expressions/assignment/destructuring/obj-prop-nested-obj-undefined-own.js
```diff
@@ -0,0 +1,15 @@
...
+description: >
+    When DestructuringAssignmentTarget is an object literal and the value is
+    `undefined`, a TypeError should be thrown.
+es6id: 12.14.5.4
+---*/
+
+var x;
+
+assert.throws(TypeError, function() {
+  ({ x: { x } }) = { x: undefined };
```
anba: Assignment expression needs to be parenthesised instead of only left-hand side expression.

test/language/expressions/assignment/destructuring/obj-prop-nested-obj-undefined.js
```diff
@@ -0,0 +1,15 @@
...
+description: >
+    When DestructuringAssignmentTarget is an object literal and the value is
+    not defined, a TypeError should be thrown.
+es6id: 12.14.5.4
+---*/
+
+var x;
+
+assert.throws(TypeError, function() {
+  ({ x: { x } }) = {};
```
anba: Assignment expression needs to be parenthesised instead of only left-hand side expression.

test/language/expressions/assignment/destructuring/obj-prop-put-const-no-strict.js
```diff
@@ -0,0 +1,18 @@
...
+es6id: 12.14.5.4
+flags: [noStrict]
+features: [const]
+---*/
+
+var value = { a: 2 };
+var result;
+const c = 1;
+
+result = { a: c } = value;
```
anba: Assignment to immutable strict binding throws a TypeError in non-strict (and strict) mode (8.1.1.1.5 SetMutableBinding, step 6). (`const` is always created as immutable strict binding, cf. 15.1.8 Runtime Semantics: GlobalDeclarationInstantiation, step 15.b.i.1).

test/language/expressions/assignment/destructuring/obj-prop-put-const-strict.js
```diff
@@ -0,0 +1,16 @@
...
+description: >
+    The assignment target should obey `const` semantics in strict mode.
+es6id: 12.14.5.4
+flags: [onlyStrict]
+features: [const]
+---*/
+
+const c = 1;
+
+assert.throws(ReferenceError, function() {
```
anba: `TypeError` according to spec (8.1.1.1.5 SetMutableBinding, step 6).

test/language/expressions/assignment/destructuring/obj-prop-nested-array-yield-expr.js
jugglinmike: Aww, I *just* finished typing up my rationale (`<sub>` tags and everything!). Posting it here for posterity I guess

```js
({ x: [x = yield] } = { x: [] });
```

- AssignmentPattern<sub>[Yield]</sub>
  - ArrayAssignmentPattern<sub>[?Yield]</sub>
    - **`[`** AssignmentElementList<sub>[?Yield]</sub> **`]`**
      - AssignmentElisionElement<sub>[?Yield]</sub>
        - Elision<sub>opt</sub> AssignmentElement<sub>[?Yield]</sub>
          - DestructuringAssignmentTarget<sub>[?Yield]</sub> Initializer<sub>[In,?Yield]opt</sub>

- Initializer<sub>[In, Yield]</sub>
  - **`=`** AssignmentExpression<sub>[?In, ?Yield]</sub>

- AssignmentExpression<sub>[In, Yield]</sub>
  - [+Yield] YieldExpression<sub>[?In]</sub>

- YieldExpression<sub>[In]</sub>
  - **`yield`**

test/language/expressions/assignment/destructuring/obj-id-init-yield-expr.js
```diff
@@ -0,0 +1,32 @@
...
+
+assert.sameValue(iterationResult.value, undefined);
```
anba: Typo: `iterationResult` -> `result`

test/language/expressions/assignment/destructuring/obj-prop-nested-array-yield-expr.js
caitp: I really regret that this found its way into the spec, but what can you do :<

jugglinmike: I'm curious, could you talk more about that? To me, this all flows naturally, but I've been looking at the grammar for too long to be a unbiased critic of rationality :P

caitp: having words that are sometimes magic and sometimes not magic -> refactoring hazard that folks have generally tried to avoid. It's going to be another one of those things that bites people :(

jugglinmike: Ah, yeah. I guess this is another reason why it's great that module and class bodies are always strict mode code :)

test/language/expressions/assignment/destructuring/array-elem-init-order.js
jugglinmike: Thanks--this is done.

test/language/expressions/assignment/destructuring/array-elem-nested-array-yield-ident-valid.js
jugglinmike: I've changed that reference to `result`

test/language/expressions/assignment/destructuring/array-elem-put-constant-no-strict.js
jugglinmike: If I understand correctly, then this test could be made valid if the constants were *not* declared in the global environment. Is that right?

I ask more out of curiosity, though, because these tests shouldn't be asserting the intricacies of constant declarations. I think a better fix would be to collapse all `const-strict`/`const-no-strict` file pairs and omit the `onlyStrict`/`noStrict` flag entirely.

test/language/expressions/assignment/destructuring/array-iteration.js
jugglinmike: Thanks--this is done.

test/language/expressions/assignment/destructuring/array-rest-iteration.js
jugglinmike: Thanks--this is done.

test/language/expressions/assignment/destructuring/array-elem-put-constant-no-strict.js
anba: > If I understand correctly, then this test could be made valid if the constants were not declared in the global environment. Is that right?

`const` declarations are always created as immutable strict bindings, so even if the test is wrapped in a function or executed in eval- or module-code, an assignment to the binding throws a `TypeError`. Or do you mean something else?

Spec references:
9.2.12 FunctionDeclarationInstantiation(func, argumentsList), step 35.b.i.1
15.1.8 Runtime Semantics: GlobalDeclarationInstantiation (script, env), step 15.b.i.1
15.2.1.16.4 ModuleDeclarationInstantiation( ) Concrete Method, step 16.a.i.1
18.2.1.2 Runtime Semantics: EvalDeclarationInstantiation( body, varEnv, lexEnv, strict), step 14.b.i.1


> I think a better fix would be to collapse all const-strict/const-no-strict file pairs and omit the onlyStrict/noStrict flag entirely.

Yes, that seems reasonable.

test/language/expressions/assignment/destructuring/array-rest-nested-obj-undefined.js
jugglinmike: You are correct. Fixing now

test/language/expressions/assignment/destructuring/obj-id-init-order.js
jugglinmike: Thanks--this is done.

test/language/expressions/assignment/destructuring/obj-id-init-yield-expr.js
jugglinmike: I've changed the references for `result` to `iterationResult`

test/language/expressions/assignment/destructuring/obj-prop-nested-array-null.js
jugglinmike: Thanks--this is done.

test/language/expressions/assignment/destructuring/obj-prop-nested-array-undefined-own.js
jugglinmike: Thanks--this is done.

test/language/expressions/assignment/destructuring/obj-prop-nested-array-undefined.js
jugglinmike: Thanks--this is done.

test/language/expressions/assignment/destructuring/obj-prop-nested-obj-null.js
jugglinmike: Thanks--this is done.

test/language/expressions/assignment/destructuring/obj-prop-nested-obj-undefined-own.js
jugglinmike: Thanks--this is done.

test/language/expressions/assignment/destructuring/obj-prop-nested-obj-undefined.js
jugglinmike: Thanks--this is done.

test/language/expressions/assignment/destructuring/obj-id-init-yield-expr.js
jugglinmike: I've changed the references for `result` to `iterationResult`

test/language/expressions/assignment/destructuring/array-elem-put-constant-no-strict.js
jugglinmike: > Or do you mean something else?

Nope--I'm was just hung up on the name GlobalDeclarationInstantiation. As you point out, every abstract operation that creates an execution environment instantiates constants in this way.

> Yes, that seems reasonable.

Great; I'll start on this now

bterlson:
These tests are really good!

Possible to add coverage for the static semantics in 12.14.5.1?

Also I think some coverage of the ArrayAssignmentPattern with ellisions would be useful... stuff like `var [a,,b] = [1,2,3]`?

bterlson:
The 2nd bullet in 12.14.5.1 is particularly interesting I think - there are a number of negative syntax tests you could invent to stress the refinement of LeftHandSideExpression to AssignmentPattern. Things like {a: 1} = {a: 1}, [a, 1] = [a], etc.

jugglinmike:
@bterlson @caitp That sounds good. As you know from our discussions on IRC, this has been a little confusing for me, so I'd like to get your feedback on an informal proposal before creating the Test262-formatted files. What do you think of:

> ### 12.14.5.1 Static Semantics: Early Errors
>
> AssignmentProperty : IdentifierReference Initializer<sub>opt</sub>
> - It is a Syntax Error if IsValidSimpleAssignmentTarget of IdentifierReference is false.

test:

```js
// negative: SyntaxError
{ 1 } = {};
```

> DestructuringAssignmentTarget : LeftHandSideExpression
> - It is a Syntax Error if LeftHandSideExpression is either an ObjectLiteral or an ArrayLiteral and if the lexical token sequence matched by LeftHandSideExpression cannot be parsed with no tokens left over using AssignmentPattern as the goal symbol.

tests:

```js
// negative: SyntaxError
[ { get x() } ] = [{}];
```

```js
// negative: SyntaxError
[ [ 1 ] ] = [[]];
```

> - It is a Syntax Error if LeftHandSideExpression is neither an ObjectLiteral nor an ArrayLiteral and IsValidSimpleAssignmentTarget(LeftHandSideExpression) is false.

test:

```js
// negative: SyntaxError
[ 1 ] = [];
```

anba:
> AssignmentProperty : IdentifierReference Initializeropt
> 
> It is a Syntax Error if IsValidSimpleAssignmentTarget of IdentifierReference is false.

This condition can only be tested in strict mode code, tests:
```javascript
"use strict";

({eval} = {});
({eval = 0} = {});

({arguments} = {});
({arguments = 0} = {});
```

> DestructuringAssignmentTarget : LeftHandSideExpression
> 
> It is a Syntax Error if LeftHandSideExpression is either an ObjectLiteral or an ArrayLiteral and 
> if the lexical token sequence matched by LeftHandSideExpression cannot be parsed with no 
> tokens left over using AssignmentPattern as the goal symbol.

The first test example is ok; the second example is basically the same one as your last example (=> `[ 1 ] = []`). Alternatives:
```javascript
[ [(a, b)] ] = [];
[ [...a, ...b] ] = [];
[ [a += 0] ] = [];

[ ...[(a, b)] ] = [];
[ ...[...a, ...b] ] = [];
[ ...[a += 0] ] = [];
```

bterlson:
Also good to have noStrict tests for the first bullet, ie. that [eval] = [1] doesn't give an error. Otherwise with @anba's correction to the first bullet I think these are good tests. (Also don't forget previously suggested ArrayAssignmentPattern ellisions and negative syntax cases around syntactically valid LeftHandSide expressions that are not valid assignment patterns)

jugglinmike:
@bterlson I've pushed up the new tests.

> negative syntax cases around syntactically valid LeftHandSide expressions that are not valid assignment patterns

I believe this is covered by the tests for the third bullet:

> It is a Syntax Error if LeftHandSideExpression is neither an ObjectLiteral nor an ArrayLiteral and IsValidSimpleAssignmentTarget(LeftHandSideExpression) is false.

jugglinmike:
This patch already contains a test for honoring `const` semantics during assignment:

```js
const c = null;

  [ c ] = [1];
});
```

...but it's missing an equivalent for `let`. According to [8.1.1.1.5 SetMutableBinding (N,V,S)](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-declarative-environment-records-setmutablebinding-n-v-s):

> 4. If the binding for N in envRec has not yet been initialized throw a ReferenceError exception.

 @bterlson what do you think of:

```js
assert.throws(ReferenceError, function() {
  [x] = [];
  let x;
});
```

```js
assert.throws(ReferenceError, function() {
  ({ x } = {});
  let x;
});
```

```js
assert.throws(ReferenceError, function() {
  ({ x: x } = {});
  let x;
});
```

...although I don't know if covering [8.1.1.1.6 GetBindingValue(N,S)](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-declarative-environment-records-getbindingvalue-n-s) for Initializers is "good measure" or "too far", as in:

```js
var x;
assert.throws(ReferenceError, function() {
  [x = y] = [];
  let y;
});
```

```js
var x;
assert.throws(ReferenceError, function() {
  ({ x = y } = {});
  let y;
});
```

```js
var x;
assert.throws(ReferenceError, function() {
  ({ x: x = y } = {});
  let y;
});
```

bterlson:
>> negative syntax cases around syntactically valid LeftHandSide expressions that are not valid assignment patterns

> I believe this is covered by the tests for the third bullet:

>> It is a Syntax Error if LeftHandSideExpression is neither an ObjectLiteral nor an ArrayLiteral and IsValidSimpleAssignmentTarget(LeftHandSideExpression) is false.

I mean cases where LHSExpr is a ObjectLiteral or ArrayLiteral but cannot be refined to an AssignmentPattern per 2nd BIC (bullet-in-clause :stuck_out_tongue_closed_eyes:). So things like `[1] = [1]` and `{a: 1} = {a: 1}`.

bterlson:
I like all the tests you propose above. I don't think testing the GetBindingValue for initializers is out of scope, this feels like good coverage to me. I think you are good to proceed on your current course :)

jugglinmike:
> I mean cases where LHSExpr is a ObjectLiteral or ArrayLiteral but cannot be refined to an AssignmentPattern per 2nd BIC (bullet-in-clause :stuck_out_tongue_closed_eyes:). So things like [1] = [1] and {a: 1} = {a: 1}

If I'm reading the spec right (a big "if" to be sure), then I think the
situation you described in words is different than the code examples you
shared.

In `[1] = [1]`, the first `[` token is consumed as part of
ArrayAssignmentPattern, so `1` is interpreted as an AssignmentElementList, then
further as an AssignmentElisionElement, then as an AssignmentElement, then a
DestructuringAssignmentTarget, then as a LeftHandSideExpression. Here, the
*third* BIC takes over and dictates that a SyntaxError be thrown. This is
covered by the test I proposed earlier and have since included:

```js
// negative: SyntaxError
[ 1 ] = [];
```

In `{a: 1} = {a: 1}`, the first `{` token is consumed as part of
ObjectAssignmentPattern, so `a: 1` is interpreted as an AssignmentPropertyList,
then further as an AssignmentProperty. Then, `1` is interpreted as an
AssignmentElement, then a DestructuringAssignmentTarget, then finally as a
LeftHandSideExpression. So here, too, the third BIC takes over.

DestructuringAssignmentTarget is only itself a object literal or array literal
in the case of "nested" destructuring. This would include `[[1]] = [[1]]` and
`{a: {a: 1}} = {a: {a: 1}}`. The parser does not attempt to refine the nested
literals to an AssignmentPattern. Instead, the runtime semantics have explicit
instructions for parsing the nested pattern. For example

> **[12.14.5.3](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-runtime-semantics-iteratordestructuringassignmentevaluation) Runtime Semantics: IteratorDestructuringAssignmentEvaluation**
>
> [...]
>
> AssignmentElement<sub>[Yield]</sub> : DestructuringAssignmentTarget Initializer<sub>opt</sub>
>
> [...]
>
> 6 If DestructuringAssignmentTarget is an ObjectLiteral or an ArrayLiteral,
>    then
>    a. Let nestedAssignmentPattern be the parse of the source text
>       corresponding to DestructuringAssignmentTarget using either
>       AssignmentPattern or AssignmentPattern[Yield] as the goal symbol
>       depending upon whether this AssignmentElement has the Yield parameter.
>    b. Return the result of performing DestructuringAssignmentEvaluation of
>       nestedAssignmentPattern with v as the argument.

Now this is testing the limits of my comfort with the spec, but I believe these
examples fail because of a combination of the second and third BIC--the nested
literals cannot be parsed as AssignmentPattern due to BIC2, and the BIC3
"promotes" this to an early error (it would still be an error without BIC3, but
it wouldn't occur until evaluation time, when the above-quoted runtime
semantics were applied). I believe the tests that @anba an I discussed are
equivalent: `({ x: [(x, y)] } = { x: [] });` and `({ x: { get x() {} } } = { x:
{} });`.

It's really tricky for me to keep this all straight, though, so I may have made
a mistake somewhere in my reasoning. Does it sound right to you?

jugglinmike:
> I like all the tests you propose above. I don't think testing the GetBindingValue for initializers is out of scope, this feels like good coverage to me. I think you are good to proceed on your current course :)

Great! I'll follow up with these tomorrow.

jugglinmike:
I just pushed up those tests for `let` behavior for SetMutableBinding and GetBindingValue. I missed one case in my previous comment, namely SetMutableBinding for an un-initialized binding in an AssignmentResetElement:

```js
assert.throws(ReferenceError, function() {
  [...x] = [];
  let x;
});
```

but it's included in the latest commit.

test/language/expressions/assignment/destructuring/obj-id-init-yield-expr.js
```diff
@@ -0,0 +1,32 @@
...
+assert.sameValue(iterationResult.done, false);
+assert.sameValue(x, undefined);
+assert.sameValue(assignmentResult, undefined);
+
+iterationResult = iter.next(3);
+
+assert.sameValue(iterationResult.value, undefined);
+assert.sameValue(iterationResult.done, true);
+assert.sameValue(x, undefined);
```
anba: s/undefined/3/

jugglinmike: Oof. Thanks.

bterlson:
Looks good, just need a squash.

jugglinmike:
@bterlson All squashed

bterlson:
Thanks!

PR #270, opened 2015-05-15
S15.4.4.5_A3.1_T1.js: Fix inconsistent error message

edgemaster:
Test for x.join(NaN) reports an error using the content of x.join("NaN"). This is confusing.


bterlson:
Good find, thanks!

PR #272, opened 2015-05-18
Enable strict mode for more built-in tests

anba:
- Remove no longer needed noStrict flags.
- Change tests to use propertyHelper.js for strict mode compatibility.
- Add tests for return value of `delete` operator, e.g. `delete Array.prototype === false`.
- Add non-writable tests for global NaN property.
- Split some tests to run in strict mode.
- Change tests with global `var length` declaration to use a different variable name for browser environments.
- Merge duplicated tests symbol-data-property-configurable-non-strict and symbol-data-property-configurable-strict.
- Merge duplicated tests symbol-data-property-writable-non-strict and symbol-data-property-writable-strict.
- (And a small change in propertyHelper to reduce code duplication in function call.)


bterlson:
LGTM!

PR #273, opened 2015-05-18
Assignment with left-hand side property accessor

anba:
The assignment operator evaluates its operands from left to right. When
the left-hand side expression is a property accessor, RequireObjectCoercible
and ToPropertyKey are called on the property accessor before the right-hand
side expression is evaluated.

### DIFF
--- a/test/language/expressions/assignment/S11.13.1_A7_T1.js
+++ b/test/language/expressions/assignment/S11.13.1_A7_T1.js
@@ -1,0 +1,38 @@
+// Copyright (C) 2015 André Bargull. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+info: Assignment Operator evaluates its operands from left to right.
+description: >
+    The left-hand side expression is evaluated before the right-hand side.
+    Left-hand side expression is MemberExpression: base[prop]. base is the
+    null value.
+---*/
+
+function DummyError() { }
+
+assert.throws(DummyError, function() {
+  var base = null;
+  var prop = function() {
+    throw new DummyError();
+  };
+  var expr = function() {
+    $ERROR("right-hand side expression evaluated");
+  };
+
+  base[prop()] = expr();
+});
+
+assert.throws(TypeError, function() {
+  var base = null;
+  var prop = {
+    toString: function() {
+      $ERROR("property key evaluated");
+    }
+  };
+  var expr = function() {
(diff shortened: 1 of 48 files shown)


bterlson:
I guess this is continuing in your series of other assignment reference tests that fail in all implementations? ;)

Success? I should think so! I wish we had a bounty program for such tests :)

anba:
> I guess this is continuing in your series of other assignment reference tests that fail in all implementations?

Hmm kind of, I was doing some performance experiments in my test runtime when I noticed the spec violations in other runtimes.

> Success? I should think so! I wish we had a bounty program for such tests :)

I wonder about the impact of these tests on actual implementations, for example whether or not it's worth it to fix the spec violations and if it will have negative effect on the general performance?

bterlson:
> I wonder about the impact of these tests on actual implementations, for example whether or not it's worth it to fix the spec violations and if it will have negative effect on the general performance? 

Indeed, and also whether the change is actually web compatible since right now we have interoperable behavior. Test262's role, as I see it, is to test what the spec says. If we have good reasons the spec should change, then we should tackle that, and at such time implementers agree on actual spec language, we can modify the tests accordingly.

domenic:
It would probably be worth setting aside some section of the readme or something that notes tests that fail in an interoperable way?

bterlson:
@Domenic I think that's a good idea. Could also add a front-matter attribute or flag for such tests, as this would allow harnesses to filter them out. Something like knownDeviation? Probably a better name than this...

anba:
This part of the specification hasn't really changed that much when compared to ES1. So it's a rather long-standing deviation between spec and implementations, already visible in the old Mozilla sources from 1998: http://lxr.mozilla.org/classic/source/js/src/jsemit.c#1659

`base[prop] += expr` is compiled to:
```
emit base
emit prop
dup2
getelem
emit expr
setelem
```
Byte code interpretation for *getelem* starts at http://lxr.mozilla.org/classic/source/js/src/jsinterp.c#2168, and next to it is *setelem*. In case you want to read multiple levels of C macros... :wink:

anba:
I've just filed https://bugs.ecmascript.org/show_bug.cgi?id=4379 to have a place to discuss possible specification changes.

bterlson:
I'm goign to pull these in. When/if the spec changes here (see also [ES bug 4379](https://bugs.ecmascript.org/show_bug.cgi?id=4379)) we can update these tests. This also applies to #275.

PR #274, opened 2015-05-18
MOP operations on non-configurable arguments-object property

anba:
Mapped arguments property is changed to non-configurable. Check arguments mapping works correctly when applying various MOP methods.

Questions:
- There is a new `arguments-object/mapped` directory, should I move the test file into that directory?
- Should I try to split the test into multiple files?


bterlson:
Yeah definitely split these into multiple files (and move the comments for each scenario into the description).

Might as well use arguments/mapped location as well. While I'm not too sold on keeping that around as a separate directory, would be better for now to keep everything together.

anba:
The test case in `argumentsNonConfigurableThenNonWritableWithInterveningSetMutableBinding` should probably be considered a spec bug. Implementers may want to propose the following change for 9.4.4.2 [[DefineOwnProperty]].

<ol>
<li> Let <var>args</var> be the arguments object.
<li> Let <var>map</var> be the value of the [[ParameterMap]] internal slot of the arguments object.
<li> Let <var>isMapped</var> be HasOwnProperty(<var>map</var>, <var>P</var>).
<li> <ins>Let <var>newArgDesc</var> be <var>Desc</var>.</ins>
<li> <ins>If <var>isMapped</var> is <strong>true</strong> and IsDataDescriptor(<var>Desc</var>) is <strong>true</strong>, then
  <ol>
  <li> If <var>Desc</var>.[[Value]] is not present and <var>Desc</var>.[[Writable]] is present and its value is <strong>false</strong>, then
    <ol>
    <li> NOTE: Update the mapped property value before removing the mapping in step 9.
    <li> Let <var>newArgDesc</var> be a copy of <var>Desc</var>.
    <li> Let <var>current</var> be Get(<var>map</var>, <var>P</var>).
    <li> Set <var>newArgDesc</var>.[[Value]] to <var>current</var>.
    </ol>
  </ol></ins>
<li> <del>Let <var>allowed</var> be OrdinaryDefineOwnProperty(<var>args</var>, <var>P</var>, <var>Desc</var>).</del>
<li value=6> <ins>Let <var>allowed</var> be OrdinaryDefineOwnProperty(<var>args</var>, <var>P</var>, <var>newArgDesc</var>).</ins>
<li> ReturnIfAbrupt(<var>allowed</var>).
<li> If <var>allowed</var> is <strong>false</strong>, return <strong>false</strong>.
<li> If the value of <var>isMapped</var> is <strong>true</strong>, then
  <ol>
  <li> If IsAccessorDescriptor(<var>Desc</var>) is <strong>true</strong>, then
    <ol>
    <li> Call <var>map</var>.[[Delete]](<var>P</var>).
    </ol>
  <li> Else
    <ol>
    <li> If <var>Desc</var>.[[Value]] is present, then
      <ol>
      <li> Let <var>setStatus</var> be Set(<var>map</var>, <var>P</var>, <var>Desc</var>.[[Value]], <strong>false</strong>).
      <li> Assert: <var>setStatus</var> is <strong>true</strong> because formal parameters mapped by argument objects are always writable.
      </ol>
    <li> If <var>Desc</var>.[[Writable]] is present and its value is <strong>false</strong>, then
      <ol>
      <li> Call <var>map</var>.[[Delete]](<var>P</var>).
      </ol>
    </ol>
  </ol>
<li> Return <strong>true</strong>.
</ol>


Simplified test case:
```javascript
(function(a) {
  a = 2;
  Object.defineProperty(arguments, "0", {writable: false});
  assert.sameValue(arguments[0], 1);
})(1);
```

anba:
Tests are now split into multiple files.

bterlson:
Wow, good find. I wonder what @allenwb thinks of your proposed spec update. I am ensuring it matches with what we actually implement but it seems reasonable at first glance.

bterlson:
@anba you shouldn't need to explictly include assert.js - it's part of the default environment.

bterlson:
Opened [a spec bug](https://bugs.ecmascript.org/show_bug.cgi?id=4371).

allenwb:
this is a spec. bug that goes that is traceable  back to ES5.
@anba's proposed spec. revision looks pretty good

anba:
> @anba you shouldn't need to explictly include assert.js - it's part of the default environment.

I'll remove the includes.

> this is a spec. bug that goes that is traceable back to ES5.

Now that the spec bug has been verified, I wonder what to do with the test case itself. I don't like to commit it as-is, because it'd be confusing if test262 tests for known spec bugs. Three possible alternatives:
 1. Remove the test case.
 1. Add a new flag to mark the test case as ~~wonky~~ asserting a known spec bug.
 1. Move the test case to a new `test/bugs` directory, which is not run by default, but merely serves as a place to hold tests for known spec bugs.

domenic:
IMO we should include a test for the *correct* behavior. But, probably not until we have a spec with it fixed to refer to...

Time to boot up the tc39/ecma262 errata.md file? (Wasn't sure if @allenwb was planning on fixing this in ES2015 but it's understandable if not.)

allenwb:
@domenic  Given the age of this bug in the spec. I don't seen any need to rush to fix it at this time in ES2015.

[The list of known ES2005 spec. bugs](https://bugs.ecmascript.org/buglist.cgi?component=Bugs%20from%20ES6&list_id=9694&product=Draft%20for%207th%20Edition&resolution=---)

bterlson:
I think we can just test for the intended semantics now. Make sure the description refers to the spec bug. @anba you want to update the test?

anba:
> @anba you want to update the test?

 Updated!

bterlson:
Thanks!

PR #275, opened 2015-05-19
Increment/Decrement with property accessor expression

anba:
The increment/decrement operator evaluates its operand expression once. When
the operand expression is a property accessor, RequireObjectCoercible
and ToPropertyKey are called on the property accessor in the correct order.


anba:
Basically the same number of tests as in #273, this time for the increment and decrement operators. The S11.x.x_A6_T3.js tests fail in Edge, JSC, V8 and Nashorn. SpiderMonkey passes all tests.

PR #277, opened 2015-05-19
Add tests for IdentifierReference as property

jugglinmike:
There are three unique error cases for this production:

1. Identifier resolution
2. Value retrieval
3. Property creation

Testing (1) was difficult. Neither [ResolveBinding](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-resolvebinding) nor [GetIdentifierReference](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-getidentifierreference) throw an error, but the latter delegates to [HasBinding](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-declarative-environment-records-hasbinding-n). HasBinding never returns an abrupt completion for Declarative Environment Records, so I had to resort to using an Object Environment Record via a `with` statement. In addition to disqualifying strict mode, this approach requires a Symbol to trigger the error. I recognize that these semantics represent a heavy requirement for this test, though, and I would welcome a simpler alternative. 

Testing (2) was straightforward--[any unresolvable reference would generate an error from GetValue](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-getvalue). 

I don't think it's possible to test (3). [CreateDataPropertyOrThrow](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-createdatapropertyorthrow) will only return an abrupt completion if [CreateDataProperty (O, P, V)](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-createdataproperty) returns an abrupt completion or `false`. I don't think either of these situations can occur because O in all cases will be a plain object with zero or more properties (all of which must be writable, enumerable, and configurable--I can't think of a way to initialize an object with any other kind of property). I'm likely missing something, but if my reasoning is sound, then I'll follow up with a bug against ECMAScript to simplify the spec.

### DIFF
--- a/test/language/expressions/object/prop-def-id-eval-error-2.js
+++ b/test/language/expressions/object/prop-def-id-eval-error-2.js
@@ -1,0 +1,22 @@
+// Copyright (C) Copyright 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 12.2.5.9
+description: >
+    Errors thrown during IdentifierReference evaluation are forwarded to the
+    runtime.
+flags: [noStrict]
+features: [Proxy]
+---*/
+
+var p = new Proxy({}, {
+  has: function () {
+    throw new Test262Error();
+  }
+});
+
+assert.throws(Test262Error, function() {
+  with (p) {
+    ({attr});
+  }
+});
--- a/test/language/expressions/object/prop-def-id-eval-error.js
+++ b/test/language/expressions/object/prop-def-id-eval-error.js
@@ -1,0 +1,21 @@
+// Copyright (C) Copyright 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 12.2.5.9
+description: >
+    Errors thrown during IdentifierReference evaluation are forwarded to the
+    runtime.
+flags: [noStrict]
+features: [Symbol, Symbol.unscopables]
+---*/
+
+var obj = {
+  attr: null,
+  get [Symbol.unscopables]() { throw new Test262Error(); }
+};
+
+assert.throws(Test262Error, function() {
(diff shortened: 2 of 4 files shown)


anba:
> I recognize that these semantics represent a heavy requirement for this test, though, and I would welcome a simpler alternative.

There are two other alternatives, but they require Proxy support. And proxies are definitely not simpler than using normal objects and Symbol.unscopables:

```javascript
var p = new Proxy({}, {
  has: function (t, pk) {
    throw "success";
  }
});

// 1. Object environment and proxy.
try {
  with (p) {
    ({attr});
  }
} catch (e) { print(e) }

// 2. Object environment of the global object and proxy.
// Assumption: |Object.prototype| is on the prototype chain of the global object.
Object.setPrototypeOf(Object.prototype, p)
try {
  try {
    ({attr});
  } catch (e) { print(e) }
} finally {
  Object.setPrototypeOf(Object.prototype, null)
}
```

bterlson:
I think the tests included are fine, although I would add a feature tag of "Symbol.unscopables" as well.

I think the first proxy test case is good to add too, actually. The 2nd I would avoid simply because global object semantics are wonky across hosts and we've had some issues with them in the past...

jugglinmike:
> although I would add a feature tag of "Symbol.unscopables" as well.

Just to be clear: `features: [Symbol, Symbol.unscopables]` -- right?

bterlson:
Yeah, I think that's best. I guess Symbol.unscopables implies symbol support so you could get away with just Symbol.unscopables I suppose... but I'm ok with both for now.

jugglinmike:
Got it

jugglinmike:
Alright, I've updated the frontmatter for the existing test and added the first test suggested by André.

bterlson:
LGTM!

jugglinmike:
Great. I've squashed the commits, so this should be good to merge

PR #279, opened 2015-05-22
fix case sensitive filenames

leobalter:
`Math.log2` is a method, so it should start with a lowercase.


bterlson:
Good call. Thanks!

PR #280, opened 2015-05-24
S15.4.4.4_A4.1 length enumerable should be checked on Array.prototype.concat

pvginkel:
The test checks Array.concat which is undefined; instead it should check Array.prototype.concat.


bterlson:
Good find, thanks! Fix is trivial so CLA is not required.

PR #282, opened 2015-05-25
Add tests for Proxy

leobalter:
this PR aims to bring all Proxy tests, including traps handlers and Proxy.revocable.

As it will require a few more days to get finished, I'm opening it now to allow earlier review on the already done parts and avoid concurrent work on the same tests.

As soon as I have this PR good for final reviews, the [WIP] prefix will be removed and all the commits will be squashed.

### DIFF
--- a/test/built-ins/Proxy/apply/call-result.js
+++ b/test/built-ins/Proxy/apply/call-result.js
@@ -1,0 +1,22 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.13
+description: >
+    Return the result from the trap method.
+info: >
+    [[Call]] (thisArgument, argumentsList)
+
+    9. Return Call(trap, handler, «target, thisArgument, argArray»).
+---*/
+
+var target = function(a, b) { return a + b; };
+var result = {};
+var handler = {
+    apply: function(t, c, args) {
+        return result;
+    }
+};
+var p = new Proxy(target, handler);
+
+assert.sameValue(p.call(), result);
--- a/test/built-ins/Proxy/create-handler-is-revoked-proxy.js
+++ b/test/built-ins/Proxy/create-handler-is-revoked-proxy.js
@@ -1,0 +1,20 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.15
+description: >
+    Proxy ( target, handler )
+    ...
+    4.  If handler is a Proxy exotic object and the value of the
+    [[ProxyHandler]] internal slot of handler is null, throw a
+    TypeError exception.
+    ...
+---*/
+
+var revocable = Proxy.revocable({}, {});
+
+revocable.revoke();
+
+assert.throws(TypeError, function() {
+    new Proxy({}, revocable.proxy);
+});
--- a/test/built-ins/Proxy/create-target-is-revoked-proxy.js
+++ b/test/built-ins/Proxy/create-target-is-revoked-proxy.js
@@ -1,0 +1,20 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.15
+description: >
+    Proxy ( target, handler )
+    ...
+    2. If target is a Proxy exotic object and the value of the
+    [[ProxyHandler]] internal slot of target is null, throw a
+    TypeError exception.
+    ...
+---*/
+
+var revocable = Proxy.revocable({}, {});
+
+revocable.revoke();
+
+assert.throws(TypeError, function() {
+    new Proxy(revocable.proxy, {});
+});
--- a/test/built-ins/Proxy/defineProperty/call-parameters.js
+++ b/test/built-ins/Proxy/defineProperty/call-parameters.js
@@ -1,0 +1,52 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.6
+description: >
+    Trap is called with handler as context and parameters are target, P, and the
+    descriptor object.
+info: >
+    [[DefineOwnProperty]] (P, Desc)
+
+    ...
+    9. Let descObj be FromPropertyDescriptor(Desc).
+    10. Let booleanTrapResult be ToBoolean(Call(trap, handler, «target, P,
+    descObj»)).
+    ...
+---*/
+
+var _handler, _target, _prop, _desc;
+var target = {};
+var descriptor = {
+    configurable: true,
+    enumerable: true,
+    writable: true,
+    value: 1
+};
+var handler = {
+    defineProperty: function(t, prop, desc) {
+        _handler = this;
+        _target = t;
+        _prop = prop;
+        _desc = desc;
+
+        return true;
+    }
+};
+var p = new Proxy(target, handler);
+
+Object.defineProperty(p, "attr", descriptor);
+
+assert.sameValue(_handler, handler);
+assert.sameValue(_target, target);
+assert.sameValue(_prop, "attr");
+
+assert.sameValue(
+    Object.keys(_desc).length, 4,
+    "descriptor arg has the same amount of keys as given descriptor"
+);
+
+assert(_desc.configurable);
+assert(_desc.writable);
+assert(_desc.enumerable);
+assert.sameValue(_desc.value, 1);
--- a/test/built-ins/Proxy/defineProperty/return-boolean-and-define-target.js
+++ b/test/built-ins/Proxy/defineProperty/return-boolean-and-define-target.js
@@ -1,0 +1,46 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.6
+description: >
+    If a property has a corresponding target object property then applying the
+    Property Descriptor of the property to the target object using
+    [[DefineOwnProperty]] will not throw an exception.
+features: [Reflect]
+includes: [propertyHelper.js]
+---*/
+
+var target = {};
+var p = new Proxy(target, {
+    defineProperty: function(t, prop, desc) {
+        return Object.defineProperty(t, prop, desc);
+    }
+});
+
+var result = Reflect.defineProperty(p, "attr", {
+    configurable: true,
+    enumerable: true,
+    writable: true,
+    value: 1
+});
+
+assert.sameValue(result, true, "result === true");
+
+verifyEqualTo(target, "attr", 1);
+verifyWritable(target, "attr");
+verifyEnumerable(target, "attr");
+verifyConfigurable(target, "attr");
+
+result = Reflect.defineProperty(p, "attr", {
+    configurable: false,
+    enumerable: false,
+    writable: false,
+    value: 2
+});
+
+assert.sameValue(result, true, "result === true");
+
+verifyEqualTo(target, "attr", 2);
+verifyNotWritable(target, "attr");
+verifyNotEnumerable(target, "attr");
+verifyNotConfigurable(target, "attr");
--- a/test/built-ins/Proxy/defineProperty/trap-is-undefined.js
+++ b/test/built-ins/Proxy/defineProperty/trap-is-undefined.js
@@ -1,0 +1,42 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.6
+description: >
+    Return target.[[DefineOwnProperty]](P, Desc) if trap is undefined.
+info: >
+    [[DefineOwnProperty]] (P, Desc)
+
+    ...
+    8. If trap is undefined, then
+        a. Return target.[[DefineOwnProperty]](P, Desc).
+    ...
+includes: [propertyHelper.js]
+---*/
+
+var target = {};
+var p = new Proxy(target, {});
+
+Object.defineProperty(p, "attr", {
+    configurable: true,
+    enumerable: true,
+    writable: true,
+    value: 1
+});
+
+verifyEqualTo(target, "attr", 1);
+verifyWritable(target, "attr");
+verifyEnumerable(target, "attr");
+verifyConfigurable(target, "attr");
+
+Object.defineProperty(p, "attr", {
+    configurable: false,
+    enumerable: false,
+    writable: false,
+    value: 2
+});
+
+verifyEqualTo(target, "attr", 2);
+verifyNotWritable(target, "attr");
+verifyNotEnumerable(target, "attr");
+verifyNotConfigurable(target, "attr");
--- a/test/built-ins/Proxy/defineProperty/trap-return-is-false.js
+++ b/test/built-ins/Proxy/defineProperty/trap-return-is-false.js
@@ -1,0 +1,27 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.6
+description: >
+    Trap returns a boolean. Checking on false values.
+info: >
+    [[DefineOwnProperty]] (P, Desc)
+
+    ...
+    12. If booleanTrapResult is false, return false.
+    ...
+features: [Reflect]
+---*/
+
+var target = {};
+var p = new Proxy(target, {
+    defineProperty: function(t, prop, desc) {
+        return 0;
+    }
+});
+
+assert.sameValue(Reflect.defineProperty(p, "attr", {}), false);
+assert.sameValue(
+    Object.getOwnPropertyDescriptor(target, "attr"),
+    undefined
+);
--- a/test/built-ins/Proxy/deleteProperty/boolean-trap-result-boolean-false.js
+++ b/test/built-ins/Proxy/deleteProperty/boolean-trap-result-boolean-false.js
@@ -1,0 +1,32 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.10
+description: >
+    [[Delete]] (P)
+
+    The result is a Boolean value.
+features: [Reflect]
+---*/
+
+var target = {};
+var p = new Proxy(target, {
+    deleteProperty: function() {
+        return 0;
+    }
+});
+
+Object.defineProperties(target, {
+    isConfigurable: {
+        value: 1,
+        configurable: true
+    },
+    notConfigurable: {
+        value: 1,
+        configurable: false
+    }
+});
+
+assert.sameValue(Reflect.deleteProperty(p, "attr"), false);
+assert.sameValue(Reflect.deleteProperty(p, "isConfigurable"), false);
+assert.sameValue(Reflect.deleteProperty(p, "notConfigurable"), false);
--- a/test/built-ins/Proxy/deleteProperty/boolean-trap-result-boolean-true.js
+++ b/test/built-ins/Proxy/deleteProperty/boolean-trap-result-boolean-true.js
@@ -1,0 +1,17 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.10
+description: >
+    [[Delete]] (P)
+
+    The result is a Boolean value.
+---*/
+
+var p = new Proxy({}, {
+    deleteProperty: function() {
+        return 1;
+    }
+});
+
+assert.sameValue(Reflect.deleteProperty(p, "attr"), true);
--- a/test/built-ins/Proxy/deleteProperty/call-parameters.js
+++ b/test/built-ins/Proxy/deleteProperty/call-parameters.js
@@ -1,0 +1,33 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.10
+description: >
+    [[Delete]] (P)
+
+    9. Let booleanTrapResult be ToBoolean(Call(trap, handler, «target, P»)).
+info: >
+    6.1.7.2 Object Internal Methods and Internal Slots
+
+    (...) Receiver is used as the this value when evaluating the code
+---*/
+
+var _handler, _target, _prop;
+var target = {
+    attr: 1
+};
+var handler = {
+    deleteProperty: function(t, prop) {
+        _handler = this;
+        _target = t;
+        _prop = prop;
+        return delete t[prop];
+    }
+};
+var p = new Proxy(target, handler);
+
+delete p.attr;
+
+assert.sameValue(_handler, handler, "handler object as the trap context");
+assert.sameValue(_target, target, "first argument is the target object");
+assert.sameValue(_prop, "attr", "second argument is the property name");
--- a/test/built-ins/Proxy/deleteProperty/targetdesc-is-not-configurable.js
+++ b/test/built-ins/Proxy/deleteProperty/targetdesc-is-not-configurable.js
@@ -1,0 +1,28 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.10
+description: >
+    [[Delete]] (P)
+
+    A property cannot be reported as deleted, if it exists as a non-configurable
+    own property of the target object.
+info: >
+    14. If targetDesc.[[Configurable]] is false, throw a TypeError exception.
+---*/
+
+var target = {};
+var p = new Proxy(target, {
+    deleteProperty: function() {
+        return true;
+    }
+});
+
+Object.defineProperty(target, "attr", {
+    configurable: false,
+    value: 1
+});
+
+assert.throws(TypeError, function() {
+    delete p.attr;
+});
--- a/test/built-ins/Proxy/enumerate/call-parameters.js
+++ b/test/built-ins/Proxy/enumerate/call-parameters.js
@@ -1,0 +1,30 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.11
+description: >
+    Trap called as trap.call(handler, target)
+info: >
+    [[Enumerate]] ()
+
+    8. Let trapResult be Call(trap, handler, «target»).
+---*/
+
+var x, _target, _handler;
+var target = {
+    attr: 1
+};
+var handler = {
+    enumerate: function(t) {
+        _target = t;
+        _handler = this;
+    }
+};
+var p = new Proxy(target, handler);
+
+try {
+    for (x in p) {}
+} catch(e) {}
+
+assert.sameValue(_handler, handler);
+assert.sameValue(_target, target);
--- a/test/built-ins/Proxy/enumerate/null-handler.js
+++ b/test/built-ins/Proxy/enumerate/null-handler.js
@@ -1,0 +1,20 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.11
+description: >
+    [[Enumerate]] ()
+
+    2. If handler is null, throw a TypeError exception.
+---*/
+
+var x;
+var p = Proxy.revocable({}, {});
+
+p.revoke();
+
+assert.throws(TypeError, function() {
+    for (x in p.proxy) {
+        x;
+    }
+});
--- a/test/built-ins/Proxy/enumerate/return-trap-result.js
+++ b/test/built-ins/Proxy/enumerate/return-trap-result.js
@@ -1,0 +1,36 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.11
+description: >
+    Trap returns a iterable result.
+info: >
+    [[Enumerate]] ()
+
+    11. Return trapResult
+includes: [compareArray.js]
+---*/
+
+var x;
+var iter = [
+    {done: false, value: 1},
+    {done: false, value: 2},
+    {done: false, value: 3},
+    {done: true, value: 42}
+];
+var target = {
+    attr: 1
+};
+var foo = { bar: 1 };
+var p = new Proxy(target, {
+    enumerate: function() {
+        return { next: function() { return iter.shift(); } };
+    }
+});
+
+var results = [];
+for (x in p) {
+    results.push(x);
+}
+
+assert(compareArray(results, [1,2,3]));
--- a/test/built-ins/Proxy/getOwnPropertyDescriptor/result-is-undefined-targetdesc-is-not-configurable.js
+++ b/test/built-ins/Proxy/getOwnPropertyDescriptor/result-is-undefined-targetdesc-is-not-configurable.js
@@ -1,0 +1,33 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.5
+description: >
+    Throws a TypeError exception if trap result is undefined and target property
+    descriptor is not configurable
+info: >
+    [[GetOwnProperty]] (P)
+
+    ...
+    14. If trapResultObj is undefined, then
+        ...
+        b. If targetDesc.[[Configurable]] is false, throw a TypeError exception.
+    ...
+---*/
+
+var target = {};
+Object.defineProperty(target, "foo", {
+    configurable: false,
+    enumerable: false,
+    value: 1
+});
+
+var p = new Proxy(target, {
+    getOwnPropertyDescriptor: function(t, prop) {
+        return;
+    }
+});
+
+assert.throws(TypeError, function() {
+    Object.getOwnPropertyDescriptor(p, "foo");
+});
--- a/test/built-ins/Proxy/getOwnPropertyDescriptor/result-is-undefined-targetdesc-is-undefined.js
+++ b/test/built-ins/Proxy/getOwnPropertyDescriptor/result-is-undefined-targetdesc-is-undefined.js
@@ -1,0 +1,31 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.5
+description: >
+    Throws a TypeError exception if trap result is undefined and target property
+    descriptor is undefined.
+info: >
+    [[GetOwnProperty]] (P)
+
+    ...
+    14. If trapResultObj is undefined, then
+        a. If targetDesc is undefined, return undefined.
+    ...
+---*/
+
+var t = {};
+var trapped;
+var p = new Proxy(t, {
+    getOwnPropertyDescriptor: function(target, prop) {
+        trapped = true;
+        return;
+    }
+});
+
+assert.sameValue(
+    Object.getOwnPropertyDescriptor(p, "attr"),
+    undefined
+);
+
+assert(trapped);
--- a/test/built-ins/Proxy/getOwnPropertyDescriptor/result-is-undefined.js
+++ b/test/built-ins/Proxy/getOwnPropertyDescriptor/result-is-undefined.js
@@ -1,0 +1,28 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.5
+description: >
+    Return undefined if trap result is undefined and target is extensible and
+    the target property descriptor is configurable.
+info: >
+    [[GetOwnProperty]] (P)
+
+    ...
+    14. If trapResultObj is undefined, then
+        ...
+        f. Return undefined.
+    ...
+---*/
+
+var target = {
+    attr: 1
+};
+
+var p = new Proxy(target, {
+    getOwnPropertyDescriptor: function(t, prop) {
+        return;
+    }
+});
+
+assert.sameValue(Object.getOwnPropertyDescriptor(p, "attr"), undefined);
--- a/test/built-ins/Proxy/getOwnPropertyDescriptor/resultdesc-return-configurable.js
+++ b/test/built-ins/Proxy/getOwnPropertyDescriptor/resultdesc-return-configurable.js
@@ -1,0 +1,30 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.5
+description: >
+    Return descriptor from trap result if it has the same value as the target
+    property descriptor.
+---*/
+
+var target = {};
+var descriptor = {
+    configurable: true,
+    enumerable: true,
+    value: 1
+};
+
+Object.defineProperty(target, "bar", descriptor);
+
+var p = new Proxy(target, {
+    getOwnPropertyDescriptor: function(t, prop) {
+        return Object.getOwnPropertyDescriptor(t, prop);
+    }
+});
+
+var proxyDesc = Object.getOwnPropertyDescriptor(p, "bar");
+
+assert(proxyDesc.configurable);
+assert(proxyDesc.enumerable);
+assert.sameValue(proxyDesc.value, 1);
+assert.sameValue(proxyDesc.writable, false);
--- a/test/built-ins/Proxy/getOwnPropertyDescriptor/resultdesc-return-not-configurable.js
+++ b/test/built-ins/Proxy/getOwnPropertyDescriptor/resultdesc-return-not-configurable.js
@@ -1,0 +1,29 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.5
+description: >
+    Return descriptor from trap result if it has the same value as the target
+    property descriptor and they're not configurable.
+---*/
+
+var target = {};
+
+Object.defineProperty(target, "attr", {
+    configurable: false,
+    enumerable: true,
+    value: 1
+});
+
+var p = new Proxy(target, {
+    getOwnPropertyDescriptor: function(t, prop) {
+        return Object.getOwnPropertyDescriptor(t, prop);
+    }
+});
+
+var proxyDesc = Object.getOwnPropertyDescriptor(p, "attr");
+
+assert.sameValue(proxyDesc.configurable, false);
+assert(proxyDesc.enumerable);
+assert.sameValue(proxyDesc.value, 1);
+assert.sameValue(proxyDesc.writable, false);
--- a/test/built-ins/Proxy/getOwnPropertyDescriptor/trap-is-not-callable.js
+++ b/test/built-ins/Proxy/getOwnPropertyDescriptor/trap-is-not-callable.js
@@ -1,0 +1,30 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.5
+description: >
+    Throws a TypeError exception if trap is not callable.
+info: >
+    [[GetOwnProperty]] (P)
+
+    ...
+    2. Let handler be the value of the [[ProxyHandler]] internal slot of O.
+    ...
+    5. Let target be the value of the [[ProxyTarget]] internal slot of O.
+    6. Let trap be GetMethod(handler, "getOwnPropertyDescriptor").
+    ...
+        7.3.9 GetMethod (O, P)
+        ...
+        2. Let func be GetV(O, P).
+        5. If IsCallable(func) is false, throw a TypeError exception.
+        ...
+---*/
+
+var target = {};
+var p = new Proxy(target, {
+    getOwnPropertyDescriptor: {}
+});
+
+assert.throws(TypeError, function() {
+    Object.getOwnPropertyDescriptor(p, "foo");
+});
--- a/test/built-ins/Proxy/getOwnPropertyDescriptor/trap-is-undefined.js
+++ b/test/built-ins/Proxy/getOwnPropertyDescriptor/trap-is-undefined.js
@@ -1,0 +1,25 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.5
+description: >
+    Return target.[[GetOwnProperty]](P) if trap is undefined.
+info: >
+    [[GetOwnProperty]] (P)
+
+    ...
+    8. If trap is undefined, then
+        a. Return target.[[GetOwnProperty]](P).
+    ...
+includes: [propertyHelper.js]
+---*/
+
+var target = {attr: 1};
+var p = new Proxy(target, {});
+
+var proxyDesc = Object.getOwnPropertyDescriptor(p, "attr");
+
+verifyEqualTo(p, "attr", 1);
+verifyWritable(p, "attr");
+verifyEnumerable(p, "attr");
+verifyConfigurable(p, "attr");
--- a/test/built-ins/Proxy/has/call-with.js
+++ b/test/built-ins/Proxy/has/call-with.js
@@ -1,0 +1,35 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.7
+description: >
+    A `with` variable check trigger trap.call(handler, target, P);
+info: >
+    [[HasProperty]] (P)
+
+    ...
+    9. Let booleanTrapResult be ToBoolean(Call(trap, handler, «target, P»)).
+    ...
+flags: [noStrict]
+---*/
+
+var _handler, _target, _prop;
+var target = {};
+var handler = {
+    has: function(t, prop) {
+        _handler = this;
+        _target = t;
+        _prop = prop;
+
+        return true;
+    }
+};
+var p = new Proxy(target, handler);
+
+with (p) {
+    (attr);
+}
+
+assert.sameValue(_handler, handler);
+assert.sameValue(_target, target);
+assert.sameValue(_prop, "attr");
--- a/test/built-ins/Proxy/has/return-true-target-prop-exists-using-with.js
+++ b/test/built-ins/Proxy/has/return-true-target-prop-exists-using-with.js
@@ -1,0 +1,25 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.7
+description: >
+    The result of [[HasProperty]] is a Boolean value and will affect has
+    checkings. True returned when target property exists;
+flags: [noStrict]
+---*/
+
+var target = {
+    attr: 1
+};
+var p = new Proxy(target, {
+    has: function(t, prop) {
+        if (prop !== "assert") {
+            return 42;
+        }
+    }
+});
+
+var attr = 0;
+with (p) {
+    assert.sameValue(attr, 1);
+}
--- a/test/built-ins/Proxy/has/return-true-target-prop-exists.js
+++ b/test/built-ins/Proxy/has/return-true-target-prop-exists.js
@@ -1,0 +1,19 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.7
+description: >
+    The result of [[HasProperty]] is a Boolean value and will affect has
+    checkings. True returned when target property exists;
+---*/
+
+var target = {
+    attr: 1
+};
+var p = new Proxy(target, {
+    has: function(t, prop) {
+        return 1;
+    }
+});
+
+assert.sameValue(("attr" in p), true);
--- a/test/built-ins/Proxy/has/return-true-without-same-target-prop.js
+++ b/test/built-ins/Proxy/has/return-true-without-same-target-prop.js
@@ -1,0 +1,16 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.7
+description: >
+    The result of [[HasProperty]] is a Boolean value and will affect has
+    checkings. True returned when target property doesn't exists;
+---*/
+
+var p = new Proxy({}, {
+    has: function(t, prop) {
+        return true;
+    }
+});
+
+assert.sameValue(("attr" in p), true);
--- a/test/built-ins/Proxy/isExtensible/return-is-boolean.js
+++ b/test/built-ins/Proxy/isExtensible/return-is-boolean.js
@@ -1,0 +1,24 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.3
+description: >
+    The trap returns a boolean result.
+---*/
+
+var target = {};
+var p = new Proxy(target, {
+    isExtensible: function(t) {
+        if ( Object.isExtensible(t) ) {
+            return 1;
+        } else {
+            return 0;
+        }
+    }
+});
+
+assert.sameValue(Object.isExtensible(p), true);
+
+Object.preventExtensions(target);
+
+assert.sameValue(Object.isExtensible(p), false);
--- a/test/built-ins/Proxy/isExtensible/return-is-different-from-target.js
+++ b/test/built-ins/Proxy/isExtensible/return-is-different-from-target.js
@@ -1,0 +1,25 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.3
+description: >
+    Throws a TypeError exception if boolean trap result is not the same as
+    target.[[IsExtensible]]() result
+info: >
+    [[IsExtensible]] ( )
+
+    ...
+    12. If SameValue(booleanTrapResult, targetResult) is false, throw a
+    TypeError exception.
+    ...
+---*/
+
+var p = new Proxy({}, {
+    isExtensible: function(t) {
+        return false;
+    }
+});
+
+assert.throws(TypeError, function() {
+    Object.isExtensible(p);
+});
--- a/test/built-ins/Proxy/isExtensible/trap-is-undefined.js
+++ b/test/built-ins/Proxy/isExtensible/trap-is-undefined.js
@@ -1,0 +1,23 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.3
+description: >
+    Return target.[[IsExtensible]]() if trap is undefined.
+info: >
+    [[IsExtensible]] ( )
+
+    ...
+    7. If trap is undefined, then
+        a. Return target.[[IsExtensible]]().
+    ...
+---*/
+
+var target = {};
+var p = new Proxy(target, {});
+
+assert.sameValue(Object.isExtensible(p), true);
+
+Object.preventExtensions(target);
+
+assert.sameValue(Object.isExtensible(p), false);
--- a/test/built-ins/Proxy/ownKeys/call-parameters-object-getownpropertynames.js
+++ b/test/built-ins/Proxy/ownKeys/call-parameters-object-getownpropertynames.js
@@ -1,0 +1,32 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.12
+description: >
+    [[OwnPropertyKeys]] ( )
+
+    8. Let trapResultArray be Call(trap, handler, «target»).
+---*/
+
+var _target, _handler;
+var target = {
+    foo: 1,
+    bar: 2
+};
+
+var handler = {
+    ownKeys: function(t) {
+        _handler = this;
+        _target = t;
+        return Object.getOwnPropertyNames(t);
+    }
+}
+var p = new Proxy(target, handler);
+
+var names = Object.getOwnPropertyNames(p);
+
+assert.sameValue(names[0], "foo");
+assert.sameValue(names[1], "bar");
+assert.sameValue(names.length, 2);
+assert.sameValue(_handler, handler);
+assert.sameValue(_target, target);
--- a/test/built-ins/Proxy/ownKeys/call-parameters-object-getownpropertysymbols.js
+++ b/test/built-ins/Proxy/ownKeys/call-parameters-object-getownpropertysymbols.js
@@ -1,0 +1,35 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.12
+description: >
+    [[OwnPropertyKeys]] ( )
+
+    8. Let trapResultArray be Call(trap, handler, «target»).
+features: [Symbol]
+---*/
+
+var _target, _handler;
+var target = {};
+var a = Symbol('a');
+var b = Symbol('b');
+
+target[a] = 1;
+target[b] = 2;
+
+var handler = {
+    ownKeys: function(t) {
+        _handler = this;
+        _target = t;
+        return Object.getOwnPropertySymbols(t);
+    }
+}
+var p = new Proxy(target, handler);
+
+var symbols = Object.getOwnPropertySymbols(p);
+
+assert.sameValue(symbols[0], a);
+assert.sameValue(symbols[1], b);
+assert.sameValue(symbols.length, 2);
+assert.sameValue(_handler, handler);
+assert.sameValue(_target, target);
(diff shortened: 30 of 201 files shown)


test/built-ins/Proxy/create-handler-is-revoked-proxy.js
```diff
@@ -0,0 +1,20 @@
...
+    TypeError exception.
+    ...
+---*/
+
+var revocable = Proxy.revocable({}, {});
+
+revocable.revoke();
+
+assert.throws(TypeError, function() {
+    new Proxy({}, revocable);
```
anba: s/revocable/revocable.proxy/

test/built-ins/Proxy/create-target-is-revoked-proxy.js
```diff
@@ -0,0 +1,20 @@
...
+    TypeError exception.
+    ...
+---*/
+
+var revocable = Proxy.revocable({}, {});
+
+revocable.revoke();
+
+assert.throws(TypeError, function() {
+    new Proxy(revocable, {});
```
anba: s/revocable/revocable.proxy/

test/built-ins/Proxy/proxy-newtarget.js
```diff
@@ -0,0 +1,28 @@
...
+        ...
+        10. Return P.
+
+---*/
+
+var p1 = new Proxy({}, {});
+
+assert.sameValue(
+    typeof p1,
+    'Object'
```
anba: Missing comma.

test/built-ins/Proxy/ownKeys/extensible-return-trap-result.js
```diff
@@ -0,0 +1,28 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.12
+description: >
+    If target is extensible, return the trap result.
+info:
```
bterlson: missing > (yaml syntax error)

test/built-ins/Proxy/ownKeys/not-extensible-missing-keys-throws.js
```diff
@@ -0,0 +1,32 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.12
+description: >
+    If target is not extensible, the result must contain all the keys of the own
+    properties of the target object.
+info:
```
bterlson: missing >

test/built-ins/Proxy/ownKeys/not-extensible-return-keys.js
```diff
@@ -0,0 +1,28 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.12
+description: >
+    If target is not extensible, the result must contain all the keys of the own
+    properties of the target object and no other values
+info:
```
bterlson: missing >

test/built-ins/Proxy/ownKeys/not-extensible-new-keys-throws.js
```diff
@@ -0,0 +1,29 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.12
+description: >
+    If target is not extensible, the result can't contain keys names not
+    contained in the target object.
+info:
```
bterlson: missing >

test/built-ins/Proxy/ownKeys/return-all-non-configurable-keys.js
```diff
@@ -0,0 +1,35 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.12
+description: >
+    The result List must contain the keys of all non-configurable own properties
+    of the target object.
+info:
```
bterlson: missing >

test/built-ins/Proxy/ownKeys/return-not-list-object-throws.js
```diff
@@ -0,0 +1,27 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.12
+description: >
+    If return is not a list object, throw a TypeError exception
+info:
```
bterlson: missing >

test/built-ins/Proxy/setPrototypeOf/boolean-trap-result-extensible-target.js
```diff
@@ -0,0 +1,30 @@
...
+
+var target = {};
+var p = new Proxy(target, {
+    setPrototypeOf: function(t, v) {
+        return v.attr;
+    }
+});
+
+var result = Object.setPrototypeOf(p, {
+    attr: 0;
```
bterlson: remove semicolon here (syntax error)

test/built-ins/Proxy/setPrototypeOf/boolean-trap-result-extensible-target.js
```diff
@@ -0,0 +1,30 @@
...
+});
+assert.sameValue(result, false);
+
+result = Object.setPrototypeOf(p, {
+    attr: 1;
```
bterlson: Same

test/built-ins/Proxy/isExtensible/return-is-boolean.js
```diff
@@ -0,0 +1,24 @@
...
+description: >
+    The trap returns a boolean result.
+---*/
+
+var target = {};
+var p = new Proxy(target, {
+    isExtensible: function(t) {
+        if ( Object.isExtensible(t) ) {
+            return 1;
+        else {
```
bterlson: Missing closing brace.

test/built-ins/Proxy/has/return-true-target-prop-exists.js
```diff
@@ -0,0 +1,24 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.7
+description: >
+    The result of [[HasProperty]] is a Boolean value and will affect has
+    checkings. True returned when target property exists;
+---*/
+
+var target = {
+    attr: 1;
```
bterlson: Remove semicolon

test/built-ins/Proxy/has/return-true-without-same-target-prop.js
```diff
@@ -0,0 +1,20 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.7
+description: >
+    The result of [[HasProperty]] is a Boolean value and will affect has
+    checkings. True returned when target property doesn't exists;
+---*/
```
bterlson: Needs onlyStrict due to with. Same applies to other tests with `with`.

test/built-ins/Proxy/has/return-true-without-same-target-prop.js
```diff
@@ -0,0 +1,20 @@
...
+var p = new Proxy({}, {
+    has: function(t, prop) {
+        return true;
+    }
+});
+
+assert.sameValue(('attr' in p), true);
+
+with (p) {
+    assert.sameValue(attr, undefined);
```
bterlson: Seems like the 'has' trap reporting true always will mean assert itself will attempt to be looked up on the with object?

test/built-ins/Proxy/getOwnPropertyDescriptor/resultdesc-return-not-configurable.js
```diff
@@ -0,0 +1,27 @@
...
+var p = new Proxy(target, {
+    getOwnPropertyDescriptor: function(t, prop) {
+        return Object.getOwnPropertyDescriptor(t, prop);
+    }
+});
+
+assert.sameValue(
+    Object.getOwnPropertyDescriptor(p, 'bar'),
+    Object.getOwnPropertyDescriptor(target, 'bar')
+});
```
bterlson: extra brace

test/built-ins/Proxy/getOwnPropertyDescriptor/trap-is-not-callable.js
```diff
@@ -0,0 +1,30 @@
...
+        ...
+---*/
+
+var target = {};
+var p = new Proxy(target, {
+    getOwnPropertyDescriptor: {}
+});
+
+assert.throws(TypeError, function() {
+    Object.getOwnPropertyDescriptor(p, 'foo'});
```
bterlson: Extra brace after 'foo'

test/built-ins/Proxy/getOwnPropertyDescriptor/resultdesc-return-configurable.js
```diff
@@ -0,0 +1,27 @@
...
+var p = new Proxy(target, {
+    getOwnPropertyDescriptor: function(t, prop) {
+        return Object.getOwnPropertyDescriptor(t, prop);
+    }
+});
+
+assert.sameValue(
+    Object.getOwnPropertyDescriptor(p, 'bar'),
+    Object.getOwnPropertyDescriptor(target, 'bar')
+});
```
bterlson: extra brace

test/built-ins/Proxy/getOwnPropertyDescriptor/result-is-undefined-targetdesc-is-undefined.js
```diff
@@ -0,0 +1,31 @@
...
+});
+
+assert.sameValue(
+    typeof Object.getOwnPropertyDescriptor(p, 'foo'),
+    'undefined'
+);
+
+assert.sameValue(
+    Object.getOwnPropertyDescriptor(p, 'foo'),
+    Object.getOwnPropertyDescriptor(t, 'foo')
```
bterlson: what is `t` supposed to be here? Probably mean to store {} as `t`?

test/built-ins/Proxy/getOwnPropertyDescriptor/result-is-undefined-targetdesc-is-not-configurable.js
```diff
@@ -0,0 +1,33 @@
...
+
+    ...
+    14. If trapResultObj is undefined, then
+        ...
+        b. If targetDesc.[[Configurable]] is false, throw a TypeError exception.
+    ...
+---*/
+
+var target = {};
+Object.definedProperty(target, 'foo', {
```
bterlson: defineProperty

test/built-ins/Proxy/deleteProperty/trap-is-undefined-no-property.js
```diff
@@ -0,0 +1,29 @@
...
+assert.sameValue(delete p.attr, true);
+assert.sameValue(delete p.notThere, true);
+assert.sameValue(
+    Object.getOwnPropertyDescriptor(target, 'attr'),
+    undefined
+);
+
+Object.defineProperty(target, 'attr', {
+    configurable: false,
+    enumerable: true
```
bterlson: missing comma

test/built-ins/Proxy/deleteProperty/trap-is-undefined.js
```diff
@@ -0,0 +1,31 @@
...
+assert.sameValue(delete p.attr, true);
+assert.sameValue(delete p.notThere, true);
+assert.sameValue(
+    Object.getOwnPropertyDescriptor(target, 'attr'),
+    undefined
+);
+
+Object.defineProperty(target, 'attr', {
+    configurable: false,
+    enumerable: true
```
bterlson: missing comma

bterlson:
A lot of syntax errors here... are you not running these before submitting PR? I'd suggest doing so if not (even if they fail).

I will take another look when this batch of issues is resolved.

bterlson:
Looks like great coverage though!

leobalter:
@bterlson, thanks! 

Due to the size of this PR, it indeed required more reviews from myself before removing the WIP prefix. 

I'll fix the pointed issues and run the tests against it.

test/built-ins/Proxy/has/return-true-without-same-target-prop.js
leobalter: you mean `[noStrict]`, right? As with doesn't work on strict mode and by the other examples I found, I'm using this flag now.

test/built-ins/Proxy/has/return-true-without-same-target-prop.js
leobalter: good call, this assertion is impossible.

test/built-ins/Proxy/has/return-true-without-same-target-prop.js
bterlson: Err yeah, meant noStrict!

bterlson:
Let me know when you're ready for a second look!

leobalter:
@bterlson Now I believe I'm ready. I also did a lot of fixes after running the tests.

test/built-ins/Proxy/defineProperty/call-parameters.js
```diff
@@ -0,0 +1,35 @@
...
+    enumerable: true,
+    writable: true,
+    value: 1
+};
+var handler = {
+    defineProperty: function(t, prop, desc) {
+        assert.sameValue(this, handler);
+        assert.sameValue(t, target);
+        assert.sameValue(prop, 'foo');
+        assert.sameValue(desc, descriptor);
```
anba: `assert.sameValue()` uses `===`, so you cannot use it here.

test/built-ins/Proxy/defineProperty/return-boolean-and-define-target.js
```diff
@@ -0,0 +1,50 @@
...
+---*/
+
+var target = {};
+var p = new Proxy(target, {
+    defineProperty: function(t, prop, desc) {
+        return Object.defineProperty(t, prop,desc);
+    }
+});
+
+var result = Object.defineProperty(p, 'foo', {
```
anba: s/Object.defineProperty/Reflect.defineProperty/ otherwise `result` is `p`.

test/built-ins/Proxy/defineProperty/trap-return-is-false.js
```diff
@@ -0,0 +1,22 @@
...
+---*/
+
+var target = {};
+var p = new Proxy(target, {
+    defineProperty: function(t, prop, desc) {
+        return 0;
+    }
+});
+
+assert.sameValue(Object.defineProperty(p, 'foo', {}), false);
```
anba: s/Object/Relect/ or alternatively Object.defineProperty + `assert.throws`.

test/built-ins/Proxy/deleteProperty/boolean-trap-result-boolean-false.js
```diff
@@ -0,0 +1,18 @@
...
+---*/
+
+var target = {};
+var p = new Proxy({}, {
+    deleteProperty: function() {
+        return 0;
+    }
+});
+
+assert.sameValue(delete target.attr, false);
```
anba: `delete p.attr` ?

test/built-ins/Proxy/deleteProperty/boolean-trap-result-boolean-true.js
```diff
@@ -0,0 +1,18 @@
...
+---*/
+
+var target = {};
+var p = new Proxy({}, {
+    deleteProperty: function() {
+        return 1;
+    }
+});
+
+assert.sameValue(delete target.attr, true);
```
anba: `delete p.attr` ?

test/built-ins/Proxy/deleteProperty/call-parameters.js
```diff
@@ -0,0 +1,27 @@
...
+var handler = {
+    deleteProperty: function(t, prop) {
+        assert.sameValue(this, handler, "handler object as the trap context");
+        assert.sameValue(t, target, "first argument is the target object");
+        assert.sameValue(prop, "attr", "second argument is the property name");
+    }
+};
+var p = new Proxy(target, handler);
+
+delete p.attr;
```
anba: Throws in strict mode.

test/built-ins/Proxy/deleteProperty/return-false.js
```diff
@@ -0,0 +1,17 @@
...
+    11. If booleanTrapResult is false, return false.
+---*/
+
+var p = new Proxy({}, {
+    deleteProperty: function() {
+        return false;
+    }
+});
+
+assert.sameValue(delete p.attr, false);
```
anba: Throws in strict mode - Reflect.deleteProperty(...) ?

test/built-ins/Proxy/deleteProperty/targetdesc-is-not-configurable.js
```diff
@@ -0,0 +1,28 @@
...
+    }
+});
+
+Object.defineProperty(target, "attr", {
+    configurable: false,
+    value: 1
+});
+
+assert.throws(TypeError, function() {
+    delete target.attr;
```
anba: `delete p.attr` ?

test/built-ins/Proxy/deleteProperty/trap-is-undefined.js
```diff
@@ -0,0 +1,29 @@
...
+    undefined
+);
+
+Object.defineProperty(target, "attr", {
+    configurable: false,
+    enumerable: true,
+    value: 1
+});
+
+assert.sameValue(delete p.attr, false);
```
anba: Throws in strict mode - Reflect.deleteProperty(..) ?

test/built-ins/Proxy/enumerate/call-parameters.js
```diff
@@ -0,0 +1,25 @@
...
+---*/
+
+var target = {
+    attr: 1
+};
+var handler = {
+    enumerate: function(t) {
+        assert.sameValue(this, handler);
+        assert.sameValue(t, target);
+        return {};
```
anba: You need to return an iterator object here.

test/built-ins/Proxy/getOwnPropertyDescriptor/resultdesc-return-configurable.js
```diff
@@ -0,0 +1,27 @@
...
+    value: 1
+});
+
+var p = new Proxy(target, {
+    getOwnPropertyDescriptor: function(t, prop) {
+        return Object.getOwnPropertyDescriptor(t, prop);
+    }
+});
+
+assert.sameValue(
```
anba: `assert.sameValue()` uses `===`, so you cannot use it here.

test/built-ins/Proxy/getOwnPropertyDescriptor/resultdesc-return-not-configurable.js
```diff
@@ -0,0 +1,27 @@
...
+    value: 1
+});
+
+var p = new Proxy(target, {
+    getOwnPropertyDescriptor: function(t, prop) {
+        return Object.getOwnPropertyDescriptor(t, prop);
+    }
+});
+
+assert.sameValue(
```
anba: Same issue here.

test/built-ins/Proxy/getOwnPropertyDescriptor/trap-is-undefined.js
```diff
@@ -0,0 +1,22 @@
...
+    ...
+    8. If trap is undefined, then
+        a. Return target.[[GetOwnProperty]](P).
+    ...
+---*/
+
+var target = {foo: 1};
+var p = new Proxy(target, {});
+
+assert.sameValue(
```
anba: And here.

test/built-ins/Proxy/has/return-true-target-prop-exists-using-with.js
```diff
@@ -0,0 +1,23 @@
...
+};
+var p = new Proxy(target, {
+    has: function(t, prop) {
+        return 1;
+    }
+});
+
+var attr = 0;
+with (p) {
+    assert.sameValue(attr, 1);
```
anba: `assert` lookup needs to be special cased in proxy handler.

test/built-ins/Proxy/isExtensible/return-is-different-from-target.js
```diff
@@ -0,0 +1,35 @@
...
+
+---*/
+
+var p = new Proxy({}, {
+    isExtensible: function(t) {
+        return !Object.isExtensible(t);
+    }
+});
+
+assert.throws(function() {
```
anba: Missing `assert.throws` exception parameter.

test/built-ins/Proxy/isExtensible/trap-is-undefined.js
```diff
@@ -0,0 +1,23 @@
...
+    [[IsExtensible]] ( )
+
+    ...
+    7. If trap is undefined, then
+        a. Return target.[[IsExtensible]]().
+    ...
+---*/
+
+var target = {};
+var p = new Proxy({}, {});
```
anba: `new Proxy(target, {})`

test/built-ins/Proxy/ownKeys/call-parameters-object-getownpropertynames.js
```diff
@@ -0,0 +1,28 @@
...
+var handler = {
+    ownKeys: function(t) {
+        assert.sameValue(this, handler);
+        assert.sameValue(t, target);
+        return Object.getOwnPropertyNames(t);
+    }
+}
+var p = new Proxy(target, handler);
+
+assert.sameValue(
```
anba: Here.

test/built-ins/Proxy/ownKeys/call-parameters-object-getownpropertysymbols.js
```diff
@@ -0,0 +1,28 @@
...
+var handler = {
+    ownKeys: function(t) {
+        assert.sameValue(this, handler);
+        assert.sameValue(t, target);
+        return Object.getOwnPropertySymbols(t);
+    }
+}
+var p = new Proxy(target, handler);
+
+assert.sameValue(
```
anba: Here.

test/built-ins/Proxy/ownKeys/call-parameters-object-keys.js
```diff
@@ -0,0 +1,28 @@
...
+var handler = {
+    ownKeys: function(t) {
+        assert.sameValue(this, handler);
+        assert.sameValue(t, target);
+        return Object.keys(t);
+    }
+}
+var p = new Proxy(target, handler);
+
+assert.sameValue(
```
anba: And here.

test/built-ins/Proxy/ownKeys/extensible-return-trap-result.js
```diff
@@ -0,0 +1,28 @@
...
+    value: true
+});
+
+var p = new Proxy(target, {
+    ownKeys: function() {
+        return ["foo", "bar"];
+    }
+});
+
+assert.sameValue(Object.keys(p), ["foo", "bar"]);
```
anba: And here.

test/built-ins/Proxy/ownKeys/not-extensible-return-keys.js
```diff
@@ -0,0 +1,28 @@
...
+
+var p = new Proxy(target, {
+    ownKeys: function() {
+        return ["foo", "bar"];
+    }
+});
+
+Object.preventExtensions(target);
+
+assert.sameValue(Object.keys(p), Object.keys(target));
```
anba: And here.

test/built-ins/Proxy/ownKeys/trap-is-undefined-no-property.js
```diff
@@ -0,0 +1,17 @@
...
+    7. If trap is undefined, then Return target.[[OwnPropertyKeys]]()
+---*/
+
+var target = {
+    foo: 1,
+    bar: 2
+};
+var p = new Proxy(target, {});
+
+assert.sameValue(Object.keys(p), Object.keys(target));
```
anba: Here.

test/built-ins/Proxy/ownKeys/trap-is-undefined.js
```diff
@@ -0,0 +1,19 @@
...
+
+var target = {
+    foo: 1,
+    bar: 2
+};
+var p = new Proxy(target, {
+    ownKeys: undefined
+});
+
+assert.sameValue(Object.keys(p), Object.keys(target));
```
anba: Here.

test/built-ins/Proxy/preventExtensions/call-parameters.js
```diff
@@ -0,0 +1,23 @@
...
+
+    ...
+    8. Let booleanTrapResult be ToBoolean(Call(trap, handler, «target»)).
+    ...
+---*/
+
+var target = {};
+var handler = {
+    preventExtensions: function(t) {
+        assert.saveValue(this, handler);
```
anba: Typo.

test/built-ins/Proxy/preventExtensions/call-parameters.js
```diff
@@ -0,0 +1,23 @@
...
+        assert.sameValue(t, target);
+    }
+};
+var p = new Proxy(target, handler);
```
anba: Missing test?

test/built-ins/Proxy/preventExtensions/return-false.js
```diff
@@ -0,0 +1,20 @@
...
+---*/
+
+var target = {};
+var p = new Proxy({}, {
+    preventExtensions: function(t) {
+        return 0;
+    }
+});
+
+assert.sameValue(Object.preventExtensions(p), false);
```
anba: Throws a TypeError. Change to Reflect.preventExtensions or use assert.throws

test/built-ins/Proxy/preventExtensions/return-true-target-is-not-extensible.js
```diff
@@ -0,0 +1,18 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.4
+description: >
+    Return boolean trap result if its true and target is not extensible.
+---*/
+
+var target = {};
+var p = new Proxy({}, {
```
anba: new Proxy(target, ...

test/built-ins/Proxy/preventExtensions/trap-is-undefined.js
```diff
@@ -0,0 +1,12 @@
...
+/*---
+es6id: 9.5.4
+description: >
+    Return target.[[PreventExtensions]]() if target is undefined.
+---*/
+
+var target = {};
+var p = new Proxy(target, {});
+
+assert.sameValue(Object.preventExtensions(p), target);
```
anba: Here.

test/built-ins/Proxy/proxy-newtarget.js
```diff
@@ -0,0 +1,28 @@
...
+        ...
+        10. Return P.
+
+---*/
+
+var p1 = new Proxy({}, {});
+
+assert.sameValue(
+    typeof p1,
+    'Object',
```
anba: s/Object/object/

test/built-ins/Proxy/set/boolean-trap-result-is-false-boolean-return-false.js
```diff
@@ -0,0 +1,20 @@
...
+
+var target = {};
+var handler = {
+    set: function(t, prop, value, receiver) {
+        return false;
+    }
+};
+var p = new Proxy(target, handler);
+
+assert.sameValue((p.attr = 'foo'), false);
```
anba: Change assignment to `Reflect.set(...)`.

test/built-ins/Proxy/set/boolean-trap-result-is-false-null-return-false.js
```diff
@@ -0,0 +1,20 @@
...
+
+var target = {};
+var handler = {
+    set: function(t, prop, value, receiver) {
+        return null;
+    }
+};
+var p = new Proxy(target, handler);
+
+assert.sameValue((p.attr = 'foo'), false);
```
anba: Change assignment to `Reflect.set(...)`.

test/built-ins/Proxy/set/boolean-trap-result-is-false-number-return-false.js
```diff
@@ -0,0 +1,20 @@
...
+
+var target = {};
+var handler = {
+    set: function(t, prop, value, receiver) {
+        return 0;
+    }
+};
+var p = new Proxy(target, handler);
+
+assert.sameValue((p.attr = 'foo'), false);
```
anba: Change assignment to `Reflect.set(...)`.

test/built-ins/Proxy/set/boolean-trap-result-is-false-string-return-false.js
```diff
@@ -0,0 +1,20 @@
...
+
+var target = {};
+var handler = {
+    set: function(t, prop, value, receiver) {
+        return '';
+    }
+};
+var p = new Proxy(target, handler);
+
+assert.sameValue((p.attr = 'foo'), false);
```
anba: Reflect.set(...)

test/built-ins/Proxy/set/boolean-trap-result-is-false-undefined-return-false.js
```diff
@@ -0,0 +1,20 @@
...
+
+var target = {};
+var handler = {
+    set: function(t, prop, value, receiver) {
+        return undefined;
+    }
+};
+var p = new Proxy(target, handler);
+
+assert.sameValue((p.attr = 'foo'), false);
```
anba: Reflect.set(...)

test/built-ins/Proxy/set/call-parameters.js
```diff
@@ -0,0 +1,25 @@
...
+        assert.sameValue(this, handler, 'handler object as the trap context');
+        assert.sameValue(t, target, 'first argument is the target object');
+        assert.sameValue(prop, 'attr', 'second argument is the property name');
+        assert.sameValue(value, 'foo', 'third argument is the new value');
+        assert.sameValue(receiver, p, 'forth argument is the proxy object');
+    }
+};
+var p = new Proxy(target, handler);
+
+p.attr = 'foo';
```
anba: Throws a TypeError in strict mode since `set` trap doesn't return `true`.

test/built-ins/Proxy/set/return-true-target-property-accessor-is-configurable-set-is-undefined.js
```diff
@@ -0,0 +1,31 @@
...
+};
+var p = new Proxy(target, handler);
+
+Object.defineProperty(target, 'attr', {
+    configurable: true,
+    set: undefined
+});
+
+assert.throws(TypeError, function() {
+    p.attr = 'bar';
```
anba: Doesn't throw.

test/built-ins/Proxy/set/return-true-target-property-accessor-is-configurable-set-is-undefined.js
```diff
@@ -0,0 +1,31 @@
...
+});
+
+assert.throws(TypeError, function() {
+    p['attr'] = 'bar';
```
anba: Doesn't throw.

test/built-ins/Proxy/set/return-true-target-property-accessor-is-not-configurable.js
```diff
@@ -0,0 +1,33 @@
...
+
+Object.defineProperty(target, 'attr', {
+    configurable: false,
+    set: function( value ) {
+        return value;
+    }
+});
+
+assert.throws(TypeError, function() {
+    p.attr = 'bar';
```
anba: Doesn't throw a TypeError.

test/built-ins/Proxy/set/return-true-target-property-accessor-is-not-configurable.js
```diff
@@ -0,0 +1,33 @@
...
+});
+
+assert.throws(TypeError, function() {
+    p['attr'] = 'bar';
```
anba: Doesn't throw a TypeError.

test/built-ins/Proxy/set/return-true-target-property-is-not-configurable.js
```diff
@@ -0,0 +1,32 @@
...
+var p = new Proxy(target, handler);
+
+Object.defineProperty(target, 'attr', {
+    configurable: false,
+    writable: true,
+    value: 'foo'
+});
+
+assert.throws(TypeError, function() {
+    p.attr = 'bar';
```
anba: Doesn't throw a TypeError.

test/built-ins/Proxy/set/return-true-target-property-is-not-configurable.js
```diff
@@ -0,0 +1,32 @@
...
+});
+
+assert.throws(TypeError, function() {
+    p['attr'] = 'bar';
```
anba: Doesn't throw a TypeError.

test/built-ins/Proxy/set/return-true-target-property-is-not-writable.js
```diff
@@ -0,0 +1,32 @@
...
+var p = new Proxy(target, handler);
+
+Object.defineProperty(target, 'attr', {
+    configurable: true,
+    writable: false,
+    value: 'foo'
+});
+
+assert.throws(TypeError, function() {
+    p.attr = 'bar';
```
anba: This assignment doesn't throw.

test/built-ins/Proxy/set/return-true-target-property-is-not-writable.js
```diff
@@ -0,0 +1,32 @@
...
+});
+
+assert.throws(TypeError, function() {
+    p['attr'] = 'bar';
```
anba: Same issue here.

test/built-ins/Proxy/setPrototypeOf/boolean-trap-result-extensible-target.js
```diff
@@ -0,0 +1,31 @@
...
+---*/
+
+var target = {};
+var p = new Proxy(target, {
+    setPrototypeOf: function(t, v) {
+        return v.attr;
+    }
+});
+
+var result = Object.setPrototypeOf(p, {
```
anba: Throws a TypeError.

test/built-ins/Proxy/setPrototypeOf/call-parameters.js
```diff
@@ -0,0 +1,27 @@
...
+var handler = {
+    setPrototypeOf: function(t, v) {
+        assert.sameValue(this, handler);
+        assert.sameValue(t, target);
+        assert.sameValue(v, val);
+    }
+};
+var p = new Proxy(target, handler);
+
+Object.setPrototypeOf(p, val);
```
anba: Throws a TypeError - missing return value in trap.

test/built-ins/Proxy/setPrototypeOf/not-extensible-target-same-target-prototype.js
```diff
@@ -0,0 +1,19 @@
...
+var target = Object.create(Array.prototype);
+var p = new Proxy(target, {
+    setPrototypeOf: function(t, v) {
+        return true;
+    }
+});
+
+Object.preventExtensions(target);
+
+assert.sameValue(Object.setPrototypeOf(p, Array.prototype), true);
```
anba: s/Object/Reflect/

test/built-ins/Proxy/setPrototypeOf/not-extensible-trap-is-false-return-false.js
```diff
@@ -0,0 +1,18 @@
...
+var target = Object.create({foo: 1});
+var p = new Proxy(target, {
+    setPrototypeOf: function(t, v) {
+        return false;
+    }
+});
+
+Object.preventExtensions(target);
+
+assert.sameValue(Object.setPrototypeOf(p, {}), false);
```
anba: s/Object/Reflect/

test/built-ins/Proxy/defineProperty/trap-is-undefined.js
```diff
@@ -0,0 +1,49 @@
...
+    enumerable: false,
+    writable: false,
+    value: 2
+});
+
+assert.sameValue(target.propertyIsEnumerable('foo'), false,
+    'target.foo is not enumerable');
+assert.sameValue(target.foo, 2);
+
+target.foo = 0;
```
anba: Reflect.set otherwise a TypeError is thrown in strict mode.

test/built-ins/Proxy/defineProperty/trap-is-undefined.js
```diff
@@ -0,0 +1,49 @@
...
+assert.sameValue(target.foo, 2, 'target.foo is not writable');
+
+assert(delete target.foo, 'false', 'target.foo is not configurable');
```
anba: s/'false'/false/ and Reflect.deleteProperty otherwise a TypeError is thrown in strict mode.

bterlson:
Looks like @anba beat me to it :) LMK when you're ready for a 3rd look ;)

test/built-ins/Proxy/apply/call-result.js
```diff
@@ -0,0 +1,23 @@
...
+info: >
+    [[Call]] (thisArgument, argumentsList)
+
+    9. Return Call(trap, handler, «target, thisArgument, argArray»).
+---*/
+
+var target = function(a, b) { return a + b; };
+var handler = {
+    apply: function(t, c, args) {
+        return 2 * t.apply(null, args);
```
jugglinmike: This test duplicates coverage with `call-parameters.js`. You can simplify it by declaring `var result = {};` at the top of the test, returning `result` from the trap, and asserting `sameValue(p.call(), result)`

jugglinmike:
> Looks like @anba beat me to it! LMK when you're ready for a 3rd look ;)

Good call. I'll hold off, too

test/built-ins/Proxy/deleteProperty/call-parameters.js
leobalter: this is a regular property, I can't see how it would throw in strict mode. It's a configurable property, matching [12.5.4.2 ](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-delete-operator-runtime-semantics-evaluation)

anba: The trap implicitly returns `undefined`, ToBoolean(undefined) = false => 12.5.4.2, step 5.e throws a TypeError.

leobalter: thanks!

I included a `return delete t[prop];` on the trap to avoid it, but it won't appear here as an outdated diff.

leobalter:
Thank you, @anba!

Fixes are all up, I'll rebase and squash commits later.

test/built-ins/Proxy/defineProperty/call-parameters.js
```diff
@@ -0,0 +1,43 @@
...
+    configurable: true,
+    enumerable: true,
+    writable: true,
+    value: 1
+};
+var handler = {
+    defineProperty: function(t, prop, desc) {
+        assert.sameValue(this, handler);
+        assert.sameValue(t, target);
+        assert.sameValue(prop, foo);
```
anba: s/foo/"foo"/

test/built-ins/Proxy/defineProperty/return-boolean-and-define-target.js
```diff
@@ -0,0 +1,51 @@
...
+    writable: false,
+    value: 2
+});
+
+assert.sameValue(result, true, "result === true");
+assert.sameValue(target.propertyIsEnumerable("foo"), false,
+    "target.foo is not enumerable");
+assert.sameValue(target.foo, 2);
+
+target.foo = 0;
```
anba: This assignment throws in strict-mode.

test/built-ins/Proxy/defineProperty/return-boolean-and-define-target.js
```diff
@@ -0,0 +1,51 @@
...
+assert.sameValue(target.foo, 2, "target.foo is not writable");
+
+assert(delete target.foo, "false", "target.foo is not configurable");
```
anba: - s/assert/assert.sameValue/
- and `delete` throws in strict mode

test/built-ins/Proxy/defineProperty/trap-return-is-false.js
```diff
@@ -0,0 +1,23 @@
...
+---*/
+
+var target = {};
+var p = new Proxy(target, {
+    defineProperty: function(t, prop, desc) {
+        return 0;
+    }
+});
+
+assert.sameValue(Reflect.defineProperty(p, foo, {}), false);
```
anba: `foo` is undefined.

test/built-ins/Proxy/deleteProperty/boolean-trap-result-boolean-false.js
```diff
@@ -0,0 +1,18 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.10
+description: >
+    [[Delete]] (P)
+
+    The result is a Boolean value.
+---*/
+
+var target = {};
```
anba: `target` is unused.

test/built-ins/Proxy/deleteProperty/boolean-trap-result-boolean-false.js
```diff
@@ -0,0 +1,18 @@
...
+var p = new Proxy({}, {
+    deleteProperty: function() {
+        return 0;
+    }
+});
+
+assert.sameValue(delete p.attr, false);
```
anba: Throws TypeError in strict mode.

test/built-ins/Proxy/deleteProperty/boolean-trap-result-boolean-true.js
```diff
@@ -0,0 +1,18 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.10
+description: >
+    [[Delete]] (P)
+
+    The result is a Boolean value.
+---*/
+
+var target = {};
```
anba: `target` is unused.

test/built-ins/Proxy/deleteProperty/targetdesc-is-not-configurable.js
```diff
@@ -0,0 +1,28 @@
...
+    [[Delete]] (P)
+
+    A property cannot be reported as deleted, if it exists as a non-configurable
+    own property of the target object.
+info: >
+    14. If targetDesc.[[Configurable]] is false, throw a TypeError exception.
+---*/
+
+var target = {};
+var p = new Proxy({}, {
```
anba: `target` is not passed to Proxy.

test/built-ins/Proxy/enumerate/return-trap-result.js
```diff
@@ -0,0 +1,26 @@
...
+        return ["a", "b", "c"][Symbol.iterator]();
+    }
+});
+
+var results = [];
+for (let x in p) {
+    results.push(x);
+}
+
+assert.sameValue(results, ["a", "b", "c"]);
```
anba: `assert.sameValue` cannot be used here. Change to compareArray (from harness/compareArray.js).

test/built-ins/Proxy/getOwnPropertyDescriptor/resultdesc-return-configurable.js
```diff
@@ -0,0 +1,30 @@
...
+        return Object.getOwnPropertyDescriptor(t, prop);
+    }
+});
+
+var proxyDesc = Object.getOwnPropertyDescriptor(p, "bar");
+
+assert(proxyDesc.configurable);
+assert(proxyDesc.enumerable);
+assert.sameValue(proxyDesc.value, 1);
+assert(proxyDesc.writable);
```
anba: This assertion isn't correct. The descriptor does not define the "writable" property, so it defaults to `false`.

test/built-ins/Proxy/getOwnPropertyDescriptor/resultdesc-return-not-configurable.js
```diff
@@ -0,0 +1,29 @@
...
+        return Object.getOwnPropertyDescriptor(t, prop);
+    }
+});
+
+var proxyDesc = Object.getOwnPropertyDescriptor(p, "attr");
+
+assert.sameValue(proxyDesc.configurable, false);
+assert(proxyDesc.enumerable);
+assert.sameValue(proxyDesc.value, 1);
+assert(proxyDesc.writable);
```
anba: This assertion isn't correct. The descriptor does not define the "writable" property, so it defaults to `false`.

test/built-ins/Proxy/ownKeys/call-parameters-object-getownpropertynames.js
```diff
@@ -0,0 +1,29 @@
...
+        assert.sameValue(this, handler);
+        assert.sameValue(t, target);
+        return Object.getOwnPropertyNames(t);
+    }
+}
+var p = new Proxy(target, handler);
+
+var names = Object.getOwnPropertyNames(p);
+
+assert.sameValues(names[0], "foo");
```
anba: s/sameValues/sameValue/ and in next two calls.

test/built-ins/Proxy/ownKeys/call-parameters-object-getownpropertysymbols.js
```diff
@@ -0,0 +1,33 @@
...
+        assert.sameValue(t, target);
+        return Object.getOwnPropertySymbols(t);
+    }
+}
+var p = new Proxy(target, handler);
+
+var symbols = Object.getOwnPropertySymbols(p);
+
+assert.sameValue(symbols[0], a);
+assert.sameValue(symbols[0], b);
```
anba: Second index should be 1 instead of 0.

test/built-ins/Proxy/ownKeys/extensible-return-trap-result.js
```diff
@@ -0,0 +1,33 @@
...
+var p = new Proxy(target, {
+    ownKeys: function() {
+        return ["foo", "bar"];
+    }
+});
+
+var keys = Object.keys(p);
+
+assert.sameValue(keys[0], "foo");
+assert.sameValue(keys[1], "bar");
```
anba: `keys.length` is 1, so this assertion isn't correct. Change `Object.getOwnPropertyNames` to `Object.keys`? Probably also in other tests ownKeys trap?

test/built-ins/Proxy/preventExtensions/return-true-target-is-not-extensible.js
```diff
@@ -0,0 +1,18 @@
...
+var target = {};
+var p = new Proxy(target, {
+    preventExtensions: function(t) {
+        return 1;
+    }
+});
+
+Object.preventExtensions(target);
+
+assert.sameValue(Object.preventExtensions(p), true);
```
anba: `Object.preventExtensions` returns its input value, change to `Reflect.preventExtensions` ?

anba:
> Thank you, @anba!
> 
> Fixes are all up, I'll rebase and squash commits later.

When the remaining issues are fixed, @bterlson or @jugglinmike should do a proper review. All I did was simply [running the tests](https://github.com/anba/es6draft#test262-test-suite).

test/built-ins/Proxy/defineProperty/return-boolean-and-define-target.js
bterlson: Also note that "false" should be `false`.

bterlson:
At this point I'm ready to sign off once the remaining technical issues are fixed.

bterlson:
I've also run the tests in Chakra, couple minor bugs (grats!) and otherwise the only failures are due to above test bugs.

Coverage wise we could stand to include tests for traps being both undefined and null (tests only check for undefined).

jugglinmike:
> At this point I'm ready to sign off once the remaining technical issues are fixed.

I've got some feedback to share, but I'm not quite done reviewing. I'll submit it tomorrow

leobalter:
@anba:

> All I did was simply [running the tests](https://github.com/anba/es6draft#test262-test-suite).

This test runner is amazing, I'm already using it.

test/built-ins/Proxy/ownKeys/extensible-return-trap-result.js
leobalter: this test is returning the trap result: 

[9.5.12 item 22](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-proxy-object-internal-methods-and-internal-slots-ownpropertykeys) `If extensibleTarget is true, return trapResult.`

leobalter: I'll may use getOwnPropertyNames as they are "working", but seems like this trap should work for Object.keys as well.

anba: `Object.keys` also calls `[[GetOwnProperty]]` to check if the property is present and enumerable (7.3.21 EnumerableOwnNames - step 5.a.i-iii). In this case the property "bar" is not-present and therefore omitted from the result set.

anba:
> This test runner is amazing, I'm already using it.

:smile:

test/built-ins/Proxy/ownKeys/extensible-return-trap-result.js
leobalter: thanks!

leobalter:
seems like I fixed all the tests. I'll be glad to know what else I should fix.

I'll use all the feedback here to improve my next contributions, so we won't have so much minor tech issues.

test/built-ins/Proxy/defineProperty/trap-is-undefined.js
```diff
@@ -0,0 +1,43 @@
...
+
+Object.defineProperty(p, "attr", {
+    configurable: true,
+    enumerable: true,
+    writable: true,
+    value: 1
+});
+
+var desc = Object.getOwnPropertyDescriptor(target, "attr");
+assert(desc.enumerable);
```
jugglinmike: Check out `harness/propertyHelper.js`--you can use the `includes` frontmatter tag to define that for tests like this.

test/built-ins/Proxy/defineProperty/trap-return-is-false.js
```diff
@@ -0,0 +1,23 @@
...
+---*/
+
+var target = {};
+var p = new Proxy(target, {
+    defineProperty: function(t, prop, desc) {
+        return 0;
+    }
+});
+
+assert.sameValue(Reflect.defineProperty(p, "foo", {}), false);
```
jugglinmike: Beyond just the return value, we'll want to ensure that the operation is canceled in this case.

test/built-ins/Proxy/enumerate/null-handler.js
```diff
@@ -0,0 +1,19 @@
...
+---*/
+
+var p = Proxy.revocable({}, {});
+
+p.revoke();
+
+assert.throws(TypeError, function() {
+    for (let x in p.proxy) {
+        x;
+    };
```
jugglinmike: Unnecessary semicolon

test/built-ins/Proxy/get/trap-is-not-undefined-accessor-get-is-undefined-throws.js
```diff
@@ -0,0 +1,36 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
```
jugglinmike: It would be more concise/consistent to omit `trap-is-not-undefined` from the name of this file. (We can take that for granted as the "normal" case.)

test/built-ins/Proxy/get/call-parameters-2.js
```diff
@@ -0,0 +1,28 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
```
jugglinmike: For consistency with the other tests for this trap, you can merge this with `get/call-parameters.js`

test/built-ins/Proxy/get/report-same-value-for-non-writable-non-configurable-property.js
```diff
@@ -0,0 +1,37 @@
...
+        a. If IsDataDescriptor(targetDesc) and targetDesc.[[Configurable]] is
+        false and targetDesc.[[Writable]] is false, then
+            i. If SameValue(trapResult, targetDesc.[[Value]]) is false, throw a
+            TypeError exception.
+---*/
+
+var target = {};
+var p = new Proxy(target, {
+    get: function() {
+        return 2;
```
jugglinmike: The file for this test is inaccurate because the test body asserte the behavior when the result is *not* the same value. I think you should keep it (but rename it) and include another test that asserts no error when the trap does return the same value.

test/built-ins/Proxy/getOwnPropertyDescriptor/result-is-undefined-targetdesc-is-undefined.js
```diff
@@ -0,0 +1,32 @@
...
+
+var t = {};
+var p = new Proxy(t, {
+    getOwnPropertyDescriptor: function(target, prop) {
+        return;
+    }
+});
+
+assert.sameValue(
+    typeof Object.getOwnPropertyDescriptor(p, "foo"),
```
jugglinmike: No need to use the `typeof` operator, you can assert equality with `undefined` directly. The second assertion is redundant. It's also a good idea to make sure the handler was invoked (since otherwise this test will pass if the trap is completely unimplemented)

test/built-ins/Proxy/getOwnPropertyDescriptor/result-is-undefined.js
```diff
@@ -0,0 +1,28 @@
...
+    foo: 1
+};
+
+var p = new Proxy(target, {
+    getOwnPropertyDescriptor: function(t, prop) {
+        return;
+    }
+});
+
+assert.sameValue(typeof Object.getOwnPropertyDescriptor(p), "undefined");
```
jugglinmike: Need to specify a second argument to `Object.getOwnPropertyDescriptor` (asserting trap execution would help here). Also no need for the `typeof` operator.

test/built-ins/Proxy/has/call-with.js
```diff
@@ -0,0 +1,30 @@
...
+info: >
+    [[HasProperty]] (P)
+
+    ...
+    9. Let booleanTrapResult be ToBoolean(Call(trap, handler, «target, P»)).
+    ...
+flags: [noStrict]
+---*/
+
+var target = { attr: 1 };
```
jugglinmike: To protect against false positives, `target` should not define an attribute `foo`

test/built-ins/Proxy/has/call-hasownproperty.js
```diff
@@ -0,0 +1,32 @@
...
+    2. Let handler be the value of the [[ProxyHandler]] internal slot of O.
+    ...
+    5. Let target be the value of the [[ProxyTarget]] internal slot of O.
+    6. Let trap be GetMethod(handler, "has").
+    ...
+    9. Let booleanTrapResult be ToBoolean(Call(trap, handler, «target, P»)).
+    ...
+---*/
+
+var target = Object.create(Array.prototype);
```
jugglinmike: `target` can be defined via a simple object literal (no need to have ArrayPrototype in its prototype chain)

test/built-ins/Proxy/isExtensible/return-is-different-from-target.js
```diff
@@ -0,0 +1,35 @@
...
+    10. Let targetResult be target.[[IsExtensible]]().
+    11. ReturnIfAbrupt(targetResult).
+    12. If SameValue(booleanTrapResult, targetResult) is false, throw a TypeError exception.
+    ...
+
+---*/
+
+var p = new Proxy({}, {
+    isExtensible: function(t) {
+        return !Object.isExtensible(t);
```
jugglinmike: It would be more direct/readable to return the literal `false` here. It's probably also worth adding an analogous test for non-extensible objects.

test/built-ins/Proxy/ownKeys/call-parameters-object-keys.js
```diff
@@ -0,0 +1,30 @@
...
+    bar: 2
+};
+
+var handler = {
+    ownKeys: function(t) {
+        assert.sameValue(this, handler);
+        assert.sameValue(t, target);
+        return Object.keys(t);
+    }
+}
```
jugglinmike: Missing semicolon

test/built-ins/Proxy/set/trap-is-undefined-no-property.js
```diff
@@ -0,0 +1,19 @@
...
+    8. If trap is undefined, then return target.[[Set]](P, V, Receiver).
+
+---*/
+
+var target = {
+    attr: 1
+};
+var p = new Proxy(target, {});
+
+p.attr = 1;
```
jugglinmike: You should set a distinct value here to ensure that VMs actually invoke target.[[Set]]

test/built-ins/Proxy/setPrototypeOf/trap-is-undefined.js
```diff
@@ -0,0 +1,14 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.2
+description: >
+    Return target.[[SetPrototypeOf]] (V) if trap is undefined.
+---*/
+
+var target = Object.create({});
```
jugglinmike: no need for `Object.create`

test/built-ins/Proxy/setPrototypeOf/not-extensible-trap-is-false-return-false.js
```diff
@@ -0,0 +1,19 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.2
+description: >
+    Return false if boolean trap result is false.
+features: [Reflect]
+---*/
+
+var target = Object.create({foo: 1});
```
jugglinmike: You can omit `Object.create` in this test

test/built-ins/Proxy/setPrototypeOf/not-extensible-target-not-same-target-prototype.js
```diff
@@ -0,0 +1,38 @@
...
+    9. Let booleanTrapResult be ToBoolean(Call(trap, handler, «target, V»)).
+    14. Let targetProto be target.[[GetPrototypeOf]]().
+    15. ReturnIfAbrupt(targetProto).
+    16. If booleanTrapResult is true and SameValue(V, targetProto) is false,
+    throw a TypeError exception.
+    ...
+
+---*/
+
+var target = Object.create({foo: 1});
```
jugglinmike: No need for `Object.create` in this test

jugglinmike:
Hey Leo, nice work! I left some specific commits in-line, but I also had some more general feedback that didn't fit with any particular file:

- Assertions within traps. It's usually best to make assertions in the main
  script body, otherwise you risk false positives. In the case of proxies, many
  of these tests will pass if the implementer does not invoke the trap under test. You
  *could* leave your assertions as they are, introduce a new counter variable,
  increment it in the trap body, and assert the counter value at the end of
  the test. The problem there is that if implementors do not honor abrupt
  completions in trap methods, the Test262Errors generated by the assertions
  will be swallowed (see the next point). It's probably best to close around variables defined in
  the global scope and asserting them at the end of each test.
- Generally, if the spec isn't returning the result of a Proxy trap
  unconditionally, it is explicitly checking if the result is an abrupt
  completion and returning that if so.  This means it's important to verify
  that abrupt completions are forwarded to the runtime. What do you think about
  adding a new test for each trap where you throw a Test262Error from the trap?
- ProxyCreate only defines a [[Call]] internal method if the target defines
  one. Another good test would be to attempt to invoke a proxy created from a
  target with no [[Call]] internal method, where you would expect a TypeError
  as per EvaluateDirectCall (12.3.4.3).
- ProxyCreate only defines a [[Construct]] internal method if the target
  defines one. Another good test would be to attempt to construct from a proxy
  created from a target with no [[Construct]] internal method, where you would
  expect a TypeError as per EvaluateNew (12.3.3.1.1).
- `apply/trap-is-not-callable.js` is a good test! We'll want an equivalent for
  [[Construct]], maybe `construct/trap-is-not-constructor.js`.
- The `defineProperty` tests don't cover steps 19 and 20 of the algorithm. For
  instance, step 19 would need something like:
    ```js
    var p = new Proxy({}, { defineProperty: function() {return true;} });
      Object.defineProperty(p, 'foo', { configurable: false });
    });
    ```

   ..and

   ```js
    var p = new Proxy(Object.preventExtensions({}), { defineProperty: function() {return true;} });
      Object.defineProperty(p, 'foo', {});
    });
    ```
- Tests that rely on Symbol values should declare `features: [Symbol]` in their
  frontmatter
- I think it would be good to have tests that ensure when the [[Delete]] trap returns a falsey
  value, the "configurable" invariant is not enforced. What do you think?
- The `trap-is-not-callable` tests are good, but we don't have them for all the traps. I think the following are missing:
  - `construct/trap-is-not-callable.js`
  - `deleteProperty/trap-is-not-callable.js`
  - `enumerate/trap-is-not-callable.js`
  - `get/trap-is-not-callable.js`
  - `ownKeys/trap-is-not-callable.js`
  - `set/trap-is-not-callable.js`
- `enumerate`
  - you should test that when this method returns a valid iterator, that
    iterator is actually used by the `for..in` loop.
  - you can simplify the prerequisites for these tests by using `var` instead
    of `let`
- I don't think this patch includes coverage for [[OwnPropertyKeys]] step 19.a . If that's right, could you add another test for it?

leobalter:
Still working. For the records, it was good to know that Object.prototype.hasOwnProperty doesn't trigger the `has` trap, regarding the spec. I detected that after moving the assertions out of the traps and going deep on the specs for `hasOwnProperty` and `hasProperty` abstracts.

Thanks, @jugglinmike.

leobalter:
@jugglinmike 

> apply/trap-is-not-callable.js is a good test! We'll want an equivalent for [[Construct]], maybe construct/trap-is-not-constructor.js

actually, the file name should remain `construct/trap-is-not-callable.js` too as the trap is gonna be checked by [`GetMethod`](https://people.mozilla.org/%7Ejorendorff/es6-draft.html#sec-getmethod) abstract.

jugglinmike:
> actually, the file name should remain construct/trap-is-not-callable.js too as the trap is gonna be checked by GetMethod abstract.

Roger that

leobalter:
Done all changes after @jugglinmike's review.

test/built-ins/Proxy/getOwnPropertyDescriptor/trap-is-undefined.js
```diff
@@ -0,0 +1,25 @@
...
+    ...
+includes: [propertyHelper.js]
+---*/
+
+var target = {attr: 1};
+var p = new Proxy(target, {});
+
+var proxyDesc = Object.getOwnPropertyDescriptor(p, "attr");
+
+assert(proxyDesc.configurable);
```
jugglinmike: It looks like you added an `includes` entry for propertyHelper.js but forgot to use it here

leobalter: solved!

test/built-ins/Proxy/enumerate/call-parameters.js
```diff
@@ -0,0 +1,27 @@
...
+
+var x, _target, _handler;
+var target = {
+    attr: 1
+};
+var handler = {
+    enumerate: function(t) {
+        _target = t;
+        _handler = this;
+        return {next: function() { return { done:true }; } };
```
evilpie: This causes a hang in Firefox that we might not be able to fix at the moment for backwards-compatibility reasons. Can you change this to `{ done: true, value: undefined }`?

leobalter: I would insist in this test for the same reason it catches an ES6 issue on a specific browser.

While this can't be addressed would it be possible to blacklist this test there?

leobalter:
Squashed all the fixes on the main commits.

test/built-ins/Proxy/enumerate/call-parameters.js
```diff
@@ -0,0 +1,27 @@
...
+    }
+};
+var p = new Proxy(target, handler);
+
+for (x in p) {}
```
evilpie: Fair enough, I have a better idea now. Add $ERROR (or something similar) inside the loop body, because it should never be executed.

test/built-ins/Proxy/enumerate/call-parameters.js
```diff
@@ -0,0 +1,30 @@
...
+    enumerate: function(t) {
+        _target = t;
+        _handler = this;
+    }
+};
+var p = new Proxy(target, handler);
+
+try {
+    for (x in p) {}
+} catch(e) {}
```
leobalter: @evilpie I made this change, so this way I don't need to care at all about the trap return and focus on this file goal: to test exclusively the trap context and parameter.

test/built-ins/Proxy/enumerate/return-trap-result-done.js
```diff
@@ -0,0 +1,22 @@
...
+
+var x;
+var p = new Proxy([1,2,3], {
+    enumerate: function(t) {
+        return {next: function() { return { done:true }; } };
+    }
+});
+
+for (x in p) {
+    $ERROR("returned iterable interface from trap is flagged as done.");
```
leobalter: @evilpie and here we have a specific test based on your feedback to check the trap return as a done iterable. This might serve as a flag to check when this feature will be fixed on Firefox or any other envs.

test/built-ins/Proxy/enumerate/return-trap-result-done.js
```diff
@@ -0,0 +1,22 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.11
+description: >
+    Trap returns a done iterable result.
```
jugglinmike: This description could use some work. Instead of "Trap returns a done iterable result", what about: "Trap returns an iterator whose IteratorResult does not contain a `value` property". Likewise, the file name could be `return-trap-result-no-value.js`

leobalter:
@bterlson

> Let me know when you're ready for a second look!

Seems like everything is finally fixed, probably good to land.

bterlson:
Yep, LGTM!

leobalter:
Yay! Thanks, everyone! Looking forward to bring the next tests.

PR #285, opened 2015-05-26
Additional tests with incorrect for-in context

anba:
- Add missing ".prototype" lookup in Array.prototype tests (Follow-up of #280)
- Add noStrict flag in global/S10.2.3_A2.2_*.js (implicit this-binding is part of the test)
- Provide correct this-binding in function/13.2-*.js
- Add missing assert.throws in "set-iterator-close-after-add-failure.js"


bterlson:
Thanks!!

PR #286, opened 2015-05-26
Add tests for MethodDefinition forms new to ES6

jugglinmike:
@bterlson There are a few things things I'd like to call out about this set of
tests:

1. File names. I know this isn't so important to you, but I'm still not happy
   with the scheme I've chosen: `generator-*` for generator method tests and
   `name-*` for "normal" method tests. The spec is organized such that the
   "normal" form doesn't have a distinct name. I chose the prefix `name`
   because I needed to distinguish from getter and setter methods (which could
   either be grouped using `accessor-*` or named distinctly using `getter-*`
   and `setter-*`). Would `normal-*` be any better?
2. ~~`Object.prototype` extensions. The tests that concern SuperProperty
   necessarily require dereferencing values on the Object prototype. They
   verify that the object being accessed is *the* Object prototype by assigning
   a custom attribute to it. This is something you've warned against in the
   past, but I think it's required here.~~ @arv [suggested an alternate approach that avoids this issue](https://github.com/tc39/test262/pull/286#discussion_r31051828)
3. Untestable details. PropertyDefinitionEvaluation ends with a call to
   DefinePropertyOrThrow (both for "normal" methods in
   [14.3.9](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-method-definitions-runtime-semantics-propertydefinitionevaluation)
   and GeneratorMethods in
   [14.4.13](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-generator-function-definitions-runtime-semantics-propertydefinitionevaluation)).
   I don't believe this is testable for the same reason I wasn't able to test
   related spec language in [my pull request for IdentifierReferences as
   properties in object
   initializers](https://github.com/tc39/test262/pull/277). Because these tests
   concern object initializers, the object *O* cannot be in a state that would
   cause [[DefineOwnProperty]] to return either *false* or an abrupt
   completion.


anba:
> Untestable details. PropertyDefinitionEvaluation ends with a call to DefinePropertyOrThrow (both for "normal" methods in 14.3.9 and GeneratorMethods in 14.4.13). I don't believe this is testable for the same reason I wasn't able to test related spec language in my pull request for IdentifierReferences as properties in object initializers.

Yeah, this is only testable in classes (`(class {static ["prototype"](){}})` throws TypError).

test/language/expressions/object/method-definition/generator-super-prop-body.js
```diff
@@ -10,8 +10,18 @@ author: Sam Mikes
...
-    *foo(a) {
-        yield super.x;
+var value = {};
+var obj;
+
+try {
+  Object.prototype.Test262Attr = value;
+  obj = {
+    *foo() {
+      return super.Test262Attr;
```
arv: could you return super.toString and ensure that it is same value as Object.prototype.toString?

arv:
LGTM

test/language/expressions/object/method-definition/generator-super-prop-body.js
jugglinmike: Great idea--I'll patch that in now

jugglinmike: Done--I've also consistently defined `toString` on the object instance itself so that these tests don't produce false positives for implementations that define SuperProperty to de-reference the object itself.

test/language/expressions/object/method-definition/generator-param-init-yield.js
```diff
@@ -0,0 +1,28 @@
...
+      return x;
+    }
+  };
+}());
+
+iter.next();
+
+iter.next('defaultViaExpression');
+
+assert.sameValue(obj.method().next(), 'defaultViaExpression');
```
anba: The last two statements seem to be wrong. 

```diff
-iter.next('defaultViaExpression');
-
-assert.sameValue(obj.method().next(), 'defaultViaExpression');
+assert.sameValue(obj.method('defaultViaExpression').next().value, 'defaultViaExpression');
+assert.sameValue(obj.method().next().value, 'defaultViaIdentifier');
```

anba: @allenwb clarified that `yield` should not be allowed as an identifier reference in generator function default parameters (https://twitter.com/awbjs/status/603251578070552576). That means this test should be changed to `negative: SyntaxError`.

jugglinmike: There is definitely a mistake here, but I think the correction looks more like:

```diff
 iter.next('defaultViaExpression');
 
-assert.sameValue(obj.method().next(), 'defaultViaExpression');
+assert.sameValue(obj.method().next().value, 'defaultViaExpression');
```

Here's my reasoning:

The first call to `next` is necessary to start the generator and is not associated with any `yield` expression. The next call to `next` binds the value of the `yield` expression in the method's initializer and completes the definition of `obj`. At this point, `obj.method` is equivalent to `(function*(x = 'defaultViaExpression') { return x; })`, so `obj.method()` will create an iterator that returns `'defaultViaExpression'`.

This is one of the more convoluted tests I've written so far, though, so the likelyhood that I still have it wrong is high

jugglinmike: > That means this test should be changed to negative: SyntaxError.

I just saw this--disregard my previous comment.

bterlson:
Ok, finally got this reviewed. Good to go!

PR #287, opened 2015-05-26
Extend coverage for GeneratorPrototype methods

jugglinmike:
Correct errors in existing tests for the `throw` method and add new tests for the `return` method.


bterlson:
LGTM, good fixes

PR #289, opened 2015-05-28
Update contribution guidelines

jugglinmike:
@bterlson It looks like the project's CONTRIBUTING.md file has fallen out of
date. I've tried to bring it in line with current expectations, though a number
of edits are more subjective than that (mostly relating to streamlining content
for new contributors). I'm submitting as a number of fine-grained commits, but
I'd be happy to squash prior to landing.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,9 +2,7 @@
 
 ## Test Case Names
 
-There is a substantial amount of variation in existing test names and that's ok.
-
-Test cases should be created in files that are named to identify a feature, API (or aspect of either), that's being tested. Previously, there was a naming system based on the specification section and algorithm step that was the focus of the test. This protocol doesn't work if the section or algorithm step changes. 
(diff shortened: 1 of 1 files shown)


jugglinmike:
@bterlson Do you want to include any code style preferences here? I'm thinking of line length in particular. Something like,

> **Code style**
> Except when the behavior under test mandates otherwise, please adhere to the following code style conventions:
> - Limit line length to 80 characters

etc.

bterlson:
This is really good. I want to pull this in now after squash :)

I don't like 80 char line width. Are there other coding styles you might want to enforce? IMO we can cover this in a second PR. This content is too good to gate on that!

jugglinmike:
@bterlson Sure; I've opened gh-297 to discuss coding style separately. As for squashing: would you prefer to land this as the individual non-fixup commits or as a single commit (containing all the non-fixup commit messages)?

bterlson:
One commit is fine IMO.

jugglinmike:
@bterlson great--this should be all set now

PR #292, opened 2015-05-30
Add tests for variable instantiation for eval in strict mode

pvginkel:
Test case 10.4.2-3-c-2-s verifies that eval can't instantiate variables in the calling context. However, it does not verify that the variable isn't matched to the global context either. Added test case 10.4.2-3-c-3-s to check for this.


bterlson:
Test looks good. Can you sign the CLA? http://tc39.github.io/test262-cla

pvginkel:
Done.

bterlson:
Thanks!

PR #293, opened 2015-06-01
Add tests for Object-Assign

yanlxu:
Add tests for Object-Assign

### DIFF
--- a/test/built-ins/Object/assign/ObjectOverride-sameproperty.js
+++ b/test/built-ins/Object/assign/ObjectOverride-sameproperty.js
@@ -1,0 +1,13 @@
+// Copyright 2015 Microsoft Corporation. All rights reserved.
+// This code is governed by the license found in the LICENSE file.
+
+/*--
+description: Object properties are assigned to target in ascending index order,
+             i.e. a later assignment to the same property overrides an earlier assignment.
+es6id:  19.1.2.1
+--*/
+
+var target = {a: 1};
+var result = Object.assign(target,{a:2},{a:"c"});
+
+assert.sameValue(result.a, "c", "The value should be 'c'.");
--- a/test/built-ins/Object/assign/Override-notstringtarget.js
+++ b/test/built-ins/Object/assign/Override-notstringtarget.js
@@ -1,0 +1,17 @@
+// Copyright 2015 Microsoft Corporation. All rights reserved.
+// This code is governed by the license found in the LICENSE file.
+
+/*---
+description: Test override of Object.Assign(target,...sources),
+             Every string from sources will be wrapped to objects, and override from the first letter(result[0]) all the time
+es6id:  19.1.2.1
+---*/
+
+var target = 12;
+var result = Object.assign(target,"aaa","bb2b","1c");
+
+assert.sameValue(Object.keys(result).length, 4 , "The length should be 4 in the final object.");
+assert.sameValue(result[0], "1", "The value should be {\"0\":\"1\"}.");
+assert.sameValue(result[1], "c", "The value should be {\"1\":\"c\"}.");
+assert.sameValue(result[2], "2", "The value should be {\"2\":\"2\"}.");
+assert.sameValue(result[3], "b", "The value should be {\"3\":\"b\"}.");
--- a/test/built-ins/Object/assign/Target-Boolean.js
+++ b/test/built-ins/Object/assign/Target-Boolean.js
@@ -1,0 +1,12 @@
+// Copyright 2015 Microsoft Corporation. All rights reserved.
+// This code is governed by the license found in the LICENSE file.
+
+/*---
+description: Test the first argument(target) of Object.Assign(target,...sources),
+             if target is Boolean,the return value should be a new object whose value is target.
+es6id:  19.1.2.1.1
+---*/
+
+var result = Object.assign(true,{a:1});
+assert.sameValue(typeof result, "object", "Return value should be an object.");
+assert.sameValue(result.valueOf(), true, "Return value should be true.");
--- a/test/built-ins/Object/assign/Target-Null.js
+++ b/test/built-ins/Object/assign/Target-Null.js
@@ -1,0 +1,10 @@
+// Copyright 2015 Microsoft Corporation. All rights reserved.
+// This code is governed by the license found in the LICENSE file.
+
+/*---
+description: Test the first argument(target) of Object.Assign(target,...sources),
+             if target is null,Should Throw a TypeError exception.
+es6id:  19.1.2.1.1
+---*/
+
+assert.throws(TypeError, function(){Object.assign(null, {a:1});});
--- a/test/built-ins/Object/assign/OnlyOneArgument.js
+++ b/test/built-ins/Object/assign/OnlyOneArgument.js
@@ -1,0 +1,13 @@
+// Copyright 2015 Microsoft Corporation. All rights reserved.
+// This code is governed by the license found in the LICENSE file.
+
+/*---
+description: test Object.Assign(target,...sources),only one argument was passed,
+             return ToObject(target)
+es6id:  19.1.2.1.3
+---*/
+
+var target = "a";
+var result = Object.assign(target);
+
+assert.sameValue(result.valueOf(), "a", "The value should be 'a'.");
--- a/test/built-ins/Object/assign/Override.js
+++ b/test/built-ins/Object/assign/Override.js
@@ -1,0 +1,27 @@
+// Copyright 2015 Microsoft Corporation. All rights reserved.
+// This code is governed by the license found in the LICENSE file.
+
+/*---
+description: Test Object.Assign(target,...sources).
+es6id:  19.1.2.1.5.c
+---*/
+
+//"a" will be an property of the final object and the value should be 1
+var target = {a:1};
+/*---
+"1a2c3" have own enumerable properties, so it Should be wrapped to objects;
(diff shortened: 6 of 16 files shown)


yanlxu:
Add tests for Object-Assign

test/built-ins/Object/Assign/Target-Undefined-Null.js
```diff
@@ -0,0 +1,21 @@
...
+    {
+        Object.assign(target,{a:1});
+        $ERROR("Should throw an exception but no error was throw");
+    }
+    catch(e)
+    {
+        assert(e.name === "TypeError" , "Expected to throw a TypeError but got " + e.name + " instead");
+    }
+}
+test(undefined);
```
jugglinmike: You can use `assert.throws` here instead

test/built-ins/Object/Assign/Override-withStringtarget.js
```diff
@@ -0,0 +1,13 @@
+/*---
+description: Test override of Object.Assign(target,...sources),
+             Every string from sources will be wrapped to objects, and override from the first letter(result[0]) all the time.
+es6id:  19.1.2.1
+author: Wei Guo
+---*/
+
+var target = "ddd";
+var result = Object.assign(target,"aaa","bb2b","1c");
```
jugglinmike: This test is invalid. String literals will be converted to String instances, and [String instances have non-writable index properties](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-stringgetindexproperty). This, combined with the fact that `Object.assign` invokes the Set internal operation with the *throw* parameter set to **true**, means that we should expect a TypeError here. (As I mentioned below, you can assert this using `assert.throws`.)

test/built-ins/Object/Assign/LengthOfAssign.js
```diff
@@ -0,0 +1,7 @@
+/*---
+description: The length property of the assign method should be 2
+es6id:  19.1.2.1
+author: Wei Guo
+---*/
+
+assert(Object.assign.length === 2 , "The length property of the assign method should be 2");
```
jugglinmike: Can you also test the property descriptor? You can use the `propertyHelper.js` file--see [the test for `Set.length`](https://github.com/tc39/test262/blob/527644d21199d5e9fcab0b958c3426cbd0d0b9d6/test/built-ins/Set/length.js) for an example

test/built-ins/Object/Assign/Override-notstringtarget.js
```diff
@@ -0,0 +1,13 @@
...
+             Every string from sources will be wrapped to objects, and override from the first letter(result[0]) all the time
+es6id:  19.1.2.1
+author: Wei Guo
+---*/
+
+var target = 12;
+var result = Object.assign(target,"aaa","bb2b","1c");
+
+assert(Object.keys(result).length === 4 , "The length should be 4 in the final object,but actual length: " + Object.keys(result).length +" .");
+assert((result[0] === "1" && result[1] === "c" && result[2]==="2" && result[3] === "b") , "Every string form sources will be wrapped to objects, and override from the first letter(result[0]), so the value should be{\"0\":\"1\",\"1\":\"c\",\"2\":\"2\",\"3\":\"b\"}, but we got:{\"0\":\""+ result[0]+
```
jugglinmike: Please use a separate assertion for each condition in this expression

jugglinmike:
Thanks for the patch! I left specific comments in-line, but I also have some more general recommendations:

  - Please include a copyright block in each test file
  - Please use `assert.sameValue` to test equality (this will allow you to use
    a shorter assertion message)
  - Name the new directory to match the method--`assign` instead of `Assign`
  - Remove trailing whitespace. (Although avoiding unnecessary white space is generally
    a code style concern, in this case, it actually interferes with the Python runner's YAML parser.)

test/built-ins/Object/Assign/Target-String-Number-Boolean.js
```diff
@@ -0,0 +1,17 @@
...
+function test(target)
+{
+    var result = Object.assign(target,{a:1});
+    assert((typeof result === "object" && result.valueOf() === target) , "Return value should be an object whose value is: " + target + ", but a "+ typeof result 
+        + " whose value is: " +result.valueOf() + ", instead." ); 
+}
+
+test("test");
+test(1);
+test(true);
```
jugglinmike: Can you make a dedicated test file for each of these value types?

test/built-ins/Object/Assign/Target-Undefined-Null.js
suwc: /*---
...
negative: TypeError
---*/
Object.assign(target,{a:1});

jugglinmike: Actually, we're trying to move away from using the `negative` frontmatter for runtime errors like this. This preference is soon-to-be-documented in the project's CONTRIBUTING.md file; see gh-289.

yanlxu:
have made changes to address PR review comments.

test/built-ins/Object/assign/ObjectOverride-sameproperty.js
```diff
@@ -0,0 +1,15 @@
...
+description: Object properties are assigned to target in ascending index order,
+			 i.e. a later assignment to the same property overrides an earlier assignment.
+es6id:  19.1.2.1
+includes: [assert.js]
+--*/
+
+var target = {a: 1};
+var result = Object.assign(target,{a:2},{a:"c"});
+
+assert.sameValue(result.a, "c", "Object.assign() will override the property with the last value of this property in the arguments, so the value should be 'c'. But we got a \'" +result.a
```
jugglinmike: [The `assert.sameValue` method prints actual and expected values in case of failure](https://github.com/tc39/test262/blob/36cbdcf8ec8cae0e494741034a6cd7f6d83f9325/harness/assert.js#L33), so the assertion messages in this patch can be simplified. Not strictly necessary, though, so @bterlson should probably weigh in before you change it.

test/built-ins/Object/assign/ObjectOverride-sameproperty.js
```diff
@@ -0,0 +1,15 @@
+// Copyright 2015 Microsoft Corporation. All rights reserved.
+// This code is governed by the license found in the LICENSE file.
+
+/*--
+description: Object properties are assigned to target in ascending index order,
+			 i.e. a later assignment to the same property overrides an earlier assignment.
+es6id:  19.1.2.1
+includes: [assert.js]
```
jugglinmike: This helper file is included by default, so an explicit entry in `includes` isn't necessary. We just recently removed this pattern from the project (see https://github.com/tc39/test262/pull/309), so it would be good to avoid adding it back.

test/built-ins/Object/assign/Target-Null.js
```diff
@@ -0,0 +1,13 @@
...
+/*---
+description: Test the first argument(target) of Object.Assign(target,...sources),
+             if target is null,Should Throw a TypeError exception.
+es6id:  19.1.2.1.1
+includes: [assert.js]
+---*/
+
+assert.throws(TypeError, function(){
+                Object.assign(null, {a:1});
+			  });
```
jugglinmike: The formatting here is a little wonky

bterlson:
@yanlxu can you remove the assert.js includes and I'll pull this in? Simplify assert messages if you wish but I'm ok with them as is.

yanlxu:
@bterlson, we have made changes to address review comments

test/built-ins/Object/assign/Override-notstringtarget.js
```diff
@@ -0,0 +1,17 @@
...
+---*/
+
+var target = 12;
+var result = Object.assign(target,"aaa","bb2b","1c");
+
+assert.sameValue(Object.keys(result).length, 4 , "The length should be 4 in the final object.");
+assert.sameValue(result[0], "1", "The value should be {\"0\":\"1\"}.");
+assert.sameValue(result[1], "c", "The value should be {\"1\":\"c\"}.");
+assert.sameValue(result[2], "2", "The value should be {\"2\":\"2\"}.");
+assert.sameValue(result[3], "b", "The value should be {\"3\":\"b\"}.");
```
rwaldron: This exposes a bug in the es6-shim implementation—great work :)

Bug filed: https://github.com/paulmillr/es6-shim/issues/348
Patch: https://github.com/paulmillr/es6-shim/pull/351

test/built-ins/Object/assign/Target-Boolean.js
```diff
@@ -0,0 +1,12 @@
+// Copyright 2015 Microsoft Corporation. All rights reserved.
+// This code is governed by the license found in the LICENSE file.
+
+/*---
+description: Test the first argument(target) of Object.Assign(target,...sources),
+             if target is Boolean,the return value should be a new object whose value is target.
```
jugglinmike: This line (and a few others) has trailing whitespace that interferes with Test262's custom YAML parser. These tests shouldn't be merged until the whitespace is removed, but maybe it's straightforward enough that @bterlson can do it on @yanlxu's behalf.

bterlson: I can do it tomorrow unless @yanlxu can get to it before me.

yanlxu:
@bterlson, we have made changes for the trailing whitespace, please help to check these files, thanks

bterlson:
Seems to work for me, pulling it in!

PR #300, opened 2015-06-03
Add tests for early errors in module syntax

jugglinmike:
@bterlson When it comes to ES6 modules, I think we're limited in what we can
practically test without a loader. Still, it seems like a good idea to get
*some* coverage into test262 because it will pave the way for more rigorous
tests in the future. For instance: harnesses will have to be made aware of
modules so they can parse the source text accordingly. I'm proposing a new
flag, `module`. I prefer this to a boolean frontmatter tag because the relation
with `onlyStrict` and `noStrict` is more clear.

Of all the tests we might write, tests for early errors seem like the most
likely to be useful to implementors today (minimal though they may be).

I've placed the test for `import`'s early error in `language/import` because it
is not technically a statement. At the same time, neither are
FunctionDeclarations, GeneratorDeclarations, ClassDeclarations, or
LexicalDeclarations, so maybe I should take this to mean a loose interpretation
of "statement". I'm happy to move that directory if so.

What do you think?

### DIFF
--- a/test/language/module-code/lex-and-var.js
+++ b/test/language/module-code/lex-and-var.js
@@ -1,0 +1,13 @@
+// Copyright (C) Copyright 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 10.2.1
+description: >
+    It is a Syntax Error if any element of the LexicallyDeclaredNames of
+    ModuleItemList also occurs in the VarDeclaredNames of ModuleItemList.
+flags: [module]
+features: [let]
+---*/
+
+let x;
+var x;
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -115,6 +115,8 @@ This tag is for boolean properties associated with the test.
 
 - **`onlyStrict`** - only run the test in strict mode (*not supported by the browser runner*)
 - **`noStrict`** - only run the test in "sloppy" mode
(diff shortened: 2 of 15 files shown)


bterlson:
language/import is fine. You're right about the declarations, maybe we should have a declarations subdirectory? :)

I'm going to pull this in. Tests will fail until some harness figures out how to test modules on a host that supports them, and that's ok.

test/language/module-code/lex-and-var.js
```diff
@@ -0,0 +1,13 @@
+// Copyright (C) Copyright 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 10.2.1
+description: >
+    It is a Syntax Error if any element of the LexicallyDeclaredNames of
+    ModuleItemList also occurs in the VarDeclaredNames of ModuleItemList.
+flags: [module]
+features: [let]
```
anba: Did you intentionally omit `negative: SyntaxError` here, because the test runners don't yet understand the `module` flag?

jugglinmike: Nope, that is a mistaken omission. I've made a pull request to correct it https://github.com/tc39/test262/pull/312 Thanks!

anba: > I've made a pull request to correct it #312 Thanks!

Great! :+1:

Issue #302, opened 2015-06-04
Remove `onlyStrict`


jugglinmike:
There are plenty of tests that specify the `"use strict";` directive *and* the `onlyStrict` flag:

    $ grep onlyStrict test/ -r -l --null | xargs -0 grep -E "^[\"']use" | wc -l
    124

(Some of these are false positives because not all tests use idiomatic indenting for function bodies.)

I was tempted to call this "unnecessary," (maybe cruft from the Sputnik days), but then I remembered that [the browser runner does not honor the `onlyStrict` flag](https://github.com/tc39/test262/blob/21b739f1dcd008e894384a909af5c3ec458a5dcc/CONTRIBUTING.md#flags). Given that [the browser runner injects each test dependency using a dedicated `<script>` tag](https://github.com/tc39/test262/blob/21b739f1dcd008e894384a909af5c3ec458a5dcc/harness/sth.js#L175-L194), that runner does not suffer from the issue that prompted [the proposal for a `raw` frontmatter flag](https://github.com/tc39/test262/issues/291). That makes this a necessary (but very confusing) pattern.

The most direct solution is to update the browser runner to honor that tag (see gh-301). But as the recently-reported gh-299 argues, by relying on the `onlyStrict` flag, the source text does not technically qualify as ECMAScript. This is a surprising trait for the official language test suite, and I thought it would be worth discussing an alternative. Namely: removing the `onlyStrict`
flag. Here's how it could work

1. Change tests to "real" ECMAScript
   - remove the `onlyStrict` frontmatter flag from tests
   - replace with a `"use strict";` directive
2. Inject includes *after* test bodies
   - Update all helper code to define only functions and to do so using FunctionDeclarations (so `assert.sameValue = function() {};` becomes `function assertSameValue() {}`)
   - Find-and-replace existing references to unhoistable references

Benefits:

1. No need for a `raw` flag
2. Source text is "real" ECMAScript

Trade-offs:

1. Time investment required
2. Less efficiency in test execution (without the `onlyStrict` flag, test runners will needlessly execute those tests in both strict mode and sloppy mode)
3. The introduction of a reliance on function hoisting semantics

Trade-off 1 is partially mitigated by using scripted transformation techniques (and given that this is a test suite, we can verify equivalence in a thorough, automated fashion).

[...]

PR #305, opened 2015-06-08
Update browser runner to honor `onlyStrict` flag

### LINKED ISSUE #301: Update browser runner to honor `onlyStrict`
From the project's CONTRIBUTING.md file:

> #### flags
> **flags**: [list]
>
> This tag is for boolean properties associated with the test.
>
> - **`onlyStrict`** - only run the test in strict mode (*not supported by the browser runner*)
> - **`noStrict`** - only run the test in "sloppy" mode

[The Python runner makes a simple transformation to the source prior to executing it based on the presence of this flag](https://github.com/tc39/test262/blob/21b739f1dcd008e894384a909af5c3ec458a5dcc/tools/packaging/test262.py#L312-L316) ([the experimental Node.js runner does this, as well](https://github.com/bterlson/test262-harness/blob/b4d8adcd0cdce26949a10fad739777214ff4d791/lib/runner.js#L47-L49)). Update the browser runner with equivalent functionality, remove explicit directives from tests that specify `onlyStrict`, and remove the caveat from the contributing guidelines.

jugglinmike:
Resolves #301


anba:
The *onlyStrict* flag should be removed for some test files instead of the `"use strict"` directive. For example in:
- function-code/10.4.3-1-10.js
- function-code/10.4.3-1-10gs.js
- function-code/10.4.3-1-12.js
- statements/function/13.0-10-s.js
- statements/function/13.0-11-s.js
- ... more

I think I'd remove `"use strict"` if *onlyStrict* flag and
  1. either global `"use strict"` directive is present
  2. or `"use strict"` directive in `function testcase(){ "use strict"; ... }` is present

In all other cases the *onlyStrict* flag should be deleted.

jugglinmike:
Didn't mean to close this, sorry about that!

I don't think I can automate that heuristic, so it may be some time before I can make the requested change. I'll comment here when I am able to do so.

anba:
> I don't think I can automate that heuristic, so it may be some time before I can make the requested change. I'll comment here when I am able to do so.

I can do that part, too.

jugglinmike:
Thanks! Actually, I have that part done. It's the removal of `onlyStrict` that is currently giving me trouble.

jugglinmike:
@anba Put another way:

"Remove `onlyStrict` from all files that **do not** declare the 'use strict' directive prologue globally nor within a function passed to the `runTestCase` helper."

I've authored a commit that does this; you can see it [here](https://github.com/jugglinmike/test262/commit/65dfa03ac33a4873bebaaaf85fca1d5ac9fd430f). I don't think removing `onlyStrict` is the best approach in these cases, though.

In addition to enabling strict mode, we're using the `onlyStrict` flag to optimize test runs. Generally, running tests in both modes ensures that strictness does not effect the semantics under test.

The value of running the tests in question in non-strict mode amounts to creating additional tests for the application of the "use strict" directive prologue (a far more limited benefit).

That said, as the author of [the proposal to remove `onlyStrict` entirely](https://github.com/tc39/test262/issues/302), maybe I'm not the best person to make this argument :) @bterlson: does what I'm saying align with your goals for the `onlyStrict` flag?

bterlson:
Yes seems to align. As far as I see it, onlyStrict is used in cases where a test is expected to pass only in strict mode. Such tests should not include a USD. If USD is used it should either be in a tested function or use the raw frontmatter flag.

Many of the removals in your PR seem to fall into this bucket. The tests would not pass without the USD at global scope. @anba I wonder what you think should be done with these tests?

anba:
I think one issue with the current setup is the meaning of the *onlyStrict* flag within certain tests. Apparently sometimes the test author assumed *onlyStrict* means: Only run this test if strict-mode is supported. 

I've pushed a couple of commits to [this branch](https://github.com/anba/test262/commits/issue-305/use-strict-directive). The commits are split up in removing the `"use strict"` directive and removing the *onlyStrict* flag. This should make it easier to identify which test files can/should be changed.

jugglinmike:
Thanks for that, @anba. I'm starting to think that this can't be done with automated processes. For
instance, in [the proposed change for language/statements/function/13.1-15-s.js](https://github.com/tc39/test262/commit/728771ef436adb4978ac1da3161f7df9a29d8d3e#diff-0731936e24629e0152c929a5caa7ea4eL17):

```diff
-flags: [onlyStrict]
 includes: [runTestCase.js]
 ---*/
 
 function testcase() {
 
         try {
             eval("'use strict';function _13_1_15_fun(eval) { }");
             return false;
         } catch (e) {
             return e instanceof SyntaxError;
         }
     }
 runTestCase(testcase);
```

The semantics under test pre-suppose strict mode. Without `strictOnly`, the test will run outside of strict mode. It will pass of course, but this additional test run will only ensure that the directive is applied correctly (which is not what this test was authored to assert). So I don't think `onlyStrict` should be removed here (the directive should be removed instead).

Then again, some tests concern the application of the directive directly, as in [language/statements/with/12.10.1-1-s.js](https://github.com/tc39/test262/commit/728771ef436adb4978ac1da3161f7df9a29d8d3e#diff-8c7cd4a8d0b73788b517c39eaea7a00eL16):

```diff
 description: >
     Strict Mode - SourceElements is evaluated as strict mode code when
     the function body of a Function constructor begins with a Strict
     Directive
-flags: [onlyStrict]
 includes: [runTestCase.js]
 ---*/
 
 function testcase() {
 
         try {
             eval("var _13_0_13_fun = new Function(\" \", \"'use strict'; eval = 42;\"); _13_0_13_fun();");
             return false;
         } catch (e) {
             return e instanceof SyntaxError;
         }
     }
 runTestCase(testcase);
```

In this case, I would actually go so far as to add `noStrict`, since the intent of the test is how the directive is applied. Running this test in strict mode doesn't buy us much (beyond ensuring that the directive does not somehow disable strict mode).

jugglinmike:
Hey @bterlson 

As we discussed, this transformation is probably a bit too nuanced to be made
in an automated fashion. I revisited this patch, starting by reverting all the
test file modifications. Then I reviewed each file that uses `onlyStrict` and
decided how to change it in context--each change took one of five forms:

1. Remove `onlyStrict` flag, add `noStrict` flag
2. Remove `onlyStrict` flag
3. Remove `onlyStrict` flag, remove one or more "use strict" directives
4. Remove one or more "use strict" directive
5. No change

I've organized the test changes into distinct commits for each type (excepting
`#5` of course). The difference between `#1` and `#2` is very subjective and
essentially amounts to test run optimization.

As part of this work, I identified a few invalid tests and a few errors in test
descriptions. I've split those corrections into separate commits in case we
want to land them in advance of this work.

I've double-checked my work by running the final result through [André's
runtime](https://github.com/anba/es6draft).

jugglinmike:
Looks like this branch won't merge cleanly anymore. I will wait to rebase until I hear back from you; please let me know if you'd like me to keep the commits as-is or squash one ore more of them together.

bterlson:
I will check this out in more detail tomorrow.

bterlson:
I think this is good to go. I didn't find any issues with the test updates (seems to have same results).

13 commits seems excessive. Might make sense to separate the web harness update from the collateral updates but otherwise I'd be ok with one giant commit too..

jugglinmike:
@bterlson While rebasing, I grouped all the strict-mode-related changes into a single commit. I left the other test collateral changes as distinct commits because they describe unrelated work, and I'm afraid they would get lost in the noise of the "strict" changes. If you prefer, I can extract those commits from this branch entirely and open a new pull request dedicated to just those changes.

Issue #126, opened 2014-12-01
Remove $FAIL?


domenic:
Seems redundant with $ERROR.

PR #307, opened 2015-06-09
Remove `$FAIL` function

jugglinmike:
This function is equivalent to `$ERROR` (which is automatically included
in test environments). Remove the harness file that defines the
function, remove references to the file from test `includes` lists, and
update scripts to instead invoke the `$ERROR` function.

Resolves gh-126

### DIFF
--- a/test/built-ins/Number/prototype/S15.7.3.1_A3.js
+++ b/test/built-ins/Number/prototype/S15.7.3.1_A3.js
@@ -5,13 +5,12 @@
 info: Number.prototype value is not +0
 es5id: 15.7.3.1_A3
 description: Checking value of Number.prototype property
-includes: [$FAIL.js]
 ---*/
 
 //CHECK#1
 try {
   (Number.prototype != 0);
-  $FAIL('#1: "(Number.prototype != 0);" lead to throwing exception. Actual: '+(Number.prototype != 0));
+  $ERROR('#1: "(Number.prototype != 0);" lead to throwing exception. Actual: '+(Number.prototype != 0));
 } catch (e) {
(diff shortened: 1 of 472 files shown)


anba:
The three new files in test/built-ins/TypedArray/from are most likely part of a different PR, right?

And `$PRINT()` should probably be removed, too. As a follow-up of course. See:
- Object/prototype/S15.2.4_A3.js
- Object/prototype/S15.2.4_A4.js
- Object/prototype/hasOwnProperty/S15.2.4.5_A7.js
- ... more

test/built-ins/Number/prototype/S15.7.3.1_A3.js
```diff
@@ -5,13 +5,12 @@
 info: Number.prototype value is not +0
 es5id: 15.7.3.1_A3
 description: Checking value of Number.prototype property
-includes: [$FAIL.js]
 ---*/
 
 //CHECK#1
 try {
   (Number.prototype != 0);
-  $FAIL('#1: "(Number.prototype != 0);" lead to throwing exception. Actual: '+(Number.prototype != 0));
```
bterlson: I imagine this is not going to be too friendly of a reported error as $ERROR will throw, causing the check in the catch block to fail as well. But, seems ok for now, it's not like $FAIL was helping anyway.

bterlson:
I guess remove the excess tests and this is g2g.

jugglinmike:
You are correct, @anba. I guess I'm trying to keep too many things in the air at once.

@bterlson I've removed those three tests and force-pushed a new commit.

PR #309, opened 2015-06-09
No need to include asserts.js

arv:


arv:
@bterlson @jugglinmike PTAL

This is causing issues with our test runner.

jugglinmike:
There are a few includes left over. This should take care of them https://github.com/arv/test262/pull/1 (apply it however you like). Otherwise this looks good to me! Always nice to simplify

PR #310, opened 2015-06-09
Remove remaining includes for `assert.js` helper

arv:


arv:
@bterlson PTAL

Here are some more from @jugglinmike

PR #313, opened 2015-06-10
Use the more common array syntax for includes

### LINKED ISSUE #295: Can't find propertyHelper.js
I just ran test262 with the SpiderMonkey shell, however during the run I get the following error:
`Error: Can't find: /home/tom/projects/test262/harness/propertyHelper.js`
However manually looking up this path the file exists.

evilpie:
This also fixes #295 for me. We use this array syntax everywhere else for includes:

### DIFF
--- a/test/built-ins/Math/acosh/acosh-descriptor.js
+++ b/test/built-ins/Math/acosh/acosh-descriptor.js
@@ -3,8 +3,7 @@
 
 /*---
 description: Testing descriptor property of Math.acosh
-includes:
-    - propertyHelper.js
+includes: [propertyHelper.js]
 es6id: 20.2.2.3
 ---*/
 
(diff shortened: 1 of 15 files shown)


jugglinmike:
I'm all for consistency, but I also think this is addressing a symptom. If the docs say "YAML", then any valid YAML should be accepted. Otherwise, we have to include caveats like, "YAML except without support for list notation". Beyond being distracting, this is likely to be forgotten by contributors who aren't having problems with list notation, meaning this fix could be accidentally broken again when new tests are added.

@anba suggested a patch to the parser [here](https://github.com/tc39/test262/issues/295#issuecomment-110744341)--does that also fix the problem for you? I think this would be a better solution because it doesn't suffer from those problems.

(An ideal solution would involve an externally-maintained and well-tested YAML parser, but I also know that the Python runner is deprecated, so maybe a quick patch like André is proposing is the most practical way forward.)

evilpie:
@anba patch fixes this problem for me as well. However landing this probably doesn't hurt.

bterlson:
Doesn't hurt :)

PR #315, opened 2015-06-11
Add tests for Object.assign

jugglinmike:
This patch omits a test for step 5.b.ii because I believe it to be unreachable. [I've filed a spec bug to verify](https://bugs.ecmascript.org/show_bug.cgi?id=4391); if I'm mistaken, I will follow up with an additional test.


jugglinmike:
Actually, this is duplicative of gh-293. I'll close it for now and re-open if I find it includes any additional coverage

PR #316, opened 2015-06-12
Improve tests for templates

jugglinmike:
In addition to expanding coverage of ES6 templates, this patch:

- reorganizes existing tests to better-reflect the coverage structure and prefer a more flat file system hierarchy
- updates frontmatter to use more accurate ES6 IDs and shorter descriptions (moving multi-line explanatory text to the "info" tag)


test/language/expressions/tagged-template/call-expression-context.js
```diff
@@ -0,0 +1,21 @@
...
+description: Invocation context for call expressions
+info: >
+    A tagged template is a function call where the arguments of the call are
+    derived from a TemplateLiteral. The actual arguments include a template
+    object and the values produced by evaluating the expressions embedded
+    within the TemplateLiteral.
+---*/
+var context;
+var fn = function() {
+  return function() {
```
caitp: this function never seems to be evaluated

jugglinmike: Good catch! This is a bug in the test as indicated by the file's name, "call-expression-context.js". This ought to be asserting that a tagged template applied to a call expression has the correct context. That means the frontmatter needs the `onlyStrict` flag and the invocation should read `fn()LITERALBACKTICKCHARACTERHERENoSubstitutionTemplateLITERALBACKTICKCHARACTERHERE`

(I need to find a better convention for using backticks within inline code samples)

caitp:
You should add a test asserting the `this` value of a tag function when it is returned from A CallExpression --- a few variations actually, one where it's just a MemberExpression leading with a CallExpression (property of return value), and another which isn't

```js
var globalOrUndefined = (function() { return this; })();

assert.same_whatever((function() {
  return function() {
    return this;
  };
})()`foo`, globalOrUndefined);

var obj = {
  tag: function() { return this; }
};
assert.same_whatever((function() {
  return obj
})().tag`foo`, obj);
```

jugglinmike:
Good ideas all around, @caitp. I've added the tests in fixup! commits (also consistently initializing `context` to `null` to help avoid false positives).

test/language/expressions/tagged-template/call-expression-context-no-strict.js
```diff
@@ -0,0 +1,31 @@
...
+assert.sameValue(context, this);
+
+fn = function() {
+  return () => { context = this; };
+};
+
+context = null;
+fn()`NoSubstitutionTemplate`;
+
+assert.sameValue(context, this);
```
caitp: `this` here is "probably" the global object in most environments, but`context` should be `fn()`'s `this`, which may not match the outer value of `this`

jugglinmike: I think I follow you. What about if I used [`fnGlobalObject`](https://github.com/tc39/test262/blob/42d16da90113c4739df3a8161ea898e80df70697/harness/fnGlobalObject.js) here, like this:

```diff
 context = null;
 fn()`NoSubstitutionTemplate`;

-assert.sameValue(context, this);
+assert.sameValue(context, fnGlobalObject());
```

bterlson:
Sorry for the delay in reviewing this. It looks good.

jugglinmike:
No worries, and thanks!

PR #319, opened 2015-06-14
The Array.prototype.find function is an ES6 function.

pvginkel:
A number of Array.prototype.find tests were missing a feature specification.


bterlson:
This is fine to have, although I think it's appropriate to consider the current directory a file is in as an entry in its feature array (replacing /prototype/ with #).

PR #322, opened 2015-06-16
Remove $PRINT helper function

jugglinmike:
@anba [requested this while reviewing another patch](https://github.com/tc39/test262/pull/307#issuecomment-110505474), and I thought it was a good idea. As noted in the commit message, this work also surfaced some tautological tests.

Commit message:

> The `$PRINT` helper function has no effect on test behavior. The
> following tests use it to report assertion outcomes (and as a result
> will fail silently):
>
> - test/language/function-code/S10.2.1_A4_T2.js
> - test/language/function-code/S10.2.1_A5.1_T1.js
> - test/language/function-code/S10.2.1_A5.1_T2.js
> - test/language/function-code/S10.2.1_A5.2_T1.js
>
> Remove the function definition and all references within tests. Update
> tests that use it as an error reporting mechanism to instead use an
> appropriate `assert` helper function.

### DIFF
--- a/harness/$PRINT.js
+++ b/harness/$PRINT.js
@@ -1,4 +1,0 @@
-//adaptors for Test262 framework
-function $PRINT(message) {
-
-}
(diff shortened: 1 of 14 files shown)


bterlson:
So nice!

Issue #129, opened 2014-12-02
Cleaning up ./harness


domenic:
Currently test/harness seems to contain three things:
- Helpers, e.g. $FAIL.js, arrayContains.js, ...
- Website assets, e.g.  jquery-1.4.2.min.js, jqueryprogressbar.js, ...
- Harness-setup files, e.g. sta.js, sth.js, cth.js, ed.js, ...

These should be separated at the very least. E.g. website assets could go into website/.

It would be good to audit and consolidate or remove the harness-setup files; IMO they probably do not belong in this repo, but perhaps belong in bterlson/test262-harness, or possibly in the repos of the various engines they are adapting for (e.g. V8 already has [their own setup file](https://code.google.com/p/chromium/codesearch#chromium/src/v8/test/test262/harness-adapt.js)).

Once those two are cleaned up it might be nice to rename the folder to something involving "helpers" or "includes" for accuracy.

raphamorim:
I think this issue can be closed. Right?

jugglinmike:
Not necessarily. In addition to moving web scripts (as completed in gh-338), this issue also describes separating "helpers" (i.e. the scripts that may optionally be included via the `includes` frontmatter) from "harness-setup files" (i.e. the scripts that are injected unconditionally). I think we need @bterlson's input on whether that is desired. If so, I'll follow up with a patch. If not, then this issue is effectively resolved.

leobalter:
This seems to be resolved for now. We can reopen it if necessary.

PR #326, opened 2015-06-18
Move includes

jugglinmike:
This is one part of the change described by gh-129. I thought it would be easier to tackle that work one piece at a time, but I'd be happy to also separate website assets from general test setup files as part of this patch. Let me know!


jugglinmike:
I talked this over with @bterlson on IRC. I'm going leave the "includes" files where they are and move the website-specific assets instead. That's substantially different than what I've submitted here, so I'll follow up with a new pull request.

PR #330, opened 2015-06-23
Fix bugs in Promise tests

jugglinmike:
This test's description concerns the behavior of `Promise.all` when the
IteratorStep abstract operation fails due to an abrupt completion
returned by the iterator's `next` method. The test body did not actually
assert that functionality.

Update the test body to correctly define the requisite iterator and
assert that the specific error created is the one thrown from the
invocation of `Promise.all`


jugglinmike:
This bug also exists in an equivalent test for `Promise.race`. I've added a commit to address it there, as well.

jugglinmike:
I've just added another fix for a third Promise test.

PR #331, opened 2015-06-24
Formalize tests for built-in Promise methods 

jugglinmike:
Uniformly and exhaustively test the generic aspects of the static- and
instance Promise methods, including:

- function `length` properties
- function `name` properties
- method property descriptors


bterlson:
Nice updates.

PR #334, opened 2015-06-25
Fix issue with math_precision.js

arv:
It was missing vars which was causing issues with our test harness
since we run all the files in strict mode.

### DIFF
--- a/harness/math_precision.js
+++ b/harness/math_precision.js
@@ -1,13 +1,11 @@
 // Copyright 2009 the Sputnik authors.  All rights reserved.
 // This code is governed by the BSD license found in the LICENSE file.
 
-function getPrecision(num)
-{
+function getPrecision(num) {
 	//TODO: Create a table of prec's,
-	//      because using Math for testing Math isn't that correct. 
-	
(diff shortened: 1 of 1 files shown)


arv:
@bterlson @jugglinmike PTAL

arv:
More background: The V8 test harness runs `d8` with `--use-strict` command line flag and therefore this file fails to parse.

jugglinmike:
This looks good to me. I don't understand how this *ever* worked in strict mode.

Also, @arv, doesn't running with `d8 --use_strict` invalidate all the tests that specify `noStrict`?

arv:
@jugglinmike `noStrict` and `onlyStrict` are taken into account.

https://github.com/v8/v8-git-mirror/blob/master/test/test262-es6/testcfg.py#L84

jugglinmike:
Oooh. I thought you meant you were running with `--use_strict` unconditionally, as in:

    test262.py --command "d8 --use_strict"

arv:
We do not use the test runner provided by tc39/test262

PR #335, opened 2015-06-26
Add tests for Map

leobalter:
Opening this PR to avoid work conflict.

This branch currently contains tests for the Map constructor (23.1.1.1). It might be landed before the tests for the following Map tests (prototype methods, etc.) if you prefer. Otherwise, I'll remove the WIP prefix after covering the remaining tests.


bterlson:
All in one PR is fine by me. But whatever works best for you is what I want!

leobalter:
> All in one PR is fine by me.

Cool! I'll do it this way. 

I just wanted to show the current WIP and avoid conflicts with anyone else working on the same feature, as this will be a long PR.

leobalter:
Completed tests for Map including Map.prototype properties.

Ready for review.

test/built-ins/Map/iterable-calls-set.js
```diff
@@ -0,0 +1,35 @@
...
+  results.push([k, v]);
+  mapSet.call(this, k, v);
+};
+
+var s = new Map(iterable);
+
+assert.sameValue(counter, 2, "`Map.prototype.set` called twice.");
+
+assert(compareArray(results[0], iterable[0]));
+assert(compareArray(results[1], iterable[1]));
```
jugglinmike: This would be a good place to assert the context of the `set` method when invoked by the Map constructor

test/built-ins/Map/iterator-item-second-entry-returns-abrupt.js
```diff
@@ -0,0 +1,48 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 23.1.1.1
+description: >
+  Closes iterator if item first entry completes abruptly.
```
jugglinmike: "first" -> "second"

test/built-ins/Map/iterator-items-are-not-object-close-iterator.js
```diff
@@ -0,0 +1,71 @@
...
+var nextItem;
+var iterable = {};
+iterable[Symbol.iterator] = function() {
+  return {
+    next: function() {
+      return { value: nextItem, done: false };
+    },
+    return: function() {
+      count += 1;
+      print(arguments);
```
jugglinmike: Looks like a debugging statement made it through :)

test/built-ins/Map/iterator-next-failure.js
```diff
@@ -0,0 +1,29 @@
...
+  Map ( [ iterable ] )
+
+  ...
+  9. Repeat
+    a. Let next be IteratorStep(iter).
+    b. ReturnIfAbrupt(next).
+features: [Symbol]
+---*/
+
+var count = 0;
```
jugglinmike: `count` is unused

test/built-ins/Map/iterator-value-failure.js
```diff
@@ -0,0 +1,35 @@
...
+
+  ...
+  9. Repeat
+    ...
+    d. Let nextItem be IteratorValue(next).
+    e. ReturnIfAbrupt(nextItem).
+features: [Symbol]
+---*/
+
+var count = 0;
```
jugglinmike: `count` is unused

test/built-ins/Map/prototype/clear/map-data-list-is-preserved.js
```diff
@@ -0,0 +1,33 @@
...
+  4. Let entries be the List that is the value of M’s [[MapData]] internal slot.
+  5. Repeat for each Record {[[key]], [[value]]} p that is an element of
+  entries,
+    a. Set p.[[key]] to empty.
+    b. Set p.[[value]] to empty.
+  6. Return undefined.
+features: [Symbol]
+---*/
+
+var count = 0;
```
jugglinmike: `count` is unused

test/built-ins/Map/newtarget.js
```diff
@@ -0,0 +1,34 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
```
jugglinmike: This test appears to be a duplicate of test/built-ins/Map/properties-of-map-instances.js

test/built-ins/Map/prototype/constructor/is-map.js
```diff
@@ -0,0 +1,10 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 23.1.3.2
+description: >
+  The initial value of Map.prototype.constructor is the intrinsic object %Map%.
+---*/
+
+assert.sameValue(Map.prototype.constructor, Map);
+assert.sameValue((new Map()).constructor, Map);
```
jugglinmike: Could you check the property descriptor here, too? Once that's done, you can probably nix the directory and just define `constructor.js` (in the same way you're doing with `Symbol.iterator.js`. I guess that file organization is a little inconsistent with the existing tests for Map, so I think it would be okay to do in a follow-up patch.

test/built-ins/Map/prototype/forEach/callback-parameters.js
```diff
@@ -0,0 +1,42 @@
...
+
+map.forEach(callback);
+
+assert.sameValue(results[0].value, 42);
+assert.sameValue(results[0].key, 'foo');
+assert.sameValue(results[0].context, map);
+
+assert.sameValue(results[1].value, 'baz');
+assert.sameValue(results[1].key, 'bar');
+assert.sameValue(results[1].context, map);
```
jugglinmike: Another relevant assertion: `assert.sameValue(results.length, 2);`

test/built-ins/Map/prototype/forEach/iterates-values-deleted-then-readded.js
```diff
@@ -0,0 +1,50 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
```
jugglinmike: This test is redundant with test/built-ins/Map/prototype/forEach/deleted-values-during-foreach.js . I suggest keeping them both but updating the latter to *only* delete a value

test/built-ins/Map/prototype/get/returns-value-normalized-zero-key.js
```diff
@@ -0,0 +1,21 @@
...
+  entries,
+    a. If p.[[key]] is not empty and SameValueZero(p.[[key]], key) is true,
+    return p.[[value]].
+  ...
+---*/
+
+var map = new Map();
+
+map.set(+0, 42);
+assert.sameValue(map.get(-0), 42);
```
jugglinmike: How about an equivalent assertion for a key value of `-0`?

test/built-ins/Map/prototype/get/returns-value.js
```diff
@@ -0,0 +1,26 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
```
jugglinmike: This looks like a less-complete version of test/built-ins/Map/prototype/get/returns-value-different-key-types.js . I think it's safe to remove it.

test/built-ins/Map/prototype/has/returns-boolean.js
```diff
@@ -0,0 +1,31 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
```
jugglinmike: I think this test is made redundant by test/built-ins/Map/prototype/has/return-false-different-key-types.js and test/built-ins/Map/prototype/has/return-true-different-key-types.js. Those other two seem preferable because they have less dependencies (they don't need `Map.prototype.clear` or `Map.prototype.delete`)

test/built-ins/Map/prototype/set/append-new-values-normalizes-zero-key.js
```diff
@@ -0,0 +1,27 @@
...
+
+var map = new Map();
+map.set(-0, 42);
+
+assert.sameValue(map.get(0), 42);
+
+map.clear();
+
+map.set(+0, 43);
+assert.sameValue(map.get(0), 43);
```
jugglinmike: Could you make a second `Map` instance for this assertion? That way implementers can benefit from this test before they have implemented `Map.prototype.clear`

test/built-ins/Map/prototype/set/append-new-values-normalizes-zero-key.js
```diff
@@ -0,0 +1,27 @@
...
+info: >
+  Map.prototype.set ( key , value )
+
+  ...
+  6. If key is −0, let key be +0.
+  7. Let p be the Record {[[key]]: key, [[value]]: value}.
+  8. Append p as the last element of entries.
+  9. Return M.
+  ...
+features: [Symbol]
```
jugglinmike: This tag can be removed

test/built-ins/Map/prototype/set/append-new-values-return-map.js
```diff
@@ -0,0 +1,31 @@
...
+info: >
+  Map.prototype.set ( key , value )
+
+  ...
+  6. If key is −0, let key be +0.
+  7. Let p be the Record {[[key]]: key, [[value]]: value}.
+  8. Append p as the last element of entries.
+  9. Return M.
+  ...
+features: [Symbol]
```
jugglinmike: This tag can be removed

test/built-ins/Map/prototype/set/append-new-values.js
```diff
@@ -0,0 +1,51 @@
...
+assert.sameValue(result.value, 42);
+assert.sameValue(result.key, null);
+
+result = results.pop();
+assert.sameValue(result.value, 2);
+assert.sameValue(result.key, s);
+
+var map2 = new Map();
+result = map2.set.call(map, 1, 'valid call');
+assert.sameValue(map.get(1), 'valid call');
```
jugglinmike: This code involving `map2` seems unrelated to the test. Can it be removed?

test/built-ins/Map/prototype/set/replaces-a-value-normalizes-zero-key.js
```diff
@@ -0,0 +1,23 @@
...
+      i. Set p.[[value]] to value.
+      ii. Return M.
+  ...
+---*/
+
+var m = new Map([[+0, 1]]);
+
+m.set(-0, 42);
+
+assert.sameValue(m.get(+0), 42, 'zero key is normalized in SameValueZero');
```
jugglinmike: How about an equivalent for a key value of `-0`?

jugglinmike:
Hey @leobalter! I have a couple comments that apply to a bunch of files, so I'll leave them here:

- In the `features` tag, `Symbol` means "this test uses the `Symbol` constructor. This is a little different than saying, "this test uses the `Symbol.iterator` well-known Symbol." For many (but not all) of these tests, you'll want to specify `features: [Symbol.iterator]` instead.
- The files named `context-is-not-object.js` and `this-not-object-throw.js` have a description that reads:

    > Throws a TypeError if context is not a non-null Object.

  ...but they include assertions for the behavior when the context is `null`. I think you can just remove "non-null" from the descriptions.

Besides the comments I've left in-line, that's all I have to say for this patch. Nice work!

leobalter:
Thanks for reviewing, that's awesome notes.

> ..but they include assertions for the behavior when the context is null. I think you can just remove "non-null" from the descriptions.

In my humble perspective from the spec: it mentions `Object` as a non-null object, as null is an Object, I want to make clear I'm asserting for all types that aren't a regular object (Array included) and are not null. `non-null` is intentionally mentioned to explain `null` should be included on the assertions for this case.

jugglinmike:
It sounds like you may be thinking of [the `typeof` operator, whose return value isn't completely in-line with the spec language](http://www.ecma-international.org/ecma-262/6.0/#sec-typeof-operator). Actually, [the spec defines `null` as a Type that is distinct from Object](http://www.ecma-international.org/ecma-262/6.0/#sec-ecmascript-data-types-and-values). I'm recommending that change to the test descriptions so the text more closely aligns with the spec language (e.g. in [Set.prototype.add](http://www.ecma-international.org/ecma-262/6.0/#sec-set.prototype.add): "2. If Type(S) is not Object, throw a TypeError exception."), where a value that is "not an Object" may be null.

leobalter:
> I'm recommending that change to the test descriptions so the text more closely aligns with the spec language 

This is very reasonable. I'll change the descriptions.

Thanks for the feedback.

test/built-ins/Map/iterator-items-are-not-object-close-iterator.js
leobalter: oops :+1:

domenic:
Note that "context" is not a correct way to refer to a function's thisArg. In the context of ES, it usually means "execution context," which is a completely different thing.

leobalter:
@domenic I fixed this and also avoided using "context" replacing it for appropriate meanings..

arv:
Another common term for `this` is receiver.

bterlson:
I think this is good for check in if you want to squash it!

leobalter:
> I think this is good for check in if you want to squash it!

done, thanks!

PR #337, opened 2015-06-26
Remove unnecessary "include" file

jugglinmike:
The `fnExists` function defines a generic way to determine if any number
of values are function instances. Because it is only used by a single
test, the additional complexity required by the generalized code (and
the organizational drawbacks to maintaining another "include" file) are
not justified. Remove the file and update the test to assert the
function's existence directly.


bterlson:
hmm for some reason this rings a bell but I also didn't find any other usage so I think this change is good.

Issue #291, opened 2015-05-30
Proposal: `raw` frontmatter flag


jugglinmike:
Many of the directive prologue tests are invalidated by the code transformation
that takes place prior to test execution. For instance, see
[test/language/directive-prologue/10.1.1-2gs.js](https://github.com/tc39/test262/blob/527644d21199d5e9fcab0b958c3426cbd0d0b9d6/test/language/directive-prologue/10.1.1-2gs.js):

```js
/*---
es5id: 10.1.1-2gs
description: >
    Strict Mode - Use Strict Directive Prologue is ''use strict''
    which lost the last character ';'
negative: NotEarlyError
flags: [noStrict]
---*/

"use strict"
throw NotEarlyError;
var public = 1;
```

First of all: this test is incorrect. Any attempt to execute that code should
fail with an *early* `SyntaxError`, not the `NotEarlyError` object during
runtime. The reason it passes right now is that test runners insert code into
the beginning of the test, so this code is never executed in strict mode.

One fix would be to simply change the `noStrict` flag to `onlyStrict`, but this
is generally an issue for tests involving directives; that substitution won't
work for tests that require more nuanced manipulation of the directive, e.g.
the also-incorrect
[test/language/directive-prologue/10.1.1-8gs.js](https://github.com/tc39/test262/blob/527644d21199d5e9fcab0b958c3426cbd0d0b9d6/test/language/directive-prologue/10.1.1-8gs.js):

```js
/*---
es5id: 10.1.1-8gs
description: >
    Strict Mode - Use Strict Directive Prologue is ''use strict';'
    which appears twice in the code
negative: NotEarlyError
flags: [noStrict]
---*/

"use strict";
"use strict";
throw NotEarlyError;
var public = 1;
```

In order to actually test the described behavior, these tests must be run
without the insertion of any code at all. Rather than opt all tests out of the
"default" helper logic, I think it makes sense to implement a new flag called
`raw` that instructs runners to do no pre-processing prior to execution.

```diff
 /*---
 es5id: 10.1.1-8gs
 description: >
     Strict Mode - Use Strict Directive Prologue is ''use strict';'
     which appears twice in the code
-negative: NotEarlyError
-flags: [noStrict]
+negative: SyntaxError
+flags: [raw]
 ---*/

 "use strict";
 "use strict";
-throw NotEarlyError;
+throw new Error("Not an early error");
 var public = 1;
```

Does this seem right to you, @bterlson?

bterlson:
[...]

Issue #128, opened 2014-12-02
Removing the Python harness


domenic:
As mentioned in #127 the Python harness should be removed.

Pieces of it are still used by V8, but we should just move those to V8 instead of keeping them in this repo.

PR #340, opened 2015-06-26
Implement `raw` flag

jugglinmike:
Some tests involving the directive prologue are invalidated by source
text transformations that insert executable code in the beginning of the
script. Implement a `raw` flag that allows these tests to opt-out of
this transformation. Update the relevant tests to use this flag (and
remove references to globals only available when code is injected).

Update the Python runner accordingly:

- Do not run tests marked as "raw" in strict mode
- Reject invalid test configurations

Update the browser runner accordingly:

- Do not modify the script body of tests marked as "raw"

### DIFF
--- a/test/language/directive-prologue/10.1.1-2gs.js
+++ b/test/language/directive-prologue/10.1.1-2gs.js
@@ -9,10 +9,10 @@ es5id: 10.1.1-2gs
 description: >
     Strict Mode - Use Strict Directive Prologue is ''use strict''
     which lost the last character ';'
-negative: NotEarlyError
-flags: [noStrict]
+negative: SyntaxError
+flags: [raw]
 ---*/
 
 "use strict"
-throw NotEarlyError;
+throw new Error("This code should not execute");
 var public = 1;
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -117,6 +117,9 @@ This tag is for boolean properties associated with the test.
 - **`noStrict`** - only run the test in "sloppy" mode
 - **`module`** - interpret the source text as [module
   code](http://www.ecma-international.org/ecma-262/6.0/#sec-modules)
+- **`raw`** - execute the test without any modification (no helpers will be
+  available); necessary to test the behavior of directive prologue; implies
+  `noStrict`
 
 #### features
 **features**: [list]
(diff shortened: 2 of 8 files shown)


arv:
Hmmm... Why is this needed? In V8 test harness we never modify the script body. When is the script body modified in the tc39/test262 harness?

jugglinmike:
@arv The Python runner injects code ([some files unconditionally](https://github.com/tc39/test262/blob/408318bfb0f2f2d93498916fc4903cd2ee7f4374/tools/packaging/test262.py#L299-L301), others [based on the `includes` frontmatter](https://github.com/tc39/test262/blob/408318bfb0f2f2d93498916fc4903cd2ee7f4374/tools/packaging/test262.py#L309)).

I suspect the V8 harness can avoid this by executing multiple files in the same environment (likely using [the mechanism that André described](https://github.com/tc39/test262/issues/291#issuecomment-108639290) in the issue where I first proposed this change). But as noted there, there's no standard mechanism for doing that. So while the [Python runner may be deprecated](https://github.com/tc39/test262/issues/128), this will continue to be a concern as long as there is a need to define a generic runner.

arv:
@jugglinmike Thanks.

test/language/directive-prologue/10.1.1-2gs.js
```diff
@@ -9,10 +9,10 @@ es5id: 10.1.1-2gs
 description: >
     Strict Mode - Use Strict Directive Prologue is ''use strict''
     which lost the last character ';'
-negative: NotEarlyError
-flags: [noStrict]
```
arv: Can we keep `noStrict` here so that we (V8) do not run this test in both strict and sloppy mode?

jugglinmike: The configuration `[raw, noStrict]` is a little redundant with the definition I'm proposing for `raw` (above):

> **`raw`** - execute the test without any modification (no helpers will be
>  available); necessary to test the behavior of directive prologue; implies
>  `noStrict`

We could do this to maintain backwards compatibility with the V8 runner, but since specifying both flags is not technically necessary, any future change could introduce a breaking test. Practically speaking, I don't think we expect too many new tests to need the `raw` flag, but it still seems safer to me for the V8 runner to update to understand the new semantics. Is that a feasible alternative? If so, I'd be happy to follow up with that change if/when this patch lands in `master`.

arv: Updating the v8 test runner seems doable.

bterlson:
Looks good if you want to squash. (Full disclosure: as usual I did not vet the python harness changes)

jugglinmike:
All squashed

PR #342, opened 2015-06-29
Make tests more strict

jugglinmike:
In ECMAScript 5, assignment to a non-reference value throws a runtime
ReferenceError. ECMAscript 6 specifies an early ReferenceError in these
cases. Tests for this behavior have been authored to pass in both cases.
Simplify these tests to describe and assert the early error.


bterlson:
Nice find.

PR #346, opened 2015-07-01
monkeyYaml: Add support for line folding

arv:
Fixes #345

Issue #345, opened 2015-07-01
monkeyYaml is confused


arv:
monkeyYaml fails on https://github.com/tc39/test262/blob/master/test/built-ins/Object/assign/ObjectOverride-sameproperty.js with the error mesage

```
Traceback (most recent call last):
  File "/mnt/data/b/build/slave/linux/build/v8/tools/run-tests.py", line 692, in <module>
    sys.exit(Main())
  File "/mnt/data/b/build/slave/linux/build/v8/tools/run-tests.py", line 518, in Main
    code = Execute(arch, mode, args, options, suites, workspace)
  File "/mnt/data/b/build/slave/linux/build/v8/tools/run-tests.py", line 618, in Execute
    for v in s.VariantFlags(t, variant_flags) ]
  File "/mnt/data/b/build/slave/linux/build/v8/test/test262-es6/testcfg.py", line 86, in VariantFlags
    test_record = self.GetTestRecord(testcase)
  File "/mnt/data/b/build/slave/linux/build/v8/test/test262-es6/testcfg.py", line 114, in GetTestRecord
    testcase.path)
  File "/mnt/data/b/build/slave/linux/build/v8/test/test262-es6/data/tools/packaging/parseTestRecord.py", line 106, in parseTestRecord
    yamlAttrParser(testRecord, attrs, name)
  File "/mnt/data/b/build/slave/linux/build/v8/test/test262-es6/data/tools/packaging/parseTestRecord.py", line 81, in yamlAttrParser
    parsed = yamlLoad(body)
  File "/mnt/data/b/build/slave/linux/build/v8/test/test262-es6/data/tools/packaging/monkeyYaml.py", line 33, in load
    raise Exception("monkeyYaml is confused at " + line)
Exception: monkeyYaml is confused at              i.e. a later assignment to the same property overrides an earlier assignment.
```

@smikes

jugglinmike:
That looks like invalid YAML. I'll submit a patch to fix it post-haste

jugglinmike:
Looks like I spoke too soon: [YAML supports line folding](http://www.yaml.org/spec/1.2/spec.html#id2779048), so monkeyYAML should probably be updated.

arv:
I'm fixing monkeyYaml. Almost done.

jugglinmike:
Great!

PR #347, opened 2015-07-01
Use "monkeyYaml" in all environments

jugglinmike:
The "monkeyYaml" parser is intended to serve as a lightweight fallback
to Python's standard YAML parser in contexts where the latter is not
available. Any intentionally-simplified implementation will necessarily
exhibit non-standard behavior for different input, so not all input
accepted by the standard parser will be accepted by "monkeyYaml". If
loaded exclusively in fallback situations, these edge cases can only be
identified (and debugged) in the environments that require the fallback.
This has allowed developers to unknowingly author tests that cause
errors.

Update the test runner to use "monkeyYaml" in all cases, ensuring more
consistent behavior across contexts and precluding this class of
regression.

This depends on gh-346

### DIFF
--- a/tools/packaging/parseTestRecord.py
+++ b/tools/packaging/parseTestRecord.py
@@ -113,12 +113,8 @@ def importYamlLoad():
     global yamlLoad
     if yamlLoad:
         return
-    try:
-        import yaml
-        yamlLoad = yaml.load
-    except:
-        monkeyYaml = loadMonkeyYaml()
-        yamlLoad = monkeyYaml.load
+    monkeyYaml = loadMonkeyYaml()
+    yamlLoad = monkeyYaml.load
 
 def loadMonkeyYaml():
     f = None


arv:
I think this is worth doing and I doubt that there will be any performance implications.

Also see: https://code.google.com/p/v8/issues/detail?id=4263

jugglinmike:
I should admit that I'm only assuming the motivation for monkeyYaml. This issue (along with gh-346 and gh-295 before it) prompts a few questions: is monkeyYaml still necessary? Which environments rely on it? Would it be possible for them to use the YAML parser from the Python standard library instead?

arv:
I would prefer if we could get all our bots to have yaml installed. Then we could remove the monkeyYaml code from this repo.

arv:
our == v8's

jugglinmike:
I actually agree with you on that. I wrote this patch under the assumption that would not be an option (but if not, I'd love to learn why).

domenic:
+1 for not dynamically switching based on runtime behavior. Whether that means removing monkeyYaml or using it always, I am ambivalent.

That said, I think the real solution here is to remove the Python stuff from this repo and make it completely V8's responsibility, since the official runner is the Node-based test262-harness.

smikes:
You are correct about the purpose behind monkeyYaml; I wrote it to bridge the gap between build envs that had Python but not pyyaml , when we switched from custom metadata to YAML based metadata.

I would prefer to drop monkeyYaml entirely if the v8 team are on board.  Domenic's solution is the ideal one, but again, requires v8 buy in to get node on all their test platforms.

domenic:
@smikes no, it doesn't require the V8 team to upgrade to the node runner; it just requires the V8 team to check the Python code that's current checked in to tc39/test262, into v8/v8-git-mirror.

smikes:
@domenic Sorry, I misread your suggestion. I thought you were proposing that v8 switch to the official test262 runner.  Indeed, it would be enough for v8 to take over the test262 adapter code.

My apologies.

arv:
Switching to node is not going to happen. There is still a chance that we might be able to update our bots to install pyyaml by default.

domenic:
Yes, we know V8 is not going to update its infrastructure to use the official runner. I am just suggesting we get the unofficial runner's code out of this repo and into the v8 repo.

arv:
@domenic I see. That might work (as long as V8 finds someone to do the work).

jugglinmike:
@arv A [recent comment of yours](https://github.com/tc39/test262/pull/334#issuecomment-115406116) has me confused about all this:

> We do not use the test runner provided by tc39/test262

I took "we" to mean "the V8 team" and "test runner" to mean "[`test262/tools/packaging/test262.py`](https://github.com/tc39/test262/blob/f4e17963f851e583e78c58a6de7e031e170a6936/tools/packaging/test262.py)". I think I'm mistaken, though, because if that were true, then it would seem like [problems with monkeyYaml](https://github.com/tc39/test262/issues/345) would not effect you at all. Maybe the V8 runner just depends on some of the Python modules here, but not the runner itself... Can you elaborate on how (and where) your projects depend on code hosted in this repository?

smikes:
In the V8 repository there is a little Python bridge module (test/test262-es6/testing.py) that loads (from test262) tools/packaging/parseTestRecord.py - which now depends unconditionally on monkeyYaml

This was hacked up by me about 15 months ago, and is now primarily maintained by V8 folks (arv, rossberg?)

smikes:
Note that the Python code is the only way to produce the "official" website in the style that existed for ES5.  So removing that implies committing to some new website production system.

PR #349, opened 2015-07-02
Add tests for String.fromCodePoint

leobalter:


test/built-ins/String/fromCodePoint/to-number-conversions.js
```diff
@@ -0,0 +1,24 @@
...
+    ...
+  6. Return the String value whose elements are, in order, the elements in the
+  List elements. If length is 0, the empty string is returned.
+
+  Ref: 7.1.3 ToNumber ( argument )
+---*/
+
+assert.sameValue(String.fromCodePoint(null), '\x00');
+assert.sameValue(String.fromCodePoint(false), '\x00');
+assert.sameValue(String.fromCodePoint(true), '\x01');
```
jugglinmike: How about including assertions for other ToNumber conversions?

    String.fromCodePoint('42');
    String.fromCodePoint({ valueOf: function() { return 31; } });

test/built-ins/String/fromCodePoint/argument-is-not-number.js
```diff
@@ -0,0 +1,33 @@
...
+
+var obj = {};
+Object.defineProperty(obj, 'item', {
+  get: function() {
+    throw new Test262Error();
+  }
+});
+
+assert.throws(Test262Error, function() {
+  String.fromCodePoint(obj.item);
```
jugglinmike: These assertions pass, but not because of the behavior of `String.fromCodePoint`. The expression `obj.item` throws a `Test262Error` prior to the invocation of the method under test.

The file name and `info` meta data suggest that you want to ensure that abrupt completions from ToNumber are returned. I can think of two ways to do  this:

    String.fromCodePoint(Symbol('s'));
    String.fromCodePoint({ valueOf: function() { throw new Test262Error(); });

test/built-ins/String/fromCodePoint/argument-is-not-number.js
```diff
@@ -0,0 +1,33 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
```
jugglinmike: I think a better name for this file would be `argument-not-number-coercible.js` or simply `argument-not-coercible.js`  because (as your other tests confirm), `String.fromCodePoint` does work with non-number values.

test/built-ins/String/fromCodePoint/return-string-value.js
```diff
@@ -0,0 +1,36 @@
...
+  6. Return the String value whose elements are, in order, the elements in the
+  List elements. If length is 0, the empty string is returned.
+---*/
+
+assert.sameValue(String.fromCodePoint(0), '\x00');
+assert.sameValue(String.fromCodePoint(42), '*');
+assert.sameValue(String.fromCodePoint(65, 90), 'AZ');
+assert.sameValue(String.fromCodePoint(0x404), '\u0404');
+assert.sameValue(String.fromCodePoint(0x2F804), '\uD87E\uDC04');
+assert.sameValue(String.fromCodePoint(194564), '\uD87E\uDC04');
```
jugglinmike: To tightly assert boundaries, a good complement to `number-is-out-of-range.js` would be `String.fromCodePoint(1114111)`

test/built-ins/String/fromCodePoint/return-string-value.js
```diff
@@ -0,0 +1,36 @@
...
+assert.sameValue(
+  String.fromCodePoint(0x1D306, 0x61, 0x1D307),
+  '\uD834\uDF06a\uD834\uDF07'
+);
+
+assert.sameValue(
```
jugglinmike: This final assertion seems unrelated (and is essentially a weaker form of the first assertion). Do you think you can remove it?

test/built-ins/String/fromCodePoint/arguments-is-empty.js
```diff
@@ -0,0 +1,26 @@
...
+    f. Append the elements of the UTF16Encoding (10.1.1) of nextCP to the end of
+    elements.
+    g. Let nextIndex be nextIndex + 1.
+  6. Return the String value whose elements are, in order, the elements in the
+  List elements. If length is 0, the empty string is returned.
+
+---*/
+
+assert.sameValue(String.fromCodePoint(), '');
+assert.sameValue(
```
jugglinmike: This final assertion seems unrelated, and it describes behavior that is in conflict with the file name. Do you think you can remove it?

jugglinmike:
Hey Leo: I left some feedback for you as in-line comments. Nice work so far!

leobalter:
@jugglinmike TY!

Everything is fixed.

leobalter:
squashed and rebased with master

leobalter:
Thank you @bterlson!

Issue #278, opened 2015-05-20
RegExp tests incorrectly verify that RegExp.prototype is a RegExp object


bterlson:
According to https://people.mozilla.org/~jorendorff/es6-draft.html#sec-properties-of-the-regexp-prototype-object, it is not.

jugglinmike:
Canonical reference: http://www.ecma-international.org/ecma-262/6.0/#sec-properties-of-the-regexp-prototype-object

@bterlson I can't find the tests mentioned in this issue's title. Maybe this has been fixed?

bterlson:
Perhaps :)

bterlson:
As in, I can't find it either.

PR #352, opened 2015-07-06
Add tests for RegExp `u` flag

jugglinmike:
I'd like to draw attention to two tests in particular since they both concern a very specific condition (for example, good ol' step 8.g.iv.5.d):

- test/built-ins/RegExp/prototype/Symbol.match/u-lastindex.js
- test/built-ins/RegExp/prototype/Symbol.replace/u-lastindex.js

Despite a careful reading, I'm still not convinced those test cases properly express the detail documented in their info/description. I decided I may have looked at it for too long and that some fresh perspective would help

@mathiasbynens Would you mind taking a look?

### DIFF
--- a/test/built-ins/RegExp/prototype/Symbol.match/get-unicode-error.js
+++ b/test/built-ins/RegExp/prototype/Symbol.match/get-unicode-error.js
@@ -1,0 +1,34 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: >
+    Errors thrown by `unicode` accessor are forwarded to the runtime for global patterns
+es6id: 21.2.5.6
+info: >
+    21.2.5.6 RegExp.prototype [ @@match ] ( string )
+
+    [...]
+    8. Else global is true,
+       a. Let fullUnicode be ToBoolean(Get(rx, "unicode")).
+       b. ReturnIfAbrupt(fullUnicode).
+features: [Symbol.match]
+---*/
+
+var nonGlobalRe = /./;
+var globalRe = /./g;
+var accessor = function() {
+  throw new Test262Error();
+};
+Object.defineProperty(nonGlobalRe, 'unicode', {
+  get: accessor
+});
+Object.defineProperty(globalRe, 'unicode', {
+  get: accessor
+});
+
+nonGlobalRe[Symbol.match]('');
+
+assert.throws(Test262Error, function() {
+  globalRe[Symbol.match]('');
+});
--- a/test/built-ins/RegExp/prototype/Symbol.match/u-advance-after-empty.js
+++ b/test/built-ins/RegExp/prototype/Symbol.match/u-advance-after-empty.js
@@ -1,0 +1,30 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: lastIndex is advanced according to width of astral symbols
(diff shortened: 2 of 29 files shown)


test/built-ins/RegExp/prototype/Symbol.match/u-lastindex.js
```diff
@@ -0,0 +1,28 @@
...
+                 [...]
+                 c. Let nextIndex be AdvanceStringIndex(S, thisIndex,
+                    fullUnicode).
+                 d. Let setStatus be Set(rx, "lastIndex", nextIndex, true).
+features: [Symbol.match]
+---*/
+
+var r = /(𝌆)?/ug;
+r[Symbol.match]('𝌆');
+assert.sameValue(r.lastIndex, 2);
```
anba: `lastIndex` is reset to `0` in 21.2.5.2.2 RegExpBuiltinExec, step 15.a.i when the loop in 21.2.5.6, step 8.g is executed the third (and last) time.

jugglinmike: Ah, okay. So I think this behavior can't be externally verified through inspection of the `lastIndex` property. What about this instead:

`test/built-ins/RegExp/prototype/Symbol.match/u-lastindex.js`, renamed to `test/built-ins/RegExp/prototype/Symbol.match/u-advance-after-empty.js`

```js
var match = /^|\udf06/ug[Symbol.match]('\ud834\udf06');

assert(match !== null);
assert.sameValue(match.length, 1);
assert.sameValue(match[0], '');
```

`test/built-ins/RegExp/prototype/Symbol.replace/u-lastindex.js`, renamed to `test/built-ins/RegExp/prototype/Symbol.replace/u-advance-after-empty.js`

```js
var str = /^|\udf06/ug[Symbol.replace]('\ud834\udf06', 'XXX');
assert.sameValue(str, 'XXX\ud834\udf06');
```

anba: Yes, that should work. Alternatively it's also possible to use `/\b/` and `/\B/` assertions to execute those steps. Like in: https://github.com/anba/es6draft/blob/master/src/test/scripts/suite/regress/bug4159.js

jugglinmike:
I've fixed the bugs that @anba identified in a fixup commit

goyakin:
These are good tests @jugglinmike. It might be better to add the following tests to improve coverage:

  * Coercion of the `unicode` property to Boolean
  * `length` of the `unicode` get accessor function
  * `Symbol.search`

jugglinmike:
Sounds good to me. I've added tests for those three behaviors

mathiasbynens:
You could also add a test to see if running `RegExp.prototype.unicode;` (i.e. `this` is `RegExp.prototype`) throws.

jugglinmike:
Hey Mathias! That test actually describes to distinct expectations:

1. `RegExp.prototype` is not a RegExp instance
2. `get RegExp.prototype.unicode` requires the "this" value to have an [[OriginalFlags]] internal slot

I think it will be more clear (and maintainable) to test `RegExp.prototype instanceof RegExp` in one place and then only test for the second behavior generically (i.e. with *any* object missing the expected internal slot) in all other cases. The former is outside the scope of this pull request (though there is an open issue for it: https://github.com/tc39/test262/issues/278), but this patch includes a test for the latter (https://github.com/tc39/test262/pull/352/files#diff-09b2f28b81fb4f5d4d84056cdced1ba6).

bterlson:
So I think this is good for :eggplant:?

jugglinmike:
Done and done

PR #353, opened 2015-07-06
Add test for built-in accessor properties names

leobalter:
After https://bugzilla.mozilla.org/show_bug.cgi?id=1180290, I'm writing these tests to catch all the built in accessor properties names.

Including:

- 21.2.4.2 get RegExp [ @@species ]
- 21.2.5.3 get RegExp.prototype.flags
- 21.2.5.4 get RegExp.prototype.global
- 21.2.5.5 get RegExp.prototype.ignoreCase
- 21.2.5.7 get RegExp.prototype.multiline
- 21.2.5.10 get RegExp.prototype.source
- 21.2.5.12 get RegExp.prototype.sticky
- 22.1.2.5 get Array [ @@species ]
- 23.1.2.2 get Map [ @@species ]
- 23.2.2.2 get Set [ @@species ]
- 24.1.3.3 get ArrayBuffer [ @@species ]
- 24.1.4.1 get ArrayBuffer.prototype.byteLength
- 24.2.4.1 get DataView.prototype.buffer
- 24.2.4.2 get DataView.prototype.byteLength
- 24.2.4.3 get DataView.prototype.byteOffset
- 25.4.4.6 get Promise [ @@species ]
- B.2.2.1.1 get Object.prototype.__proto__
- B.2.2.1.2 set Object.prototype.__proto__

Skipped TypedArray tests in order to avoid a premature test strategy definition for them.

PR #354, opened 2015-07-08
Add tests for RegExp `y` flag

jugglinmike:
- I have omitted a test for the name of [`get
  RegExp.prototype.sticky`](http://www.ecma-international.org/ecma-262/6.0/#sec-get-regexp.prototype.sticky)
  since @leobalter has that covered in gh-353.
- ~~[The `sticky` flag will trigger a Set operation to the `lastIndex` property
  in two
  cases](http://www.ecma-international.org/ecma-262/6.0/#sec-regexpbuiltinexec).
  The spec guards each case with an explicit call to ReturnIfAbrupt, but [due
  to the way RegExp objects are
  allocated](http://www.ecma-international.org/ecma-262/6.0/#sec-regexpalloc),
  I don't believe this operation can ever fail (even for wacky RegExp
  subclasses). If anyone has any ideas on how to test that behavior, I'd love
  to add the relevant tests. Otherwise, I'll follow up with a proposal to
  simplify the next version of the spec by removing those lines (or maybe even
  replacing with Asserts) on the ECMAScript bug tracker~~ (@anba pointed me in the right direction below)

### DIFF
--- a/test/built-ins/RegExp/prototype/Symbol.match/get-sticky-coerce.js
+++ b/test/built-ins/RegExp/prototype/Symbol.match/get-sticky-coerce.js
@@ -1,0 +1,54 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Boolean coercion of `sticky` property
+es6id: 21.2.5.6
+info: >
+    21.2.5.6 RegExp.prototype [ @@match ] ( string )
+
+    [...]
+    5. Let global be ToBoolean(Get(rx, "global")).
+    6. ReturnIfAbrupt(global).
+    7. If global is false, then
+       a. Return RegExpExec(rx, S).
+
+    21.2.5.2.2 Runtime Semantics: RegExpBuiltinExec ( R, S )
+
+    [...]
+    8. Let sticky be ToBoolean(Get(R, "sticky")).
+features: [Symbol, Symbol.match]
+---*/
+
+var r = /a/;
+Object.defineProperty(r, 'sticky', { writable: true });
+
+r.sticky = undefined;
+assert.notSameValue(r[Symbol.match]('ba'), null);
+
+r.sticky = null;
+assert.notSameValue(r[Symbol.match]('ba'), null);
+
+r.sticky = true;
+assert.sameValue(r[Symbol.match]('ba'), null);
+
+r.sticky = false;
+assert.notSameValue(r[Symbol.match]('ba'), null);
+
+r.sticky = NaN;
+assert.notSameValue(r[Symbol.match]('ba'), null);
+
+r.sticky = 0;
+assert.notSameValue(r[Symbol.match]('ba'), null);
+
+r.sticky = 86;
+assert.sameValue(r[Symbol.match]('ba'), null);
+
+r.sticky = '';
+assert.notSameValue(r[Symbol.match]('ba'), null);
+
+r.sticky = Symbol();
+assert.sameValue(r[Symbol.match]('ba'), null);
+
+r.sticky = {};
+assert.sameValue(r[Symbol.match]('ba'), null);
--- a/test/built-ins/RegExp/prototype/Symbol.match/get-sticky-err.js
+++ b/test/built-ins/RegExp/prototype/Symbol.match/get-sticky-err.js
@@ -1,0 +1,33 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Behavior if error is thrown when accessing `sticky` property
+es6id: 21.2.5.6
+info: >
+    21.2.5.6 RegExp.prototype [ @@match ] ( string )
+
+    [...]
+    5. Let global be ToBoolean(Get(rx, "global")).
+    6. ReturnIfAbrupt(global).
+    7. If global is false, then
+       a. Return RegExpExec(rx, S).
+
+    21.2.5.2.2 Runtime Semantics: RegExpBuiltinExec ( R, S )
+
+    [...]
+    8. Let sticky be ToBoolean(Get(R, "sticky")).
+    9. ReturnIfAbrupt(sticky).
+features: [Symbol.match]
+---*/
+
+var r = /./;
+Object.defineProperty(r, 'sticky', {
+  get: function() {
+    throw new Test262Error();
+  }
+});
+
+assert.throws(Test262Error, function() {
+  r[Symbol.match]();
+});
--- a/test/built-ins/RegExp/prototype/Symbol.match/y-fail-return.js
+++ b/test/built-ins/RegExp/prototype/Symbol.match/y-fail-return.js
@@ -1,0 +1,33 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Stops match execution after first match failure
+es6id: 21.2.5.6
+info: >
+    21.2.5.6 RegExp.prototype [ @@match ] ( string )
+
+    [...]
+    5. Let global be ToBoolean(Get(rx, "global")).
+    6. ReturnIfAbrupt(global).
+    7. If global is false, then
+       a. Return RegExpExec(rx, S).
+
+    21.2.5.2.2 Runtime Semantics: RegExpBuiltinExec ( R, S )
+
+    [...]
+    4. Let lastIndex be ToLength(Get(R,"lastIndex")).
+    [...]
+    8. Let sticky be ToBoolean(Get(R, "sticky")).
+    [...]
+    15. Repeat, while matchSucceeded is false
+        [...]
+        b. Let r be matcher(S, lastIndex).
+        c. If r is failure, then
+           i. If sticky is true, then
+              [...]
+              3. Return null.
+features: [Symbol.match]
+---*/
+
+assert.sameValue(/b/y[Symbol.match]('ab'), null);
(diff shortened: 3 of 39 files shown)


anba:
Set() can fail when `lastIndex` was made non-writable:
Steps 15.c.i.1-2: `re = /a/y; re.lastIndex = 1; Object.freeze(re); re.exec("b");`
Steps 18.a-b: `re = /a/y; Object.freeze(re); re.exec("a");`

jugglinmike:
> Set() can fail when lastIndex was made non-writable

Thanks! I've added the relevant tests

test/built-ins/RegExp/prototype/Symbol.match/y-set-lastindex.js
```diff
@@ -0,0 +1,29 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Sets the `lastIndex` property to the index of the first match
```
goyakin: This should say "end index" instead of "index" to be fully correct.

test/built-ins/RegExp/prototype/Symbol.match/y-init-lastindex.js
```diff
@@ -0,0 +1,35 @@
...
+features: [Symbol.match]
+---*/
+
+var r = /./y;
+var match;
+r.lastIndex = 1;
+
+match = r[Symbol.match]('abc');
+
+assert(match !== null);
```
goyakin: Using `assert.notSameValue` here would give a more descriptive output including the actual and expected values.

test/built-ins/RegExp/prototype/Symbol.match/y-fail-return.js
```diff
@@ -0,0 +1,33 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Stops match execution after first match failure
```
goyakin: Similar to this test case, should we add another one making sure that the loop at **8. g.** stops at the first non-match when the **global** flag is present?

test/built-ins/RegExp/prototype/Symbol.match/y-fail-return.js
```diff
@@ -0,0 +1,33 @@
...
+es6id: 21.2.5.6
+info: >
+    21.2.5.6 RegExp.prototype [ @@match ] ( string )
+
+    [...]
+    5. Let global be ToBoolean(Get(rx, "global")).
+    6. ReturnIfAbrupt(global).
+    7. If global is false, then
```
goyakin: Similar to this test case, should we add another one making sure that the loop at 8. g. stops at the first non-match when the global flag is present?

test/built-ins/RegExp/prototype/Symbol.replace/y-fail-return.js
```diff
@@ -0,0 +1,31 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Stops match execution after first match failure
+es6id: 21.2.5.8
+info: >
+    21.2.5.8 RegExp.prototype [ @@replace ] ( string, replaceValue )
+
+    [...]
+    13. Repeat, while done is false
```
goyakin: Similar to my comment about the test case for the existence of the **global** flag with **Symbol.match**, it would provide additional coverage if we also have a test case for **Symbol.replace**.

An example assertion would be `assert.sameValue(/a/gy[Symbol.replace]('aaba', 'c'), 'ccba')`.

goyakin:
In addition to my inline comments, it would make the test cases more comprehensive if we also check:

* Coersion of the **sticky** flag to boolean. According to step **8** in section **RegExpBuiltinExec  (21.2.5.2.2)**, implementations should use `ToBoolean`.
* **RegExp.prototype\[@@search\] (21.2.5.9)** with the **sticky** flag. For example, `assert.sameValue(/b/y[Symbol.search]('ab'), -1)` should pass.

test/built-ins/RegExp/prototype/sticky/this-regexp.js
```diff
@@ -0,0 +1,24 @@
...
+
+assert.sameValue(/./.sticky, false);
+assert.sameValue(/./i.sticky, false);
+assert.sameValue(/./g.sticky, false);
+assert.sameValue(/./gi.sticky, false);
+
+assert.sameValue(/./y.sticky, true);
+assert.sameValue(/./iy.sticky, true);
+assert.sameValue(/./yg.sticky, true);
+assert.sameValue(/./iyg.sticky, true);
```
goyakin: It would be better if we also have assertions when the RegExp object is created using `new RegExp`.

test/built-ins/RegExp/prototype/Symbol.match/get-sticky-err.js
```diff
@@ -0,0 +1,33 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Error thrown when accessing `sticky` property
```
goyakin: This makes it seem like the exception is **always** thrown. We should make it clear in the description that it is thrown when the accessor throws it.

test/built-ins/RegExp/prototype/Symbol.match/y-set-lastindex.js
jugglinmike: No problem

test/built-ins/RegExp/prototype/Symbol.match/y-init-lastindex.js
jugglinmike: I wasn't aware that we had that method!

test/built-ins/RegExp/prototype/Symbol.match/y-fail-return.js
jugglinmike: Sounds good

test/built-ins/RegExp/prototype/Symbol.replace/y-fail-return.js
jugglinmike: Agreed. Done

test/built-ins/RegExp/prototype/Symbol.match/get-sticky-err.js
jugglinmike: Couldn't hurt

test/built-ins/RegExp/prototype/sticky/this-regexp.js
jugglinmike: This would improve coverage, so it's hard to say "no", but the same would go for all the RegExp tests (and all the tests for prototype methods of primitive types). I don't think it makes sense to duplicate every assertion (one for "X when value created via literal expression" and one for "X when value created via constructor"), nor do I think that is what you are suggesting. There are probably cases where the construction approach might have some bearing on the behavior under test, but I think those should be clear before testing both forms (otherwise there will always be ambiguity as to whether such duplication is preferable).

jugglinmike:
Thanks for the input, @goyakin!

test/built-ins/RegExp/prototype/Symbol.match/get-sticky-coerce.js
```diff
@@ -0,0 +1,54 @@
...
+    [...]
+    8. Let sticky be ToBoolean(Get(R, "sticky")).
+features: [Symbol, Symbol.match]
+---*/
+
+var r = /a/;
+Object.defineProperty(r, 'sticky', { writable: true });
+
+r.sticky = undefined;
+assert.notSameValue(r[Symbol.match]('ba'), null);
```
goyakin: Shouldn't the ones that coerce to `false` result in `['a']`?

test/built-ins/RegExp/prototype/Symbol.replace/y-set-lastindex.js
```diff
@@ -0,0 +1,28 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Sets the `lastIndex` property to the index of the first match
```
goyakin: Like in the file for **Symbol.match**, it should say "end index" rather than "index". We should also update other files.

test/built-ins/RegExp/prototype/sticky/this-regexp.js
goyakin: Makes sense. Maybe we can check the constructor accepts the `y` flag, like it is done in `test/built-ins/RegExp/15.10.4.1-4.js`. What do you think?

test/built-ins/RegExp/prototype/flags/y-attr-err.js
```diff
@@ -0,0 +1,25 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: >
+    Error thrown when retrieving `sticky` attribute
```
goyakin: This should say "Behavior if error is thrown ..." to match the other files.

test/built-ins/RegExp/prototype/Symbol.replace/get-sticky-coerce.js
```diff
@@ -0,0 +1,53 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Error thrown when accessing `sticky` property
```
goyakin: This doesn't correctly reflect the intention of the test.

goyakin:
Thanks for the updates, @jugglinmike. I left inline comments for minor updates.

test/built-ins/RegExp/prototype/sticky/this-regexp.js
jugglinmike: Sounds good to me

jugglinmike:
Alright, comments updated

goyakin:
LGTM

bterlson:
Squash in preparation for pull.

jugglinmike:
All set

PR #355, opened 2015-07-08
B.1.4 and Unicode RegExps

anba:
Extension Annex B, "B.1.4 Regular Expressions Patterns" does not apply for Unicode RegExps.

### DIFF
--- a/test/built-ins/RegExp/unicode_restricted_identity_escape.js
+++ b/test/built-ins/RegExp/unicode_restricted_identity_escape.js
@@ -2,7 +2,7 @@
 // This code is governed by the BSD license found in the LICENSE file.
 
 /*---
-description: Extension B.1.4 is not applied for Unicode RegExp
+description: B.1.4 is not applied for Unicode RegExp - Identity escape with basic latin characters
 info: >
     The compatibility extensions defined in B.1.4 Regular Expressions Patterns
     are not applied for Unicode RegExps.
--- a/test/built-ins/RegExp/unicode_restricted_identity_escape.js
+++ b/test/built-ins/RegExp/unicode_restricted_identity_escape.js
@@ -1,0 +1,65 @@
+// Copyright (C) 2015 André Bargull. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Extension B.1.4 is not applied for Unicode RegExp
+info: >
+    The compatibility extensions defined in B.1.4 Regular Expressions Patterns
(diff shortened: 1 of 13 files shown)


test/built-ins/RegExp/unicode_restricted_identity_escape.js
```diff
@@ -0,0 +1,65 @@
...
+//
+// AtomEscape[U] :: CharacterEscape[?U]
+// CharacterEscape[U] :: IdentityEscape[?U]
+for (var cu = 0x00; cu <= 0x7f; ++cu) {
+  var s = String.fromCharCode(cu);
+  if (!isAlphaDigit(s) && !isSyntaxCharacter(s) && s !== "/") {
+    assert.throws(SyntaxError, function() {
+      RegExp("\\" + s, "u");
+    }, "Invalid IdentityEscape '\\" + s + "'");
+  }
```
jugglinmike: What do you think about also asserting accepted values?

```diff
   if (!isAlphaDigit(s) && !isSyntaxCharacter(s) && s !== "/") {
     assert.throws(SyntaxError, function() {
       RegExp("\\" + s, "u");
     }, "Invalid IdentityEscape '\\" + s + "'");
-  }
+  } else {
+    RegExp("\\" + s, "u");
+  }
}

anba: The current if-condition is quite permissive, for example it skips over `A` because a future ES edition might add the [`\A` assertion](http://perldoc.perl.org/perlre.html#Assertions).

I could add two more test files:
1. Valid IdentityEscape in Unicode RegExp (`\<SyntaxCharacter>` and `\/`)
2. Invalid IdentityEscape (`\<Alpha>` except for known escapes like `\B` etc., that means a new test for the forbidden RegExp extension rule described in 16.1)

anba:
Added two additional files:
- test valid IdentityEscape
- test IdentityEscape for letters not allowed (16.1 Forbidden Extensions)

goyakin:
@anba, could you add messages to assertions? Given that the test harness doesn't report the line number when there is a failure, it won't be obvious which assertion failed in a given file. This will also give us more information in the test execution log.

Moreover, I think we should differentiate the file descriptions. Almost all of them currently say:

    Extension B.1.4 is not applied for Unicode RegExp

Otherwise, the tests look good to me.

anba:
Thanks for reviewing! Added assertion messages and changed test descriptions.

goyakin:
Thanks @anba. Everything LGTM.

PR #356, opened 2015-07-08
Clean-up test comments and remove invalid tests

anba:
- 15.10.2.15-3-1 and 15.10.2.15-3-2 are both invalid (U+002F (SOLIDUS) instead of U+005C (REVERSE SOLIDUS) was used to start an escape sequence).
- 15.10.4.1-2 and 15.10.2.2-1 are identical tests, delete the latter.
- Update tests to avoid using runTestCase() function.
- Update multiple test infos to point to correct algorithm step (CharacterRange step 6 instead of step 1).


bterlson:
:heart:

PR #357, opened 2015-07-08
Intl402 ed2

anba:
Update tests for ECMA-402, 2nd. edition compliance.

### DIFF
--- a/harness/testIntl.js
+++ b/harness/testIntl.js
@@ -1148,6 +1148,43 @@ function testValidDateTimeComponentValue(component, value) {
 
 
 /**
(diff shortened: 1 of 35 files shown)


kunalspathak:
LGTM.

bterlson:
@anba looks good to me too. Thanks!!

anba:
Cool! I was just about saying my ES6^H2015 test implementation now passes all but one test (which fails because of an ICU4J bug) :smile:, but then I realized recent commits introduced a couple of regressions (in test262 itself), probably related to 10e0d977ecc402d02571b4b42bf550281f297872 :cry: .

PR #358, opened 2015-07-08
Extend test coverage for Promise built-in

jugglinmike:
I've omitted test for the name of [get Promise [ @@species ]](http://www.ecma-international.org/ecma-262/6.0/#sec-get-promise-@@species) since @leobalter has that covered in gh-353.


bterlson:
Looks great, pulling it in. This also found a few bugs in Chakra so good work :-D

jugglinmike:
Awesome!

Issue #196, opened 2015-03-13
testing ECMAScript parsers


jmdyck:
I'm wondering if it would be possible to increase test262's usefulness for testing ECMAScript parsers. Presumably, every positive test in the suite is a positive test for a parser, but not so for negative tests. I'm guessing that the negative parser tests are (pretty much?) a subset of the tests marked `negative: SyntaxError`, but I don't think there's a way to auto-detect which ones.

(According to https://github.com/tc39/test262/blob/master/CONTRIBUTING.md :

```
To assert that an error is thrown during lexing or parsing, before
any lines of JavaScript are executed, use the following pattern:

/*
 * @negative ^((?!NotEarlyError).)*$
 */

throw NotEarlyError;
var var = var;
```

But note that (a) this uses the old header syntax, and (b) no "negative:" field in the suite currently uses that pattern, so I suspect it's stale advice.)

What I had in mind is a new field for the file-header (say, "parser-negative:") to designate which tests are negative for parsers. The value for the field could be an indication of roughly at what 'level' the error occurs (lexical, syntactic, early error, others?) to accommodate parsers of different "depths". E.g., when testing a parser that checks for early errors, you'd expect tests matching

```
 /parser-negative: (lexical|syntactic|early error)/
```

to cause an error, but when testing a parser that doesn't, you'd only expect

```
 /parser-negative: (lexical|syntactic)/
```

tests to cause errors. (I realize that not all parsers would conform to such a categorization of errors, but at least it would be a better starting point than now.)

(I suppose, if tests had references back to the pertinent section(s) of the spec, that might supply the necessary info, e.g. a pointer to a Syntax section vs an Early Errors section. But I haven't heard any discussion of adding such references lately, so I'm doubtful it'll happen any time soon.)

jugglinmike:
I think this is a great idea, and I'd like to help make it happen! I don't
quite understand the categorization of errors: "lexical", "syntactic", and
"early error", though. My intuition would be to allow for two types:
`ReferenceError` and `SyntaxError`, so I'd like to know more about the
distinction you're proposing.

jmdyck:
(The 'negative' field has values 'SyntaxError' and 'ReferenceError', so presumably already handles your intuition.)

The thing is, a SyntaxError can arise in various different ways:
  1. during tokenization;
  2. during parsing; or
  3. when checking for early errors.

Moreover, these can happen "statically" (for the top-level Script or Module) or 4. "dynamically" (for a call to eval() or Function()).

[...]

PR #360, opened 2015-07-08
Negative reform

jugglinmike:
Re-factor tests for runtime errors to express their expectations using
`assert.throws`. This change is in service of gh-196 as it almost completely
limits the use of the `negative` frontmatter to only those tests that fail
prior to code execution.

Some tests explicitly concern the generation of errors in global code--these
cannot be refactored use use the `assert.throws` helper method without
fundamentally changing the behavior under test:

- test/language/statements/const/global-use-before-initialization-in-declaration-statement.js
- test/language/statements/const/global-use-before-initialization-in-prior-statement.js
- test/language/statements/let/global-use-before-initialization-in-declaration-statement.js
- test/language/statements/let/global-use-before-initialization-in-prior-statement.js

I think we can address this with finer-grained control over `negative`, but
that extension can be designed and implemented separately.

It probably makes sense to squash this into a single commit prior to landing,
but I left the history as-is in case seeing the changes broken down by error
type aids the review process.

### DIFF
--- a/test/language/line-terminators/S7.3_A3.4_T1.js
+++ b/test/language/line-terminators/S7.3_A3.4_T1.js
@@ -5,8 +5,7 @@
 info: Single line comments can not contain PARAGRAPH SEPARATOR (U+2029) inside
 es5id: 7.3_A3.4_T1
 description: Insert PARAGRAPH SEPARATOR (\u2029) into single line comment
-negative: ReferenceError
+negative: SyntaxError
 ---*/
 
-// CHECK#1
-eval("// single line \u2029 comment");
+// single line PS> ??? (invalid)
--- a/test/built-ins/Array/prototype/splice/S15.4.4.12_A6.1_T2.js
+++ b/test/built-ins/Array/prototype/splice/S15.4.4.12_A6.1_T2.js
@@ -5,7 +5,6 @@
 info: Array.prototype.splice sets `length` on `this`
 es5id: 15.4.4.12_A6.1_T2
 description: Array.prototype.splice throws if `length` is read-only
-negative: TypeError
 ---*/
 
 var a = [0, 1, 2];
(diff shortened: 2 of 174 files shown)


test/language/line-terminators/S7.3_A3.4_T1.js
```diff
@@ -5,8 +5,7 @@
...
 es5id: 7.3_A3.4_T1
 description: Insert PARAGRAPH SEPARATOR (\u2029) into single line comment
-negative: ReferenceError
+negative: SyntaxError
 ---*/
 
-// CHECK#1
-eval("// single line \u2029 comment");
+// single line PS>
??? (invalid)
```
leobalter: It's not easy to tell this line contains a PS character, but still looks better than using eval.

Also, it seems the expected error should be the ReferenceError, which you can catch:

```js
try { // single line PS>  (invalid)
} catch(e) { console.log(e) }
```

I just removed the ??? characters, responsible for the SyntaxError. The ReferenceError occours because invalid is not defined, as the PS breaks the line.

Maybe:

```js
// single line PS>  throw new Test262Error()
```

leobalter:
Just a single comment with a suggestion. Everything else LGTM.

test/language/line-terminators/S7.3_A3.4_T1.js
jugglinmike: > It's not easy to tell this line contains a PS character, but still looks better than using eval.

Yup. Refactoring like this isn't going to be pretty, but `eval` forces it to be a runtime error, which I'd rather avoid.

> Also, it seems the expected error should be the ReferenceError, which you can catch:

Since the behavior under test is a consequence of the grammar, I'd much rather express the test in a way that does not rely on runtime semantics.

bterlson:
This is great stuff! I guess we should open an issue for what to do with tests that throw runtime errors from global scope (where they can't be handled).

Feel free to squash, I will pull.

jugglinmike:
Thanks, @bterlson! All squashed

PR #361, opened 2015-07-09
Add tests for function `name` attribute

jugglinmike:
This changeset includes tests for most invocations of the SetFunctionName
abstract operation in the ES2015 specification. Practical testing
considerations preclude the introduction of tests for certain invocations:

- The project is still vetting methods to sustainably test the semantics of the
  Destructuring Binding pattern across all valid productions.
  - 13.3.3.6 Runtime Semantics: IteratorBindingInitialization
  - 13.3.3.7 Runtime Semantics: KeyedBindingInitialization
- Without a loader, there is no way to access a function object declared in an
  ExportDeclaration, so `name` assignment cannot be tested in these
  cases
  - 14.1.19 Runtime Semantics: InstantiateFunctionObject
  - 14.4.12 Runtime Semantics: InstantiateFunctionObject
  - 14.5.15 Runtime Semantics: BindingClassDeclarationEvaluation
  - 15.2.3.11 Runtime Semantics: Evaluation


anba:
The PR looks good to me. Possible additional tests from https://github.com/anba/es6draft/tree/master/src/test/scripts/suite/semantic/function/name :

- `class` with `static get name() { ... }` property and also with `name` as (non-trivial) computed property name to make sure the `HasOwnProperty(rval, " name" )` checks from the spec are performed at runtime
- `class` with explicit `constructor() {}` property to test `(f = class { constructor() {} }).name === "f"` and not `(f = class { constructor() {} }).name === "constructor"`
- (static) method definitions in classes
- chained calls to `bind()`, e.g. `(function f(){}).bind().bind().name === "bound bound f"`

jugglinmike:
Good ideas all around, @anba. I've added the tests you suggested (and fixed the tests for MethodDefinitions in object initializers, too).

bterlson:
Squash it up! :eggplant:

jugglinmike:
I'm having trouble confirming that an eggplant is actually a squash. Anyway, this is all one commit now.

bterlson:
Thanks!

PR #362, opened 2015-07-09
Re-organize generator tests

jugglinmike:
The `test/language/generators/` directory contained a single file that
concerned generators derived both from syntactic form and from a
built-in function.

Refactor this test into two files and place each in the appropriate
directory.


ianwjhalliday:
:+1: LGTM

PR #365, opened 2015-07-10
Add tests for String.prototype.codePointAt

leobalter:


test/built-ins/String/prototype/codePointAt/position-equal-or-more-than-array-size-returns-undefined.js
```diff
@@ -0,0 +1,20 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 21.1.3.3
+description: >
+  If pos >= size, return undefined
```
jugglinmike: This test seems redundant given `position-equal-or-more-than-size-returns-undefined.js`. I appreciate your interest in covering the Array case, but since there is no explicit "array branch" concerning the logic under test, it's probably better to leave this one out. I suppose we could keep this and instead duplicate all the other tests for the array case, but that seems excessive.

test/built-ins/String/prototype/codePointAt/position-to-integer-less-than-zero-returns-undefined.js
```diff
@@ -0,0 +1,20 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 21.1.3.3
+description: >
+  If pos < 0, return undefined
```
jugglinmike: This file (along with `position-to-integer-value-equal-or-more-than-size-returns-undefined.js`) tests two different steps (4 and 7) at once. It's probably better to keep the tests directed: one test for out-of-bounds numbers, one test for in-bounds numbers, and a final for string-to-number conversion. Admittedly, this will mean we won't have an explicit test for a negative number specified as a string, but that seems like a class of edge case that I don't think Test262 can hope to protect against.

test/built-ins/String/prototype/codePointAt/return-abrupt-from-symbol-pos-to-integer.js
```diff
@@ -0,0 +1,20 @@
...
+  1. Let O be RequireObjectCoercible(this value).
+  2. Let S be ToString(O).
+  3. ReturnIfAbrupt(S).
+  4. Let position be ToInteger(pos).
+  5. ReturnIfAbrupt(position).
+features: [Symbol]
+---*/
+
+assert.throws(TypeError, function() {
+  'abc'.codePointAt(Symbol(1));
```
jugglinmike: In order to avoid false positives and to improve readability, it's generally best to limit the code within the callback to `assert.throws` as much as possible. In this case, I'm thinking that the test could pass if the implementation incorrectly threw a TypeError from `Symbol(1)` or `Symbol()` in the case of `return-abrupt-from-this-as-symbol.js` (maybe because the description is not a string? I dunno) It's definitely unlikely (and not something we can protect against in all cases), but it can't hurt to move the constructor invocation outside as `var s = new Symbol(1);`. It also makes the test slightly more readable, because there's less ambiguity as to the expression that is expected to generate the error.

test/built-ins/String/prototype/codePointAt/return-first-code-unit.js
```diff
@@ -0,0 +1,36 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 21.1.3.3
+description: >
+  Return first code unit value of the element at index position if first >=
+  0xD800 or first <= 0xDBFF and second < 0xDC00 or second > 0xDFFF.
```
jugglinmike: This test is good, but it took me a little while to figure out what is going on here. A better name/description would say something surrogate pairs--maybe "Returns code point of LeadSurrogate if not followed by a valid TrailSurrogate."

jugglinmike:
I left a few suggestions for minor changes, but other than that: looking good!

test/built-ins/String/prototype/codePointAt/position-equal-or-more-than-array-size-returns-undefined.js
leobalter: I'll leave this out and add a note to make a whole array case on this, as it's good to first prioritize the other string methods.

bterlson:
I think this is good after a squash.

leobalter:
squashed and rebased, thanks!

PR #367, opened 2015-07-14
Add tests for String.prototype.normalize

leobalter:


jugglinmike:
Another nice batch of tests.

@bterlson Leo's approach to Unicode's Annex 15 (and the one I'm leaning
towards) is nominal. A "complete" approach would require using
`String.prototype.normalize` to test the entirety of that specification. I
thought maybe this question could be answered by the approach taken in existing
tests, but it seems to be a novel issue (other spec references in the
bibliography are non-normative). Do you have any thoughts about a general
policy for testing externally-referenced specifications?

test/built-ins/String/prototype/normalize/return-abrupt-from-form-as-symbol.js
```diff
@@ -0,0 +1,21 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 21.1.3.12
+description: >
+  Returns abrupt from ToString(this) as a Symbol.
```
goyakin: Shouldn't `ToString(this)` be `ToString(form)`?

test/built-ins/String/prototype/normalize/return-normalized-string-using-default-parameter.js
```diff
@@ -0,0 +1,22 @@
...
+  8. Let ns be the String value that is the result of normalizing S into the
+  normalization form named by f as specified in
+  http://www.unicode.org/reports/tr15/tr15-29.html.
+  9. Return ns.
+---*/
+
+var s = '\u1E9B\u0323';
+
+assert.sameValue(s.normalize(), '\u1E9B\u0323');
+assert.sameValue(s.normalize(undefined), '\u1E9B\u0323');
```
goyakin: This file seems to be missing the newline at EOF.

leobalter:
@goyakin, thanks! I fixed them.

test/built-ins/String/prototype/normalize/return-abrupt-from-form.js
```diff
@@ -0,0 +1,24 @@
...
+  6. ReturnIfAbrupt(f).
+---*/
+
+var o = {
+  toString: function() {
+    throw new Test262Error();
+  }
+};
+
+assert.throws(Test262Error, function() {
```
goyakin: I think it would be better if we add a test checking successful coercion to String to complement this.

test/built-ins/String/prototype/normalize/return-abrupt-from-form.js
```diff
@@ -0,0 +1,24 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 21.1.3.12
+description: >
+  Returns abrupt from ToString(form)
```
goyakin: @leobalter, what do you think about merging `form` coercion failure tests into a single file?

test/built-ins/String/prototype/normalize/return-abrupt-from-this.js
```diff
@@ -0,0 +1,23 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 21.1.3.12
+description: >
+  Returns abrupt from ToString(this)
```
goyakin: Like `form` String coercion tests, maybe we can combine these into a single file?

test/built-ins/String/prototype/normalize/this-is-null-throws.js
```diff
@@ -0,0 +1,16 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 21.1.3.12
+description: >
+  Throws TypeError when `this` is null
```
goyakin: Shall we merge Object coercion failure tests into a single file, too?

goyakin:
Thank @leobalter. I left inline comments about possibly merging related files, but otherwise, LGTM.

test/built-ins/String/prototype/normalize/return-abrupt-from-form.js
leobalter: this is a decision I've made from a previous PR discussion.

I avoid merging tests where I can isolate new ES6 features, as the ones using Symbol.

leobalter: I'm ok to merge, but that should affect the other open PRs and we could also track the others already merged.

test/built-ins/String/prototype/normalize/return-abrupt-from-form.js
leobalter: at b0c6084

I limited the only two valid forms as it aims only for the ToString coercion there.

leobalter:
@goyakin, ty! 

The remaining comments needs to wait for https://github.com/tc39/test262/pull/367#discussion_r35116918 to be addressed.

test/built-ins/String/prototype/normalize/return-abrupt-from-form.js
goyakin: I see. It should be OK to keep the files separate then.

leobalter:
Comments solved. I'm ready to rebase and squash unless we get more changes to apply.

bterlson:
Squash away!

test/built-ins/String/prototype/normalize/return-normalized-string-from-coerced-form.js
```diff
@@ -0,0 +1,38 @@
...
+---*/
+
+var s = '\u00C5\u2ADC\u0958\u2126\u0344';
+var nfc = '\xC5\u2ADD\u0338\u0915\u093C\u03A9\u0308\u0301';
+var nfd = 'A\u030A\u2ADD\u0338\u0915\u093C\u03A9\u0308\u0301';
+var o = {
+  toString: function() { return 'NFC'; }
+};
+
+assert.sameValue(s.normalize(undefined), nfc, 'form is undefined');
```
goyakin: This is duplicate. We have this test case in `return-normalized-string-using-default-parameter.js`, so we can just rely on that.

leobalter: good catch! That leaded me to another approach: I removed the undefined parameter test from `eturn-normalized-string-using-default-parameter.js` and left this one, as I'm asserting the coercion.

goyakin: `undefined` isn't coerced though. In **step 4** of the method, `form` is set to `"NFC"` before coercion it is `undefined`.

leobalter: my bad.

leobalter: fixed.

goyakin:
Thanks @leobalter. Everything LGTM.

PR #369, opened 2015-07-15
Reinforce test for the use of ToLength(lastIndex) in RegExp

littledan:
Previously, test262 had only a test that ensured that ToLength (for
example, rather than ToInteger) was used in test'ing a RegExp, not
in calls to exec. Although in the ES5 and ES2015 specs there is only
one code path, in some implementations, ToLength is called in from
separate code paths. This patch makes a new test for exec'ing a
RegExp and ensures that ToLength is called.

### DIFF
--- a/test/built-ins/RegExp/prototype/test/S15.10.6.3_A1_T22.js
+++ b/test/built-ins/RegExp/prototype/test/S15.10.6.3_A1_T22.js
@@ -1,0 +1,38 @@
+// Copyright 2009 the Sputnik authors.  All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+info: >
+    RegExp.prototype.test behavior depends on the lastIndex property:
+    ToLength(lastIndex) is the starting point for the search, so
+    negative numbers result in searching from 0.
+es5id: 15.10.6.3_A1_T22
+description: "Set lastIndex to -1 and call /(?:ab|cd)\\d?/g.test(\"aacd22 \")"
+---*/
+
+var __re = /(?:ab|cd)\d?/g;
+__re.lastIndex=-1;
+var __executed = __re.test("aacd22 ");
+
+//CHECK#1
+if (!__executed) {
+	$ERROR('#1: __re = /(?:ab|cd)\\d?/g; __re.lastIndex=-1; __executed = __re.test("aacd22 "); __executed === true');
+}
+
+//CHECK#2
+if (__re.lastIndex !== 5) {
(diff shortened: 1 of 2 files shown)


bterlson:
Can you update the copyright on the new file? (Presumably not copyright Sputnik authors!)

test/built-ins/RegExp/prototype/test/S15.10.6.3_A1_T22.js
```diff
@@ -0,0 +1,38 @@
...
+description: "Set lastIndex to -1 and call /(?:ab|cd)\\d?/g.test(\"aacd22 \")"
+---*/
+
+var __re = /(?:ab|cd)\d?/g;
+__re.lastIndex=-1;
+var __executed = __re.test("aacd22 ");
+
+//CHECK#1
+if (!__executed) {
+	$ERROR('#1: __re = /(?:ab|cd)\\d?/g; __re.lastIndex=-1; __executed = __re.test("aacd22 "); __executed === true');
```
leobalter: how about using the assert methods instead of `$ERROR`?

leobalter:
this is also good to merge on the es5-tests branch.

test/built-ins/RegExp/prototype/test/S15.10.6.3_A1_T22.js
bterlson: I'm not a stickler if it's updating an old tests (although this one does add a new one in the old style). Can't hurt though.

bterlson:
I think these tests will fail on es5 implementations since ToLength was added in ES6?

littledan:
Right, they should fail, since it was calling ToUint32. I had to mark the related test in the ES5 branch as a known failure in V8's test262 rig.

littledan:
Fixed the copyright statement

bterlson:
Did you push the new commits? Not seeing them.

bterlson:
@littledan in case you didn't get the notification :)

littledan:
@bterlson I think I did push the commit. At https://github.com/littledan/test262/commit/9d2451eb7247d6e78653973f2f388bb0602f492c I see the new copyright statement.

bterlson:
@littledan, I think the test file you modified should retain its original copyright. The new test file should not be copyright the Sputnik authors, though.

littledan:
@bterlson Git saw the change in a funny way, but in actuality, I just moved the Sputnik file to another location, and made a new file in the old location. The copyright statements reflect that.

bterlson:
Ahh ok. Sorry for the confusion!

PR #370, opened 2015-07-15
Add tests for well-known Symbol, @@hasInstance

jugglinmike:
Assert the behavior of the built-in method on the FunctionPrototype and
the usage of the Symbol by the `instanceof` operator.


bterlson:
This looks good. :eggplant:

jugglinmike:
All set!

PR #371, opened 2015-07-15
Add tests for @@isConcatSpreadable

jugglinmike:
`test/built-ins/Array/prototype/concat/is-concat-spreadable-val-undefined.js` is prone to false positives, but I want to make sure that implementers don't simply coerce any "own" property.

### DIFF
--- a/test/built-ins/Array/prototype/concat/is-concat-spreadable-val-falsey.js
+++ b/test/built-ins/Array/prototype/concat/is-concat-spreadable-val-falsey.js
@@ -1,0 +1,55 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 22.1.3.1
+description: >
+    The `Symbol.isConcatSpreadable` property is defined and coerces to `false`
+info: >
+    1. Let O be ToObject(this value).
+    2. ReturnIfAbrupt(O).
+    3. Let A be ArraySpeciesCreate(O, 0).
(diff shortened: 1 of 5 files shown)


test/built-ins/Array/prototype/concat/is-concat-spreadable-val-falsey.js
```diff
@@ -0,0 +1,55 @@
...
+    ES6 22.1.3.1.1: Runtime Semantics: IsConcatSpreadable ( O )
+
+    1. If Type(O) is not Object, return false.
+    2. Let spreadable be Get(O, @@isConcatSpreadable).
+    3. ReturnIfAbrupt(spreadable).
+    4. If spreadable is not undefined, return ToBoolean(spreadable).
+features: [Symbol.isConcatSpreadable]
+---*/
+
+var item = [1];
```
farzonl: could you change this to var item = [1,2];

First let me state that this test case is correct, b\c of the array ref compares, but I think it needs to be more explicitly clearing what is going on here. if the array you are adding/concating has length 2 and the length of the concated array is length 1 it is more abundantly clear that a spread did not happen.

farzonl:
Everything else LGTM

test/built-ins/Array/prototype/concat/is-concat-spreadable-val-falsey.js
jugglinmike: Happy to oblige. Also: welcome to Test262!

bterlson:
:eggplant: whenever!

jugglinmike:
All set!

PR #374, opened 2015-07-15
Fix 11.13.1-4-29gs.js incorrect comment explanation (master)

### LINKED ISSUE #372: Incorrect rationale for test 11.13.1-4-29gs.js
Located at test/suite/ch11/11.13/11.13.1/ in es5-tests and test/language/expressions/assignment/ on master.

Test comment claims that `Math.PI = 20;` should throw a SyntaxError in strict mode. It should actually throw a TypeError.

Patches coming.

edgemaster:
SyntaxError => TypeError for writing to a non-writeable field.

Fixes #372 for master

### DIFF
--- a/test/language/expressions/assignment/11.13.1-4-29gs.js
+++ b/test/language/expressions/assignment/11.13.1-4-29gs.js
@@ -7,7 +7,7 @@
 /*---
 es5id: 11.13.1-4-29gs
 description: >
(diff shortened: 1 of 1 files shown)


bterlson:
The test is strange in that it will pass for any error thrown I guess? Should probably use assert.throws, but updating the docs is a good first step. Thanks!

PR #376, opened 2015-07-16
update tests for Array.from

yanlxu:


test/built-ins/Array/from/Array.from_arity.js
```diff
@@ -7,6 +7,19 @@ description: >
     The Array.from() method creates a new Array instance
     from an array-like or iterable object.
     https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from
+info: >
```
jugglinmike: I know you didn't author the description, but since you are changing the frontmatter, would you mind also shortening this description to one line, [as recommended in the project's contribution guidelines](https://github.com/tc39/test262/blob/d62f3a82b97dba22a0be2798fbe34548e96a439f/CONTRIBUTING.md#description)?

test/built-ins/Array/from/argument-passed-null.js
```diff
@@ -0,0 +1,13 @@
+// Copyright 2015 Microsoft Corporation. All rights reserved.
+// This code is governed by the license found in the LICENSE file.
+
+/*---
+description: this argument is passed as null
+es6id: 22.1.2.1
+---*/
+
+function mapFn(value, index) {
+    assert("mapFn is not expected to be called. Called for " + value + " and index " + index);
```
jugglinmike: The idiomatic way to express an unreachable code path is to use [the `$ERROR` functon](https://github.com/tc39/test262/blob/d62f3a82b97dba22a0be2798fbe34548e96a439f/CONTRIBUTING.md#test-environment). This will make the test source and the produced error easier to understand.

test/built-ins/Array/from/argument-passed-null.js
```diff
@@ -0,0 +1,13 @@
...
+/*---
+description: this argument is passed as null
+es6id: 22.1.2.1
+---*/
+
+function mapFn(value, index) {
+    assert("mapFn is not expected to be called. Called for " + value + " and index " + index);
+}
+
+assert.throws(TypeError, function(){Array.from.call(null);});
```
jugglinmike: It looks like you have forgotten to specify `mapFn` here.

test/built-ins/Array/from/calling-from-valid-2.js
```diff
@@ -0,0 +1,25 @@
...
+
+/*---
+description: Calling from with a valid map function with thisArg
+es6id: 22.1.2.1
+---*/
+
+var array = [ 0, 1, -2, 4, -8, 16, -32, 64, -128, 256, -512, 1024 ];
+var arrayIndex = -1;
+var obj = {
+    mapFn (value, k) {
```
jugglinmike: Given that we expect implementors to be transitioning from ES5 to ES6, it's generally best to limit the number of new language features to just those that are necessary to express the behavior under test. In this case, implementations that do not support MethodDeclarations won't be able to use this test. Could you define `mapFn` using a FunctionDeclaration instead? This would also be more consistent with the related test in this patch, `calling-from-valid-1.js`

test/built-ins/Array/from/create-typedarray-from.js
```diff
@@ -0,0 +1,13 @@
+// Copyright 2015 Microsoft Corporation. All rights reserved.
+// This code is governed by the license found in the LICENSE file.
+
+/*---
+description: Create a TypedArray from another TypedArray
```
jugglinmike: This test is identical to `test/built-ins/Array/from/create-array-from.js`

test/built-ins/Array/from/calling-from-valid-1.js
```diff
@@ -0,0 +1,24 @@
...
+// This code is governed by the license found in the LICENSE file.
+
+/*---
+description: Calling from with a valid map function without thisArg
+es6id: 22.1.2.1
+---*/
+
+var array = [ 0, 1, -2, 4, -8, 16, -32, 64, -128, 256, -512, 1024 ];
+var arrayIndex = -1;
+var globalThis = this;
```
jugglinmike: The approach preferred in recent tests involving callback-supporting APIs is to include a dedicated test for strict mode (using the `onlyStrict` [frontmatter flag](https://github.com/tc39/test262/blob/d62f3a82b97dba22a0be2798fbe34548e96a439f/CONTRIBUTING.md#flags)) and a separate test for non-strict mode (using the `noStrict` frontmatter flag). Each of these tests can be more explicit about expectations for the `thisValue`. This protects against false positives where the value of `this` may be incorrect in the global scope.

test/built-ins/Array/from/mapfn-attached-typedarray.js
```diff
@@ -0,0 +1,25 @@
+// Copyright 2015 Microsoft Corporation. All rights reserved.
+// This code is governed by the license found in the LICENSE file.
+
+/*---
+description: mapFn is attached to the TypedArray itself
```
jugglinmike: I don't understand this test. The file name and description reference TypedArray, but that intrinsic is not used in the test body. Also, the modification of the ArrayPrototype built-in doesn't seem necessary.

test/built-ins/Array/from/mapfn-returns-integer.js
```diff
@@ -0,0 +1,30 @@
...
+
+/*---
+description: mapFn returns a non-integer
+es6id: 22.1.2.1
+---*/
+
+var array = [ 2, 4, 8, 16, 32, 64, 128 ];
+var mapObj = {
+    arrayIndex : -1,
+    mapFn(value, index) {
```
jugglinmike: Here also: the MethodDeclaration is not necessary for this test

test/built-ins/Array/from/mapfn-returns-integer.js
```diff
@@ -0,0 +1,30 @@
+// Copyright 2015 Microsoft Corporation. All rights reserved.
+// This code is governed by the license found in the LICENSE file.
+
+/*---
+description: mapFn returns a non-integer
```
jugglinmike: The filename (`mapfn-returns-integer.js`) does not match the description ("mapFn returns a non-integer"). Since the test body clearly concerns the behavior when an object is returned, probably it's just the filename that needs changing.

test/built-ins/Array/from/mapfn-returns-integer.js
```diff
@@ -0,0 +1,30 @@
+// Copyright 2015 Microsoft Corporation. All rights reserved.
+// This code is governed by the license found in the LICENSE file.
+
+/*---
+description: mapFn returns a non-integer
+es6id: 22.1.2.1
+---*/
+
+var array = [ 2, 4, 8, 16, 32, 64, 128 ];
+var mapObj = {
```
jugglinmike: The use of `mapObject` here does not seem necessary. Instead, you could just specify a `mapFn` declaration and a variable for `arrayIndex`. This would make the test more direct, easier to read, and easier to debug.

test/built-ins/Array/from/Array.from-name.js
```diff
@@ -0,0 +1,30 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
```
leobalter: the contributor name doesn't match with the other files.

test/built-ins/Array/from/mapfn-throws-exception.js
```diff
@@ -0,0 +1,25 @@
...
+// This code is governed by the license found in the LICENSE file.
+
+/*---
+description: mapFn throws an exception
+es6id: 22.1.2.1
+---*/
+
+var array = [ 2, 4, 8, 16, 32, 64, 128 ];
+function mapFn(value, index, obj) {
+    throw "Expected error";
```
jugglinmike: If you use `throw new Test262Error();` here, then you can express this test's assertion more simply:

```js
  Array.from(array, mapFn);
});
```

Note that there is no need to test the return value in this case, since if the CallExpression throws an exception, we can take it for granted that no assignment would follow.

test/built-ins/Array/from/mapfn-throws-exception.js
```diff
@@ -0,0 +1,25 @@
...
+
+    return 127;
```
jugglinmike: This statement is unreachable

test/built-ins/Array/from/source-object-length.js
```diff
@@ -0,0 +1,25 @@
...
+var obj = {
+    length : 4,
+    0 : 2,
+    1 : 4,
+    2 : 0,
+    3 : 16
+}
+delete obj[2];
+var a = Array.from.call(Object, obj);
+assert.sameValue(typeof(a), "object", "The returned type is expected to be object.");
```
jugglinmike: Based on the invocation of `Array.from`, this appears to be a test for the use of a constructor other than `Array`. That's an important feature to test, but it should be done in a dedicated file with a name/frontmatter that describe the feature specifically.

Also, the expectation should be expressed in a way that would fail if the Array constructor were used incorrectly (note that `typeof Array.from([])` is *also* "object", so it would be better to check `Array.from.call(Object, []).constructor`)

By the way: since `typeof` is an operator, you do not need to wrap the operand in parenthesis.

test/built-ins/Array/from/source-object-length.js
```diff
@@ -0,0 +1,25 @@
...
+
+var array = [2, 4, 0, 16];
+var expectedArray = [2, 4, , 16];
+var obj = {
+    length : 4,
+    0 : 2,
+    1 : 4,
+    2 : 0,
+    3 : 16
+}
```
jugglinmike: Missing semicolon here

test/built-ins/Array/from/source-object-missing.js
```diff
@@ -0,0 +1,21 @@
...
+es6id: 22.1.2.1
+---*/
+
+var array = [2, 4, , 16];
+var obj = {
+    length: 4,
+    0: 2,
+    1: 4,
+    3: 16
+}
```
jugglinmike: Missing semicolon here

test/built-ins/Array/from/source-array-boundary.js
```diff
@@ -0,0 +1,32 @@
...
+---*/
+
+var array = [ Number.MAX_VALUE, Number.MIN_VALUE, Number.NaN, Number.NEGATIVE_INFINITY, Number.POSITIVE_INFINITY ];
+var mapObj = {
+    arrayIndex : -1,
+    mapFn(value, index) {
+        this.arrayIndex++;
+        if (index !== 2) {
+            assert.sameValue(value, array[this.arrayIndex], "Value mismatch in mapFn at index " + index + ".");
+        } else {
```
jugglinmike: This branch is unnecessary. The assertion in the `if` block can be made for every iteration, so the `if..else` statement can be replaced with the body of the `if` block.

jugglinmike:
@yanlxu Many of these tests assign to an undeclared variable `j`, which means they will always fail in strict mode. This mistake would be reported by all the major implementations. If you would like help learning to test your patches, then I would be happy to do so. We might even be able to update the contribution guidelines with some information about this!

yanlxu:
have made changes to address PR review comments.

bterlson:
Thanks @yanlxu!

PR #378, opened 2015-07-16
Add tests for Array.of

leobalter:
I also updated the old tests.


test/built-ins/Array/of/construct-this-with-the-number-of-arguments.js
```diff
@@ -0,0 +1,36 @@
...
+  length = len;
+  hits++;
+}
+
+Array.of.call(C);
+assert.sameValue(length, 0, '`Array.of.call(C);` called `new C(0)`');
+assert.sameValue(hits, 1, 'Called constructor once');
+
+Array.of.call(C, 'a', 'b')
+assert.sameValue(length, 2, '`Array.of.call(C, "a", "b"));` called `new C(2)`');
```
jugglinmike: Might as well check `hits` after the second and third invocation, too.

test/built-ins/Array/of/creates-a-new-array-from-arguments.js
```diff
@@ -0,0 +1,42 @@
...
+    e. Increase k by 1.
+  9. Let setStatus be Set(A, "length", len, true).
+  10. ReturnIfAbrupt(setStatus).
+  11. Return A.
+includes: [compareArray.js]
+---*/
+
+var a1 = Array.of('Mike', 'Rick', 'Leo');
+assert(
+  compareArray(a1, ['Mike', 'Rick', 'Leo']),
```
jugglinmike: Between the two different approaches this test uses to compare arrays, I prefer. the second. It doesn't rely on an external function or iteration semantics. It's more characters to type, but we don't expect this test to change any time soon :) Either approach is technically correct, so I think the most important thing is to consistently use one or the other.

test/built-ins/Array/of/does-not-use-prototype-properties.js
```diff
@@ -0,0 +1,35 @@
...
+description: Array.of does not use prototype properties for arguments.
+info: >
+  It defines elements rather than assigning to them.
+---*/
+
+Object.defineProperty(Array.prototype, "0", {
+  set: function(v) {
+    $ERROR('Should define own properties');
+  },
+  get: function(v) {
```
jugglinmike: Here and below: no need to specify a parameter for the `get` method

test/built-ins/Array/of/return-a-new-array-object.js
```diff
@@ -0,0 +1,40 @@
...
+result = Array.of.call(undefined);
+assert(
+  result instanceof Array,
+  'this is not a constructor'
+);
+
+result = Array.of.call(Math.cos);
+assert(
+  result instanceof Array,
+  'this is a builtin functions with no [[Construct]] slot'
```
jugglinmike: Here and below: "a builtin functions" -> "a builtin function"

jugglinmike:
I just have a few minor recommendations. Nice work!

leobalter:
Done. Thanks, Mike!

ianwjhalliday:
LGTM

bterlson:
:eggplant:

leobalter:
done, thanks!

PR #387, opened 2015-07-22
Add tests for default parameters

leobalter:


leobalter:
Did some changes on fixup commits. Referenced the [13.3.3.7](http://www.ecma-international.org/ecma-262/6.0/#sec-runtime-semantics-keyedbindinginitialization) on general cases and added the test for properties values.

test/language/default-parameters/call-define-values.js
```diff
@@ -0,0 +1,35 @@
...
+  return [a, b, c, d];
+}
+
+results = fn();
+
+assert.sameValue(results[0], 1, 'apply default values #1');
+assert.sameValue(results[1], null, 'apply default values #2');
+assert.sameValue(results[2], o, 'apply default values #3');
+
+// Ref: https://bugzilla.mozilla.org/show_bug.cgi?id=777060
```
jugglinmike: I've referenced Mozilla's BugZilla tracker for many personal projects, but
    because these are the canonical ECMAScript tests, I think we should derive
    all justification from the spec itself. You were right about the spec being
    tricky here; I think the justification goes something like this:

    > 24. If hasDuplicates is true, then
    >     [...]
    > 25. Else,
    >     a. Let formalStatus be IteratorBindingInitialization for formals with
    >        iteratorRecord and env as arguments.

    > ES6 13.3.3.6 Runtime Semantics: IteratorBindingInitialization
    >
    > SingleNameBinding : BindingIdentifier Initializeropt
    >
    > [...]
    > 6. If Initializer is present and v is undefined, then
    >    [...]

Seeing that justification makes an improvement more obvious: this test is
    good because it ensures that the default value is not used when a
    non-undefined positional argument is specified. It would be even better if
    it ensured that the initializer is not evaluated in those cases.

test/language/default-parameters/call-define-values.js
```diff
@@ -0,0 +1,35 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 13.3.3.7
+description: >
+  Function call define default values to arguments
+info: >
+  13.3.3.7 Runtime Semantics: KeyedBindingInitialization
+
+  When undefined is passed for environment it indicates that a PutValue
```
jugglinmike: This text is irrelevant to the behavior under test.

test/language/default-parameters/call-returns-reference-error.js
```diff
@@ -0,0 +1,15 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 13.3.3.7
+description: >
+  Throws a ReferenceError from a unsolvable reference as the default parameter.
+---*/
+
+function fn(a = unresolvableReference) {
+  return 42;
```
jugglinmike: You can simplify this a bit by removing the`return` statement

test/language/default-parameters/method-definitions.js
```diff
@@ -0,0 +1,14 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 14.3
+description: Set default parameters on method definitions
+---*/
+
+var o = {
+  m: function(a = 1, b = 2) {
```
jugglinmike: This is not a method definition.

test/language/default-parameters/replace-default-values.js
```diff
@@ -0,0 +1,17 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.2.1
+description: >
+  Replace default values
```
jugglinmike: This could use more formal justification, e.g.

    > 24. If hasDuplicates is true, then
    >     [...]
    > 25. Else,
    >     a. Let formalStatus be IteratorBindingInitialization for formals with
    >        iteratorRecord and env as arguments.

    > ES6 13.3.3.6 Runtime Semantics: IteratorBindingInitialization
    >
    > SingleNameBinding : BindingIdentifier Initializeropt
    >
    > [...]
    > 6. If Initializer is present and v is undefined, then
    >    [...]

Which suggests a good extension: why not test specifying all the falsey
    values as arguments (to be sure the implementor isn't doing a type coercion
    when applying defaults)

test/language/default-parameters/rest-parameters.js
```diff
@@ -0,0 +1,11 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 14.1
+description: >
+  Rest parameters can not have a default
+---*/
+
+assert.throws(SyntaxError, function() {
+  eval('function fn(...args = [1]) {}');
```
jugglinmike: This should be expressed as literal code with the `negative` frontmatter tag

test/language/default-parameters/scope.js
```diff
@@ -0,0 +1,32 @@
...
+function fn2(a = y) {
+  var y = 1;
+}
+
+// y is only defined on the inner scope of fn2
+assert.throws(ReferenceError, function() {
+  fn2();
+});
+
+function fn3(a = function() { return typeof z === 'undefined'; }) {
```
jugglinmike: I think it might be better to declare a variable binding for `z` in the
    outer scope (and then `return z;`), just to ensure that the *correct*
    environment record is in effect.

test/language/default-parameters/scope.js
```diff
@@ -0,0 +1,32 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 13.3.3.7
+description: >
+  Expressions as parameter default values share the top level context only.
```
jugglinmike: This test could use more formal justification. Something like:

    > 9.2.12 FunctionDeclarationInstantiation(func, argumentsList)
    >
    > 27. If hasParameterExpressions is false, then
    >     [...]
    > 28. Else,
    >     a. NOTE A separate Environment Record is needed to ensure that
    >        closures created by expressions in the formal parameter list do
    >        not have visibility of declarations in the function body.
    >     b. Let varEnv be NewDeclarativeEnvironment(env).
    >     c. Let varEnvRec be varEnv’s EnvironmentRecord.
    >     d. Set the VariableEnvironment of calleeContext to varEnv.

test/language/default-parameters/syntaxerror-on-duplicates.js
```diff
@@ -0,0 +1,41 @@
...
+  and BoundNames of FormalParameterList contains any duplicate elements.
+flags: [noStrict]
+---*/
+
+function valid1(a, a) {}
+var valid2 = function(a,a) {};
+(function(a,a) {});
+
+assert.throws(SyntaxError, function() {
+  eval("function fn(a, a = 1) {}");
```
jugglinmike: Each of these assertions should be expressed as a separate file containing literal code and the `negative` frontmatter tag

test/language/default-parameters/syntaxerror-on-duplicates.js
```diff
@@ -0,0 +1,41 @@
...
+  FormalParameters : FormalParameterList
+
+  It is a Syntax Error if IsSimpleParameterList of FormalParameterList is false
+  and BoundNames of FormalParameterList contains any duplicate elements.
+flags: [noStrict]
+---*/
+
+function valid1(a, a) {}
+var valid2 = function(a,a) {};
+(function(a,a) {});
```
jugglinmike: I don't think there is much value in testing two different cases for FunctionExpressions.

test/language/default-parameters/undefined-parameters-doesnt-replace-default.js
```diff
@@ -0,0 +1,17 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 13.3.3.7
+description: >
+  Undefined parameters doesn't replace default values.
```
jugglinmike: The test name/description should concern "undefined arguments", not
    "undefined parameters". And since the subject is plural, the verb should be
    "don't".

jugglinmike: I think this could use more formal justification

    > 24. If hasDuplicates is true, then
    >     [...]
    > 25. Else,
    >     a. Let formalStatus be IteratorBindingInitialization for formals with
    >        iteratorRecord and env as arguments.

    > ES6 13.3.3.6 Runtime Semantics: IteratorBindingInitialization
    >
    > SingleNameBinding : BindingIdentifier Initializeropt
    >
    > [...]
    > 6. If Initializer is present and v is undefined, then
    >    [...]

jugglinmike:
Hey Leo: I left some comments in-line, but I had a couple other thoughts that didn't belong to any single file, so I'll include them here:

- The tests `test/language/default-parameters/previous-parameters.js` and
  `test/language/default-parameters/not-initilialized-parameters-throws.js` are
  directly related, but that isn't reflected in their names. To make them more
  discoverable, I recommend `param-ref-initialized.js` and
  `param-ref-uninitialized.js`

- We should add tests for:
  - section 9.2.12:
    - step 22: substeps a and b (don't forget to specify `noStrict` for these)
    - steps 28.f.i: substeps 4, 5.a, and 5.b (although to be honest, I don't know if 28.f.i.5.b is testable. @anba: you're much better at this than I am; can you think of an example of where that step comes into play?)
  - Default parameters' effect on function length (as per ES6 14.1.6 Static Semantics: ExpectedArgumentCount), e.g.
    - `(function (x = 0) {}).length`
    - `(function (x = 0, y) {}).length`
    - `(function (x, y = 0) {}).length`
    - `(function (x, y = 0, z) {}).length`

anba:
> steps 28.f.i: substeps 4, 5.a, and 5.b (although to be honest, I don't know if 28.f.i.5.b is testable. @anba: you're much better at this than I am; can you think of an example of where that step comes into play?)

I think the *ReturnIfAbrupt* step in 28.f.i.5.b is a spec bug. It's also marked in my local copy of the spec, unfortunately past-André only wrote `???` instead of a proper comment describing what's wrong with that call.

test/language/default-parameters/call-define-values.js
leobalter: The justification is the hardest point which I avoided. But it's now more than reasonable to write it.

Thanks!

leobalter:
I really appreciate the feedback and I'm sorry this PR came too sloppy.

I'll fix it.

test/language/default-parameters/undefined-parameters-doesnt-replace-default.js
leobalter: I removed this file due to the update on `replace-default-values.js`

leobalter:
@jugglinmike:

>  I recommend param-ref-initialized.js and param-ref-uninitialized.js

a0672d6

----

> We should add tests for ...

On the way.

The inline comments are addressed at dcbb97d

leobalter:
@jugglinmike done!

test/language/arguments-object/mapped/simple-parameters-list-provide-mapped-arguments-non-strict.js
```diff
@@ -0,0 +1,41 @@
...
+  _a = a;
+
+  b = 43;
+  return arguments[1];
+}
+
+var result = fn(false, false);
+
+assert.sameValue(result, 43, 'arguments are mapped');
+assert.sameValue(_a, 42, 'parameters are bound to `arguments`');
```
rwaldron: This is definitely a useful test, but I have two thoughts: 

1. This isn't a default parameters test
2. It might be a duplicate

If it's a not a duplicate, can you make this test in a separate PR?

test/language/default-parameters/arrow-function.js
```diff
@@ -0,0 +1,11 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 14.2
+description: >
+  Default parameter values on arrow functions
+---*/
+
+var a = (x = 40, y = 2) => { return x + y };
+
+assert.sameValue(a(), 42);
```
rwaldron: Duplicates

- https://github.com/tc39/test262/blob/master/test/language/expressions/arrow-function/syntax/arrowparameters-cover-initialize-1.js

test/language/default-parameters/function-constructor.js
```diff
@@ -0,0 +1,13 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 19.2.1.1
+description: >
+  Create new Function with default parameters
+---*/
+
+var fn = new Function('a = 42', 'b = "foo"', 'return [a, b]');
```
rwaldron: :clap:

rwaldron:
- Remove the file marked duplicate
- Follow up on the question here: https://github.com/tc39/test262/pull/387/files#r38490733

Once those are addressed, this is good to go!

test/language/arguments-object/mapped/simple-parameters-list-provide-mapped-arguments-non-strict.js
leobalter: roger that

leobalter: Duplicate confirmed:

- https://github.com/tc39/test262/blob/master/test/language/arguments-object/10.6-10-c-ii-1.js
- https://github.com/tc39/test262/blob/master/test/language/arguments-object/10.6-10-c-ii-2.js

leobalter:
fixed, rebased and squashed. The fixup commits are saved here for references: https://github.com/bocoup/test262/tree/default-parameters-fixup

PR #388, opened 2015-07-22
Update tests for Array#find and Array#findIndex

leobalter:


test/built-ins/Array/prototype/find/array-altered-during-loop.js
```diff
@@ -0,0 +1,46 @@
...
+arr.find(function(kValue) {
+  if (results.length === 0) {
+    arr.splice(1, 1);
+  }
+  results.push(kValue);
+});
+
+assert.sameValue(results.length, 3, 'predicate called three times');
+assert.sameValue(results[0], 'Shoes');
+assert.sameValue(results[1], 'Bike');
```
jugglinmike: Couldn't hurt to assert the value of `results[2]`

test/built-ins/Array/prototype/find/predicate-call-parameters.js
```diff
@@ -0,0 +1,60 @@
...
+    e. ReturnIfAbrupt(testResult).
+  ...
+---*/
+
+var arr = ['Mike', 'Rick', 'Leo'];
+
+var results = [];
+
+arr.find(function(kValue, k, O) {
+  results.push({
```
jugglinmike: If you push the `arguments` object (instead of a collection of the expected formal parameters), you can also assert that the predicate wasn't incorrectly invoked with additional arguments (via `results[0].length`).

test/built-ins/Array/prototype/find/predicate-call-parameters.js
```diff
@@ -0,0 +1,60 @@
...
+assert.sameValue(result.O, arr);
+assert.sameValue(result.thisArg, undefined);
+
+result = results[2];
+assert.sameValue(result.kValue, 'Leo');
+assert.sameValue(result.k, 2);
+assert.sameValue(result.O, arr);
+assert.sameValue(result.thisArg, undefined);
+
+assert.sameValue(results.length, 3);
```
jugglinmike: Better to test `results.length` first (that way, if the number is less than expected, a descriptive assertion error will be thrown instead of a TypeError for `results[2]`)

jugglinmike: Oops, I mean, "instead of a TypeError for `results[2].kValue`"

test/built-ins/Array/prototype/find/predicate-call-parameters.js
```diff
@@ -0,0 +1,60 @@
...
+
+var o = {};
+results = [];
+arr.find(function() {
+  results.push(this);
+}, o);
+
+assert.sameValue(results[0], o);
```
jugglinmike: It's unlikely to be a problem, but for consistency's sake, you may want to test `results.length` here as well

test/built-ins/Array/prototype/find/predicate-called-for-each-array-property.js
```diff
@@ -0,0 +1,26 @@
...
+  ...
+  6. If thisArg was supplied, let T be thisArg; else let T be undefined.
+  7. Let k be 0.
+  8. Repeat, while k < len
+    ...
+    d. Let testResult be ToBoolean(Call(predicate, T, «kValue, k, O»)).
+  ...
+---*/
+
+var arr = [undefined, , , 'foo'];
```
jugglinmike: Very good!

test/built-ins/Array/prototype/find/return-abrupt-from-this-length.js
```diff
@@ -0,0 +1,37 @@
...
+    throw new Test262Error();
+  },
+  configurable: true
+});
+// predicate fn is given to avoid false positives
+assert.throws(Test262Error, function() {
+  [].find.call(o, function() {});
+});
+
+delete o.length;
```
jugglinmike: You can simplify this test a little by just making a new object to use as the "this" value for the second invocation. Might be slightly safer, too (since it avoids shared state)

test/built-ins/Array/prototype/find/return-abrupt-from-this-length.js
```diff
@@ -0,0 +1,37 @@
...
+
+var o = {};
+
+Object.defineProperty(o, 'length', {
+  get: function() {
+    throw new Test262Error();
+  },
+  configurable: true
+});
+// predicate fn is given to avoid false positives
```
jugglinmike: Since your expectation is for the user-defined `Test262Error`, you don't have to worry about false positives (a missing predicate would throw a `TypeError`)

test/built-ins/Array/prototype/find/return-abrupt-from-this.js
```diff
@@ -0,0 +1,21 @@
...
+description: >
+  Return abrupt from ToObject(this value).
+info: >
+  22.1.3.8 Array.prototype.find ( predicate[ , thisArg ] )
+
+  1. Let O be ToObject(this value).
+  2. ReturnIfAbrupt(O).
+---*/
+
+// predicate fn is given to avoid false positives
```
jugglinmike: Excellent! Very thoughtful!

test/built-ins/Array/prototype/findIndex/array-altered-during-loop.js
```diff
@@ -0,0 +1,46 @@
...
+arr.findIndex(function(kValue) {
+  if (results.length === 0) {
+    arr.splice(1, 1);
+  }
+  results.push(kValue);
+});
+
+assert.sameValue(results.length, 3, 'predicate called three times');
+assert.sameValue(results[0], 'Shoes');
+assert.sameValue(results[1], 'Bike');
```
jugglinmike: Couldn't hurt to assert the value of `results[2]`

test/built-ins/Array/prototype/findIndex/predicate-call-parameters.js
```diff
@@ -0,0 +1,59 @@
...
+    d. Let testResult be ToBoolean(Call(predicate, T, «kValue, k, O»)).
+  ...
+---*/
+
+var arr = ['Mike', 'Rick', 'Leo'];
+
+var results = [];
+
+arr.findIndex(function(kValue, k, O) {
+  results.push({
```
jugglinmike: If you push the `arguments` object (instead of a collection of the expected formal parameters), you can also assert that the predicate wasn't incorrectly invoked with additional arguments (via `results[0].length`).

test/built-ins/Array/prototype/findIndex/predicate-call-parameters.js
```diff
@@ -0,0 +1,59 @@
...
+assert.sameValue(result.O, arr);
+assert.sameValue(result.thisArg, undefined);
+
+result = results[2];
+assert.sameValue(result.kValue, 'Leo');
+assert.sameValue(result.k, 2);
+assert.sameValue(result.O, arr);
+assert.sameValue(result.thisArg, undefined);
+
+assert.sameValue(results.length, 3);
```
jugglinmike: Better to test `results.length` first (that way, if the number is less than  expected, a descriptive assertion error will be thrown instead of a TypeError for `results[2].kValue`)

test/built-ins/Array/prototype/findIndex/predicate-call-parameters.js
```diff
@@ -0,0 +1,59 @@
...
+
+var o = {};
+results = [];
+arr.findIndex(function() {
+  results.push(this);
+}, o);
+
+assert.sameValue(results[0], o);
```
jugglinmike: It's unlikely to be a problem, but for consistency's sake, you may want to test `results.length` here as well

test/built-ins/Array/prototype/findIndex/return-abrupt-from-this-length.js
```diff
@@ -0,0 +1,37 @@
...
+    throw new Test262Error();
+  },
+  configurable: true
+});
+// predicate fn is given to avoid false positives
+assert.throws(Test262Error, function() {
+  [].findIndex.call(o, function() {});
+});
+
+delete o.length;
```
jugglinmike: You can simplify this test a little by just making a new object to use as the "this" value for the second invocation. Might be slightly safer, too (since it avoids shared state)

jugglinmike:
The tests for each of the two methods are highly related, so my feedback applies to both. I've double-commented in case it helps you keep track.

Nice work!

test/built-ins/Array/prototype/find/return-abrupt-from-this-length.js
leobalter: indeed. Actualy it's better to remove the predicate argument so we can also assert the order.

test/built-ins/Array/prototype/find/predicate-call-parameters.js
```diff
@@ -0,0 +1,60 @@
...
+    k: k,
+    O: O
+  });
+});
+
+var result = results[0];
+assert.sameValue(result.kValue, 'Mike');
+assert.sameValue(result.k, 0);
+assert.sameValue(result.O, arr);
+assert.sameValue(result.thisArg, undefined);
```
leobalter: another bug: this was passing as I didn't set thisArg.

I'll need to move this assertion to 2 separate files, becuase of [non-]strict modes.

leobalter:
@jugglinmike done.

ianwjhalliday:
LGTM

bterlson:
:eggplant:

leobalter:
done, thanks!

PR #389, opened 2015-07-24
Add and update tests for Array.prototype.fill

leobalter:


test/built-ins/Array/prototype/fill/fill-values.js
```diff
@@ -0,0 +1,92 @@
...
+  ToInteger(end).
+  9. ReturnIfAbrupt(relativeEnd).
+  10. If relativeEnd < 0, let final be max((len + relativeEnd),0); else let
+  final be min(relativeEnd, len).
+  11. Repeat, while k < final
+    a. Let Pk be ToString(k).
+    b. Let setStatus be Set(O, Pk, value, true).
+    c. ReturnIfAbrupt(setStatus).
+    d. Increase k by 1.
+  12. Return O.
```
jugglinmike: These are the most critical assertions for the behavior of this method, but I don't think they all belong in the same file. A lot of the detail/justification is difficult to see when they are grouped together like this (and when the assertion description text takes the form of "X #n"). What do you think about breaking it up a bit? Usually I need to experiment with the tests before I can see a reasonable structure, but my first try would involve a separate file for "relativeStart" and "relativeEnd".

test/built-ins/Array/prototype/fill/return-abrupt-from-setting-property-value.js
```diff
@@ -0,0 +1,28 @@
...
+---*/
+
+var a1 = [];
+Object.freeze(a1);
+
+// won't break on an empty array.
+a1.fill(1);
+
+var a2 = [0];
+Object.freeze(a2);
```
jugglinmike: `Object.freeze` makes sense for the first case because it ensures that there is no property setting at all. But for the negative case, we should be more specific about exactly *which* operation we expect to take place. You could do this by defining `0` as an accessor property that throws a `Test262Error` when set.

test/built-ins/Array/prototype/fill/return-abrupt-from-this-length.js
```diff
@@ -0,0 +1,36 @@
...
+  get: function() {
+    throw new Test262Error();
+  },
+  configurable: true
+});
+assert.throws(Test262Error, function() {
+  [].fill.call(o, 1);
+});
+
+delete o.length;
```
jugglinmike: You can simplify this test by creating a new object to use as the "this" value for the second invocation of `fill`

jugglinmike:
All done!

test/built-ins/Array/prototype/fill/fill-values.js
leobalter: relativeStart and relativeEnd is a first good separation. I'll split this way and see if it's good to go further and split again.

They're together as they came this way from the old tests. I had the same impression but I wanted to check first it this would smell in a code review, as it just did.

leobalter: a50398a

test/built-ins/Array/prototype/fill/return-abrupt-from-setting-property-value.js
leobalter: 0db9901

leobalter:
changes done! Thanks

ianwjhalliday:
LGTM

bterlson:
:eggplant:

leobalter:
done, thanks!

PR #390, opened 2015-07-28
Add tests for Array.prototype.copyWithin

leobalter:


test/built-ins/Array/prototype/copyWithin/coerced-values.js
```diff
@@ -0,0 +1,92 @@
...
+  ...
+  11. If end is undefined, let relativeEnd be len; else let relativeEnd be
+  ToInteger(end).
+  ...
+includes: [compareArray.js]
+---*/
+
+assert(
+  compareArray(
+    [0, 1, 2, 3].copyWithin(undefined, undefined),
```
jugglinmike: To be more consistent with the rest of your tests, these assertions should be grouped into three separate files: one for step 5, one for step 8, and one for step 11.

test/built-ins/Array/prototype/copyWithin/fill-holes.js
```diff
@@ -0,0 +1,15 @@
...
+es6id: 22.1.3.3
+description: >
+  Loop from each property, even empty holes.
+includes: [compareArray.js]
+---*/
+
+assert(
+  compareArray(
+    [0, 1, , , 1].copyWithin(0, 1, 4),
+    [1, , , , 1]
```
jugglinmike: [The `compareArray` helper does not take holes into account](https://github.com/tc39/test262/blob/258d2124ab44d210cb92e3211fd321337fba9525/harness/compareArray.js). This means this test isn't as specific as it ought to be because the following would *also* pass (incorrectly):

  ```js
  assert(
    compareArray(
      [0, 1, , , 1].copyWithin(0, 1, 4),
      [1, undefined, undefined, undefined, 1]
    )
  );
  ```

Better to explicitly check each element (and the array length) in this case.

test/built-ins/Array/prototype/copyWithin/return-abrupt-from-this-length.js
```diff
@@ -0,0 +1,37 @@
...
+  4. ReturnIfAbrupt(len).
+---*/
+
+var o1 = {};
+
+Object.defineProperty(o1, 'length', {
+  get: function() {
+    throw new Test262Error();
+  },
+  configurable: true
```
jugglinmike: This line can be removed

jugglinmike:
Hey Leo!

A couple tests are missing a "features" frontmatter entry for `Symbol`:

- test/built-ins/Array/prototype/copyWithin/return-abrupt-from-end-as-symbol.js
- test/built-ins/Array/prototype/copyWithin/return-abrupt-from-start-as-symbol.js

leobalter:
fixup commit for the fixes, next is coming for the missing tests

leobalter:
Oops, I misread your comment. Everything is fixed now.

rwaldron:
@leobalter @bterlson LGTM (very thorough)

PR #391, opened 2015-07-28
Add tests for Reflect

leobalter:
This branch will add tests for the entire Reflect object and its static methods.

~~The current commits are ready to review, while I'll still write the tests for [26.1.6](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-reflect.get) to [26.1.14](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-reflect.setprototypeof) on the following day(s).~~

### DIFF
--- a/test/built-ins/Reflect/construct/newtarget-is-not-constructor-throws.js
+++ b/test/built-ins/Reflect/construct/newtarget-is-not-constructor-throws.js
@@ -1,0 +1,30 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 26.1.2
+description: >
+  Throws a TypeError if `newTarget` is not a constructor.
+info: >
+  26.1.2 Reflect.construct ( target, argumentsList [, newTarget] )
+
+  ...
+  2. If newTarget is not present, let newTarget be target.
+  3. Else, if IsConstructor(newTarget) is false, throw a TypeError exception.
+  ...
+---*/
+
+assert.throws(TypeError, function() {
+  Reflect.construct(function() {}, [], 1);
+});
+
+assert.throws(TypeError, function() {
+  Reflect.construct(function() {}, [], null);
+});
+
+assert.throws(TypeError, function() {
+  Reflect.construct(function() {}, [], {});
+});
+
+assert.throws(TypeError, function() {
+  Reflect.construct(function() {}, [], Date.now);
+});
--- a/test/built-ins/Reflect/construct/return-with-newtarget-argument.js
+++ b/test/built-ins/Reflect/construct/return-with-newtarget-argument.js
@@ -1,0 +1,29 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 26.1.2
+description: >
+  Return target result using newTarget argument.
+info: >
+  26.1.2 Reflect.construct ( target, argumentsList [, newTarget] )
+
+  ...
+  2. If newTarget is not present, let newTarget be target.
+  ...
+  6. Return Construct(target, args, newTarget).
+---*/
+
+var o = {};
+var internPrototype;
+function fn() {
+  this.o = o;
+  internPrototype = Object.getPrototypeOf(this);
+}
+
+var result = Reflect.construct(fn, [], Array);
+assert.sameValue(Object.getPrototypeOf(result), Array.prototype);
+assert.sameValue(
+  internPrototype, Array.prototype,
+  'prototype of this from within the constructor function is Array.prototype'
+);
+assert.sameValue(result.o, o);
--- a/test/built-ins/Reflect/defineProperty/define-properties.js
+++ b/test/built-ins/Reflect/defineProperty/define-properties.js
@@ -1,0 +1,39 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 26.1.3
+description: >
+  Define properties from the attributes object.
+info: >
+  26.1.3 Reflect.defineProperty ( target, propertyKey, attributes )
+
+  ...
+  6. Return target.[[DefineOwnProperty]](key, desc).
+includes: [propertyHelper.js]
+---*/
+
+var o = {};
+var desc;
+
+Reflect.defineProperty(o, 'p1', {
+  value: 42,
+  writable: true,
+  enumerable: true
+});
+
+assert.sameValue(o.p1, 42);
+
+verifyWritable(o, 'p1');
+verifyNotConfigurable(o, 'p1');
+verifyEnumerable(o, 'p1');
+
+var f1 = function() {};
+var f2 = function() {};
+Reflect.defineProperty(o, 'p2', {
+  get: f1,
+  set: f2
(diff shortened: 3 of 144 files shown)


leobalter:
All the tests are done.

Thanks @anba, your runtime helped me a lot to find typos and other errors. ++

anba:
> your runtime helped me a lot to find typos and other errors

Awesome! I'm glad it was of use to you! :smile:

test/built-ins/Reflect/construct/newtarget-is-not-constructor-throws.js
```diff
@@ -0,0 +1,30 @@
...
+assert.throws(TypeError, function() {
+  Reflect.construct(function() {}, [], null);
+});
+
+assert.throws(TypeError, function() {
+  Reflect.construct(function() {}, [], {});
+});
+
+assert.throws(TypeError, function() {
+  Reflect.construct(function() {}, [], Date.now);
```
jugglinmike: Good thinking!

test/built-ins/Reflect/construct/return-with-newtarget-argument.js
```diff
@@ -0,0 +1,26 @@
...
+---*/
+
+var o = {};
+function fn() {
+  this.o = o;
+}
+
+var result = Reflect.construct(fn, []);
+
+assert.sameValue(result.o, o);
```
jugglinmike: This first assertion is unnecessary because this patch includes a dedicated test file for that behavior.

test/built-ins/Reflect/construct/return-with-newtarget-argument.js
```diff
@@ -0,0 +1,26 @@
...
+
+result = Reflect.construct(fn, [], Array);
+assert.sameValue(Object.getPrototypeOf(result), Array.prototype);
```
jugglinmike: This assertion should be followed by a check for `this.o`

test/built-ins/Reflect/construct/return-with-newtarget-argument.js
```diff
@@ -0,0 +1,26 @@
...
+
+  ...
+  2. If newTarget is not present, let newTarget be target.
+  ...
+  6. Return Construct(target, args, newTarget).
+---*/
+
+var o = {};
+function fn() {
+  this.o = o;
```
jugglinmike: It wouldn't hurt to assert the prototype of `this` from within the constructor function (to ensure it isn't being set after construction). If you do that, make sure to take steps to ensure that the test can't pass if the implementor swallows the assertion error.

test/built-ins/Reflect/defineProperty/define-properties.js
```diff
@@ -0,0 +1,41 @@
...
+var o = {};
+var desc;
+
+Reflect.defineProperty(o, 'p1', {
+  value: 42,
+  writable: true,
+  enumerable: true
+});
+
+desc = Object.getOwnPropertyDescriptor(o, 'p1');
```
jugglinmike: This line is unnecessary

test/built-ins/Reflect/deleteProperty/define-properties.js
```diff
@@ -0,0 +1,20 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
```
jugglinmike: The name of this file is incorrect

test/built-ins/Reflect/deleteProperty/return-abrupt-from-property-key.js
```diff
@@ -0,0 +1,24 @@
...
+---*/
+
+var p = {
+  toString: function() {
+    throw new Test262Error();
+  }
+};
+
+assert.throws(Test262Error, function() {
+  Reflect.deleteProperty({}, p, {});
```
jugglinmike: This method only defines two parameters.

test/built-ins/Reflect/get/return-value-from-receiver.js
```diff
@@ -0,0 +1,51 @@
...
+---*/
+
+var o1 = {};
+var receiver = {
+  p: 42
+};
+
+Object.defineProperty(o1, 'p', {
+  get: function() {
+    return this.p;
```
jugglinmike: This test is risky because it will cause stack overflows if the implementation isn't correct. A failure is definitely warranted in that case, but crashing seems a bit harsh :P Why not `return this.somethingElse;` in the accessor function for the `p` property?

test/built-ins/Reflect/preventExtensions/prevent-extensions.js
```diff
@@ -0,0 +1,27 @@
...
+
+  9.1.4 [[PreventExtensions]] ( )
+
+  1. Set the value of the [[Extensible]] internal slot of O to false.
+  ...
+---*/
+
+var o = {};
+Reflect.preventExtensions(o);
+assert.sameValue(Object.isExtensible(o), false, 'object is not extensible');
```
jugglinmike: In addition to "taking their word for it," (i.e. just checking the value of `Object.isExtensible`), I think we should also ensure that the object actually cannot be extended.

test/built-ins/Reflect/set/set-value-on-data-descriptor.js
```diff
@@ -0,0 +1,47 @@
...
+  o1.p, 42,
+  'sets the new value'
+);
+
+var o2 = {p: 43};
+var receiver = {p: 44};
+var result = Reflect.set(o2, 'p', 42, receiver);
+assert.sameValue(result, true, 'returns true on a successful setting');
+assert.sameValue(o2.p, 43, 'with a receiver, does not set a value on target');
+assert.sameValue(receiver.p, 42, 'sets the new value on the receiver');
```
jugglinmike: We write the strangest code for this project, don't we?

jugglinmike:
Hi Leo!

I left some in-line comments. Here are a few high-level ideas that don't fit with any one file:

- For `Reflect.construct`, we should also have a test to ensure that the
  specified arguments are used in the constructor invocation.
- It would be good to add a positive test where the `propertyKey` argument is
  specified as a Symbol with a "bad" `toString` method. This will ensure that
  implementors are following `ToPropertyKey` and not just `ToString`.

test/built-ins/Reflect/set/set-value-on-data-descriptor.js
leobalter: ROFL

leobalter:
> It would be good to add a positive test where the propertyKey argument is specified as a Symbol with a "bad" toString method.

If propertyKey is a aymbol, it will return the Symbol regardless of it's toString method. Also, I can't change it as a symbol is not extensible. 

http://www.ecma-international.org/ecma-262/6.0/index.html#sec-topropertykey

What I am doing is creating additional tests to assure we can have propertyKey with symbol values.

leobalter:
Done. thanks, @jugglinmike!

rwaldron:
@jugglinmike this LGTM, but I'd like you to sign off

leobalter:
@kunalspathak, there are lots of fixup commits on this PR with fixes after the review from @jugglinmike. 

Maybe it's better to review from the PR's diff at this point.

kunalspathak:
LGTM.

goyakin:
@leobalter, let me know once you address @kunalspathak's comments and apply the fixups, and I'll merge these changes.

leobalter:
addressed with commits fixed and rebased.

PR #392, opened 2015-07-29
Replace runTestCase with assert.throws

anba:
I've split the changes into multiple commits so it's easier to review them directly on GitHub. For the same reason I didn't change line indents, if necessary follow-up commits can re-indent the files. 

I plan to remove the `runTestCase` function from all test files, but first I'm going to wait until I get feedback on this PR.

743 files changed, 2269 insertions(+), 7891 deletions(-)


bterlson:
So I think this PR is good to go, seems fine to me. I didn't review everything but did review a random sample. I noticed that the callback passed to assert.throws is indented oddly (usually far too much indentation) - is that a diff artifact or intentional?

smikes:
I have manually reviewed the first four commits, making one comment of substance (and two comments that cancel out).  I will look at built-ins/Array and built-ins tomorrow morning.

Looks amazing so far.  This is very needed work, thanks for taking it on.

anba:
> I noticed that the callback passed to assert.throws is indented oddly (usually far too much indentation) - is that a diff artifact or intentional?

It's intentional, I didn't want to convert the files to `assert.throws` and re-indent at the same time. I can prepare a follow-up PR to clean up the line indentations.

anba:
> Looks amazing so far. This is very needed work, thanks for taking it on.

Great, thanks!

anba:
Updated commit messages as discussed in #396

anba:
@bterlson I'll merge the "Change assertion message" commits into the original commits when reviewing is done. Ok?

anba:
Merged commits.

PR #393, opened 2015-07-29
Improve compatibility for engines without support for Annex B and non-standard features

anba:
Improve compatibility for engines without support for Annex B and non-standard features
- `Function#caller` is not supported in some engines (e.g. Nashorn), added a new `caller` features flag
- Moved (or split) tests which require Annex B support to annexB folder
- And other small changes


test/built-ins/RegExp/S15.10.2.6_A4_T7.js
```diff
@@ -7,13 +7,13 @@ info: >
     AssertionTester closure that takes a State argument x and performs the ...
 es5id: 15.10.2.6_A4_T7
 description: >
-    Execute /\B\[^z]{4}\B/.test("devil arise\tforzzx\nevils") and
+    Execute /\B\[^z\]{4}\B/.test("devil arise\tforzzx\nevils") and
     check results
 ---*/
 
-var __executed = /\B\[^z]{4}\B/.test("devil arise\tforzzx\nevils");
+var __executed = /\B\[^z\]{4}\B/.test("devil arise\tforzzx\nevils");
```
goyakin: I think the original regular expression was wrong here. The intention of this file is to test `\B`, but given that there is no `]` in the string, this test would always fail. Therefore, I think we should change the RegExp to `/\B[^z]{4}\B/` instead by removing the `\` before the `[`.

test/built-ins/Function/15.3.5.4_2-12gs.js
```diff
@@ -7,6 +7,7 @@ description: >
     Strict mode - checking access to non-strict function caller from
     non-strict function (eval includes strict directive prologue)
 flags: [noStrict]
+features: [caller]
```
goyakin: We have more files that use the `caller` property on the `Function` than the ones changed here. Is there a particular reason only these files were updated but the rest were left as they were? If there isn't, we should add the `caller` feature to the others also.

test/built-ins/RegExp/S15.10.2_A1_T1.js
```diff
@@ -20,20 +20,20 @@ var TextSE = "[^<]+";
 var UntilHyphen = "[^-]*-";
 var Until2Hyphens = UntilHyphen + "([^-]" + UntilHyphen + ")*-";
 var CommentCE = Until2Hyphens + ">?";
-var UntilRSBs = "[^]]*]([^]]+])*]+";
-var CDATA_CE = UntilRSBs + "([^]>]" + UntilRSBs + ")*>";
+var UntilRSBs = "[^]\\]*\\]([^]\\]+\\])*\\]+";
```
goyakin: I think this should be `[^\\]]*\\](?:[^\\]]+\\])*\\]+` instead. This seems to have been taken from [http://www.cs.sfu.ca/~cameron/REX.html](Robert D. Cameron's website) where there are also non-JavaScript versions. Those make the intention clearer.

Similar thing applies to `CDATA_CE` and `MarkupDeclCE`, too.

test/annexB/RegExp-leading-escape-BMP.js
```diff
@@ -0,0 +1,29 @@
+// Copyright 2009 the Sputnik authors.  All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+info: >
+    RegularExpressionFirstChar :: BackslashSequence :: \NonTerminator,
+    RegularExpressionChars :: [empty], RegularExpressionFlags :: [empty]
+es5id: 7.8.5_A1.4_T2
+es6id: 11.8.5
```
goyakin: This should say `B.1.4` instead. This also applies to the other new files in the **annexB** folder.

test/built-ins/RegExp/S15.10.2.10_A5.1_T1.js
```diff
@@ -11,12 +11,8 @@ description: "Tested string is \"~`!@#$%^&*()-+={[}]|\\\\:;'<,>./?\" + '\"'"
...
-do {
-   k++;
-   var arr = new RegExp("\\" + non_ident[k], "g").exec(non_ident);   
-} while ((arr !== null) && (arr[0] === non_ident[k]))
-
-if (non_ident.length !== k) {
-   $ERROR('#1: IdentityEscapeSequence :: SourceCharacter but not IdentifierPart');
+for (var k = 0; k < non_ident.length; ++k) {
+  var arr = new RegExp("\\" + non_ident[k], "g").exec(non_ident);
+  assert(arr !== null, "No match for character: " + non_ident[k]);
```
goyakin: I think it would be better to use `assert.notSameValue(arr, null, ...)` here. We would get more information in the execution log this way.

test/annexB/RegExp-leading-escape.js
```diff
@@ -0,0 +1,16 @@
...
+    RegularExpressionFirstChar :: BackslashSequence :: \NonTerminator,
+    RegularExpressionChars :: [empty], RegularExpressionFlags :: [empty]
+es5id: 7.8.5_A1.4_T1
+es6id: 11.8.5
+description: Check similar to (/\1/.source === "\\1")
+---*/
+
+assert.sameValue(/\1/.source, "\\1");
+assert.sameValue(/\a/.source, "\\a");
+assert.sameValue(/\;/.source, "\\;");
```
goyakin: This line and the next are also tested in their original file. Removing these would remove duplication. @anba, what do you think?

The same comment applies to `test/annexB/RegExp-trailing-escape.js`, too.

goyakin:
@anba, in addition to my inline comments, there seems to be merge conflicts in two frontmatter sections. Everything else looks good to me.

rwaldron:
@anba needs a rebase in addition to @goyakin's review

test/built-ins/Function/15.3.5.4_2-12gs.js
anba: The other tests work without any changes because they
- either have been updated in the past to work in engines without `caller` (https://github.com/tc39/test262/blob/master/test/built-ins/Function/15.3.5.4_2-16gs.js#L23)
- or the default `caller` property from `Function.prototype` (8.2.2 CreateIntrinsics, step 12) gives the correct result (that means a `TypeError` is thrown).

test/annexB/RegExp-leading-escape-BMP.js
anba: The file is a test for the lexical grammar (11.8.5), not the regular expression grammar (21.2.1). And since the lexical grammar is not changed for B.1.4, the `es6id` field should be set to 11.8.5, shouldn't it?

anba:
Addressed all review comments (except for `caller` and the es6id, see above), squashed and rebased the PR.

PR #394, opened 2015-07-31
Use guillemets («») to quote string values - fixes #174

### LINKED ISSUE #174: String values in assert.sameValue should be quoted
This looks weird:

```
FAIL test/built-ins/Symbol/species/Symbol.species.in_RegExp.js
     RegExp[Symbol.species] exists per spec
     Exp: no error
     Got: Error: Expected SameValue(returnThis, get [Symbol.species]) to be true
```

This looks better:
```
FAIL test/built-ins/Symbol/species/Symbol.species.in_RegExp.js
     RegExp[Symbol.species] exists per spec
     Exp: no error
     Got: Error: Expected SameValue("returnThis", "get [Symbol.species]") to be true
```

Comparing empty strings also looks weird without quotes:

```
     Got: Error: Expected SameValue(, expected) to be true
```

smikes:
Expected/actual values should be delimited in a way that
isn't confusable with actual string quotes " '

### DIFF
--- a/harness/assert.js
+++ b/harness/assert.js
@@ -30,7 +30,7 @@ assert.sameValue = function (actual, expected, message) {
         message += ' ';
     }
 
-    message += 'Expected SameValue(' + String(actual) + ', ' + String(expected) + ') to be true';
+    message += 'Expected SameValue(«' + String(actual) + '», «' + String(expected) + '») to be true';
 
     $ERROR(message);
 };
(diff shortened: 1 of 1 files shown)


leobalter:
this is covering every value, not only empty strings, so why not extend it to `assert` and `assert.throws` messages?

As an alternative, the french quotes could be applied only to empty strings.

bterlson:
assert could be updated, but throws doesn't present any user data so seems unnecessary there. This is a good start though, so pulling it in!

PR #396, opened 2015-07-31
Replace runTestCase in built-ins/Object

anba:
Next batch of updates to replace `runTestCase`, similar to #392.


test/built-ins/Object/create/15.2.3.5-0-2.js
```diff
@@ -4,12 +4,6 @@
...
-includes: [runTestCase.js]
 ---*/
 
-function testcase() {
-  if (Object.create.length === 2) {
-    return true;
-  }
- }
-runTestCase(testcase);
+assert.sameValue(Object.create.length, 2, 'Object.create.length: ');
```
leobalter: These assertion messages may work fine on failures, but maybe it's better to set it little bit more specific like `"Object.create.length === 2"` on any report uses.

jugglinmike: What do you mean by "report uses"?

leobalter: We don't use any custom reporter, but I may see some future within a website running the tests and reporting each assertion. These messages wouldn't make much change. 

In other perspective: I like (and I'm used to) assertion messages explaining what they do. Here, these messages are only functional on failure cases for the current reporter. And in this case, even the space in the end could be trimmed.

jugglinmike: It sounds like you would like the test report to include actual and expected values, even for satisfied assertions. If Test262 does ever implement a reporter, couldn't this information be emitted programatically? I ask because if so, that would let us avoid duplicating information when writing assertion messages.

leobalter: in this case I wouldn't use the `:`, as `Object.create.length` is fine as the assertion message.

jugglinmike: Ah, okay. I was confused about the first example your provided. I think removing the colon would be fine.

anba: I'll remove the trailing colon and space characters from all error messages. Kind of related:
 - Error messages for `assert(<condition>)` are currently formatted as `'<condition> !== true'`. Should I drop the `!== true` part?
 - Single or double quotes for error message?

leobalter: I like the `'<condition> !== true'` format

> Single or double quotes for error message?

Last time I checked, I saw single quotes recommended in this repo.

anba:
Updated assertion messages.

anba:
@bterlson I'll merge the "Change assertion message" commits into the original commits when reviewing is done. Ok?

bterlson:
This is G2G. I honestly didn't take a close look but by this point I think you have the process down. This applies to the other RTC removal PRs as well :)

anba:
Merged commits.

PR #397, opened 2015-07-31
Move intl tests to sub-folders

anba:
Moved Intl tests to sub-folders, so it's easier to navigate and add new tests.


bterlson:
Ooooo, pretty!!

PR #398, opened 2015-07-31
String.prototype.to{Lower,Upper} tests for special casing

anba:
Changes in "test/built-ins/String/prototype":
- String.prototype.to{LowerCase, UpperCase, LocaleLowerCase, LocaleUpperCase} are specified to implement the locale-insensitive case mappings defined in SpecialCasings.txt, add new tests for this requirement.
- The tests for toLowerCase and toLocaleLowerCase resp. toUpperCase and toLocaleUpperCase are the same. This is intentional, the methods need to implement the same functionality. 

Changes in "test/intl402/String/prototype":
- String.prototype.to{LocaleLowerCase, LocaleUpperCase} are additionally specified to implement the locale-sensitive case mappings defined in SpecialCasings.txt, add new tests for this requirement.

Notes:
  1. ~~This PR currently depends on #397~~ (No longer applies, re-based and force pushed.)
  1. The test cases also include the new ES2015 (and ECMA-402, v2) requirement to iterate over string code points instead of code units in String.prototype.to{LowerCase, UpperCase, LocaleLowerCase, LocaleUpperCase}.

### DIFF
--- a/test/built-ins/String/prototype/toLocaleLowerCase/special_casing.js
+++ b/test/built-ins/String/prototype/toLocaleLowerCase/special_casing.js
@@ -1,0 +1,136 @@
+// Copyright (C) 2015 André Bargull. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: >
+    Check if String.prototype.toLocaleLowerCase supports mappings defined in SpecialCasings
+info: >
+    The result must be derived according to the locale-insensitive case mappings in the Unicode Character
+    Database (this explicitly includes not only the UnicodeData.txt file, but also all locale-insensitive
+    mappings in the SpecialCasings.txt file that accompanies it).
+es5id: 15.5.4.17
+es6id: 21.1.3.20
+---*/
+
+// SpecialCasing.txt, except for conditional mappings.
+
+assert.sameValue("\u00DF".toLocaleLowerCase(), "\u00DF", "LATIN SMALL LETTER SHARP S");
+
+// Locale-sensitive for Turkish and Azeri.
+// assert.sameValue("\u0130".toLocaleLowerCase(), "\u0069\u0307", "LATIN CAPITAL LETTER I WITH DOT ABOVE");
+
+assert.sameValue("\uFB00".toLocaleLowerCase(), "\uFB00", "LATIN SMALL LIGATURE FF");
+assert.sameValue("\uFB01".toLocaleLowerCase(), "\uFB01", "LATIN SMALL LIGATURE FI");
+assert.sameValue("\uFB02".toLocaleLowerCase(), "\uFB02", "LATIN SMALL LIGATURE FL");
+assert.sameValue("\uFB03".toLocaleLowerCase(), "\uFB03", "LATIN SMALL LIGATURE FFI");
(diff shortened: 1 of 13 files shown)


test/intl402/String/prototype/toLocaleUpperCase/special_casing_Lithuanian.js
```diff
@@ -0,0 +1,188 @@
...
+  "\u1ECB\u0307".toLocaleUpperCase("lt"),
+  "\u1ECA",
+  "COMBINING DOT ABOVE preceded by LATIN SMALL LETTER I WITH DOT BELOW (U+1ECB)"
+);
+assert.sameValue(
+  "\uD835\uDC22\u0307".toLocaleUpperCase("lt"),
+  "\uD835\uDC22",
+  "COMBINING DOT ABOVE preceded by MATHEMATICAL BOLD SMALL I (U+1D422 = D835 DC22)"
+);
+for (var i = 0; i < Soft_Dotted.length; ++i) {
```
goyakin: Aren't the tests above duplicated in the `for` loop?

test/built-ins/String/prototype/toLocaleLowerCase/special_casing.js
```diff
@@ -0,0 +1,136 @@
...
+
+/*---
+description: >
+    Check if String.prototype.toLocaleLowerCase supports mappings defined in SpecialCasings
+info: >
+    The result must be derived according to the locale-insensitive case mappings in the Unicode Character
+    Database (this explicitly includes not only the UnicodeData.txt file, but also all locale-insensitive
+    mappings in the SpecialCasings.txt file that accompanies it).
+es5id: 15.5.4.16
+es6id: 21.1.3.22
```
goyakin: The IDs for the `toLocale...` versions of the functions are the same here as the non-locale versions. Could you update them to match the spec?

test/intl402/String/prototype/toLocaleUpperCase/special_casing_Lithuanian.js
```diff
@@ -0,0 +1,188 @@
...
+);
+assert.sameValue(
+  "J\u0307".toLocaleUpperCase("lt"),
+  "J\u0307",
+  "COMBINING DOT ABOVE preceded by LATIN CAPITAL LETTER J"
+);
+
+
+// Code points with Soft_Dotted property (Unicode 5.1, PropList.txt)
+var Soft_Dotted = [
```
goyakin: Even though the variable name `Soft_Dotted` matches what's in the Unicode spec, we should change it to `softDotted` to keep the variable naming consistent across the project.

goyakin: @anba, other than minor comments I left inline, everything else looks good.

test/intl402/String/prototype/toLocaleUpperCase/special_casing_Lithuanian.js
anba: Yes, they are. I think I started with the explicit tests and only later added the `for` loop. I've removed them in the fixup.

anba:
Addressed review comments, rebased and squashed.

rwaldron:
Thanks @anba

PR #399, opened 2015-08-04
Add tests for @@match

jugglinmike:
Introduce tests for the well-known Symbol, `@@match`. This effects the behavior of `String.prototype.match` and is also used by the IsRegExp abstract operation.

@bterlson I wanted to call your attention to one test in particular: built-ins/String/prototype/match/invoke-builtin-match.js. I'm modifying a builtin there, which I know you don't really care for. I think it's required in this case because I wanted to assert that the runtime invokes the `Symbol.match` method of an internally-created RegExp instance. Let me know if you can think of a better approach (or if you'd rather we do without that test).


rwaldron:
@goyakin @bterlson can you take a look at @jugglinmike's question above?

bterlson:
Sorry I missed the question. I think modifying the built-in is perfectly fine. Since each test is run in isolation on a new environment, pollution isn't really a concern. Since I can't think of a better way to test these semantics I think this is fine!

test/built-ins/String/prototype/match/cstm-matcher-invocation.js
```diff
@@ -0,0 +1,33 @@
...
+       a. Let matcher be GetMethod(regexp, @@match).
+       b. ReturnIfAbrupt(matcher).
+       c. If matcher is not undefined, then
+          i. Return Call(matcher, regexp, «O»).
+features: [Symbol.match]
+---*/
+
+var obj = {};
+var returnVal = {};
+var callCount = 0;
```
goyakin: Did you intent to have an assertion for this?

test/built-ins/RegExp/prototype/Symbol.match/builtin-coerce-global.js
```diff
@@ -0,0 +1,97 @@
...
+var r = /./;
+var val, result;
+Object.defineProperty(r, 'global', {
+  get: function() {
+    return val;
+  }
+});
+
+val = false;
+result = r[Symbol.match]('ab');
```
goyakin: Rather than duplicating the call to `@@match` and the assertions, can you create a function that does all these? I see two different outcomes, so we should have two functions.

test/built-ins/RegExp/prototype/Symbol.match/builtin-coerce-lastindex-err.js
```diff
@@ -0,0 +1,34 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Behavior when coercion of `lastIndex` attribute throws an error
```
goyakin: We aren't actually testing the coercion here, but the `Get` call for `lastIndex`.

test/built-ins/RegExp/from-regexp-like-flag-override.js
```diff
@@ -0,0 +1,55 @@
...
+
+Object.defineProperty(obj, 'flags', {
+  get: function() {
+    $ERROR('The `flags` property value should not be referenced.');
+  }
+});
+
+obj[Symbol.match] = true;
+result = new RegExp(obj, 'g');
+assert.sameValue(result.source, 'source text');
```
goyakin: Can you add assertion messages? It won't be immediately obvious which one of these assertions failed otherwise.

jugglinmike: Sure!

test/built-ins/String/prototype/match/cstm-matcher-invocation.js
jugglinmike: Yes I did--adding now

test/built-ins/RegExp/prototype/Symbol.match/builtin-coerce-global.js
jugglinmike: I noticed the duplication as well, but I'm reluctant to exploit the pattern with additional code. I don't expect people to be reading this code very often, but it is far more likely that someone will have to read it than that someone will have to modify it. This is why most of the tests I've written for this project do not make assertions indirectly. I'd rather take an approach that lends itself to transparency and that does not introduce additional abstractions. It leads to more typing, but I'm pretty good with my editor's copy-and-paste facilities :P

But I may be getting ahead of myself because there are reasons to prefer abstraction beyond maintainability. For instance, is file size a concern?

test/built-ins/String/prototype/match/invoke-builtin-match.js
```diff
@@ -0,0 +1,36 @@
...
+  args = arguments;
+  return returnVal;
+};
+
+try {
+  result = 'target'.match('string source');
+
+  assert(thisVal instanceof RegExp);
+  assert.sameValue(thisVal.source, 'string source');
+  assert.sameValue(thisVal.flags, '');
```
goyakin: Since `lastIndex` is supposed to be set to `0` by `RegExpCreate`, could you add an assertion also for that?

test/built-ins/RegExp/prototype/Symbol.match/builtin-get-global-err.js
```diff
@@ -0,0 +1,43 @@
...
+      throw new Test262Error();
+    }
+  }
+});
+
+assert.throws(Test262Error, function() {
+  r[Symbol.match]('');
+});
+
+assert.sameValue(callCount, 2);
```
goyakin: Given that the `Test262Error` is thrown when `callCount` becomes `2`, this is covered by the assertion above.

test/built-ins/RegExp/prototype/Symbol.match/builtin-success-g-set-lastindex-err.js
```diff
@@ -0,0 +1,47 @@
...
+---*/
+
+var r = /b/;
+var callCount = 0;
+
+Object.defineProperty(r, 'lastIndex', { writable: false });
+Object.defineProperty(r, 'global', {
+  get: function() {
+    callCount += 1;
+    if (callCount === 1) {
```
goyakin: @jugglinmike, what do you think about shortening the `if` statement as `callCount > 1`?

test/built-ins/RegExp/prototype/Symbol.match/builtin-success-g-set-lastindex.js
```diff
@@ -0,0 +1,45 @@
...
+features: [Symbol.match]
+---*/
+
+var r = /b/;
+var callCount = 0;
+
+Object.defineProperty(r, 'global', {
+  get: function() {
+    callCount += 1;
+    if (callCount === 1) {
```
goyakin: Same comment about shortening the `if` statement applies to here also.

test/built-ins/RegExp/prototype/Symbol.match/coerce-global.js
```diff
@@ -0,0 +1,66 @@
...
+var result;
+Object.defineProperty(r, 'global', { writable: true });
+
+r.global = undefined;
+result = r[Symbol.match]('aa');
+assert.notSameValue(result, null);
+assert.sameValue(result.length, 1);
+
+r.global = null;
+result = r[Symbol.match]('aa');
```
goyakin: Could you add the assertions?

test/built-ins/RegExp/prototype/Symbol.match/g-match-empty-coerce-lastindex-err.js
```diff
@@ -0,0 +1,55 @@
...
+        valueOf: function() {
+          throw new Test262Error();
+        }
+      };
+      return thisMatch;
+    }
+  };
+};
+
+r.lastIndex = 0;
```
goyakin: We can remove this as it's already done by the `RegExp` constructort.

test/built-ins/RegExp/prototype/Symbol.match/g-match-empty-coerce-lastindex-err.js
```diff
@@ -0,0 +1,55 @@
...
+nextMatch = 'a non-empty string';
+r[Symbol.match]('');
```
goyakin: @jugglinmike, what do you think about splitting this to a separate file as it isn't testing the failure condition?

test/built-ins/RegExp/prototype/Symbol.match/g-match-empty-set-lastindex-err.js
```diff
@@ -0,0 +1,52 @@
...
+      return thisMatch;
+    }
+  };
+};
+var r, nextMatch;
+
+r = /./g;
+r.exec = exec;
+nextMatch = 'a non-empty string';
+r[Symbol.match]('');
```
goyakin: Split into a separate file?

test/built-ins/RegExp/prototype/Symbol.match/builtin-coerce-global.js
goyakin: My concern is readability. Since we have multiple assertions, it is harder here compared to other tests to match the expected outcome for a given input.

If we had for example two functions called `assertCoercible` and `assertNotCoercible`, it would be easy to scan through the file since it would say:

    function assertNotCoercible(global) {
        val = global;

        result = r[Symbol.match]('ab');

        assert.notSameValue(result, null);
        assert.sameValue(result.length, 1);
        assert.sameValue(result[0], 'a');
    }

    ...

    assertNotCoercible(false);
    assertNotCoercible('');
    ...
    assertCoercible(true);
    assertCoercible('truthy');

I'm fine with the version of the code now because, as you said, I don't expect people to read it often.

goyakin:
@jugglinmike, I've added a few more minor comments, but everything else looks good to me.

test/built-ins/String/prototype/match/invoke-builtin-match.js
jugglinmike: Sure

test/built-ins/RegExp/prototype/Symbol.match/builtin-get-global-err.js
jugglinmike: The assertion against `callCount` ensures that the accessor was invoked exactly twice. But in light of the `assert.throws`, that would only catch the most bizarre implementation bugs. I'm fine with removing it.

test/built-ins/RegExp/prototype/Symbol.match/builtin-success-g-set-lastindex-err.js
jugglinmike: I'm not sure of the rationale for that change, but I'm happy to oblige!

test/built-ins/RegExp/prototype/Symbol.match/builtin-success-g-set-lastindex.js
jugglinmike: Sure

test/built-ins/RegExp/prototype/Symbol.match/coerce-global.js
jugglinmike: Sure

test/built-ins/RegExp/prototype/Symbol.match/g-match-empty-coerce-lastindex-err.js
jugglinmike: Alrighty

test/built-ins/RegExp/prototype/Symbol.match/g-match-empty-coerce-lastindex-err.js
jugglinmike: No problem. The new file uses `$ERROR` instead of `throw new Test262Error()` to be a little more clear about the intention of that test.

test/built-ins/RegExp/prototype/Symbol.match/g-match-empty-set-lastindex-err.js
jugglinmike: Sure. I'm unable to make use of `$ERROR` in this case, so the expectation of "does not throw a `TypeError` is a little implicit.

jugglinmike:
Thanks, @goyakin. I think I've addressed all your feedback

goyakin:
Thanks @jugglinmike. If you apply the fixups, we can then merge these changes.

jugglinmike:
Sure sure. I've created [the `bocoup/symbol-match-presquash` branch](https://github.com/bocoup/test262/tree/symbol-match-presquash) to reference the commits prior to squashing for posterity.

PR #401, opened 2015-08-05
Remove invalid Array.from test

anba:
`ArrayBuffer` cannot be constructed from an array literal and `ArrayBuffer` instances are not a valid argument value for `Array.from` (no length property, not iterable).


leobalter:
LGTM

bterlson:
Should we instead test the opposite behavior rather than deleting the test?

leobalter:
@bterlson: I'm currently working on an update for Array.from tests, I can write this one as well.

bterlson:
Ok cool!

Issue #297, opened 2015-06-03
Coding style


jugglinmike:
While reviewing a recent pull request (gh-293), I found myself wanting to give feedback regarding coding style--things like:
- Maximum column width
- Quotation mark usage (single or double)
- White space (hard vs. soft, and width if soft)

...but I had to stop myself because I didn't know the actual policy of the project. I think that any open source project has a lot to gain from consistently-formatted source code. 

Test262 concerns the syntax of the language itself, so there will need to be provisions for tests that explicitly require formatting code that break any specific style recommendation. I don't think this effects the utility of a style guide (especially not for one as brief as I'm suggesting) as such tests will be few and far between.

Although the current level of consistency across tests is a little tough to gauge, I would like to introduce a new standard for future contributions and document these expectations in the project's contribution guidelines.

bterlson:
Agree, these would be good to have (aside from tests for specific sequences of whitespaces or what-have-you).

* Line width of 80 seems excessive. Maybe 100? I'd like to say no hard wraps but I guess some people use editors where this doesn't work nicely?
* Single quotes ftw?!?
* 4-space tabs seems the standard in most tests I can find. Might as well formalize that.
* Probably want a point about asserting one thing per assert (Eg. don't use Boolean operators in your assertion).

Anything else we might want?

domenic:
I think the key thing is that the style should be enforceable and preferably auto-fixable. jscs is probably the best tool for such things these days. Maybe eslint since it would allow custom rules like the don't-use-boolean-operators.

bterlson:
I think automating this is more trouble than it's worth. A fair number of files aren't parseable. It also seems likely to clobber deliberate styles unless the developer is careful.

domenic:
I wouldn't necessarily want to apply it retroactively, but rather to incoming files.

bterlson:
I understand, but if you want a file for some reason to differ from the enforced styles, you would have to add your new file to some exclude list. Or if the file can't be parsed (eg. every negative syntax error). Seems like a lot of overhead for what is really easy for authors to do (once documented :-P)

jugglinmike:
> - Line width of 80 seems excessive. Maybe 100? I'd like to say no hard wraps but I guess some people use editors where this doesn't work nicely?

Soft wrap can be nice for text, especially cases like frontmatter description/info which we don't expect to change often (so we won't miss line-by-line diffs). In my experience, hard wrapping makes source code more readable.

> - Single quotes ftw?!?
[...]

PR #404, opened 2015-08-05
Update tests for Array.from

leobalter:
This branch updates Array.from tests with some fixes as:

- fixed a false positive where `argument-passed-null.js` was returning a TypeError from 22.1.2.1 step 5
- created 2 tests from `argument-passed-null` to assert this as null and items as null
- fixed reference info on `Array.from_arity.js`
- fix tests at `Array.from_forwards-length-for-array-likes`
- static checks on `calling-from-valid-1-onlyStrict.js`
- static checks on `calling-from-valid-1-noStrict.js`
- static checks on `calling-from-valid-2.js`
- removed `create-typedarray-from.js`, that test belong to es6id 22.2.2.1
- removed `mapfn-invalid-typeerror-1.js` - false positive from 22.1.2.1 step 5
- moved `mapfn-invalid-typeerror-2.js` to more spread tests where `mapfn` is not callable
- created a test when `items` is an ArrayBuffer instance. Ref #401
- ...

I also noticed some files using CRLF instead of LF. There is no specific rule about this, but this might be something else addressed later on #297.

@jugglinmike would you confirm if `source-object-iterator-1.js` and `source-object-iterator-2.js` are duplicates from the Symbol.iterator tests we have?


leobalter:
rebased to solve conflicts after changes from #425

test/built-ins/Array/from/mapfn-is-not-callable-typeerror.js
```diff
@@ -0,0 +1,34 @@
...
+  Array.from([], 'string');}
+);
+
+assert.throws(TypeError, function() {
+  Array.from([], true);}
+);
+
+assert.throws(TypeError, function() {
+  Array.from([], 42);}
+);
```
rwaldron: These look funny...

rwaldron:
Thanks @leobalter for the solid cleanup!

PR #406, opened 2015-08-06
Fix missing variable declaration

### LINKED ISSUE #405: Missing variable declaration
The declaration for i on https://github.com/tc39/test262/blob/master/harness/testIntl.js#L1196 is missing. That makes the usage of i global, which is bad, and makes the code not run in strict mode sometimes.

leobalter:
Closes #405


littledan:
LGTM

PR #407, opened 2015-08-06
Add tests for well-known Symbol: @@split

jugglinmike:
Some of these tests are pretty complicated, so I'm especially interested in
ideas for simplification. Unfortunately, many concern semantics that only seem
to be verifiable using odd RegExp-like objects; this is why there is such heavy
usage of `Symbol.species`.

### DIFF
--- a/test/built-ins/RegExp/prototype/Symbol.split/coerce-flags-err.js
+++ b/test/built-ins/RegExp/prototype/Symbol.split/coerce-flags-err.js
@@ -1,0 +1,32 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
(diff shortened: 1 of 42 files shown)


rwaldron:
@jugglinmike I've reviewed these twice this week, hoping to come up with some useful feedback to your question, but these look straight forward to me. Can you point a specific test or set of tests that you'd like me to give more attention?

goyakin:
Thanks @jugglinmike for the tests.

I don't see anything that could be done regarding simplification, so I'm merging the code as it is. I don't know if you have any upcoming changes for this, but if you do, it'd be great if you could submit them as a separate pull request.

jugglinmike:
Mostly I just wanted confirmation that the complexity was necessary, so thanks to you both for that. If I come up with any ideas for simplification, I'll let you know!

PR #408, opened 2015-08-06
Tests for changes introduced in ES2015 (Annex E)

anba:
- String case functions iterate over code points
- Has called before Get in Array.p.reverse
- Add test for web-compat Array.p.splice behaviour; Plus missing test for no arguments case
- ToObject no longer applied to this-value in Array.p.toLocaleString
- ToObject no longer applied to this-value in Object.p.toLocaleString
- Add tests for Object.p.propertyIsEnumerable and symbol property keys
- Add tests for Object.p.hasOwnProperty and symbol property keys
- Test property descriptor attributes of message property
- Tests for RegExp constructor checks
- Date constructor when called with date object
- TimeClip never returns negative zero

### DIFF
--- a/test/built-ins/Error/message_property.js
+++ b/test/built-ins/Error/message_property.js
@@ -1,0 +1,24 @@
+// Copyright (C) 2015 André Bargull. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Error constructor creates own message property
+info: >
+  19.5.1.1 Error ( message )
+
+  ...
+  4.
+    ...
+    c. Let msgDesc be the PropertyDescriptor{[[Value]]: msg, [[Writable]]: true, [[Enumerable]]: false, [[Configurable]]: true}.
+    d. Let status be DefinePropertyOrThrow(O, "message", msgDesc).
+es6id: 19.5.1.1
+includes: [propertyHelper.js]
+---*/
+
+var message = "my-message";
+var error = new Error(message);
+
+verifyEqualTo(error, "message", message);
+verifyNotEnumerable(error, "message");
+verifyWritable(error, "message");
+verifyConfigurable(error, "message");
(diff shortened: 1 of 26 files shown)


test/built-ins/Error/message_property.js
```diff
@@ -0,0 +1,24 @@
...
+---*/
+
+var message = "my-message";
+var error = new Error(message);
+var desc = Object.getOwnPropertyDescriptor(error, "message");
+
+assert.sameValue(desc.value, message, "Property descriptor [[Value]]");
+assert.sameValue(desc.writable, true, "Property descriptor [[Writable]]");
+assert.sameValue(desc.enumerable, false, "Property descriptor [[Enumerable]]");
+assert.sameValue(desc.configurable, true, "Property descriptor [[Configurable]]");
```
rwaldron: These should use the `verify*` property helper functions: 

```
includes:
    - propertyHelper.js
```

```js
verifyWritable(error, "message");
verifyNotEnumerable(error, "message");
verifyConfigurable(error, "message");
```
Example: https://github.com/tc39/test262/blob/master/test/built-ins/Array/from/Array.from-descriptor.js

test/built-ins/Error/message_property_native_error.js
```diff
@@ -0,0 +1,32 @@
...
+  var nativeError = nativeErrors[i];
+
+  var message = "my-message";
+  var error = new nativeError(message);
+  var desc = Object.getOwnPropertyDescriptor(error, "message");
+
+  assert.sameValue(desc.value, message, "Property descriptor [[Value]]");
+  assert.sameValue(desc.writable, true, "Property descriptor [[Writable]]");
+  assert.sameValue(desc.enumerable, false, "Property descriptor [[Enumerable]]");
+  assert.sameValue(desc.configurable, true, "Property descriptor [[Configurable]]");
```
rwaldron: Apply same use of `verify*` property helper functions here.

rwaldron:
Once the descriptor tests are updated, this is good to go

anba:
Updated both tests to use the propertyHelper functions and rebased the PR.

rwaldron:
@anba thank you again!

PR #412, opened 2015-08-07
Add tests for well-known Symbol: @@search

jugglinmike:
@bterlson same as gh-399, I believe modification of a built-in is required by one test (same reason, different method: `built-ins/String/prototype/search/invoke-builtin-search.js`).

### DIFF
--- a/test/built-ins/RegExp/prototype/Symbol.search/set-lastindex-init-err.js
+++ b/test/built-ins/RegExp/prototype/Symbol.search/set-lastindex-init-err.js
@@ -1,0 +1,52 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+es6id: 21.2.5.9
+description: >
+    Behavior when error thrown while initially setting `lastIndex` property
+info: >
+    [...]
+    7. Let status be Set(rx, "lastIndex", 0, true).
+    8. ReturnIfAbrupt(status).
+features: [Symbol.search]
+---*/
+
+var callCount;
+var poisonedLastIndex = {
+  get lastIndex() {
+    callCount += 1;
+  },
+  set lastIndex(_) {
+    throw new Test262Error();
+  }
+};
+var nonWritableLastIndex = {
+  get lastIndex() {
+    callCount += 1;
+  },
+  // This method defined to avoid false positives from unrelated TypeErrors
+  exec: function() {
+    return null;
+  }
+};
+
+callCount = 0;
+assert.throws(Test262Error, function() {
+  RegExp.prototype[Symbol.search].call(poisonedLastIndex);
+});
+assert.sameValue(
+  callCount,
+  1,
+  'Property value was accessed before being set ("poisoned" lastIndex)'
+);
+
+callCount = 0;
+assert.throws(TypeError, function() {
+  RegExp.prototype[Symbol.search].call(nonWritableLastIndex);
+});
+assert.sameValue(
+  callCount,
+  1,
+  'Property value was accessed before being set (non-writable lastIndex)'
+);
--- a/test/built-ins/RegExp/prototype/Symbol.search/success-get-index-err.js
+++ b/test/built-ins/RegExp/prototype/Symbol.search/success-get-index-err.js
@@ -1,0 +1,27 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+es6id: 21.2.5.9
+description: >
+    Behavior when error thrown while accessing `index` property of match result
+info: >
+    [...]
+    14. Return Get(result, "index").
+features: [Symbol.search]
+---*/
+
+var poisonedIndex = {
+  get index() {
+    throw new Test262Error();
+  }
+};
+var fakeRe = {
+  exec: function() {
+    return poisonedIndex;
+  }
+};
+
+assert.throws(Test262Error, function() {
+  RegExp.prototype[Symbol.search].call(fakeRe);
+});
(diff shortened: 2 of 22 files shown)


test/built-ins/String/prototype/search/cstm-search-get-err.js
```diff
@@ -0,0 +1,24 @@
...
+es6id: 21.1.3.15
+info: >
+    [...]
+    3. If regexp is neither undefined nor null, then
+       a. Let searcher be GetMethod(regexp, @@search).
+       b. ReturnIfAbrupt(searcher).
+features: [Symbol.search]
+---*/
+
+var poisonedSplit = {};
```
goyakin: The suffix should be `Search` instead of `Split`.

test/built-ins/String/prototype/search/cstm-search-invocation.js
```diff
@@ -0,0 +1,33 @@
...
+       a. Let searcher be GetMethod(regexp, @@search).
+       b. ReturnIfAbrupt(searcher).
+       c. If searcher is not undefined, then
+          i. Return Call(searcher, regexp, «O»)
+features: [Symbol.search]
+---*/
+
+var regexp = {};
+var returnVal = {};
+var callCount = 0;
```
goyakin: `callCount` isn't used. Did you intend to verify its value after the call to `search`? Given that we verify other values, it may not be useful.

test/built-ins/RegExp/prototype/Symbol.search/success-return-val.js
```diff
@@ -0,0 +1,15 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+es6id: 21.2.5.9
+description: Return value following successful match
+info: >
+    [...]
+    14. Return Get(result, "index").
```
goyakin: Having another test to verify getting `index` from a **result** object returned from a custom `exec` would make it more comprehensive.

test/built-ins/RegExp/prototype/Symbol.search/success-get-index-err.js
```diff
@@ -0,0 +1,27 @@
...
+features: [Symbol.search]
+---*/
+
+var poisonedIndex = {
+  get index() {
+    throw new Test262Error();
+  }
+};
+var fakeRe = {
+  exec: function() {
```
goyakin: @jugglinmike, what do you think about adding new tests to verify that the result object returned from `exec` is either `Object` or `Null`?

Step `5.d.` in Section `21.2.5.2.1` (RegExpExec) says:

    If Type(result) is neither Object or Null, throw a TypeError exception.

test/built-ins/String/prototype/search/cstm-search-invocation.js
```diff
@@ -0,0 +1,33 @@
...
+
+/*---
+description: Invocation of @@search property of user-supplied objects
+es6id: 21.1.3.15
+info: >
+    [...]
+    3. If regexp is neither undefined nor null, then
+       a. Let searcher be GetMethod(regexp, @@search).
+       b. ReturnIfAbrupt(searcher).
+       c. If searcher is not undefined, then
```
goyakin: @jugglinmike, could you add another test to verify that the built-in search is used if `searcher` is `undefined`?

goyakin:
@jugglinmike, I left some comments inline, but otherwise, everything else looks good to me.

test/built-ins/RegExp/prototype/Symbol.search/set-lastindex-init-err.js
```diff
@@ -0,0 +1,52 @@
...
+  },
+  set lastIndex(_) {
+    throw new Test262Error();
+  }
+};
+var nonWritableLastIndex = {
+  get lastIndex() {
+    callCount += 1;
+  },
+  // Thie method defined to avoid false positives from unrelated TypeErrors
```
goyakin: Typo. `Thie` should be `This`.

test/built-ins/String/prototype/search/cstm-search-get-err.js
jugglinmike: Right-o

test/built-ins/String/prototype/search/cstm-search-invocation.js
jugglinmike: That was my intent. I've included the assertion rather than remove the value, since it guards against the (however unlikely) case of repeated invocations of the method.

test/built-ins/RegExp/prototype/Symbol.search/success-return-val.js
jugglinmike: Couldn't hurt!

test/built-ins/RegExp/prototype/Symbol.search/success-get-index-err.js
jugglinmike: Sounds good to me

test/built-ins/String/prototype/search/cstm-search-invocation.js
jugglinmike: Sure thing

test/built-ins/RegExp/prototype/Symbol.search/set-lastindex-init-err.js
jugglinmike: Fixed

jugglinmike:
@goyakin Thanks for the review! I think I've addressed all your feedback.

rwaldron:
@goyakin This LGTM whenever you're ready

goyakin:
@jugglinmike, this is good to go once you apply the fixups.

jugglinmike:
@goyakin No problem! I've created [the `bocoup/symbol-search-presquash-2`branch](https://github.com/bocoup/test262/tree/symbol-search-presquash-2) to reference the commits prior to squashing for posterity.

goyakin:
Thanks @jugglinmike for the tests.

PR #413, opened 2015-08-07
Replace runTestCase in built-ins (part 2)

anba:
Tweaking the regular expressions I'm using to update the test files revealed some more files which are applicable for automatic rewrites.


anba:
Rebased and added another commit to replaced all other uses of `runTestCase` in "test/built-ins".

PR #415, opened 2015-08-11
Replace runTestCase when used with try-finally

anba:
Do we care at all about restoring built-ins after tests have finished? If it's not important or required to restore built-ins, the `finally` blocks in these tests can be removed. But for the time being I've left them there.


jugglinmike:
It's a good question. If we do, the safe approach requires enough code to maybe warrant a helper function:

```js
function extendBuiltin(builtin, prop, value, cb) {
  var original = builtIn[prop];
  var wasDefined = Object.hasOwnProperty.call(original, prop);

  builtIn[prop] = value;

  try {
    cb();
  } finally {
    if (wasDefined) {
      builtIn[prop] = original;
    } else {
      delete builtIn[prop];
    }
  }
}
```

domenic:
We almost certainly do not. The tests explicitly each run in a clean context.

bterlson:
Not necessary IMO. Test262 tests assume a clean environment per run. Running without this isolation will result in a number of failures already.

anba:
Great! Then I'll remove those pesky `finally` blocks. :smile:

anba:
Removed the `finally` blocks and replaced the commits.

PR #417, opened 2015-08-13
Remove harness/runTestCase.js

anba:
Can be applied after #413, #415 and #416 are pulled in.


anba:
Missed a couple of files in the annexB directory. These still need to be updated before runTestCase.js can be removed.

rwaldron:
All set https://github.com/tc39/test262/commit/03d2f68c9c185970a9b7cee5806e8733188db360

anba:
> All set 03d2f68

Thanks @rwaldron !

rwaldron:
@anba no problem & thank you for doing all the cleanup work for this

PR #421, opened 2015-08-20
add 'onlyStrict' flags to frontmatter of strict-mode tests for classes

suwc:
add 'onlyStrict' flags to frontmatter of strict-mode tests for classes


anba:
The implicit strict mode of classes is part of the test, therefore the `onlyStrict` flag must not be set.

Relevant spec chapter:

ES2015, 10.2.1 Strict Mode Code
> All parts of a ClassDeclaration or a ClassExpression are strict mode code.

PR #430, opened 2015-09-11
Fix #429: writing to const variables is only an error in strict mode.

### LINKED ISSUE #429: Writing to const variables is only a TypeError in strict mode
See the following tests:

https://github.com/tc39/test262/blob/master/test/language/statements/const/syntax/const-invalid-assignment-next-expression-for.js
https://github.com/tc39/test262/blob/master/test/language/statements/const/syntax/const-invalid-assignment-statement-body-for-in.js
https://github.com/tc39/test262/blob/master/test/language/statements/const/syntax/const-invalid-assignment-statement-body-for-of.js

Per [the ES2015 spec](http://www.ecma-international.org/ecma-262/6.0/#sec-for-statement-runtime-semantics-labelledevaluation), `const` variables are created with `CreateImmutableBinding`. Writes to immutable bindings silently [fail outside of strict mode](http://www.ecma-international.org/ecma-262/6.0/#sec-declarative-environment-records-setmutablebinding-n-v-s).

Working on PR which will fix these tests and also test the non-strict behavior.

bakkot:
Fix #429.

Also add tests for silent failure of writes to const variables outside of strict mode.

Also, in the case that writing fails silently in strict mode, the for-statement test will now exit instead of entering an infinite loop.

### DIFF
--- a/test/language/statements/const/syntax/const-invalid-assignment-next-expression-for-non-strict.js
+++ b/test/language/statements/const/syntax/const-invalid-assignment-next-expression-for-non-strict.js
@@ -1,0 +1,20 @@
+// Copyright (C) 2015 Rick Waldron. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 13.6.3.7_S5.a.i
+description: >
+    const: invalid assignment in next expression silently fails in non-strict mode
+flags: [noStrict]
+---*/
+
+let hasRun = false;
+let savedValue;
+for (const i = 0; i < 1; i++) {
+  savedValue = i;
+  if (hasRun) {
+    break;
+  }
+  hasRun = true;
+}
+
+assert.sameValue(savedValue, 0);
--- a/test/language/statements/const/syntax/const-invalid-assignment-next-expression-for-strict.js
+++ b/test/language/statements/const/syntax/const-invalid-assignment-next-expression-for-strict.js
@@ -3,9 +3,16 @@
 /*---
 es6id: 13.6.3.7_S5.a.i
 description: >
-    const: invalid assignment in next expression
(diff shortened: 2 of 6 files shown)


michaelficarra:
:+1: LGTM

ianwjhalliday:
This is incorrect.

There is no silent failure for writing to consts in the spec.  CreateImmutableBinding is given true for the second argument for all const variables, marking the binding as "strict".  Attempts to write to a strict binding always throws.

See:
[Note in Table 15 for CreateImmutableBinding(N,S)](http://www.ecma-international.org/ecma-262/6.0/#table-15)
[8.1.1.1.3 CreateImmutableBinding(N,S)](http://www.ecma-international.org/ecma-262/6.0/#sec-declarative-environment-records-createimmutablebinding-n-s)
[8.1.1.1.5 SetMutableBinding(N,V,S)](http://www.ecma-international.org/ecma-262/6.0/#sec-declarative-environment-records-setmutablebinding-n-v-s)

ianwjhalliday:
Note V8 has promised to change to adhere to the spec here.  Mozilla SM is working on it now.  IE 11 and Edge adhere to the spec.  Not sure what state other ES6 implementations are in.

bakkot:
Ah, my mistake, thanks.

If it would be worthwhile, I could do the "in the case that writing fails silently in strict mode, the for-statement test will now exit instead of entering an infinite loop." part as a separate PR, but it doesn't make that much difference.

ianwjhalliday:
No worries.

Eliminating the infinite loop that would occur in a incorrect implementation sounds like a good idea.  I do not know if the test harnesses already timeout long running tests.  @bterlson ?

bterlson:
In some cases timeout can be achieved but it is not reliable so fixing the infinite loop would be good regardless.

PR #447, opened 2015-12-02
Improve test coverage for various Promise methods

anba:


anba:
Added 'class' features tags. Thanks for reviewing @cscott !

rwaldron:
@anba :clap:

PR #450, opened 2015-12-03
Remove unnecessary ES2015 syntax; add feature flags where appropriate.

cscott:
This allows these tests to be more easily reused to test Promise implementations in isolation (for example, `core-js` and `es6-shim`).


rwaldron:
Nice one @cscott!

PR #452, opened 2015-12-12
Don't stomp on the object in `verifyConfigurable` test.

cscott:
Just as we do for `isWritable`, revert the change after we've verified the property.

This helps us run tests in environments without full isolation between tests, as well as protecting test authors from shooting themselves in the foot inadvertently.

### DIFF
--- a/harness/propertyHelper.js
+++ b/harness/propertyHelper.js
@@ -1,5 +1,6 @@
 
 function isConfigurable(obj, name) {
+    var desc = Object.getOwnPropertyDescriptor(obj, name);
     try {
         delete obj[name];
     } catch (e) {
@@ -7,7 +8,12 @@ function isConfigurable(obj, name) {
             $ERROR("Expected TypeError, got " + e);
         }
     }
-    return !Object.prototype.hasOwnProperty.call(obj, name);
+    var result = !Object.prototype.hasOwnProperty.call(obj, name);
+    if (result) {
+        // Put things back the way we found them.
+        Object.defineProperty(obj, name, desc);
+    }
+    return result;
 }
 
 function isEnumerable(obj, name) {


harness/propertyHelper.js
```diff
@@ -7,7 +8,12 @@ function isConfigurable(obj, name) {
             $ERROR("Expected TypeError, got " + e);
         }
     }
-    return !Object.prototype.hasOwnProperty.call(obj, name);
+    var result = !Object.prototype.hasOwnProperty.call(obj, name);
+    if (result) {
+        // Put things back the way we found them.
+        Object.defineProperty(obj, name, desc);
```
goyakin: This call would fail when the property doesn't originally exist given that `desc` would be `undefined`. Instead of checking the result of the `hasOwnProperty` call in the `if` condition, we should check the descriptor.

jugglinmike:
Given that this test uses one aspect of the property descriptor API to test property descriptors themselves, I think there's a nontrivial opportunity for false positives in buggy implementations. It would be good to see unit tests for the new behavior added to [the project's `test/harness` directory](https://github.com/tc39/test262/tree/07eafc665174345700cdfab40a138e2922fd3659/test/harness) so that implementors can trace failures faster (e.g. they will be able to differentiate between bugs in the behavior under test and bugs in the test framework itself).

harness/propertyHelper.js
```diff
@@ -7,7 +8,12 @@ function isConfigurable(obj, name) {
             $ERROR("Expected TypeError, got " + e);
         }
     }
-    return !Object.prototype.hasOwnProperty.call(obj, name);
+    var result = !Object.prototype.hasOwnProperty.call(obj, name);
+    if (result) {
```
ljharb: should this just be `if (result && desc)`?

harness/propertyHelper.js
```diff
@@ -1,5 +1,6 @@
 
 function isConfigurable(obj, name) {
+    var desc = Object.getOwnPropertyDescriptor(obj, name);
```
ljharb: does this test only deal with own properties (it seems like it due to the `delete`)? if so, can this entire test not simply `return !desc || !desc.configurable;`?

anba: The function is using `delete` instead of just `Object.getOwnPropertyDescriptor` to catch implementation errors where properties are reported as configurable through the `Object.getOwnPropertyDescriptor` API, but act like non-configurable when `delete` is used. 

(We should totally add comments to the propertyHelper.js functions to explain why the `Object.getOwnPropertyDescriptor` API isn't sufficient for the use cases in test262.)

ljharb: ok - in that case, then `getOwnPropertyDescriptor` won't work in cases where the property is on an object higher up the prototype chain.

We'd have to walk up the [[Prototype]], and then preserve the object and descriptor that owned the property - so they can be restored later after the `delete`.

domenic: I don't think this is worthwhile. test262 tests are meant to be run in isolation, and I appreciate that the verify helpers make this very clear. Relying on hosts to implement all the semantics necessary for restoring a property descriptor is not sound testing practice; it makes a lot of assumptions about correct implementation which are intimately related to the very assumptions we are trying to test here.

I think this PR is better off closed, or converted into a PR with comments. I'd also support not restoring the originals in the other verify helpers, to improve testing hygiene in those.

ljharb: Tests can't be run in isolation if they're going to be usefully run in a browser context - this seems like a philosophical concern that affects the entirety of the test262 process.

In other words, it's going to be impractical to run every test in a separate realm in a non-command-line context :-/

domenic: False. Browser tests for test262 have always run each in a separate iframe. Anything else is not supported.

ljharb: That's good to know - I'm not aware of any workable in-browser tests for test262 yet, or else I'd be running them on the es6-shim. Pointers commented on https://github.com/paulmillr/es6-shim/issues/278 would be appreciated.

cscott: Right -- there is no actual implementation of test isolation or test262 integration yet.  In the interim, I've been identifying tests which mutate the context object and/or trying to be more gentle about it, so we can run as many of the tests as possible in more-standard test environments.

It wouldn't be a bad idea to add a `destructive` (or `nondestructive`) property tag to tests which  stomp on the global state (or which are asserted not to do so).

WRT whether *all* the property tests should be destructive -- IMO this actually greatly complicates the writing of tests.  You have to carefully order your tests so as not to shoot yourself in the foot.  The reason I proposed this PR is that I think it's better to support the test authors by not stomping on the environment unless absolutely necessary, rather than leave around unexploded ordinance for them to stumble into.

bterlson: There is an implementation of test isolation in the browser, it's the old test262 harness, though it can't run ES6 tests. The isolation is not hard - you just use iframes. Yes, these destructive helpers make it harder to write tests, but what it forces you to do is good - think very hard about side effects, and test different things in different files. Test262 is not a typical test harness - its high order bit is not convenience for test authors but instead value for implementations (especially half-done or buggy ones).

IMO this should be closed.

ljharb: @bterlson according to https://github.com/tc39/test262/pull/453#discussion-diff-47529015, this forces my test262 tests to have a specific ordering. That seems like a footgun that applies to any test author, regardless of environment. (If that comment is wrong, however, please comment on that thread?)

domenic: You are probably testing too many things in one file

ljharb: If you click the link, you'll note that that file has a single line, `verifyConfigurable`, yet @anba says that file has to run last.

ljharb: oops, nevermind, you're right

ljharb: you're right, i think i'd misunderstood the comment. i guess i can just rearrange the descriptor checks, but it still seems like a footgun. side effects ◉︵◉

bterlson: It seems like that test has three lines, each testing one field of the descriptor. verifyConfigurable should go last in that case.

jugglinmike: There's a fine distinction between the type of dangers from destructive
assertions. I agree with @bterlson about test hygiene, but I also recognize the
"footgun" described by @ljharb.

One approach that might address both concerns is by encapsulating all property
descriptor checks into a single assert method. It could still be destructive
(limiting complexity and promoting test hygiene), but it would abstract away
the gotchas of the common case (testing multiple descriptors on the same
object).

We last spoke about this in March, leaving the door open for a future
improvement:

https://github.com/tc39/test262/pull/191#issuecomment-77442739

bterlson: I'd be fine with such a helper! Though note that it only partially addressed the general footgun problem as I understand it. `verifyDescriptor(obj, prop, { writable: true, configurable: false, enumerable: true})` followed by anything else doing something with obj.prop will fail (or perhaps spuriously pass).

leobalter:
@cscott @goyakin @bterlson @jugglinmike what's the conclusion this PR should take?

leobalter:
When I was running the tests for this branch I've found this example from `test/built-ins/TypedArrays/internals/DefineOwnProperty/non-extensible-redefine-key.js`:

```js
var sample = new TA([42, 43]);
  sample.foo = true;
  sample.bar = true;

  Object.preventExtensions(sample);

  assert.sameValue(
    Reflect.defineProperty(sample, "foo", {value:42}),
    true,
    "data descriptor"
  );

  assert.sameValue(sample.foo, 42);
  verifyEnumerable(sample, "foo");
  verifyWritable(sample, "foo");
  verifyConfigurable(sample, "foo");
...
```

`defineProperty` will work here because I'm redefining a pre-existing property, but as I'm preventing extensions in sample.

With these changes applied, verifyConfigurable will return the abrupt completion from trying to define a new descriptor. This comes from [ValidateAndApplyPropertyDescriptor](https://tc39.github.io/ecma262/#sec-validateandapplypropertydescriptor). Where it checks if the object is extensible only if I'm adding a new property.

---

With this case, I really prefer not creating more side effects than we are already doing. Tests should be atomic and a property should be consumed once for each case. Trying to recover the previous status will add an extra functionality that would open a path for anti-patterns.

What I believe we should add instead is a helper for to check a property at once, so we don't need to worry on the order for calling `verify{Enumerable, Writable, Configurable}`.

Example:

```js
verifyProperty(sample, 'foo', { value: 42, enumerable: true, writable: false, configurable: false })
```

This is a case for another PR, anyway.

---

I'm closing this PR due to the long time with no further answer and believe I don't think it's currently a fit. Even though, I appreciate the contribution and efforts.

PR #453, opened 2015-12-13
Object.{values,entries} tests

ljharb:
I'm putting this PR up now for https://github.com/tc39/proposal-object-values-entries to hopefully work out any kinks before the next meeting, and to aid in implementation in various browsers.

I'll be happy to correct any tweaks pointed out, no matter how minor, including adding missing test cases. Thanks!

### DIFF
--- a/test/built-ins/object/entries/exception-during-enumeration.js
+++ b/test/built-ins/object/entries/exception-during-enumeration.js
@@ -4,19 +4,18 @@
 /*---
 description: Object.entries should terminate if getting a value throws an exception
 es7id: pending
-includes: [Test262Error.js]
 author: Jordan Harband
 ---*/
 
 var trappedKey = {
     get a() {
-        throw new Test262Error('This error should be re-thrown');
+        throw new RangeError('This error should be re-thrown');
     },
     get b() {
         $ERROR('Should not try to get the second element');
     }
 };
 
-assert.throws(Test262Error, function () {
+assert.throws(RangeError, function () {
     Object.entries(trappedKey);
 });
--- a/test/built-ins/object/entries/exception-during-enumeration.js
+++ b/test/built-ins/object/entries/exception-during-enumeration.js
@@ -1,0 +1,21 @@
+// Copyright (C) 2015 Jordan Harband. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Object.entries should terminate if getting a value throws an exception
+includes: [Test262Error.js]
+author: Jordan Harband
+---*/
+
+var trappedKey = {
+    get a() {
+        throw new Test262Error('This error should be re-thrown');
+    },
+    get b() {
+        $ERROR('Should not try to get the second element');
+    }
+};
+
+assert.throws(Test262Error, function () {
+    Object.entries(trappedKey);
+});
--- a/test/built-ins/object/entries/exception-during-enumeration.js
+++ b/test/built-ins/object/entries/exception-during-enumeration.js
@@ -3,6 +3,7 @@
 
 /*---
 description: Object.entries should terminate if getting a value throws an exception
+es7id: pending
 includes: [Test262Error.js]
 author: Jordan Harband
 ---*/
--- a/test/built-ins/object/entries/function-property-descriptor.js
+++ b/test/built-ins/object/entries/function-property-descriptor.js
@@ -1,0 +1,12 @@
+// Copyright (C) 2015 Jordan Harband. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Object.entries should be writable, non-enumerable, and configurable
+author: Jordan Harband
+includes: [propertyHelper.js]
+---*/
+
+verifyConfigurable(Object, 'entries');
+verifyNotEnumerable(Object, 'entries');
+verifyWritable(Object, 'entries');
--- a/test/built-ins/object/entries/function-property-descriptor.js
+++ b/test/built-ins/object/entries/function-property-descriptor.js
@@ -8,6 +8,6 @@ author: Jordan Harband
 includes: [propertyHelper.js]
 ---*/
 
-verifyConfigurable(Object, 'entries');
 verifyNotEnumerable(Object, 'entries');
 verifyWritable(Object, 'entries');
+verifyConfigurable(Object, 'entries');
--- a/test/built-ins/object/entries/function-property-descriptor.js
+++ b/test/built-ins/object/entries/function-property-descriptor.js
@@ -3,6 +3,7 @@
 
 /*---
 description: Object.entries should be writable, non-enumerable, and configurable
+es7id: pending
 author: Jordan Harband
 includes: [propertyHelper.js]
 ---*/
--- a/test/built-ins/object/entries/getter-adding-key.js
+++ b/test/built-ins/object/entries/getter-adding-key.js
@@ -2,7 +2,7 @@
 // This code is governed by the BSD license found in the LICENSE file.
 
 /*---
-description: Object.entries sees a new element added by a getter that is hit during iteration
+description: Object.entries does not see a new element added by a getter that is hit during iteration
 es7id: pending
 author: Jordan Harband
 ---*/
@@ -18,15 +18,12 @@ var bAddsC = {
 var result = Object.entries(bAddsC);
 
 assert.sameValue(Array.isArray(result), true, 'result is an array');
-assert.sameValue(result.length, 3, 'result has 3 items');
+assert.sameValue(result.length, 2, 'result has 2 items');
 
 assert.sameValue(Array.isArray(result[0]), true, 'first entry is an array');
 assert.sameValue(Array.isArray(result[1]), true, 'second entry is an array');
-assert.sameValue(Array.isArray(result[2]), true, 'third entry is an array');
 
 assert.sameValue(result[0][0], 'a', 'first entry has key "a"');
 assert.sameValue(result[0][1], 'A', 'first entry has value "A"');
 assert.sameValue(result[1][0], 'b', 'second entry has key "b"');
 assert.sameValue(result[1][1], 'B', 'second entry has value "B"');
-assert.sameValue(result[2][0], 'c', 'third entry has key "c"');
-assert.sameValue(result[2][1], 'C', 'third entry has value "C"');
--- a/test/built-ins/object/entries/getter-adding-key.js
+++ b/test/built-ins/object/entries/getter-adding-key.js
@@ -3,6 +3,7 @@
 
 /*---
 description: Object.entries sees a new element added by a getter that is hit during iteration
+es7id: pending
 author: Jordan Harband
 ---*/
 
--- a/test/built-ins/object/entries/getter-adding-key.js
+++ b/test/built-ins/object/entries/getter-adding-key.js
@@ -1,0 +1,31 @@
+// Copyright (C) 2015 Jordan Harband. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Object.entries sees a new element added by a getter that is hit during iteration
+author: Jordan Harband
+---*/
+
+var bAddsC = {
+    a: 'A',
+    get b() {
+        this.c = 'C';
+        return 'B';
+    }
+};
+
+var result = Object.entries(bAddsC);
+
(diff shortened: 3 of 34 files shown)


test/built-ins/object/entries/exception-during-enumeration.js
```diff
@@ -0,0 +1,22 @@
+// Copyright (C) 2015 Jordan Harband. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Object.entries should terminate if getting a value throws an exception
+es7id: pending
+includes: [Test262Error.js]
```
anba: Test262Error.js is auto-included.

test/built-ins/object/entries/exception-during-enumeration.js
```diff
@@ -0,0 +1,22 @@
...
+author: Jordan Harband
+---*/
+
+var trappedKey = {
+    get a() {
+        throw new Test262Error('This error should be re-thrown');
```
anba: `$ERROR` also throws a Test262Error, so you want to throw something other than a Test262Error.

test/built-ins/object/entries/function-property-descriptor.js
```diff
@@ -0,0 +1,13 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Object.entries should be writable, non-enumerable, and configurable
+es7id: pending
+author: Jordan Harband
+includes: [propertyHelper.js]
+---*/
+
+verifyConfigurable(Object, 'entries');
```
anba: verifyConfigurable needs to be last (#452).

test/built-ins/object/entries/getter-adding-key.js
```diff
@@ -0,0 +1,32 @@
...
+    get b() {
+        this.c = 'C';
+        return 'B';
+    }
+};
+
+var result = Object.entries(bAddsC);
+
+assert.sameValue(Array.isArray(result), true, 'result is an array');
+assert.sameValue(result.length, 3, 'result has 3 items');
```
anba: Added keys should __not__ be present in the result array.

test/built-ins/object/entries/primitive-strings.js
```diff
@@ -0,0 +1,20 @@
...
+var result = Object.entries('abc');
+
+assert.sameValue(Array.isArray(result), true, 'result is an array');
+assert.sameValue(result.length, 3, 'result has 3 items');
+
+assert.sameValue(result[0][0], '0', 'first entry has key "0"');
+assert.sameValue(result[0][1], 'a', 'first entry has value "a"');
+assert.sameValue(result[1][0], '1', 'second entry has key "1"');
+assert.sameValue(result[1][1], 'b', 'second entry has value "b"');
+assert.sameValue(result[2][0], '2', 'second entry has key "2"');
```
anba: Typo: second -> third

test/built-ins/object/entries/tamper-with-global-object.js
```diff
@@ -0,0 +1,19 @@
...
+    Object.entries should not have its behavior impacted by modifications to the global property Object
+es7id: pending
+author: Jordan Harband
+---*/
+
+function fakeObject() {
+    $ERROR('The overriden version of Object was called!');
+}
+
+var global = Function('return this;')();
```
anba: Preferable: Include fnGlobalObject.js and use `fnGlobalObject()` to retrieve the global object.

test/built-ins/object/entries/tamper-with-global-object.js
```diff
@@ -0,0 +1,19 @@
...
+global.Object = fakeObject;
+
+assert.sameValue(Object, fakeObject, 'Sanity check failed: could not modify the global Object');
+assert.sameValue(Object.entries(1).length, 0, 'Expected number primitive to have zero entries');
```
anba: You need to use the original `Object` to access `Object.entries`.

test/built-ins/object/values/exception-during-enumeration.js
```diff
@@ -0,0 +1,22 @@
+// Copyright (C) 2015 Jordan Harband. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Object.values should terminate if getting a value throws an exception
+es7id: pending
+includes: [Test262Error.js]
```
anba: Test262Error.js is auto-included.

test/built-ins/object/values/exception-during-enumeration.js
```diff
@@ -0,0 +1,22 @@
...
+author: Jordan Harband
+---*/
+
+var trappedKey = {
+    get a() {
+        throw new Test262Error('This error should be re-thrown');
```
anba: `$ERROR` also throws a Test262Error, so you want to throw something other than a Test262Error.

test/built-ins/object/values/function-property-descriptor.js
```diff
@@ -0,0 +1,13 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Object.values should be writable, non-enumerable, and configurable
+es7id: pending
+author: Jordan Harband
+includes: [propertyHelper.js]
+---*/
+
+verifyConfigurable(Object, 'values');
```
anba: verifyConfigurable needs to be last (#452).

test/built-ins/object/values/getter-adding-key.js
```diff
@@ -0,0 +1,25 @@
...
+    get b() {
+        this.c = 'C';
+        return 'B';
+    }
+};
+
+var result = Object.values(bAddsC);
+
+assert.sameValue(Array.isArray(result), true, 'result is an array');
+assert.sameValue(result.length, 3, 'result has 3 items');
```
anba: Added keys should __not__ be present in the result array.

test/built-ins/object/values/primitive-strings.js
```diff
@@ -0,0 +1,17 @@
...
+---*/
+
+var result = Object.values('abc');
+
+assert.sameValue(Array.isArray(result), true, 'result is an array');
+assert.sameValue(result.length, 3, 'result has 3 items');
+
+assert.sameValue(result[0], 'a', 'first value is "a"');
+assert.sameValue(result[1], 'b', 'second value is "b"');
+assert.sameValue(result[2], 'c', 'second value is "c"');
```
anba: Typo: second -> third

test/built-ins/object/values/symbols-omitted.js
```diff
@@ -0,0 +1,25 @@
...
+var obj = { key: symValue };
+obj[enumSym] = value;
+Object.defineProperty(obj, nonEnumSym, { enumerable: false, value: value });
+
+var result = Object.values(obj);
+
+assert.sameValue(Array.isArray(result), true, 'result is an array');
+assert.sameValue(result.length, 1, 'result has 1 item');
+
+assert.sameValue(result[0][1], symValue, 'first value is `symValue`');
```
anba: Typo: `result[0][1]` -> `result[0]`

test/built-ins/object/values/tamper-with-global-object.js
```diff
@@ -0,0 +1,19 @@
...
+    Object.values should not have its behavior impacted by modifications to the global property Object
+es7id: pending
+author: Jordan Harband
+---*/
+
+function fakeObject() {
+    $ERROR('The overriden version of Object was called!');
+}
+
+var global = Function('return this;')();
```
anba: Preferable: Include fnGlobalObject.js and use `fnGlobalObject()` to retrieve the global object.

test/built-ins/object/values/tamper-with-global-object.js
```diff
@@ -0,0 +1,19 @@
...
+global.Object = fakeObject;
+
+assert.sameValue(Object, fakeObject, 'Sanity check failed: could not modify the global Object');
+assert.sameValue(Object.values(1).length, 0, 'Expected number primitive to have zero values');
```
anba: You need to use the original `Object` to access `Object.values`.

anba:
And tests using Proxies could be useful to check implementations perform the operations in the correct order. For example something simple like:
```js
var log = "";
var object = {a: 0, b: 0, c: 0};
var handler = {
  get: function(target, propertyKey, receiver) {
    assert.sameValue(target, object, "get target");
    assert.sameValue(receiver, proxy, "get receiver");
    log += "|get:" + propertyKey;
    return target[propertyKey];
  },
  getOwnPropertyDescriptor: function(target, propertyKey) {
    assert.sameValue(target, object, "getOwnPropertyDescriptor");
    log += "|getOwnPropertyDescriptor:" + propertyKey;
    return Object.getOwnPropertyDescriptor(target, propertyKey);
  },
  ownKeys: function(target) {
    assert.sameValue(target, object, "ownKeys");
    log += "|ownKeys";
    return Object.getOwnPropertyNames(target);
  }
};
var check = {
  get: function(target, propertyKey, receiver) {
    assert(propertyKey in target, "handler check: " + propertyKey);
    return target[propertyKey];
  }
};
var proxy = new Proxy(object, new Proxy(handler, check));
var result = Object.values(proxy);
assert.sameValue(log, "|ownKeys|getOwnPropertyDescriptor:a|get:a|getOwnPropertyDescriptor:b|get:b|getOwnPropertyDescriptor:c|get:c", log);
```

test/built-ins/object/entries/tamper-with-global-object.js
ljharb: oops, good call

test/built-ins/object/entries/getter-adding-key.js
ljharb: That's not how the spec language is written - it explicitly allows for that, just like the algorithm for `Reflect.enumerate`.

anba: The current spec has this note:
> [...] any property added or made enumerable during enumeration is not guaranteed to be included in the output;

But:
1. Notes are not normative.
2. The specification algorithm does not consider new properties. (O.\[[OwnPropertyKeys]\]() returns a fixed snapshot of the current property keys, so newly added properties are not part of the _ownKeys_ list.)

anba: The [complete note](https://github.com/tc39/proposal-object-values-entries/blob/4eddea7cc48ab719847043f6427a012558ee6ce8/spec.emu#L67) should be removed, because it doesn't reflect the actual algorithm and introduces unnecessary implementation dependant behaviour.

test/built-ins/object/entries/tamper-with-global-object.js
ljharb: I added `Object.entries` to `fakeObject` instead

test/built-ins/object/entries/getter-adding-key.js
ljharb: Convincing points. Note removed in https://github.com/tc39/proposal-object-values-entries/commit/9bd195495030e4a40b7b558dc857ce57549b056d

test/built-ins/object/values/tamper-with-global-object.js
ljharb: I added `Object.values` to `fakeObject` instead

test/built-ins/object/values/function-property-descriptor.js
ljharb: Ouch - if possible I'd prefer to just wait for that to be merged rather than adding a brittle workaround to test pollution

ljharb:
Proxy tests added, thanks for the suggestion!

test/built-ins/object/entries/function-property-descriptor.js
ljharb: https://github.com/tc39/test262/pull/453#discussion_r47566837

test/built-ins/object/entries/getter-adding-key.js
anba: Yay! :+1:

ljharb:
I believe the only remaining issue is the `verifyConfigurable` ordering, which I'd prefer to be obviated by #452.

Please let me know if there's anything else!

test/built-ins/object/entries/function-property-descriptor.js
ljharb: fixed

ljharb:
Thanks @goyakin!

PR #455, opened 2015-12-16
Add tests for instanceof operator when prototype property is primitive or getter

anba:
New tests because implementation reality doesn't match spec. And browsers don't have a common behaviour so it's also not possible to change the spec to match browsers.

### DIFF
--- a/test/language/expressions/instanceof/primitive-prototype-with-object.js
+++ b/test/language/expressions/instanceof/primitive-prototype-with-object.js
@@ -1,0 +1,32 @@
+// Copyright (C) 2015 André Bargull. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+es6id: 12.9.3
+description: >
+  Throws a TypeError if `prototype` property is not an Object.
+info: >
+  12.9.3 Runtime Semantics: Evaluation
+  RelationalExpression : RelationalExpression instanceof ShiftExpression
+    ...
+    7. Return InstanceofOperator(lval, rval).
+
+    12.9.4 Runtime Semantics: InstanceofOperator(O, C)
+    ...
+    6. Return OrdinaryHasInstance(C, O).
+
+    7.3.19 OrdinaryHasInstance
+    ...
+    3. If Type(O) is not Object, return false.
+    4. Let P be Get(C, "prototype").
+    5. ReturnIfAbrupt(P).
+    6. If Type(P) is not Object, throw a TypeError exception.
+    ...
+---*/
+
+// Check with primitive "prototype" property on non-constructor function.
+Function.prototype.prototype = "";
+
+assert.throws(TypeError, function() {
+  [] instanceof Function.prototype;
+});
--- a/test/language/expressions/instanceof/prototype-getter-with-object.js
+++ b/test/language/expressions/instanceof/prototype-getter-with-object.js
@@ -1,0 +1,47 @@
+// Copyright (C) 2015 André Bargull. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+es6id: 12.9.3
+description: >
+  "prototype" property is retrieved when left-hand side expression in `instanceof` is object.
+info: >
+  12.9.3 Runtime Semantics: Evaluation
+  RelationalExpression : RelationalExpression instanceof ShiftExpression
+    ...
+    7. Return InstanceofOperator(lval, rval).
+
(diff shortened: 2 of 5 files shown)


test/language/expressions/instanceof/primitive-prototype-with-object.js
```diff
@@ -0,0 +1,31 @@
...
+    5. ReturnIfAbrupt(P).
+    6. If Type(P) is not Object, throw a TypeError exception.
+    ...
+---*/
+
+Function.prototype.prototype = "";
+
+assert.throws(TypeError, function() {
+  [] instanceof Function.prototype;
+});
```
leobalter: this looks similar to `test/language/expressions/instanceof/S15.3.5.3_A2_T2.js` and `test/language/expressions/instanceof/S15.3.5.3_A2_T6.js` and these tests (from the PR) might even replace those old versions.

test/language/expressions/instanceof/prototype-getter-with-object.js
```diff
@@ -0,0 +1,39 @@
...
+
+var getterCalled = false;
+
+Object.defineProperty(Function.prototype, "prototype", {
+  get: function() {
+    assert.sameValue(getterCalled, false, "'prototype' getter already called");
+    getterCalled = true;
+    return Array.prototype;
+  },
+  configurable: true
```
leobalter: configurable is not necessary.

test/language/expressions/instanceof/prototype-getter-with-object.js
```diff
@@ -0,0 +1,39 @@
...
+    4. Let P be Get(C, "prototype").
+    5. ReturnIfAbrupt(P).
+    ...
+---*/
+
+var getterCalled = false;
+
+Object.defineProperty(Function.prototype, "prototype", {
+  get: function() {
+    assert.sameValue(getterCalled, false, "'prototype' getter already called");
```
leobalter: "'prototype' getter called once" would improve the assertion message as valid for passing and failing cases.

This is personal opinion but I believe it's always good to inform the goal of the assertion rather than just yielding when it fails.

test/language/expressions/instanceof/prototype-getter-with-primitive.js
```diff
@@ -0,0 +1,35 @@
...
+
+    12.9.4 Runtime Semantics: InstanceofOperator(O, C)
+    ...
+    6. Return OrdinaryHasInstance(C, O).
+
+    7.3.19 OrdinaryHasInstance
+    ...
+    3. If Type(O) is not Object, return false.
+    4. Let P be Get(C, "prototype").
+    5. ReturnIfAbrupt(P).
```
leobalter: I know it's only a non-normative descriptive info, but this test ends on step 3. `If Type(O) is not Object, return false.`

Maybe cutting these 2 additional lines is easier to follow up?

test/language/expressions/instanceof/primitive-prototype-with-object.js
leobalter: You may want to duplicate this test throwing a custom Error from `Function.prototype.prototype` getter? 

Defined as in `test/language/expressions/instanceof/prototype-getter-with-primitive.js` but throwing a Test262Error.

anba: I think we want to keep the old and the new tests to catch this V8 bug:
```
V8 version 4.9.0 (candidate)
d8> function f(){} f.prototype = 0; [] instanceof f
(d8):1: TypeError: Function has non-object prototype '0' in instanceof check
function f(){} f.prototype = 0; [] instanceof f
                                              ^
TypeError: Function has non-object prototype '0' in instanceof check
    at (d8):1:47

d8> var f = Function.prototype; f.prototype = 0; [] instanceof f
false
```

I'll add these notes to all new tests:
```js
// Check with primitive "prototype" property on non-constructor function.
Function.prototype.prototype = "";

// The "prototype" property for constructor functions is a non-configurable data-property,
// therefore we need to use a non-constructor function to install the getter.
Object.defineProperty(Function.prototype, "prototype", { ...
```

leobalter: sounds great!

anba:
Applied all review comments and rebased the PR. Thanks for reviewing @leobalter !

leobalter:
LGTM

goyakin:
Thanks @anba.

PR #457, opened 2015-12-22
Coverage for length and name function properties

anba:
- Add missing tests for "length" property of built-in functions
- Add missing tests for "name" property of built-in functions
- Add basic surface tests for NativeErrors
- Add basic surface tests for TypedArrays

Notes:
- Already uses the updated DataView function lengths from https://github.com/tc39/ecma262/pull/266 (ES2016 Draft 2015-12-20)


leobalter:
@anba, would you mind splitting this in several commits? At least one commit - or even more - for each topic you already mentioned would be way easier to review.

leobalter:
@anba I'm looking at the changes locally and that's a lot of great work there. Can we have the `%TypedArray%` and `TypedArrays` files on a isolated commit? I'm working on the TypedArray coverage and that would be perfect to avoid duplicated work.

anba:
@leobalter Done!

leobalter:
The %TypedArray% and TypedArrays commits seems great.

Beyond the note on the info field (discussed above) I have only a single note: 

the helper variable [TypedArray](https://github.com/anba/test262/commit/2da3da338950e3b005e9f790111890de8ae43ff9#diff-f3608cb9efa521711bbebf0d3b85c4ecR22) is used on a commit where it's not already defined (of course it happened after the rebase I asked).

I still need to look at 7888891 to finish the review for this PR, but the split was very helpful.

goyakin:
I didn't go through every single file, but looked at some, and they look good to me except for the issue @leobalter pointed out. @leobalter, do you see anything else? If not, I'll merge this in once @anba fixes that issue.

leobalter:
As mentioned on the notes, the DataView tests on 7888891 are using the new length rule applied on https://github.com/tc39/ecma262/pull/266. 

I would keep those tests using the ES6 length and add new ones with a `es7id: pending` tag for the new normative rules. The own info on the DataView tests reflects the ES6 behavior: _is equal to the largest number of named arguments shown in the subclause headings for the function description, including optional parameters_

leobalter:
well, as the new normative is already there, the only update we should need instead of adding is replacing the current tags for `es7id: pending` and update the info with the [new text](https://github.com/tc39/ecma262/pull/266/files#diff-3540caefa502006d8a33cb1385720803R21679)

anba:
Rebased and addressed all review comments:
- Changed info text in TypedArrays/**/prototype/constructor.js
- Changed info text DataView/prototype/**/length.js to use ES2016 function length defaults [1]
- Replaced es6id with "es7id:pending" in DataView/prototype/**/length.js [1, 2]
- Moved change for harness/testTypedArray.js to correct commit

[1] only float32, float64, int16, int32, uint16 and uint32 methods
[2] I've used es7id:pending instead of the current ES2016 section number, because planned additions to ES2016 could invalidate those ids.

PR #463, opened 2016-01-07
Improve Promise coverage: Promise Resolve Function

jugglinmike:
See http://www.ecma-international.org/ecma-262/6.0/#sec-promise-resolve-functions

This depends on gh-461


test/built-ins/Promise/resolve/resolve-self.js
```diff
@@ -0,0 +1,46 @@
...
+  resolve = _resolve;
+  reject = _reject;
+});
+var P = function(executor) {
+  executor(resolve, reject);
+  return promise;
+};
+
+Promise.resolve.call(P, promise)
+  .then(function() {
```
boingoing: Do we need to set the prototype property of P to Promise.prototype in order for the constructed value to have [[prototype]] which has Promise.prototype.then function?

P.prototype = Promise.prototype;

test/built-ins/Promise/resolve-poisoned-then-deferred.js
```diff
@@ -0,0 +1,40 @@
...
+---*/
+
+var value = {};
+var poisonedThen = Object.defineProperty({}, 'then', {
+  get: function() {
+    throw value;
+  }
+});
+var promise = new Promise(function(_resolve) {
+  resolve = _resolve;
```
goyakin: In strict mode, this doesn't create a global variable and the call at the end of the file fails because of that.

goyakin:
@jugglinmike Other than the two issues mentioned above, everything else LGTM.

test/built-ins/Promise/resolve/resolve-self.js
jugglinmike: I don't believe this is necessary because the `P` "constructor" (such as it is) returns a `Promise` instance, so (per [the behavior of [[Construct]]](http://www.ecma-international.org/ecma-262/6.0/#sec-ecmascript-function-objects-construct-argumentslist-newtarget)) the constructed value is that same `Promise` instance.

test/built-ins/Promise/resolve-poisoned-then-deferred.js
jugglinmike: Right you are. I've corrected this in a fixup commit.

test/built-ins/Promise/resolve/resolve-self.js
goyakin: @jugglinmike You're right. When @boingoing and I discussed this, it looks like we misinterpreted the spec.

PR #469, opened 2016-01-15
Basic coverage for the %TypedArray% object

leobalter:
This already handles compatibility with the tests at #457.

The idea on this PR is to improve the initial structure for TypedArray tests. 

A more comprehensive coverage is coming on other PRs, testing every spec steps.


test/built-ins/TypedArray/Symbol.species/result.js
```diff
@@ -0,0 +1,18 @@
...
+
+  1. Return the this value.
+features: [TypedArray, Symbol.species]
+includes: [testTypedArray.js]
+---*/
+
+var value = {};
+var getter = Object.getOwnPropertyDescriptor(TypedArray, Symbol.species).get;
+
+assert.sameValue(getter.call({}), value);
```
anba: Typo -> `getter.call(value)`

test/built-ins/TypedArray/length-invalid.js
```diff
@@ -0,0 +1,22 @@
...
+  5. Let elementLength be ToLength(numberLength).
+  6. ReturnIfAbrupt(elementLength).
+  7. If SameValueZero(numberLength, elementLength) is false, throw a RangeError
+  exception.
+  ...
+features: [TypedArray]
+includes: [testTypedArray.js]
+---*/
+
+assert.throws(RangeError, function() {
```
anba: This test throws a `TypeError` in ES2016 (https://github.com/tc39/ecma262/pull/247).

test/built-ins/TypedArray/Symbol.species/result.js
leobalter: f22217b, thanks!

test/built-ins/TypedArray/length-invalid.js
leobalter: It took me a while to think about this case and as I discussed with @jugglinmike: this is still a valid es2015 test flagged with a es6id.

I looked at the changes on the new specs but it's not published/released. So that makes a certainly confusing situation where this test is valid even if we have another file covering the expected behavior with a `es7id: pending` tag. 

From here, I believe the best case would be stick with this test and maybe add a note on info mentioning this is expected to change on es2016.

anba: > I looked at the changes on the new specs but it's not published/released.

The change is present in the ES2016 2015-12-20 draft, [22.2.1.1, step 3](http://tc39.github.io/ecma262/#sec-typedarray):
> <ol><li value=3> If SameValue(NewTarget, here) is true, throw a TypeError exception.

leobalter: bf8c6f8

I included the test for the es2016 behavior but didn't exclude the es6. Both are valid, IMHO.

test/built-ins/TypedArray/prototype/Symbol.toStringTag/prop-desc.js
```diff
@@ -0,0 +1,23 @@
...
+  function is undefined.
+
+  This property has the attributes { [[Enumerable]]: false, [[Configurable]]:
+  true }.
+includes: [propertyHelper.js, testTypedArray.js]
+features: [TypedArray]
+---*/
+
+var TypedArrayPrototype = TypedArray.prototype;
+var desc = Object.getOwnPropertyDescriptor(TypedArrayPrototype, 'byteLength');
```
jugglinmike: This file's description (and some of its body) concerns the unrelated`byteLength` property. Since this patch already has a file for the `byteLength`-related assertions, you can simply delete them from here.

jugglinmike:
Besides that one in-line comment, I think we should avoid over-specifying the features under test. Since these tests are placed in the `built-ins/TypedArray/` directory, I don't see much value in specifying "TypedArray" as a feature. This was my doing, though, so I wrote the patch to fix it: https://github.com/bocoup/test262/pull/59

Otherwise, looks good to me!

leobalter:
Thanks, @jugglinmike! bocoup#59 is already merged.

I still need to fix https://github.com/tc39/test262/pull/469#discussion-diff-50027590, pinging when it's ready.

test/built-ins/TypedArray/prototype/Symbol.toStringTag/prop-desc.js
```diff
@@ -0,0 +1,22 @@
...
+description: >
+  "@@toStringTag" property of TypedArrayPrototype
+info: >
+  %TypedArray%.prototype.byteLength is an accessor property whose set accessor
+  function is undefined.
+
+  This property has the attributes { [[Enumerable]]: false, [[Configurable]]:
+  true }.
+includes: [propertyHelper.js, testTypedArray.js]
+---*/
```
leobalter: It's also missing a features tag

leobalter:
https://github.com/tc39/test262/pull/469#discussion-diff-50027590 fixed at 5f17999.

test/built-ins/TypedArray/length-invalid.js
goyakin: We now have two different versions of the same test, one for ES6 and another for ES7. One of them is going to always fail, but it depends on which version an engine implements.

Given that Test262 is a living test suite (i.e., accepting tests starting from ECMA262 Stage 3), I think we should remove the ES6 version of this test (`test/built-ins/TypedArray/length-invalid.js`). @leobalter, what do you think?

goyakin:
The tests look good to me other than the duplication of one test I mentioned inline.

leobalter:
@goyakin thanks! With #472 solved it's clear and easy to fix this.

leobalter:
ed9cb69 removes the conflicting test. I'm rebasing and squashing this branch and will update when it's ready.

leobalter:
Ok, it's now rebased and squashed on 2 commits which I believe provide the necessary information for the ES2016 test.

PR #471, opened 2016-01-19
Update README.md

littledan:


README.md
```diff
@@ -1,7 +1,9 @@
 ## Test262 is the official ECMAScript (ECMA-262) conformance suite.
 
+Test262 tests conformance to the continually maintained draft future ECMAScript standard found at http://tc39.github.io/ecma262/ , together with any Stage 3 or later TC39 proposals. It is maintained by Gorkem Yakin (@goyakin) with significant contributions from many in the ECMAScript community.
+
```
rwaldron: When was this agreed to?

littledan: Sorry if I got this wrong. I was trying to document the state of the world, especially since the old test262 site is gone. As for test262 being a reflection of the draft spec, that was agreed at the September TC39 meeting https://github.com/rwaldron/tc39-notes/blob/dd8bb263aef908509c91f58e4912482a75c1c74d/es7/2015-09/sept-24.md#8-test262-update . For Gorkem being the maintainer--are you another maintainer, or are there other, additional maintainers? I apologize for any omission. Brian Terlson used to be a maintainer of test262, but he delegated this to Gorkem recently. It's useful to have maintainers listed here so that people can bug a maintainer if they aren't getting a code review.

rwaldron: I was referring to this: 

> Test262 tests conformance to the continually maintained draft future ECMAScript standard found at http://tc39.github.io/ecma262/ , together with any Stage 3 or later TC39 proposals


I think we have different understanding of how that is actually implemented. The _published_ specification must have tests in which implementers can test against \*. test262 should also contain tests that fit this: 

>  Stage 3 or later TC39 proposals.

However, if those Stage 3 (or later) features are in fact _changes_ to previously published editions of the standard, then it's not clear to me how to proceed. I would expect my assertion above\* to still hold to some degree. Tests for published specification **must** be written and published in this project, then updated to match whatever the new semantics according to the changes defined within the Stage 3 (or later) feature.

As to the maintainer "title", I don't care who is called what, but as a T39 member and someone who is very active in this project, I think it does make sense to also list me as someone that can be contacted for submission review.

domenic: > The published specification must have tests in which implementers can test against *

The agreement was that is well served by a tag created at publication time. But master of test262 reflects master of ecma262 + stage 3 proposals.

bterlson: Unless we want to create a tag for ES6 before any ES7 tests are checked in (and thus not include many new ES6 tests in the ES6 tag), this would work, otherwise we will require branches and integration thereof.

domenic: I think we should just create a tag for the day Ecma published 262 6th edition. That's the tests for 6th edition. Anything else is the tests for future editions. Our coverage might not be very good for 6th edition but that's fine.

domenic: Also,

> The published specification must have tests in which implementers can test against *

Implementers agreed at the meeting they preferred master + stage 3, not the published snapshot. Tests for the published snapshot are useless for implementers.

bterlson: > I think we should just create a tag for the day Ecma published 262 6th edition. That's the tests for 6th edition. Anything else is the tests for future editions. Our coverage might not be very good for 6th edition but that's fine.

I'm ok with a tag, but note that this strategy will mean that the es6 tag contains es7 tests. Generally I expect we will get tests for the v.next features before ECMA ratifies the current version.

littledan: In addition to a tag, I'd be OK if someone wants to, say, maintain a branch within this repository to backport certain tests to the ES6 tag, as we get new tests written that fill in the gaps for ES6. This will take a bunch of extra work to test specifically for things in ES2015 which were changed later and are errors, but if it's useful for someone, then that's fine. I don't think we need to maintain this in master, though--that would take a lot of metadata and multiple copies of tests with disagreeing results. It's not just that subsequent versions add more tests, but they also invalidate earlier test results. But for master, let's stay on the ecma262 master.

leobalter: > I think we should just create a tag for the day Ecma published 262 6th edition. That's the tests for 6th edition. Anything else is the tests for future editions. Our coverage might not be very good for 6th edition but that's fine.

Although it looks simple, that would be very ineffective. We still lack the coverage for important features of ES6 today and a tag from the day it was published would not give a reasonable value to use it.

With less effort, a filter using the frontmatter tags is still more effective regarding what we still have today.

The best case scenario would be having all the tests before each final release is published. Even if it demands an extra time from feature champions to support the tests before each spec goes to Stage 4. That won't be a strategy to save the tests for ES2015, but would help to avoid the same issue when the next version is released.

domenic: > Although it looks simple, that would be very ineffective. We still lack the coverage for important features of ES6 today and a tag from the day it was published would not give a reasonable value to use it.

That's fine. There's no value in the old versions anyway.

> a filter using the frontmatter tags is still more effective regarding what we still have today.

Effective for whom? Literally zero consumers of test262 are asking for this.

bterlson: We, for one, would appreciate the ability to get conformance data against a particular version of the spec :) I think maintaining a front-matter attribute isn't a bad idea for this purpose. It's even useful information in other cases (like, answering the question "when was this behavior introduced??") However, it won't be perfect as tests may be removed when made invalid by later versions. This seems good.

domenic: Hmm, OK. Well, Google would not, and does not intend to contract Bocoup to do any work on maintaining that data.

bterlson: > Well, Google would not, and does not intend to contract Bocoup to do any work on maintaining that data.

Ok, so what does this mean practically speaking? If the test262 authors agree here that we want to continue requiring this metadata, what will Google do? Cancel the contract? Fork the project over a trivial amount of work to support more use cases?

littledan: Sorry about this direction the conversation has gone in. I'm fine with someone adding these metadata tags. I'd argue that they are relatively low priority compared to improving test coverage and ensuring the test suite is maintainable, however.

bterlson: > I'd argue that enhancing tests for older versions of ECMAScript is relatively low priority compared to expanding coverage for the ToT spec, however.

I would be very surprised if anyone disagreed with this ;)

> To properly test old versions, we would really need version ceilings, not just version floors.

Yeah, and I think this is fine. Previously valid tests made invalid are deleted (or updated to have the new versions' metadata attribute). Otherwise, running tests that only have an esNid where N <= the version you want to test should be a very close approximation.

PR #473, opened 2016-01-19
Add tests for iterator expression in ForIn/Of head

jugglinmike:
Although the `for..in` statement allows Expressions to define the
iterator, only an AssignmentExpression may occupy this position in the
`for..of` statement.


ianwjhalliday:
Good catch, that's subtle.  ChakraCore got this wrong Microsoft/ChakraCore#147.

LGTM

jugglinmike:
So did V8 https://bugs.chromium.org/p/v8/issues/detail?id=4692

It's pretty easy to overlook!

PR #474, opened 2016-01-22
Fix trivial typo in test

littledan:
Date.prototype.toGMTString.name should be "toGMTString", not "toUTCString".

### DIFF
--- a/test/annexB/Date/prototype/toGMTString/name.js
+++ b/test/annexB/Date/prototype/toGMTString/name.js
@@ -19,7 +19,7 @@ info: >
 includes: [propertyHelper.js]
 ---*/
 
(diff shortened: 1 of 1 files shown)


anba:
`"toUTCString"` is the correct name, because `Date.prototype.toGMTString` is actually the same function object as `Date.prototype.toUTCString`. From [B.2.4.3 Date.prototype.toGMTString](http://tc39.github.io/ecma262/#sec-date.prototype.togmtstring):
> The function object that is the initial value of Date.prototype.toGMTString is the same function object that is the initial value of Date.prototype.toUTCString.

And we also seem to miss the test that `Date.prototype.toUTCString === Date.prototype.toGMTString`. At least I haven't found one in test/annexB/Date/prototype/toGMTString.

PR #475, opened 2016-01-22
Tests for ES2015/2016 Completion Reform

jugglinmike:
I initially set out to test the statement completion values that ES2015 modified. Note that I included tests for `for..of` despite this statement not being specified previously. My thinking is that eager implementors may have simply copied the ES5 completion values for `for..in` prior to addressing ES2015 completion reform.

As I was finalizing the patch, [I learned](https://bugs.chromium.org/p/v8/issues/detail?id=4695) that [completion values have been further modified by the ES2016 draft](https://github.com/tc39/ecma262/pull/101). This patch reflects the state of the latest draft. (I've included the fix as a separate commit so retrieval will be easier in the event that we agree on a method to maintain tests for prior editions.)

Satisfying the ES2016 specification required the removal of two test cases; as far as I can tell, it is no longer possible to assert behavior for those specific conditions. I'd be happy to be shown otherwise, though!

@goyakin You'll see that I'm proposing somewhat generic "empty completion" tests that assert behavior for both "break" and "continue" completions. I'm on the fence about further factoring those into distinct files for each completion type (given that the spec doesn't make a distinction), so I'm deferring to your judgment on that.


goyakin:
@jugglinmike Regarding `break` and `continue` assertions being in the same file, I think it's fine. I'll merge the PR as it is.

PR #476, opened 2016-01-23
Stop test262 from hanging Spidermonkey with an unimportant error message

### LINKED ISSUE #458: S15.4.4.10_A3_T1.js and S15.4.4.10_A3_T2.js hang SpiderMonkey for a long time
We don't yet use ToLength instead of ToUint32 in Array#slice. So we end up calling the first `$ERROR`. That contains `... Actual:'+arr`, which is going to stringify that massive array. This is really slow for us. You could totally keep this test and just remove the `+arr` part.

evilpie:
This fixes #458 and should allow Spidermonkey to successfully run test262 to completion.

Somebody has to check if signed a CLA before, I can't remember.


leobalter:
LGTM

Issue #477, opened 2016-01-24
Definition for frontmatter id tags


leobalter:
As mentioned at https://github.com/tc39/test262/issues/472#issuecomment-173006824:

> @littledan: Rather than IDs which point to specific spec sections, I think it'd be much more useful to have ids that are the anchors within the document. These are much more likely to be consistent across releases. You could call these esids.

Based on this, and as it's easy to locate the section ids on tc39/ecma262 on each of the `emu-clause` tags, which are expected to be unique, I believe we can use the respective hash id for the Frontmatter's `id` and further versions:

```yaml
id: sec-completion-record-specification-type
```

this value can be converted to links that could be found anywhere:

- https://tc39.github.io/ecma262/#sec-completion-record-specification-type
- http://www.ecma-international.org/ecma-262/6.0/index.html#sec-constructor-properties-of-the-global-object
- mylocalcopy#sec-constructor-properties-of-the-global-object

This reference works good and it's still compatible with the previous formats, as won't conflict with the recently adopted `es7id: pending` value, which can be easily replaced by `id: pending` and it's still useful for anything on Stage 3 and it's not on the draft yet.

@goyakin, do you like this idea?

domenic:
I think it should just be "id"

leobalter:
Yes, it makes sense. I'll update the title and example.

goyakin:
I think this is a good idea. The [test harness](https://github.com/bterlson/test262-harness/) doesn't use the current `es6id`, so we can start using the new one without any issues. I don't know if there's any other tool depending on `es6id` though.

I also agree with @littledan that this way we wouldn't have to keep up with changes in section numbers.

littledan:
+1

jugglinmike:
I'm a little late to the party, but I'd like to know more about the decision to use the name `id` instead of `esid` as suggested by @littledan. The latter is a few more characters, but I think it's preferable. Many tests will share the same value, and that contradicts commonly-held assumptions about "IDs". `esid` (or even `specid`) makes it clear that this value is reference to an identifier of some other document (and also gives a hint as to *which* document it refers).

There wasn't any discussion about why the change was made, so I thought I'd ask here.

littledan:
Any of those names sounds good to me.

leobalter:
I would like to reopen this issue.

[...]

PR #479, opened 2016-01-26
Add tests for async functions

bterlson:
Been sitting on these for too long. Please take a look and let me know what needs fixing up and where people would like to see more coverage.


test/built-ins/AsyncFunction/AsyncFunction-is-subclass.js
```diff
@@ -0,0 +1,15 @@
...
+/*---
+author: Brian Terlson <brian.terlson@microsoft.com>
+es7id: pending
+description: >
+  %AsyncFunction% is a subclass of Function
+---*/
+async function foo() { };
+var AsyncFunction = foo.constructor;
+assert.sameValue(Object.getPrototypeOf(AsyncFunction), Function, "Prototype of constructor");
+assert.sameValue(Object.getPrototypeOf(AsyncFunction.prototype), Function.prototype, "Prototype of constructor's prototype");
```
rwaldron: The assertion messages in these two could be clearer, but no one should block on that.

rwaldron:
Since `AsyncFunction` is not exposed as a property of the global object, as `Function` is, there should be a single test that asserts this

test/built-ins/AsyncFunction/AsyncFunction-name.js
```diff
@@ -0,0 +1,12 @@
...
+
+/*---
+author: Brian Terlson <brian.terlson@microsoft.com>
+es7id: pending
+description: >
+  %AsyncFunction% has a name of "AsyncFunction".
+---*/
+
+var AsyncFunction = async function foo() { }.constructor;
+assert.sameValue(AsyncFunction.name, "AsyncFunction");
```
rwaldron: Needs the descriptor verification assertions

test/built-ins/AsyncFunction/instance-has-name.js
```diff
@@ -0,0 +1,12 @@
...
+
+/*---
+author: Brian Terlson <brian.terlson@microsoft.com>
+es7id: pending
+description: Async function declarations have a name property
+---*/
+
+async function foo () { };
+
+assert.sameValue(foo.name, "foo");
```
rwaldron: Needs the descriptor verification assertions

test/language/expressions/async-arrow-function/arrow-returns-promise.js
```diff
@@ -0,0 +1,15 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+author: Brian Terlson <brian.terlson@microsoft.com>
+es7id: pending
+description: >
+  Async arrow functions return promises
+---*/
+
+var p = (async () => await 1 + await 2)();
```
rwaldron: Is it worthwhile to include an `instanceof` check?

test/language/expressions/await/await-BindingIdentifier-nested.js
```diff
@@ -0,0 +1,15 @@
+// Copyright 2016 Microsoft, Inc. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+author: Brian Terlson <brian.terlson@microsoft.com>
+es7id: pending
+description: >
+  Await is allowed as an identifier in functions nested in async functions
```
rwaldron: Should read: "is not"

test/language/expressions/await/await-awaits-thenable-not-callable.js
```diff
@@ -0,0 +1,19 @@
...
+author: Brian Terlson <brian.terlson@microsoft.com>
+es7id: pending
+description: >
+  Await can await any thenable. If the thenable's then is not callable,
+  await evaluates to the thenable
+---*/
+
+async function foo() {
+  var thenable = { then: 42 };
+  var res = await thenable
```
rwaldron: Semi-colon?

test/language/expressions/await/await-in-nested-generator.js
```diff
@@ -0,0 +1,20 @@
+// Copyright 2016 Microsoft, Inc. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+author: Brian Terlson <brian.terlson@microsoft.com>
+es7id: pending
+description: >
+  Await is allowed as an identifier in functions nested in async functions
```
rwaldron: "identifier in functions nested in async functions" => "identifier in **generator** functions nested in async functions"

test/language/expressions/object/method-definition/early-errors-object-method-NSPL-with-USD.js
```diff
@@ -0,0 +1,13 @@
...
+
+/*---
+author: Brian Terlson <brian.terlson@microsoft.com>
+es7id: pending
+description: >
+  It is a Syntax Error if ContainsUseStrict of AsyncConciseBody is *true* and IsSimpleParameterList of ArrowParameters is *false*.
+negative: SyntaxError
+---*/
+({
+  foo(x = 1) {"use strict"}
```
rwaldron: Was this intended to be an ArrowFunction?

test/language/expressions/object/method-definition/object-method-returns-promise.js
```diff
@@ -0,0 +1,15 @@
+// Copyright 2016 Microsoft, Inc. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+author: Brian Terlson <brian.terlson@microsoft.com>
+es7id: pending
+description: >
+  Async function expressions return promises
```
rwaldron: "function expressions" => "method definition"

test/language/statements/async-function/declaration-returns-promise.js
```diff
@@ -0,0 +1,13 @@
...
+/*---
+author: Brian Terlson <brian.terlson@microsoft.com>
+es7id: pending
+description: >
+  Async functions return promises
+---*/
+
+async function foo() { };
+var p = foo();
+assert(p instanceof Promise, "async functions return promise instances");
```
rwaldron: Ah, I see this now—ignore my quesion from an earlier test

rwaldron:
It's possible that I simply missed this, but I was looking for a negative test that shows a failure when the AwaitExpression is missing an operand, eg:

```js
async function foo() {
  await;
}
```

test/built-ins/AsyncFunction/instance-construct.js
```diff
@@ -0,0 +1,16 @@
...
+/*---
+author: Brian Terlson <brian.terlson@microsoft.com>
+es7id: pending
+description: >
+  Async function instances are not constructors and do not have a
+  [[Construct]] slot.
+---*/
+
+assert.throws(TypeError, function() {
+  async function foo() { }
```
jugglinmike: Recommend moving this statement outside of the assertion body, in the unlikely event that the implementation throws something like `new TypeError('Terrible function name')`

test/language/expressions/async-arrow-function/arrow-returns-promise.js
jugglinmike: Or even more precisely, a `Object.getPrototypeOf` check

test/language/expressions/async-arrow-function/arrow-returns-promise.js
```diff
@@ -0,0 +1,15 @@
...
+p.then(function (v) {
+  assert.sameValue(v, 3);
+  $DONE();
+}, $DONE);
```
jugglinmike: I recently started explicitly invoking `$DONE` with a error description from within a closure, just in case the promise was rejected with out a reason (or with any falsey reason) since [the helper wouldn't report an error in those cases](https://github.com/tc39/test262/blob/fbad34959ffedb02c5b8c040e0f992a6fc006f36/harness/doneprintHandle.js#L5)

test/language/expressions/async-function/early-errors-expression-formals-contains-super-call.js
```diff
@@ -0,0 +1,11 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+author: Brian Terlson <brian.terlson@microsoft.com>
+es7id: pending
+description: It is a syntax error if FormalParameters contains SuperProperty is true
+negative: SyntaxError
+---*/
+
+(async function foo (foo = super()) { let bar; });
```
jugglinmike: The `let` binding isn't necessary here

test/language/expressions/async-function/early-errors-expression-formals-contains-super-property.js
```diff
@@ -0,0 +1,11 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+author: Brian Terlson <brian.terlson@microsoft.com>
+es7id: pending
+description: It is a syntax error if FormalParameters contains SuperCall is true
+negative: SyntaxError
+---*/
+
+(async function foo (foo = super.foo) { let bar; });
```
jugglinmike: The `let` binding isn't necessary here

test/language/expressions/await/await-awaits-thenable-not-callable.js
```diff
@@ -0,0 +1,19 @@
...
+  assert.sameValue(res, thenable);
+}
+
+foo().then($DONE, $DONE);
```
jugglinmike: This relies on Promise semantics, which are not under test. To some extent, there's no avoiding this, but we can mitigate some issues by incrementing a global `callCount` variable after the assertion, and asserting *that* prior to invoking `$DONE`.

Then again, this is ES7, so maybe we have to start taking correct Promise operation for granted...

test/language/expressions/await/await-in-function.js
```diff
@@ -0,0 +1,12 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+author: Brian Terlson <brian.terlson@microsoft.com>
+es7id: pending
+description: >
+  Await is an identifier in a function
+---*/
+
+function foo(await) { return await }
```
jugglinmike: Missing semi-colon for the `return` statment

test/language/expressions/await/await-in-generator.js
```diff
@@ -0,0 +1,12 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+author: Brian Terlson <brian.terlson@microsoft.com>
+es7id: pending
+description: >
+  Await in a generator is an identifier
+---*/
+
+function* foo(await) { yield await };
```
jugglinmike: Missing semi-colon for the `yield` ExpressionStatement

test/language/expressions/await/await-in-nested-function.js
```diff
@@ -0,0 +1,20 @@
...
+author: Brian Terlson <brian.terlson@microsoft.com>
+es7id: pending
+description: >
+  Await is allowed as an identifier in functions nested in async functions
+---*/
+
+var await;
+async function foo() {
+  function bar() {
+    await = 1;
```
jugglinmike: Nice one!

test/language/statements/async-function/evaluation-body-that-throws-after-await.js
```diff
@@ -0,0 +1,23 @@
...
+  Errors thrown from the async function body reject the returned promise
+---*/
+
+async function foo() {
+  await Promise.resolve();
+  throw 1;
+}
+
+foo().then(function () {
+  $ERROR("Should not be called");
```
jugglinmike: You'll want to call `$DONE("Should not be called")` since [`$ERROR` simply throws an error](https://github.com/tc39/test262/blob/fbad34959ffedb02c5b8c040e0f992a6fc006f36/harness/sta.js#L16-L19), which in this context would reject the promise but not signal test failure.

test/language/statements/async-function/evaluation-body-that-throws.js
```diff
@@ -0,0 +1,21 @@
...
+description: >
+  Errors thrown from the async function body reject the returned promise
+---*/
+
+async function foo() {
+  throw 1;
+}
+
+foo().then(function () {
+  $ERROR("Should not be called");
```
jugglinmike: Same as above w/r/t `$ERROR`

test/language/statements/async-function/evaluation-body.js
```diff
@@ -0,0 +1,18 @@
...
+author: brian terlson <brian.terlson@microsoft.com>
+es7id: pending
+description: >
+  async function bodies are executed immediately (unlike generators)
+---*/
+
+let called;
+async function foo() {
+  called = true;
+  await new promise();
```
jugglinmike: I guess it doesn't matter much here, but `new promise()` will trigger a ReferenceError

test/language/statements/async-function/evaluation-default-that-throws.js
```diff
@@ -0,0 +1,19 @@
...
+es7id: pending
+description: >
+  If a default expression throws, the promise is rejected.
+info: >
+  This is different from generators which will throw the error out of the generator
+  when it is called.
+---*/
+async function foo(x = null++) {}
+foo().then(function () {
+  $ERROR('promise should be rejected');
```
jugglinmike: Same dealie with `$ERROR`

test/language/statements/async-function/syntax-declaration-no-line-terminator.js
```diff
@@ -0,0 +1,13 @@
+// Copyright 2016 Microsoft, Inc. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+author: Brian Terlson <brian.terlson@microsoft.com>
+es7id: pending
+description: Async function declarations cannot have a line break after `async`
+info: Reference error is thrown due to looking up async in strict mode
+flags: [onlyStrict]
+negative: ReferenceError
```
jugglinmike: Since this is a runtime error, could you assert this with `assert.throws`? I still have hopes for gh-196

jugglinmike:
These look good, Brian! I had a few ideas for additional tests:

- Assigning to `await` expression or an async function (validating observance
  of the changes to IsValidSimpleAssignmentTarget)
- SyntaxError for "async generators", e.g. `async function* g() {}`
- If I'm reading the spec correctly, then aync functions should not define a
  `prototype` property, so:

      assert.sameValue(async function() {}.prototype, undefined);

test/language/expressions/async-arrow-function/early-errors-arrow-arguments-in-formal-parameters.js
```diff
@@ -0,0 +1,12 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+author: Brian Terlson <brian.terlson@microsoft.com>
+es7id: pending
+description: It is a SyntaxError if FormalParameters contains arguments
+negative: SyntaxError
+---*/
+
+async(arguments) => {  }
```
ianwjhalliday: This isn't in the spec, is it meant to be? I just undid this check for general arrow functions

test/language/expressions/async-arrow-function/early-errors-arrow-eval-in-formal-parameters.js
```diff
@@ -0,0 +1,11 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+author: Brian Terlson <brian.terlson@microsoft.com>
+es7id: pending
+description: It is a SyntaxError if FormalParameters contains eval
+negative: SyntaxError
+---*/
+
+async(eval) => {  }
```
ianwjhalliday: Ditto not in spec

test/language/expressions/async-arrow-function/early-errors-arrow-arguments-in-formal-parameters.js
bterlson: This is true in strict mode. Should be onlyStrict.

test/language/expressions/async-arrow-function/early-errors-arrow-eval-in-formal-parameters.js
bterlson: Ditto.

test/language/statements/async-function/evaluation-default-that-throws.js
```diff
@@ -0,0 +1,19 @@
...
+/*---
+author: Brian Terlson <brian.terlson@microsoft.com>
+es7id: pending
+description: >
+  If a default expression throws, the promise is rejected.
+info: >
+  This is different from generators which will throw the error out of the generator
+  when it is called.
+---*/
+async function foo(x = null++) {}
```
anba: `null++` is an early reference error, so the parser will already reject this test file.

bterlson:
Additional tests required:

* Mapped vs. unmapped arguments object
* this-value tests
* non-negative super tests

test/language/expressions/object/method-definition/early-errors-object-method-NSPL-with-USD.js
bterlson: No, this is a method definition.

bterlson:
Addressed review feedback!

test/language/statements/async-function/evaluation-this-value-global.js
```diff
@@ -0,0 +1,17 @@
...
+
+/*---
+author: Brian Terlson <brian.terlson@microsoft.com>
+es7id: pending
+description: >
+  The this value is set to the global object when not passed in sloppy mode.
+flags: [noStrict]
+---*/
+
+var glob = Function("return this;")();
```
jugglinmike: Could you use the `fnGlobalObject.js` helper here? Using that consistently allows implementers to replace with a host-specific equivalent in cases where the Function constructor is disallowed.

test/language/statements/async-function/syntax-declaration-no-line-terminator.js
```diff
@@ -0,0 +1,14 @@
+// Copyright 2016 Microsoft, Inc. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+author: Brian Terlson <brian.terlson@microsoft.com>
+es7id: pending
+description: Async function declarations cannot have a line break after `async`
+info: Reference error is thrown due to looking up async in strict mode
+flags: [onlyStrict]
```
jugglinmike: I believe this test is equally valid for non-strict mode

test/language/statements/class/definition/methods-async-super-call-param.js
```diff
@@ -0,0 +1,24 @@
...
+class A {
+  async method() {
+    return 'sup';
+  }
+}
+
+class B extends A {
+  async method(x = super.method()) {
+    var y = await super.method();
+    assert.sameValue(y, 'sup');
```
jugglinmike: We should probably include an assertion for the value of `x`. Maybe the default value should be `await super.method()`? Or should that be a separate test?

jugglinmike:
Hi Brian! I've left a couple comments in-line. Some of my feedback doesn't
apply to one specific file, so I'll include it here:

- The latest changeset introduces a few files named `test.txt` that should be
  removed.
- There have been two recent changes to expected test structure that concern
  the files in this patch:
  1. We have moved from `es7id` to the more generic `esid` (see
     https://github.com/tc39/test262/issues/477#issuecomment-178131536 and
     https://github.com/tc39/test262/pull/516)
  2. Asynchronous tests must now specify the `async` tag (see
     https://github.com/tc39/test262/pull/504)

jugglinmike:
Another recent change in `master` affects this patch: @michaelficarra's work with Function.prototype.toString. Could you add tests for async functions [following his example](https://github.com/tc39/test262/commit/6b7cbb50350e478a543da40d02492aa8f45a792d)?

leobalter:
I just pinged @bterlson today on irc, I hope we can have this merged soon!

bterlson:
Sorry missed the notifications on this thread. I will make updates hopefully this week!

test/language/statements/async-function/evaluation-this-value-global.js
jugglinmike: Actually, scratch that. Since gh-568 and gh-595, we are now encouraging the use of the `this` keyword in the global scope to access the global object.

test/language/statements/async-function/syntax-declaration-no-line-terminator.js
bterlson: It isn't because async; will just be an identifier lookup that won't throw...

test/language/statements/class/definition/methods-async-super-call-param.js
bterlson: We already have tests for awaiting super in a method, so I just deleted the second await and changed the assertion to x.

test/language/statements/async-function/syntax-declaration-no-line-terminator.js
jugglinmike: I think the ReferenceError occurs outside of strict mode

>     ExpressionStatement : Expression ;
>
>     1. Let exprRef be the result of evaluating Expression.
>     2. Return ? GetValue(exprRef). 

and

>     6.2.3.1 GetValue (V)
>
>     1. ReturnIfAbrupt(V).
>     2. If Type(V) is not Reference, return V.
>     3. Let base be GetBase(V).
>     4. If IsUnresolvableReference(V) is true, throw a ReferenceError exception.

which is distinct from

>     PutValue (V, W)
>
>     1. ReturnIfAbrupt(V).
>     2. ReturnIfAbrupt(W).
>     3. If Type(V) is not Reference, throw a ReferenceError exception.
>     4. Let base be GetBase(V).
>     5. If IsUnresolvableReference(V) is true, then
>        a. If IsStrictReference(V) is true, then
>             i. Throw a ReferenceError exception.
>        b. Let globalObj be GetGlobalObject().
>        c. Return ? Set(globalObj, GetReferencedName(V), W, false).

bterlson:
Addressed all the review feedback here with the exception of the toString tests (which will not run in Chakra due to [other bugs](https://github.com/Microsoft/ChakraCore/issues/971) and so are harder for me to write).

leobalter:
LGTM. 

I'm ok to have the toString tests in a follow-up PR.

leobalter:
oops, just another detail: `test/language/expressions/async-function/test.txt`

jugglinmike:
I believe the ReferenceError in `syntax-declaration-no-line-terminator.js` also occurs outside of strict mode. My rationale is [here](https://github.com/tc39/test262/pull/479#discussion_r62938070).

littledan:
These tests are great. @caitp wrote hundreds of procedurally generated syntax tests at https://codereview.chromium.org/1841543003/diff/480001/test/cctest/test-parsing.cc . Any chance at integrating these? Also semantics tests at https://codereview.chromium.org/1895603002 but you might already have equivalent coverage (look at the files starting with test/mjsunit/harmony).

test/language/statements/async-function/syntax-declaration-no-line-terminator.js
bterlson: Agree, agree. Of course. Thank you!

bterlson:
I have addressed the remaining feedback (again). Should be good to go.

I did not explore the v8 tests, but I don't think that is on the critical path to landing this PR personally!

test/language/statements/class/definition/methods-async-super-call-param.js
```diff
@@ -0,0 +1,24 @@
...
+
+class A {
+  async method() {
+    return 'sup';
+  }
+}
+
+class B extends A {
+  async method(x = super.method()) {
+    assert.sameValue(x, 'sup');
```
littledan: Maybe you want 'await x' here.

bterlson: sure do! this test will fail.

bterlson:
Fixed. I now think this is ready to go :)

test/language/expressions/object/method-definition/early-errors-object-method-arguments-in-formal-parameters.js
```diff
@@ -0,0 +1,12 @@
+// Copyright 2016 Microsoft, Inc. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+author: Brian Terlson <brian.terlson@microsoft.com>
+esid: pending
+description: It is a SyntaxError if FormalParameters contains arguments
+negative: SyntaxError
```
caitp: This needs to be `onlyStrict` like early-errors-arrow-arguments-in-formal-parameters.js --- this restriction is only enforced in strict mode

test/language/expressions/object/method-definition/early-errors-object-method-eval-in-formal-parameters.js
```diff
@@ -0,0 +1,12 @@
+// Copyright 2016 Microsoft, Inc. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+author: Brian Terlson <brian.terlson@microsoft.com>
+esid: pending
+description: It is a SyntaxError if FormalParameters contains eval in strict mode
+negative: SyntaxError
```
caitp: same thing here --- this restriction is only enforced in strict mode, needs [onlyStrict]

test/language/statements/class/definition/early-errors-class-method-eval-in-formal-parameters.js
```diff
@@ -0,0 +1,12 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+author: Brian Terlson <brian.terlson@microsoft.com>
+esid: pending
+description: It is a SyntaxError if FormalParameters contains eval in strict mode
+negative: SyntaxError
+---*/
+class Foo {
+  async foo(eval) { }
```
caitp: ditto here

test/language/expressions/async-function/early-errors-expression-body-contains-super-call.js
```diff
@@ -0,0 +1,11 @@
+// Copyright 2016 Microsoft, Inc. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+author: Brian Terlson <brian.terlson@microsoft.com>
+esid: pending
+description: It is a syntax error if AsyncFunctionBody contains SuperProperty is true
```
caitp: description is misleading, this is a SuperCall

test/language/expressions/async-function/early-errors-expression-formals-contains-super-call.js
```diff
@@ -0,0 +1,11 @@
+// Copyright 2016 Microsoft, Inc. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+author: Brian Terlson <brian.terlson@microsoft.com>
+esid: pending
+description: It is a syntax error if FormalParameters contains SuperProperty is true
```
caitp: SuperCall

test/built-ins/Function/prototype/toString/async-method.js
```diff
@@ -0,0 +1,15 @@
...
+esid: pending
+description: Function.prototype.toString on an async method
+features: [async-functions]
+---*/
+
+let f = { /* before */async f /* a */ ( /* b */ ) /* c */ { /* d */ }/* after */ }.f;
+let g = { /* before */async /* a */ [ /* b */ "g" /* c */ ] /* d */ ( /* e */ ) /* f */ { /* g */ }/* after */ }.g;
+
+assert.sameValue(f.toString(), "async f /* a */ ( /* b */ ) /* c */ { /* d */ }");
+assert.sameValue(g.toString(), "async f /* a */ [ /* b */ \"g\" /* c */ ] /* d */ ( /* e */ ) /* f */ { /* g */ }");
```
jugglinmike: This assertion is faulty because "f" is not applicable here. The expected value should be:

    "async /* a */ [ /* b */ \"g\" /* c */ ] /* d */ ( /* e */ ) /* f */ { /* g */ }"

jugglinmike:
I've reviewed this patch one more time and found one final detail holding us
back. We've been at this for quite some time, and the fix is straightforward,
so if you'd like, I can fix it as I merge--let me know!

littledan:
@jugglinmike Seems like a good idea to me for you to fix as you merge it. Thanks!

leobalter:
I'm applying this fix

leobalter:
I applied the fix, rebased, squashed and merged this.

:eggplant:

PR #480, opened 2016-01-28
Document frontmatter's id tag

### LINKED ISSUE #477: Definition for frontmatter id tags

leobalter:
Closes #477

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -81,19 +81,19 @@ For best practices on how to use the negative tag please see Handling Errors and
 #### es5id
 **es5id**: [es5-test-id]
 
-This tag identifies the portion of the ECMAScript 5.1 standard that is tested by this test.  It was automatically generated for tests that were originally written for the ES5 version of the test suite and are now part of the ES6 version.
+This tag identifies the section number from the portion of the ECMAScript 5.1 standard that is tested by this test.  It was automatically generated for tests that were originally written for the ES5 version of the test suite and are now part of the ES6 version.
 
 When writing a new test for ES6, it is only necessary to include this tag when the test covers a part of the ES5 spec that is incorporated into ES6. All other tests should specify the `es6id` (see below) instead.
 
 #### es6id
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -89,6 +90,13 @@ When writing a new test for ES6, it is only necessary to include this tag when t
 
 This tag identifies the portion of the ECMAScript 6 standard that is tested by this test.
```
goyakin: The new `id` is an anchor in the spec whereas the old ones are section numbers. I think we should update the old ones' descriptions to call out the difference.

leobalter: :+1: I will do it

leobalter: 8375822

@goyakin tell me if it needs improvement.

goyakin:
Merged as cb754d2450d682afc25c6d263cb2a7ff17862a7d

Issue #470, opened 2016-01-18
Procedural Test Generation


jugglinmike:
The following language features are observable from a number of distinct syntactic forms\*:

- Destructuring assignment (3)
- Spead operator (4)
- Default parameters (16)
- Destructuring binding (18)

Testing a given feature following this project's current practices would involve authoring highly-related tests that differed only in the syntactic form. For instance, completely testing [13.3.3.6](http://www.ecma-international.org/ecma-262/6.0/#sec-destructuring-binding-patterns-runtime-semantics-iteratorbindinginitialization) Step 6.b alone would require 18 separate test files scattered throughout the project file hierarchy. The semantics behind destructuring binding are very involved, so this 1-to-18 ratio makes for a large surface area for testing. These tests would vary according to a well-defined pattern.

Although writing such tests is a dull task, that cost is paid once and enjoyed many times. *Maintaining* those tests, on the other hand, is a more concerning prospect. When tests are added in the future, there is nothing to support a contributor in maintaining parity across all the different syntactic forms (or even recognizing that as an expectation).

One solution\*\* would be to maintain files describing the tests abstractly and use a tool to generate the related tests. I spoke about this at the July 2015 TC-39 meeting ([slides here](https://jugglinmike.github.io/presentations/2015/tc39-test262-update/)), and [the committee's conclusion at that time was to "continue exploration."](https://github.com/rwaldron/tc39-notes/blob/master/es7/2015-07/july-30.md)

I've been doing just that, and here is what I think we'll need:

1. "Region" handling. We should be able to write a test template that defines
   one or more insertion points, which can then be expanded with values
   specified in external "test case" files.
2. Frontmatter specification. Files should be able to control the YAML value
   emitted in the generated tests: `es6id`, `description`, `info`, `features`
   and `flags`.
3. Generation of syntactically invalid JavaScript. We'll want to use this tool
   to assert SyntaxErrors, as well!

And although this is not a hard requirement, I think it would be good if the generated test files document from which source files they were derived.

Here's how I think the solution should behave:

1. Accept as input one or more "test case" files
2. Parse these files for the following information:
   - the location of a set of test template files
   - values for the "regions" that should be used to expand the templates
3. Load the test templates as specified in `#2` and expand the "regions" with
   the provided values
4. Output each file to a location on disk according to meta data in both the
   "test case" and test template

[...]

Issue #306, opened 2015-06-09
Proposal: Automated Tests


jugglinmike:
I'd like to write some tests for test file content. As a start, these would
assert:
- Valid YAML frontmatter syntax
- Basic frontmatter tag requirements
  - Presence of `description` tag
  - Valid configurations
    - Mutual exclusivity of `strictOnly` and `noStrict`
    - At least one of `es5id` and `es6id`
- Presence of copyright block

These assertions are admittedly a little weak, but even superficial checks like
these would have prevented past errors (e.g. gh-177, gh-225, gh-144). Because
test262 does not currently offer any automated tests, this work would involve
introducing the tooling and practices. That would reduce the friction in
implementing more substantial/helpful tests in the future (see for instance,
gh-276 and gh-297).

This will require some discussion about dependencies, but I wanted to see if
the basic proposal is desired before getting into technical details.

jugglinmike:
This could also enforce a restriction on description length that is [currently requested in the project's contribution guidelines](https://github.com/tc39/test262/blob/12cc01484eb4253894d7e9a3478445382e80b262/CONTRIBUTING.md#description).

leobalter:
#546 will have us one step closer to solve this.

The items you have listed here sound reasonable to me.

rwaldron:
This was addressed by implementing a linter that runs in our CI

PR #481, opened 2016-01-29
Prototype Test Generation Tool & Workflow

jugglinmike:
This patch represents the latest in [an ongoing design discussion](https://github.com/tc39/test262/issues/470) on procedural test generation in Test262. It contains a number of technical decisions (which I will detail in just a minute), but it also involves a workflow for the continued maintenance of these tests.

Before getting in to technical details, I would like to call out the most crucial aspects of this proposal:

- Development moves to a `dev` branch, `master` is used for publishing "built" tests
- We use [the TravisCI service](https://travis-ci.org/) (free for open source projects) to ensure that the distributed files always reflect the latest "source" files

I have three reasons for recommending a `dev` branch. First, it avoids any interruption in service for the consumers of Test262--they will continue to find every test within the `test/` subdirectory of the `master` branch. Second, it minimizes confusion for contributors--we can configure GitHub to render the `dev` branch by default, so contributors are less likely to accidentally edit built files. Third, it minimizes the risk of trying this approach (at any time, we can discard the `dev` branch and return to editing files in `master` directly, forgetting that some test files were originally created as the result of a build process).

Under this workflow, if we merge a patch like this to the `dev` branch:

https://github.com/jugglinmike/test262/commit/dev-gen-ex-fix

...we could expect `master` to be automatically updated with a patch like this:

https://github.com/jugglinmike/test262/commit/dev-gen-ex-rebuild

Now, as for technical details, I can identify six decisions that require validation:

1. DSL/Syntax
2. Programming language
3. Workflow automation
4. Metadata syntax
5. Directory structure 
6. File extension

I am interested in finding the best solution for each of these problems, regardless of the initial state of this proposal. That said, my decisions are not without reason, so I'll include some explanation here as a way to get the discussion started.

**DSL/Syntax** Test templates define injection points via C-style multi-line comments that are further delimited by curly braces (`{` and `}`). For example:

    function f(/*{ bindings }*/) {}

Test "case" files define values for those insertion points via single-line comments prefixed with a dash (`-`) character:

    //- bindings
    a, b = null, c

I selected this pattern so that source files could be tokenized as JavaScript (even though they can't necessarily be fully parsed as such). Requiring additional delimiters allows comments to be used for documentation purposes within the source files.

**Programming language** I built the prototype using Python largely because there was precedence in this project (see the deprecated test runner and the website build system). Node.js seems like an obvious alternative, but I can think of a few reasons to avoid a JavaScript VM. First, it may be taken as an endorsement of that VM specifically (which ECMA may want to avoid). Second, there is some risk in self-hosting the project (a bug in V8 may lead to a bug in the tests that V8 could not detect). That honestly seems pretty protracted, though.

One benefit of Node.js (beyond familiarity) is that there are far more advanced JavaScript code analysis tools available for that platform. This pull request contains a naive sort of tokenizer of my own design, but it's not clear if the use case we're discussing actually requires anything more powerful.

In any event, the tool is less than 400 LOC, so re-writing in some other scripting language should be trivial. (And with some assurance about the platform, I could justify taking time to write tests for the tool itself.)

**Workflow automation** I've selected GNUMake as it is a widely-used, long-lived, Free Software solution. That said, it may be a little onerous to use in Windows (my shallow research suggests it is not very difficult, but I haven't had the chance to verify that).

**Metadata syntax** Just like the "static" tests in `master` now, test template and "case" files both specify metadata using YAML frontmatter. There's a lot of inventing going on here, and I saw this as a good opportunity to re-use existing convention. Doing so promotes consistency in test writing and makes it easier to trace values from generated files back to their source.

**Directory structure** I've moved all current test files to `src/static/` and created a new directory for sources of generated tests: `src/cases/`. Although the files are technically tests, they are very much the subject of this project, so `src/` seems more fitting (this also makes `test/` available for future introduction of tests-for-the-tests as proposed in gh-297 and gh-306). Please remember that this does *not* effect the location of the "built" tests in the `master` branch--they will continue to exist in `test/`.

Each directory within `src/cases/` defines "case" files along with one or more directories for template classes. A case uses meta data to specify the template that it should be used to expand. For instance:

    /*---
    desc: Spread operator applied to the only argument with a valid iterator
    template: default
    ---*/

...or:

    /*---
    desc: Spread operator following other arguments when IteratorValue fails
    template: error
    ---*/

**File extension** The test cases/templates are not necessarily valid JavaScript, so `.js` seems inappropriate. Since I'm building off of prior research by @bterlson, I selected the extension `.hashes`. That also seems inappropriate because (following @littledan's [suggestion](https://github.com/tc39/test262/issues/470#issuecomment-172775743), the syntax no longer uses hash marks.

...and I'm certainly open to discussing any other decisions I'm missing.

---

This patch also includes actual test cases for a few language features (destructuring binding and the spread operator). Neither of these are very thorough; I've included them as a proof-of-concept.

I should mention that I plan to use this solution (in one form or another) to author tests immediately following its acceptance. So if you can find the time to read through all this and offer some feedback (even simply, "LGTM"), then the sooner the better!

Thanks!


goyakin:
@jugglinmike, at a high level, this looks good to me. I have a few initial observations.

I doesn't look like it's easy to use `make` on Windows without setting up either Cygwin or MSYS. Since we have a few simple commands in `Makefile`, I think we should convert it to a Python script in order to reduce the number of dependencies.

Regarding the Git branches, I think it'd be better to keep everything in the `master` branch.

> I have three reasons for recommending a dev branch. First, it avoids any interruption in service for the consumers of Test262--they will continue to find every test within the test/ subdirectory of the master branch.

I don't think this is a problem. Since you're proposing to move the current test files to `src` and keep the generated files under `test`, tools running the tests won't need to be changed even when we keep everything in `master`.

> Second, it minimizes confusion for contributors--we can configure GitHub to render the dev branch by default, so contributors are less likely to accidentally edit built files.

Wouldn't this actually cause people to edit built files when they clone it to their machines since they'll see the `master` branch automatically instead of `dev`?

Even if they accidentally edit built files under `test` directory, it'll be easier to guide people to just move the files to `src` if everything is in `master` instead of first switching to the `dev` branch before doing that.

> Third, it minimizes the risk of trying this approach (at any time, we can discard the dev branch and return to editing files in master directly, forgetting that some test files were originally created as the result of a build process).

If everything is in `master`, we can still do this by just deleting the `src` directory.

As for the file extension, what do you think about using different extensions for templates and case files so that we can distinguish them when we look at the filenames? For example, `.template` for templates and `.case` for case files.

I still have a few things to go through and I'm hoping to finish them by tomorrow. The design other than the above (programming language, directory structure, etc.) looks good to me, so I don't expect there'll be anything major.

jugglinmike:
> Since we have a few simple commands in Makefile, I think we should convert it
> to a Python script in order to reduce the number of dependencies.

That's a great idea. I'll get right on it.

> Regarding the Git branches, I think it'd be better to keep everything in the
> `master` branch.

Everything you've said on this topic sounds right to me. I have a few remaining
concerns with the "`master`-only" solution:

1. The `test/` directory will not be available for automated tests. That said,
   we obviously don't have any yet, and it's not clear to me if there's any
   interest in adding them. I suppose that we could always place such tests in
   a sub-directory of `tools/`, though.
2. Another confusion common among projects that check in compiled assets is
   that, even when contributors identify the correct location for edits, they
   often mistakenly compile those assets on their own and check those in as
   well. We don't have a formal release process, so maybe we're alright with
   that (although it also increases noise for reviewers). This is mostly to
   say: there's a potential for confusion when we introduce *any* process that
   checks in built assets, so I probably shouldn't have raised it as a
   motivating factor.
3. The commit history will be noisy. We'll have commits for novel test material
   mixed in with commits of generated files. We may be able to filter by
   author, but that's only a solution if the CI machine is the sole source of
   such commits.

I have to admit some personal bias towards checking in built assets for *any*
reason, though, so that bias may be clouding my judgment here. The concerns
listed above are not major, so if you're comfortable with those trade-offs,
then I'm willing to let this go.

> As for the file extension, what do you think about using different extensions
> for templates and case files so that we can distinguish them when we look at
> the filenames? For example, `.template` for templates and `.case` for case
> files.

Works for me. Patch incoming.

Thanks for taking the time to look over all this!

jugglinmike:
@goyakin I've re-implemented the Makefile in Python and updated the file extensions as per your suggestion. I've also added some documentation to the `CONTRIBUTING.md` file. It's a little terse, but that file is growing quite long as it is... still, I'm happy to add more as you see fit.

I'll await your final confirmation before re-structuring the build/deployment process to a "`master`-only" approach.

jugglinmike:
@goyakin I've just now added some high-level acceptance tests. This will make verifying bugs and preventing regressions easier in the future.

jugglinmike:
I've just now added support for license text in template and case files. I'm assuming that we do *not* want license information in the generated files, since they are not themselves authored by a human.

littledan:
+1 to the high-level design described at the top.

littledan:
@goyakin @bterlson What do you think of this change? It looks good to me.

jugglinmike:
@goyakim I'll need to modify this approach before we move forward.

We've been discussing a solution that maintains the "source" tests in a new
directory named `src/`, which itself contains two sub-directories: `src/cases/`
(for procedurally-generated tests) and `src/static/` (for "literal" test
files). This solution exposes all tests (both "static" and built) as files
checked in to the top-level `test/` directory.

The problem here is that the vast majority of content (approaching 200
megabytes) will reside in `src/static/`, so redundant copies in `test/` would
nearly double the size of the repository. This will have a significant effect
on resource requirements for consumers (bandwidth, download time, storage
space, etc.). It probably goes without saying that this cannot be addressed by
using a shallow git clone.

I can think of three solutions to this problem:

1. Maintaining so-called "static" tests in the `test/` directory,
  "case"/"template" files in the `src/` directory, and generating test files
  into `test/`
2. A dedicated "dev" branch as discussed earlier in this thread
3. A dedicated publication channel, separate from the git repository

For `#1`, source files will exist in two different top-level directories, and
(in the case of the "static" files in `test/`) they will be located
*side-by-side* with generated files. Although we are careful to generate each
test with a note describing its provenance, I expect this will still cause
confusion (more than we discussed earlier in this thread). Mixing source files
and generated files will also make common workflow tasks like directory
"cleaning" more difficult.

I think either of the other two options would be preferable.

- With a dedicated "dev" branch, git will recognize the equivalence between
  `src/static/` on the "dev" branch and `test/` on the "master" branch. It will
  maintain only information about file name changes, keeping the repository
  (relatively) small
- With a dedicated publication channel, we can avoid these concerns altogether.
  git can technically serve as a mechanism to distribute code, but I think the
  discussion we're having now demonstrates that using it this way is an
  over-extension of responsibilities.

The shortcoming with the "dev" branch is that we would still be checking in
built files, so the more heavily we rely on test generation, the larger the
repository will become. Consumers can mitigate this by using shallow clones,
but contributors will find themselves cloning/fetching a larger and larger
repository.

My recommendation would be to use a "dev" branch for the short term. Creating a
separate publication channel will require additional infrastructure and
coordination with consumers. We can defer all of that until later, when we have
some confidence that the utility of the test generation approach warrants the
effort.

littledan:
I vote for not doing something that would require a large amount of work--a big portion of the effort here is to save work, and you've already implemented a tool which does that; we're just talking about automating executing that. It doesn't seem like it should require reorganizing everything in a time-consuming way. On the other hand, I also don't think we need to think about the resource usage of tests so much--this is just text, so doubling the size of text is still pretty small. Git logs are something I would worry more about--being able to figure out where a test came from by searching around in git is really useful.

For v8, switching to a different publication channel, if it's just a different branch on GitHub, would be really easy; I don't know about other engines. I bet they would also have a manual 'roll' process which would make this not too hard as long as you tell each of the consumers.

jugglinmike:
> On the other hand, I also don't think we need to think about the resource
> usage of tests so much--this is just text, so doubling the size of text is
> still pretty small.

I initially thought so too, but the `test/` directory is currently 200
megabytes. From the "just text" perspective, that seems very large to me. By
doubling it, we'll be approaching half a gigabyte. And possibly worse, nothing
short of a forced rebase will allow us to shrink the repository size back down.

I point this out because we've identified the ability to jettison the
generation tool as a strength of the design, but this decision cannot be undone
as easily.

> if it's just a different branch on GitHub, would be really easy

My suggestion about a new publication channel was motivated by problems of
publishing via git, no matter the branch. So this would only be appropriate if
we used another system, e.g. Amazon S3 or possibly npm. This would require a
non-trivial amount of work, which is why I recommended deferring on this path
until we had some more confidence in the test generation tool/workflow.

jugglinmike:
@leobalter and I just met with @goyakim to discuss this. Here is the approach we decided to take:

1. Keep the current tests (so called "static" files) in the `test/` directory
2. Store the input for the test generation tool ("case" and "template" files)
   in a new top-level directory named `src/`. Update the tool to output
   generated files into the `test/` directory (with the expectation that those
   files will be checked in to the repository)
3. This structure will result in generated files existing side-by-side with
   "static" files. Make the following changes to mitigate concerns regarding
   this detail:
   1. Update the test generation tool's output to include a frontmatter tag
      that documents the fact that the file was generated (e.g. `flags:
      [generated]`)
   2. Update the "clean" task to only remove files with the above flag
4. Factor out the work for automated CI-based deployment into a separate pull
   request

I will be implementing these changes, squashing the commits, and rebasing this pull request within the next day or so.

jugglinmike:
Superseded by gh-545.

I created a new pull request because this one targets the "dev" branch, which we have decided not to use (for now).

PR #484, opened 2016-02-01
Add Object.getOwnPropertyDescriptors tests

ljharb:
Tests for the stage 3 [proposal](https://github.com/tc39/proposal-object-getownpropertydescriptors) for `Object.getOwnPropertyDescriptors`.

### DIFF
--- a/harness/proxyTrapsHelper.js
+++ b/harness/proxyTrapsHelper.js
@@ -3,7 +3,7 @@ function allowProxyTraps(overrides) {
         return function () { throw new Test262Error(msg); };
     }
     if (!overrides) { overrides = {}; }
-    var traps = {
+    return {
         getPrototypeOf: overrides.getPrototypeOf || throwTest262Error('[[GetPrototypeOf]] trap called'),
         setPrototypeOf: overrides.setPrototypeOf || throwTest262Error('[[SetPrototypeOf]] trap called'),
         isExtensible: overrides.isExtensible || throwTest262Error('[[IsExtensible]] trap called'),
--- a/harness/proxyTrapsHelper.js
+++ b/harness/proxyTrapsHelper.js
@@ -1,7 +1,7 @@
-function throwTest262Error(msg) {
-    return function () { throw new Test262Error(msg); };
-}
 function allowProxyTraps(overrides) {
+    function throwTest262Error(msg) {
+        return function () { throw new Test262Error(msg); };
+    }
     if (!overrides) { overrides = {}; }
     var traps = {
         getPrototypeOf: overrides.getPrototypeOf || throwTest262Error('[[GetPrototypeOf]] trap called'),
--- a/harness/proxyTrapsHelper.js
+++ b/harness/proxyTrapsHelper.js
@@ -1,0 +1,22 @@
+function throwTest262Error(msg) {
+    return function () { throw new Test262Error(msg); };
+}
+function allowProxyTraps(overrides) {
+    if (!overrides) { overrides = {}; }
+    var traps = {
+        getPrototypeOf: overrides.getPrototypeOf || throwTest262Error('[[GetPrototypeOf]] trap called'),
+        setPrototypeOf: overrides.setPrototypeOf || throwTest262Error('[[SetPrototypeOf]] trap called'),
+        isExtensible: overrides.isExtensible || throwTest262Error('[[IsExtensible]] trap called'),
+        preventExtensions: overrides.preventExtensions || throwTest262Error('[[PreventExtensions]] trap called'),
+        getOwnPropertyDescriptor: overrides.getOwnPropertyDescriptor || throwTest262Error('[[GetOwnProperty]] trap called'),
+        has: overrides.has || throwTest262Error('[[HasProperty]] trap called'),
+        get: overrides.get || throwTest262Error('[[Get]] trap called'),
+        set: overrides.set || throwTest262Error('[[Set]] trap called'),
+        deleteProperty: overrides.deleteProperty || throwTest262Error('[[Delete]] trap called'),
+        defineProperty: overrides.defineProperty || throwTest262Error('[[DefineOwnProperty]] trap called'),
+        enumerate: throwTest262Error('[[Enumerate]] trap called: this trap has been removed'),
+        ownKeys: overrides.ownKeys || throwTest262Error('[[OwnPropertyKeys]] trap called'),
+        apply: overrides.apply || throwTest262Error('[[Call]] trap called'),
+        construct: overrides.construct || throwTest262Error('[[Construct]] trap called')
+    };
+}
--- a/test/built-ins/Object/getOwnPropertyDescriptors/duplicate-keys.js
+++ b/test/built-ins/Object/getOwnPropertyDescriptors/duplicate-keys.js
@@ -1,0 +1,31 @@
+// Copyright (C) 2016 Jordan Harband. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Object.getOwnPropertyDescriptors on a proxy with duplicate ownKeys should work
+esid: pending
+author: Jordan Harband
+features: [Proxy]
+---*/
+
+var i = 0;
+var descriptors = [
+  { enumerable: false, value: "A1", writable: true, configurable: true },
+  { enumerable: true, value: "A2", writable: true, configurable: true }
+];
+var log = '';
+var proxy = new Proxy({}, {
+  ownKeys() {
+    log += 'ownKeys|';
+    return ['DUPLICATE', 'DUPLICATE', 'DUPLICATE'];
+  },
+  getOwnPropertyDescriptor(t, name) {
+    log += 'getOwnPropertyDescriptor:' + name + '|';
+    return descriptors[i++];
+  }
+});
+
+var result = Object.getOwnPropertyDescriptors(proxy);
+assert.sameValue(result.hasOwnProperty('DUPLICATE'), true);
+assert.sameValue(result.DUPLICATE, undefined);
+assert.sameValue(log, 'ownKeys|getOwnPropertyDescriptor:DUPLICATE|getOwnPropertyDescriptor:DUPLICATE|getOwnPropertyDescriptor:DUPLICATE');
--- a/test/built-ins/Object/getOwnPropertyDescriptors/duplicate-keys.js
+++ b/test/built-ins/Object/getOwnPropertyDescriptors/duplicate-keys.js
@@ -3,7 +3,7 @@
 
 /*---
 description: Object.getOwnPropertyDescriptors on a proxy with duplicate ownKeys should work
-esid: pending
+id: pending
 author: Jordan Harband
 features: [Proxy]
 ---*/
--- a/test/built-ins/Object/getOwnPropertyDescriptors/inherited-properties-omitted.js
+++ b/test/built-ins/Object/getOwnPropertyDescriptors/inherited-properties-omitted.js
@@ -3,16 +3,17 @@
 
 /*---
 description: Object.getOwnPropertyDescriptors does not see inherited properties.
-es7id: pending
+esid: pending
 author: Jordan Harband
 ---*/
 
-var F = function G() {};
+var F = function () {};
 F.prototype.a = {};
 F.prototype.b = {};
 
 var f = new F();
-f.b = {}; // shadow the prototype
+var bValue = {};
+f.b = bValue; // shadow the prototype
 Object.defineProperty(f, 'c', {
   enumerable: false,
   configurable: true,
@@ -28,15 +29,15 @@ assert.sameValue(!!result.c, true, 'c has a descriptor');
 assert.sameValue(result.b.enumerable, true, 'b is enumerable');
 assert.sameValue(result.b.configurable, true, 'b is configurable');
 assert.sameValue(result.b.writable, true, 'b is writable');
-assert.sameValue(result.b.value, f.b, 'b’s value is f.b');
+assert.sameValue(result.b.value, bValue, 'b’s value is `bValue`');
 
 assert.sameValue(result.c.enumerable, false, 'c is enumerable');
 assert.sameValue(result.c.configurable, true, 'c is configurable');
 assert.sameValue(result.c.writable, false, 'c is writable');
-assert.sameValue(result.c.value, f.c, 'c’s value is f.c');
+assert.sameValue(result.c.value, f.c, 'c’s value is `f.c`');
 
 assert.sameValue(
   Object.keys(result).length,
-  Object.getOwnPropertyNames(f).length,
+  2,
   'result has same number of own property names as f'
 );
--- a/test/built-ins/Object/getOwnPropertyDescriptors/inherited-properties-omitted.js
+++ b/test/built-ins/Object/getOwnPropertyDescriptors/inherited-properties-omitted.js
@@ -1,0 +1,42 @@
+// Copyright (C) 2016 Jordan Harband. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Object.getOwnPropertyDescriptors does not see inherited properties.
+es7id: pending
+author: Jordan Harband
+---*/
+
+var F = function G() {};
+F.prototype.a = {};
+F.prototype.b = {};
+
+var f = new F();
+f.b = {}; // shadow the prototype
+Object.defineProperty(f, 'c', {
+  enumerable: false,
+  configurable: true,
+  writable: false,
+  value: {}
+}); // solely an own property
+
+var result = Object.getOwnPropertyDescriptors(f);
+
+assert.sameValue(!!result.b, true, 'b has a descriptor');
+assert.sameValue(!!result.c, true, 'c has a descriptor');
+
+assert.sameValue(result.b.enumerable, true, 'b is enumerable');
+assert.sameValue(result.b.configurable, true, 'b is configurable');
+assert.sameValue(result.b.writable, true, 'b is writable');
+assert.sameValue(result.b.value, f.b, 'b’s value is f.b');
+
+assert.sameValue(result.c.enumerable, false, 'c is enumerable');
+assert.sameValue(result.c.configurable, true, 'c is configurable');
+assert.sameValue(result.c.writable, false, 'c is writable');
+assert.sameValue(result.c.value, f.c, 'c’s value is f.c');
+
+assert.sameValue(
+  Object.keys(result).length,
+  Object.getOwnPropertyNames(f).length,
+  'result has same number of own property names as f'
+);
--- a/test/built-ins/Object/getOwnPropertyDescriptors/inherited-properties-omitted.js
+++ b/test/built-ins/Object/getOwnPropertyDescriptors/inherited-properties-omitted.js
@@ -3,7 +3,7 @@
 
 /*---
 description: Object.getOwnPropertyDescriptors does not see inherited properties.
-esid: pending
+id: pending
 author: Jordan Harband
 ---*/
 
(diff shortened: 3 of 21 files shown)


test/built-ins/Object/getOwnPropertyDescriptors/inherited-properties-omitted.js
```diff
@@ -0,0 +1,42 @@
+// Copyright (C) 2016 Jordan Harband. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Object.getOwnPropertyDescriptors does not see inherited properties.
+es7id: pending
+author: Jordan Harband
+---*/
+
+var F = function G() {};
```
jugglinmike: The `G` identifier is not relevant for this test

test/built-ins/Object/getOwnPropertyDescriptors/inherited-properties-omitted.js
```diff
@@ -0,0 +1,42 @@
...
+
+var result = Object.getOwnPropertyDescriptors(f);
+
+assert.sameValue(!!result.b, true, 'b has a descriptor');
+assert.sameValue(!!result.c, true, 'c has a descriptor');
+
+assert.sameValue(result.b.enumerable, true, 'b is enumerable');
+assert.sameValue(result.b.configurable, true, 'b is configurable');
+assert.sameValue(result.b.writable, true, 'b is writable');
+assert.sameValue(result.b.value, f.b, 'b’s value is f.b');
```
jugglinmike: As written, it's possible to pass this assertion by returning the "shadowed"
value of `b`. It seems highly unlikely that an implementor would get
"enumerable", "configurable", and "writable" correct but make an error on
"value", to be sure. All the same (to be extra cautious), I'd recommend
creating a binding for the "shadowing" value and expressing this expectation
with that reference.

test/built-ins/Object/getOwnPropertyDescriptors/symbols-omitted.js
```diff
@@ -0,0 +1,37 @@
+// Copyright (C) 2016 Jordan Harband. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Object.getOwnPropertyDescriptors does not include Symbol keys.
```
jugglinmike: Shouldn't this read, "Object.getOwnPropertyDescriptors *includes* Symbol keys." ?

test/built-ins/Object/getOwnPropertyDescriptors/tamper-with-global-object.js
```diff
@@ -0,0 +1,21 @@
...
+author: Jordan Harband
+---*/
+
+function fakeObject() {
+    $ERROR('The overriden version of Object was called!');
+}
+fakeObject.getOwnPropertyDescriptors = Object.getOwnPropertyDescriptors;
+fakeObject.keys = Object.keys;
+
+var global = Function('return this;')();
```
jugglinmike: This logic is encapsulated in [the `fnGlobalObject.js` helper function](https://github.com/tc39/test262/blob/fbad34959ffedb02c5b8c040e0f992a6fc006f36/harness/fnGlobalObject.js)

jugglinmike:
Looking good, @ljharb! I left some in-line notes for minor alterations. One patch-wide suggestion concerns the `es7id` tag. Given that this is not targeting ES7 (and that [we're moving to a version-independent identification scheme](https://github.com/tc39/test262/issues/477)), these tests should probably specify `esid: pending` instead.

jugglinmike:
Oh, and one more test! For 

>     [...]
>     3. Let descriptors be ! ObjectCreate( %ObjectPrototype%).
>     [...]
>     5. Return descriptors. 

We're want something like,

```js
assert.sameValue(
  Object.getPrototypeOf(Object.getOwnPropertyDescriptors({})),
  Object.prototype
);
```

test/built-ins/Object/getOwnPropertyDescriptors/symbols-omitted.js
ljharb: oops sorry, left over from a copypaste :-) will fix

test/built-ins/Object/getOwnPropertyDescriptors/inherited-properties-omitted.js
ljharb: Good call, I'll store it in a local reference.

test/built-ins/Object/getOwnPropertyDescriptors/tamper-with-global-object.js
ljharb: Thanks - can I just refer to `fnGlobalObject`, or do I need to add a directive to bring it into scope?

ljharb:
Thanks, I'll make the requested changes

test/built-ins/Object/getOwnPropertyDescriptors/tamper-with-global-object.js
jugglinmike: Good question. I sometimes forget myself. [Judging from existing tests](https://github.com/tc39/test262/blob/fbad34959ffedb02c5b8c040e0f992a6fc006f36/test/built-ins/RegExp/prototype/Symbol.replace/fn-invoke-this-no-strict.js#L19), it looks like you'll need to use the `include` tag

ljharb: nvm, i found the includes directive

ljharb:
Updated!

caitp:
here's one you could try:

```js
var i = 0;
var descriptors = [
  { enumerable: false, value: "A1", writable: true, configurable: true },
  { enumerable: true, value: "A2", writable: true, configurable: true }
];
var log = "";
var proxy = new Proxy({}, {
  ownKeys() {
    log += "ownKeys|";
    return ["DUPLICATE", "DUPLICATE", "DUPLICATE"];
  },
  getOwnPropertyDescriptor(t, name) {
    log += "getOwnPropertyDescriptor:" + name + "|";
    return descriptors[i++];
  }
});

var result = Object.getOwnPropertyDescriptors(proxy);
assert.sameValue(result.hasOwnProperty("DUPLICATE"), true);
assert.sameValue(result.DUPLICATE, undefined);
assert.sameValue(log, "ownKeys|getOwnPropertyDescriptor:DUPLICATE|getOwnPropertyDescriptor:DUPLICATE|getOwnPropertyDescriptor:DUPLICATE");
```

This exposes some currently broken code in v8, and from IRC discussions, v8 will probably have to fix its bug :d

ljharb:
@jugglinmike just noticed your edit - IDs updated to "id".

test/built-ins/Object/getOwnPropertyDescriptors/inherited-properties-omitted.js
```diff
@@ -0,0 +1,43 @@
...
+assert.sameValue(result.b.value, bValue, 'b’s value is `bValue`');
+
+assert.sameValue(result.c.enumerable, false, 'c is enumerable');
+assert.sameValue(result.c.configurable, true, 'c is configurable');
+assert.sameValue(result.c.writable, false, 'c is writable');
+assert.sameValue(result.c.value, f.c, 'c’s value is `f.c`');
+
+assert.sameValue(
+  Object.keys(result).length,
+  Object.getOwnPropertyNames(f).length,
```
goyakin: @ljharb, what do you think about using the constant `2` here instead of `Object.getOwnPropertyNames`? That way, we won't depend on the correctness of the function.

test/built-ins/Object/getOwnPropertyDescriptors/observable-operations.js
```diff
@@ -0,0 +1,45 @@
...
+};
+var check = {
+  get: function (target, propertyKey, receiver) {
+    assert(propertyKey in target, "handler check: " + propertyKey);
+    return target[propertyKey];
+  }
+};
+var proxy = new Proxy(object, new Proxy(handler, check));
+var result = Object.getOwnPropertyDescriptors(proxy);
+assert.sameValue(log, "|ownKeys|getOwnPropertyDescriptor:a|getOwnPropertyDescriptor:b|getOwnPropertyDescriptor:c", log);
```
goyakin: Should the message be the String `"log"` instead of the Object `log`?

test/built-ins/Object/getOwnPropertyDescriptors/observable-operations.js
```diff
@@ -0,0 +1,45 @@
...
+  },
+  deleteProperty: function (oTarget, sKey) {
+    throw new Test262Error('properties should not be deleted');
+  },
+  defineProperty: function (oTarget, sKey, oDesc) {
+    throw new Test262Error('properties should not be defined');
+  },
+  set: function (oTarget, sKey, vValue) {
+    throw new Test262Error('properties should not be assigned');
+  }
```
goyakin: Should we check for other Proxy methods, e.g., `preventExtensions`, too? We don't need to define them as properties on the handler. Instead, we can log all property `Get`s via the `check` object and assert the log at the end.

test/built-ins/Object/getOwnPropertyDescriptors/inherited-properties-omitted.js
ljharb: sure, np

test/built-ins/Object/getOwnPropertyDescriptors/observable-operations.js
ljharb: no, `log` is the stateful closed-over string that's defined higher up in the file

test/built-ins/Object/getOwnPropertyDescriptors/observable-operations.js
ljharb: Can you elaborate on what you mean here? I could totally add `preventExtensions` (it might be nice to have a generic helper to create a proxy that by default disallows all traps)

test/built-ins/Object/getOwnPropertyDescriptors/observable-operations.js
goyakin: I meant the third argument which is the assertion message :smile:

ljharb: isn't that the only way to make the actual `log` content show up in the failure output? if it's redundant than sure, the string "log" is perfect

goyakin: Expected and actual values are appended to the message. If you look at the definition of `assert.sameValue`, it has:

    message += 'Expected SameValue(«' + String(actual) + '», «' + String(expected) + '») to be true'

test/built-ins/Object/getOwnPropertyDescriptors/observable-operations.js
goyakin: var checkLog = '';
    var check = {
        get: function (target, propertyKey, receiver) {
            checkLog += 'propertyKey|';
            // Rest is the same as what you currently have.
        }
    };

    // Run Object.getOwnPropertyDescriptors
    // ...

    assert.sameValue(checkLog, '...');

As long as `checkLog` has only `getOwnPropertyDescriptor` and `ownKeys`, and in the correct order, I don't think we would need to define `deleteProperty` or any unused proxy handler method and still cover all of them.

ljharb: if it called one of the other traps tho, nothing would be appended to `checkLog` - how would i know if `deleteProperty` was called for example, since it would fall through to the object?

goyakin: When `proxy.deleteProperty(...)` is called, wouldn't `check`'s `get` function be called beforehand to retrieve `deleteProperty` and we'd have it in `checkLog`?

ljharb: no, "deleteProperty" isn't a method on proxy, it's what gets called with i do `delete obj[key]` - which doesn't invoke any getters

goyakin: You're right.

I think your suggestion about creating a helper for proxy creation is a good idea. Could you add one and use it here?

ljharb: i'll take a look at the existing helpers and give it a shot

jugglinmike: We also maintain tests for the assertions to try to avoid false positives:

https://github.com/tc39/test262/tree/d549225f9151f1ea46bfcbc0b7b1447875da6b50/test/harness

ljharb: Currently, things that require Proxy, Symbol, etc use `features: [Proxy]` to indicate that the test should only be run when Proxy is available. If I make a helper, would consumers of that helper have to remember to add `features: [Proxy]`? Or, could the helper somehow include that?

goyakin: Yes, they would need to add `features: [Proxy]` to each test file.

ljharb: ok, then instead of making a helper that provides a Proxy, I'll make one that provides an object with pre-set proxy traps that can be selectively overridden :+1:

ljharb: I've updated the PR with a helper, including tests for the helper. Please review!

ljharb:
@goyakin @jugglinmike friendly ping! :-D

jugglinmike:
> Since we use the list of traps in multiple files, I think we should create a helper to enumerate them. This would be similar to testWithTypedArrayConstructors defined in harness/testTypedArray.js. We can use it like:
>
> ```js
> testWithTraps(assertTrapThrows);
> ```

I disagree. These tests are themselves for the a helper function. Introducing a helper to test a helper seems a little too abstract (and would necessitate tests for the helper for the tests for the helper :P).

test/harness/proxytrapshelper-overrides.js
```diff
@@ -0,0 +1,71 @@
...
+    set: function () {},
+    deleteProperty: function () {},
+    defineProperty: function () {},
+    enumerate: function () {},
+    ownKeys: function () {},
+    apply: function () {},
+    construct: function () {},
+});
+
+function assertTrapSucceeds(trap) {
```
jugglinmike: It's not enough to check that the trap didn't throw because we also want to be
sure that the specified function was invoked. If you want to enforce a loose
contract, you could assert this with a counter variable for each provided trap.
If you don't mind stricter tests (or just want to avoid all those counters),
you could

```diff
-var traps = allowProxyTraps({
-var input = {
   getPrototypeOf: function () {},
   setPrototypeOf: function () {},

 ...
   apply: function () {},
   construct: function () {},
-});
+};
+var traps = allowProxyTraps(input);
```

...and then

```diff
-function assertTrapSucceeds(trap) {
+function assertCustom(trap) {
+  if (input[trap] !== traps[trap]) {
+    throw new Test262Error('Expected custom trap');
+  }
+}
```

ljharb: That's a fair point - although I'd say it *is* enough here since there's a parallel test that asserts that it does throw when an override isn't defined. I'll update with this change.

ljharb: Updated!

goyakin:
> I disagree. These tests are themselves for the a helper function. Introducing a helper to test a helper seems a little too abstract (and would necessitate tests for the helper for the tests for the helper :P).

I can see how testing a helper for a helper might be weird, but we don't have to have the additional tests. We enumerate the traps manually in two places and need to manually verify that all of them are covered in both anyway.

Moreover, when a new trap is added, we won't have to look for all the files where all traps are used if we have the helper.

We currently need all traps only in two files, so this isn't much of a problem though.

jugglinmike:
Looks good to me!

harness/proxyTrapsHelper.js
```diff
@@ -0,0 +1,22 @@
...
+        has: overrides.has || throwTest262Error('[[HasProperty]] trap called'),
+        get: overrides.get || throwTest262Error('[[Get]] trap called'),
+        set: overrides.set || throwTest262Error('[[Set]] trap called'),
+        deleteProperty: overrides.deleteProperty || throwTest262Error('[[Delete]] trap called'),
+        defineProperty: overrides.defineProperty || throwTest262Error('[[DefineOwnProperty]] trap called'),
+        enumerate: throwTest262Error('[[Enumerate]] trap called: this trap has been removed'),
+        ownKeys: overrides.ownKeys || throwTest262Error('[[OwnPropertyKeys]] trap called'),
+        apply: overrides.apply || throwTest262Error('[[Call]] trap called'),
+        construct: overrides.construct || throwTest262Error('[[Construct]] trap called')
+    };
```
goyakin: We need to return `traps` from the function :wink:

test/harness/proxytrapshelper-overrides.js
```diff
@@ -0,0 +1,74 @@
...
+    if (typeof traps[trap] !== 'function') {
+        throw new Test262Error('trap ' + trap + ' is not a function');
+    }
+    if (traps[trap] !== overrides[trap]) {
+        throw new Test262Error('trap ' + trap + ' was not overriden in allowProxyTraps');
+    }
+    var threw = false;
+    try {
+        traps[trap]();
+        threw = true;
```
goyakin: `threw` should be set inside the `catch` block instead.

ljharb: whoops, good catch, thanks

harness/proxyTrapsHelper.js
ljharb: whoops again :-) thanks

ljharb:
Thank you! Onward towards [stage 4](https://github.com/tc39/proposal-object-getownpropertydescriptors/issues/9)!

test/built-ins/Object/getOwnPropertyDescriptors/symbols-included.js
```diff
@@ -0,0 +1,37 @@
...
+var nonEnumSym = Symbol('nonenum');
+var symValue = Symbol('value');
+
+var obj = { key: symValue };
+obj[enumSym] = value;
+Object.defineProperty(obj, nonEnumSym, { enumerable: false, value: value });
+
+var result = Object.getOwnPropertyDescriptors(obj);
+
+assert.sameValue(Object.keys(result).length, 3, 'obj has 3 descriptors');
```
caitp: Object.keys() does not include Symbols, only Strings --- so this doesn't really work. See https://build.chromium.org/p/tryserver.v8/builders/v8_linux64_asan_rel/builds/14411/steps/Test262%20-%20no%20variants/logs/symbols-included/text

anba: #518 fixes this issue.

test/built-ins/Object/getOwnPropertyDescriptors/duplicate-keys.js
```diff
@@ -0,0 +1,31 @@
...
+  getOwnPropertyDescriptor(t, name) {
+    log += 'getOwnPropertyDescriptor:' + name + '|';
+    return descriptors[i++];
+  }
+});
+
+var result = Object.getOwnPropertyDescriptors(proxy);
+assert.sameValue(result.hasOwnProperty('DUPLICATE'), true);
+assert.sameValue(result.DUPLICATE, undefined);
+assert.sameValue(log, 'ownKeys|getOwnPropertyDescriptor:DUPLICATE|getOwnPropertyDescriptor:DUPLICATE|getOwnPropertyDescriptor:DUPLICATE');
```
caitp: Missing a trailing `|` at the end of the expectation string here!

test/built-ins/Object/getOwnPropertyDescriptors/symbols-included.js
caitp: alright, leave a note on my PR I guess

test/built-ins/Object/getOwnPropertyDescriptors/duplicate-keys.js
ljharb: whoops, thanks

PR #485, opened 2016-02-01
Add tests for _TypedArray_ constructors

leobalter:


test/built-ins/TypedArrays/buffer-argument-byteoffset-throws-from-modulo-element-size.js
```diff
@@ -0,0 +1,31 @@
...
+  Throws a RangeError if ToInteger(byteOffset) modulo elementSize is not 0
+info: >
+  22.2.4.5 TypedArray ( buffer [ , byteOffset [ , length ] ] )
+
+  This description applies only if the TypedArray function is called with at
+  least one argument and the Type of the first argument is Object and that
+  object has an [[ArrayBufferData]] internal slot.
+
+  ...
+  9. If offset modulo elementSize ≠ 0, throw a RangeError exception.
```
jugglinmike: This is actually step 10 (the danger of tracking a moving target)

test/built-ins/TypedArrays/called-with-length-return-object.js
```diff
@@ -0,0 +1,35 @@
...
+  defaultProto [ , length ])
+
+  ...
+  7. Return obj
+includes: [testTypedArray.js]
+---*/
+
+testWithTypedArrayConstructors(function(TA) {
+  var typedArray = new TA(4);
+  var bytesPerElement = TA.BYTES_PER_ELEMENT;
```
jugglinmike: `bytesPerElement` is not used in this test

test/built-ins/TypedArrays/called-with-length-return-object.js
```diff
@@ -0,0 +1,35 @@
...
+  var length = typedArray.length;
+
+  assert.sameValue(length, 4, "length");
+  assert.sameValue(
```
jugglinmike: For consistency with the other tests, I'd include an assertion for the value of the instance's `constructor` property

test/built-ins/TypedArrays/called-with-typedarray-return-new-instance.js
```diff
@@ -0,0 +1,31 @@
...
+
+includes: [testTypedArray.js]
+---*/
+
+var length = 10;
+var typedArraySample = new Int8Array(length);
+
+testWithTypedArrayConstructors(function(TA) {
+  var typedArray = new TA(typedArraySample);
+  var bytesPerElement = TA.BYTES_PER_ELEMENT;
```
jugglinmike: `bytesPerElement` is not used in this test

test/built-ins/TypedArrays/get-prototype-with-buffer-argument-returns-abrupt.js
```diff
@@ -0,0 +1,47 @@
...
+features: [Reflect]
+includes: [testTypedArray.js]
+---*/
+
+var buffer = new ArrayBuffer(8);
+
+var newTarget = function() {}.bind(null);
+Object.defineProperty(newTarget, "prototype", {
+  get() {
+    throw new Test262Error();
```
jugglinmike: Nice one!

test/built-ins/TypedArrays/object-argument-length-excessive-throws.js
```diff
@@ -0,0 +1,30 @@
...
+  This description applies only if the TypedArray function is called with at
+  least one argument and the Type of the first argument is Object and that
+  object does not have either a [[TypedArrayName]] or an [[ArrayBufferData]]
+  internal slot.
+
+  ...
+  6. Perform ? AllocateTypedArrayBuffer(O, len).
+  ...
+includes: [testTypedArray.js]
+features: [Symbol]
```
jugglinmike: You can remove the `features` tag for this test

test/built-ins/TypedArrays/object-argument-returns.js
```diff
@@ -0,0 +1,32 @@
...
+  "3": "7",
+  "4": Symbol("1"),
+  length: 4
+};
+
+testWithTypedArrayConstructors(function(TA) {
+  var typedArray = new TA(obj);
+  assert.sameValue(typedArray.length, 4);
+  assert.sameValue(typedArray[2], 42);
+  assert.sameValue(typedArray[3], 7);
```
jugglinmike: - This seems like a good place to validate property descriptors
- We should also assert the value of properties 0, 1, and 4

test/built-ins/TypedArrays/other-typedarray-ctor-argument-buffer-ctor-species-not-ctor.js
```diff
@@ -0,0 +1,45 @@
...
+features: [Symbol.species]
+---*/
+
+var sample1 = new Int8Array();
+var sample2 = new Int16Array();
+
+testWithTypedArrayConstructors(function(TA) {
+  var sample = TA === Int8Array ? sample2 : sample1;
+  var ctor = {};
+  var o = { m() {} };
```
jugglinmike: It's not worth changing now, but in the future, I recommend storing the method name directly--`var m = { m() {} }.m;` makes the test a little more clear because the reader doesn't have to learn that the object value is just an artifact of the method creation process... or just use `var m = eval;` :)

test/built-ins/TypedArrays/same-typedarray-ctor-argument-buffer-ctor-species-prototype-abrupt.js
```diff
@@ -0,0 +1,62 @@
...
+features: [Symbol.species]
+---*/
+
+testWithTypedArrayConstructors(function(TA) {
+  var sample = new TA();
+  var ctor = {};
+
+  sample.buffer.constructor = ctor;
+
+  ctor[Symbol.species] = function(){}.bind(null);
```
jugglinmike: For consistency with `other-typedarray-ctor-argument-buffer-ctor-species-prototype-abrupt.js`, you should assert that the constructor was not called

test/built-ins/TypedArrays/other-typedarray-ctor-argument-buffer-returns-abrupt-from-constructor-value.js
```diff
@@ -0,0 +1,90 @@
...
+---*/
+
+var sample1 = new Int8Array();
+var sample2 = new Int16Array();
+
+testWithTypedArrayConstructors(function(TA) {
+  var sample = TA === Int8Array ? sample2 : sample1;
+  Object.defineProperty(sample.buffer, "constructor", {
+    get: function() {
+      return 1;
```
jugglinmike: It doesn't look like this test needs to define accessor functions--won't data properties also work?

test/built-ins/TypedArrays/buffer-argument-is-referenced.js
```diff
@@ -0,0 +1,29 @@
...
+  ...
+  15. Set O's [[ViewedArrayBuffer]] internal slot to buffer.
+  ...
+includes: [testTypedArray.js]
+---*/
+
+testWithTypedArrayConstructors(function(TA) {
+  var bpe = TA.BYTES_PER_ELEMENT;
+
+  var buffer = new ArrayBuffer(bpe * 4);
```
jugglinmike: It's not necessary to multiply by 4

test/built-ins/TypedArrays/buffer-argument-is-referenced.js
```diff
@@ -0,0 +1,29 @@
...
+
+  var ta1 = new TA(buffer);
+  var ta2 = new TA(buffer);
+
+  assert.sameValue(ta1.buffer, ta2.buffer);
```
jugglinmike: We should `assert.sameValue(ta1.buffer, buffer)` to protect against cases where the property isn't assigned at all

test/built-ins/TypedArrays/same-typedarray-ctor-argument-returns-new-cloned-typedarray.js
```diff
@@ -0,0 +1,29 @@
...
+  23. Return O.
+includes: [testTypedArray.js]
+---*/
+
+testWithTypedArrayConstructors(function(TA) {
+  var sample = new TA(7);
+  var typedArray = new TA(sample);
+
+  assert.sameValue(typedArray.length, 7);
+  assert.notSameValue(typedArray, sample);
```
jugglinmike: You should assert that the array buffers are unique as well

jugglinmike:
Hey Leo!

I've left some specific notes as in-line comments on this pull request, but I
also had some ideas that don't belong to any specific line:

- The file naming scheme is a little inconsistent. Some names are prefixed with
  descriptions of the constructor under test, but others begin with a
  description of the behavior. To make discovery easier for others, I think we
  should choose one pattern and stick with it. My strong preference is to use
  the constructor as a prefix (since it mimics the general pattern of the rest
  of Test262's hierarchy), but I'd be open to arguments for the latter.
  Consistency is the important thing.
- By testing against maximum/minimum values across *all* typed arrays, the precision of many of these tests is degraded. For instance, following code (adapted from `buffer-argument-excessive-offset-throws.js`):

  ```js
  var buffer = new ArrayBuffer(8);
  
  testWithTypedArrayConstructors(function(TA) {
      new TA(buffer, 16);
    });
  });
  ```

  Triggers the error by invoking `Int8Array` with an offset of 16, even though
  for that constructor and that buffer, an offset of 9 would prove the intended
  behavior with greater precision.

  I would recommend calculating edge cases using `TA.BYTES_PER_ELEMENT` in
  order to produce more fine-grained assertions. For instance:

  ```js
  testWithTypedArrayConstructors(function(TA) {
    var buffer = new ArrayBuffer(TA.BYTES_PER_ELEMENT);
      new TA(buffer, 2 * TA.BYTES_PER_ELEMENT);
    });
  });
  ```

  Although calculated values aren't as immediately readable as literals, the
  intent of the test is much more clear when the bounds are derived from a
  formula.
- Some ideas for additional tests:
  - `new TA(-0)` to verify that SameValueZero (and not SameValue) is being used
    internally
  - Asserting the numeric order of traversal across properties in `new
    TA(arrayLike)`
  - Tests for the CloneArrayBuffer internal operation (it's not currently
    covered, as far as I can tell)

test/built-ins/TypedArrays/buffer-argument-is-referenced.js
leobalter: I'm doing it, but keeping this assertion to ensure the buffer is not unassigned anywhere else.

test/built-ins/TypedArrays/object-argument-returns.js
leobalter: `assert.sameValue(typedArray.length, 4);` asserts there's no property 4 (obj.length === 4 on purpose), but I'll create a check to assert it returns undefined.

test/built-ins/TypedArrays/buffer-argument-is-referenced.js
jugglinmike: Oh of course! I didn't mean to suggest otherwise

leobalter: ed387f2 

Thanks for the suggestion! I amplified it to cover NaN handling.

leobalter:
e74fbbb and ed387f2 covers the inline feedback. I'm working on the remaining.

leobalter:
> Tests for the CloneArrayBuffer internal operation (it's not currently covered, as far as I can tell)

They already are, at least they're easier to find after the renaming. `typedarray-arg-same-ctor-*`

> new TA(-0) to verify that SameValueZero (and not SameValue) is being used internally

that's a good one!

> Asserting the numeric order of traversal across properties in new TA(arrayLike)

It might be covered on some of the `object-arg-*` files.

> The file naming scheme is a little inconsistent

I hope you like it now. I have to say it was way easier for me to browser the files after this change.

> I would recommend calculating edge cases using TA.BYTES_PER_ELEMENT in order to produce more fine-grained assertions

Maybe I should still improve it, but I got some changes regarding this.

---

Thanks for the review, @jugglinmike! It's ready for another (final?) round.

anba:
The negative zero tests is already covered in https://github.com/tc39/test262/blob/master/test/built-ins/TypedArray/negative-zero-byteoffset.js

leobalter:
@anba you're right, I was trying to find them on the TypedArrays folder. I'm removing this last commit.

anba:
Please feel free to move the file into the TypedArrays folder. IIRC I've added that test before the ES2016 typed array constructor changes.

leobalter:
@anba done at 75952be

goyakin:
Thanks @leobalter

PR #486, opened 2016-02-02
Add tests for [[SetPrototypeOf]] on immutable prototype exotic objects

leobalter:
Ref tc39/ecma262#308

The second commit covers the suggestion at https://github.com/tc39/ecma262/pull/308/files#r50536782 and it might be squashed or discarded regarding the discussion there.

This is an anticipated PR and needs to wait until the effective merge on the PR for tc39/ecma262.

cc @littledan

### DIFF
--- a/test/built-ins/Object/prototype/extensibility.js
+++ b/test/built-ins/Object/prototype/extensibility.js
@@ -1,0 +1,29 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+id: sec-properties-of-the-object-prototype-object
+description: >
+  Object.prototype is still extensible and may have extensions prevented
+info: >
+  19.1.3 Properties of the Object Prototype Object
+
+  The value of the [[Prototype]] internal slot of the Object prototype object is
+  null and the initial value of the [[Extensible]] internal slot is true.
+---*/
+
+assert(
+  Object.isExtensible(Object.prototype),
+  "Object.prototype is extensible"
+);
+
+assert.sameValue(
+  Object.preventExtensions(Object.prototype),
+  Object.prototype,
+  "Object.prototype may have extensions prevented"
+);
+
+assert.sameValue(
+  Object.isExtensible(Object.prototype),
+  false,
+  "Object.prototype is not extensible after a preventExtensions operation"
+);
--- a/test/built-ins/Object/prototype/proto.js
+++ b/test/built-ins/Object/prototype/proto.js
@@ -1,0 +1,14 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+id: sec-properties-of-the-object-prototype-object
+description: >
+  The value of the [[Prototype]] internal slot of Object.prototype is null
+info: >
+  19.1.3 Properties of the Object Prototype Object
+
+  The value of the [[Prototype]] internal slot of the Object prototype object is
+  null and the initial value of the [[Extensible]] internal slot is true.
+---*/
+
+assert.sameValue(Object.getPrototypeOf(Object.prototype), null);
--- a/test/built-ins/Object/prototype/setPrototypeOf-with-different-values.js
+++ b/test/built-ins/Object/prototype/setPrototypeOf-with-different-values.js
@@ -1,0 +1,52 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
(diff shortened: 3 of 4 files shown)


test/built-ins/Object/setPrototypeOf/immutable-prototype-exotic-object-same-value.js
```diff
@@ -0,0 +1,18 @@
...
+  19.1.3 Properties of the Object Prototype Object
+
+  ...
+  The Object prototype object is an immutable prototype exotic object.
+---*/
+
+var ObjProto = Object.prototype;
+
+assert.sameValue(
+  Object.setPrototypeOf(ObjProto, ObjProto)
```
jugglinmike: Won't this create a cycle in the prototype chain? (And also, the `assert.sameValue` call's argument list is missing a semicolon.)

jugglinmike: Ah, I see: it's not cyclic because of the new definition of [[SetPrototypeOf]]. But in any case, when [[SetPrototypeOf]] returns `false`, `Object.setPrototypeOf` should throw a TypeError. https://tc39.github.io/ecma262/#sec-object.setprototypeof

test/built-ins/Reflect/setPrototypeOf/immutable-prototype-exotic-object-same-value.js
```diff
@@ -0,0 +1,18 @@
...
+  19.1.3 Properties of the Object Prototype Object
+
+  ...
+  The Object prototype object is an immutable prototype exotic object.
+---*/
+
+var ObjProto = Object.prototype;
+
+assert.sameValue(
+  Reflect.setPrototypeOf(ObjProto, ObjProto)
```
jugglinmike: Missing comma here, as well

test/built-ins/Object/setPrototypeOf/immutable-prototype-exotic-object-same-value.js
leobalter: This test case will be valid only if the suggestion on https://github.com/tc39/ecma262/pull/308/files#r50536782 gets merged.

test/built-ins/Reflect/setPrototypeOf/immutable-prototype-exotic-object-same-value.js
leobalter: > missing a ~~semicolon~~comma

:) I got this typo when I removed a misplaced semicolon :+1:

leobalter: 1ffcc2d

test/built-ins/Object/setPrototypeOf/immutable-prototype-exotic-object.js
```diff
@@ -0,0 +1,19 @@
...
+  The Object prototype object is an immutable prototype exotic object.
+---*/
+
+assert.throws(TypeError, function() {
+  Object.setPrototypeOf(Object.prototype, {});
+});
+
+assert.throws(TypeError, function() {
+  Object.setPrototypeOf(Object.prototype, null);
+});
```
goyakin: @leobalter, could you add assertion messages to distinguish the two?

test/built-ins/Reflect/setPrototypeOf/immutable-prototype-exotic-object.js
```diff
@@ -0,0 +1,20 @@
...
+---*/
+
+assert.sameValue(
+  Reflect.setPrototypeOf(Object.prototype, {}),
+  false
+);
+assert.sameValue(
+  Reflect.setPrototypeOf(Object.prototype, null),
+  false
+);
```
goyakin: These ones need assertions messages too.

goyakin:
Other than the assertion messages, everything else LGTM.

littledan:
These tests look good. You might also want to write a test to ensure that `Object.prototype` isn't preventExtension'd.

leobalter:
>  You might also want to write a test to ensure that Object.prototype isn't preventExtension'd.

This made me ask myself if these tests should be placed on the Object/prototype folder, and it seems the right thing to do, as we're not testing the behavior of Object.setPrototypeOf or Reflect.setPrototypeOf, but the internal [[setPrototypeOf]] of Object.prototype. We (should) already have proper tests for the Object/Reflect methods regarding the return of an object internal method.

I'll move the tests to Object/prototype folder and include tests to assert Object.prototype is extensible, it may be preventExtension'd and [[setPrototypeOf]] does not mess with its extensibility.

leobalter:
done, with rebased and squashed commits

leobalter:
New rebase with fixes and adding a check for `Object.getPrototypeOf(Object.prototype)`

Issue #466, opened 2016-01-08
Testing Runtime Internals


jugglinmike:
There are a few aspects of the ECMAScript 2015 specification that concern implementation details which are not directly observable from the JavaScript runtime:
- Tail call optimization
- Template string caching

As far as I know, the only reliable way to test these things is through putting the runtime under severe stress and asserting that it does not crash (in the case of tail call optimization) or does crash (in the case of template string cachine). For instance, to assert tail call optimization, we might write:

``` js
function f(x) {
  if (x === 0) {
    return 0;
  }
  return f(x - 1);
}
f(9007199254740991);
```

Only runtimes which implemented proper tail calls would execute this code to completion without crashing.

Test262 already has facilities to test crashes (implemented to support negative syntax tests), but this reflects a fundamentally different approach to testing. I believe there are two considerations that need to be addressed before we can test these features.
### 1: Satisfiability

As written, tests like the above would be take an extremely long time to complete. I've selected the largest possible integer value as the number of iterations, although a smaller value would do.

The chosen value is completely arbitrary. We basically need to use a value that is "high enough" to prove that the same stack frame is being re-used but not so high as to make the test prohibitively slow. This value could be dependent on the runtime, since some systems will have less available memory than others.

The same concerns exist for testing template caching, although the bottleneck there is the heap size.

Test262 already exposes functionality that may be useful here. Many tests reference a global function named `$DONE`. [The test runner is expected to modify its expectations in the presence of such a reference](https://github.com/tc39/test262/blob/1c1a75eead413ab5152e825d3c551908667e405f/tools/packaging/test262.py#L215-L221) because these files are testing asynchronous behavior. In addition, [the behavior of this function is defined at test execution time](https://github.com/tc39/test262/blob/1c1a75eead413ab5152e825d3c551908667e405f/tools/packaging/test262.py#L301)--implementors are expected to provide the name of the host-specific printing function.

We could introduce a similar pattern for these stress tests. Global values like `$MAX_HEAP` and `$MAX_STACK` could be defined in the same way, allowing implementors to select appropriate values for their runtime.

``` js
function f(x) {
  if (x === 0) {
    return 0;
  }
  return f(x - 1);
}
f($MAX_STACK);
```

Even this approach is a little imprecise: the values cannot describe specific sizes of memory because there are no standard values for the size of JavaScript objects/stack frames. I don't think that level of precision is necessary, though. So long as implementors can specify an appropriate order of magnitude, then we can avoid excessive test execution times.
[...]

PR #487, opened 2016-02-02
Add tests for tail-call optimization

jugglinmike:
I'm using the testing strategy discussed in gh-466 to assert proper tail call optimization. As noted in that discussion, runtime behavior in the event of a stack overflow is not currently specified. The `callCount` variable in these tests is intended to guard against bizarre runtime behavior in these cases--even if the failed recursion does not throw a runtime error or crash outright, the final assertion will fail.

Unfortunately, the lack of standard behavior in the event of stack overflow makes asserting the negative case (i.e. "no optimization occurs") impossible. Many runtimes throw an error of some sort, and I think that is reasonable behavior to standardize. To that end, [I have submitted a patch to the specification](https://github.com/tc39/ecma262/pull/319), and I have tagged a full set of "negative" tests that hinge on that behavior: https://github.com/jugglinmike/test262/commit/tco-negative If the language patch is accepted, I will swoon and then submit a pull request with these "negative" tests.

### DIFF
--- a/harness/tco-helper.js
+++ b/harness/tco-helper.js
@@ -1,0 +1,4 @@
+// This defines the number of consecutive recursive function calls that must be
+// made in order to prove that stack frames are properly destroyed according to
+// ES2015 tail call optimization semantics.
+var $MAX_ITERATIONS = 100000;
--- a/test/language/expressions/call/tco-call-args.js
+++ b/test/language/expressions/call/tco-call-args.js
@@ -1,0 +1,20 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+description: Expression is a candidate for tail-call optimization.
+id: static-semantics-hasproductionintailposition
+flags: [onlyStrict]
+features: [tail-call-optimization]
+includes: [tco-helper.js]
+---*/
+
+var callCount = 0;
+(function f(n) {
+  if (n === 0) {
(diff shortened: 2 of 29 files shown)


tcare:
@jugglinmike it would be great if we reach consensus on the negative tests. Did you make a decision on how you would implement the organization strategy mentioned in #466 to avoid adding the stress of these tests? Would this be on an opt-in or opt-out basis?

The tests themselves look good.

jugglinmike:
> Did you make a decision on how you would implement the organization strategy mentioned in #466 to avoid adding the stress of these tests? Would this be on an opt-in or opt-out basis?

This patch and the one referenced for "negative" tests both denote stress tests via a frontmatter flag, meaning they are "opt out". This is the lowest-friction approach--just like with the "iteration count" value, I was motivated to use the simplest solution in the absence of any request for more. But also just like with the "iteration count" value, now that the tests are written, I'm happy to iterate if you have a case for something different.

tcare:
Thanks for the clarification. @bterlson conveyed to me that this number might be based on some data from existing implementations. In the absence of that, we might want to indicate that it will need to be tuned for the tested implementation until we find a solution for testing this completely.

I'm ok with opt-out for now. I see valid arguments for both, but I lean towards completeness by default until we have a resolution of #466 or some expectation of stress limits.

littledan:
I confirmed with other members of the V8 team that 100k recursions makes sense for reliably blowing the stack.

PR #488, opened 2016-02-03
Add tests for TypedArrays `from` and `of`

leobalter:


test/built-ins/TypedArray/from/length.js
```diff
@@ -2,7 +2,7 @@
 // This code is governed by the BSD license found in the LICENSE file.
 
 /*---
-es6id: 22.2.2.1
+id: sec-%typedarray%.from
```
rwaldron: Thanks for updating all of these

leobalter: it was necessary as they changed.

es6id is still valid and up to date if the tests are in conformance with released spec without changes on the draft or any stage 3 feat.

rwaldron:
Are these meant to replace those found in https://github.com/tc39/test262/blob/master/test/built-ins/TypedArray/ ? For example: https://github.com/tc39/test262/blob/master/test/built-ins/TypedArray/from/prop-desc.js + https://github.com/bocoup/test262/blob/typedarray-from-of/test/built-ins/TypedArrays/from/inherited.js is the same as https://github.com/bocoup/test262/blob/typedarray-from-of/test/built-ins/TypedArrays/from/prop-desc.js

test/built-ins/TypedArrays/of/argument-number-value-throws.js
```diff
@@ -0,0 +1,33 @@
...
+    c. Perform ? Set(newObj, Pk, kValue, true).
+  ...
+includes: [testTypedArray.js]
+---*/
+
+var obj1 = {
+  valueOf() {
+    return 42;
+  }
+}
```
rwaldron: semicolon

leobalter:
That's something I'm bringing to discussion here: 0a4d1d5 and 9729c8b

As we are testing the same behaviors of `from` and `of` on every step that are not tested directly on `%TypedArray%`, it seems slightly reasonable to repeat these tests on each TypedArray constructor.

The test repetition looks ugly, but I believe the real value is testing `from` and `of` on the TypedArray constructor, rather than on `%TypedArray%`.

test/built-ins/TypedArrays/of/argument-number-value-throws.js
leobalter: f1938b1

test/built-ins/TypedArrays/of/invoked-as-method.js
```diff
@@ -0,0 +1,26 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+id: sec-%typedarray%.of
+description: >
+  "of" cannot be invoked as a method of %TypedArray%
```
rwaldron: This description doesn't make sense.

```js
var uint8s = Uint8Array.of(0x100, 0x111);
uint8s; // [0, 17]

var uint8s = new Uint8Array([0x100, 0x111])
uint8s; // [0, 17]
```

leobalter:
Just want to conclude saying I'm game to use the tests on 0a4d1d5 and 9729c8b to remove the pre-existent on the `TypedArray` folder.

test/built-ins/TypedArrays/of/invoked-as-method.js
```diff
@@ -0,0 +1,26 @@
...
+  1. Let newTypedArray be ? Construct(constructor, argumentList).
+  ...
+includes: [testTypedArray.js]
+---*/
+
+testWithTypedArrayConstructors(function(TA) {
+  assert.throws(TypeError, function() {
+    TA.of();
+  });
+});
```
rwaldron: Can you explicitly cite the spec that this is testing—I don't think this should throw any errors, right? It should just produce an empty %TypedArray%

test/built-ins/TypedArrays/of/invoked-as-method.js
leobalter: oops. Fixed at 208b8c3

rwaldron:
> As we are testing the same behaviors of from and of on every step that are not tested directly on %TypedArray%, it seems slightly reasonable to repeat these tests on each TypedArray constructor.

I buy that. As long as it was intentional and for the good of thorough coverage, I'm all for it.

leobalter:
Running some tests I found this particular behavior on some runtimes:

`_TypedArray_​.from.name` is returning an empty string and `​_TypedArray_​.from.length` is returning `0`. They should be `"from"` and `1`. 

The respective tests return different values on `%TypedArray%.from`. So yes, it seems necessary to repeat the tests.

rwaldron:
> [...] it seems necessary to repeat the tests.

That's an awesome find—thanks for the follow up :)

test/built-ins/TypedArrays/from/mapfn-is-not-callable.js
```diff
@@ -0,0 +1,26 @@
...
+  ...
+  3. If mapfn was supplied and mapfn is not undefined, then
+    a. If IsCallable(mapfn) is false, throw a TypeError exception.
+  ...
+includes: [testTypedArray.js]
+features: [Symbol.iterator]
+---*/
+
+var arrayLike = {};
+Object.defineProperty(arrayLike, Symbol.iterator, {
```
jugglinmike: No need for the `Symbol.iterator` property

test/built-ins/TypedArray/from/mapfn-is-not-callable.js
```diff
@@ -0,0 +1,26 @@
...
+  ...
+  3. If mapfn was supplied and mapfn is not undefined, then
+    a. If IsCallable(mapfn) is false, throw a TypeError exception.
+  ...
+includes: [testTypedArray.js]
+features: [Symbol.iterator]
+---*/
+
+var arrayLike = {};
+Object.defineProperty(arrayLike, Symbol.iterator, {
```
jugglinmike: No need for the `Symbol.iterator` property

test/built-ins/TypedArrays/from/arylk-to-length-error.js
```diff
@@ -0,0 +1,26 @@
...
+  ...
+  7. Let len be ? ToLength(? Get(arrayLike, "length")).
+  ...
+features: [Symbol.toPrimitive]
+includes: [testTypedArray.js]
+---*/
+
+var arrayLike = { length: {} };
+
+arrayLike.length[Symbol.toPrimitive] = function() {
```
jugglinmike: You can simplify this test a bit by defining the `length` property as `{ valueOf: function() { throw new Test262Error(); } }`

test/built-ins/TypedArrays/from/mapfn-arguments.js
```diff
@@ -0,0 +1,34 @@
...
+includes: [testTypedArray.js]
+---*/
+
+var source = [42, 43, 44];
+
+testWithTypedArrayConstructors(function(TA) {
+  var results = [];
+  var mapfn = function(kValue, k) {
+    results[k] = kValue;
+    return kValue;
```
jugglinmike: I would remove the `return` statement because it isn't necessary

test/built-ins/TypedArrays/from/mapfn-arguments.js
```diff
@@ -0,0 +1,34 @@
...
+  ...
+includes: [testTypedArray.js]
+---*/
+
+var source = [42, 43, 44];
+
+testWithTypedArrayConstructors(function(TA) {
+  var results = [];
+  var mapfn = function(kValue, k) {
+    results[k] = kValue;
```
jugglinmike: This test is a little imprecise for a few reasons:

1. By using the second argument as an array key, we are unable to distinguish between string and number values. The implementation could be invoking the callback with, e.g., `"1"` instead of `1`, and this test would still pass.
2. We cannot determine the order that function calls were made
3. There is no assertion of the number of arguments

This could be addressed by storing the values of each parameter and the `arguments` length as elements of three independent arrays.

test/built-ins/TypedArrays/from/mapfn-is-not-callable.js
```diff
@@ -0,0 +1,26 @@
...
+  get: function() {
+    throw new Test262Error();
+  }
+});
+
+assert.throws(TypeError, function() {
```
jugglinmike: This assertion should be made from within a callback function passed to `testWithTypedArrayConstructors`

test/built-ins/TypedArrays/from/mapfn-this-without-thisarg-non-strict.js
```diff
@@ -0,0 +1,39 @@
...
+    ...
+    c. If mapping is true, then
+      i. Let mappedValue be ? Call(mapfn, T, « kValue, k »).
+  ...
+includes: [testTypedArray.js]
+flags: [noStrict]
+---*/
+
+var source = [42, 43];
+var global = (function() {
```
jugglinmike: It's probably best to use [the `fnGlobalObject` helper function](https://github.com/tc39/test262/blob/d549225f9151f1ea46bfcbc0b7b1447875da6b50/harness/fnGlobalObject.js) for this

test/built-ins/TypedArrays/from/new-instance-using-custom-ctor.js
```diff
@@ -0,0 +1,29 @@
...
+includes: [testTypedArray.js]
+---*/
+
+var source = [42, 43, 42];
+
+testWithTypedArrayConstructors(function(TA) {
+  var called = 0;
+
+  var ctor = function(len) {
+    called++;
```
jugglinmike: The counter should be incremented *after* the assertion. That way, if the assertion fails and the runtime swallows the completion, we'll know something went wrong.

test/built-ins/TypedArrays/of/new-instance-using-custom-ctor.js
```diff
@@ -0,0 +1,27 @@
...
+description: >
+  Return a new TypedArray using a custom Constructor
+includes: [testTypedArray.js]
+---*/
+
+testWithTypedArrayConstructors(function(TA) {
+  var called = 0;
+
+  var ctor = function(len) {
+    called++;
```
jugglinmike: The counter should be incremented *after* the assertion. That way, if the assertion fails and the runtime swallows the completion, we'll know something went wrong.

test/built-ins/TypedArrays/from/set-value-abrupt-completion.js
```diff
@@ -0,0 +1,45 @@
...
+  var mapfn = function(kValue) {
+    lastValue = kValue;
+    return kValue;
+  };
+
+  assert.throws(Test262Error, function() {
+    TA.from(source, mapfn);
+  });
+
+  assert.sameValue(lastValue, obj, "interrupted source iteration");
```
jugglinmike: Very nice

test/built-ins/TypedArrays/from/source-value-is-symbol-throws.js
```diff
@@ -0,0 +1,23 @@
...
+  ...
+includes: [testTypedArray.js]
+features: [Symbol]
+---*/
+
+var s = Symbol("1");
+
+testWithTypedArrayConstructors(function(TA) {
+  assert.throws(TypeError, function() {
+    var result = TA.from([s]);
```
jugglinmike: The variable declaration may be removed

test/built-ins/TypedArrays/of/argument-number-value-throws.js
```diff
@@ -0,0 +1,33 @@
...
+};
+var obj2 = {
+  valueOf() {
+    throw new Test262Error();
+  }
+};
+
+testWithTypedArrayConstructors(function(TA) {
+  assert.throws(Test262Error, function() {
+    TA.of(obj1, obj2);
```
jugglinmike: You might use the same pattern from `set-value-abrupt-completion.js` to ensure correct iteration cancelling: have each setter close around a `lastValue` binding, and update the invocation to read: `TA.of(obj1, obj2, obj1);`

jugglinmike:
Hey Leo! I just completed a review of the patch. I've left some comments
in-line for you, but I have some more general feedback, too:

> Running some tests I found this particular behavior on some runtimes:
>
> `_TypedArray_.from.name` is returning an empty string and
> `_TypedArray_.from.length` is returning 0. They should be `"from"` and `1`.
>
> The respective tests return different values on `%TypedArray%.from`. So yes,
> it seems necessary to repeat the tests.

It's good that you caught this, but I think @rwaldron is right when it comes to
repeating the tests for properties of the static methods. The error here
isn't that `_TypedArray_.from.name` is different than `%TypedArray%.from.name`;
it's that `_TypedArray_` defines a "own" property named `from` at all. As
written, the assertions (and resulting failures) make it harder to see that
fact.

Each Typed Array constructor inherits from the generic TypedArray constructor.
This generic constructor defines the static methods under test, so assertions
about the property values do not need to be repeated for each subclass.

So the following tests are redundant with the equivalent tests in the
`TypedArray` directory:

- test/built-ins/TypedArrays/from/name.js
- test/built-ins/TypedArrays/from/length.js
- test/built-ins/TypedArrays/of/length.js
- test/built-ins/TypedArrays/of/name.js

Although this is only *provided that each Typed Array subclass does not define
its own `from`/`of` properties*. It certainly shouldn't, but we don't have an
assertion for that. The closest we have is:

- test/built-ins/TypedArrays/from/inherited.js
- test/built-ins/TypedArrays/of/inherited.js

but by simply asserting the value as retrieved from a MemberExpression like
`Int8Array.of.name`, we can't be sure that the implementation hasn't done
something like, e.g. `Int8Array.of = TypedArray.of;`.

Since we already have the following tests:

- prototype chain of each of the Typed Array constructors (see `built-ins/TypedArrays/*/proto.js`)
- property descriptor on the TypedArray constructor (see `built-ins/TypedArray/{of,from}/prop-desc.js`)

What we really need these two `inherited.js` tests to assert is that each
constructor does *not* define an "own" property named `from`/`of`,
respectively.

---

I'm really torn about this extension to `testWithTypedArrayConstructors`.
On the one hand, it is very simple, and (despite what you might think) I
don't really like having a lot of abstraction in test logic. On the other
hand, it reduces the responsibility of the function so far as to make it
almost meaningless; you could use `Array.prototype.forEach` without much
more code.

One alternative would be to make the extension a little more domain-specific:
the second argument could simply take the value "int" or
"float". This would limit repetition and make the test files more obvious,
but it would also require a little more abstraction. I don't know if it's an
improvement, but I'm curious to hear your thoughts!

---

I also had some ideas for additional tests:

1. `from` with a sparse array
2. Boundaries. We should assert proper behavior for, for instance,
   `Int8Array.of(-129, -128, 127, 128)` and `Uint16Array.from([-1, 0, 65535, 65536])`
   You might be able to get away with some tricky `BYTES_PER_ELEMENT`-based
   math from within the `testWithTypedArrayConstructors`, but this seems like
   an acceptable case for maintaining distinct files for each subclass.

test/built-ins/TypedArrays/from/mapfn-is-not-callable.js
leobalter: it was intentional. The next step with an abrupt completion would be the getting Symbol.iterator from the source arg.

```
6. Let arrayLike be ? IterableToArrayLike(source).
```

I'm asserting the TypeError happens and it comes before the Test262Error.

test/built-ins/TypedArray/from/mapfn-is-not-callable.js
leobalter: same as https://github.com/tc39/test262/pull/488#discussion_r52700655

test/built-ins/TypedArrays/from/arylk-to-length-error.js
leobalter: that was a clone from `test/built-ins/TypedArray/from/arylk-to-length-error.js`. I changed both for consistency.

b3dcc95 and cc76879

test/built-ins/TypedArrays/from/mapfn-is-not-callable.js
leobalter: does it work for you if I add a comment above this definition?

test/built-ins/TypedArrays/from/mapfn-arguments.js
leobalter: d048eb1

test/built-ins/TypedArrays/from/mapfn-arguments.js
leobalter: 8f5408f

test/built-ins/TypedArrays/of/argument-number-value-throws.js
leobalter: I ended up using the valueOf call on the lack of a mapfn.

272d58d

test/built-ins/TypedArrays/from/source-value-is-symbol-throws.js
leobalter: 272d58d

test/built-ins/TypedArrays/from/set-value-abrupt-completion.js
leobalter: thanks :)

test/built-ins/TypedArrays/of/new-instance-using-custom-ctor.js
leobalter: 272d58d

test/built-ins/TypedArrays/from/new-instance-using-custom-ctor.js
leobalter: that's brilliant! 

272d58d

test/built-ins/TypedArrays/from/mapfn-this-without-thisarg-non-strict.js
leobalter: 272d58d

test/built-ins/TypedArrays/from/mapfn-is-not-callable.js
leobalter: 272d58d

leobalter:
I fixed from the inline feedback, now it's time to run through the [last comment](https://github.com/tc39/test262/pull/488#issuecomment-180491022)

leobalter:
> What we really need these two inherited.js tests to assert is that each
constructor does not define an "own" property named from/of,
respectively.

Fixed at 8dfccae

leobalter:
> I'm really torn about this extension to testWithTypedArrayConstructors

I was as well, but I have a slight preference on not creating a specific map.

The way I changed this method allows the direct use of defined typedarrays and it let's it be explicit when calling the function.

a forEach would be easy, but I wanted to repeat the same aspects on the original method, with that I would end repeating too much for every file.

test/built-ins/TypedArrays/from/mapfn-is-not-callable.js
jugglinmike: Ah, now I understand. That's a great detail to check! But as written, it *is* a
little implicit. A comment would help, but an explicit assertion might be even
better. Like a comment, ensuring that a counter was not incremented would make
the test easier to read, but it would also make the failure more clear.

```js
assert.sameValue(callCount, 0, 'The @@iterator method should not be invoked.');
```

leobalter: 2067a78, I also included some extra assertions for different types.

leobalter: of course I missed null

leobalter:
a548ba9 has tests for the sparse array.

Some other fixes came along and I'll need to finish with the boundaries test, which I'll need an extra time to design a good strategy.

goyakin:
Merged as 76080eac08d5295658f9e8abe471b18b4e80e24d.

Leo, just FYI, I replaced `TA.hasOwnProperty("of")` with `TA.hasOwnProperty("from")` in `test/built-ins/TypedArrays/from/inherited.js` before merging.

PR #490, opened 2016-02-04
Move tests for object literals

jugglinmike:
Test262 defines tests for expression-producing syntactic forms within
the `language/expressions/` directory. Most tests for object literals
conform to this structure, but 12 such tests were added to the
`language/object-literal/` directory. Move these tests to the canonical
location for object literals.


leobalter:
LGTM

PR #491, opened 2016-02-04
Fix location based return value for Date on subclassing tests

### LINKED ISSUE #489: test262 tests need to take different executing time zones into account
The ECMAScript multiargument Date constructor http://tc39.github.io/ecma262/#sec-date-year-month-date-hours-minutes-seconds-ms takes its arguments in terms of local time . As a result, the expression `new Date(1859, 10, 24).getUTCDate()` gives 23 in Africa/Eurasia and 24 in the Americas. This affects the accuracy of certain tests, including language/statements/class/subclass/builtin-objects/Date/regular-subclassing .

leobalter:
Closes #489

After some quick searches through the repo, it seems that was the only case with this particular issue.

The tests for several Date.prototype methods needs improvement, as most of them does not have any tests for returned values.

### DIFF
--- a/test/language/statements/class/subclass/builtin-objects/Date/regular-subclassing.js
+++ b/test/language/statements/class/subclass/builtin-objects/Date/regular-subclassing.js
@@ -20,10 +20,10 @@ info: >
 
 class D extends Date {}
 
-var d1 = new D(1859, '10', 24);
-assert.sameValue(d1.getUTCFullYear(), 1859);
-assert.sameValue(d1.getUTCMonth(), 10);
-assert.sameValue(d1.getUTCDate(), 24);
+var d1 = new D(1859, '10', 24, 11);
+assert.sameValue(d1.getFullYear(), 1859);
+assert.sameValue(d1.getMonth(), 10);
+assert.sameValue(d1.getDate(), 24);
 
 var d2 = new D(-3474558000000);
 assert.sameValue(d2.getUTCFullYear(), 1859);
@@ -32,6 +32,6 @@ assert.sameValue(d2.getUTCDate(), 24);
 
 var d3 = new D();
 var d4 = new Date();
-assert.sameValue(d3.getUTCFullYear(), d4.getUTCFullYear());
-assert.sameValue(d3.getUTCMonth(), d4.getUTCMonth());
-assert.sameValue(d3.getUTCDate(), d4.getUTCDate());
+assert.sameValue(d3.getFullYear(), d4.getFullYear());
+assert.sameValue(d3.getMonth(), d4.getMonth());
+assert.sameValue(d3.getDate(), d4.getDate());


test/language/statements/class/subclass/builtin-objects/Date/regular-subclassing.js
```diff
@@ -20,18 +20,18 @@ info: >
...
+assert.sameValue(d1.getMonth(), 10);
+assert.sameValue(d1.getDate(), 24);
 
 var d2 = new D(-3474558000000);
-assert.sameValue(d2.getUTCFullYear(), 1859);
-assert.sameValue(d2.getUTCMonth(), 10);
-assert.sameValue(d2.getUTCDate(), 24);
+assert.sameValue(d2.getFullYear(), 1859);
+assert.sameValue(d2.getMonth(), 10);
+assert.sameValue(d2.getDate(), 24);
```
goyakin: These should stay as their UTC variants since the single argument constructor interprets the argument in UTC.

test/language/statements/class/subclass/builtin-objects/Date/regular-subclassing.js
```diff
@@ -20,18 +20,18 @@ info: >
...
 
 var d3 = new D();
 var d4 = new Date();
-assert.sameValue(d3.getUTCFullYear(), d4.getUTCFullYear());
-assert.sameValue(d3.getUTCMonth(), d4.getUTCMonth());
-assert.sameValue(d3.getUTCDate(), d4.getUTCDate());
+assert.sameValue(d3.getFullYear(), d4.getFullYear());
+assert.sameValue(d3.getMonth(), d4.getMonth());
+assert.sameValue(d3.getDate(), d4.getDate());
```
goyakin: I just realized these could potentially fail if the date changes after `d3` is created but before we reach the constructor call for `d4`. Should we change these to use `getTime` instead and also include 1 or 2 second delta in the comparison?

leobalter: There's a very small change of this bring an error, if d3 and d4 are instantiated between the very moment the day changes. It's indeed a problem, but I've done this way believing it is still better than a result manipulation.

I'm calling these methods here because I want at least minimal assertions for the constructor parameters and handling with Date internals, if I could just stop time.

goyakin: If we compare the time values instead, wouldn't that be enough to verify that the Date internals are initialized correctly? Since we already use `getDate` etc. together with other variants of the constructor, we already cover how the time is parsed.

This change isn't needed for this PR by the way. It's just something to consider for the future.

test/language/statements/class/subclass/builtin-objects/Date/regular-subclassing.js
goyakin: I forgot to mention that the date assertion fails for me since `getDate` returns `23` whereas `getUTCDate` returns `24`.

test/language/statements/class/subclass/builtin-objects/Date/regular-subclassing.js
leobalter: > If we compare the time values instead, wouldn't that be enough to verify that the Date internals are initialized correctly?

Yes, I tried to make it short enough to assert the super constructor follow the arguments, especially the first ones.

I agree this might be improved on the future, finding better approaches to assert the internals initialization.

goyakin:
@leobalter I can merge this PR if you update the assertions for `d2`.

leobalter:
done!

PR #493, opened 2016-02-04
Replace es7id tags with id

leobalter:
There's a single file with a remaining es7id tag supposed to be removed on #485

### DIFF
--- a/test/built-ins/DataView/prototype/setFloat32/length.js
+++ b/test/built-ins/DataView/prototype/setFloat32/length.js
@@ -2,7 +2,7 @@
 // This code is governed by the BSD license found in the LICENSE file.
 
 /*---
-es7id: pending
+id: id: sec-dataview.prototype.setfloat32
 description: >
   DataView.prototype.setFloat32.length is 2.
 info: >
--- a/test/built-ins/DataView/prototype/setFloat32/length.js
+++ b/test/built-ins/DataView/prototype/setFloat32/length.js
@@ -2,7 +2,7 @@
 // This code is governed by the BSD license found in the LICENSE file.
 
 /*---
-id: id: sec-dataview.prototype.setfloat32
+id: sec-dataview.prototype.setfloat32
 description: >
   DataView.prototype.setFloat32.length is 2.
 info: >
--- a/test/built-ins/DataView/prototype/setInt16/length.js
+++ b/test/built-ins/DataView/prototype/setInt16/length.js
@@ -2,7 +2,7 @@
 // This code is governed by the BSD license found in the LICENSE file.
 
 /*---
-id: id: sec-dataview.prototype.setint16
+id: sec-dataview.prototype.setint16
 description: >
   DataView.prototype.setInt16.length is 2.
 info: >
--- a/test/built-ins/DataView/prototype/setInt16/length.js
+++ b/test/built-ins/DataView/prototype/setInt16/length.js
@@ -2,7 +2,7 @@
 // This code is governed by the BSD license found in the LICENSE file.
 
 /*---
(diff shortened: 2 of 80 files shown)


test/built-ins/Object/entries/exception-during-enumeration.js
```diff
@@ -2,8 +2,8 @@
 // This code is governed by the BSD license found in the LICENSE file.
 
 /*---
+id: sec-object.entries
```
goyakin: We should wait until the proposal gets merged into the draft to use the ID. This one doesn't match what's [currently in the proposal](https://github.com/tc39/proposal-object-values-entries/blob/master/index.html#L63) anyway.

test/built-ins/Object/values/exception-during-enumeration.js
```diff
@@ -2,8 +2,8 @@
 // This code is governed by the BSD license found in the LICENSE file.
 
 /*---
+id: sec-object.values
```
goyakin: Same as `Object.entries`, we should wait until the proposal gets merged into the draft to use the ID.

test/built-ins/DataView/prototype/setFloat32/length.js
```diff
@@ -2,7 +2,7 @@
 // This code is governed by the BSD license found in the LICENSE file.
 
 /*---
-es7id: pending
+id: id: sec-dataview.prototype.setfloat32
```
goyakin: Second `id` is redundant :wink:

test/built-ins/DataView/prototype/setInt16/length.js
```diff
@@ -2,7 +2,7 @@
 // This code is governed by the BSD license found in the LICENSE file.
 
 /*---
-es7id: pending
+id: id: sec-dataview.prototype.setint16
```
goyakin: Second id is redundant :wink:

test/built-ins/Object/entries/exception-during-enumeration.js
leobalter: I noticed that, but I'm following the pattern used on the current spec, as:

- https://tc39.github.io/ecma262/#sec-object.assign
- https://tc39.github.io/ecma262/#sec-object.create
- https://tc39.github.io/ecma262/#sec-object.defineproperties
- ...

My suggestion is that we should keep it, I may open a PR on the proposal you linked **or** change the ids to `pending` until they are merged on the spec.

test/built-ins/DataView/prototype/setInt16/length.js
leobalter: 428d1ae

oops

test/built-ins/Object/entries/exception-during-enumeration.js
goyakin: It actually makes sense to use section IDs in tests even at Stage 3 since the spec text should be near complete at that point, including the section IDs.

I talked to @bterlson and he is going to make sure that section IDs are correct in the spec starting at Stage 3 :smile:

goyakin:
Merged as 428418d5c7556aa1302ae40ec141f6ccc74bc02b. Thanks Leo for the cleanup :smile:

I made some changes before merging:

  1. Prefixed the section ID in `test/built-ins/DataView/prototype/getFloat64/length.js` with `id:` since it was missing.
  2. Ignored the change in `test/built-ins/TypedArray/invoked-as-ctor-with-arguments.js` since that file has been deleted.
  3. Replaced the `es7id` tags introduced with ac7711e95f4951efc4aa81fb7fcc20ce133caaeb.

PR #494, opened 2016-02-05
Add tests for zero conversion on Array#indexOf and Array#lastIndexOf

### LINKED ISSUE #435: SameValue is too strict and used incorrectly for some tests
It looks like the patch at https://github.com/tc39/test262/commit/9ccc6639360ea28161d9eb97b40a99ef846ee2bc replaced === in a number of tests with SameValue. While it's nice to tighten things up in general, I'm not sure if the changes are all valid.

For example, in the test built-ins/Array/prototype/indexOf/15.4.4.14-5-9 , -0 is given as the starting index, and 0 is expected as the output (when the answer is found at index 0). By my reading of the spec, -0 would be the right answer (though it's a little ambiguous since the font makes it look like we left float-world and we're in math-land). Would it make sense to change the SameValue test back to ===, or use -0 in this particular test?

leobalter:
Closes #435

### DIFF
--- a/test/built-ins/Array/prototype/indexOf/fromindex-zero-conversion.js
+++ b/test/built-ins/Array/prototype/indexOf/fromindex-zero-conversion.js
@@ -1,0 +1,16 @@
+// Copyright 2016 The V8 project authors. All rights reserved.
+// This code is governed by the license found in the LICENSE file.
+/*---
+id: sec-array.prototype.indexof
+description: >
+  Return +0 when fromIndex is -0 and return index refers to the first position
+info: >
+  22.1.3.12 Array.prototype.indexOf ( searchElement [ , fromIndex ] )
+
+  ...
+  6. If n ≥ 0, then
+    a. If n is -0, let k be +0; else let k be n.
+  ...
+---*/
+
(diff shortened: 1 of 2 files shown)


anba:
Isn't this already covered by the existing tests?
https://github.com/tc39/test262/blob/master/test/built-ins/Array/prototype/indexOf/15.4.4.14-5-9.js#L11
https://github.com/tc39/test262/blob/master/test/built-ins/Array/prototype/lastIndexOf/15.4.4.15-5-9.js#L12

leobalter:
I wanted to make these tests explicitly about the -0 conversion. My first idea was to remove these tests you mentioned, but I left them as their goal is to assert the value of 'fromIndex' works as -0, not the return signal.

I know assert.sameValue handles -0 and +0 comparisons, but it's something that shouldn't be enforced as we are checking the precise values here.

anba:
Ok.

jugglinmike:
The year listed in the license is incorrect for both of these files.

goyakin:
LGTM except the year in the license as @jugglinmike pointed out.

leobalter:
My error was so 2015.

It's fixed!

PR #497, opened 2016-02-08
Enable parallel test execution in console runner

tschneidereit:
Adds a `-j`/`--workers-count` parameter to `tools/packaging/test262.py`, defaulting to `[number of cores] - 1`.

Speeds up running the test suite by about ~3x on my 4-core machine, with the SpiderMonkey shell. This could certainly be optimized more by just appending test results to per-thread lists and merging them at the end, but it's better than nothing.

### DIFF
--- a/CONSOLE-RUNNER.md
+++ b/CONSOLE-RUNNER.md
@@ -40,6 +40,7 @@ Name | Action
 -----|-------
 -h, --help | displays a brief help message
 --command=COMMAND | **required** command which invokes javascript engine to be tested
+-j, --workers-count | Number of tests to run in parallel (defaults to number of cores - 1)
 --tests=TESTS | path to the test suite; default is current directory
 --cat | don't execute tests, just print code that would be run
 --summary | generate a summary at end of execution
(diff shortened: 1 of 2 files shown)


rwaldron:
```
$ sysctl -n hw.ncpu
8

$ time tools/packaging/test262.py --workers-count 8 --command "js" language/
...
real  2m27.561s
user  7m54.752s
sys 6m20.952s


$ time tools/packaging/test262.py --command "js" language/
...
real  2m9.082s
user  7m43.122s
sys 6m3.655s
```

In both cases I did nothing else with the machine, aside from watch—am I doing something wrong?

jugglinmike:
@rwaldron I believe the second invocation would default to using 7 jobs on your machine, so the difference might be small enough to be overtaken by other random fluctuations. I wonder if you would see a more pronounced difference with `--workers-count 4` or `--workers-count 1`.

rwaldron:
@jugglinmike I actually tried 4 to begin with: 

```

$ time tools/packaging/test262.py --workers-count 4 --command "js" language/
real 2m19.231s
user  7m25.795s
sys 5m23.747s
```

rwaldron:
Let me also try 1 vs 4 vs 8

rwaldron:
>  I believe the second invocation would default to using 7 jobs on your machine

Yes, that's definitely the part I missed. Whoops!

tschneidereit:
Yeah, this isn't a work load where virtual cores add much value, so there's not much to be gained by going from 4 to 8 on a machine with 4 physical, but 8 logical cores.

rwaldron:
@jugglinmike @tschneidereit I can now completely reproduce the ~3x improvement: 

```
$ time tools/packaging/test262.py --workers-count 1 --command "js" language/
...
real	6m50.030s
user	4m5.791s
sys	2m38.871s
```


This is awesome—nice one!

jugglinmike:
@tschneidereit I'm not really familiar with systems programming, but I'm curious: what kind of work *would* benefit from virtual cores?

tschneidereit:
>
> @tschneidereit <https://github.com/tschneidereit> I'm not really familiar
> with systems programming, but I'm curious: what kind of work *would*
> benefit from virtual cores?
>
Virtual cores are only useful if multiple threads running in parallel can
improve utilization of a single CPU core. That can happen if the CPU's
pipeline is stalled less often, which is more likely if the threads have
different characteristics. As in this case all threads are very similar in
work load, hyper threading doesn't gain much. See [this wikipedia article](
https://en.wikipedia.org/wiki/Hyper-threading) for some more details.

jugglinmike:
Thanks, @tschneidereit!

PR #499, opened 2016-02-10
Reorganize tests for Annex B extensions

jugglinmike:
The "mainline" tests in Test262 are converging on a more formal
structure. Files are organized as tests for either either "language"
(e.g. syntax-driven) or "built-in" (e.g. API-driven). "Language" test
locations are themselves structured according to whether the syntactic
form under test is an Expression or a Statement.

To limit ambiguity when locating/adding tests, re-organize the tests for
Annex B extensions to match this structure.


leobalter:
LGTM.

++ on improving tests discoverability

Issue #443, opened 2015-11-24
Missing promise coverage


domenic:
V8 has a number of subtle promise bugs which @littledan and I are reasonably sure are not caught by test262 tests. (We have most test262 tests on for promises, and the ones we have disabled don't seem like they would catch these bugs.)
- https://code.google.com/p/v8/issues/detail?id=4162
- https://code.google.com/p/v8/issues/detail?id=3641
- https://code.google.com/p/v8/issues/detail?id=4539

It would be good to not only add tests to catch these bugs, but also do a bit of investigation to see which parts of the spec these bugs stress, and why test262's fairly-exhaustive coverage misses them. That might reveal a few other potentially missing tests or V8 bugs.

anba:
Just added #445 which covers built-in function tests, IIRC V8 does not pass the "*-function-nonconstructor.js" tests.

anba:
#447 adds more Promise tests, including a few which don't pass in current V8.

jugglinmike:
@domenic Following up on @anba's contribution, we further expanded test coverage
for Promises in gh-461, gh-462, gh-463, gh-464, and gh-465 based on an analysis
of the abstract operations. Here's a diagram that, despite all appearances, was
authored by a grown man:

![promise-state-machine](https://cloud.githubusercontent.com/assets/677252/12960045/24f20ec6-d007-11e5-8e9f-79be200e8c62.jpg)

Because of the amount of re-use in abstract operations, the Promise test suite
now has a fair number of closely-related tests.

[v8/4162](https://bugs.chromium.org/p/v8/issues/detail?id=4162) Describes a failure to observe two distinct behaviors:

1. accessing the `then` property of an object-coercible resolution value
2. invoking such a method if it exists

In the first case, we are asserting the property access (along with the
handling of resulting abrupt completions) with the following tests:

- [Promise/resolve-poisoned-then-immed.js](https://github.com/tc39/test262/blob/5cb97c293b4a44ee1c31961a07663993c521fb5f/test/built-ins/Promise/resolve-poisoned-then-immed.js)
- [Promise/resolve-poisoned-then-deferred.js](https://github.com/tc39/test262/blob/5cb97c293b4a44ee1c31961a07663993c521fb5f/test/built-ins/Promise/resolve-poisoned-then-deferred.js) (this matches the reported case)
[...]

PR #500, opened 2016-02-10
Promise: Add tests to disallow faulty optimization

jugglinmike:
Add tests that assert behavior when a Promise is resolved with another
Promise whose `then` method has been overridden. Because all objects
with a `then` method are treated equivalently, the presence of a
[[PromiseState]] internal slot should have no effect on program
behavior.

These tests guard against a faulty optimization originally implemented
in V8:

https://bugs.chromium.org/p/v8/issues/detail?id=3641


jugglinmike:
(This is in service of gh-443.)

PR #501, opened 2016-02-10
Promise: Add test for constructor access count

jugglinmike:
Assert that the `constructor` property of the "this" value of
`Promise.prototype.then` is accessed exactly once. This guards against
implementations where repeated access is used instead of reference
passing (possibly motivated by convenience).

Repeated access of this kind was demonstrated by V8's implementation of
the specification:

https://bugs.chromium.org/p/v8/issues/detail?id=4539


jugglinmike:
(This is in service of gh-443.)

PR #504, opened 2016-02-12
Make asynchronous test configuration explicit

jugglinmike:
For asynchronous tests, the contract between test file and test runner
is implicit: runners are expected to inspect the source code for
references to a global `$DONE` identifier.

Promote a more explicit contract between test file and test runner by
introducing a new frontmatter "tag", `async`. This brings asynchronous
test configuration in-line with other configuration mechanisms and also
provides a more natural means of test filtering.

The modifications to test files was made programatically using the
`grep` and `sed` utilities:

    $ grep "\$DONE" test/ -r --files-with-match --null | \
        xargs -0 sed -i 's/^\(flags:\s*\)\[/\1[async, /g'
    $ grep "\$DONE" test/ -rl --null | \
        xargs -0 grep -E '^flags:' --files-without-match --null | \
        xargs -0 sed -i 's/^---\*\//flags: [async]\n---*\//'

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -128,6 +128,8 @@ This tag is for boolean properties associated with the test.
 - **`raw`** - execute the test without any modification (no helpers will be
   available); necessary to test the behavior of directive prologue; implies
   `noStrict`
+- **`async`** - defer interpretation of test results until after the invocation
+  of the global $DONE` function
 
 #### features
 **features**: [list]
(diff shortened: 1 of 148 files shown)


littledan:
+1, but let's maintain the contract that $DONE has to appear in the code for these async tests also to make it easier for diverse test users.

leobalter:
@littledan the [contributing.md](https://github.com/tc39/test262/pull/504/files#diff-6a3371457528722a734f3c51d9238c13) is saying the async flag will defer the tests until $DONE is called. I believe this is solved.

LGTM.

Issue #382, opened 2015-07-16
Encoding Expected Error Type


jugglinmike:
As part of the ongoing effort to make Test262 more useful to JavaScript parser implementors (see gh-196), [we recently updated most tests for runtime errors to use the relatively new `assert.throws` helper method instead of the `negative` frontmatter](https://github.com/tc39/test262/pull/360). As mentioned there, not _all_ runtime errors could be refactored in this way; specifically, tests that concern errors thrown from the global scope must produce an uncaught error in order to be verified.

This means that the `negative` tag is still somewhat overloaded; it describes expected runtime errors, parser errors, and "early errors". If tests made this distinction explicit, then parser implementors would be able to automatically consume the Test262 tests.

We should update the frontmatter format to allow for this distinction. While we're at it, it wouldn't hurt to make a further distinction between parser errors. This was originally raised in gh-196; it may be relevant if tests are asserting parser errors generated from direct grammar application (invalid syntax; only SyntaxErrors), or generated due to explicit "Static Semantics" (which may be SyntaxErrors or ReferenceErrors).

One idea:

``` diff
 /*---
-negative: ReferenceError
+runtimeError: ReferenceError
 ---*/
 let x = x;
```

``` diff
 /*---
-negative: SyntaxError
+parserError: SyntaxError
 ---*/
 let x =;
```

``` diff
 /*---
-negative: SyntaxError
+parserError: early:SyntaxError
 ---*/
 let x, x;
```

Although the above makes room for invalid configurations that test runners (or [tests for the tests](https://github.com/tc39/test262/issues/306)) would have to guard against. Another approach would help avoid this and also curtail the proliferation of top-level configuration attributes:

``` diff
 /*---
-negative: ReferenceError
+negative:
+  stage: runtime
+  type: Referenceerror
 ---*/
 let x = x;
```

``` diff
 /*---
-negative: SyntaxError
+negative:
+  stage: parse
+  type: SyntaxError
 ---*/
 let x =;
```

``` diff
 /*---
-negative: SyntaxError
+negative:
+  stage: early
+  type: SyntaxError
 ---*/
 let x, x;
```

@bterlson @jmdyck Do either of these seem good to you? Or maybe something else?

jmdyck:
The second approach makes sense to me.

(Note that in issue #196, I distinguished between 'lexical' and 'syntactic'. If this is deemed a useful distinction, it could be accommodated with another value for the 'stage' field.)

jmdyck:
[...]

Issue #299, opened 2015-06-03
onlyStrict/noStrict: non-conformant?


jmdyck:
CONTRIBUTING.md says:
- **`onlyStrict`** - only run the test in strict mode (_not supported by the browser runner_)
- **`noStrict`** - only run the test in "sloppy" mode

This implies that non-/strictness is a mode of the implementation, but that disagrees with the spec:

> A complete ECMAScript program may be composed of both strict mode and non-strict mode ECMAScript source text units. In this case, strict mode only applies when actually executing code that is defined within a strict mode source text unit.
> In order to conform to this specification, an ECMAScript implementation must implement both the full unrestricted ECMAScript language and the strict variant of the ECMAScript language as defined by this specification. In addition, an implementation must support the combination of unrestricted and strict mode source text units into a single composite program.

So do `onlyStrict` and `noStrict` assume some non-conformant behavior of the implementation? If so, that should be stated explicitly (and the options should perhaps be deprecated/eliminated). In any event, these options need to be better explained.

bterlson:
I agree they could be documented better. Let me attempt to explain.

ECMAScript code (when parsed with the Script goal symbol) can be run in either strict mode or non strict mode at the global level. Because most tests are not sensitive to strict mode and hence should run identically in both modes, it is up to the harness to control the strictness by prepending `"use strict"` when we want to run in strict mode. Tests my specify `onlyStrict` when harnesses should only run the test under strict mode only and `noStrict` when the harness should run the test under sloppy mode only. The default (now, thanks to some impressive work from @smikes and @anba) is to run every test under both modes.

The existing browser runner doesn't do any appending of "strict" afaict, so will always just run tests in sloppy mode. That is the meaning of the note.

Does this explain thing better?

jmdyck:
> ECMAScript code (when parsed with the Script goal symbol) can be run in either strict mode or non strict mode at the global level.

This sounds like a statement that would be supportable by reference to the spec, but I gather it isn't meant to be?

> it is up to the harness to control the strictness by prepending "use strict" when we want to run in strict mode.

Okay, so it sounds like:
* The content of any given test file is not necessarily an "ECMAScript source text" (in the spec's terminology).
* Rather, the test harness is allowed/expected to take the file's content, transform it, and supply the result to the implementation-under-test as the source code of a script.
* One such transformation is the insertion of a Use Strict Directive:
  * For `onlyStrict`, the harness should insert a USD in order to obtain the expected result.
[...]

PR #508, opened 2016-02-12
Add documentation on interpreting tests

jugglinmike:
I'm interested in formalizing the contract between Test262 and the implementors
who maintain harnesses to execute it. For instance, implementors currently have
to infer that:

- most tests rely on a file named `harness/sta.js`.
- some tests mutate global state so as to make realm sharing unacceptable
- test code must be evaluated in the global scope

I expect we will see more benefit from a document like this as we continue to
expand coverage in Test262, particularly for aspects of the environment which
cannot be directly controlled from ECMAScript code. For instance: 

- module resolution semantics
- cross-script concerns (e.g. a host-provided `$CREATESCRIPT` function)
- cross-realm concerns (e.g. a host-provided `$CREATEREALM` function)

...but I think I've covered all of the current requirements.

Commit message:

> The project's CONTRIBUTING.md was written with test authors in mind. It
> contains details on non-technical metadata (e.g. "author" and "es6id"),
> helper function usage, and preferred code structure. In addition, it
> elides certain low-level technical details on the requirements for the
> runtime environment.
>
> Introduce a new document targeted towards those executing the tests.
> Formalize all expectations regarding how the runtime environment should
> be defined, how metadata should be interpreted, and how results should
> be understood. This information has overlap with the CONTRIBUTING.md
> file, but it also contains details that are irrelevant to test authors.
>
> This document can serve as a more formal contract between Test262 and
> the implementors who consume it. This allows Test262 to unambiguously
> document future modifications to the formal requirements which in turn
> supports consumers who maintain their own test harnesses.

### DIFF
--- a/INTERPRETING.md
+++ b/INTERPRETING.md
@@ -39,7 +39,15 @@ properties of the global scope prior to test execution.
 
 Unless configured otherwise (via the `noStrict`, `onlyStrict`, or `raw` flags),
 each test must be executed twice: once in ECMAScript's non-strict mode, and
-again in ECMAScript's strict mode.
+again in ECMAScript's strict mode. To run in strict mode, the test contents
+must be modified prior to execution--[a "use strict"
+directive](https://tc39.github.io/ecma262/#sec-directive-prologues-and-the-use-strict-directive)
+must be inserted as the initial character sequence of the file, followed by a
+semicolon (`;`) and newline character (`\n`):
+
+    "use strict";
+
+This must precede any additional text modifications described by test metadata.
 
 ## Test Results
 
--- a/INTERPRETING.md
+++ b/INTERPRETING.md
@@ -37,10 +37,10 @@ properties of the global scope prior to test execution.
 
 ### Strict Mode
 
-Unless configured otherwise (via the `noStrict`, `onlyStrict`, or `raw` flags),
-each test must be executed twice: once in ECMAScript's non-strict mode, and
-again in ECMAScript's strict mode. To run in strict mode, the test contents
-must be modified prior to execution--[a "use strict"
+Unless configured otherwise (via the `noStrict`, `onlyStrict`, `module`, or
+`raw` flags), each test must be executed twice: once in ECMAScript's non-strict
+mode, and again in ECMAScript's strict mode. To run in strict mode, the test
+contents must be modified prior to execution--[a "use strict"
 directive](https://tc39.github.io/ecma262/#sec-directive-prologues-and-the-use-strict-directive)
 must be inserted as the initial character sequence of the file, followed by a
 semicolon (`;`) and newline character (`\n`):
@@ -102,6 +102,8 @@ The `flags` attribute is an optional value that specifies one or more of the
 following strings:
 
 - **`onlyStrict`** The test must be executed just once--in strict mode, only.
+  This must be accomplished using the transformation described in the section
+  titled "Strict Mode".
 
   *Example*
 
@@ -117,6 +119,8 @@ following strings:
   ```
 
 - **`noStrict`** The test must be executed just once--in non-strict mode, only.
+  In other words, the transformation described by the section titled "Strict
+  Mode" must **not** be applied to these tests.
 
   *Example*
 
@@ -134,6 +138,10 @@ following strings:
 
 - **`module`** The test source code must be interpreted as [module
   code](http://www.ecma-international.org/ecma-262/6.0/#sec-types-of-source-code).
+  In addition, this flag negates the default requirement to execute the test
+  both in strict mode and in non-strict mode. In other words, the
+  transformation described by the section titled "Strict Mode" must **not** be
+  applied to these tests.
 
   *Example*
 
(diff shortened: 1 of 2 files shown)


INTERPRETING.md
```diff
@@ -0,0 +1,175 @@
...
+
+- **`print`** A function that exposes the string value of its first argument to
+  the test runner. This is used as a communication mechanism for asynchronous
+  tests (via the `async` flag, described below).
+
+### Strict Mode
+
+Unless configured otherwise (via the `noStrict`, `onlyStrict`, or `raw` flags),
+each test must be executed twice: once in ECMAScript's non-strict mode, and
+again in ECMAScript's strict mode.
```
jmdyck: This doc needs to better explain what this means. According to the spec, strictness is a property of regions of source code, not an execution mode of the processor than can be enabled or disabled at will. See issue #299 and especially https://github.com/tc39/test262/issues/299#issuecomment-108620354 where I take a stab at defining what test262 means by "executing a test in (non-)strict mode".

jmdyck:
Thanks for creating this doc!

INTERPRETING.md
bterlson: Given that the test must be executed as global code the only option is to run the test with a global use strict directive. Saying this explicitly is a good idea.

INTERPRETING.md
```diff
@@ -0,0 +1,175 @@
...
+
+## Test Results
+
+By default, tests signal failure by generating an uncaught exception. If
+execution completes without generating an exception, the test must be
+interpreted as "passing." Any uncaught exception must be interpreted as test
+failure. These semantics may be modified by any test according to the metadata
+declared within the test itself (via the `negative` attribute and the `async`
+flag, described below).
```
bterlson: Lack of exception doesn't not necessarily mean pass. An async test might not call done and fail.

INTERPRETING.md
```diff
@@ -0,0 +1,175 @@
...
+
+## Metadata
+
+Each test file may define metadata that describe additional requirements. This
+information is delimited by the token sequence `/*---` and `---*/` and is
+structured as [YAML](http://yaml.org/).
```
bterlson: We may want to stick to a small subset of YAML and document it. Some may prefer to hand-write a quick parser than use a complete YAML parser.

INTERPRETING.md
jugglinmike: @jmdyck I'm glad you're digging this! I've added some text to be more explicit
about enabling strict mode. I've taken a cue from your comments in gh-299,
refraining from referring to the input as "source text" since (as you
mentioned) that term has a specific meaning in this context.

> To run in strict mode, the test contents must be modified prior to execution--[a "use strict" directive](https://tc39.github.io/ecma262/#sec-directive-prologues-and-the-use-strict-directive) must be inserted as the initial character sequence of the file, followed by a semicolon (`;`) and newline character (`\n`):
>
>     "use strict";
>
> This must precede any additional text modifications described by test metadata.

What do you think?

INTERPRETING.md
jugglinmike: My goal is for the text to be accurate without being repetitive or undirected/wandering. I tried to call out the contradictory behavior by beginning the section with "By default, [...]" and by ending it with a reference to the `negative` and `async` flags. Can you recommend a better way to address this?

INTERPRETING.md
jugglinmike: Sure! I'm not really familiar with YAML, but from reading the spec, I believe this is a reasonably complete description of the subset we use:

> [...] structured using a subset of [YAML](http://yaml.org/):
>
> - Block mappings
> - Flow scalars (including both single- and double- quoted styles)
> - Flow sequences
> - Block sequences
> - Block scalars
>
> For example:
>
> ```yaml
> first: scalar value
> second: "`quoted` scalar value"
> third: [a, flow, sequence]
> fourth:
>    - a
>    - block
>    - sequence
> fourth: |
>     A block scalar (without newline folding)
> fifth: >
>     Another block scalar (with newline folding)
> ```

Does this look good to you? (By the way, this will have an impact on the design of gh-382.)

jugglinmike:
@goyakin I just realized that I authored this under the assumption that gh-504 would be merged. We should hold off on merging this until we merge/close that pull request.

jugglinmike:
I just identified some lingering vaguery with the document's treatment of strict mode:

- The "Strict Mode" instructions did not include `module` in the list of
  metadata flags that negate them
- The instructions for the `noStrict` and `onlyStrict` flags did not reference
  the (now improved) "Strict Mode" section
- The instructions for the `module` flag did not specify that the tests should
  not be transformed (it's unnecessary to run these with a global "use strict"
  directive)

I've pushed up another commit to address this.

INTERPRETING.md
leobalter: the async flag has landed now, maybe it solves this problem.

leobalter:
gh-504 is merged and this PR is LGTM.

@bterlson we still need your feedback on the inline comments above.

Issue #438, opened 2015-11-04
Missing test for-of with non-destructuring, non-simple LHS


caitp:
Example: [v8:4418](https://code.google.com/p/v8/issues/detail?id=4418)

jugglinmike:
Hey Cait! I've submitted a pull request for this early error plus a few more over at gh-510.

leobalter:
this is fixed by 5a77ac8 and #510

PR #510, opened 2016-02-16
Add tests for IterationStatement early errors

jugglinmike:
gh-438 describes an early error that is not currently verified by Test262. It looks as though *none* of the early errors for IterationStatements ([here](https://tc39.github.io/ecma262/#sec-semantics-static-semantics-early-errors), [here](https://tc39.github.io/ecma262/#sec-for-statement-static-semantics-early-errors), and [here](https://tc39.github.io/ecma262/#sec-for-in-and-for-of-statements-static-semantics-early-errors)) are tested. This patch contains tests for each (that's an iteration joke).

Doing this cleanly involved a little housekeeping around file names (beyond gh-499). I've organized those changes into a separate commit.

Resolves gh-438.


test/language/statements/for-in/head-const-bound-names-in-stmt.js
```diff
@@ -0,0 +1,15 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+description: The body may not re-declare variables declared in the head
+negative: SyntaxError
+info: |
```
leobalter: we've been using the `>` character on the other parts, it's good to use them here for consistency.

test/language/statements/for-in/head-const-bound-names-in-stmt.js
```diff
@@ -0,0 +1,15 @@
...
+    It is a Syntax Error if any element of the BoundNames of ForDeclaration
+    also occurs in the VarDeclaredNames of Statement.
+id: sec-for-in-and-for-of-statements
+es6id: 13.7
+---*/
+
+for (const x in {}) {
+  var x;
```
leobalter: this is a great test and a great feature, it would be challenging to interpret this code otherwise.

test/annexB/language/statements/for-in/labeled-fn-stmt-lhs.js
```diff
@@ -0,0 +1,11 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+description: It is a Syntax Error if IsLabelledFunction(Statement) is true.
+negative: SyntaxError
+id: sec-labelled-function-declarations
+es6id: B.3.2
+---*/
+
+var x;
```
leobalter: this line is not necessary

test/language/statements/for-in/head-const-bound-names-in-stmt.js
```diff
@@ -0,0 +1,15 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+description: The body may not re-declare variables declared in the head
+negative: SyntaxError
+info: |
+    It is a Syntax Error if any element of the BoundNames of ForDeclaration
+    also occurs in the VarDeclaredNames of Statement.
+id: sec-for-in-and-for-of-statements
+es6id: 13.7
```
leobalter: we don't need to have both ids, but it's not an early error if you keep it. (intended joke)

test/language/statements/for-in/head-const-bound-names-let.js
```diff
@@ -0,0 +1,13 @@
...
+description: The declaration may not contain a binding for `let`
+negative: SyntaxError
+info: |
+    It is a Syntax Error if the BoundNames of ForDeclaration contains "let".
+id: sec-for-in-and-for-of-statements
+es6id: 13.7
+flags: [noStrict]
+---*/
+
+for (const let in {}) {}
```
leobalter: this is an early error on strict mode too, right?

leobalter:
I loved the use of label recursion (on the labelledItem).

Just a few minor comments, this already looks great.

test/annexB/language/statements/for-in/labeled-fn-stmt-lhs.js
jugglinmike: You are correct. Three other tests for SyntaxErrors also needlessly included variable declarations. I've removed them all in 9be1481

test/language/statements/for-in/head-const-bound-names-in-stmt.js
jugglinmike: The `>` block literal delimiter is used for "folded" block literals, meaning
newlines will be stripped from the specified value. For the most part, we've
been using newlines to format the contents of our tests' `info` fields. My use
of `|` here is intentional; it may be inconsistent with the existing tests, but
I think correctness trumps consistency.

test/language/statements/for-in/head-const-bound-names-let.js
jugglinmike: It is an early error, but due to another part of the specification (strict mode
restrictions on identifiers). This test declares `noStrict` in order to keep
the semantics under test limited to the IterationStatement's early errors (we
maintain explicit tests for strict mode identifier restrictions elsewhere).

test/language/statements/for-in/head-const-bound-names-in-stmt.js
leobalter: Do you mean the newline characters?

That's interesting, thanks!

leobalter:
LGTM. :shipit:

caitp:
The V8 issue referenced in #438 is specifically about `for (<PropertyReference> of <Expression>)`, eg `x.y of [1, 2, 3]`, etc

jugglinmike:
Ah, I guess I was confused because the original report used a MemberExpression within a parenthesized expression--it was `for-in/head-lhs-cover.js` and `for-of/head-lhs-cover.js` that I had in mind. Anyway, I've added analogous tests for a MemberExpression in that position.

caitp:
I believe the parenthesized MemberExpression is valid too, due to being a valid simple assignment target  -- the new tests look good tho

test/language/statements/for-in/head-lhs-member.js
```diff
@@ -0,0 +1,19 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: >
+    Head's AssignmentExpression may be a MemberExpression
+id: sec-for-in-and-for-of-statements-static-semantics-early-errors
+es6id: 13.7
```
leobalter: I believe `sec-for-in-and-for-of-statements` and `13.7.5` fit best on the id fields. (the same applies for the other new test file)

test/annexB/language/statements/do-while/labeled-fn-stmt.js
```diff
@@ -0,0 +1,10 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+description: It is a Syntax Error if IsLabelledFunction(Statement) is true.
+negative: SyntaxError
+id: sec-labelled-function-declarations
+es6id: B.3.2
+---*/
+
+do label1: label2: function f() {} while (false)
```
goyakin: Isn't `LabelledItem: FunctionDeclaration` disallowed **only** in **strict** mode? I believe this file needs the `onlyStrict` flag because of that.

test/language/statements/for/head-const-bound-names-in-stmt.js
```diff
@@ -0,0 +1,18 @@
...
+description: The body may not re-declare variables declared in the head
+info: |
+    IterationStatement :
+        for ( LexicalDeclaration Expressionopt ; Expressionopt ) Statement
+
+    It is a Syntax Error if any element of the BoundNames of LexicalDeclaration
+    also occurs in the VarDeclaredNames of Statement.
+negative: SyntaxError
+id: sec-for-in-and-for-of-statements
+es6id: 13.7
```
goyakin: The IDs for files under `test/language/statements/for` point to the wrong section.

test/language/statements/for-in/head-lhs-member.js
goyakin: I agree.

test/language/statements/for-in/head-const-bound-names-in-stmt.js
goyakin: :smile:

@jugglinmike If you decide to keep the `es6id`s, we should update them to point to the more specific sections (e.g., `13.7.5` instead of `13.7`).

goyakin:
@jugglinmike Other than the minor comments I left inline, this PR LGTM.

test/language/statements/for/head-const-bound-names-in-stmt.js
jugglinmike: My mistake. Fix is on it's way

test/language/statements/for-in/head-lhs-member.js
jugglinmike: Alrighty

test/annexB/language/statements/do-while/labeled-fn-stmt.js
jugglinmike: This is a little tricky, but I believe this particular early error applies
regardless of language mode.

[Section
13.3.1](http://www.ecma-international.org/ecma-262/6.0/#sec-labelled-statements-static-semantics-early-errors)
describes an early error for labelled statements generally:

> LabelledItem : FunctionDeclaration
>
>     It is a Syntax Error if any source text matches this rule.
> 
> NOTE An alternative definition for this rule is provided in B.3.2.

[Section
B.3.2](http://www.ecma-international.org/ecma-262/6.0/#sec-labelled-function-declarations)
modifies that general rule:

> It is a Syntax Error if any strict mode source code matches this rule.

This clearly limits the error to strict mode. But the early error under test
specifically concerns the statement as is appears within an IterationStatement.
[Section
13.7.1.1](http://www.ecma-international.org/ecma-262/6.0/#sec-labelled-statements-static-semantics-early-errors)
reads:

> It is a Syntax Error if IsLabelledFunction(Statement) is true.
>
> NOTE It is only necessary to apply this rule if the extension specified in
> B.3.2 is implemented.

Here, the language mode is *not* taken into account.

So by my reading, the only place where a labelled function declaration should
not throw an error is when it appears in a StatementList of non-strict mode
code that does not implement Annex B:

Position           | Mode       | Annex B         | Error?
-------------------|------------|-----------------|--------
StatementList      | Non-strict | Not implemented | Yes
StatementList      | Non-strict | Implemented     | No
StatementList      | Strict     | Not implemented | Yes
StatementList      | Strict     | Implemented     | Yes
IterationStatement | Non-strict | Not implemented | Yes
IterationStatement | Non-strict | Implemented     | Yes
IterationStatement | Strict     | Not implemented | Yes
IterationStatement | Strict     | Implemented     | Yes

Does this sound right to you?

goyakin: I see, this should be an error regardless of the mode for `IterationStatement`. In that case, we should change the `id` to point to `sec-semantics-static-semantics-early-errors` (`13.7.1.1`) instead.

goyakin:
One more thing, could you convert `id`s in the frontmatter to `esid`?

test/annexB/language/statements/do-while/labeled-fn-stmt.js
jugglinmike: Yes, I think you're right.

jugglinmike:
Sure!

jugglinmike:
I've also added an `info` field to describe the intention of the AnnexB tests, since the issue is so subtle:

> Although Annex B describes an extension which permits labelled function declarations outside of strict mode, this early error is applied regardless of the language mode.

goyakin:
Merged as 5a77ac86a95e54b0e1d03588e1642e04652b9429.

@jugglinmike FYI, since Annex B doesn't apply, I moved the files under `test/annexB/language` to `test/language`.

Issue #512, opened 2016-02-19
Race Condition in Console Runner


jugglinmike:
A race condition in the console runner's threading logic causes it to run some
test cases more than once and to not run others at all.

This example demonstrates a test run using 2 parallel threads (via `-j 2`). I'm
using a shell pipeline to highlight repeated runs:

    $ ./tools/packaging/test262.py --command js language/reserved-words -j 2 | sort | uniq -c | sort | tail
    2 language/reserved-words/7.6.1-7-10 passed in non-strict mode
    2 language/reserved-words/7.6.1-7-11 passed in non-strict mode
    2 language/reserved-words/7.6.1-7-1 passed in non-strict mode
    2 language/reserved-words/7.6.1-7-8 passed in non-strict mode
    2 language/reserved-words/7.6.1-7-8 passed in strict mode
    2 language/reserved-words/7.6.1-8-3 passed in strict mode
    2 language/reserved-words/7.6.1-8-6 passed in strict mode
    2 language/reserved-words/S7.6.1_A1.1 failed in strict mode as expected
    2 language/reserved-words/S7.6.1_A2 passed in non-strict mode
    3 language/reserved-words/7.6.1-8-1 passed in strict mode

Executing the tests within a single thread (via `-j 1`) does not present this
problem:

    $ ./tools/packaging/test262.py --command js language/reserved-words -j 1 | sort | uniq -c | sort | tail
    1 language/reserved-words/7.6.1-8-9 passed in non-strict mode
    1 language/reserved-words/7.6.1-8-9 passed in strict mode
    1 language/reserved-words/S7.6.1_A1.1 failed in non-strict mode as expected
    1 language/reserved-words/S7.6.1_A1.1 failed in strict mode as expected
    1 language/reserved-words/S7.6.1_A1.2 failed in non-strict mode as expected
    1 language/reserved-words/S7.6.1_A1.2 failed in strict mode as expected
    1 language/reserved-words/S7.6.1_A1.3 failed in non-strict mode as expected
    1 language/reserved-words/S7.6.1_A1.3 failed in strict mode as expected
    1 language/reserved-words/S7.6.1_A2 passed in non-strict mode
    1 language/reserved-words/S7.6.1_A2 passed in strict mode

We have a couple options for how to handle this: fix the bug or remove the
optimization. Patches and rationale incoming.

PR #513, opened 2016-02-19
Console Runner: Revert Parallelization and Formally Deprecate

jugglinmike:
I've heard it mentioned that the Python "console runner" utility is deprecated,
but there is no documentation in this project to suggest this. If it is the
case, we should be more clear about it in order to better set expectations for
consumers and contributors.

A decision to deprecate would also motivate the removal of a recent feature
addition to the console runner--support for parallelization. To date, three
bugs have been found in that implementation. As an optimization to a deprecated
tool, the maintenance overhead is not justified.

Resolves gh-512.

### DIFF
--- a/CONSOLE-RUNNER.md
+++ b/CONSOLE-RUNNER.md
@@ -40,7 +40,6 @@ Name | Action
 -----|-------
 -h, --help | displays a brief help message
 --command=COMMAND | **required** command which invokes javascript engine to be tested
--j, --workers-count | Number of tests to run in parallel (defaults to number of cores - 1)
 --tests=TESTS | path to the test suite; default is current directory
 --cat | don't execute tests, just print code that would be run
 --summary | generate a summary at end of execution
@@ -50,9 +49,9 @@ Name | Action
 --unmarked_default=MODE | mode to use for tests that are not marked **onlyStrict** or **noStrict** ; MODE can be `strict` or `non_strict` or `both`
 --logname=LOGNAME | write output to file (in addition to stdout)
 --junitname=JUNITNAME | write test results to file in JUnit XML format
---loglevel=LOGLEVEL | set log level, primarily useful for debugging `test262.py`
+--loglevel=LOGLEVEL | set log level, primarily useful for debugging `test262.py` 
 --print-handle=FUNC | enable async test logging via javascript function e.g., `console.log`
-
+ 
 ### Usage Notes
 
 Non-option arguments are used as filters to match test names.  If no filters are found, the whole test suite is run.
@@ -67,7 +66,7 @@ The COMMAND argument can be a quoted string.  This is useful when testing ECMASc
 
 ```
 $ test262.py --command="node --harmony" es6
-```
+``` 
 
 #### Async Tests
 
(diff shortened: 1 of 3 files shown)


goyakin:
We at Microsoft don't use the Python runner. If nobody is actually is using it, I'd be in favor of **officially** deprecating it instead of [fixing the race condition](https://github.com/tc39/test262/pull/514).

tschneidereit:
Clearly I'm not smart enough to write multithreaded code :(

Still, the console runner being the only way to properly automate running these tests in a shell (and thus with reasonable speed), I'm opposed to deprecating it. I would in fact like to use it to run test262 in our pre-commit tests.

If people are strongly opposed to having a console runner, then I'll just import what I need into our repository and only use test262 as the source of tests. I'd find that unfortunate, though.

jugglinmike:
I think that offering a console runner has always been an over-extension of this project's responsibilities. We can't serve everyone, so some consumers will invariably build their own runner anyway.

So while I understand your desire to have an executable packaged with Test262, it seems healthier in the long run to place that responsibility elsewhere. As you mention, you are free to maintain a customized copy of the runner in your project. In that case, you'll be happy to know that we're formalizing the expectations for test interpretation--see gh-508.

You may not want to take on that burden, though. As luck would have it, @bterlson has built [test262-harness](https://github.com/bterlson/test262-harness) an excellent replacement using Node.js--maybe that will suit your needs?

tschneidereit:
Ok, I can't say I agree or fully understand why hosting a runner isn't a
useful part of this project, but as you say I can just fork and maintain a
copy.


> You may not want to take on that burden, though. As luck would have it,
> @bterlson <https://github.com/bterlson> has built test262-harness
> <https://github.com/bterlson/test262-harness> an excellent replacement
> using Node.js--maybe that will suit your needs?
>

Thanks for the pointer. Doesn't help me though, because our buildsystem
doesn't have node.js where I'd need it.

littledan:
Before officially deprecating this tool being part of test262, let's consult Apple to see if they use it or plan to use it. If two browsers find it useful, then it would make sense to stay. If it's just for one, then we might consider deprecating it.

I think it's within scope of this project to make sure it's usable. It comes down to details whether this test runner being maintained as part of this project is important, rather than any abstract principle, I think.

kmiller68:
We don't have any current plans to use the Python test runner. I'll be integrating test262 into our existing infrastructure to make running tests simpler. That said, if the console running is being actively maintained I don't see any reason to deprecate or remove it from the project.

littledan:
I don't really see the point in leaving something in the repository frozen and deprecated. Why don't we either let Till maintain it here for Mozilla's use, or remove it?

tschneidereit:
>
> I don't really see the point in leaving something in the repository frozen
> and deprecated. Why don't we either let Till maintain it here for Mozilla's
> use, or remove it?
>

I'd be ok with either. One thing to note might be that people do seem to
use it, otherwise the issues I introduced wouldn't have mattered, right?

jugglinmike:
TC-39 agreed to deprecate the runner in 2014. This decision wasn't
well-communicated, so there was some contention a few months later when
@domenic filed gh-128 to remove the runner. The discussion died off without a
change to TC-39's decision or to the runner itself.

I came across the bugs when implementing new features for the test runner:
first for the proposed `async` flag, and later for executing module code. I
suggested formal deprecation in light of the long-standing TC-39 decision and
because I was afraid I was devoting energy to a rarely-used/updated tool.

littledan:
Thanks for the reference to the history. Let's bring this up at the next TC39 meeting to see if we should reverse the deprecation decision. To address the concern about wasted maintanence work: Whether or not the runner is deprecated, I don't think @jugglinmike is personally responsible for maintaining it. But it would definitely be unfortunate if Mozilla and V8 (for the inner parts Andreas mentioned) needed to maintain separate copies of the same code.

PR #514, opened 2016-02-19
Console Runner: Fix Race Condition

jugglinmike:
When spawning threads from within a `for` loop, the original
parallelization implementation relied on a function closure to access
each `TestCase` instance.  Due to the concurrent and non-deterministic
nature of thread processing, the value of this reference at execution
time was unstable.

Explicitly specify the `TestCase` instance to the `Thread` constructor,
ensuring that each thread is run with the intended value, regardless of
execution order.

Resolves gh-512.

### DIFF
--- a/tools/packaging/test262.py
+++ b/tools/packaging/test262.py
@@ -591,7 +591,7 @@ class TestSuite(object):
       log_lock = None
 
     for case in cases:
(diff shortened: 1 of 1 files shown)


goyakin:
Closing in favor of #513

PR #515, opened 2016-02-19
Add tests for module code

jugglinmike:
I've tried to cover all the aspects of ES2015 modules that do *not* depend on a loader. As you might expect, it's not very much!

Test262 already maintains a few tests for module code (specifically: tests for early errors). Unfortunately, early errors from correct interpretation of module code are indistinguishable from parsing errors from incorrectly interpreting as script code. In other words, consumers can pass most of these tests by ignoring the `module` flag and interpreting the source text as script code.

The significance of this changeset is that it introduces the first tests (`language/module-code/strict-mode.js` and `language/reserved-words/await-module.js`) that runtimes cannot pass by simply interpreting module code as script code. I'm hoping that having all the negative tests in place initially will still help avoid errors as implementors begin add support for modules.

### DIFF
--- a/CONSOLE-RUNNER.md
+++ b/CONSOLE-RUNNER.md
@@ -41,6 +41,7 @@ Name | Action
 -h, --help | displays a brief help message
 --command=COMMAND | **required** command which invokes javascript engine to be tested
 -j, --workers-count | Number of tests to run in parallel (defaults to number of cores - 1)
+--module_command | command which invokes the JavaScript engine to be tested for tests labeled as ES2015 modules
 --tests=TESTS | path to the test suite; default is current directory
 --cat | don't execute tests, just print code that would be run
 --summary | generate a summary at end of execution
--- a/test/language/eval-code/export.js
+++ b/test/language/eval-code/export.js
@@ -1,0 +1,24 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+description: The `export` declaration may not appear within eval code
+id: sec-scripts
+flags: [module]
+info: |
+     Eval code is the source text supplied to the built-in eval function. More
+     precisely, if the parameter to the built-in eval function is a String, it
+     is treated as an ECMAScript Script. The eval code for a particular
+     invocation of eval is the global code portion of that Script.
+
+     A.5 Scripts and Modules
+
+     Script:
+         ScriptBodyopt
+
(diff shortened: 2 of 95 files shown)


leobalter:
In a coverage exercise, 82abd59 is already covering that Module code runs on strict mode. For the test purpose it's enough for Modules.

Additional tests should only serve the purpose to cover strict code tests on a module code. I believe they also belong to a special folder instead of `test/language/module-code`

jugglinmike:
@leobalter It sounds like you are recommending additional tests, but I don't quite understand what you have in mind. Can you elaborate?

leobalter:
I believe we may have additional tests, but they're not the on the goal for this PR.

jugglinmike:
This morning, I looked into more meaningful ways to test for strict mode. It would be possible to use [the proposed test generation tool](https://github.com/tc39/test262/pull/481) to exhaustively test *every* strict mode behavior for every strict-mode-enabling form.

This may be too many tests, though. It would essentially be the cross product of [10.2.1](https://tc39.github.io/ecma262/#sec-strict-mode-code) and [Annex C](https://tc39.github.io/ecma262/#sec-strict-mode-of-ecmascript). The former set is quite large when you include all the permutations of valid directive prologues (which would mean a full set of test files for each test currently resisting in [`test/language/directive-prologue`](https://github.com/tc39/test262/tree/master/test/language/directive-prologue)).

leobalter:
> to exhaustively test every strict mode behavior for every strict-mode-enabling form

I'm glad we have plans for the test generation tool, it's indeed necessary to create and manage these tests.

goyakin:
@jugglinmike I'll merge this PR with the current set of tests. We can add the additional ones later.

Also, I plan on merging [your PR deprecating the console runner](https://github.com/tc39/test262/pull/513). Given that, what do you think about reverting [the changes to the runner](https://github.com/bocoup/test262/commit/13e10f3628c5a6db269d848716632c0a461efcee) in this PR? I didn't go through those, but if you want to keep them, I'll take a look.

jugglinmike:
I wouldn't mind if we landed this without the feature addition for the runner. It's not really ergonomic, anyway. Would you be able to remove it before merging? Or would you rather I remove it and force-push to this branch?

goyakin:
I'll remove it before merging then :wink:

goyakin:
Merged as b35b39c4aafef364d6a7d67865a427b74c5057ff. Thanks @jugglinmike.

PR #517, opened 2016-02-22
Add tests for TypedArrays.prototype methods - part 1

leobalter:
- Add tests to check \_TypedArrays\_.prototype methods inherit from %TypedArray%
- Fix a %TypedArray%.prototype.length test
- Add tests fro %TypedArray%.prototype get methods (buffer, byteLength, byteOffset, length, @@toStringTag)

The other methods will be covered on extra PRs to minimize review overload.


leobalter:
I removed the bad commit.

leobalter:
and included some tests for missing descriptors coverage.

leobalter:
I added some tests on 993d79c which are quite interesting.

It looks like buffer, byteLength and byteOffset could check for a [[TypedArrayName]] internal slot.

test/built-ins/TypedArray/prototype/Symbol.toStringTag/this-has-no-viewedarraybuffer-internal.js
```diff
@@ -0,0 +1,26 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 22.2.3.31
+description: |
+  Return undefined when `this` does not have a [[ViewedArrayBuffer]] internal
```
goyakin: `ViewedArrayBuffer` should be `TypedArrayName` instead.

test/built-ins/TypedArray/prototype/Symbol.toStringTag/this-is-not-object-strict-mode.js
```diff
@@ -0,0 +1,24 @@
...
+es6id: 22.2.3.31
+description: Return undefined when `this` is null or undefined
+info: >
+  22.2.3.31 get %TypedArray%.prototype [ @@toStringTag ]
+
+  1. Let O be the this value.
+  2. If Type(O) is not Object, return undefined.
+  ...
+includes: [testTypedArray.js]
+flags: [onlyStrict]
```
goyakin: It doesn't seem like the result would be different in **sloppy** mode. If that's the case, I think we should merge this file into `this-is-not-object.js`.

test/built-ins/TypedArray/prototype/length/return-length.js
```diff
@@ -0,0 +1,27 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 22.2.3.18
```
goyakin: `22.2.3.18` is the section ID in the ES 2016 draft. It's `22.2.3.17` in ES6.

test/built-ins/TypedArray/prototype/Symbol.toStringTag/this-is-not-object-strict-mode.js
goyakin: There are other tests like this in this PR. If we merge this file with the other one, we should do the same for other tests too.

goyakin:
@leobalter I left some minor comments inline, but otherwise, this PR LGTM.

test/built-ins/TypedArrays/prototype/Symbol.toStringTag/inherited.js
```diff
@@ -0,0 +1,13 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-get-%typedarray%.prototype-@@tostringtag
+description: >
+  _TypedArray_.prototype[@@toStringTag] is inherited from %TypedArray%
+  _TypedArray_.prototype has no own property @@toStringTag
+includes: [testTypedArray.js]
+---*/
```
jugglinmike: `features: [Symbol.toStringTag]`

leobalter:
> What do you think about adding features: [DataView] to the tests that use the DataView constructor?

I remember on other PR I mentioned that for ArrayBuffer, as the constructors are connected we decided to remove it. Maybe it makes more sense for DataView, I'll do it.

test/built-ins/TypedArray/prototype/Symbol.toStringTag/this-is-not-object-strict-mode.js
leobalter: the results are not different by other reason, where `this` would be the global object, and the global does not have a [[TypedArrayName]]. I'm testing step 2 here. 

The respective tests on sloppy mode are already covered on `invoked-as-func.js`

goyakin: You're right. It slipped my mind that `null` and `undefined` are replaced with the global object.

leobalter:
done!

leobalter:
the fixup references can be seen at bdea419 and 4aec964.

I'll push the squashed and rebased commits.

test/built-ins/TypedArray/prototype/Symbol.toStringTag/this-is-not-object-strict-mode.js
jugglinmike: Actually, that's only true for ECMAScript function objects. For built-in function objects like the ones under test, OrdinaryCallBindThis is *not* used. This means that both inside strict mode and outside strict mode, when either `null` or `undefined` is specified as a "this" value, that is the actual value during function execution (not the global object).

leobalter: that's interesting, thanks!

leobalter: fixed!

goyakin:
Merged as e9c7608f6f42c2b4542e553569ef4446a1766170. Thanks @leobalter.

PR #518, opened 2016-02-23
Fix test bugs

anba:
I didn't understand the test in "object-arg-throws-setting-typedarray-property.js", so I've simply removed it. Maybe @leobalter can shed some light on that one. :grinning:


test/built-ins/TypedArrays/object-arg-throws-setting-typedarray-property.js
```diff
@@ -1,73 +0,0 @@
...
-  7. Return ? OrdinaryToPrimitive(input, hint).
-
-  OrdinaryToPrimitive
-
-  ...
-  5. For each name in methodNames in List order, do
-    a. Let method be ? Get(O, name).
-    b. If IsCallable(method) is true, then
-      i. Let result be ? Call(method, O).
-      ii. If Type(result) is not Object, return result.
```
leobalter: you're right, it returns here.

Maybe I should have used a custom `toString` function (or even a custom `join` function).

littledan:
I ran into this too! A single-element TypedArray converts to a Number just fine, for better or worse, so V8 fails this "test".

ljharb:
@anba thanks for fixing the `getOwnPropertyDescriptors` tests!

PR #520, opened 2016-02-26
Add validation tests for TypedArray instance methods

leobalter:
+ basic tests for TypedArray instances iterator methods.

Tests for detachedBuffer will come in an individual PR.


jugglinmike:
- The "return-itor.js" tests all use the `sample` TypedArray as the test
  subject, but they should be using the `typedArray` object because that is the
  instance under test.
- @goyakin's [suggestion on strict
  mode](https://github.com/tc39/test262/pull/517#discussion_r54494067) from
  gh-517 applies to this patch, as well.

leobalter:
> The "return-itor.js" tests all use the sample TypedArray as the test subject

Oops. False positives are dangerous.

leobalter:
WRT the strict mode:

> the results are not different by other reason, where this would be the global object, and the global does not have a [[TypedArrayName]]. I'm testing step 2 here.
> 
> The respective tests on sloppy mode are already covered on invoked-as-func.js

leobalter:
fixed the return-itor tests on a rebased commit.

leobalter:
fixed the tests on strict mode

leobalter:
-550 lines of code...

test/built-ins/TypedArray/prototype/toString/this-is-not-object.js
```diff
@@ -0,0 +1,54 @@
...
+
+  22.2.3.15 %TypedArray%.prototype.join ( separator )
+
+  This function is not generic. ValidateTypedArray is applied to the this value
+  prior to evaluating the algorithm. If its result is an abrupt completion that
+  exception is thrown instead of evaluating the algorithm.
+
+  22.2.3.5.1 Runtime Semantics: ValidateTypedArray ( O )
+
+  1. If Type(O) is not Object, throw a TypeError exception.
```
goyakin: `%TypedArray%.prototype.join` isn't called if `this` isn't a `TypedArray`, so `ValidateTypedArray` doesn't apply in this case.

Since `this` goes through `ToObject` according to `Array.prototype.toString`, `undefined` and `null` should cause a `TypeError`, but the other types are valid.

test/built-ins/TypedArray/prototype/toString/this-is-not-typedarray-instance.js
```diff
@@ -0,0 +1,45 @@
...
+  22.2.3.15 %TypedArray%.prototype.join ( separator )
+
+  This function is not generic. ValidateTypedArray is applied to the this value
+  prior to evaluating the algorithm. If its result is an abrupt completion that
+  exception is thrown instead of evaluating the algorithm.
+
+  22.2.3.5.1 Runtime Semantics: ValidateTypedArray ( O )
+
+  1. If Type(O) is not Object, throw a TypeError exception.
+  2. If O does not have a [[TypedArrayName]] internal slot, throw a TypeError
```
goyakin: Because of the same reason as `this-is-not-object.js`, we shouldn't get a `TypeError`.

goyakin:
@leobalter This PR LGTM except the tests for `%TypedArray%.prototype.toString`.

test/built-ins/TypedArray/prototype/toString/this-is-not-object.js
leobalter: there's no validation on toString before calling join. 

The full description on toString omits the validation:

```

22.2.3.29 %TypedArray%.prototype.toString ( )

The initial value of the %TypedArray%.prototype.toString data property is the same built-in function object as the Array.prototype.toString method defined in 22.1.3.28.
```

Then Array.prototype.toString will be called first, getting the `join` method:

```
22.1.3.28 Array.prototype.toString ( )

When the toString method is called, the following steps are taken:

    1. Let array be ? ToObject(this value).
    2. Let func be ? Get(array, "join").
    3. If IsCallable(func) is false, let func be the intrinsic function %ObjProto_toString%.
    4. Return ? Call(func, array). 
```

goyakin: Since `Get` would return `undefined` for, for example, a Number, `IsCallable` would return `false`, causing `%ObjProto_toString%` to be called.

`Array.prototype.toString.call(42)` returns `"[object Number]"`, so `TypedArray.prototype.toString.call(42)` should too.

leobalter: Right you are.

Thanks for your patience, @goyakin! I'll fix it.

leobalter: It's fixed now at 9bcfd28.

goyakin: @leobalter, what do you think about simply checking for `TypedArray.prototype.toString === Array.prototype.toString` instead? According to the spec, they should have the same value.

leobalter: you are right, again. :+1: 

I removed the prop-desc as it's already covered on `built-ins/TypedArray/prototype/toString.js`

leobalter:
ping @jugglinmike @goyakin. Can we merge this?

jugglinmike:
Certainly. See b26190f1ce79bdb3624a30d937284baa1b144430 & 83b27c9beb7b3ea78b319148b5c82a5411a333b1

leobalter:
thanks!

PR #523, opened 2016-02-29
Fix bugs in Object.getOwnPropertyDescriptors() tests

### LINKED ISSUE #522: Object.getOwnPropertyDescriptors test bugs
See comments left at https://github.com/tc39/test262/pull/484/files#r54413026 and https://github.com/tc39/test262/pull/484/files#r54412155

I'll send a pull request to fix these, but unfortunately it means we need to disable these tests until the next v8 roll at least :\

caitp:
Closes #522


caitp:
I guess these are dealt with in #518, so that's cool

PR #524, opened 2016-02-29
Add tests for IteratorClose in dstr assignment

jugglinmike:
@littledan recently discovered a hole in Test262's coverage of destructuring assignment: no tests exist to assert the use of the IteratorClose interal operation in [DestructuringAssignmentEvaluation](http://www.ecma-international.org/ecma-262/6.0/#sec-runtime-semantics-destructuringassignmentevaluation).

I've tried address all the missing cases in this patch. To assist review, here's the naming scheme:

Prefix             | Grammar production
-------------------|-------------------
`array-empty-`     | ArrayAssignmentPattern : [ ]
`array-elision-`   | ArrayAssignmentPattern : [ Elision ]
`array-rest-`      | ArrayAssignmentPattern : [ Elisionopt AssignmentRestElement ]
`array-elem-`      | ArrayAssignmentPattern : [ AssignmentElementList ]
`array-elem-trlg-` | ArrayAssignmentPattern : [ AssignmentElementList , Elisionopt AssignmentRestElementopt ]

Suffix             | Intent
-------------------|-------
`-abpt-close-err`  | The assignment evaluation returns an abrupt completion, and the iterator's `return` method throws an error
`-abpt-close-skip` | The assignment evaluation returns an abrupt completion, but IteratorClose is not invoked
`-abpt-close`      | The assignment evaluation returns an abrupt completion, and IteratorClose is invoked as specified
`-get-err`         | Abrupt completion returned from GetIterator
`-nrml-close-err`  | The assignment evaluation completes, and the iterator's `return` method throws an error
`-nrml-close-null` | The assignment evaluation completes, and the iterator's `return` method returns a non-Object value (there is no corresponding `-abpt-` suffix because the algorithm does not reference the return value in those cases)
`-nrml-close-skip` | The assignment evaluation completes, but IteratorClose is not invoked
`-nrml-close`      | The assignment evaluation completes, and IteratorClose is invoked as specified

Not every production uses all of the suffixes; I simplified the file name where the specificity was not necessary (e.g. `array-elision-iter-abpt.js`).


anba:
Can you also add tests to ensure `Return` completions are handled and call `IteratorClose`? 

```js
var finallyExecuted = false;
function* g() {
  [a = yield] = function*() {
    try {
      yield;
    } finally {
      finallyExecuted = true;
    }
  }();
}
var it = g();
it.next();
it.return();
assert(finallyExecuted);
```

jugglinmike:
Sounds good, @anba. This warrants a change to the naming scheme. I'll rename
`*-abpt-*` to `*-thrw-*` to make room for equivalent `*-rtrn-*` tests. I'm
going to try a more direct approach in order to limit the semantics under test:

```js
var iterable = {};
var returnCount = 0;
var unreachable = 0;
iterable[Symbol.iterator] = function() {
  return {
    next: function() {},
    return: function() {
      returnCount += 1;
      return {};
    }
  };
};
function* g() {
  [x[yield]] = iterable;
  unreachable += 1;
};
var it = g();
it.next();
it.return();
assert.sameValue(returnCount, 1);
assert.sameValue(unreachable, 0);
```

Does that look valid to you?

anba:
> Does that look valid to you?

Yes, only needs two small changes:

-  `next` needs to return an iterator result: `next: function() { return {value: undefined, done: false}; }` 
- and I assume `[x[yield]]` should be `[x = yield]` ?

jugglinmike:
`[x[yield]]` was intentional. I believe it will also exercise the code path we are interested in here, though unlike `[x = yield]`, it occurs prior to iteration, obviating the need for a valid `next` method. I may be mistaken, though!

anba:
Ah ok, I see what you're aiming at. You can avoid specifying a `next` method if you use `[(yield).x] = iterable;`.

`[x[yield]] = iterable;` will report a reference error because `x` is not defined.

jugglinmike:
Or even `[{}[yield]] = iterable;`. We're really golfing, now :)

jugglinmike:
@anba Okay, new tests added. As I'm sure you expected, these ended up being a
novel combination of the tests cases for normal completions and those for
"throw" completions.

They are distinct from both "normal" and "throw" because, to the best of my
knowledge, it's impossible to interrupt the iteration protocol with a "return"
completion (so no files matching `*-rtrn-close-skip,js`)... But like "normal"
completions, the effect of returning a non-Object value from the `return`
method is observable (which is why there are `*-rtrn-close-null.js` tests). And
finally, you can't trigger a "return" completion during evaluation of the
elision pattern, so the catch-all `-abpt.js` files did not need to be split
into `-rtrn.js` and `-thrw.js` files.

Those were fun to write--thanks for the input!

@goyakim This branch now has `fixup!` commits that modify the naming scheme
described in the first commit message. Prior to merging, I would like to squash
and update the commit message accordingly, so please let me know when you are
ready to land this patch.

leobalter:
:100: @jugglinmike 

LGTM.

PR #525, opened 2016-03-01
Improve coverage %ThrowTypeError% and default class constructors

anba:
Increases the coverage for %ThrowTypeError% and adds an _interesting_ test case for default class constructors (https://bugzilla.mozilla.org/show_bug.cgi?id=1234702).

### DIFF
--- a/test/built-ins/ThrowTypeError/name.js
+++ b/test/built-ins/ThrowTypeError/name.js
@@ -1,0 +1,17 @@
+// Copyright (C) 2016 André Bargull. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-%throwtypeerror%
+description: >
+  %ThrowTypeError% is an anonymous function.
+info: >
+  %ThrowTypeError% ( )
+
+  The %ThrowTypeError% intrinsic is an anonymous built-in function
+  object that is defined once for each Realm.
+---*/
+
+var ThrowTypeError = Object.getOwnPropertyDescriptor(function(){ "use strict"; return arguments; }(), "callee").get;
+
+assert.sameValue(Object.prototype.hasOwnProperty.call(ThrowTypeError, "name"), false);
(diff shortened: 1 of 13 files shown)


test/built-ins/ThrowTypeError/name.js
```diff
@@ -0,0 +1,17 @@
...
+info: >
+  %ThrowTypeError% ( )
+
+  The %ThrowTypeError% intrinsic is an anonymous built-in function
+  object that is defined once for each Realm.
+---*/
+
+var ThrowTypeError = Object.getOwnPropertyDescriptor(function(){ "use strict"; return arguments; }(), "callee").get;
+
+assert.sameValue(Object.prototype.hasOwnProperty.call(ThrowTypeError, "name"), false);
```
leobalter: Isn't this affected bythe var binding?

https://tc39.github.io/ecma262/#sec-variable-statement-runtime-semantics-evaluation

```
VariableDeclaration: BindingIdentifier Initializer 

  1.Let bindingId be StringValue of BindingIdentifier.
  2.Let lhs be ? ResolveBinding(bindingId).
  3.Let rhs be the result of evaluating Initializer.
  4.Let value be ? GetValue(rhs).
  5.If IsAnonymousFunctionDefinition(Initializer) is true, then
    a.Let hasNameProperty be ? HasOwnProperty(value, "name").
    b.If hasNameProperty is false, perform SetFunctionName(value, bindingId).
```

test/built-ins/ThrowTypeError/unique-per-realm-unmapped-args.js
```diff
@@ -0,0 +1,28 @@
...
+info: >
+  %ThrowTypeError% ( )
+
+  The %ThrowTypeError% intrinsic is an anonymous built-in function
+  object that is defined once for each realm.
+---*/
+
+var ThrowTypeError = Object.getOwnPropertyDescriptor(function(){ "use strict"; return arguments; }(), "callee").get;
+
+// Test with unmapped arguments object from strict-mode function.
```
leobalter: this comment - as in the other unique-per-realm tests - could be part of the description. It is even ok to have more than a single line description if necessary.

PR #526, opened 2016-03-01
Add missing surface tests for %TypedArray%.prototype.includes

leobalter:


leobalter:
pinging @jugglinmike @goyakin

jugglinmike:
The new `TypedArray.prototype.toString` tests aren't necessary--Test262 maintains equivalent tests against `String.prototype.toString`, and a single test for equality of the two methods in `built-ins/TypedArray/toString.js`. You could test in *either* location, but it's probably a good idea to mimic the organization of the spec (where the algorithm is defined for the String.prototype property and simply referenced by the TypedArray.prototype property. It might be worth removing the other previously-existing tests in that directory, too.

leobalter:
I removed the commit for .toString and renamed the branch

> It might be worth removing the other previously-existing tests in that directory, too.

Do you mean the tests for length and name?

jugglinmike:
Yessir

leobalter:
I've opened so many PRs on TypedArrays that I need to confirm if I did it anywhere else or if I can simply remove that folder.

jugglinmike:
Thanks!

PR #527, opened 2016-03-01
Add tests for detached buffer validation on TypedArray methods

leobalter:
these tests requires an extra implementation on the runners. Ref `harness/detachArrayBuffer.js`

### DIFF
--- a/test/built-ins/TypedArray/prototype/buffer/detached-buffer.js
+++ b/test/built-ins/TypedArray/prototype/buffer/detached-buffer.js
@@ -1,0 +1,20 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-get-%typedarray%.prototype.buffer
+description: The getter method does not throw with a detached buffer
+info: >
+  22.2.3.1 get %TypedArray%.prototype.buffer
+
+  ...
+  4. Let buffer be the value of O's [[ViewedArrayBuffer]] internal slot.
+  5. Return buffer.
+includes: [testTypedArray.js, detachArrayBuffer.js]
+---*/
+
+testWithTypedArrayConstructors(function(TA) {
+  var buffer = new ArrayBuffer(8);
+  var sample = new TA(buffer, 0, 1);
+  $DETACHBUFFER(sample.buffer);
+  assert.sameValue(sample.buffer, buffer);
+});
--- a/test/built-ins/TypedArray/prototype/join/detached-buffer.js
+++ b/test/built-ins/TypedArray/prototype/join/detached-buffer.js
@@ -19,7 +19,7 @@ includes: [testTypedArray.js, detachArrayBuffer.js]
 ---*/
 
 var obj = {
-  valueOf() {
+  toString: function() {
     throw new Test262Error();
   }
 };
--- a/test/built-ins/TypedArray/prototype/join/detached-buffer.js
+++ b/test/built-ins/TypedArray/prototype/join/detached-buffer.js
@@ -1,0 +1,33 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-%typedarray%.prototype.join
+description: Throws a TypeError if this has a detached buffer
+info: >
+  22.2.3.15 %TypedArray%.prototype.join ( separator )
+
+  This function is not generic. ValidateTypedArray is applied to the this value
+  prior to evaluating the algorithm. If its result is an abrupt completion that
+  exception is thrown instead of evaluating the algorithm.
+
+  22.2.3.5.1 Runtime Semantics: ValidateTypedArray ( O )
+
+  ...
+  5. If IsDetachedBuffer(buffer) is true, throw a TypeError exception.
+  ...
+includes: [testTypedArray.js, detachArrayBuffer.js]
+---*/
+
+var obj = {
+  valueOf() {
+    throw new Test262Error();
+  }
+};
+
+testWithTypedArrayConstructors(function(TA) {
+  var sample = new TA(1);
+  $DETACHBUFFER(sample.buffer);
+  assert.throws(TypeError, function() {
+    sample.join(obj);
+  });
+});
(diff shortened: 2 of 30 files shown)


leobalter:
I rearranged the commits to isolate the harness/detachArrayBuffer.js file in a single commit.

There are also rebased with the last changes on master.

test/built-ins/TypedArray/prototype/buffer/detached-buffer.js
```diff
@@ -0,0 +1,21 @@
...
+info: >
+  22.2.3.1 get %TypedArray%.prototype.buffer
+
+  ...
+  4. Let buffer be the value of O's [[ViewedArrayBuffer]] internal slot.
+  5. Return buffer.
+includes: [testTypedArray.js, detachArrayBuffer.js]
+---*/
+
+var buffer = new ArrayBuffer();
```
jugglinmike: This invocation needs to specify a valid buffer size.

jugglinmike:
We want *something* defined at `harness/detachArrayBuffer.js` by default--that
gives us an opportunity to communicate the requirement more clearly
(implementers will see your error message, not an OS-level warning about
non-existent files). Instead of testing for `ArrayBuffer.transfer`, though, we
should use some future-proof reference--maybe `Test262DetachArrayBuffer`. It's
doesn't really matter what we choose; any reference that is unlikely to be
standardized will do. In addition, an in-line comment that documents the
expected behavior of the function will be helpful.

Speaking of which, I think the usage of the helper is incorrect here. Every
invocation specifies a TypedArray as the first argument, but the name of the
helper (and the implementation provided) strongly suggest that it should be
invoked with an ArrayBuffer instance.

Up until now, we've been using the `built-ins/TypedArray` directory for tests
that do *not* use any particular TypedArray subclass. The tests that assert the
behavior of all the subclasses together (using
`testWithTypedArrayConstructors`) are generally located in
`built-ins/TypedArrays`. The placement of these new files contradicts that
pattern, though. Is there a specific reason to do this?

test/built-ins/TypedArray/prototype/buffer/detached-buffer.js
leobalter: I almost got it valid! https://github.com/tc39/ecma262/pull/410 :)

test/built-ins/TypedArray/prototype/buffer/detached-buffer.js
```diff
@@ -0,0 +1,21 @@
...
+
+testWithTypedArrayConstructors(function(TA) {
+  var sample = new TA(buffer);
+  $DETACHBUFFER(sample);
```
jugglinmike: Also, it might be a good idea to create a new buffer for each iteration. Implementations might define guards against detaching an already-detached buffer (es6draft does this). It's tricky because we can't account for all behaviors of a non-standard function, but it seems like this is a likely error we could avoid easily.

leobalter:
> we should use some future-proof reference--maybe Test262DetachArrayBuffer

Isn't the `$DETACHBUFFER` enough for this reference? I could remote the calls for `ArrayBuffer.transfer` inside, but we don't need to use another reference inside.

> Speaking of which, I think the usage of the helper is incorrect here. Every
invocation specifies a TypedArray as the first argument, but the name of the
helper (and the implementation provided) strongly suggest that it should be
invoked with an ArrayBuffer instance.

sure thing

> The placement of these new files contradicts that pattern, though. Is there a specific reason to do this?

The tests we have on `built-ins/TypedArrays/prototype/*` are checking only they inherit from `TypedArray.prototype`. All the other tests are currently being addressed to `built-ins/TypedArray/prototype` as I'm testing the methods, regardless of the constructor, but I'm indeed using samples of each constructor designed to be used with them.

test/built-ins/TypedArray/prototype/buffer/detached-buffer.js
leobalter: f483954

jugglinmike:
> Isn't the $DETACHBUFFER enough for this reference? I could remote the calls for ArrayBuffer.transfer inside, but we don't need to use another reference inside.

Actually, yes. I was over-thinking the amount of indirection we'd need. Let's
remove ArrayBuffer.transfer, though, and throw in all cases (despite being a
likely solution, ArrayBuffer.transfer may still change in ways that invalidate
our usage here).

> I'm indeed using samples of each constructor designed to be used with them.

Sounds good to me

leobalter:
> Let's remove ArrayBuffer.transfer, though, and throw in all cases

Doing it. It's funny how this is a 100% guaranteed way to land valid but failing tests.

---

fixing a typo from a past comment as it bugs me: *I could __remove__ the calls for `ArrayBuffer.transfer`* (s/remote/remove)

leobalter:
done. 5441ac3

test/built-ins/TypedArray/prototype/join/detached-buffer.js
```diff
@@ -0,0 +1,33 @@
...
+  22.2.3.5.1 Runtime Semantics: ValidateTypedArray ( O )
+
+  ...
+  5. If IsDetachedBuffer(buffer) is true, throw a TypeError exception.
+  ...
+includes: [testTypedArray.js, detachArrayBuffer.js]
+---*/
+
+var obj = {
+  valueOf() {
```
jugglinmike: This is a good idea, but it doesn't serve the intended purpose. The spec calls ToString, so the method to define here is `toString`.

test/built-ins/TypedArray/prototype/lastIndexOf/detached-buffer.js
```diff
@@ -0,0 +1,33 @@
...
+  22.2.3.5.1 Runtime Semantics: ValidateTypedArray ( O )
+
+  ...
+  5. If IsDetachedBuffer(buffer) is true, throw a TypeError exception.
+  ...
+includes: [testTypedArray.js, detachArrayBuffer.js]
+---*/
+
+var searchElement = {
+  valueOf() {
```
jugglinmike: `lastIndexOf` operates according to strict equality, so this method won't be called even when the buffer is *not* detached.

leobalter:
inline comments fixed at d806df0

jugglinmike:
Merged to `master` at commits 70c7375be83137015da2c5d6d29a3c40ad3cb5b1 and ce503e638f3e018d493f97d7e2284c0137f5d39e. Thanks!

PR #531, opened 2016-03-09
Add tests for `debugger` statement

jugglinmike:
The runtime semantics of this statement are host-defined and therefore
untestable, but the statement's affect on the formal grammar should be
consistent across all implementations.


leobalter:
This is great!

PR #532, opened 2016-03-09
Add test for evaluating Function#toString's output

jugglinmike:


michaelficarra:
@jugglinmike You can probably close this, as the committee has decided that the currently specified semantics are undesirable. But you could also wait until https://github.com/tc39/test262/pull/553 is merged to be safe.

jugglinmike:
Since implementation reality makes adhering to this language infeasible, and since @michaelficarra's patch for the improved specification has landed (gh-553), I'm happy to close this without merging.

PR #534, opened 2016-03-10
Removes special handling for UTC timezone in toISOString/15.9.5.43-0-1…

mythrialle:
…3.js

### DIFF
--- a/test/built-ins/Date/prototype/toISOString/15.9.5.43-0-13.js
+++ b/test/built-ins/Date/prototype/toISOString/15.9.5.43-0-13.js
@@ -8,11 +8,8 @@ description: >
     date is outside the valid range of time. 
 ---*/
 
-        var date, dateStr;
-
+  // As specified in ES5 15.9.1.14, time > 8.64e15 is not in the valid range.
+  var date = new Date(8.64e15 + 1);
 assert.throws(RangeError, function() {
-                // As specified in ES5 15.9.1.14, time > 8.64e15 is not in the
-                // valid range.
-                date = new Date(8.64e15 + 1);
-                dateStr = date.toISOString();
+                date.toISOString();
 });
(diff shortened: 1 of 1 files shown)


jugglinmike:
Thanks for the report and the patch! You're right; this test is pretty wonky. I think we can simplify the fix a little more, though. What do you think about specifying a millisecond value to the Date constructor? This value will not be interpreted according to the local system time, obviating the need to adjust according to `Date.prototype.getTimezoneOffset`.

You could get the millisecond value through `Date.UTC`, or (even better) hard-coding the value `8.64e15 + 1` (as specified in [ES5 15.9.1.14](https://es5.github.io/#x15.9.1.14)).

mythrialle:
Thanks for the suggestion. It sounds good. I updated the branch. I also updated the comment that we are using absolute value of time. Please take a look and let me know any comments.

jugglinmike:
Great! There's just one more thing: the Date constructor invocation is inside
the callback function for `assert.throws`. It's possible that an implementation
could incorrectly throw a RangeError when creating a Date instance like this,
and as written, this test would still pass in that case. Would you mind moving
the constructor invocation outside of the assertion's callback? (Also, while
we're cleaning up: there's no need for the `dateStr` variable.)

I know this isn't new from your previous submission, so my apologies for not
mentioning this sooner.

mythrialle:
Thanks. Sorry, I missed the date constructor last time. Fixed both of them. Let me know if its ok.

jugglinmike:
This looks great! The last step (and another thing I almost forgot) is to sign the CLA: https://tc39.github.io/test262-cla/ . If you've already done so, then we don't need anything further.

@bterlson @goyakin Can either of you verify the signature here? I don't have access to the database.

goyakin:
@mythrialle Thanks for signing the CLA.

PR #536, opened 2016-03-10
Add tests for the internal methods of Integer Indexed exotic objects

leobalter:


harness/detachArrayBuffer.js
```diff
@@ -0,0 +1,11 @@
+function $DETACHBUFFER(buffer) {
+  if (ArrayBuffer.transfer) {
+    ArrayBuffer.transfer(buffer);
+  } else {
+    try {
+      eval("%ArrayBufferNeuter(buffer)");
```
jugglinmike: This is implementation-specific code, so it probably doesn't belong in Test262

test/built-ins/TypedArrays/internals/DefineOwnProperty/detached-buffer.js
```diff
@@ -0,0 +1,124 @@
...
+
+  assert.sameValue(
+    Reflect.defineProperty(sample, "-0", desc),
+    false,
+    "Return false before Detached Buffer check when value is -0"
+  );
+
+  assert.sameValue(
+    Reflect.defineProperty(sample, "0", {
+      configurable: true
```
jugglinmike: In order to really prove that the order of operations are being observed, these should specify a `value` field.

test/built-ins/TypedArrays/internals/DefineOwnProperty/key-is-greater-than-last-index.js
```diff
@@ -0,0 +1,58 @@
...
+      configurable: false,
+      enumerable: true,
+      writable: true
+    }),
+    false,
+    "numericIndex > length"
+  );
+
+  assert.sameValue(
+    Reflect.defineProperty(sample, "Infinity", {
```
jugglinmike: This assertion does not demonstrate the semantics under test. It belongs in `DefineOwnProperty/key-is-not-integer.js`

test/built-ins/TypedArrays/internals/DefineOwnProperty/key-is-lower-than-zero.js
```diff
@@ -0,0 +1,45 @@
...
+      configurable: true,
+      enumerable: true,
+      writable: true
+    }),
+    false,
+    "-1"
+  );
+
+  assert.sameValue(
+    Reflect.defineProperty(sample, "-Infinity", {
```
jugglinmike: This assertion does not demonstrate the semantics under test. It belongs in `DefineOwnProperty/key-is-not-integer.js`

test/built-ins/TypedArrays/internals/DefineOwnProperty/key-is-numericindex.js
```diff
@@ -0,0 +1,42 @@
...
+      writable: true
+    }),
+    true
+  );
+
+  assert.sameValue(sample[0], 8, "property value was set");
+  var desc = Object.getOwnPropertyDescriptor(sample, "0");
+
+  assert.sameValue(desc.value, 8, "desc.value");
+  assert.sameValue(desc.writable, true, "property is writable");
```
jugglinmike: This should use the `verifyWritable` helper function

test/built-ins/TypedArrays/internals/DefineOwnProperty/key-is-symbol.js
```diff
@@ -0,0 +1,54 @@
...
+      enumerable: true
+    }),
+    true,
+    "return true after defining property"
+  );
+
+  var desc = Object.getOwnPropertyDescriptor(sample, s2);
+  assert.sameValue(desc.get, fnget, "accessor's get");
+  assert.sameValue(desc.set, fnset, "accessor's set");
+  assert.sameValue(desc.enumerable, true);
```
jugglinmike: This assertion should be expressed with the property helper methods (so you'll have to file that other patch first :) )

test/built-ins/TypedArrays/internals/Get/detached-buffer-key-is-not-numeric-index.js
```diff
@@ -0,0 +1,27 @@
...
+includes: [testTypedArray.js, detachArrayBuffer.js]
+---*/
+
+testWithTypedArrayConstructors(function(TA) {
+  var sample = new TA([42, 43]);
+  $DETACHBUFFER(sample.buffer);
+
+  assert.sameValue(sample.undef, undefined);
+
+  sample.foo = "test262";
```
jugglinmike: It's probably safer to set the `foo` property before detaching the buffer, since (incorrect) errors resulting from "setting on a detached buffer" would be confusing in this test.

test/built-ins/TypedArrays/internals/GetOwnProperty/index-prop-desc.js
```diff
@@ -0,0 +1,36 @@
...
+
+  var desc0 = Reflect.getOwnPropertyDescriptor(sample, 0);
+  assert.sameValue(desc0.value, 42, "value", "desc0.value === 42");
+  assert.sameValue(desc0.writable, true, "index descriptor is writable [0]");
+  verifyEnumerable(sample, "0", "index descriptor is enumerable [0]");
+  verifyNotConfigurable(sample, "0", "index descriptor is not configurable [0]");
+
+  var desc1 = Reflect.getOwnPropertyDescriptor(sample, 1);
+  assert.sameValue(desc1.value, 43, "value", "desc1.value === 43");
+  assert.sameValue(desc1.writable, true, "index descriptor is writable [1]");
```
jugglinmike: `verifyWritable` here

test/built-ins/TypedArrays/internals/HasProperty/abrupt-from-ordinary-has-parent-hasproperty.js
```diff
@@ -0,0 +1,63 @@
...
+    "OrdinaryHasProperty does not affect numericIndex properties [0]"
+  );
+  assert.sameValue(
+    Reflect.has(sample, 1), false,
+    "OrdinaryHasProperty does not affect numericIndex properties [1]"
+  );
+
+  assert.throws(Test262Error, function() {
+    Reflect.has(sample, "foo");
+  }, "");
```
jugglinmike: This looks like a mistakenly-omitted assertion message

jugglinmike:
Hey Leo!

I've just finished reviewing this patch. I left some specific suggestions
in-line. Here are a few additional thoughts that don't apply to any one line:

- Some (but not all) of these tests take the invariants of the internal methods
  for granted. It's not enough to verify that a given `Reflect` method returns
  the expected value; we'll have to also assert that we can observe the
  expected side effect. For example:

  ```diff
     var sample = new TA(2);
   
     assert.sameValue(
       Reflect.defineProperty(sample, "0", {
         value: 42,
         configurable: true,
         enumerable: true,
         writable: true
       }),
       false
     );
  +  assert.sameValue(sample[0], 0);
  ```
- There are some cases of ES6 syntax in these tests. Because TypedArrays are
  themselves an ES6 feature, I think it's worthwhile to avoid this. Doing so
  will allow implementers to access these tests without the syntax extensions,
  which is relevant for VMs that are incrementally adding support for ES6, but
  also for libraries that shim typed arrays.
- It doesn't look like the tests for [[DefineOwnProperty]] assert that
  IntegerIndexedElementSet (not just Set) is called. We should include tests
  that ensure expected behavior for non-integer values and integer values that
  are out-of-bounds. The tests for [[Set]] guard against calls to OrdinarySet,
  but they would also benefit from assertions regarding invalid values.
- I think there are some interesting tests to write regarding
  CanonicalNumericIndexString. For instance, in
  HasProperty/detached-buffer-key-is-not-number.js, you assert that no error is
  thrown for a detached buffer when the input string is non-numeric. A slightly
  more nuanced details is when the string is numeric, but not the canonical
  representation: `Reflect.has(sample, '+1')`, `Reflect.has(sample, '1.0')`,
  `Reflect.has(sample, '0.0000001')` or `Reflect.has(sample, 1000000000000000000000)`
- The following tests will pass even if OrdinaryHasProperty is used.

  - HasProperty/key-is-greater-than-last-index.js
  - HasProperty/key-is-lower-than-zero.js
  - HasProperty/key-is-minus-zero.js
  - HasProperty/key-is-not-integer.js

  To avoid false positives, they should define a property on the prototype.
- Some tests modify the TypedArray prototype, while others re-set the
  [[Prototype]] of individual Integer Indexed exotic objects. Although the
  second approach is necessary when you need to use a Proxy trap, I think the
  first approach is preferable in all other cases because it requires fewer new
  language features.

harness/detachArrayBuffer.js
leobalter: I forgot to flag this file as a "please do not merge this commit" as this file is also introduced on #527. I'll fix it on the other PR, or better: maybe have an individual PR for this file required by these PRs.

test/built-ins/TypedArrays/internals/DefineOwnProperty/detached-buffer.js
leobalter: it returns `true` if `value` field is not present, before the check for the detachedBuffer:

```
9.4.5.3 [[DefineOwnProperty]] ( P, Desc)
...
xi. If Desc has a [[Value]] field, then
    1. Let value be Desc.[[Value]].
    2. Return ? IntegerIndexedElementSet(O, intIndex, value).
xii. Return true.
```

I've set a value field to throw a custom error in case it's evaluated before with ToNumber.

jugglinmike: I'm saying that by specifying a `value` field, we can be certain that step 3.b.viii comes before step 3.b.xi

leobalter: afa2c0f

test/built-ins/TypedArrays/internals/DefineOwnProperty/key-is-greater-than-last-index.js
leobalter: 4d20de0

test/built-ins/TypedArrays/internals/DefineOwnProperty/key-is-lower-than-zero.js
leobalter: 4d20de0

test/built-ins/TypedArrays/internals/DefineOwnProperty/key-is-numericindex.js
leobalter: this helper is not helpful with TypedArray objects.

leobalter: instead of sending a custom value to `verifyWritable` to test it is really writable it seems more reasonable to check only the descriptor. On cases like this, I already have a specific test to set a new value.

test/built-ins/TypedArrays/internals/DefineOwnProperty/key-is-symbol.js
leobalter: waiting for #543

test/built-ins/TypedArrays/internals/Get/detached-buffer-key-is-not-numeric-index.js
leobalter: ebbe80f

test/built-ins/TypedArrays/internals/GetOwnProperty/index-prop-desc.js
leobalter: same problem as in https://github.com/tc39/test262/pull/536#discussion_r55903190

test/built-ins/TypedArrays/internals/HasProperty/abrupt-from-ordinary-has-parent-hasproperty.js
leobalter: d091f3c

leobalter: a bug ate my comments

leobalter:
> Some (but not all) of these tests take the invariants of the internal methods
for granted. It's not enough to verify that a given Reflect method returns
the expected value; we'll have to also assert that we can observe the
expected side effect

d0e062a and c1e9ab7

leobalter:
> There are some cases of ES6 syntax in these tests. Because TypedArrays are themselves an ES6 feature, I think it's worthwhile to avoid this. Doing so will allow implementers to access these tests without the syntax extensions, which is relevant for VMs that are incrementally adding support for ES6, but also for libraries that shim typed arrays.

cbaea98, 	80e01dc, 	9e56604, 2eb036b, 87292a5

leobalter:
> I think there are some interesting tests to write regarding CanonicalNumericIndexString. For instance, in HasProperty/detached-buffer-key-is-not-number.js, you assert that no error is thrown for a detached buffer when the input string is non-numeric. A slightly more nuanced details is when the string is numeric, but not the canonical representation: Reflect.has(sample, '+1'), Reflect.has(sample, '1.0'), Reflect.has(sample, '0.0000001') or Reflect.has(sample, 1000000000000000000000) 

This is big and will require more time. I'll finish the other parts of the feedback first.

leobalter:
> The following tests will pass even if OrdinaryHasProperty is used. (....) To avoid false positives, they should define a property on the prototype.

 4c8c32e

leobalter:
> Some tests modify the TypedArray prototype, while others re-set the [[Prototype]] of individual Integer Indexed exotic objects. Although the second approach is necessary when you need to use a Proxy trap, I think the first approach is preferable in all other cases because it requires fewer new language features.

ca100dc and 684e44c

leobalter:
> I think there are some interesting tests to write regarding CanonicalNumericIndexString. For instance, in HasProperty/detached-buffer-key-is-not-number.js, you assert that no error is thrown for a detached buffer when the input string is non-numeric. A slightly more nuanced details is when the string is numeric, but not the canonical representation: Reflect.has(sample, '+1'), Reflect.has(sample, '1.0'), Reflect.has(sample, '0.0000001') or Reflect.has(sample, 1000000000000000000000) 

29a1e9a 4901010 26c3880 c87c2a9 3426748

leobalter:
I believe I've finished addressing the feedback.

I'll squash the current fixup commits on the next step (before merge or if I have more feedback).

test/built-ins/TypedArrays/internals/HasProperty/detached-buffer-key-is-symbol.js
```diff
@@ -0,0 +1,31 @@
...
+var s2 = Symbol("bar");
+
+testWithTypedArrayConstructors(function(TA) {
+  var sample = new TA([42, 43]);
+  Object.defineProperty(sample, s1, { value: "baz" });
+
+  $DETACHBUFFER(sample.buffer);
+
+  assert.sameValue(Reflect.has(sample, s1), false);
+  assert.sameValue(Reflect.has(sample, s2), true);
```
jugglinmike: The Symbol values should be switched for these assertions to hold

test/built-ins/TypedArrays/internals/DefineOwnProperty/key-is-not-canonical-index.js
```diff
@@ -0,0 +1,77 @@
...
+    assert.sameValue(
+      Reflect.defineProperty(sample, key, dataDesc),
+      true,
+      "return true after defining data property [" + key + "]"
+    );
+
+    assert.sameValue(sample[key], 42, "value is set to [" + key + "]");
+    verifyWritable(sample, key);
+    verifyConfigurable(sample, key);
+    // verifyNotEnumerable(sample, key);
```
jugglinmike: It seems like you should be able to assert that these properties are non-enumerable. Is that not true?

test/built-ins/TypedArrays/internals/DefineOwnProperty/detached-buffer.js
```diff
@@ -0,0 +1,134 @@
...
+    }),
+    false,
+    "Return false before Detached Buffer check with accessor descriptor"
+  );
+
+  assert.sameValue(
+    Reflect.defineProperty(sample, "0", {
+      configurable: false,
+      enumerable: true,
+      writable: false
```
jugglinmike: In order for this assertion to pass, "writable" should be set to `true`. Also, it's probably safer to use a different property name for each assertion--implementers who make a mistake here could see misleading errors and/or false positives if these tests try to re-set the same property descriptor (given that it is sometimes being set to `configurable: false`).

jugglinmike:
Hey Leo!

I just looked over your latest changes. I've left some feedback in-line, but I
also had a few comments that don't apply to any single line of code:

The new assertions in these files are incorrect

- DefineOwnProperty/key-is-minus-zero.js
- DefineOwnProperty/key-is-not-integer.js

...though I'm not sure of the intent. For example:

```js
  var sample = new TA([42, 43]);

  assert.sameValue(
    Reflect.defineProperty(sample, "0.1", {
      value: 42,
      configurable: true,
      enumerable: true,
      writable: true
    }),
    false,
    "0.1"
  );
  assert.sameValue(sample[0], 0, "'0.1' - does not change the value for [0]");
```

The value of `sample[0]` should be `42` both before and after the call to
`Reflect.defineProperty`. The assertion message correctly describes this
expectation, but the expected value does not. Changing the `0` to `42` will fix
the assertion, but you'll need to update the `value` of the rejected property
descriptor to be sure that it was actually rejected. It also seems like another
test for `sample["0.1"]` (in this case) would be relevant. What do you think?

---

These files use `"'0.0000001"` (among other values) for a property name.:

- Get/key-is-not-canonical-index.js
- GetOwnProperty/key-is-not-canonical-index.js
- Set/key-is-not-canonical-index.js
- HasProperty/key-is-not-canonical-index.js

This is not invalid, but it is also not really in-line with the "spirit" of the
tests--the value `"0.0000001"` is much more likely to catch implementation bugs
related to the CanonicalNumericIndexString operation.

leobalter:
Inline comments fixed on ce441ec

leobalter:
other fixes at 7d386d7 addressing 

- DefineOwnProperty/key-is-minus-zero.js
- DefineOwnProperty/key-is-not-integer.js

leobalter:
> These files use "'0.0000001" (among other values) for a property name.

That was a mistake, kudos to my "US-International" keyboard layout and my ignorance for a shortcut key to alternate it with a plain US layout while I'm writing code.

Fixed at ed58dd3

test/built-ins/TypedArrays/internals/DefineOwnProperty/key-is-not-canonical-index.js
```diff
@@ -0,0 +1,77 @@
...
+    assert.sameValue(
+      Reflect.defineProperty(sample, key, dataDesc),
+      true,
+      "return true after defining data property [" + key + "]"
+    );
+
+    assert.sameValue(sample[key], 42, "value is set to [" + key + "]");
+    verifyWritable(sample, key);
+    verifyConfigurable(sample, key);
+    verifyNotEnumerable(sample, key);
```
jugglinmike: `verifyConfigurable` is destructive, so this assertion will fail. Can you move it two lines up, just before `verifyWritable`?

test/built-ins/TypedArrays/internals/DefineOwnProperty/key-is-not-integer.js
```diff
@@ -0,0 +1,124 @@
...
+      enumerable: true,
+      writable: true
+    }),
+    false,
+    "0.1"
+  );
+  assert.sameValue(sample[0], 0, "'0.1' - does not change the value for [0]");
+  assert.sameValue(
+    sample["0.1"],
+    undefined,
```
jugglinmike: This line (and a few more in this file) has trailing whitespace

test/built-ins/TypedArrays/internals/DefineOwnProperty/key-is-not-canonical-index.js
leobalter: 24ffeed

test/built-ins/TypedArrays/internals/DefineOwnProperty/key-is-not-integer.js
leobalter: 8a331d4

jugglinmike:
This is looking good, Leo. I just went hunting for areas of missing coverage, and I found one small thing and one kind of larger thing:

**Small thing:** Extensibility

Step 11 of [IntegerIndexdObjectCreate](https://tc39.github.io/ecma262/#sec-integerindexedobjectcreate) sets the object's [[Extensible]] internal slot to "true". We'll want a test asserting that. As a bonus, you might add a test which attempts to make an Indexed Integer exotic object non-extensible. The spec allows this, but 5.3.0 incorrectly throws (we can rest easy, though: it's since been fixed in V8).

**Larger thing:** Value conversion

Currently, values of integer properties are selected to be acceptable for *any* TypedArray, e.g. positive integers between 0 and 127. This makes sense because you are iterating over all of the constructors in these tests, so if you choose a static value, that value has to behave consistently in all cases.

The trouble is, an important behavior of Typed Arrays is what happens at the boundaries defined for the given constructor. For example:

```js
print(new Uint8Array([128])[0]); // 128
print(new  Int8Array([128])[0]); // -128

print(new Uint8ClampedArray([-1])[0]); // 0
print(new        Uint8Array([-1])[0]); // 255
```

...although it's not *just* boundaries of on the integer scale:

```js
print(new Float64Array([1.1])[0]) // 1.1
print(new   Int32Array([1.1])[0]) // 1
```

It's easy to forget this because the spec language is factored in such a way as to make all the constructors seem identical. But really, all that indirection with the strange-looking "%TypedArray%" reference is to facilitate exactly this distinction.

So as written, these tests do not demonstrate the unique behavior of the internal methods as controlled by the constructor used to create them. To prove this, I made the following change to `harness/testWithTypedArrays.js`:

```diff
 /**
  * Array containing every typed array constructor.
  */
+function createFaker(Ctor) {
+  class Fake extends Int8Array {}
+  Fake.name = Ctor.name;
+  Fake.prototype.BYTES_PER_ELEMENT = Ctor.BYTES_PER_ELEMENT;
+  return Fake;
+}
 var typedArrayConstructors = [
-  Float64Array,
-  Float32Array,
-  Int32Array,
-  Int16Array,
-  Int8Array,
-  Uint32Array,
-  Uint16Array,
-  Uint8Array,
-  Uint8ClampedArray
+  createFaker(Float64Array),
+  createFaker(Float32Array),
+  createFaker(Int32Array),
+  createFaker(Int16Array),
+  createFaker(Int8Array),
+  createFaker(Uint32Array),
+  createFaker(Uint16Array),
+  createFaker(Uint8Array),
+  createFaker(Uint8ClampedArray)
 ];
```

That is: instead of iterating over each of the typed array constructors, iterate over a set of "fake" constructors that consistently produce instances of `Int8Array` but *report an incorrect element size*. All tests currently pass with this change applied.

The spec language that dictates this is [GetValueFromBuffer](https://tc39.github.io/ecma262/#sec-getvaluefrombuffer) and [SetValeInBuffer](https://tc39.github.io/ecma262/#sec-setvalueinbuffer). For the purposes of this patch, the relevant callers are [IndexedIntegerElementGet](https://tc39.github.io/ecma262/#sec-integerindexedelementget) and [IndexedIntegerElementSet](https://tc39.github.io/ecma262/#sec-integerindexedelementset).

Building up a testing methodology now will be helpful for you in your upcoming work on the Typed Array prototype methods--note how `set` and `slice` also use those internal operations. When the time comes to tests those methods, we'll want to assert the same semantics are observed.

leobalter:
> The spec allows this, but 5.3.0 incorrectly throws (we can rest easy, though: it's since been fixed in V8).

what is 5.3.0?

> Small thing: Extensibility

9772de5 and #557

leobalter:
> Larger thing: Value conversion

This is highly valuable. Would you mind if we cover this in a separate PR? I can even prioritize this work before the other Typedarray instance methods.

jugglinmike:
> what is 5.3.0?

Sorry: I meant to write **Node** 5.3.0.

> This is highly valuable. Would you mind if we cover this in a separate PR? I can even prioritize this work before the other Typedarray instance methods.

This pull request is a bit long in the tooth as it is. Since the tests in the current form are valid, I think it's fine to revise with a follow-up patch.

leobalter:
Ok, I've opened #558 to address it. I can rebase and squash if it's ready.

jugglinmike:
Merged to `master` at commit 3cb20b9df59daeb850588bffcfa0c2c880395e10. Thanks!

PR #537, opened 2016-03-11
Update S15.8.2.1_A2.js

Yaffle:


leobalter:
The ideal would be creating a new test file for other specific cases.

On https://tc39.github.io/ecma262/#sec-math.abs, we have a specific step for -0, so a check for +0 would part of: "Returns the absolute value of x; the result has the same magnitude as x but has positive sign." instead of "If x is -0, the result is +0."

Plus, we could have some better coverage for this method with other values.

Yaffle:
@leobalter 
Why are there not a specific step for `-0` and `-Infinity`?

leobalter:
> Why are there not a specific step for -0 and -Infinity?

I can't follow your question.

The main step of Math.abs is `Returns the absolute value of x;`. There are specific cases for specific values, like `-0` (returning `+0`) and `-Infinity` returning `Infinity`. 

What I'm suggesting for this PR is to create a new file to test the +0 instead of mixing it with the tests for -0.

Yaffle:
@leobalter , you said:
>On https://tc39.github.io/ecma262/#sec-math.abs, we have a specific step for -0, so a check for +0 would part of: "Returns the absolute value of x; the result has the same magnitude as x but has positive sign." 

"-0" also has a sign and magnitude so I do not see why it is specific, but "+0" is not.
Also I cannot understand why you like to create so many files in this test suite. It is a little harder to open all files using github to see all their contents.

leobalter:
> Also I cannot understand why you like to create so many files in this test suite. It is a little harder to open all files using github to see all their contents.

This is a historical convention on this project, changing the convention requires extra work where we need consensus and bandwidth to change the project structure. If a tool (GitHub) is not the best to browser the files, it's easier to fix the tool or use another one.

Chapter [6.1.6](https://tc39.github.io/ecma262/#sec-ecmascript-language-types-number-type) specifies the differences from `-0` and `+0`.

Yaffle:
OK, will close this request; exsume me for spending your time and for trash pull request

Yaffle:
Chrome Mobile 30 and Konqueror 4.13 have this bug.
But you cannot include the check in test262. Great!

leobalter:
> Chrome Mobile 30 and Konqueror 4.13 have this bug.
> But you cannot include the check in test262. Great!

You can file a bug on their respective bug trackers.

I only asked you to follow the current conventions we have on this project. Your contribution is very appreciated when you make it constructive.

Yaffle:
@leobalter , 
But I think, the test for `+0` should not be in a separate file, because `+0` case is also a special case as `-0` and `-Infinity`, the Java docs ( https://docs.oracle.com/javase/8/docs/api/java/lang/Math.html#abs-double ), for example, have `+0` and `+Infinity` as special cases too.

PR #539, opened 2016-03-11
Update meta data: `id` to `esid`

jugglinmike:
The project's expected frontmatter tag name changed while these files
were under review.


leobalter:
LGTM

Issue #529, opened 2016-03-03
test262 still contains invalid tests for precision of Math operations


littledan:
built-ins/Math/cos/S15.8.2.7_A7
built-ins/Math/sin/S15.8.2.16_A7
built-ins/Math/tan/S15.8.2.18_A7
built-ins/Math/exp/S15.8.2.8_A6

We discussed these tests at the July 2015 TC39 meeting. There was agreement that they are invalid for the current spec text, but @sunfishcode might champion spec changes here. Until then, though, all that's valid to test for are the exact boundary conditions listed in the specification.

littledan:
@goyakin @jugglinmike What do you think about these tests?

jugglinmike:
I agree that those tests ought to be removed. I'm happy to file a pull request
that does this, but first: there is another set of suspicious-looking tests for
the trig functions:

- [test/built-ins/Math/cos/S15.8.2.7_A6.js](https://github.com/tc39/test262/blob/007d3b1eb76ccc7760f641d8e956e5da838b586f/test/built-ins/Math/cos/S15.8.2.7_A6.js)
- [test/built-ins/Math/sin/S15.8.2.16_A6.js](https://github.com/tc39/test262/blob/007d3b1eb76ccc7760f641d8e956e5da838b586f/test/built-ins/Math/sin/S15.8.2.16_A6.js)
- [test/built-ins/Math/tan/S15.8.2.18_A6.js](https://github.com/tc39/test262/blob/007d3b1eb76ccc7760f641d8e956e5da838b586f/test/built-ins/Math/tan/S15.8.2.18_A6.js)

It seems to me that "an implementation-dependent approximation" may not
necessarily be periodic to the (seemingly-arbitrary) precision asserted by
those tests.

Should we remove those files, as well?

littledan:
Yes, these seem invalid to me too (v8 just happens to pass them :)

jugglinmike:
Yesterday I added still more tests to the chopping block in gh-540. We are very limited in what we can assert according to the current spec text, but I would be interested in contributing to an effort to improve the specification in this regard. @sunfishcode do you have any interest in work like that?

jugglinmike:
Resolved via gh-540.

@sunfishcode I'd still be interested in working with you on the spec here--even if you could spare a little time to set me off on the right foot!

sunfishcode:
@jugglinmike I unfortunately don't have time to do any spec work for this myself right now.

[...]

PR #540, opened 2016-03-11
Remove over-specified tests

jugglinmike:
Except for a small set of expected input/output pairs, both ES5 and
ES2015 define the expected return value of these methods in terms of an
"implementation-dependent approximation." This makes it inappropriate to
enforce expectations for specific values, even if expressed imprecisely.


jugglinmike:
This resolves gh-529.

jugglinmike:
I've removed more over-specified tests. I've also added new tests for the Math
constants that assert their type (which I believe is the most Test262 can do
right now).

test/built-ins/Math/E/value.js
```diff
@@ -0,0 +1,15 @@
...
+es6id: 20.2.1.1
+info: >
+    The Number value for e, the base of the natural logarithms, which is
+    approximately 2.7182818284590452354.
+
+    The precision of this approximation is host-defined.
+---*/
+
+assert.sameValue(typeof Math.E, 'number');
+assert.notSameValue(Math.E, NaN);
```
leobalter: can we at least assert something like `2.7 < Math.E < 2.8` ? 

It's hard to say this is valid because approximation is relative. 3 is way closer to Math.E than 1000.

leobalter:
This is merged on b698c8b3ab2980cc1282cb08796d0b76337bf7d9 and f7aa31b41faee25823bb38b7ca53b104cf53debd

Thanks

PR #541, opened 2016-03-11
Add syntax tests for declaration restrictions

jugglinmike:
Assert that declarations are not permitted in positions reserved for
statements only.


leobalter:
This is great! LGTM

PR #542, opened 2016-03-14
Encode Expected Error "stage"

jugglinmike:
Update "negative" tests to include information on the expected stage of evaluation. My goal with this change is to make Test262 more easily consumable by JavaScript parsers.

This changeset includes:

- Changes to existing "negative" tests for SyntaxErrors. The documented advice
  for testing exceptions is to prefer `assert.throws` wherever possible. As
  noted in the first commit message, there were some lingering files that
  expressed runtime errors using the `negative` frontmatter.
- Updates to the project's custom YAML parser. Some dependents (e.g. V8) use
  this Python module directly.
- Updates the Python "console runner." Although the Python console runner has
  been formally deprecated as per TC-39's wishes, it has not been removed.
  Until such time as it *is* removed, it should continue to function with the same input.
- Removal of the `NotEarlyError` object (and associated documentation). By
  encoding error stage information in test metadata, test runners can enforce
  this concern. This lessens the amount of domain-specific knowledge required
  to contribute to Test262 and avoids inconsistencies between negative tests.


jugglinmike:
@jmdyck This is a "first pass" solution for gh-196. It does not make a distinction between lexical, syntactic, and early errors, but I thought it would be a good starting point. It lowers the barrier for those distinctions to be made in the future; with this in place, extending support will require less infrastructure/documentation work.

leobalter:
Can we change `stage` to be optional with `early` as the default value?

jmdyck:
@jugglinmike Thanks! So the next step would be to change "early" in test metadata to "lexical" or "syntactic" (or leave it as "early") as appropriate?

jugglinmike:
Yup. I guess "early" is a bit misleading as a catch-all value, because only a subset of the current negative tests describe early errors. Do you have a suggestion for a better name to use in the mean time? Maybe `stage: parse`?

jmdyck:
`parse` isn't great as a catch-all, because the spec doesn't view early error detection as part of parsing. If you just want a placeholder to use until we know (for each test) what the proper value of `stage` should be, then I'd be inclined to use something like "?" or "TBD".

leobalter:
the "early" value seems better than ? or TBD, even if it's not perfect, it might help on the migration to the next plans as well.

I believe the changes are good and it solves the problem on the project's test runner but it's good to check how it works on other runners. @littledan, @bterlson would you provide some feedback here, please?

bterlson:
It is not supported but it can be once this PR is accepted (fairly trivial). Am fine with using "early" for now.

littledan:
The infrastructure changes should be fine for us to implement.

Not: the word 'stage' makes me think of TC39 stages, which might make sense to represent in frontmatter. How about phase?

jugglinmike:
Technically speaking, "stage" as a nested property of the "negative" attribute would not collide with a future top-level "stage" property. But of course, it wouldn't be any less confusing!

I don't expect to have time to make this change until this weekend, but here are my plans:

- Stick with the generic "early" value for now
- Change "stage" to "phase"
- Update to include all "negative" tests that have landed since this pull request was originally authored

If there are any further requests for changes, please let me know. I'm not able to advance this work very quickly, so I'd like to minimize round trips in the review process where possible

jugglinmike:
I felt the requested changes were substantial enough to warrant a new pull request, so I've opened gh-655 to continue the discussion. Thanks for your help so far, everyone!

PR #543, opened 2016-03-14
Fix verifyEnumerable helper to account for properties with Symbol values

leobalter:


test/harness/propertyhelper-verifyenumerable-not-enumerable.js
```diff
@@ -6,26 +6,52 @@ description: >
...
+  }
+}
+
+if (threwStringKey === false) {
+  throw new Test262Error(
+    'Expected a Test262Error, but no error was thrown for string key.'
+  );
+}
+
+if (threwStringKey === false) {
```
jugglinmike: This should be `threwSymbolKey`

test/harness/propertyhelper-verifynotenumerable-enumerable.js
```diff
@@ -26,6 +32,26 @@ try {
...
+  }
+}
+
+if (threwStringKey === false) {
+  throw new Test262Error(
+    'Expected a Test262Error, but no error was thrown for string key.'
+  );
 }
+
+if (threwStringKey === false) {
```
jugglinmike: Same here--this should be `threwSymbolKey`

jugglinmike:
Generally, I think it would be better to make these new assertions from separate files. Doing so will improve readability (since independent assertions will not be interleaved) but more importantly, it will allow hosts that do not implement Symbol to verify the correctness of these helper functions.

leobalter:
Thanks for the feedback, I applied the suggestion on the rebased commit.

leobalter:
can we merge this?

PR #544, opened 2016-03-14
assert.throws fails if second argument is not a function

leobalter:


jugglinmike:
Merged to `master` at commit e7f6cd7f867ad02de59181d37267b9bb40bb3c5e. Thanks!

PR #545, opened 2016-03-16
Introduce test generation tool

jugglinmike:
This is a continuation of the work from gh-470 and gh-481. Please reference
those issues for background on the design of this proposal.

As noted in the documentation, the input for procedurally-generated tests is
stored in a new top-level directory named `src/`. Tests are generated by
invoking the cross-platform `make.py` script, and placed in the `test/`
directory (alongside the human-authored test files). The generated files are
expected to be checked in to the repository.

This patch introduces a new feature in addition to what we discussed in gh-481:
a `-p`/`--parents` flag. This functions in the same way as it does in the Unix
`mkdir` utility: without it, the generation tool will refuse to create new
directories (which it previously did silently and by default). We sometimes
rename subdirectories of `test/`. When we do this, we risk forgetting to change
one or more generated test's targets, thereby defining related files in two
separate directories. By making automatic directory creation "off by default,"
the generation tool will fail until we have updated all source files that
target the "old" directory.

I've also included a small set of test cases/templates for demonstration
purposes.


bterlson:
This seems good to me. Also, I want to get this coverage so unless @goyakin has concerns this is good to go IMO.

leobalter:
I'll merge this, the generation tool can be improved and it doesn't break anything. That opens the path for #546 (Automate deployment)

PR #548, opened 2016-03-17
Avoid false positives

jugglinmike:
As written, the test for `Math.random` would pass if the runtime
erroneously produced a non-numeric value. Add the necessary assertions
to guard against this case.


leobalter:
LGTM

Issue #519, opened 2016-02-24
RFC: Testing Module Semantics


jugglinmike:
@aklein @boingoing and @syg I understand you are all working on ES2015 module implementations, so if you have a moment, your input would be especially helpful here. Thanks!

Because ES2015 intentionally omits necessary details around module resolution (see [HostResolveImportedModule](https://tc39.github.io/ecma262/#sec-hostresolveimportedmodule)), functional tests cannot be written based on the specification alone.

In order to author meaningful tests, Test262 contributors and runtime implementors have to come to an agreement for additional semantics. It should be clear that conformance to this contract is a prerequisite only for Test262 and not for ECMAScript compliance generally.

[I recently proposed a new document named `INTERPRETNG.md`](https://github.com/tc39/test262/pull/508) to house exactly this kind of information. If that is accepted, we can extend it with the following text:

> **Modules.** Test262 includes tests for ECMAScript 2015 module code. In order to interpret these tests, consumers must expose an implementation of [HostResolveImportedModule](https://tc39.github.io/ecma262/#sec-hostresolveimportedmodule) that conforms to the following specification:
> - The normal return value must be an instance of [a Source Text Module
>   Record](https://tc39.github.io/ecma262/#sec-source-text-module-records).
> - If a Module Record corresponding to the pair referencingModule, specifier
>   does not exist or cannot be created, a `SyntaxError` must be thrown.

But the most contentious detail (as I see it) is resolving module specifiers. The design should be as simple as possible in order to limit the amount of custom behavior necessary to execute the tests. We can take a cue from common file system/web conventions to produce a design that can work out-of-the-box for many consumers.

Here's what I have in mind:

> All module specifiers used by Test262 begin with the character sequence `./`. The remaining characters should be interpreted as the name of a file within the same directory as the file under test.
> 
> For example, consider a test file located at `test/language/import/nested/index.js` with the following contents:
> 
> ``` js
> import * as ns from './dep.js';
> ```
> 
> Implementors should attempt to resolve this module specifier by loading a file located at `test/language/import/nested/dep.js`.

I think this will give authors just enough power to test the semantics while promoting a healthy file structure, and it limits the imposition on consumers. One potential issue is that it means Test262 will define non-test files alongside test files. I can think of three ways around this:
1. Additional meta-data for files that do not describe tests with (e.g. `flags:
   [not-a-test]`)
2. Different resolution semantics: define a top-level "fixtures" directory, and
   specify that loaders should resolve all modules from there.
[...]

PR #553, opened 2016-03-21
add Function.prototype.toString tests, conforming to proposed revision

michaelficarra:
See the proposal at https://tc39.github.io/Function-prototype-toString-revision/

Tests still missing (because I don't know how to author them):

* line terminator normalisation
* unnamed hoistable and class declarations in export default position


michaelficarra:
For the missing line terminator normalisation test, I can just use 3 separate test files that each use different line endings but perform the same test, but I wasn't sure if that was the best way to do it.

leobalter:
The current tests look good so far, there are some minor details that can be addressed:

Some files are current suffixed buy numbers. They could be improved replacing the numbers by the current content meaning, like `test/built-ins/Function/prototype/toString/class-declaration-1.js` => `test/built-ins/Function/prototype/toString/class-declaration-implied-ctor.js` following the current file description.

Also, using `esid` tag instead of `es6id` on the frontmatter, their values can be `pending` while they not have a current reference on the specs.

For the line terminator, can we use references to `\r` and `\n` or their unicode scaped values? You've probably already tried or thought about it. I'll see if I may come with a more constructive suggestion.

> unnamed hoistable and class declarations in export default position

@jugglinmike is currently working on the tests for modules and using locally the test generation tool, that could become some added case to generate more files concerning export (and maybe import) positions.

leobalter:
Also, these tests does not seem to cover the same concern from #532 yet where it evaluates the result of toString's output and reuse it as a new fn. 

We could have one of these PRs covering the desirable results for that operation.

jugglinmike:
These look good!

> - unnamed hoistable and class declarations in export default position

I'm fairly confident that testing default export values will require module
loading, so I'd suggest holding off on those tests until we formalize the
requirements (first gh-508 and then gh-519). Once this lands, we can make a new
issue to track that so we don't forget.

> For the missing line terminator normalisation test, I can just use 3 separate
> test files that each use different line endings but perform the same test,
> but I wasn't sure if that was the best way to do it.

I think that is what I would do. The duplication is kind of a drag, but it seems
manageable for the cases we need to cover.

A few ideas for additional tests:

- Complex class heritage. Maybe a function literal?
- Non-simple parameters. Destructuring/defaults
- Additional unicode normalization. Tiny edge case, but what about using some
  superfluous leading zeros in the source--`\u{000062}`?

jugglinmike:
> We could have one of these PRs covering the desirable results for that operation.

I don't think this is necessary. Just like the spec language they concern, these tests supersede those in gh-532. Asserting the exact string value is far more explicit; once that is verified, any additional tests for evaluation would really be tests for `eval`, not `Function.prototype.toString`.

leobalter:
> any additional tests for evaluation would really be tests for eval, not Function.prototype.toString.

That's great, thanks!

michaelficarra:
Checklist for myself:

- [X] replace the numbers in filenames with something meaningful
- [ ] use `esid` tag instead of `es6id`
- [ ] create test files for line terminator normalisation
- [ ] complex class heritage
- [ ] non-simple parameter lists
- [X] leading zeroes in Unicode escape sequences

michaelficarra:
@jugglinmike Done. Let me know if you want me to squash the commits (I had been amending until recently anyway).

test/built-ins/Function/prototype/toString/class-declaration-explicit-ctor.js
```diff
@@ -0,0 +1,13 @@
...
+/*---
+esid: sec-runtime-semantics-bindingclassdeclarationevaluation
+description: Function.prototype.toString on a class declaration (explicit constructor)
+---*/
+
+/* before */class /* a */ A /* b */ extends /* c */ B /* d */ { /* e */ constructor /* f */ ( /* g */ ) /* h */ { /* i */ ; /* j */ } /* k */ m /* l */ ( /* m */ ) /* n */ { /* o */ } /* p */ }/* after */
+
+assert.sameValue(A.toString(), "class /* a */ A /* b */ extends /* c */ B /* d */ { /* e */ constructor /* f */ ( /* g */ ) /* h */ { /* i */ ; /* j */ } /* k */ m /* l */ ( /* m */ ) /* n */ { /* o */ } /* p */ }");
+
+function B{}
```
jugglinmike: Typo here. I think a class declaration would work, but so would adding a parameter list.

michaelficarra: Thanks. This used to be a class declaration, but I wanted to avoid the TDZ error. I will add a parameter list.

jugglinmike:
Perfect! I think this patch is now ready to go.

We only merge tests for proposals that have reached stage three, so I'm going
to mark this pull request with the labels "ready-to-land", "future", and
"waiting-approval". Hopefully things go smoothly for you next week; then you'll
be way ahead of the game for stage four!

@bterlson: you may want to coordinate with Michael to ensure that, depending on
merge order, either your patch at gh-479 or his patch here also introduces
tests for `toString` called on objects created from the various async function
productions.

michaelficarra:
@jugglinmike That is a good point. I opened a PR against the async functions proposal for compatibility: https://github.com/tc39/ecmascript-asyncawait/pull/90. I'll leave it to @bterlson to add the tests to #479 once that's merged.

michaelficarra:
@jugglinmike I just added tests for a new addition to the proposal: preservation of names of well-known intrinsic function objects (see https://github.com/tc39/Function-prototype-toString-revision/pull/10). Please let me know if it looks alright.

test/built-ins/Function/prototype/toString/intrinsics.js
```diff
@@ -0,0 +1,25 @@
...
+  Proxy, RangeError, ReferenceError, RegExp, Set, String, Symbol, SyntaxError, TypeError,
+  Uint8Array, Uint8ClampedArray, Uint16Array, Uint32Array, URIError, WeakMap, WeakSet,
+};
+
+const NATIVE_FUNCTION_RE = /\bfunction\b[\s\S]+\b\w+\b[\s\S]*\([\s\S]*\)[\s\S]*\{[\s\S]*\[[\s\S]*\bnative\b[\s\S]+\bcode\b[\s\S]*\][\s\S]*\}/;
+
+for (let intrinsicName in intrinsics) {
+  let intrinsic = intrinsics[intrinsicName];
+  let str = Function.prototype.toString.call(intrinsic);
+  assertSameValue(typeof str, "string");
```
jugglinmike: `assertsameValue` -> `assert.sameValue`

test/built-ins/Function/prototype/toString/intrinsics.js
```diff
@@ -0,0 +1,25 @@
...
+
+let intrinsics = {
+  Array, ArrayBuffer, Boolean, DataView, Date, decodeURI, decodeURIComponent, encodeURI,
+  encodeURIComponent, Error, eval, EvalError, Float32Array, Float64Array, Function, Int8Array,
+  Int16Array, Int32Array, isFinite, isNaN, Map, Number, Object, parseFloat, parseInt, Promise,
+  Proxy, RangeError, ReferenceError, RegExp, Set, String, Symbol, SyntaxError, TypeError,
+  Uint8Array, Uint8ClampedArray, Uint16Array, Uint32Array, URIError, WeakMap, WeakSet,
+};
+
+const NATIVE_FUNCTION_RE = /\bfunction\b[\s\S]+\b\w+\b[\s\S]*\([\s\S]*\)[\s\S]*\{[\s\S]*\[[\s\S]*\bnative\b[\s\S]+\bcode\b[\s\S]*\][\s\S]*\}/;
```
jugglinmike: This regular looks too permissive for the syntax defined by your proposal:

> NativeFunction :
>     function BindingIdentifieropt ( FormalParameters ) { [ native code ] }

For funsies, I wrote a function that satisfies the RegExp but violates the
spec:

    function eval ( x = 'function() { [ native code ] }' ) { 'lol'; }

It seems like:

- most of the `\S` classes may be removed (the one in the FormalParameters
  position seems necessary)
- the pattern should be anchored with `^\s*` and `\s*$`
- the `\b` classes may be removed

jugglinmike:
We should plan on extending the helper function in `testBuiltInObject.js` with
an assertion for this. I don't think it's appropriate to do this now, though.
I expect that at this stage, doing so would just cause consumers to disable the
tests, effectively resulting in *worse* coverage and potentially *greater*
margin for error.

---

I was tempted to suggest breaking this into separate files for each intrinsic,
packing all the assertions into a helper function and re-using code that way.
My thinking was that this organization would be far more discoverable for
future test writers. For example, if WeakRef lands, the authors will be more
likely to implement a test for this behavior if an equivalent exists in the
directory of all the other well-known functions. It seems unlikely that they
will know to update this list here.

But now I'm thinking that the specification isn't actually organized in that
way, and we should prefer the structure of that document when possible.

That said, have you given any thought to extending [section
9.3](http://www.ecma-international.org/ecma-262/6.0/#sec-built-in-function-objects)
to define the value of the [[SourceText]] slot for all built-in function
objects? I may be biased by these test organization concerns, but that almost
seems like a more natural place for some of this spec text.

---

It's been fun to watch this specification (and the tests) develop!

test/built-ins/Function/prototype/toString/intrinsics.js
michaelficarra: That's fine, I expect invalid strings to pass. The point is that a valid string cannot fail the regex. Remember, comments are allowed in any whitespace position. Without employing a full-blown tokeniser here, I think this is the best we can do.

jugglinmike: Ohh I forgot about comments! I agree that this is probably the best we can do.

michaelficarra:
> But now I'm thinking that the specification isn't actually organized in that way, and we should prefer the structure of that document when possible.

Agreed, and that is why I ended up organising it this way. Plus, it's easier and more discoverable.

> That said, have you given any thought to extending section 9.3 to define the value of the [[SourceText]] slot for all built-in function objects?

I'm not interested in doing that, as it would force browsers to return an exact string, and I'm not sure they all currently return the exact same string. That's a discussion that, if you would like to have it, should be had in an issue on the proposal itself.

jugglinmike:
> That's a discussion that, if you would like to have it, should be had in an issue on the proposal itself.

Definitely. I guess I started thinking out loud--I didn't mean to distract!

michaelficarra:
@jugglinmike I just added tests for bound functions (whoops!) and functions nested in expression position within other functions (the only ones I'm aware of are class heritage and computed property names). Since this proposal reached stage 3 yesterday, I think we're ready to merge.

jugglinmike:
Looks good, Michael! Regarding shared code: Test262 does have a mechanism for
doing this. If you define a file like `harness/nativeFunctionMatcher.js`, then
that file will be inserted at the beginning of any test that specifies the following meta
data:

```js
/*---
includes: [nativeFunctionMatcher.js]
---*/
```

Beyond avoiding duplication (a nobel cause if there ever was one), this gives
you a chance to "prove" the correctness of your regular expression with tests.
See the `test/harness/` directory for the tests for all the other helper
functions. Try to limit the language semantics/library functions you use there;
we want to keep passing those tests as a meaningful "baseline".

michaelficarra:
@jugglinmike I've extracted the regex to `harness/nativeFunctionMatcher.js`, included it in the F.p.toString tests, and added tests for it. Let me know if I need to do anything else before merging.

michaelficarra:
This should be ready to merge.

jugglinmike:
Squashed and merged to `master` at commit 6b7cbb50350e478a543da40d02492aa8f45a792d. Thanks for your patience, Michael!

PR #554, opened 2016-03-21
Remove surface tests for %TypedArray%.prototype.toString

leobalter:
Ref gh-526

cc @jugglinmike


jugglinmike:
Looks like I may have spoken too soon. Test262 currently defines a test for the `name` property of `Array.prototype.toString`, but *not* for the `length` property. In order to safely remove both of these TypedArray tests, we'll have to re-implement the latter as a test against `Array.prototype.toString`.

When I merge this patch, I plan on adding an explanation to the commit message description: "Because Test262 asserts the strict equality of `Array.prototype.toString` and `TypedArrayPrototype.toString`, tests for properties of the object do not need to be reproduced in both location." Sound good?

leobalter:
I updated the branch with new tests for Array.prototype.toString.length.

> When I merge this patch, I plan on adding an explanation to the commit message description: "Because Test262 asserts the strict equality of Array.prototype.toString and TypedArrayPrototype.toString, tests for properties of the object do not need to be reproduced in both location." Sound good?

It sounds great. Just a minor typo: s/`TypedArrayPrototype`/`TypedArray.Prototype`

jugglinmike:
Merged to `master` at commit c15079e7415463f4b799bb9fda964a757dca8e83. Thanks, Leo!

PR #555, opened 2016-03-22
Correct example in contribution guidelines

jugglinmike:
As written, the example for asserting runtime errors is written with an
early error. Because the error is expected to be reported prior to
program execution, the `assert.throws` function cannot be used to detect
it.

Demonstrate the usage of the helper function with a runtime error.


leobalter:
LGTM

PR #556, opened 2016-03-23
Add tests for typedarrays foreach and map

leobalter:


test/built-ins/TypedArray/prototype/forEach/callbackfn-detachbuffer.js
```diff
@@ -0,0 +1,39 @@
...
+---*/
+
+testWithTypedArrayConstructors(function(TA) {
+  var loops = 0;
+  var sample = new TA(2);
+
+  assert.throws(TypeError, function() {
+    sample.forEach(function() {
+      loops++;
+      $DETACHBUFFER(sample.buffer);
```
jugglinmike: This test will pass if `$DETACHBUFFER` throws a TypeError. Moving the increment
expression below the invocation of `$DETACHBUFFER` will protect against this,
but it then the test would pass if the callback was (incorrectly) invoked
twice, and the *second* invocation of `$DETACHBUFFER` threw a TypeError (which
we've already seen with some implementations of that function).

I think this means you need to separate counter variables.

test/built-ins/TypedArray/prototype/forEach/callbackfn-this.js
```diff
@@ -0,0 +1,56 @@
...
+  assert.sameValue(results1[1], expected, "without thisArg - [1]");
+  assert.sameValue(results1[2], expected, "without thisArg - [2]");
+
+  var results2 = [];
+
+  sample.forEach(function() {
+    results2.push(this);
+  }, thisArg);
+
+  assert.sameValue(results2.length, 3, "resuts2");
```
jugglinmike: Typo: "resuts2" -> "results2"

test/built-ins/TypedArray/prototype/map/callbackfn-this.js
```diff
@@ -0,0 +1,49 @@
...
+  assert.sameValue(results1[2], expected, "without thisArg - [2]");
+
+  var results2 = [];
+
+  sample.map(function() {
+    results2.push(this);
+    return 0;
+  }, thisArg);
+
+  assert.sameValue(results2.length, 3, "resuts2");
```
jugglinmike: Typo: "resuts2" -> "results2"

test/built-ins/TypedArray/prototype/map/arraylength-internal.js
```diff
@@ -0,0 +1,44 @@
...
+
+testWithTypedArrayConstructors(function(TA) {
+  var sample1 = new TA(42);
+  var loop = 0;
+
+  Object.defineProperty(sample1, "length", {value: 1});
+
+  sample1.map(function(v) {
+    loop++;
+    return v;
```
jugglinmike: No need for a formal parameter or a return statement (here and below)

test/built-ins/TypedArray/prototype/map/callbackfn-arguments-with-thisarg.js
```diff
@@ -0,0 +1,46 @@
...
+var thisArg = {};
+
+testWithTypedArrayConstructors(function(TA) {
+  var sample = new TA([42, 43, 44]);
+
+  var results = [];
+
+  sample.map(function(v) {
+    results.push(arguments);
+    return v;
```
jugglinmike: This return statement is also superfluous

test/built-ins/TypedArray/prototype/map/callbackfn-detachbuffer.js
```diff
@@ -0,0 +1,33 @@
...
+  assert.throws(TypeError, function() {
+    sample.map(function() {
+      loops++;
+      $DETACHBUFFER(sample.buffer);
+      return 1;
+    });
+  });
+
+  assert.sameValue(loops, 1, "callbackfn called only once");
+  assert.sameValue(loop[0], 0, "returned value is not set");
```
jugglinmike: `loop` is not defined in this file, so this test always fails.

test/built-ins/TypedArray/prototype/map/callbackfn-not-called-on-empty.js
```diff
@@ -0,0 +1,27 @@
...
+    c. Let mappedValue be ? Call(callbackfn, T, « kValue, k, O »).
+  ...
+includes: [testTypedArray.js]
+---*/
+
+testWithTypedArrayConstructors(function(TA) {
+  var called = 0;
+
+  new TA().map(function() {
+    return calls++;
```
jugglinmike: No need to `return` here

test/built-ins/TypedArray/prototype/map/return-new-typedarray-from-empty-length.js
```diff
@@ -0,0 +1,33 @@
...
+
+testWithTypedArrayConstructors(function(TA) {
+  var sample = new TA(0);
+
+  var result = sample.map(function(v) {
+    return v;
+  });
+
+  assert.notSameValue(result, sample, "new typedArray object");
+  assert.sameValue(result.constructor, sample.constructor, "same constructor");
```
jugglinmike: Using `TA` in place of `sample.constructor` would be more direct. Also an `instanceof` check would be more rigorous. Maybe even `Object.getPrototypeOf`.

jugglinmike:
These tests should also assert the value of the array(s) created to store
per-iteration values:

- forEach/callbackfn-arguments-with-thisarg.js
- forEach/callbackfn-arguments-without-thisarg.js
- map/callbackfn-arguments-with-thisarg.js
- map/callbackfn-arguments-without-thisarg.js

---

This code:

```js
var expected = (function() { return this; })();
var thisArg = {};
```

May be removed from the following files:

- forEach/callbackfn-arguments-with-thisarg.js
- forEach/callbackfn-arguments-without-thisarg.js
- map/callbackfn-arguments-with-thisarg.js
- map/callbackfn-arguments-without-thisarg.js

---

A few ideas for additional coverage:

- forEach & map: Modifying integer property values during iteration
- forEach & map: Not visiting non-integer property values (e.g. not just
  `length`, but something custom)
- map: value conversion as per my comment on [the integer indexed
  exotic](https://github.com/tc39/test262/pull/536#issuecomment-200455389). For
  that patch, we decided to move forward without those tests "for now" in order
  to help you stay on top of all the review feedback. Since it's still early in
  the process for this patch, I think we should tackle those concerns for this
  method from within this patch

test/built-ins/TypedArray/prototype/forEach/callbackfn-detachbuffer.js
leobalter: fixed at ~~4d8907e~~, 4c80af6

leobalter:
Feedback from inline comments applied at 4de77c5 and 7eb5f4c

leobalter:
> These tests should also assert the value of the array(s) created to store per-iteration values:

Isn't that covered on the `{map,forEach}/callback-this`?

leobalter:
> This code (...) May be removed from the following files: ...

fixed at aedfde0 and e475ea9

I'll work on the other ideas later but they are all good.

jugglinmike:
> > These tests should also assert the value of the array(s) created to store
> > per-iteration values:
> 
> Isn't that covered on the `{map,forEach}/callback-this`?

Apologies, I mispoke. I meant to request that those tests also assert the
*length* of those arrays (I didn't mention the `callbackfn-this` tests because
they already have such assertions).

leobalter:
> I meant to request that those tests also assert the length of those arrays

fixed at 4dfda13 and 7dd4199

leobalter:
I believe I finally fixed everything. I can squash the fixups if you prefer. For now I'm avoiding forced updates.

leobalter:
also, #571 covers the value conversion cases

test/built-ins/TypedArray/prototype/map/arraylength-internal.js
```diff
@@ -0,0 +1,42 @@
...
+includes: [testTypedArray.js]
+---*/
+
+testWithTypedArrayConstructors(function(TA) {
+  var sample1 = new TA(42);
+  var loop = 0;
+
+  Object.defineProperty(sample1, "length", {value: 1});
+
+  sample1.map(function(v) {
```
jugglinmike: The formal parameter can be removed (here and below)

jugglinmike:
One last comment, then I think we can merge this!

leobalter:
done, thanks!

test/built-ins/TypedArray/prototype/map/arraylength-internal.js
```diff
@@ -0,0 +1,42 @@
...
+  var sample2 = new TA(4);
+  Object.defineProperty(sample2, "length", {
+    get: function() {
+      throw new Test262Error(
+        "Does not return abrupt getting length property"
+      );
+    }
+  });
+
+  sample2.map(function(v) {
```
jugglinmike: This parameter is the "and below" I was referring to in my previous comment

leobalter: oops, I'll do a search

leobalter: fixed and found another one.

leobalter:
rebased and squashed commits can be found at the `typedarray-foreach-rebase` branch.

jugglinmike:
Squashed and merged to `master` at commits:

- f3cc1fb983a44a0c524dd846698f19b276918057
- e01371e12a1b1cd974f17bd5a35e22799720089a

leobalter:
thanks!
​

PR #560, opened 2016-03-29
Module code: syntax validation

jugglinmike:
Assert that module code is parsed as specified.


goyakin:

PR #561, opened 2016-03-29
Module code: early errors

jugglinmike:
Assert that relevant early errors are reported following the parsing of
module code.


goyakin:

PR #563, opened 2016-03-29
Module semantics: evaluation

jugglinmike:
Files whose name ends in `_.js` are not themselves valid Test262 tests
and should not be interpreted as such by test runners.

---

The tests for the GetBindingValue method of the module Environment
Record are very minimal. This is because GetBindingValue is necessary to
assert any aspect of binding creation/initialization/mutation. In this
way, GetBindingValue is being implicitly tested by every test that
references a binding value.


goyakin:

PR #564, opened 2016-03-29
Add tests for `String#padStart` and `String#padEnd`

ljharb:
(https://github.com/tc39/proposal-string-pad-start-end/issues/23)

### DIFF
--- a/test/built-ins/String/prototype/padEnd/exception-not-object-coercible.js
+++ b/test/built-ins/String/prototype/padEnd/exception-not-object-coercible.js
@@ -3,7 +3,9 @@
 
 /*---
 esid: sec-string.prototype.padend
-description: String#padEnd should fail if given a null or undefined value
+description: >
+  String#padEnd should fail if given a null or undefined value,
+  or an object not coercible to a string.
 author: Jordan Harband
 ---*/
 
@@ -13,4 +15,17 @@ assert.throws(TypeError, function () {
 
 assert.throws(TypeError, function () {
     String.prototype.padEnd.call(undefined);
+});
+
+var notCoercible = {
+    toString: function () {
+        throw new Test262Error('attempted toString');
+    },
+    valueOf: function () {
+        throw new Test262Error('attempted valueOf');
+    }
+};
+
+assert.throws(Test262Error, function () {
+    String.prototype.padEnd.call(notCoercible);
 });
--- a/test/built-ins/String/prototype/padEnd/exception-not-object-coercible.js
+++ b/test/built-ins/String/prototype/padEnd/exception-not-object-coercible.js
@@ -1,0 +1,16 @@
+// Copyright (C) 2016 Jordan Harband. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-string.prototype.padend
+description: String#padEnd should fail if given a null or undefined value
+author: Jordan Harband
+---*/
+
+assert.throws(TypeError, function () {
+    String.prototype.padEnd.call(null);
+});
+
+assert.throws(TypeError, function () {
+    String.prototype.padEnd.call(undefined);
+});
--- a/test/built-ins/String/prototype/padEnd/exception-not-object-coercible.js
+++ b/test/built-ins/String/prototype/padEnd/exception-not-object-coercible.js
@@ -20,9 +20,6 @@ assert.throws(TypeError, function () {
 var notCoercible = {
     toString: function () {
         throw new Test262Error('attempted toString');
-    },
-    valueOf: function () {
-        throw new Test262Error('attempted valueOf');
     }
 };
 
--- a/test/built-ins/String/prototype/padEnd/observable-operations.js
+++ b/test/built-ins/String/prototype/padEnd/observable-operations.js
@@ -26,17 +26,17 @@ var receiver = createPrimitiveObserver('receiver', {}, 'abc');
 
 var fillString = createPrimitiveObserver('fillString', {}, 'def');
 
-var maxLength = createPrimitiveObserver('maxLength', {}, 11);
+var maxLength = createPrimitiveObserver('maxLength', 11, {});
 
-var result = String.prototype.padEnd.call(receiver, fillString, maxLength);
+var result = String.prototype.padEnd.call(receiver, maxLength, fillString);
 
 assert.sameValue(result, 'abcdefdefde');
 
 assert.sameValue(log, '|' + [
     'toString:receiver',
     'valueOf:receiver',
-    'toString:maxLength',
     'valueOf:maxLength',
+    'toString:maxLength',
     'toString:fillString',
     'valueOf:fillString'
 ].join('|'), log);
--- a/test/built-ins/String/prototype/padEnd/observable-operations.js
+++ b/test/built-ins/String/prototype/padEnd/observable-operations.js
@@ -7,6 +7,8 @@ description: String#padEnd should perform observable operations in the correct o
 author: Jordan Harband
 ---*/
 
+var log = "";
+
 function createPrimitiveObserver(name, string, value) {
   return {
     toString: function () {
@@ -20,10 +22,9 @@ function createPrimitiveObserver(name, string, value) {
   };
 };
 
-var log = "";
-var receiver = createPrimitiveObserver('receiver', 'abc');
+var receiver = createPrimitiveObserver('receiver', {}, 'abc');
 
-var fillString = createPrimitiveObserver('fillString', 'def');
+var fillString = createPrimitiveObserver('fillString', {}, 'def');
 
 var maxLength = createPrimitiveObserver('maxLength', {}, 11);
 
@@ -31,4 +32,11 @@ var result = String.prototype.padEnd.call(receiver, fillString, maxLength);
 
 assert.sameValue(result, 'abcdefdefde');
 
-assert.sameValue(log, "|toString:receiver|valueOf:maxLength|toString:fillString", log);
+assert.sameValue(log, '|' + [
+    'toString:receiver',
+    'valueOf:receiver',
+    'toString:maxLength',
+    'valueOf:maxLength',
+    'toString:fillString',
+    'valueOf:fillString'
+].join('|'), log);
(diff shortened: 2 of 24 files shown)


test/built-ins/String/prototype/padEnd/exception-not-object-coercible.js
```diff
@@ -0,0 +1,16 @@
...
+
+/*---
+esid: sec-string.prototype.padend
+description: String#padEnd should fail if given a null or undefined value
+author: Jordan Harband
+---*/
+
+assert.throws(TypeError, function () {
+    String.prototype.padEnd.call(null);
+});
```
leobalter: you may store `String.prototype.padEnd` on a variable outside of this function so we avoid false positives when `padEnd` is still `undefined`.

jugglinmike:
Looking good, @ljharb! The tests look valid, I just have a few ideas for
extending coverage (and one request for formatting).

The `?` shorthand is really convenient, but it tends to hide branches in the
algorithm. As written, these tests only cover one of the case where an internal
operation produces an abrupt completion. This *could* be done by asserting a
TypeError when a Symbol is specified for the `this` value, and the second
argument (separate tests, of course). Call me old fashioned, but I still prefer
an ordinary object whose `toString`/`valueOf` method throws a
Test262Error--expressing the intention in that way will make the tests portable
to shimmed ES5 environments.

I like the "observable operations" tests! If you want to get really fancy (and
who doesn't want to get fancy?), you could configure these to return ordinary
objects for the "preferred" operation--that would let you observe the "hint"
behavior of ToPrimitive.

The proposal says "If fillString is `undefined`, [...]", and in cases like
this, I like to explicitly test other falsey values to make sure that there is
no ToBoolean funny business going on.

The tests that have multi-line descriptions should use `>` rather than `|` as
a block delimiter because newlines are not significant. (This is a distinction
I personally discovered only recently, so many of my old commits are incorrect
in this regard).

ljharb:
@jugglinmike I'll add separate tests that throw on the ToString and ToLength branches (which the existing observable ops tests ensure are properly called). I'll also add more non-string + falsy fillString values to cover what you've indicated, and fix the multiline descriptions.

How could I observe the "hint" behavior of `toPrimitive`? Or do you mean, just the nested toString/valueOf call? That seems like it could recurse forever - at which point would I stop?

jugglinmike:
> How could I observe the "hint" behavior of toPrimitive? Or do you mean, just
> the nested toString/valueOf call? That seems like it could recurse forever -
> at which point would I stop?

Sorry, that was a bit unclear. Here's the text I had in mind:

>     [...]
>     3. If hint is "string", then
>        a. Let methodNames be « "toString", "valueOf" ».
>     4. Else,
>        a. Let methodNames be « "valueOf", "toString" ».
>     5. For each name in methodNames in List order, do
>     [...]

By defining the `toString` methods to return a non-primitive value, you'll be
able to observe that `valueOf` is called following `toString` in such cases.

```diff
-var receiver = createPrimitiveObserver('receiver', 'abc');
+var receiver = createPrimitiveObserver('receiver', {}, 'abc');

-var fillString = createPrimitiveObserver('fillString', 'def');
+var fillString = createPrimitiveObserver('fillString', {}, 'def');
```

ljharb:
ah, i see. ok cool, i'll make the changes today

ljharb:
Updated!

test/built-ins/String/prototype/padEnd/observable-operations.js
```diff
@@ -0,0 +1,51 @@
...
+    createPrimitiveObserver('toPrimitiveMaxLengthValueOf', {}, 11)
+);
+
+var result = String.prototype.padEnd.call(receiver, fillString, maxLength);
+
+assert.sameValue(result, 'defdefdeabc');
+
+assert.sameValue(log, '|' + [
+    'toString:receiver',
+    'toString:toPrimitiveABC',
```
leobalter: I believe `valueOf:receive` happens here as the result of `toString:receive` is an Object and the the valueOf will return undefined.

test/built-ins/String/prototype/padEnd/observable-operations.js
```diff
@@ -0,0 +1,51 @@
...
+
+var maxLength = createPrimitiveObserver(
+    'maxLength',
+    createPrimitiveObserver('toPrimitiveMaxLengthToString', {}, {}),
+    createPrimitiveObserver('toPrimitiveMaxLengthValueOf', {}, 11)
+);
+
+var result = String.prototype.padEnd.call(receiver, fillString, maxLength);
+
+assert.sameValue(result, 'defdefdeabc');
```
leobalter: because of `valueOf` returning undefined on `receiver` and `fillString`, this assertion will not pass.

leobalter: result is undefined

ljharb: hmm… the chain is such that `receiver` should ToPrimitive to `'abc'`, and `fillString` to `'def'`

leobalter: on https://tc39.github.io/ecma262/#sec-toprimitive

```
...
7. Return ? OrdinaryToPrimitive(input, hint). 
```

```
3. If hint is "string", then
  a. Let methodNames be « "toString", "valueOf" ».
...
5. For each name in methodNames in List order, do
  a. Let method be ? Get(O, name).
  b. If IsCallable(method) is true, then
    i. Let result be ? Call(method, O).
    ii. If Type(result) is not Object, return result.
...
```

So, step 5.b.ii will check the result for `receiver.toString` is an Object (`createPrimitiveObserver('toPrimitiveABC', {}, 'abc')`), and will continue the loop getting result from `receiver.valueOf`.

ljharb: understood and fixed, thanks!

test/built-ins/String/prototype/padEnd/observable-operations.js
ljharb: understood and fixed, thanks!

test/built-ins/String/prototype/padEnd/observable-operations.js
```diff
@@ -0,0 +1,42 @@
...
+  };
+};
+
+var receiver = createPrimitiveObserver('receiver', {}, 'abc');
+
+var fillString = createPrimitiveObserver('fillString', {}, 'def');
+
+var maxLength = createPrimitiveObserver('maxLength', {}, 11);
+
+var result = String.prototype.padEnd.call(receiver, fillString, maxLength);
```
leobalter:
```diff
-var result = String.prototype.padStart.call(receiver, fillString, maxLength);
+var result = String.prototype.padStart.call(receiver, maxLength, fillString);
```

leobalter: ops (this is padEnd, not padStart, but the same argument order applies)

test/built-ins/String/prototype/padStart/observable-operations.js
```diff
@@ -0,0 +1,42 @@
...
+  };
+};
+
+var receiver = createPrimitiveObserver('receiver', {}, 'abc');
+
+var fillString = createPrimitiveObserver('fillString', {}, 'def');
+
+var maxLength = createPrimitiveObserver('maxLength', {}, 11);
+
+var result = String.prototype.padStart.call(receiver, fillString, maxLength);
```
leobalter:
```diff
-var result = String.prototype.padStart.call(receiver, fillString, maxLength);
+var result = String.prototype.padStart.call(receiver, maxLength, fillString);
```

test/built-ins/String/prototype/padEnd/observable-operations.js
```diff
@@ -0,0 +1,42 @@
...
+assert.sameValue(result, 'abcdefdefde');
+
+assert.sameValue(log, '|' + [
+    'toString:receiver',
+    'valueOf:receiver',
+    'toString:maxLength',
+    'valueOf:maxLength',
+    'toString:fillString',
+    'valueOf:fillString'
+].join('|'), log);
```
leobalter:
```diff
 assert.sameValue(log, '|' + [
     'toString:receiver',
     'valueOf:receiver',
-    'toString:maxLength',
     'valueOf:maxLength',
+    'toString:maxLength',
     'toString:fillString',
     'valueOf:fillString'
 ].join('|'), log);
```

test/built-ins/String/prototype/padEnd/observable-operations.js
```diff
@@ -0,0 +1,42 @@
...
+      return value;
+    }
+  };
+};
+
+var receiver = createPrimitiveObserver('receiver', {}, 'abc');
+
+var fillString = createPrimitiveObserver('fillString', {}, 'def');
+
+var maxLength = createPrimitiveObserver('maxLength', {}, 11);
```
leobalter:
```diff
-var maxLength = createPrimitiveObserver('maxLength', {}, 11);
+var maxLength = createPrimitiveObserver('maxLength', 11, {});
```

leobalter: the same applies to padStart/observable-operations.js

test/built-ins/String/prototype/padEnd/observable-operations.js
leobalter: the same applies to padStart/observable-operations.js

test/built-ins/String/prototype/padEnd/exception-not-object-coercible.js
```diff
@@ -0,0 +1,31 @@
...
+    String.prototype.padEnd.call(undefined);
+});
+
+var notCoercible = {
+    toString: function () {
+        throw new Test262Error('attempted toString');
+    },
+    valueOf: function () {
+        throw new Test262Error('attempted valueOf');
+    }
```
leobalter: valueOf is not reachable. You already verify this on the observable order tests.

ljharb: ok, should i remove it?

test/built-ins/String/prototype/padEnd/observable-operations.js
ljharb: aha, thanks

test/built-ins/String/prototype/padEnd/exception-not-object-coercible.js
ljharb: removed

leobalter:
LGTM

leobalter:
Squashed and merged at c95e673, thanks!

ljharb:
Thank you! :-D

PR #565, opened 2016-04-01
Add tests for module namespace objects

jugglinmike:
Assert correct behavior of the own properties of module namespace objects and the essential internal methods of module namespace exotic objects.

In the process of writing these tests, I identified two potential improvements to the specification:

- [gh-507: [editorial] Assert normal completion
  value](https://github.com/tc39/ecma262/pull/507)
- [gh-510: [editorial] Specify order of module namespace
  keys](https://github.com/tc39/ecma262/pull/510) 

Although those patches are still under review, I wrote this changeset under the assumption that they are correct. If I was mistaken, I will update the tests here as appropriate.


goyakin:

PR #571, opened 2016-04-11
Add tests for conversion operations on TypedArrays

leobalter:


leobalter:
extra info: passing on Chakra, V8, SM and ES6Draft

harness/testTypedArray.js
```diff
@@ -46,3 +46,365 @@ function testWithTypedArrayConstructors(f, selected) {
...
 }
+
+/**
+ * Helper for conversion operations on TypedArrays, the expected values
+ * properties are indexed in order to match the respective value for each
+ * TypedArray constructor
+ * @param  {Function} fn - the function to call for each constructor and value.
+ *                         will be called with the constructor, value and
+ *                         expected value.
+ * @return {[type]} - follows the return of testWithTypedArrayConstructors if
```
jugglinmike: `testWithTypedArrayConstructors` only returns `undefined`, so this documentation may be more distracting than helpful. It's probably safest to return nothing, so that future changes to `testWithTypedArrayConstructors` don't have unintended consequences here.

leobalter: 0e6e107

leobalter:
added dd5a23f WRT gh-576

leobalter:
I just added more numbers

jugglinmike:
I'm always reluctant to rely on intricate logic in tests, but I think it may be warranted here given the sheer number of relevant test cases.

One change I suggest concerns initial values. For instance, in `test/built-ins/TypedArray/prototype/map/return-new-typedarray-conversion-operation.js`, the test subject is initialized via the "length" constructor: `var sample = new TA(1);`. This sample will initially satisfy the assertion: `assert.sameValue(sample[0], 0)` (related: gh-579). This could potentially hide errors for implementations that ignore certain values.

In other words, there is no way to tell if the second statement in the following code actually has a side effect:

```js
var x = new Int8Array(1);
x[0] = 0.6;
```

Since we know both `0` and `1` are valid values for all typed arrays, maybe something like this will suffice:

```js
if (expected === 0) {
  sample[0] = 1;
}
```

This partially relies on the semantics under test, though, so it might also be prudent to assert that this initial value is set correctly--something intentionally vague would do here: `assert.notSameValue(sample[0], 0);`

leobalter:
I agree we need to protect our tests but I believe that setting the initial value to a specific number would help, like `var sample = new TA([42]);`. Knowing that 42 is a number we won't have on the values list or if we have it in a future, it's only a single case, it feels better than adding more logic.

I changed it according to your suggestion at  6c6d258, letting the my idea to be discussed or simply ignored if it's not good.

jugglinmike:
Another alternative would be to extend the helper functions to provide an initial value. Then we could avoid the extra logic without relying on a "magic value" like 42.

leobalter:
4142632

jugglinmike:
Don't forget to make note of the new parameter in the function's documentation!

leobalter:
77828a9

jugglinmike:
Wonderful :) This looks good to me

jugglinmike:
Merged to `master` at commit 04a3c28f7d3552df2d40de53a860e0feea0a4909

Issue #569, opened 2016-04-07
Encoding Tests' Targeted Spec Version


jugglinmike:
On gh-472 back in January, we discussed the potential difficulties with maintaining a test suite for a "living" specification--what, if anything, should Test262 do to account for the fact that specification revisions invalidate existing tests. At the time, consensus seemed to favor the simple-but-progressive solution: "delete the tests."

While I had some objections to that conversation's resolution, consensus seemed strong enough for me to let it go. [A recent inquiry](https://twitter.com/dshaw/status/717728678159855616) has prompted me to re-open the discussion as due dilligence to the project's current/potential consumers.

I have two use cases in mind:

First are implementors who are specifically targeting an older edition of the spec. While writing tests for ES2015, we sometimes find omissions in coverage for ES5 features. If TC-39's advice to these implementors is, ["use the tests at this tag,"](https://github.com/tc39/test262/pull/471#discussion_r50332324) it would be cutting them off from new tests written for older features (see, for example, gh-549 and gh-550). So this solution sends a clear signal that ECMA is not interested in the correctness of "legacy" runtimes.

Second, I can imagine that there are some projects which do not have the resources to track TC-39's new yearly release cycle. Or, even if they do, they may prefer to take a more conservative approach, waiting to commit to new features until formal ratification from ECMA. The practice of [deleting tests that are invalidated by the current draft](https://github.com/tc39/test262/issues/472#issuecomment-173006747) allows for inconsistency where it was previously forbidden. Even if you have no sympathy for so-called "legacy" runtimes, I think it is another matter to degrade the test suite's value for implementors honoring the latest publication.

To be honest, I don't have enough familiarity with the ecosystem of runtimes to say if either of those groups are large enough influence the decision here. But I suspect that they are the same people who would tend to be under-represented in TC-39 meetings. So although [Chakra, V8, and SpiderMonkey may be well-served by a "living test suite,"](https://github.com/tc39/test262/pull/471/files#r50334280) I don't want to take that as sufficient without considering these other (potential) interests. If you folks see potential in this line of reasoning, I would be willing to collect more conclusive data on implementors' preferences.

The label "living" is a bit misleading because it suggests any alternative would be slow, irrelevant, or otherwise "dead." I think we may be able to have our cake and eat it, too. My idea (as ever) revolves around more test metadata.

To keep `master` in a state that reflects the very latest draft but retains value for less progressive implementors, we would only need to author tests in a slightly different way:
[...]

PR #572, opened 2016-04-12
Improve test organization and coverage for `eval`

jugglinmike:
This work involved re-organizing (and in some cases, refactoring) existing
tests. I'm submitting re-naming and re-factoring operations as distinct commits
in order to more clearly demonstrate the provinance of each file in the new
structure.

This changeset relies on gh-570.


jugglinmike:
I've just pushed two new tests to assert that local bindings created via `eval` may be deleted. These tests only apply to non-strict direct eval. This patch already had coverage for the configurability of global bindings (accounting for non-strict indirect eval), and it's not possible to verify this behavior for strict eval in any case (due to restrictions on the `delete` operator in strict mode code).

test/language/eval-code/indirect/var-env-var-non-strict.js
```diff
@@ -0,0 +1,15 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-performeval
+es5id: 10.4.2-2-c-1
```
leobalter: are these files supposed to keep `es5id`?

jugglinmike: Yes. At the moment, we're still considering using that tag as a indicator that the test applies to ES5: https://github.com/tc39/test262/issues/569. As I mentioned there, I'd like to simplify that requirement, but in the mean time, I decided to just include the es5id.

leobalter:
rebased, squashed the fixup commits and pushed to master.

Issue #528, opened 2016-03-03
Revisit date test errors


littledan:
Has anyone followed up on https://bugs.ecmascript.org/show_bug.cgi?id=293 ? It looks like it's still based on California date values, and the fix suggested in that bug looks good to me. @jugglinmike @leobalter

littledan:
Ping. These tests continue to be incorrect.

jugglinmike:
I'll look into this presently

PR #573, opened 2016-04-12
Correct invalid tests for the Date constructor

jugglinmike:
The millisecond representation of a given Date instance is dependent on
the local system's time zone settings. In order to pass consistently
across contexts, tests for this value must take the system configuration
into account.

Introduce a test harness utility function to encapsulate these concerns.
Re-use this function across all test files that assert the exact
millisecond representation of Date instances.

@littledan As noted in the original BugZilla report, this approach complicates
the tests, but it doesn't quite depend on the semantics under test, so it
should be "safe."

That said, another solution occurred to me while writing this: we could make a
specific time zone (say, Greenwich Mean Time) a formal requirement to executing
Test262. This seems like the same class of problems I am trying to address in
gh-508. That is: limiting runtime variability through requirements
specification.

...But maybe it's a little too presumptuous to ask consumers to re-configure
their system clocks before running the tests. It certainly seems like overkill
for these 6 files, but doing so would let us forget about this detail entirely
(instead of carrying around another helper function and remembering to use it).

Resolves gh-528 (and https://bugs.ecmascript.org/show_bug.cgi?id=293 before it)

### DIFF
--- a/test/built-ins/Date/S15.9.3.1_A5_T2.js
+++ b/test/built-ins/Date/S15.9.3.1_A5_T2.js
@@ -1,7 +1,7 @@
 // Copyright 2009 the Sputnik authors.  All rights reserved.
 // This code is governed by the BSD license found in the LICENSE file.
-
 /*---
+esid: sec-date-constructor
 info: >
     The [[Value]] property of the newly constructed object
     is set by following steps:
--- a/test/built-ins/Date/S15.9.3.1_A5_T2.js
+++ b/test/built-ins/Date/S15.9.3.1_A5_T2.js
@@ -14,56 +14,29 @@ info: >
     TimeClip(UTC(Result(11)))
 es5id: 15.9.3.1_A5_T2
 description: 3 arguments, (year, month, date)
-includes:
-    - numeric_conversion.js
-    - Date_constants.js
-    - Date_library.js
+includes: [assertRelativeDateMs.js]
 ---*/
 
-if (-2209046400000 !== new Date(1899, 11, 31).valueOf()) {
-  $ERROR("#1: Incorrect value of Date");
-}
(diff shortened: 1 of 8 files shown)


test/harness/assertRelativeDateMs.js
```diff
@@ -0,0 +1,80 @@
...
+try {
+  assertRelativeDateMs(new Date(-1), 0);
+} catch (err) {
+  thrown = err;
+}
+if (!thrown) {
+  $ERROR('Expected error, but no error was thrown.');
+} else if (thrown.constructor !== Test262Error) {
+  $ERROR('Expected error of type Test262Error.');
+}
```
leobalter: why we don't use assert.throws here? I'm assuming you're avoiding the use of other harness, is that right?

jugglinmike: Yes

test/built-ins/Date/S15.9.3.1_A5_T2.js
```diff
@@ -14,56 +14,29 @@ info: >
     TimeClip(UTC(Result(11)))
 es5id: 15.9.3.1_A5_T2
 description: 3 arguments, (year, month, date)
-includes:
-    - numeric_conversion.js
-    - Date_constants.js
-    - Date_library.js
+includes: [assertRelativeDateMs.js]
```
leobalter: this file doesn't have an esid flag

leobalter:
LGTM, my comments above shouldn't block anything.

test/built-ins/Date/S15.9.3.1_A5_T2.js
jugglinmike: Good catch. Fixed.

leobalter:
@littledan I'm +1 to merge this. Would you confirm this fixes #528?

PR #577, opened 2016-04-13
Add tests for TryStatement binding restrictions

jugglinmike:
This changeset includes tests for early errors and those generated
dynamically by eval. It also accounts for relevant AnnexB extensions.

### DIFF
--- a/test/language/eval-code/direct/var-env-lower-lex-catch-non-strict.js
+++ b/test/language/eval-code/direct/var-env-lower-lex-catch-non-strict.js
@@ -1,0 +1,20 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-evaldeclarationinstantiation
+description: Variable collision with lexical binding in lower scope
+info: >
+    Annex B extensions permit re-declarations from FunctionDeclaration,
+    VariableStatement, the VariableDeclarationList of a for statement, and the
+    ForBinding of a for-in statement. Bindings from the ForBinding of a for-of
+    statement are restricted regardless of the application of Annex B.
+flags: [noStrict]
+---*/
+
+assert.throws(SyntaxError, function() {
+  try {
+    throw null;
+  } catch (err) {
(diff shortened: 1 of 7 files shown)


test/language/eval-code/direct/var-env-lower-lex-catch-non-strict.js
```diff
@@ -0,0 +1,20 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-evaldeclarationinstantiation
```
leobalter: missing `es6id: 18.2.1.2`?

jugglinmike: Couldn't hurt! Although given the use case, I'm hopeful that we may be able to simplify this information soon--see my comment here https://github.com/tc39/test262/issues/569#issuecomment-211431837

leobalter:
LGTM

test/language/statements/try/early-catch-duplicates.js
```diff
@@ -0,0 +1,14 @@
...
+/*---
+esid: sec-try-statement-static-semantics-early-errors
+es6id: 13.15.1
+description: >
+    It is a Syntax Error if BoundNames of CatchParameter contains any duplicate
+    elements.
+negative: SyntaxError
+---*/
+
+$ERROR('This code should not be executed.');
```
leobalter: Some [other tests](https://github.com/tc39/test262/pull/566/files) use the `throw new Test262Error()`. Both cases are correct, but we should define one way to do it for consistency.

jugglinmike: [Still others](https://github.com/tc39/test262/blob/c230b7f307ecc33974b79998638496cd27039bb3/test/language/directive-prologue/10.1.1-8gs.js) use `NotEarlyError`. I agree that we could use some consistency, but my long-term goal is to avoid the concern altogether. See the discussion at gh-542 for details on how encoding the error "stage" in the frontmatter would take this burden off of test writers.

PR #578, opened 2016-04-14
Add tests for %TypedArray%.prototype.find

leobalter:


test/built-ins/TypedArray/prototype/find/predicate-call-parameters.js
```diff
@@ -0,0 +1,62 @@
...
+  var results = [];
+  var result;
+
+  sample.foo = "bar"; // Ignores non integer index properties
+
+  sample.find(function() {
+    results.push(arguments);
+  });
+
+  assert.sameValue(results.length, 3, "predicated is called for each index");
```
jugglinmike: "predicated" -> "predicate"

test/built-ins/TypedArray/prototype/find/predicate-may-detach-buffer.js
```diff
@@ -0,0 +1,55 @@
...
+  var loops = 0;
+
+  assert.throws(TypeError, function() {
+    sample.find(function() {
+      loops++;
+      $DETACHBUFFER(sample.buffer);
+    });
+  }, "throws a TypeError getting a value from the detached buffer");
+
+  assert.sameValue(loops, 1, "predicated is called once");
```
jugglinmike: "predicated" -> "predicate"

test/built-ins/TypedArray/prototype/find/predicate-may-detach-buffer.js
```diff
@@ -0,0 +1,55 @@
...
+testWithTypedArrayConstructors(function(TA) {
+  var sample = new TA(2);
+  var loops = 0;
+
+  assert.throws(TypeError, function() {
+    sample.find(function() {
+      loops++;
+      $DETACHBUFFER(sample.buffer);
+    });
+  }, "throws a TypeError getting a value from the detached buffer");
```
jugglinmike: This test will pass if the implementation-defined `$DETACHBUFFER` method itself throws a TypeError. We should assert a normal completion from that call.

test/built-ins/TypedArray/prototype/find/return-undefined-if-predicate-returns-false-value.js
```diff
@@ -0,0 +1,58 @@
...
+  assert.sameValue(result, undefined, "ToBoolean(empty string)");
+
+  result = sample.find(function(val) { return undefined; });
+  assert.sameValue(result, undefined, "ToBoolean(undefined)");
+
+  result = sample.find(function(val) { return null; });
+  assert.sameValue(result, undefined, "ToBoolean(null)");
+
+  result = sample.find(function(val) { return 0; });
+  assert.sameValue(result, undefined, "ToBoolean(0)");
```
jugglinmike: How about another assertion for -0?

jugglinmike:
Looking good, Leo! Besides my in-line comments, I was thinking that another good test would be the behavior of `find` when elements are changed during its traversal. This could cover mutation of previously-visited elements, the current element, and unvisited elements.

test/built-ins/TypedArray/prototype/find/predicate-may-detach-buffer.js
leobalter: > implementation-defined $DETACHBUFFER method itself throws a TypeError

That's a very specific case and there's no apparent reason for doing it, unless it can't load the current method that throws a Test262Error.

leobalter:
fixed inline comments at a10ec25.

a86cb25 has the extra test for values changed on the during traversal

jugglinmike:
This isn't exactly what I had in mind for in-place modification. When it comes
to the `find` method, the "interesting" aspect of in-place modification is how
it effects the semantics of the current search operation. Modifying the current
element or a previously-visited element (even if the new value satisfies the
predicate) should have no effect on the result. On the other hand, changing an
element that has yet to be visited *can* effect the result. (You could imagine
an buggy implementation that cached all the element values before
iterating--this test should catch such non-standard behavior).

Does that make sense?

leobalter:
yes! sure! I fixed it here and at the findIndex tests

test/built-ins/TypedArray/prototype/find/predicate-may-detach-buffer.js
jugglinmike: I don't mean to say that throwing an error is logical. After all, most of the
tests we write are intended to ensure that implementors are not doing something
illogical. We have to be particularly careful with `$DETACHBUFFER` because it
"wraps" non-standard and under-specified functionality, so its behavior is
highly variable.

leobalter: > We should assert a normal completion from that call.

Now I got it. I'll do it.

leobalter: 6108abb (I repeated this on #581)

test/built-ins/TypedArray/prototype/find/predicate-call-changes-value.js
```diff
@@ -30,14 +30,49 @@ includes: [compareArray.js, testTypedArray.js]
...
+    return val === 7;
+  });
+  assert.sameValue(result, undefined, "value not found - changed after call");
+
+  sample = new TA(arr);
+  result = sample.find(function() {
+    sample[0] = 7;
+    return true;
+  });
+  assert.sameValue(result, 1, "find() returns previous found value");
```
jugglinmike: Nice one!h

jugglinmike:
Looks good to me!

jugglinmike:
Rebased, squashed, and merged to`master` at commit 9aa4dced8d8afbe114b16650187ebb14c3e21e01. Thanks, Leo!

PR #580, opened 2016-04-15
Correct tests for native functions' toString value

jugglinmike:
Modify the regular expression for native functions' toString value to
satisfy all test cases. Correct the test file's reference to the harness
file. Re-format the test file's assertions to aid debugging in the event
of failure.

@michaelficarra Would you mind taking a look?


harness/nativeFunctionMatcher.js
```diff
@@ -1 +1 @@
-const NATIVE_FUNCTION_RE = /\bfunction\b[\s\S]+\b\w+\b[\s\S]*\([\s\S]*\)[\s\S]*\{[\s\S]*\[[\s\S]*\bnative\b[\s\S]+\bcode\b[\s\S]*\][\s\S]*\}/;
+const NATIVE_FUNCTION_RE = /\bfunction\b(\s+[\s\S]+|\s*)\([\s\S]*\)[\s\S]*\{[\s\S]*\[[\s\S]*\bnative\b[\s\S]+\bcode\b[\s\S]*\][\s\S]*\}/;
```
michaelficarra: I think you can get away with just deleting `[\s\S]+\b\w+\b` from the original regexp here.

michaelficarra:
LGTM otherwise. 👍

harness/nativeFunctionMatcher.js
jugglinmike: You are correct! I've done exactly that in a follow-up commit

jugglinmike:
Thanks for the review, Michael. In addition to your suggested simplification, I've moved the documentation describing the regular expression to the harness file itself.

michaelficarra:
Even better! :+1:

PR #581, opened 2016-04-15
Add tests for %TypedArray%.prototype.findIndex

leobalter:


test/built-ins/TypedArray/prototype/findIndex/predicate-call-this-non-strict.js
```diff
@@ -0,0 +1,49 @@
...
+  sample.findIndex(function() {
+    result = this;
+  });
+
+  assert.sameValue(result, T, "without thisArg, predicate this is the global");
+
+  var o = {};
+  sample.findIndex(function() {
+    result = this;
+  }, o);
```
jugglinmike: I just thought of another assertion that might be good here (and in the tests for `find`): specifying `undefined` for `thisArg`. In non-strict mode, the "this" value should *still* be the global object. Do you think that's worth testing?

test/built-ins/TypedArray/prototype/findIndex/predicate-is-not-callable-throws.js
```diff
@@ -0,0 +1,64 @@
...
+    sample.findIndex(null);
+  }, "null");
+
+  assert.throws(TypeError, function() {
+    sample.findIndex(undefined);
+  }, "undefined");
+
+  assert.throws(TypeError, function() {
+    sample.findIndex(false);
+  }, "true");
```
jugglinmike: "true" -> "false"

test/built-ins/TypedArray/prototype/findIndex/return-negative-one-if-predicate-returns-false-value.js
```diff
@@ -0,0 +1,55 @@
...
+  assert.sameValue(result, -1, "ToBoolean(string)");
+
+  result = sample.findIndex(function(val) { return undefined; });
+  assert.sameValue(result, -1, "ToBoolean(undefined)");
+
+  result = sample.findIndex(function(val) { return null; });
+  assert.sameValue(result, -1, "ToBoolean(null)");
+
+  result = sample.findIndex(function(val) { return 0; });
+  assert.sameValue(result, -1, "ToBoolean(0)");
```
jugglinmike: Mind adding an assertion for -0 here, too?

jugglinmike:
Just a few minor requests for this one, Leo, then it should be good to go.

test/built-ins/TypedArray/prototype/findIndex/predicate-call-this-non-strict.js
leobalter: +1. I could tell the non existent argument is handled as undefined, but it's precise adding a specific test for it.

test/built-ins/TypedArray/prototype/findIndex/predicate-is-not-callable-throws.js
leobalter: oops. at least the assertion is valid.

jugglinmike: Yup!

leobalter:
Thanks! fixes at 496f0ac

test/built-ins/TypedArray/prototype/findIndex/predicate-call-this-non-strict.js
```diff
@@ -39,14 +39,16 @@ testWithTypedArrayConstructors(function(TA) {
...
+
+  result = null;  
   sample.findIndex(function() {
     result = this;
   }, undefined);
 
   assert.sameValue(result, T, "predicate this is the global when thisArg is undefined");
 
   var o = {};
+  result = null;
```
jugglinmike: Good catch

jugglinmike:
Looks good to me!

jugglinmike:
Rebased, squashed, and merged to `master` at commit bc0a40c51ce9dad6a0fe4f5f8967559d7b707d8b. Thanks, Leo!

PR #583, opened 2016-04-18
[generation] Support changing to existing files

jugglinmike:
When inspecting previously-generated files, a new `Test` instance should
be used. This avoids over-writing the in-memory representation of the
latest test, and allows previously-existing test files to be partially
updated according to subsequent changes in their respective source/case
files.

(I discovered this oversight while updating the yet-to-be-submitted first batch of procedurally-generated tests.)


leobalter:
LGTM

PR #584, opened 2016-04-18
Add tests for destructuring binding (procedurally generated)

jugglinmike:
Although the bulk of the work here is in new test cases, this change set
includes two updates to previously-accepted files:

- Adding the `esid` frontmatter tag (which was designed in the time since the
  files were first written)
- Adding omitted destructuring binding forms (which I simply did not originally
  identify)

To help facilitate review, I've organized all procedurally-generated changes
into dedicated commits.


src/dstr-binding/ary-init-iter-close.case
```diff
@@ -0,0 +1,37 @@
...
+      return {};
+    }
+  };
+};
+//- elems
+[x]
+//- vals
+iter
+//- body
+assert.sameValue(doneCallCount, 1);
```
leobalter: can we use any message on these?

On the arrow-function example there's another assertion with the same expected value (`assert.sameValue(callCount, 1);`) and a message would be very helpful to identify the assertion if it fails.

test/language/expressions/arrow-function/dstr-ary-init-iter-close.js
```diff
@@ -0,0 +1,73 @@
...
+
+var callCount = 0;
+var f;
+f = ([x]) => {
+  assert.sameValue(doneCallCount, 1);
+  callCount = callCount + 1;
+};
+
+f(iter);
+assert.sameValue(callCount, 1);
```
leobalter: here's the generated file I mentioned on my previous comment.

test/language/expressions/arrow-function/dstr-ary-init-iter-close.js
```diff
@@ -0,0 +1,73 @@
...
+    13.3.3.5 Runtime Semantics: BindingInitialization
+    
+    BindingPattern : ArrayBindingPattern
+    
+    [...]
+    4. If iteratorRecord.[[done]] is false, return ? IteratorClose(iterator,
+       result).
+    [...]
+    
+---*/
```
leobalter: the generation tool is creating some noise: 

- trailing whitespace on empty lines (from the template and case files)
- adding an extra blank line (8 to 10 and 48 to 50)
- `info: |` converted to `info: >`

leobalter: these are issues that can be addressed in a follow-up patch.

src/dstr-binding/ary-init-iter-close.case
```diff
@@ -0,0 +1,37 @@
...
+info: |
+    13.3.3.5 Runtime Semantics: BindingInitialization
+
+    BindingPattern : ArrayBindingPattern
+
+    [...]
+    4. If iteratorRecord.[[done]] is false, return ? IteratorClose(iterator,
+       result).
+    [...]
+features: [Symbol.iterator]
```
leobalter: shouldn't destructuring binding be listed too?

src/dstr-binding/ary-init-iter-close.case
jugglinmike: That's a good idea!

src/dstr-binding/ary-init-iter-close.case
jugglinmike: Sounds good to me

test/language/expressions/arrow-function/dstr-ary-init-iter-close.js
jugglinmike: I appreciate your patience on this, but I'd personally like to correct those problems ASAP--even before landing this work. I've submitted a patch (see https://github.com/tc39/test262/pull/601) to do just that.

jugglinmike:
Now that gh-601 has landed (thanks @leobalter and @goyakin), the generated
files in this changeset should be updated. This will require rebasing this
branch, which I am happy to do upon request. Of course, my reviewer may do this
as well--just run `make.py` (I recommend squashing all the changes to generated
files into a single commit).

leobalter:
@jugglinmike would you help with the rebase, please?

jugglinmike:
Sure; I have rebased this branch as I suggested above

leobalter:
I merged it intentionally with a merge commit as I didn't want to squash the generated tests.

Great work, Mike!

PR #587, opened 2016-04-18
Add tests for Annex B "function in block" semantics (procedurally generated)

jugglinmike:
In addition to increasing Test262's coverage of Annex B "function in block"
semantics, this change set includes two extensions to the test generation tool.
These relate to requirements that were only apparent after beginning work on
these tests.


leobalter:
@goyakin would you help on this review, please?

leobalter:
LGTM, the changes at 613d33a (case directories) are nice!

PR #589, opened 2016-04-19
Add tests for %TypedArray%.prototype.subarray

leobalter:


test/built-ins/TypedArray/prototype/subarray/detached-buffer.js
```diff
@@ -0,0 +1,64 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-%typedarray%.prototype.subarray
+description: Subarray may return a new empty instance
```
jugglinmike: This description doesn't seem quite right

test/built-ins/TypedArray/prototype/subarray/detached-buffer.js
```diff
@@ -0,0 +1,64 @@
...
+}
+
+testWithTypedArrayConstructors(function(TA) {
+  var sample = new TA(2);
+  begin = false;
+  end = false;
+
+  $DETACHBUFFER(sample.buffer);
+  assert.throws(TypeError, function() {
+    sample.subarray(o1, o1);
```
jugglinmike: This test cannot pass; the second parameter should be `o2`

test/built-ins/TypedArray/prototype/subarray/tointeger-begin.js
```diff
@@ -0,0 +1,42 @@
...
+
+  assert(compareArray(sample.subarray(false), [40, 41, 42, 43]), "false");
+  assert(compareArray(sample.subarray(true), [41, 42, 43]), "true");
+
+  assert(compareArray(sample.subarray(NaN), [40, 41, 42, 43]), "NaN");
+  assert(compareArray(sample.subarray(null), [40, 41, 42, 43]), "null");
+  assert(compareArray(sample.subarray(undefined), [40, 41, 42, 43]), "undefined");
+
+  assert(compareArray(sample.subarray(1.1), [41, 42, 43]), "1.1");
+  assert(compareArray(sample.subarray(0.1), [40, 41, 42, 43]), "0.1");
```
jugglinmike: These two assertions basically "prove" the same detail. A useful addition would be a value like `0.6` or `1.5`--these would prove that "floor" (not "round") is being used.

And on that note, a negative floating-point number would be useful here, too. Note how the algorithm converts values like `-1.1` to `-1`--a careless implementation might skip the absolute value part and convert to `-2` instead.

test/built-ins/TypedArray/prototype/subarray/return-abrupt-from-end-symbol.js
```diff
@@ -0,0 +1,25 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-%typedarray%.prototype.subarray
+description: Return abrupt from ToInteger(end), begin is symbol
```
jugglinmike: "begin is a symbol" -> "end is a Symbol"

test/built-ins/TypedArray/prototype/subarray/tointeger-end.js
```diff
@@ -0,0 +1,44 @@
...
+  assert(compareArray(sample.subarray(0, false), []), "false");
+  assert(compareArray(sample.subarray(0, true), [40]), "true");
+
+  assert(compareArray(sample.subarray(0, NaN), []), "NaN");
+  assert(compareArray(sample.subarray(0, null), []), "null");
+  assert(compareArray(sample.subarray(0, undefined), [40, 41, 42, 43]), "undefined");
+
+  assert(compareArray(sample.subarray(0, 1.1), [40]), "1.1");
+  assert(compareArray(sample.subarray(0, 0.1), []), "0.1");
+  assert(compareArray(sample.subarray(0, 0.6), []), "0.6");
```
jugglinmike: Hey look! It's 0.6! (I promise I wrote the above comment before I saw this test.) My suggestion about negative floating-point numbers applies here, too.

jugglinmike:
One very important property of `subarray` that is not currently tested is the
fact the newly-created TypedArray instance has *the same array buffer*. Could
you add a test that verifies this? I think it would be a good idea to check the
`buffer` property values *and* the behavior (e.g. that modifications made
through one instance are visible from the other).

We're also missing tests for the interaction with `Symbol.species`--there are a lot of steps in [SpeciesConstructor](https://tc39.github.io/ecma262/#sec-speciesconstructor) that we should make sure are being hit. Unless you are altering this with your latest spec work...?

jugglinmike:
Finally, I have some reservations about the `testRes` functions. (This concerns
general testing philosophy,, so I had to write a lot. Please bear with me.)

Those functions assert a whole bunch of details about the object under test,
many of which are unrelated to the file's stated intention. I'm not sure this
is a good practice.

By keeping assertions organized and directed, we can have consistent
expectations as maintainers about how tests ought to be authored. The more we
include assertions for details that "ought to be correct, too" but which are
not clearly related to the test, the more subjective our expectations become.

For instance: unlike `results-with-different-length.js`, `minus-zero.js` does
not include assertions for the prototype of the newly-created TypedArray
instance. A reviewer might ask, "Why not? Can you add those assertions there,
too?" But I think it would be a pretty frustrating review process (and one that
produced bulky tests) if we started going back-and-forth about all these
"extra" assertions. If instead, we are consistent in how we encode
expectations--precisely-named files dedicated to single behaviors--then there
is no ambiguity about "what should I be testing here?"

I'm not saying that we shouldn't assert these details. You've identified
important aspects of the behavior of `subarray`, and we definitely want to
ensure that implementors get them right. I just think we need to assert each in
one place.

This is a tricky subject, though, because I am essentially requesting *less*
coverage. For instance, I'm suggesting we remove the specific tests for both
`Int8Array([1, 2, 3]).subarray(0, -2).constructor` *and* `Int8Array([1, 2,
3]).subarray(0, -3).constructor`. It's technically possible that implementors
get one of these wrong, and if you take my advice, then Test262 will not report
this bug.

...but I don't think we can ever hope to catch bugs that specific! If we tried,
then we would probably never finish testing the `subarray` method, let alone
the rest of the language. In my mind, the benefit of clear expectations (which
we get from small, directed tests) is much more valuable than the possibility
of catching completely random bugs (which we get from large, "scattershot"
tests).

This is all very subjective, though. I'm curious to hear what you think!

leobalter:
I applied the fixes and tried to resolve the suggestion on the last comment (with smaller test files).

I still need to cover the SpeciesConstructor.

leobalter:
56bfd79 this might cover the SpeciesConstructor

jugglinmike:
When you say, "might cover," do you mean that you are still working? Or is now a good time for me to take another look?

leobalter:
It is a good time to take another look.

I meant that commit should be enough to reasonably cover the SpeciesConstructor parts.

test/built-ins/TypedArray/prototype/subarray/speciesctor-get-ctor-abrupt.js
```diff
@@ -0,0 +1,39 @@
...
+  3. Let constructor be ? SpeciesConstructor(exemplar, defaultConstructor).
+  ...
+
+  7.3.20 SpeciesConstructor ( O, defaultConstructor )
+
+  1. Assert: Type(O) is Object.
+  2. Let C be ? Get(O, "constructor").
+  3. If C is undefined, return defaultConstructor.
+  ...
+includes: [testTypedArray.js, compareArray.js]
```
jugglinmike: No need for compareArray.js here

test/built-ins/TypedArray/prototype/subarray/speciesctor-get-ctor-inherited.js
```diff
@@ -0,0 +1,40 @@
...
+  3. Let constructor be ? SpeciesConstructor(exemplar, defaultConstructor).
+  ...
+
+  7.3.20 SpeciesConstructor ( O, defaultConstructor )
+
+  1. Assert: Type(O) is Object.
+  2. Let C be ? Get(O, "constructor").
+  3. If C is undefined, return defaultConstructor.
+  ...
+includes: [testTypedArray.js, compareArray.js]
```
jugglinmike: No need for compareArray.js here

test/built-ins/TypedArray/prototype/subarray/speciesctor-get-ctor-returns-throws.js
```diff
@@ -0,0 +1,62 @@
...
+
+  sample.constructor = 42;
+  assert.throws(TypeError, function() {
+    sample.subarray(0);
+  }, "42");
+
+  sample.constructor = "1";
+  assert.throws(TypeError, function() {
+    sample.subarray(0);
+  }, "'1");
```
jugglinmike: Missing single quote

test/built-ins/TypedArray/prototype/subarray/speciesctor-get-ctor.js
```diff
@@ -0,0 +1,40 @@
...
+  3. Let constructor be ? SpeciesConstructor(exemplar, defaultConstructor).
+  ...
+
+  7.3.20 SpeciesConstructor ( O, defaultConstructor )
+
+  1. Assert: Type(O) is Object.
+  2. Let C be ? Get(O, "constructor").
+  3. If C is undefined, return defaultConstructor.
+  ...
+includes: [testTypedArray.js, compareArray.js]
```
jugglinmike: No need for compareArray.js here

test/built-ins/TypedArray/prototype/subarray/speciesctor-get-species-custom-ctor-args.js
```diff
@@ -0,0 +1,56 @@
...
+testWithTypedArrayConstructors(function(TA) {
+  var sample = new TA([40, 41, 42]);
+  var calls = 0;
+  var expectedOffset = TA.BYTES_PER_ELEMENT;
+  var result;
+
+  sample.constructor = {};
+  sample.constructor[Symbol.species] = function(buffer, offset, length) {
+    result = arguments;
+    return new TA(buffer, offset, length);
```
jugglinmike: How about an assertion that the `this` value in this function is an instance of the function itself?  Maybe in another file? Or rename this one to "ctor-invocation"?

jugglinmike:
Hey Leo! I've left a few notes in-line; here are some thoughts that didn't fit
with any one line:

More than just asserting that the accessor is called as expected, these two
tests ought to verify that the `defaultConstructor` is used in these cases. (In
other words: step 3 of SpeciesConstructor.)

- speciesctor-get-ctor-inherited.js
- speciesctor-get-ctor.js

The file named `speciesctor-get-species-use-default-ctor-throws.js` does not
concern a "throw" completion, so it should be renamed.

test/built-ins/TypedArray/prototype/subarray/speciesctor-get-ctor-returns-throws.js
leobalter: that's my editor trolling with my international keyboard setup.

leobalter:
fixes applied, I'll add an inline comment on a specific case

leobalter:
no extra comments, just got confused with my own tests for a moment.

test/built-ins/TypedArray/prototype/subarray/speciesctor-get-ctor-inherited.js
```diff
@@ -0,0 +1,52 @@
...
+
+  assert.sameValue(
+    Object.getPrototypeOf(result),
+    Object.getPrototypeOf(sample),
+    "use defaultCtor on an undefined return - getPrototypeOf check"
+  );
+  assert.sameValue(
+    result.constructor,
+    undefined,
+    "used defaultCtor but still checks the inherited .constructor"
```
jugglinmike: This is a little confusing, and maybe too imprecise. As written, it's possible
for the test to pass if the instance stored in `result` has an "own" property
whose value is `undefined`. Unlikely, I know, but I think it's worth addressing
to improve the test's clarity. I can think of three ways to resolve this:

- Add an assertion using `hasOwnProperty`
- Add an assertion that `calls` is now `2`
- Define the `constructor` accessor property on the `sample` instance--not the
  `TA` prototype

The third approach is the most consistent with the rest of your work (see
`speciesctor-get-ctor.js`, for instance), so maybe that's the best choice.

jugglinmike:
One last thing here, Leo!

test/built-ins/TypedArray/prototype/subarray/speciesctor-get-ctor-inherited.js
leobalter: I fixed this in 1091944

> Add an assertion using hasOwnProperty 

That wouldn't be very clear as this file is testing the Get("constructor") on an inherited property.

>  Add an assertion that calls is now 2

I did it on a slightly different approach, setting a custom value to calls and than checking a get on `result.constructor` to verify it is using that property.

> Define the constructor accessor property on the sample instance--not the TA prototype

That's the scope for `speciesctor-get-ctor` only.

jugglinmike: I don't think I explained myself very well. Let me try again.

The confusing part for me was that the assertion is "proving" that
`result.constructor` is `undefind`, but there are many ways this could be true.
It could be that the instance has an "own" property whose value is `undefined`,
it could be that the instance has no such "own" property, but that its
prototype does, or it could be that a property with that name does not exist
anywhere in the prototype chain at all.

The assertion message helps explain what is happening, but I was suggesting
explicit assertions that would also verify that.

- An assertion using `hasOwnProperty` would clearly demonstrate that the value
  `undefined` is *not* coming directly from the `result` object itself.

    ```js
    assert.sameValue(Object.hasOwnProperty.call(result, "constructor"), false);
    ```

- An assertion that `calls` is now `2` would demonstrate that the previous
  assertion had triggered the invocation of the accessor property.

    ```js
    assert.sameValue(calls, 2);
    ```

  The newest assertion  does not demonstrate this because it re-sets the
  `calls` variable. So as written, it seems like an unrelated test for accessor
  properties. Asserting the value `2` *does* introduce a slight dependency
  between the assertions, and I can appreciate if you want to avoid that. You
  could avoid this by re-setting `calls` to `0` immediately before the
  assertion.

That said, we are definitely going too far into tiny details. I haven't
forgotten that the test is valid, all this discussion is just about a minor
readability improvement. I think it's fine to leave it as-is.

> That's the scope for speciesctor-get-ctor only.

Ah, of course.

goyakin:
Merged as 851be61.

PR #592, opened 2016-04-21
Fix test for TypedArray.prototype.forEach

leobalter:


jugglinmike:
Technically, this would have caught the same bug either way, but the failure message definitely would have been confusing.

Merged to `master` at commit bc8ae6e0aeca3fd446367091ea1f2549e6087a9f

PR #593, opened 2016-04-21
Add tests for %TypedArray%.prototype.{every, some}

leobalter:


test/built-ins/TypedArray/prototype/every/get-length-uses-internal-arraylength.js
```diff
@@ -0,0 +1,48 @@
...
+
+  sample.every(function() {
+    calls++;
+    return true;
+  });
+
+  assert.sameValue(getCalls, 0, "ignores length properties");
+
+  // Is that reachable or necessary? At least we know it worked.
+  assert.sameValue(calls, 2, "interactions are not affected by custom length");
```
leobalter: @jugglinmike ^^ (in-line comment made pre-PR)

jugglinmike:
Hi Leo,

Because of the symmetry of these tests, almost all of my feedback applies in
two places. Rather than duplicating all my comments for each file, I'll just
leave my suggestions here:

- `returns-false-if-any-cb-returns-false.js` You can remove the "features" tag
- `callbackfn-no-interaction-over-non-integer.js` For consistency with
  `callbackfn-arguments-without-thisarg.js`, I suggest simply pushing the
  `arguments` object (instead of an array literal)
- `callbackfn-set-value-during-interaction.js` One detail missing from these
  tests is whether the modifications propagate to the iterated value. An
  implementor might incorrectly cache the values up-front--in these cases,
  `sample` could be modified (and this test would pass), but the value of `val`
  (or "kValue" in the algorithm) would be "stale".

Finally, the file `some/returns-false-if-any-cb-returns-false.js` should be
renamed to `some/returns-false-if-every-cb-returns-false.js`.

...that's all!

leobalter:
> callbackfn-no-interaction-over-non-integer.js For consistency with callbackfn-arguments-without-thisarg.js, I suggest simply pushing the arguments object (instead of an array literal)

that was following the tests for forEach and map, I'll update for them as well

leobalter:
Fixes are up. 

Would you take a look on [this part](https://github.com/tc39/test262/pull/593#discussion-diff-60645870)?

test/built-ins/TypedArray/prototype/some/values-are-not-cached.js
```diff
@@ -0,0 +1,41 @@
...
+
+  ...
+  6. Repeat, while k < len
+    ..
+    c. If kPresent is true, then
+      i. Let kValue be ? Get(O, Pk).
+      ii. Let testResult be ToBoolean(? Call(callbackfn, T, « kValue, k, O »)).
+  ...
+includes: [testTypedArray.js]
+features: [Reflect.set]
```
jugglinmike: We can remove this tag

test/built-ins/TypedArray/prototype/some/values-are-not-cached.js
```diff
@@ -0,0 +1,41 @@
...
+---*/
+
+testWithTypedArrayConstructors(function(TA) {
+  var sample = new TA([42, 43, 44]);
+  var calls = 0;
+
+  sample.map(function(v, i) {
```
jugglinmike: This should use `some`, not `map`

jugglinmike:
> Would you take a look on [this
> part](https://github.com/tc39/test262/pull/593#discussion-diff-60645870)?

I don't think that is strictly necessary, but I suppose it couldn't hurt to
leave as-is. We should remove the comment in either case.

Anyway, I've left just a couple minor requests. Once those are resolved, this should be good to go.

leobalter:
Done!

jugglinmike:
Beautiful :)

goyakin:
Merged as 35c00ca.

PR #594, opened 2016-04-22
Increase coverage for Symbol constructor

jugglinmike:
I realized that we were missing a test for `new Symbol()` while review one of
@leobalter's pull requests. When I went to add it, I found that still other
tests were missing, and that some of the existing tests weren't organized
correctly.

> Add new tests for invocation requirements; re-organize existing files
> according to the language feature under test.

### DIFF
--- a/test/built-ins/Symbol/desc-to-string-symbol.js
+++ b/test/built-ins/Symbol/desc-to-string-symbol.js
@@ -1,0 +1,19 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-symbol-constructor
+es6id: 19.4.1
+description: The first argument is coerced to a String value
+info: |
+    1. If NewTarget is not undefined, throw a TypeError exception.
+    2. If description is undefined, let descString be undefined.
+    2. Else, let descString be ? ToString(description).
+    3. Return a new unique Symbol value whose [[Description]] value is
+       descString.
+---*/
+
+var s = Symbol('1');
+
+assert.throws(TypeError, function() {
+  Symbol(s);
+});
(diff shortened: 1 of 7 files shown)


test/built-ins/Symbol/desc-to-string.js
```diff
@@ -0,0 +1,88 @@
...
+
+calls = '';
+assert.throws(TypeError, function() {
+  Symbol(val);
+});
+assert.sameValue(calls, 'toStringvalueOf');
+
+assert.throws(TypeError, function() {
+  Symbol(Symbol('1'));
+});
```
leobalter: it would be good to add a message to these throws assertions, this way it's easier to keep track what exactly failed.

Also, this last assertion seems isolated to the other cases, can we have it in an extra file? It's a good justification for the fact it's not possible to make a Symbol from another Symbol.

leobalter:
just a single inline comment above, the other parts look good to me

test/built-ins/Symbol/desc-to-string.js
jugglinmike: Sounds good to me

test/built-ins/Symbol/desc-to-string-symbol.js
```diff
@@ -0,0 +1,19 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-symbol-constructor
+es6id: 19.4.1
+description: The first argument is coerced to a String value
```
leobalter: This could use a distinct description from [`test/built-ins/Symbol/desc-to-string.js`](test/built-ins/Symbol/desc-to-string.js
)

jugglinmike: I beat you to the punch!

leobalter: :8ball:

leobalter:
LGTM

goyakin:
Merged as 35c00ca. Thanks @jugglinmike.

Issue #568, opened 2016-04-04
Reducing reliance on fnGlobalObject.js


jugglinmike:
While reviewing [gh-567: "Add `System.global` tests"](https://github.com/tc39/test262/pull/567), I was about to suggest the use of [the `fnGlobalObject` helper function](https://github.com/tc39/test262/blob/755b0d61171a7b16fd81979bc8d6f24bb832ecc6/harness/fnGlobalObject.js) instead of a simple `this` reference to test the property value.

I'm not sure this is necessary, though, and really, I'm not sure `fnGlobalObject` is necessary in *any* case that Test262 currently uses it.

Test runners are required to execute test contents in the global scope--anything else would invalidate certain tests. The `this` keyword has long (always?) been specified as returning the global object when evaluated in global code.

If this is true, I would be in favor of simplifying the ~200 tests that include `fnGlobalObject.js` to reference `this` directly, and (more importantly) reducing the domain-specific knowledge required to author new tests.

I'd like to keep the helper around, though, because it will still be necessary for accessing the global from module code (see [gh-562: "Module semantics: declaration instantiation"](https://github.com/tc39/test262/pull/562)), and it seems possible that security restrictions (e.g. CSP in web browsers) may lead some consumers to look for a mechanism to override that function.

If there is some detail that I'm overlooking, then I'm happy to leave the usage as-is (although I may want to document the use case(s) as an in-line comment in `fnGlobalObject.js` itself).

From ES5's [11.1.1 The this Keyword](https://es5.github.io/#x11.1.1):

> The this keyword evaluates to the value of the ThisBinding of the current
> execution context.

And from [10.4.1.1 Initial Global Execution Context](https://es5.github.io/#x10.4.1.1):

> The following steps are performed to initialize a global execution context
> for ECMAScript code C:
>
> 1. Set the VariableEnvironment to the Global Environment.
> 2. Set the LexicalEnvironment to the Global Environment.
> 3. Set the ThisBinding to the global object.

leobalter:
> If this is true, I would be in favor of simplifying the ~200 tests that include fnGlobalObject.js to reference this directly, and (more importantly) reducing the domain-specific knowledge required to author new tests.

Agreed.

> I'd like to keep the helper around, though, because it will still be necessary for accessing the global from module code (see gh-562: "Module semantics: declaration instantiation"), and it seems possible that security restrictions (e.g. CSP in web browsers) may lead some consumers to look for a mechanism to override that function.

+1 to keep the helper 

[...]

PR #595, opened 2016-04-22
Reduce reliance on `fnGlobalObject.js`

jugglinmike:
This harness function is not necessary in the majority of cases in which
it is used. Remove its usage to simplify tests and decrease the amount
of domain-specific knowledge necessary to contribute to the test suite.

Persist the harness function itself for use by future tests for ES2015
modules (such a helper is necessary for tests that are interpreted as
module code).

This resolves gh-568

### DIFF
--- a/test/built-ins/Array/isArray/15.4.3.2-1-15.js
+++ b/test/built-ins/Array/isArray/15.4.3.2-1-15.js
@@ -4,7 +4,6 @@
 /*---
 es5id: 15.4.3.2-1-15
 description: Array.isArray applied to the global object
-includes: [fnGlobalObject.js]
 ---*/
 
-assert.sameValue(Array.isArray(fnGlobalObject()), false, 'Array.isArray(fnGlobalObject())');
+assert.sameValue(Array.isArray(this), false, 'Array.isArray(fnGlobalObject())');
--- a/test/built-ins/Array/isArray/15.4.3.2-1-15.js
+++ b/test/built-ins/Array/isArray/15.4.3.2-1-15.js
@@ -6,4 +6,4 @@ es5id: 15.4.3.2-1-15
 description: Array.isArray applied to the global object
 ---*/
 
-assert.sameValue(Array.isArray(this), false, 'Array.isArray(fnGlobalObject())');
+assert.sameValue(Array.isArray(this), false, 'Array.isArray(this)');
--- a/test/built-ins/Array/prototype/every/15.4.4.16-5-21.js
+++ b/test/built-ins/Array/prototype/every/15.4.4.16-5-21.js
@@ -4,15 +4,15 @@
 /*---
 es5id: 15.4.4.16-5-21
 description: Array.prototype.every - the global object can be used as thisArg
-includes: [fnGlobalObject.js]
 ---*/
 
+var global = this;
         var accessed = false;
 
         function callbackfn(val, idx, obj) {
             accessed = true;
-            return this === fnGlobalObject();
+            return this === global;
         }
 
-assert([11].every(callbackfn, fnGlobalObject()), '[11].every(callbackfn, fnGlobalObject()) !== true');
+assert([11].every(callbackfn, global), '[11].every(callbackfn, fnGlobalObject()) !== true');
 assert(accessed, 'accessed !== true');
--- a/test/built-ins/Array/prototype/every/15.4.4.16-5-21.js
+++ b/test/built-ins/Array/prototype/every/15.4.4.16-5-21.js
@@ -14,5 +14,5 @@ var global = this;
             return this === global;
         }
 
-assert([11].every(callbackfn, global), '[11].every(callbackfn, fnGlobalObject()) !== true');
+assert([11].every(callbackfn, global), '[11].every(callbackfn, global) !== true');
 assert(accessed, 'accessed !== true');
--- a/test/built-ins/Array/prototype/every/15.4.4.16-7-c-i-23.js
+++ b/test/built-ins/Array/prototype/every/15.4.4.16-7-c-i-23.js
@@ -6,7 +6,6 @@ es5id: 15.4.4.16-7-c-i-23
 description: >
     Array.prototype.every - This object is an global object which
     contains index property
-includes: [fnGlobalObject.js]
 ---*/
 
         function callbackfn(val, idx, obj) {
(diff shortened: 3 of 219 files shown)


test/built-ins/Array/isArray/15.4.3.2-1-15.js
```diff
@@ -4,7 +4,6 @@
 /*---
 es5id: 15.4.3.2-1-15
 description: Array.isArray applied to the global object
-includes: [fnGlobalObject.js]
 ---*/
 
-assert.sameValue(Array.isArray(fnGlobalObject()), false, 'Array.isArray(fnGlobalObject())');
+assert.sameValue(Array.isArray(this), false, 'Array.isArray(fnGlobalObject())');
```
leobalter: the assertion message still has the old method

test/built-ins/Array/prototype/every/15.4.4.16-7-c-i-23.js
```diff
@@ -17,8 +16,8 @@ includes: [fnGlobalObject.js]
             }
         }
 
-            var oldLen = fnGlobalObject().length;
-            fnGlobalObject()[0] = 11;
-            fnGlobalObject().length = 1;
+            var oldLen = this.length;
```
leobalter: I know this is not the focus for this PR but this line is unnecessary.

test/built-ins/Array/prototype/filter/15.4.4.20-2-15.js
```diff
@@ -4,18 +4,18 @@
...
-            fnGlobalObject()[1] = 11;
-            fnGlobalObject()[2] = 9;
-            fnGlobalObject().length = 2;
-            var newArr = Array.prototype.filter.call(fnGlobalObject(), callbackfn);
+            var oldLen = global.length;
+            global[0] = 12;
+            global[1] = 11;
+            global[2] = 9;
+            global.length = 2;
+            var newArr = Array.prototype.filter.call(global, callbackfn);
```
leobalter: you can use `this` instead of `global` here, right?

test/built-ins/Array/prototype/filter/15.4.4.20-9-c-iii-28.js
```diff
@@ -6,11 +6,11 @@ es5id: 15.4.4.20-9-c-iii-28
 description: >
     Array.prototype.filter - return value of callbackfn is the global
     object
-includes: [fnGlobalObject.js]
 ---*/
 
+var global = this;
         function callbackfn(val, idx, obj) {
-            return fnGlobalObject();
+            return global;
```
leobalter: this whole test seems unnecessary. It's testing a `ToBoolean(global) === true`. That can be done with a simple `{}`.

test/built-ins/Array/prototype/lastIndexOf/15.4.4.15-1-17.js
```diff
@@ -4,13 +4,12 @@
...
 
-            var oldLen = fnGlobalObject().length;
-            fnGlobalObject()[1] = targetObj;
-            fnGlobalObject().length = 3;
+            var oldLen = this.length;
+            this[1] = targetObj;
+            this.length = 3;
 
-assert.sameValue(Array.prototype.lastIndexOf.call(fnGlobalObject(), targetObj), 1, 'Array.prototype.lastIndexOf.call(fnGlobalObject(), targetObj)');
+assert.sameValue(Array.prototype.lastIndexOf.call(this, targetObj), 1, 'Array.prototype.lastIndexOf.call(this, targetObj)');
```
leobalter: this is another case that could be done with an ordinary object.

And that's another case that shouldn't be focus of this PR.

test/built-ins/Array/prototype/some/15.4.4.17-5-21.js
```diff
@@ -4,11 +4,11 @@
...
 ---*/
 
+var global = this;
         function callbackfn(val, idx, obj) {
-            return this === fnGlobalObject();
+            return this === global;
         }
 
-assert([11].some(callbackfn, fnGlobalObject()), '[11].some(callbackfn, fnGlobalObject()) !== true');
+assert([11].some(callbackfn, this), '[11].some(callbackfn, fnGlobalObject()) !== true');
```
leobalter: a remaining `fnGlobalObject()` in the message

test/built-ins/Infinity/S15.1.1.2_A3_T1.js
```diff
@@ -5,14 +5,14 @@
...
+includes: [propertyHelper.js]
 ---*/
 
 // CHECK#1
-verifyNotConfigurable(fnGlobalObject(), "Infinity");
+verifyNotConfigurable(this, "Infinity");
 
 try {
-  if (delete fnGlobalObject().Infinity !== false) {
+  if (delete this.Infinity !== false) {
```
leobalter: this is another off-topic. This case is unnecessary, already covered at line 12 with verifyNotConfigurable

test/built-ins/Array/prototype/every/15.4.4.16-5-21.js
```diff
@@ -4,15 +4,15 @@
...
         var accessed = false;
 
         function callbackfn(val, idx, obj) {
             accessed = true;
-            return this === fnGlobalObject();
+            return this === global;
         }
 
-assert([11].every(callbackfn, fnGlobalObject()), '[11].every(callbackfn, fnGlobalObject()) !== true');
+assert([11].every(callbackfn, global), '[11].every(callbackfn, fnGlobalObject()) !== true');
```
leobalter: remaining fnGlobalObject on the message

leobalter:
a few comments above, the current tests are looking good.

we can delete `harness/fnGlobalObject.js` as well, right?

There are some cases declaring a `oldLen` to store `global.length` variable which is totally unnecessary. This is not the focus of this PR, but if it's not handled here, they should be removed on a follow up patch. See https://github.com/tc39/test262/pull/595#discussion-diff-60796648

There are some other tests requiring some updates for good. Once again, not in this PR. See https://github.com/tc39/test262/pull/595#discussion-diff-60797444

test/built-ins/Array/isArray/15.4.3.2-1-15.js
jugglinmike: Whoops! I'll update all three assertion messages

test/built-ins/Array/prototype/every/15.4.4.16-7-c-i-23.js
jugglinmike: Agreed on both counts

test/built-ins/Array/prototype/filter/15.4.4.20-2-15.js
jugglinmike: Definitely. Change incoming

jugglinmike:
> we can delete `harness/fnGlobalObject.js` as well, right?

I'd rather not--the utility function is still useful for module code, where
the global `this` is `undefined`. I tried to document this concern in the
commit message, but I'd be happy to re-word that if you think it could be more
clear.

I agree that it would be good to clean these tests up, and also that it is
better left for another patch.

Thanks for the review!

leobalter:
> the utility function is still useful for module code

That's right, I forgot about it.

> Persist the harness function itself for use by future tests for ES2015
modules (such a helper is necessary for tests that are interpreted as
module code).

My bad I started the code review minutes after I opened this PR (locally first). This message is great.

leobalter:
LGTM

goyakin:
Merged as 7f88f29.

PR #596, opened 2016-04-22
Add tests for return value of @@species accessors

jugglinmike:


leobalter:
We could add tests for missing prop-desc, asserting the _set_ accessor is __undefined__ as in [here](https://github.com/tc39/test262/blob/master/test/built-ins/Promise/Symbol.species/prop-desc.js)

jugglinmike:
Hey Leo: I've added the assertions that you requested.

leobalter:
LGTM

goyakin:
Merged as 85e6ab8.

PR #598, opened 2016-04-22
Update tests for TypedArrays foreach and map

leobalter:
Based on the feedback from gh-593

### DIFF
--- a/test/built-ins/TypedArray/prototype/forEach/values-are-not-cached.js
+++ b/test/built-ins/TypedArray/prototype/forEach/values-are-not-cached.js
@@ -13,7 +13,6 @@ info: >
   this object's [[ArrayLength]] internal slot is accessed in place of performing
   a [[Get]] of "length"
 includes: [testTypedArray.js]
-features: [Reflect.set]
 ---*/
 
 testWithTypedArrayConstructors(function(TA) {
--- a/test/built-ins/TypedArray/prototype/forEach/values-are-not-cached.js
+++ b/test/built-ins/TypedArray/prototype/forEach/values-are-not-cached.js
@@ -17,7 +17,6 @@ includes: [testTypedArray.js]
 
 testWithTypedArrayConstructors(function(TA) {
   var sample = new TA([42, 43, 44]);
-  var calls = 0;
 
   sample.forEach(function(v, i) {
     if (i < sample.length - 1) {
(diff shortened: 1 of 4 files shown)


test/built-ins/TypedArray/prototype/forEach/values-are-not-cached.js
```diff
@@ -0,0 +1,32 @@
...
+  Integer indexed values are not cached before interaction
+info: >
+  22.2.3.12 %TypedArray%.prototype.forEach ( callbackfn [ , thisArg ] )
+
+  %TypedArray%.prototype.forEach is a distinct function that implements the same
+  algorithm as Array.prototype.forEach as defined in 22.1.3.10 except that the
+  this object's [[ArrayLength]] internal slot is accessed in place of performing
+  a [[Get]] of "length"
+includes: [testTypedArray.js]
+features: [Reflect.set]
```
leobalter: removing...

test/built-ins/TypedArray/prototype/forEach/values-are-not-cached.js
```diff
@@ -0,0 +1,31 @@
...
+  %TypedArray%.prototype.forEach is a distinct function that implements the same
+  algorithm as Array.prototype.forEach as defined in 22.1.3.10 except that the
+  this object's [[ArrayLength]] internal slot is accessed in place of performing
+  a [[Get]] of "length"
+includes: [testTypedArray.js]
+---*/
+
+testWithTypedArrayConstructors(function(TA) {
+  var sample = new TA([42, 43, 44]);
+  var calls = 0;
```
goyakin: I don't think this was intended to be used in this test. If so, can we remove it?

test/built-ins/TypedArray/prototype/map/values-are-not-cached.js
```diff
@@ -0,0 +1,26 @@
...
+description: >
+  Integer indexed values changed during iteration
+info: >
+  22.2.3.19 %TypedArray%.prototype.map ( callbackfn [ , thisArg ] )
+includes: [testTypedArray.js]
+---*/
+
+testWithTypedArrayConstructors(function(TA) {
+  var sample = new TA([42, 43, 44]);
+  var calls = 0;
```
goyakin: We should remove this one too if we don't intend to use it.

goyakin:
@leobalter Other than my comments for the `calls` variables, everything else LGTM.

jugglinmike:
I agree with Gorkem; this looks good to merge once those variables are removed.

leobalter:
fixes are up! Thanks!

goyakin:
Merged as 84d9507.

PR #599, opened 2016-04-25
Add tests for %TypedArray%.prototype.toLocaleString

leobalter:


test/built-ins/TypedArray/prototype/toLocaleString/calls-tolocalestring-from-each-value.js
```diff
@@ -0,0 +1,52 @@
...
+
+var arr = [42, 0];
+var expected = ["hacks1", "hacks2"].join(separator);
+
+testWithTypedArrayConstructors(function(TA) {
+  var sample = new TA(arr);
+  calls = [];
+  assert.sameValue(sample.toLocaleString(), expected, "returns expected value");
+  assert(
+    compareArray(new TA(calls), sample),
```
leobalter: this is the best I got to handle number conversions.

jugglinmike: I am having trouble understanding the intent of this test. How is it different than `test/built-ins/TypedArray/prototype/toLocaleString/calls-valueof-from-each-value.js`? Could we cover this detail by extending that file with an assertion for the `this` value?

test/built-ins/TypedArray/prototype/toLocaleString/return-abrupt-from-firstelement-tolocalestring.js
```diff
@@ -0,0 +1,43 @@
...
+  5. Let firstElement be ? Get(array, "0").
+  6. If firstElement is undefined or null, then
+    a. Let R be the empty String.
+  7. Else,
+    a. Let R be ? ToString(? Invoke(firstElement, "toLocaleString")).
+includes: [testTypedArray.js]
+---*/
+
+var calls;
+var separator = ["", ""].toLocaleString();
```
jugglinmike: It looks like this variable can be removed

test/built-ins/TypedArray/prototype/toLocaleString/return-abrupt-from-nextelement-tolocalestring.js
```diff
@@ -0,0 +1,46 @@
...
+    b. Let nextElement be ? Get(array, ! ToString(k)).
+    c. If nextElement is undefined or null, then
+      i. Let R be the empty String.
+    d. Else,
+      i. Let R be ? ToString(? Invoke(nextElement, "toLocaleString")).
+includes: [testTypedArray.js]
+---*/
+
+var calls = 0;
+var separator = ["", ""].toLocaleString();
```
jugglinmike: It looks like this variable can be removed

test/built-ins/TypedArray/prototype/toLocaleString/calls-tolocalestring-from-each-value.js
goyakin: @jugglinmike Doesn't `test/built-ins/TypedArray/prototype/toLocaleString/calls-valueof-from-each-value.js` check that the value returned from `toLocaleString` goes through `ToString` whereas this test checks whether `toLocaleString` is called correctly? I would keep them separate since their intentions are different.

goyakin: @leobalter Why do we create a new TypedArray here (i.e., `new TA(calls)`)? Wouldn't `compareArray(calls, arr)` be enough since the values in `arr` don't overflow any of the array types?

goyakin:
@leobalter Other than what's mentioned inline, everything else LGTM.

test/built-ins/TypedArray/prototype/toLocaleString/calls-tolocalestring-from-each-value.js
leobalter: because `calls` has the ToObject value of each number.

the `this` values on `Number#toLocaleString` are cast to objects, I'm reusing the typedArray to recast them back to numbers.

See step 6.b.i on [9.2.1.2 OrdinaryCallBindThis](https://tc39.github.io/ecma262/#sec-ordinarycallbindthis)

This is gonna happen after step 7.a of [Array#toLocaleString](https://tc39.github.io/ecma262/#sec-array.prototype.tolocalestring) will [`Invoke`](https://tc39.github.io/ecma262/#sec-invoke) returning [`Call(func, V, argumentsList).`](https://tc39.github.io/ecma262/#sec-call) which will return `? F.[[Call]](V, argumentsList)`. You can find `OrdinaryCallBindThis` being called on step 6 of [9.2.1 `[[Call]]`](https://tc39.github.io/ecma262/#sec-ecmascript-function-objects-call-thisargument-argumentslist).

leobalter: > @jugglinmike How is it different than `test/built-ins/TypedArray/prototype/toLocaleString/calls-valueof-from-each-value.js`?

This assertion covers the `this` value on the `Number#toLocaleString`. It looks like it could be part of that file or `test/built-ins/TypedArray/prototype/toLocaleString/calls-tostring-from-each-value.js`, but I preferred leaving it on this one. I can move it if necessary.

leobalter:
fixes are up removing the unnecessary variables.

test/built-ins/TypedArray/prototype/toLocaleString/calls-tolocalestring-from-each-value.js
jugglinmike: I'm comfortable keeping the files separate

jugglinmike:
Merged to `master` at commit 177b2f6018f0abb24150154386f715fc9e1bd641. Thanks!

PR #600, opened 2016-04-26
Add tests for ArraySpeciesCreate

jugglinmike:
Assert the expected behavior of the ArraySpeciesCreate abstract
operation in the 5 Array instance methods from which it is invoked.

This change set does not include tests for ECMAScript realm
considerations because Test262 does not currently expose a mechanism for
interfacing with realms.

### DIFF
--- a/test/built-ins/Array/prototype/filter/create-species-non-ctor.js
+++ b/test/built-ins/Array/prototype/filter/create-species-non-ctor.js
@@ -1,0 +1,37 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 22.1.3.7
+esid: sec-array.prototype.filter
+description: >
+    Behavior when the @@species attribute is a non-constructor object
+info: |
+    [...]
+    5. Let A be ? ArraySpeciesCreate(O, 0).
+    [...]
+
+    9.4.2.3 ArraySpeciesCreate
+
+    [...]
+    5. Let C be ? Get(originalArray, "constructor").
+    [...]
+    7. If Type(C) is Object, then
+       a. Let C be ? Get(C, @@species).
+       b. If C is null, let C be undefined.
+    [...]
+    9. If IsConstructor(C) is false, throw a TypeError exception.
+features: [Symbol.species]
+---*/
+
+var a = [];
+var callCount = 0;
+var cb = function() { callCount += 1; };
+var result;
+
+a.constructor = {};
+a.constructor[Symbol.species] = parseInt;
+
+assert.throws(TypeError, function() {
+  a.filter(cb);
+});
+assert.sameValue(callCount, 0);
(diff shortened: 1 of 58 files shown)


test/built-ins/Array/prototype/filter/create-species-non-ctor.js
```diff
@@ -0,0 +1,37 @@
...
+       b. If C is null, let C be undefined.
+    [...]
+    9. If IsConstructor(C) is false, throw a TypeError exception.
+features: [Symbol.species]
+---*/
+
+var a = [];
+var callCount = 0;
+var cb = function() { callCount += 1; };
+var result;
```
goyakin: `result` isn't needed.

goyakin:
@jugglinmike A `result` variable is declared in a few test files, but they aren't actually used, so we should remove them. Everything else LGTM.

jugglinmike:
Thanks for the review, Gorkem! I've removed all occurrence of unused variables (this includes `result` plus a few function parameters).

jugglinmike:
Looks like I forgot to push those "fixup" commits (thanks for letting me know, @leobalter). This branch now includes the requested changes.

test/built-ins/Array/prototype/splice/create-species-neg-zero.js
```diff
@@ -0,0 +1,50 @@
...
+    4. If relativeStart < 0, let actualStart be max((len + relativeStart), 0);
+       else let actualStart be min(relativeStart, len).
+    5. If the number of actual arguments is 0, then
+       [...]
+    6. Else if the number of actual arguments is 1, then
+       [...]
+    7. Else,
+       a. Let insertCount be the number of actual arguments minus 2.
+       b. Let dc be ? ToInteger(deleteCount).
+       c. Let actualDeleteCount be min(max(dc, 0), len - actualStart).
```
leobalter: This is just cosmetic, but it is cloudy for me. I believe `actualDeleteCount` becomes `+0` here on the max operation. In any way this is not observable before the step 2 of 9.4.2.3.

test/built-ins/Array/prototype/splice/create-species-neg-zero.js
```diff
@@ -0,0 +1,50 @@
...
+  args = arguments;
+};
+var a = [];
+a.constructor = {};
+a.constructor[Symbol.species] = Ctor;
+
+a.splice(0, -0);
+
+assert.sameValue(args.length, 1);
+assert.sameValue(args[0], 0);
```
leobalter: and I always need to remind myself `sameValue` handles -0 and +0 as different values.

test/built-ins/Array/prototype/splice/create-species-neg-zero.js
jugglinmike: I disagree. The reason I included that step in this file is to justify the test. "max" does not distinguish between negative zero and positive zero, so the result of "max(0, -0)" is implementation-dependent.

leobalter: > so the result of "max(0, -0)" is implementation-dependent.

I can't say I love the implementation-dependent parts, but it's reasonable. Thanks for the quick answer.

PR #601, opened 2016-04-26
Reduce noise in output of test generation tool

jugglinmike:
This patch corrects a number of errors in the test generation tool which led to
undesirable artifacts in the output files, as identified by @leobalter in the
review of gh-584.

### DIFF
--- a/tools/generation/lib/template.py
+++ b/tools/generation/lib/template.py
@@ -20,7 +20,11 @@ def indent(text, prefix = '    '):
     else:
         lines = text.split('\n')
 
-    return prefix + ('\n' + prefix).join(lines)
+    indented = map(
+        lambda line: line if len(line) == 0 else prefix + line,
+        lines)
+
+    return '\n'.join(indented)
 
 class Template:
     def __init__(self, filename):
--- a/tools/generation/lib/template.py
+++ b/tools/generation/lib/template.py
@@ -119,7 +119,7 @@ class Template:
         features += case_values['meta'].get('features', [])
         features += self.attribs['meta'].get('features', [])
         if len(features):
-            lines += ['features: ' + yaml.dump(features)]
+            lines += ['features: ' + yaml.dump(features).strip()]
 
         flags = ['generated']
         flags += case_values['meta'].get('flags', [])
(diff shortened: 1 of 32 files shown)


tools/generation/test/fixtures/normal/info-multiline-folding.template
```diff
@@ -0,0 +1,11 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+name: First template name
+path: normal/info-multiline-folding-
+es6id: 1.2.3
+info: |
+    This is an "info" field with an
+
+    empty line
+---*/
```
leobalter: how about leaving a trailing whitespace in the frontmatter to see it is removed and another in the code where it probably should be kept?

jugglinmike: This is a good idea, but I disagree with the expected behavior. I think that trailing whitespace should be preserved in all cases. I want to keep the test generator as transparent as possible, so that the output can be clearly traced back to the input. Let me know if you have a different interpretation.

Barring that, I'll update this patch to preserve trailing white space in the frontmatter in cases where it is present in the source material. I'll include the tests you've mentioned, of course.

leobalter: > I think that trailing whitespace should be preserved in all cases.

++

Thanks!

jugglinmike:
Okay, I've addressed the consistency issue. Autosquashing this change log will produce conflicts. Resolution should be pretty clear, but I'm happy to do the honors and force-push the result to this branch. Let me know!

tools/generation/lib/template.py
```diff
@@ -147,9 +151,7 @@ def _frontmatter(self, case_filename, case_values):
 
         if len(info):
             lines.append('info: |')
-            lines += map(
-                lambda line: re.sub(trailingWhitePattern, r'\1', line),
-                info)
+            lines += info
```
leobalter: nice move

leobalter:
> I'm happy to do the honors and force-push the result to this branch. 

Please.

+1 to the new tests.

jugglinmike:
All set

PR #602, opened 2016-04-26
Add tests for %TypedArray%.prototype.slice

leobalter:


test/built-ins/TypedArray/prototype/slice/detached-buffer-get-custom-ctor.js
```diff
@@ -0,0 +1,36 @@
...
+      i. Let Pk be ! ToString(k).
+      ii. Let kValue be ? Get(O, Pk).
+  ...
+includes: [testTypedArray.js, detachArrayBuffer.js]
+features: [Symbol.species]
+---*/
+
+testWithTypedArrayConstructors(function(TA) {
+  var sample = new TA();
+  var calls = 0;
```
jugglinmike: `calls` is unused

test/built-ins/TypedArray/prototype/slice/detached-buffer-get-custom-ctor.js
```diff
@@ -0,0 +1,36 @@
...
+
+  sample.constructor = {};
+  sample.constructor[Symbol.species] = function(count) {
+    var other = TA === Int8Array ? Uint8Array : Int8Array;
+    $DETACHBUFFER(sample.buffer);
+    return new other(count);
```
jugglinmike: I think you can simplify this by just using `TA` instead of creating `other`

test/built-ins/TypedArray/prototype/slice/results-with-empty-length.js
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  testRes(sample.slice(4), "begin == length");
+  testRes(sample.slice(5), "begin > length");
+
+  testRes(sample.slice(4, 4), "begin == length, end == length");
+  testRes(sample.slice(5, 4), "begin > length, end == length");
+
+  testRes(sample.slice(4, 4), "begin == length, end > length");
+  testRes(sample.slice(5, 4), "begin > length, end > length");
```
jugglinmike: It looks like you meant to specify `5` as the second argument in these two calls (otherwise, the assertions are identical to the two that precede them).

test/built-ins/TypedArray/prototype/slice/speciesctor-get-species-custom-ctor-invocation.js
```diff
@@ -0,0 +1,61 @@
...
+    ...
+  4. Return newTypedArray.
+includes: [testTypedArray.js]
+features: [Symbol.species]
+---*/
+
+testWithTypedArrayConstructors(function(TA) {
+  var sample = new TA([40, 41, 42]);
+  var calls = 0;
+  var expectedOffset = TA.BYTES_PER_ELEMENT;
```
jugglinmike: The two variables above are unused

test/built-ins/TypedArray/prototype/slice/set-values-with-bit-precision.js
```diff
@@ -0,0 +1,40 @@
...
+---*/
+
+var buffer = new ArrayBuffer(32);
+
+testWithTypedArrayConstructors(function(TA) {
+  var sample = new TA(buffer);
+  sample[0] = 0.6;
+  sample[1] = 0.6;
+  sample[2] = 0.6;
+  sample[3] = 0.6;
```
jugglinmike: I'm way out of my element here (TypedArray joke), but I don't think this test will catch violations of the quoted algorithm steps. I believe we need to use values like NaN, Infinity, and -Infinity.

@littledan: this seems like something you could help us with--do you know the best way to verify expected behavior here?

test/built-ins/TypedArray/prototype/slice/arraylength-internal.js
```diff
@@ -0,0 +1,39 @@
...
+    getCalls++;
+    return 0;
+  }
+};
+
+Object.defineProperty(TypedArray.prototype, "length", desc);
+
+testWithTypedArrayConstructors(function(TA) {
+  var sample = new TA([42, 43]);
+  getCalls = 0;
```
jugglinmike: It's not enough to re-set the counter here because the [[ArrayLength]] value is something you are trying to test.

As written, the call to `compareArray` will always return `true` because it operates as though it were given two array-likes with length `0`. For instance, you can insert this statement on line 33, and the test will continue to pass:

    result[0] = 9;

leobalter:
fixes are up, still waiting for more ideas on https://github.com/tc39/test262/pull/602#discussion_r61342578

test/built-ins/TypedArray/prototype/slice/arraylength-internal.js
leobalter: I did a manual check skipping compareArray. It's a good thing typedArrays are not sparse.

jugglinmike:
I did some research on the motivation for the "bit-level encoding" text, and I've come up with a set of tests that verify it consistently--see gh-623. If that looks good to you, then we can remove that file from this patch.

leobalter:
Yes, I'll remove that file. ++ to the tests on gh-623

leobalter:
done!

test/built-ins/TypedArray/prototype/slice/detached-buffer-get-custom-ctor.js
```diff
@@ -0,0 +1,34 @@
...
+  var sample = new TA();
+
+  sample.constructor = {};
+  sample.constructor[Symbol.species] = function(count) {
+    $DETACHBUFFER(sample.buffer);
+    return new TA(count);
+  };
+
+  assert.throws(TypeError, function() {
+    sample.slice();
```
jugglinmike: I believe this test is incorrect:

- step 3: len is 0
- step 4: relativeStart is 0
- step 5: k is 0
- step 6: relativeEnd is 0
- step 7: final is 0
- step 14.b: "k < final" is false

...meaning step 14.b.ii is never executed, so there should be no TypeError.

test/built-ins/TypedArray/prototype/slice/detached-buffer-get-ctor.js
```diff
@@ -0,0 +1,32 @@
...
+testWithTypedArrayConstructors(function(TA) {
+  var sample = new TA();
+
+  Object.defineProperty(sample, "constructor", {
+    get: function() {
+      $DETACHBUFFER(sample.buffer);
+    }
+  });
+  assert.throws(TypeError, function() {
+    sample.slice();
```
jugglinmike: I believe this test is incorrect:

- step 3: len is 0
- step 4: relativeStart is 0
- step 5: k is 0
- step 6: relativeEnd is 0
- step 7: final is 0
- step 8: count is 0
- step 15: "count > 0" is false

...meaning step 15.b is never executed, so there should be no TypeError.

jugglinmike:
Hey Leo,

I believe two of these tests are incorrect. I've left my specific rationale as an in-line comment, but in both cases, we'll need to initialize the `sample` instance to have at least one element in order to verify the behavior under test.

I'm curious: does Edge pass these tests as they are written? Regardless, I'd recommend adding corresponding test cases that verify that a TypeError is *not* thrown when slicing an array of length 0.

leobalter:
I'll fix it. Meanwhile there's another case I missed for detached buffers, following a similar code from `slice/detached-buffer-get-custom-ctor.js`:

```
  22.2.3.24 %TypedArray%.prototype.slice ( start, end )

  ...
  9. Let A be ? TypedArraySpeciesCreate(O, « count »).
  ...

  22.2.4.7 TypedArraySpeciesCreate ( exemplar, argumentList )

  ...
  3. Let constructor be ? SpeciesConstructor(exemplar, defaultConstructor).
  4. Return ? TypedArrayCreate(constructor, argumentList).

  22.2.4.6 TypedArrayCreate ( constructor, argumentList )

  1. Let newTypedArray be ? Construct(constructor, argumentList).
  2. Perform ? ValidateTypedArray(newTypedArray).
  ...
```

leobalter:
Applied the fixes. 

I'll get my windows laptop to test these methods with Chakra.

jugglinmike:
That won't be necessary for the review; I was just wondering

leobalter:
Sure, that's why I worked on the fixes first. :)

test/built-ins/TypedArray/prototype/slice/detached-buffer-zero-count-custom-ctor-same-targettype.js
```diff
@@ -0,0 +1,45 @@
...
+  sample = new TA(0);
+  sample.constructor = ctor;
+  result = sample.slice();
+  assert.sameValue(result.length, 0, "#1: result.length");
+  assert.notSameValue(result.buffer, sample.buffer, "#1: creates a new buffer");
+  assert.sameValue(result.constructor, TA, "#1: ctor");
+
+  sample = new TA(4);
+  sample.constructor = ctor;
+  result = sample.slice(1, 1);
```
jugglinmike: Nice :)

jugglinmike:
Looks good, Leo. This last set of commits introduced a couple cases of trailing whitespace. This was so objective that I figured it would be better to fix that myself, rather than interrupt your current work. Hope you don't mind!

leobalter:
of course I don't. Thanks!

jugglinmike:
I think the general problem of testing bit precision is sufficiently distinct to call this patch complete. We can continue discussion around testing that functionality in gh-623.

Merged to `master` at commit bace781a5b4c2c5e91062311cc8267be04e3ba20. Thanks, Leo!

PR #606, opened 2016-04-27
Add tests for %TypedArray%.prototype.filter

leobalter:


test/built-ins/TypedArray/prototype/filter/speciesctor-get-species-custom-ctor-length.js
```diff
@@ -0,0 +1,46 @@
...
+  sample.constructor = {};
+  sample.constructor[Symbol.species] = function() {
+    return new TA(customCount);
+  };
+
+  customCount = 2;
+  result = sample.filter(function() { return true; });
+  assert.sameValue(result.length, customCount, "length == count");
+
+  customCount = 5
```
jugglinmike: Missing a semicolon here

test/built-ins/TypedArray/prototype/filter/callbackfn-not-called-on-empty.js
```diff
@@ -0,0 +1,26 @@
...
+    ...
+    c. Let selected be ToBoolean(? Call(callbackfn, T, « kValue, k, O »)).
+  ...
+includes: [testTypedArray.js]
+---*/
+
+testWithTypedArrayConstructors(function(TA) {
+  var called = 0;
+
+  var result1 = new TA().filter(function() {
```
jugglinmike: This variable may be removed

test/built-ins/TypedArray/prototype/filter/speciesctor-get-species-custom-ctor-invocation.js
```diff
@@ -0,0 +1,61 @@
...
+    ...
+  4. Return newTypedArray.
+includes: [testTypedArray.js]
+features: [Symbol.species]
+---*/
+
+testWithTypedArrayConstructors(function(TA) {
+  var sample = new TA([40, 42, 42]);
+  var calls = 0;
+  var expectedOffset = TA.BYTES_PER_ELEMENT;
```
jugglinmike: The two variables above are unused

test/built-ins/TypedArray/prototype/filter/speciesctor-get-species-custom-ctor-returns-another-instance.js
```diff
@@ -0,0 +1,53 @@
...
+  3. If argumentList is a List of a single Number, then
+    ...
+  4. Return newTypedArray.
+includes: [testTypedArray.js, compareArray.js]
+features: [Symbol.species]
+---*/
+
+testWithTypedArrayConstructors(function(TA) {
+  var sample = new TA([40]);
+  var other = new Int8Array([1, 0, 1]);
```
jugglinmike: If I'm reading the file name/description correctly, then this test specifically concerns the behavior when the @@species constructor is different from that of the instance. If that's true, we should special-case Int8Array here

jugglinmike:
Hey Leo--I left a few suggestions for minor changes in-line. I also have some feedback that concerns multiple files:

1. There are some newline-termination inconsistencies. Could you make sure each file ends with one newline?
2. TypedArraySpeciesCreate is called iteration, which is different than `map`. This is observable even when the internal operation returns an abrupt completion. I think this could be a source of bugs because some implementors might perform part of that operation "early". What do you think about
   asserting that the callback function was invoked the correct number of times
   before each of the observable operations in TypedArraySpeciesCreate?

leobalter:
> What do you think about asserting that the callback function was invoked the correct number of times before each of the observable operations in TypedArraySpeciesCreate?

Isn't it covered at [`test/built-ins/TypedArray/prototype/filter/callbackfn-called-before-species.js`](https://github.com/tc39/test262/pull/606/files#diff-ea916de279a392f2e38228c1dd803662R1)

leobalter:
The other fixes are up.

jugglinmike:
Yes, but only partially--that test covers the "@@species" property access, but not the "constructor" property access

leobalter:
I see. I added this test for the constructor: 56f9fc4

jugglinmike:
Thanks! I think this is ready for master

jugglinmike:
Rebased, squashed, and merged to `master` at commit e6f4b2083439dc1b1917a95be30400de2639250d. Thanks, Leo!

PR #607, opened 2016-04-28
Add tests for Lexical Environment management

jugglinmike:
This patch contains a fair amount of duplication, but I didn't feel it
warranted the use of the new generation tool for two reasons:

- The repeated sets are small (1, 2, and 4 files)
- I don't expect there to be much maintenance around these semantics (unlike,
  say, destructuring binding, where spec changes will necessitate addition
  tests in the future)

That said, the duplication is still kind of noisy for reviewers. In the hopes
of making my process a little more clear (and hopefully making the review
slightly less dreadful), I've organized the repetitive tests into dedicated
commits. The first commit in this patch contains the novel tests, the other
commits contain tests that were copied for those to cover various function
forms.

I don't think we necessarily need to merge these as distinct commits, but I'm
hoping that this separation will make it easier for my reviewers to set their
expectations when reading through the files.

Commit message:

> Add tests that assert the management of the running execution context's
> LexicalEnvironment and VariableEnvironment components, as created by the
> following abstract operations:
> 
> - NewDeclarativeEnvironment
> - NewObjectEnvironment
> - NewFunctionEnvironment
> 
> Many tests require the use of non-strict direct eval, meaning they may not be
> run in strict mode. This does not effect coverage because the semantics in
> these cases are not observable from strict mode code.
> 
> Some situations require the creation of a binding, but this binding has no
> relevance to the test itself. In these cases, use names consisting solely of
> the underscore character (`_`).
> 
> Avoid the use of Block statements wherever possible, as these trigger the
> creation of additional environments which may interfere with the behavior under
> test.

### DIFF
--- a/test/language/expressions/arrow-function/scope-body-lex-distinct.js
+++ b/test/language/expressions/arrow-function/scope-body-lex-distinct.js
@@ -1,0 +1,48 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-functiondeclarationinstantiation
+description: >
+    Creation of new lexical environment (distinct from the variable
+    environment) for the function body outside of strict mode
+info: |
+    [...]
+    29. If strict is false, then
+        a. Let lexEnv be NewDeclarativeEnvironment(varEnv).
+        b. NOTE: Non-strict functions use a separate lexical Environment Record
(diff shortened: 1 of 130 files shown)


leobalter:
@goyakin would you help on this review, please? It's long enough to have someone else on the process.

leobalter:
Finally checked every single file of the new 130 tests. LGTM. Nice work.

PR #608, opened 2016-04-28
Add tests for Annex B extns to numeric literals

jugglinmike:


jugglinmike:
V8 is currently failing `test/annexB/language/literals/numeric/non-octal-decimal-integer-strict.js`

https://bugs.chromium.org/p/v8/issues/detail?id=4973

jugglinmike:
On second thought, I believe that the `-strict`-named variants belong in `test/language/literals/numeric`, as they are equally valid for contexts that do not honor Annex B. I will move them in just a moment.

PR #610, opened 2016-04-28
Add tests for Annex B extns to string literals

jugglinmike:
This change set does not include a test for restrictions relating to
template literals because such a test already exists in the project.

While a form of this test for string literals in strict mode code
existed previously, it is less precise and relies on unrelated
semantics. Remove the previous form and replace with a more direct
version.


test/language/literals/string/legacy-octal-escape-sequence-strict.js
```diff
@@ -0,0 +1,34 @@
...
+    FourToSeven :: one of
+      4 5 6 7
+
+    This definition of EscapeSequence is not used in strict mode or when
+    parsing TemplateCharacter.
+flags: [onlyStrict]
+negative: SyntaxError
+---*/
+
+'\1';
```
leobalter: This is amazing! :)

leobalter:
LGTM

avdg:
Duplicate of https://github.com/tc39/test262/blob/master/test/language/literals/string/S7.8.4_A4.3_T1.js?

PR #617, opened 2016-05-04
Add tests for %TypedArray%.prototype.set

leobalter:


test/built-ins/TypedArray/prototype/set/array-arg-return-abrupt-from-src-get-value.js
```diff
@@ -0,0 +1,47 @@
...
+  var sample = new TA([1, 2, 3, 4]);
+
+  assert.throws(Test262Error, function() {
+    sample.set(obj);
+  });
+
+  assert(
+    compareArray(sample, [42, 43, 3, 4]),
+    "values are set until exception"
+  );
```
jugglinmike: Very nice--I'm glad to see this concern covered in so many tests!

test/built-ins/TypedArray/prototype/set/array-arg-return-abrupt-from-src-length.js
```diff
@@ -0,0 +1,45 @@
...
+  length: {
+    valueOf: function() {
+      throw new Test262Error();
+    }
+  }
+};
+
+var obj2 = {
+  length: {
+    valueOf: function() {
```
jugglinmike: Based on the assertion message below, this function should be defined as the `toString` property

test/built-ins/TypedArray/prototype/set/array-arg-src-tonumber-value-type-conversions.js
```diff
@@ -0,0 +1,49 @@
...
+    return 42;
+  }
+};
+
+// undefined and NaN covered on typedArrayConversions
+var arr = ["1", "", false, true, null, obj1, obj2, [], [1]];
+
+testWithTypedArrayConstructors(function(TA) {
+  var sample = new TA(arr.length);
+  var expected = new TA(arr);
```
jugglinmike: This test is a little suspicious because it uses SetValueInBuffer to test itself. Since all the TypedArrays convert the expected values in the same way, you can avoid this by hard-coding them.

test/built-ins/TypedArray/prototype/set/array-arg-targetbuffer-detached-on-get-src-value-throws.js
```diff
@@ -0,0 +1,44 @@
...
+  });
+  Object.defineProperty(obj, 2, {
+    get: function() {
+      throw new Test262Error("Should not get other values");
+    }
+  });
+
+  assert.throws(TypeError, function() {
+    sample.set(obj);
+  });
```
jugglinmike: As written, this test correctly ensures that the TypeError is thrown after the buffer is detached and before the next element is visited. It does not ensure that the IsDetachedBuffer check comes *before* SetValue, though. If we check the elements of `sample` following the error (as you do in so many other tests), we can be sure of the correct order of operations there, as well.

test/built-ins/TypedArray/prototype/set/array-arg-targetbuffer-detached-throws.js
```diff
@@ -0,0 +1,28 @@
...
+includes: [testTypedArray.js, detachArrayBuffer.js]
+---*/
+
+testWithTypedArrayConstructors(function(TA) {
+  var sample = new TA(2);
+  $DETACHBUFFER(sample.buffer);
+
+  assert.throws(TypeError, function() {
+    sample.set([1]);
+  });
```
jugglinmike: Could you add an assertion to demonstrate that this occurs before ToObject(array)?

jugglinmike:
These two files assert that a TypeError is thrown in response to detached
buffers:

- typedarray-arg-srcbuffer-detached-throws.js
- typedarray-arg-targetbuffer-detached-throws.js

...but this patch also includes similar tests with names matching the pattern
`*-detached-during-tointeger-offset-throws.js`. The latter tests seem like more
precise versions of the former. Do the `-detached-throws` tests demonstrate
anything unique?

---

The files named `typedarray-arg-set-values-*` describe the most interesting
functionality of this method. I think we can do more to test the value
conversion that takes place. Here's an idea for step 28 (where the srcType and
targetType differ):

```js
testTypedArrayConversions((TA, value, expected, initial) => {
  if (TA === Float64Array) {
    return;
  }
  var src = new Float64Array([value]);
  var target = new TA(1);

  target.set(src);

  assert.sameValue(target[0], expected);
});
```

It's not perfect because it doesn't cover Float64Array as the target (and it
doesn't help with step 27), but I hope it gets my point across. Do you see what
I'm trying to say?

test/built-ins/TypedArray/prototype/set/array-arg-targetbuffer-detached-on-get-src-value-throws.js
leobalter: > If we check the elements of sample following the error (as you do in so many other tests), we can be sure of the correct order of operations there, as well.

I wanted to do that, but sample is detached and [I can't get](https://tc39.github.io/ecma262/#sec-integerindexedelementget) its final values.

I could check the first value was set, but I believe this is already covered in another order test.

jugglinmike: Oh, of course!

leobalter:
>  Here's an idea for step 28 (where the srcType and targetType differ)

That's step 27: _SameValue(srcType, targetType) is **false**_. It can be fixed, thou.

leobalter:
tc39/ecma262#558

leobalter:
fixes are up

jugglinmike:
Looking good!

jugglinmike:
Merged to `master` at commit 2c7c98943977a06e4d125eb287f91e7173864644. Thanks!

PR #618, opened 2016-05-04
Add tests for %TypedArray%.prototype.sort

leobalter:
Most parts are intentionally missing because they are implementation-dependent.


test/built-ins/TypedArray/prototype/sort/arraylength-internal.js
```diff
@@ -0,0 +1,38 @@
...
+  Object.defineProperty(sample, "length", desc);
+
+  var result = sample.sort();
+
+  assert.sameValue(getCalls, 0, "ignores length properties");
+  assert(
+    compareArray(result, sample),
+    "result is not affected by custom length"
+  );
+});
```
jugglinmike: This file is missing a trailing newline

test/built-ins/TypedArray/prototype/sort/comparefn-calls.js
```diff
@@ -0,0 +1,41 @@
...
+  };
+
+  sample.sort(comparefn);
+
+  assert(calls.length > 0, "calls comparefn");
+  calls.forEach(function(args) {
+    assert.sameValue(args[0], expectedThis, "comparefn is called no specific this");
+    assert.sameValue(args[1].length, 2, "comparefn is always called with 2 args");
+    assert.sameValue(args[1][0], 42, "x is a listed value");
+    assert.sameValue(args[1][0], 42, "y is a listed value");
```
jugglinmike: Interesting!

test/built-ins/TypedArray/prototype/sort/comparefn-same-return-unsort-values.js
```diff
@@ -0,0 +1,34 @@
...
+var comparefn = function() {
+  return 0;
+};
+
+testWithTypedArrayConstructors(function(TA) {
+  var arr, sample;
+
+  arr = [4, 3, 2, 1];
+  sample = new TA(arr).sort(comparefn);
+  assert(compareArray(sample, arr), "[4, 3, 2, 1]");
```
jugglinmike: We may not be able to assert this. From Array.prototype.sort:

  > The sort is not necessarily stable (that is, elements that compare equal do
  > not necessarily remain in their original order).

  and further down:

  > Unless the sort order is specified above to be implementation-defined, the
  > returned object must have the following two characteristics:
  >
  > - There must be some mathematical permutation π of the nonnegative integers
  >   less than len, such that for every nonnegative integer j less than len,
  >   if property old[j] existed, then new[π(j)] is exactly the same value as
  >   old[j].  But if property old[j] did not exist, then new[π(j)] does not
  >   exist. 
  > - Then for all nonnegative integers j and k, each less than len, if
  >   SortCompare(old[j], old[k]) < 0 (see SortCompare below), then new[π(j)] <
  >   new[π(k)].

test/built-ins/TypedArray/prototype/sort/detached-buffer-comparefn.js
```diff
@@ -0,0 +1,37 @@
...
+  2. If the argument comparefn is not undefined, then
+    a. Let v be ? Call(comparefn, undefined, « x, y »).
+    b. If IsDetachedBuffer(buffer) is true, throw a TypeError exception.
+    ...
+  ...
+includes: [testTypedArray.js, detachArrayBuffer.js]
+---*/
+
+testWithTypedArrayConstructors(function(TA) {
+  var sample = new TA(2);
```
jugglinmike: With an array length of 2, this test can't tell the difference between these two implementations:

  1. If `compareFunction` detaches the buffer, immediately return an abrupt
     completion (the correct behavior)
  2. Sort the array, and if the buffer is detached at the end of the sorting
     operation, return an abrupt completion (incorrect behavior)

`#2` seems unlikely, but if we're going to have a `calls` variable, we might as well use it!

test/built-ins/TypedArray/prototype/sort/sortcompare-with-no-tostring.js
```diff
@@ -0,0 +1,32 @@
...
+---*/
+
+var origToString = Number.prototype.toString;
+
+testWithTypedArrayConstructors(function(TA) {
+  var sample = new TA([4, 3, 2, 1]);
+
+  Number.prototype.toString = function() {
+    throw new Test262Error();
+  };
```
jugglinmike: This is a good idea, but I think it is unlikely to catch the kind of bugs
  that you have in mind. The ToString internal operation does not invoke
  Number.prototype.toString, so even if SortCompare for TypedArrays incorrectly
  used the semantics for Arrays, this test would pass.

  One way to ensure that ToString is not being called on the elements is to
  select values that would sort differently when interpreted as their string
representations, e.g.

  ```js
  var sample = new TA([20, 100, 3]);
  ```

test/built-ins/TypedArray/prototype/sort/sorted-values-nan.js
```diff
@@ -0,0 +1,31 @@
...
+  NOTE: Because NaN always compares greater than any other value, NaN property
+  values always sort to the end of the result when comparefn is not provided.
+includes: [testTypedArray.js, compareArray.js]
+---*/
+
+testWithTypedArrayConstructors(function(TA) {
+  var sample;
+
+  sample = new TA([2, NaN, NaN, 0, 1]).sort();
+  assert(compareArray(sample.slice(0, 3), [0, 1, 2]), "case #1");
```
jugglinmike: Maybe I'm being too sensitive, but I'm concerned about using different
  prototype methods to test this one because consumers may not have those
  available yet. Would you mind writing out the assertions in "long form"?

test/built-ins/TypedArray/prototype/sort/sorted-values.js
```diff
@@ -0,0 +1,63 @@
...
+  sample = new TA([0.5, 0, 1.5, 1]).sort();
+  assert(compareArray(sample, [0, 0.5, 1, 1.5]), "non integers");
+
+  sample = new TA([0.5, 0, 1.5, -0.5, -1, -1.5, 1]).sort();
+  assert(compareArray(sample, [-1.5, -1, -0.5, 0, 0.5, 1, 1.5]), "non integers + negatives");
+
+  sample = new TA([1, 0, -0, 2]).sort();
+  assert(compareArray(sample, [0, 0, 1, 2]), "0 and -0");
+  assert.sameValue(1 / sample[0], -Infinity, "-0 goes before +0");
+  assert.sameValue(1 / sample[1], Infinity, "+0 goes after -0");
```
jugglinmike: Since `+0` and `-0` compare equal, and since `sort` is not stable, I don't
think we can assert this. I think it's a great idea to test it (and make
sure that `-0` isn't accidentally converted to `+0`), but the assertion
will have to be "fuzzier".

test/built-ins/TypedArray/prototype/sort/sorted-values.js
```diff
@@ -0,0 +1,63 @@
...
+
+  sample = new TA([3, 4, Infinity, -Infinity, 1, 2]).sort();
+  assert(compareArray(sample, [-Infinity, 1, 2, 3, 4, Infinity]), "infinities");
+
+  sample = new TA([2, NaN, NaN, 0, 1]).sort();
+  assert(compareArray(sample.slice(0, 3), [0, 1, 2]), "NaN compares greater 1");
```
jugglinmike: I think these assertions may be removed since we have `sorted-values-nan.js` dedicated to this case

jugglinmike:
What a tricky method to test!

test/built-ins/TypedArray/prototype/sort/sorted-values.js
leobalter: I'm quite disappointed with the specs saying sort is not stable.

leobalter:
fixes are up

jugglinmike:
Great!

jugglinmike:
Merged to `master` at commit a0cd3b07fcd7a2936391702c162c129426b8f885. Thanks!

PR #621, opened 2016-05-04
Remove duplicate front matter keys

### LINKED ISSUE #620: Two tests fail test262-harness due to duplicated front matter keys
These two tests

test/built-ins/String/prototype/match/length.js
test/language/expressions/object/method-definition/generator-super-prop-param.js

contain duplicated keys in their front matter material currently.

This crashes test262-harness with an uncaught error.

ianwjhalliday:
Fixes #620

### DIFF
--- a/test/built-ins/String/prototype/match/length.js
+++ b/test/built-ins/String/prototype/match/length.js
@@ -2,11 +2,12 @@
 // This code is governed by the BSD license found in the LICENSE file.
 
 /*---
-info: The length property of the match method is 1
 es5id: 15.5.4.10_A11
 es6id: 21.1.3.11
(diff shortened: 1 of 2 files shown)


leobalter:
LGTM

leobalter:
I just need to confirm the [test262 CLA](http://tc39.github.io/test262-cla) is already signed. A CLA bot would be very helpful.

bterlson:
@ianwjhalliday is Microsoft and so should be good to go. Also, trivial contributions do not require CLA. Trivial is a gut-check (and better to be safe than sorry) but I think metadata updates are fine from non-member, non-CLA-signatories.

PR #623, opened 2016-05-06
Add tests for NaN handling with Typed Arrays

jugglinmike:
Given the highly technical nature of this aspect of the specification, the
chances that I have made a mistake here are higher than normal--I'm still a
little concerned that these tests are overly strict.

@littledan can you verify that these tests accurately demonstrate the
conditions required by the spec?

For context, SpiderMonkey fails:

- `built-ins/TypedArray/prototype/map/return-new-typedarray-conversion-operation-consistent-nan.js`
- `built-ins/TypedArray/prototype/slice/bit-precision.js`

while V8 fails:

- `built-ins/TypedArrays/internals/Set/conversion-operation-consistent-nan.js`

Commit message:

> Ensure that NaN values are canonicalized consistently by all invocations
> of SetValueInBuffer. Also ensure that `%TypedArray%.prototype.set` and
> `%TypedArray%.prototype.slice` preserve the bit-level encoding of the
> source data.
>
> Use a set of experimentally-derived expressions known to produce NaN
> values with distinct bit patterns in various platforms.

### DIFF
--- a/test/built-ins/TypedArray/prototype/fill/fill-values-conversion-operations-consistent-nan.js
+++ b/test/built-ins/TypedArray/prototype/fill/fill-values-conversion-operations-consistent-nan.js
@@ -63,7 +63,7 @@ function body(FloatArray) {
 
     sampleBytes = new Uint8Array(sample.buffer);
     controlBytes = new Uint8Array(control.buffer);
-    assert(compareArray(sampleBytes, controlBytes));
+    assert(compareArray(sampleBytes, controlBytes), 'NaN value #' + idx);
   }
 }
 
--- a/test/built-ins/TypedArray/prototype/fill/fill-values-conversion-operations-consistent-nan.js
+++ b/test/built-ins/TypedArray/prototype/fill/fill-values-conversion-operations-consistent-nan.js
@@ -1,0 +1,70 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-%typedarray%.prototype.fill
+es6id: 22.2.3.8
+description: Consistent canonicalization of NaN values
+info: >
+  22.2.3.8 %TypedArray%.prototype.fill (value [ , start [ , end ] ] )
+
+  %TypedArray%.prototype.fill is a distinct function that implements the same
+  algorithm as Array.prototype.fill as defined in 22.1.3.6 except that the this
+  object's [[ArrayLength]] internal slot is accessed in place of performing a
+  [[Get]] of "length". The implementation of the algorithm may be optimized with
+  the knowledge that the this value is an object that has a fixed length and
+  whose integer indexed properties are not sparse. However, such optimization
+  must not introduce any observable changes in the specified behaviour of the
+  algorithm.
+
+  ...
+
+  22.1.3.6 Array.prototype.fill (value [ , start [ , end ] ] )
+
+  ...
+  7. Repeat, while k < final
+    a. Let Pk be ! ToString(k).
+    b. Perform ? Set(O, Pk, value, true).
+  ...
+
+  24.1.1.6 SetValueInBuffer ( arrayBuffer, byteIndex, type, value [ ,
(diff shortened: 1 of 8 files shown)


test/built-ins/TypedArray/prototype/fill/fill-values-conversion-operations-consistent-nan.js
```diff
@@ -0,0 +1,70 @@
...
+
+  for (idx = 0; idx < distinctNaNs.length; ++idx) {
+    someNaN = distinctNaNs[idx];
+    control = new FloatArray([someNaN, someNaN, someNaN]);
+
+    sample.fill(someNaN);
+
+    sampleBytes = new Uint8Array(sample.buffer);
+    controlBytes = new Uint8Array(control.buffer);
+    assert(compareArray(sampleBytes, controlBytes));
```
leobalter: ìdx` can be used as the assertion message to help finding any error.

test/built-ins/TypedArray/prototype/fill/fill-values-conversion-operations-consistent-nan.js
```diff
@@ -0,0 +1,70 @@
...
+
+function body(FloatArray) {
+  var sample = new FloatArray(3);
+  var control, idx, someNaN, sampleBytes, controlBytes;
+
+  for (idx = 0; idx < distinctNaNs.length; ++idx) {
+    someNaN = distinctNaNs[idx];
+    control = new FloatArray([someNaN, someNaN, someNaN]);
+
+    sample.fill(someNaN);
```
leobalter: the distinct NaN is not necessarily preserved. On SetValueInBuffer: "rawValue may be set to any implementation chosen IEEE 754-2008 binary32 format Not-a-Number"

In GetValueInBuffer, it returns the NaN value: `If value is an IEEE 754-2008 binary32 NaN value, return the NaN Number value.` which is an [an IEEE 754-2008 “Not-a-Number” value](https://tc39.github.io/ecma262/#sec-terms-and-definitions-nan). It doesn't specify which.

The precision is missed or at least implementation dependent when you get the values from the buffer in the following assertions.

leobalter: Wow, you got it:

```
24.1.1.5 GetValueFromBuffer ( arrayBuffer, byteIndex, type [ , isLittleEndian ] )

...
11. If the first code unit of type is "U", then
  a. Let intValue be the byte elements of rawValue concatenated and interpreted as a bit string encoding of an unsigned little-endian binary number.
```

That's such an amazing test, thank you!

leobalter:
The current tests are great. My only [comment left](https://github.com/tc39/test262/pull/623#discussion_r62348197) is on the assertion messages to provide a clear info in case of failures. `compareArray` is not helpful with it.

jugglinmike:
I've extended the assertion messages as requested.

test/built-ins/TypedArray/prototype/map/return-new-typedarray-conversion-operation-consistent-nan.js
```diff
@@ -0,0 +1,60 @@
...
+  var sampleBytes, resultBytes;
+
+  var result = sample.map(function(value) {
+    return value;
+  });
+
+  sampleBytes = new Uint8Array(sample.buffer);
+  resultBytes = new Uint8Array(result.buffer);
+
+  assert(compareArray(sampleBytes, resultBytes));
```
littledan: I dunno about this test; I'm pretty sure GetValueFromBuffer is allowed to return a different NaN from what you Set into it. https://tc39.github.io/ecma262/#sec-getvaluefrombuffer says "return the NaN Number value." not "the one that was in the buffer".

At least that must have been the intent of the committee. Here, of anyplace, NaN canonicalization must be allowed.

jugglinmike: Ah, yes. I think I can still salvage this test, though. I'll have a patch ready for your review later today.

PR #624, opened 2016-05-09
Add tests for Annex B extns to global object

jugglinmike:
Extend tests for the `escape` and `unescape` built-ins to assert their
behavior. Rely on the latest ECMAScript draft, which specifies that
`escape` returns hexadecimal codes in upper case.


leobalter:
LGTM!

PR #625, opened 2016-05-09
Add tests for Annex B extn: O.p.__proto__

jugglinmike:


test/annexB/built-ins/Object/prototype/__proto__/set-non-object.js
```diff
@@ -0,0 +1,18 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-object.prototype.__proto__
+es6id: B.2.2.1
+description: Called on a value that is object-coercible but not an Object
+info: >
+    1. Let O be ? RequireObjectCoercible(this value).
+    2. If Type(proto) is neither Object nor Null, return undefined.
```
leobalter: This undefined return is from step 3

```
3. If Type(O) is not Object, return undefined.
```

test/annexB/built-ins/Object/prototype/__proto__/set-immutable.js
```diff
@@ -0,0 +1,18 @@
...
+    [...]
+    4. Let status be ? O.[[SetPrototypeOf]](proto).
+    5. If status is false, throw a TypeError exception.
+---*/
+
+Object.prototype.__proto__ = null;
+
+assert.throws(TypeError, function() {
+  Object.prototype.__proto__ = {};
+});
```
leobalter: this is so tricky

leobalter:
One comment above to fix that doesn't affect the tests behavior.

LGTM

jugglinmike:
I've extended the meta data of `set-immutable.js` as requested.

PR #626, opened 2016-05-10
Add tests for Annex B `Object.prototype` methods

jugglinmike:
This patch has a large number of files, but there is a good deal of symmetry in
the tests (just like the specification text they exercise).

### DIFF
--- a/test/annexB/built-ins/Object/prototype/__defineGetter__/define-non-extensible.js
+++ b/test/annexB/built-ins/Object/prototype/__defineGetter__/define-non-extensible.js
@@ -1,0 +1,21 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-additional-properties-of-the-object.prototype-object
+description: Behavior when "this" value is not extensible
+info: |
+    [...]
+    5. Perform ? DefinePropertyOrThrow(O, key, desc).
+---*/
+
+var noop = function() {};
+var thrower = function() {
+  throw new Test262Error();
+};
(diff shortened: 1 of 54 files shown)


leobalter:
You're 3 lines of code close to get it l33t.

jugglinmike:
Let's find three assertions that I should add :P

test/annexB/built-ins/Object/prototype/__defineGetter__/define-non-extensible.js
```diff
@@ -0,0 +1,21 @@
...
+description: Behavior when "this" value is not extensible
+info: |
+    [...]
+    5. Perform ? DefinePropertyOrThrow(O, key, desc).
+---*/
+
+var noop = function() {};
+var thrower = function() {
+  throw new Test262Error();
+};
```
leobalter: this is defined but not used

test/annexB/built-ins/Object/prototype/__defineSetter__/define-non-extensible.js
```diff
@@ -0,0 +1,21 @@
...
+description: Behavior when "this" value is not extensible
+info: |
+    [...]
+    5. Perform ? DefinePropertyOrThrow(O, key, desc).
+---*/
+
+var noop = function() {};
+var thrower = function() {
+  throw new Test262Error();
+};
```
leobalter: defined but not used as well

leobalter:
Instead of finding 3 extra lines we are probably getting rid of 6.

jugglinmike:
I somehow missed this; sorry for the delay. We'll get 1337 next time

PR #627, opened 2016-05-10
Extend tests for Annex B String HTML methods

jugglinmike:
- Add tests for abrupt completion handling from ToString operations
- Add tests for method property descriptors
- Remove redundant assertion


leobalter:
LGTM

> Remove redundant assertion

Thanks for that as well. I wish we could have available time to reformat all the legacy test files, but this small cleanup is a good step forward.

PR #630, opened 2016-05-10
Add tests for DataView.prototype.getFloat32

leobalter:
These tests will be similar to on the other `DataView#get*` methods. I'm opening this to reuse any feedback on the following tests (new PRs).

I'm not so happy with the tests returning a NaN. Have you got any ideas to improve it, @jugglinmike?

### DIFF
--- a/test/built-ins/DataView/prototype/getFloat32/detached-buffer-after-integer-byteoffset.js
+++ b/test/built-ins/DataView/prototype/getFloat32/detached-buffer-after-integer-byteoffset.js
@@ -1,0 +1,38 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-dataview.prototype.getfloat32
+es6id: 24.2.4.5
+description: >
+  Detached buffer is checked after checking If numberIndex ≠ getIndex or
+  getIndex < 0,
+info: |
+  24.2.4.5 DataView.prototype.getFloat32 ( byteOffset [ , littleEndian ] )
+
+  1. Let v be the this value.
+  2. If littleEndian is not present, let littleEndian be false.
+  3. Return ? GetViewValue(v, byteOffset, littleEndian, "Float32").
+
+  24.2.1.1 GetViewValue ( view, requestIndex, isLittleEndian, type )
+
+  ...
+  6. If numberIndex ≠ getIndex or getIndex < 0, throw a RangeError exception.
+  ...
+  8. Let buffer be the value of view's [[ViewedArrayBuffer]] internal slot.
+  9. If IsDetachedBuffer(buffer) is true, throw a TypeError exception.
+  ...
+includes: [detachArrayBuffer.js]
+---*/
+
+var buffer = new ArrayBuffer(1);
+var sample = new DataView(buffer, 0);
+
+$DETACHBUFFER(buffer);
+assert.throws(RangeError, function() {
+  sample.getFloat32(1.1);
+});
+
+assert.throws(RangeError, function() {
+  sample.getFloat32(-1);
+});
--- a/test/built-ins/DataView/prototype/getFloat32/detached-buffer-after-integer-byteoffset.js
+++ b/test/built-ins/DataView/prototype/getFloat32/detached-buffer-after-integer-byteoffset.js
@@ -25,7 +25,7 @@ info: |
 includes: [detachArrayBuffer.js]
 ---*/
 
-var buffer = new ArrayBuffer(1);
+var buffer = new ArrayBuffer(6);
 var sample = new DataView(buffer, 0);
 
 $DETACHBUFFER(buffer);
--- a/test/built-ins/DataView/prototype/getFloat32/detached-buffer.js
+++ b/test/built-ins/DataView/prototype/getFloat32/detached-buffer.js
@@ -1,0 +1,31 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-dataview.prototype.getfloat32
+es6id: 24.2.4.5
+description: >
+  Throws a TypeError if buffer is detached
+info: |
+  24.2.4.5 DataView.prototype.getFloat32 ( byteOffset [ , littleEndian ] )
+
+  1. Let v be the this value.
+  2. If littleEndian is not present, let littleEndian be false.
+  3. Return ? GetViewValue(v, byteOffset, littleEndian, "Float32").
+
+  24.2.1.1 GetViewValue ( view, requestIndex, isLittleEndian, type )
+
+  ...
+  8. Let buffer be the value of view's [[ViewedArrayBuffer]] internal slot.
+  9. If IsDetachedBuffer(buffer) is true, throw a TypeError exception.
+  ...
+includes: [detachArrayBuffer.js]
+---*/
+
+var buffer = new ArrayBuffer(1);
+var sample = new DataView(buffer, 0);
+
+$DETACHBUFFER(buffer);
+assert.throws(TypeError, function() {
+  sample.getFloat32(0);
+});
--- a/test/built-ins/DataView/prototype/getFloat32/return-nan.js
+++ b/test/built-ins/DataView/prototype/getFloat32/return-nan.js
@@ -26,21 +26,20 @@ info: |
   ...
   8. If isLittleEndian is false, reverse the order of the elements of rawValue.
   ...
-features: [DataView.prototype.setFloat32]
+features: [DataView.prototype.setUint8]
 ---*/
 
-var buffer = new ArrayBuffer(12);
+var buffer = new ArrayBuffer(8);
 var sample = new DataView(buffer, 0);
 
-sample.setFloat32(0, 39);
-sample.setFloat32(4, NaN);
-sample.setFloat32(8, 42);
+sample.setUint8(0, 127);
+sample.setUint8(1, 192);
+sample.setUint8(2, 0);
+sample.setUint8(3, 0);
+sample.setUint8(4, 255);
+sample.setUint8(5, 192);
+sample.setUint8(6, 0);
+sample.setUint8(7, 0);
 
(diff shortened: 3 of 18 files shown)


test/built-ins/DataView/prototype/getFloat32/return-nan.js
```diff
@@ -0,0 +1,43 @@
...
+  the corresponding step in the SetValueInBuffer abstract operation.
+  8. If isLittleEndian is false, reverse the order of the elements of rawValue.
+  ...
+features: [DataView.prototype.setFloat32]
+---*/
+
+var buffer = new ArrayBuffer(4);
+var sample = new DataView(buffer, 0);
+
+sample.setFloat32(0, NaN);
```
jugglinmike: You could improve this test by increasing the size of the buffer and using `setUint8` to set a few different NaN bit patterns.

test/built-ins/DataView/prototype/getFloat32/return-values-custom-offset.js
```diff
@@ -0,0 +1,60 @@
...
+assert.sameValue(sample.getFloat32(0, false), 13323083, "0, false");
+assert.sameValue(sample.getFloat32(1, false), 13323084, "1, false");
+assert.sameValue(sample.getFloat32(2, false), 13323340, "2, false");
+assert.sameValue(sample.getFloat32(3, false), 13388877, "3, false");
+assert.sameValue(sample.getFloat32(4, false), 53556532, "4, false");
+assert.sameValue(sample.getFloat32(0, true), 13323083, "0, true");
+assert.sameValue(sample.getFloat32(1, true), 53292332, "1, true");
+assert.sameValue(sample.getFloat32(2, true), 53554476, "2, true");
+assert.sameValue(sample.getFloat32(3, true), 214222000, "3, true");
+assert.sameValue(sample.getFloat32(4, true), 215270592, "4, true");
```
jugglinmike: This test file seems to be expressing the same expectation as `return-values.js`. Am I missing a detail that makes the two files distinct?

jugglinmike:
Step 7 of GetValueFromBuffer cannot be observed by calling this method because
isLittleEndian is always set to some value in setFloat32. Since it is
irrelevant, we should remove it from the test meta data.

Some tests use `setFloat32` while others use `setUint8`. I think we should be
consistent unless there is some variable condition that influences the
decision. `setUint8` seems more precise to me, so maybe that's the best option.

Many tests have duplicate assertions for the value of the `littleEndian`
parameter. Since that value only effects a single step of the algorithm under
test, I don't think the redundancy is warranted.

leobalter:
> Some tests use setFloat32 while others use setUint8. I think we should be
consistent unless there is some variable condition that influences the
decision. 

The cases I use `setFloat32` are meant to set Float32 `Infinity` and `NaN` values. 

> setUint8 seems more precise to me, so maybe that's the best option.

`setUint8` is used anywhere else, I'm reusing it on the tests for the other methods.

> Many tests have duplicate assertions for the value of the littleEndian
parameter. Since that value only effects a single step of the algorithm under
test, I don't think the redundancy is warranted.

I have some duplicates on the ToBolean, I'll do a search and possibly a cleanup

test/built-ins/DataView/prototype/getFloat32/return-values-custom-offset.js
leobalter: It uses the offset position (see line 49).

The littleEndian argument also helps telling how `rawBytes` are organized.

test/built-ins/DataView/prototype/getFloat32/return-nan.js
leobalter: I added more tests to check the NaN positioned in the middle of a buffer.

I also want to add a few tests to check the same IEEE NaN is stored, but that will come for `setFloat32`. The following is a preview:

```js
sample.setFloat(0, 0/0);
sample.setFloat(4, Infinity/Infinity);
sample.setFloat(8, -(0,0));
sample.setFloat(12, Math.pow(-1, 0.5));
sample.setFloat(16, -Math.pow(-1, 0.5));

var uint8s = [];
var i;

for ( i = 0; i < 20; i++ ) {
  uint8s.push(sample.getUint8(i));
}

// check the 4 bytes are the same on every NaN
```

leobalter:
fixes are up!

test/built-ins/DataView/prototype/getFloat32/return-values-custom-offset.js
jugglinmike: > It uses the offset position (see line 49).

Got it!

> The littleEndian argument also helps telling how rawBytes are organized.

Doesn't the Float32 value itself tell you exactly how rawBytes are organized?

jugglinmike:
> I added more tests to check the NaN positioned in the middle of a buffer.

I don't think I explained myself very well. Here is the text we want to verify:

> If value is an IEEE 754-2008 binary32 NaN value, return the NaN Number value.

While this test certainly demonstrates that, moving the position of the NaN
value does not really improve the test.

The important (though tiny) word here is "an". The significance is that there
are many encodings for NaN, and if `getFloat32` encounters *any* of them, it
should return the JavaScript value NaN.

So when I suggested "using setUint8 to set a few different NaN bit patterns", I
meant something like:

```js
var buffer = new ArrayBuffer(8);
var sample = new DataView(buffer, 0);

sample.setUint8(0, 127);
sample.setUint8(1, 192);
sample.setUint8(2, 0);
sample.setUint8(3, 0);
sample.setUint8(4, 255);
sample.setUint8(5, 192);
sample.setUint8(6, 0);
sample.setUint8(7, 0);

assert.sameValue(sample.getFloat32(0), NaN);
assert.sameValue(sample.getFloat32(4), NaN);
```

Does that seem right to you?

---

> I also want to add a few tests to check the same IEEE NaN is stored, but that
> will come for setFloat32. The following is a preview:

Actually, that strategy is not valid. The text for SetValueInBuffer reads:

> If value is NaN, rawValue may be set to any implementation chosen IEEE
> 754-2008 binary32 format Not-a-Number encoding. An implementation must always
> choose the same encoding for each implementation distinguishable NaN value.

Again, the distinction here is tiny: "the same encoding for **each** [...]
value," not "the same encoding for **every** [...] value." That means that
different NaN values may have bit patterns, but the decision has to be
consistent. So in the sample code you provided, it would be perfectly
acceptable if each of the 4-byte sets in the `uint8s` array were different.

---

> The cases I use setFloat32 are meant to set Float32 Infinity and NaN values. 

I noticed that, but I still don't understand the reason for the change.
Couldn't you set those values with `setUint8` as well?

test/built-ins/DataView/prototype/getFloat32/return-values-custom-offset.js
leobalter: > Doesn't the Float32 value itself tell you exactly how rawBytes are organized?

I set the values on the first sample, without the offset.

leobalter:
```js
sample.setUint8(0, 127);
sample.setUint8(1, 192);
sample.setUint8(2, 0);
sample.setUint8(3, 0);
```

As we talked on a private chat, I can't find a source to tell this group of Uint8 represents a NaN.

I don't feel comfortable using these numbers arbitrarily, but at least we guarantee a non-redundant way to test `DataView#setFloat32(n, NaN)` in another test.

leobalter:
patches are up for NaN and Infinity

leobalter:
got some extra tests for -0

test/built-ins/DataView/prototype/getFloat32/detached-buffer-after-integer-byteoffset.js
```diff
@@ -0,0 +1,38 @@
...
+  ...
+  6. If numberIndex ≠ getIndex or getIndex < 0, throw a RangeError exception.
+  ...
+  8. Let buffer be the value of view's [[ViewedArrayBuffer]] internal slot.
+  9. If IsDetachedBuffer(buffer) is true, throw a TypeError exception.
+  ...
+includes: [detachArrayBuffer.js]
+---*/
+
+var buffer = new ArrayBuffer(1);
```
jugglinmike: This ArrayBuffer should be initialized with at least 4 bytes to ensure that the RangeError is due to the buffer being detached and not the element size exceeding the buffer boundary.

test/built-ins/DataView/prototype/getFloat32/detached-buffer.js
```diff
@@ -0,0 +1,31 @@
...
+  24.2.1.1 GetViewValue ( view, requestIndex, isLittleEndian, type )
+
+  ...
+  8. Let buffer be the value of view's [[ViewedArrayBuffer]] internal slot.
+  9. If IsDetachedBuffer(buffer) is true, throw a TypeError exception.
+  ...
+includes: [detachArrayBuffer.js]
+---*/
+
+var buffer = new ArrayBuffer(1);
```
jugglinmike: The same goes for this test

test/built-ins/DataView/prototype/getFloat32/detached-buffer-after-integer-byteoffset.js
jugglinmike: I misunderstood; a detached buffer would trigger a TypeError, so there is no danger of false positive there. Although it may be good to use a valid buffer size in order to differentiate between the RangeErrors.

test/built-ins/DataView/prototype/getFloat32/detached-buffer.js
jugglinmike: Scratch that. This is asserting the TypeError, so we're good here

test/built-ins/DataView/prototype/getFloat32/detached-buffer-after-integer-byteoffset.js
leobalter: I moved to 6 bytes as 1.1 may be cast to 1 or 2, so we assert the RangeError is not due to the buffer size.

jugglinmike: Roger roger

jugglinmike:
Merged to `master` at commit 47b323cf8986f0084df2b2451a0bf6aa844f5a8a. Thanks, Leo!

PR #631, opened 2016-05-11
Add tests for Annex B extns to Date.prototype

jugglinmike:
I've removed a couple of surface tests for `getYear` and `setYear` since their
weak assertions are covered by the new set of tests. I've also removed a few
tests for `toGMTString`, preferring a simple test for object equality with
`toUTCString`.


leobalter:
LGTM

PR #633, opened 2016-05-11
Add tests for Annex B extn: __proto__ in obj init

jugglinmike:


test/language/expressions/object/__proto__-permitted-dup.js
```diff
@@ -0,0 +1,34 @@
...
+  ['__proto__']: null,
+  __proto__() {},
+  get __proto__() { return 33; },
+  set __proto__(_) { return 44; }
+};
+
+var desc = Object.getOwnPropertyDescriptor(obj, '__proto__');
+
+assert.sameValue(desc.get(), 33);
+assert.sameValue(desc.set(), 44);
```
leobalter: this test is good, but (╯°□°）╯︵ ┻━┻ after this `obj`.

leobalter:
LGTM

PR #634, opened 2016-05-11
Dataview.prototype.get* methods

leobalter:
The feedback from #630 was also applied on this branch.

This and #630 covers the get* methods.

### DIFF
--- a/test/built-ins/DataView/prototype/getFloat64/minus-zero.js
+++ b/test/built-ins/DataView/prototype/getFloat64/minus-zero.js
@@ -1,0 +1,46 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-dataview.prototype.getfloat64
+es6id: 24.2.4.6
+description: >
+  Return -0
+info: |
+  24.2.4.6 DataView.prototype.getFloat64 ( byteOffset [ , littleEndian ] )
+
+  1. Let v be the this value.
+  2. If littleEndian is not present, let littleEndian be false.
+  3. Return ? GetViewValue(v, byteOffset, littleEndian, "Float64").
+
+  24.2.1.1 GetViewValue ( view, requestIndex, isLittleEndian, type )
+
+  ...
+  14. Let bufferIndex be getIndex + viewOffset.
+  15. Return GetValueFromBuffer(buffer, bufferIndex, type, isLittleEndian).
+  ...
+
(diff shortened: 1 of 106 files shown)


test/built-ins/DataView/prototype/getFloat64/return-nan.js
```diff
@@ -0,0 +1,73 @@
...
+
+sample.setUint8(0, 127);
+sample.setUint8(1, 251);
+sample.setUint8(2, 0);
+sample.setUint8(3, 0);
+sample.setUint8(4, 0);
+sample.setUint8(5, 0);
+sample.setUint8(6, 0);
+sample.setUint8(7, 0);
+assert.sameValue(sample.getFloat64(0), NaN, "127, 251, 0, ...");
```
jugglinmike: Very nice

test/built-ins/DataView/prototype/getFloat64/minus-zero.js
```diff
@@ -0,0 +1,46 @@
...
+sample.setUint8(2, 0);
+sample.setUint8(3, 0);
+sample.setUint8(4, 0);
+sample.setUint8(5, 0);
+sample.setUint8(6, 0);
+sample.setUint8(7, 0);
+
+var result = sample.getFloat64(0);
+assert.sameValue(result, 0);
+assert.sameValue(1 / result, -Infinity);
```
jugglinmike: Just like the operation in the spec (and as you realized in gh-630),
SameValue accounts for negative zero. As written, the first assertion is
incorrect; we'll need to compare `result` to `-0` instead.

jugglinmike:
For the following pairs of files, the chosen values  do not demonstrate the distinction between Int16 and Uint16:

- {getUint16,getInt16}/return-values-custom-offset.js
- {getUint32,getInt32}/return-values-custom-offset.js

Could you select byte values that differ between those two views?

leobalter:
> Could you select byte values that differ between those two views?

Great idea. I just did it, thanks!

jugglinmike:
Well in that case, I believe our work here is done.

leobalter:
yay!

jugglinmike:
Merged to `master` at the following commits:

- 9565f01 Add tests for DataView.prototype.getUint8
- 1567b68 Add tests for Dataview.prototype.getInt8
- 5e767dc Add tests for DataView.prototype.getUint16
- 673b63e Add tests for DataView.prototype.getInt16
- 7f8e559 Add tests for DataView.prototype.getUint32
- 0d5b6ea Add tests for DataView.prototype.getInt32
- c1cdf88 Add tests for DataView.prototype.getFloat64
- 47b323c Add tests for DataView.prototype.getFloat32

PR #635, opened 2016-05-12
Add tests for DataView.prototype.setFloat32

leobalter:


test/built-ins/DataView/prototype/setFloat32/set-nans.js
```diff
@@ -0,0 +1,83 @@
...
+    endian order. If value is NaN, rawValue may be set to any implementation
+    chosen IEEE 754-2008 binary32 format Not-a-Number encoding. An
+    implementation must always choose the same encoding for each implementation
+    distinguishable NaN value.
+  ...
+  11. Store the individual bytes of rawBytes into block, in order, starting at
+  block[byteIndex].
+  12. Return NormalCompletion(undefined).
+features: [DataView.prototype.getUint8]
+includes: [nans.js]
```
leobalter: @jugglinmike these tests are breaking everywhere, even on es6draft, the best part is: in different parts.

test/built-ins/DataView/prototype/setFloat32/set-values-return-undefined.js
```diff
@@ -0,0 +1,161 @@
...
+  block[byteIndex].
+  12. Return NormalCompletion(undefined).
+features: [DataView.prototype.getFloat32]
+---*/
+
+var buffer = new ArrayBuffer(4);
+var sample = new DataView(buffer, 0);
+
+var values = [
+  127,         // 2 ** 7 - 1
```
jugglinmike: There is a lot of duplication between this file and `harness/testTypedArray.js`. If we follow this pattern for the remaining DataView methods, this duplication will be compounded. Have you given any thought to re-factoring the harness function so that the data can be re-used?

test/built-ins/DataView/prototype/setFloat32/to-boolean-littleendian.js
```diff
@@ -0,0 +1,58 @@
...
+sample.setFloat32(0, 3, null);
+assert.sameValue(sample.getFloat32(0), 3, "null");
+sample.setFloat32(0, 4, 0);
+assert.sameValue(sample.getFloat32(0), 4, "0");
+sample.setFloat32(0, 5, "");
+assert.sameValue(sample.getFloat32(0), 5, "the empty string");
+
+// True
+sample.setFloat32(0, 6, {});
+assert.sameValue(sample.getFloat32(0, true), 6, "{}");
```
jugglinmike: This test unnecessarily depends on additional semantics of `getFloat32`. It would be simpler (and more literal) if it did not use `getFloat32`'s second parameter.

leobalter: changing it here does not solve a similar - but bigger - problem in `set-values-return-undefined.js` where it depends on the getFloat32's second parameter as well.

These set tests depend a lot on their respective `DataView#get` methods, I don't like this but it was the best approach I found. At least the get methods does not depend on any set methods rather than `setUint8`.

I'll change it in this file as suggested, but I don't thing it helps considering the other test file I mentioned above.

test/built-ins/DataView/prototype/setFloat32/set-values-return-undefined.js
leobalter: Yes, I thought about it. I ended up with this repetitive approach because I didn't want to include the typedArrays harness for DataView tests. So that would require a split the harness files into at least:

- `byteConversionValues.js`
- `testTypedArrays.js`

The tests for conversions on TypedArrays would need to include both, otherwise `testTypedArrayConversions` would fail, regardless the file I move it, as it needs the values and the `testWithTypedArrayConstructors` method.

I don't like this harness subdependency. My other suggestion would be having this byteConversionValues.js file for being used only on DataView tests, not changing the typedArrays harness, but that's a similar repetition.

I'll open another PR for this.

leobalter:
fixes are up.

jugglinmike:
Looks good to me!

leobalter:
🙌  thanks

jugglinmike:
Squashed and merged to `master` at da67e6b3056d56b8bc024d98583fd97274f1f644. Thanks!

PR #636, opened 2016-05-13
Add tests for generator intrinsic default proto

jugglinmike:


leobalter:
LGTM, I just wonder if we should add an extra one for generator methods.

PR #640, opened 2016-05-17
Correct test for NaN canonicalization

jugglinmike:
Because implementations are free to select any valid NaN value during
GetValueFromBuffer, tests concerning semantics for consistent NaN value
encoding cannot rely on values returned from that abstract operation.

Update the test for `%TypedArray%.prototype.map` to set the same NaN
values via `map` as set in the "control" array.

This problem was [originally reported](https://github.com/tc39/test262/pull/623#discussion_r63443319) by @littledan. Dan: what do you think of this approach?


tcare:
Looks good.

littledan:
LGTM

PR #641, opened 2016-05-18
Add tests for TypedArrays#includes

leobalter:


test/built-ins/TypedArray/prototype/includes/fromIndex-equal-or-greater-length-returns-false.js
```diff
@@ -0,0 +1,35 @@
...
+  8. Return false.
+includes: [testTypedArray.js]
+---*/
+
+testWithTypedArrayConstructors(function(TA) {
+  var sample;
+
+  sample = new TA(42);
+  assert.sameValue(sample.includes(0, 42), false);
+  assert.sameValue(sample.includes(0, 43), false);
```
jugglinmike: This test would be more interesting if the array contained the searchElement (just like in `fromIndex-infinity.js`)

leobalter: it contains the element, all of them are 0s.

jugglinmike: Arg, you are correct. I got confused by the second parameter, somehow.

jugglinmike:
Nice work, Leo!

test/built-ins/TypedArray/prototype/includes/fromIndex-equal-or-greater-length-returns-false.js
leobalter: I revisited this file several times because of the second parameter as well, but they are still good assertions, I believe.

leobalter:
Thanks :)

jugglinmike:
Squashed and merged to `master` at commit 56b988883e3e1819cdc98b21c125be4c80f2bc24. Thanks, Leo! :dragon_face:

PR #642, opened 2016-05-18
Add tests for Intl.getCanonicalLocales

zbraniecki:
This ports a bunch of tests we wrote for SpiderMonkey implementation into test262 suite - https://dxr.mozilla.org/mozilla-central/source/js/src/tests/Intl

You can see background story of why those tests have been chosen here: https://bugzilla.mozilla.org/show_bug.cgi?id=1263040

### DIFF
--- a/test/intl402/Intl/getCanonicalLocales/duplicates.js
+++ b/test/intl402/Intl/getCanonicalLocales/duplicates.js
@@ -1,0 +1,19 @@
+// Copyright 2016 Mozilla Corporation. All rights reserved.
+// This code is governed by the license found in the LICENSE file.
+
+/*---
+esid: sec-intl.getcanonicallocales
+description: Tests the getCanonicalLocales function for duplicate locales scenario.
+info: |
+  8.2.1 Intl.getCanonicalLocales (locales)
+  1. Let ll be ? CanonicalizeLocaleList(locales).
+  2. Return CreateArrayFromList(ll).
+includes: [compareArray.js]
+---*/
+
+assert(compareArray(
+  Intl.getCanonicalLocales(
+    ['ab-cd', 'ff', 'de-rt', 'ab-Cd']), ['ab-CD', 'ff', 'de-RT']));
+
+var locales = Intl.getCanonicalLocales(["en-US", "en-US"]);
+assert(compareArray(locales, ['en-US']), 'en-US');
--- a/test/intl402/Intl/getCanonicalLocales/has-property.js
+++ b/test/intl402/Intl/getCanonicalLocales/has-property.js
@@ -1,0 +1,30 @@
+// Copyright 2016 Mozilla Corporation. All rights reserved.
+// This code is governed by the license found in the LICENSE file.
+
+/*---
+esid: sec-intl.getcanonicallocales
+description: Test Intl.getCanonicalLocales.name for step 7.b. 
+info: |
+  9.2.1 CanonicalizeLocaleList (locales)
+    7. Repeat, while k < len.
+      b. Let kPresent be HasProperty(O, Pk).
+features: [Proxy]
+---*/
+
+var locales = {
+  '0': 'en-US',
+  '1': 'pt-BR',
+  length: 2
+};
+
+var p = new Proxy(locales, {
+  has: function(_, prop) {
+    if (prop === '0') {
+      throw new Test262Error();
+    }
+  }
+});
+
+assert.throws(Test262Error, function() {
+  Intl.getCanonicalLocales(p);
+});
--- a/test/intl402/Intl/getCanonicalLocales/locales-is-not-a-string.js
+++ b/test/intl402/Intl/getCanonicalLocales/locales-is-not-a-string.js
@@ -1,0 +1,30 @@
+// Copyright 2016 Mozilla Corporation. All rights reserved.
+// This code is governed by the license found in the LICENSE file.
+
+/*---
+esid: sec-intl.getcanonicallocales
+description: Tests for scenario where locales is not a string
+info: |
+  8.2.1 Intl.getCanonicalLocales (locales)
+  1. Let ll be ? CanonicalizeLocaleList(locales).
+  2. Return CreateArrayFromList(ll).
+includes: [compareArray.js]
+features: [Symbol]
+---*/
+
+var gCL = Intl.getCanonicalLocales;
+
+function assertArray(l, r) {
+  assert(compareArray(l, r), r);
+}
+
+assertArray(gCL(), []);
+assertArray(gCL(false), []);
+assertArray(gCL(true), []);
+assertArray(gCL(Symbol("foo")), []);
+assertArray(gCL(NaN), []);
+assertArray(gCL(1), []);
+
+Number.prototype[0] = "en-US";
+Number.prototype.length = 1;
+assertArray(gCL(NaN), ["en-US"]);
--- a/test/intl402/Intl/getCanonicalLocales/main.js
+++ b/test/intl402/Intl/getCanonicalLocales/main.js
@@ -1,0 +1,34 @@
+// Copyright 2016 Mozilla Corporation. All rights reserved.
+// This code is governed by the license found in the LICENSE file.
+
+/*---
+esid: sec-intl.getcanonicallocales
+description: Tests for existance and behavior of Intl.getCanonicalLocales
+info: |
+  8.2.1 Intl.getCanonicalLocales (locales)
+  1. Let ll be ? CanonicalizeLocaleList(locales).
+  2. Return CreateArrayFromList(ll).
+includes: [compareArray.js]
+---*/
+
+var gCL = Intl.getCanonicalLocales;
+
+function assertArray(l, r) {
+  assert(compareArray(l, r), r);
+}
+
+assertArray(gCL(), []);
+
+assertArray(gCL('ab-cd'), ['ab-CD']);
+
+assertArray(gCL(['ab-cd']), ['ab-CD']);
+
+assertArray(gCL(['ab-cd', 'FF']), ['ab-CD', 'ff']);
+
+assertArray(gCL({'a': 0}), []);
+
+assertArray(gCL({}), []);
+
+assertArray(gCL(['ar-ma-u-ca-islamicc']), ['ar-MA-u-ca-islamicc']);
+
+assertArray(gCL(['th-th-u-nu-thai']), ['th-TH-u-nu-thai']);
--- a/test/intl402/Intl/getCanonicalLocales/name.js
+++ b/test/intl402/Intl/getCanonicalLocales/name.js
@@ -1,0 +1,20 @@
+// Copyright 2016 Mozilla Corporation. All rights reserved.
+// This code is governed by the license found in the LICENSE file.
+
+/*---
+esid: sec-intl.getcanonicallocales
+description: Intl.getCanonicalLocales.name value and descriptor. 
+info: |
+  8.2.1 Intl.getCanonicalLocales (locales)
+  1. Let ll be ? CanonicalizeLocaleList(locales).
+  2. Return CreateArrayFromList(ll).
+includes: [propertyHelper.js]
+---*/
+
+assert.sameValue(Intl.getCanonicalLocales.name, 'getCanonicalLocales',
+  'The value of `Intl.getCanonicalLocales.name` is `"getCanonicalLocales"`'
+);
+
+verifyNotEnumerable(Intl.getCanonicalLocales, 'name');
+verifyNotWritable(Intl.getCanonicalLocales, 'name');
+verifyConfigurable(Intl.getCanonicalLocales, 'name');
--- a/test/intl402/Intl/getCanonicalLocales/overriden-arg-length.js
+++ b/test/intl402/Intl/getCanonicalLocales/overriden-arg-length.js
@@ -1,0 +1,89 @@
+// Copyright 2016 Mozilla Corporation. All rights reserved.
+// This code is governed by the license found in the LICENSE file.
+
+/*---
+esid: sec-intl.getcanonicallocales
+description: Test Intl.getCanonicalLocales for step 5. 
+info: |
+  9.2.1 CanonicalizeLocaleList (locales)
+    5. Let len be ? ToLength(? Get(O, "length")).
+includes: [compareArray.js]
+features: [Symbol]
+---*/
+
+var locales = {
+  '0': 'en-US',
+};
+
+Object.defineProperty(locales, "length", {
+  get: function() { throw new Test262Error() }
+});
+
+assert.throws(Test262Error, function() {
+  Intl.getCanonicalLocales(locales);
+}, "should throw if locales.length throws");
+
+var locales = {
+  '0': 'en-US',
+  '1': 'pt-BR',
+};
+
+Object.defineProperty(locales, "length", {
+  get: function() { return "1" }
+});
+
+assert(compareArray(Intl.getCanonicalLocales(locales), ['en-US']),
+  "should return one element if locales.length is '1'");
+
+var locales = {
+  '0': 'en-US',
+  '1': 'pt-BR',
+};
+
+Object.defineProperty(locales, "length", {
+  get: function() { return 1.3 }
+});
+
+assert(compareArray(Intl.getCanonicalLocales(locales), ['en-US']),
+  "should return one element if locales.length is 1.3");
+
+var locales = {
+  '0': 'en-US',
+  '1': 'pt-BR',
+};
+
+Object.defineProperty(locales, "length", {
+  get: function() { return Symbol("1.8") }
+});
+
+assert.throws(TypeError, function() {
+  Intl.getCanonicalLocales(locales);
+}, "should throw if locales.length is a Symbol");
+
+var locales = {
+  '0': 'en-US',
+  '1': 'pt-BR',
+};
+
+Object.defineProperty(locales, "length", {
+  get: function() { return -Infinity }
+});
+
+assert(compareArray(Intl.getCanonicalLocales(locales), []),
+  "should return empty array if locales.length is -Infinity");
+
+var locales = {
+  length: -Math.pow(2, 32) + 1
+};
+
+Object.defineProperty(locales, "0", {
+  get: function() { throw new Error("must not be gotten!"); }
+})
+
+assert(compareArray(Intl.getCanonicalLocales(locales), []),
+  "should return empty array if locales.length is a negative value");
+
+var count = 0;
+var locs = { get length() { if (count++ > 0) throw 42; return 0; } };
+var locales = Intl.getCanonicalLocales(locs); // shouldn't throw 42
+assert.sameValue(locales.length, 0);
--- a/test/intl402/Intl/getCanonicalLocales/to-string.js
+++ b/test/intl402/Intl/getCanonicalLocales/to-string.js
@@ -1,0 +1,20 @@
+// Copyright 2016 Mozilla Corporation. All rights reserved.
+// This code is governed by the license found in the LICENSE file.
+
+/*---
+esid: sec-intl.getcanonicallocales
+description: Test Intl.getCanonicalLocales.name for step 7.c.iii 
+info: |
+  9.2.1 CanonicalizeLocaleList (locales)
+    7. Repeat, while k < len.
+      c. If kPresent is true, then
+        iii. Let tag be ? ToString(kValue).
+includes: [compareArray.js]
+---*/
+
+var locales = {
+  '0': { toString: function() { locales[1] = 'pt-BR'; return 'en-US'; }},
+  length: 2
+};
+
+assert(compareArray(Intl.getCanonicalLocales(locales), [ "en-US", "pt-BR" ]));
--- a/test/intl402/Intl/getCanonicalLocales/weird-cases.js
+++ b/test/intl402/Intl/getCanonicalLocales/weird-cases.js
@@ -1,0 +1,25 @@
+// Copyright 2016 Mozilla Corporation. All rights reserved.
+// This code is governed by the license found in the LICENSE file.
+
+/*---
+esid: sec-intl.getcanonicallocales
+description: Tests the getCanonicalLocales function for weird tags.
+info: |
+  8.2.1 Intl.getCanonicalLocales (locales)
+  1. Let ll be ? CanonicalizeLocaleList(locales).
+  2. Return CreateArrayFromList(ll).
+includes: [compareArray.js]
+---*/
+
+var weirdCases =
+  [
+   "x-u-foo",
+   "en-x-u-foo",
+   "en-a-bar-x-u-foo",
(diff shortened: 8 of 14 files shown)


zbraniecki:
I'm so excited. It'll be the first new test in Intl in 2 years! :D

caridy:
LGTM

zbraniecki:
@jugglinmike is that ready for merge?

leobalter:
While the tests are technically correct, they file format is wrong.

The first step would be splitting each group of assertions in different files which can be placed in a new `test/built-ins/intl402/Intl/getCanonicalLocales/` folder.

If we can use a file name to identify each test case is better. The id as a file name is an old format for previous tests.

I believe we can also use `harness/compareArray.js` on tests using `assertShallowArray`.

I know this might give some extra work, but this format purpose is to provide a better report with specific purpose files, regarding a failed assertion prevents the remaining assertions to run in the same file, and it's usual to find partial implementations or specific regressions on any runtime.

zbraniecki:
@leobalter - does it look better now?

test/intl402/Intl/getCanonicalLocales/main.js
```diff
@@ -0,0 +1,48 @@
...
+    $ERROR("Intl.getCanonicalLocales doesn't exist.");
+}
+
+if (Intl.getCanonicalLocales.length !== 1) {
+    $ERROR("Intl.getCanonicalLocales.length should be 1.");
+}
+
+let gCL = Intl.getCanonicalLocales;
+
+assertEq(Intl.getCanonicalLocales.length, 1);
```
leobalter: `assertEq` does not exist. You can use `assert.sameValue( value, expected [ , message ] );`

test/intl402/Intl/getCanonicalLocales/overriden-arg-length.js
```diff
@@ -0,0 +1,17 @@
...
+  8.2.1 Intl.getCanonicalLocales (locales)
+  1. Let ll be ? CanonicalizeLocaleList(locales).
+  2. Return CreateArrayFromList(ll).
+includes: [compareArray.js]
+---*/
+
+var count = 0;
+var locs = { get length() { if (count++ > 0) throw 42; return 0; } };
+var locales = Intl.getCanonicalLocales(locs); // shouldn't throw 42
+assertEq(locales.length, 0);
```
leobalter: here's another occurrence of `assertEq`. You can use `assert.sameValue( value, expected [ , message ] );`

test/intl402/Intl/getCanonicalLocales/duplicates.js
```diff
@@ -0,0 +1,19 @@
...
+  2. Return CreateArrayFromList(ll).
+includes: [compareArray.js]
+---*/
+
+assert(compareArray(
+  Intl.getCanonicalLocales(
+    ['ab-cd', 'ff', 'de-rt', 'ab-Cd']), ['ab-CD', 'ff', 'de-RT']));
+
+var locales = Intl.getCanonicalLocales(["en-US", "en-US"]);
+assert(compareArray(locales, ['en-US']));
```
leobalter: When I suggested using compareArray I forgot to mention it does not make clear when it fails. In cases we have 2 or more assertions in the same file, is good to provide an assertion message so the implementor knows where it's failing. This can be made as:

`assert(compareArray(locales, ['en-US']), 'en-US');`

This way makes it easy to map what is failing.

test/intl402/Intl/getCanonicalLocales/main.js
```diff
@@ -0,0 +1,48 @@
...
+includes: [compareArray.js]
+---*/
+
+if (typeof Intl.getCanonicalLocales !== "function") {
+    $ERROR("Intl.getCanonicalLocales doesn't exist.");
+}
+
+if (Intl.getCanonicalLocales.length !== 1) {
+    $ERROR("Intl.getCanonicalLocales.length should be 1.");
+}
```
leobalter: we have a non-explicit pattern for existence that can be split in 3 test files:

Those files would be:

- property type (function) and its descriptor. [example](https://github.com/tc39/test262/blob/master/test/built-ins/Array/prototype/find/find.js)
- length's value, and descriptor. [example](https://github.com/tc39/test262/blob/master/test/built-ins/Array/prototype/find/length.js)
- name's value and descriptor. [example](https://github.com/tc39/test262/blob/master/test/built-ins/Array/prototype/find/name.js)

test/intl402/Intl/getCanonicalLocales/overriden-set.js
```diff
@@ -0,0 +1,19 @@
...
+  2. Return CreateArrayFromList(ll).
+includes: [compareArray.js]
+---*/
+
+Object.defineProperty(Array.prototype, 0, { set() { throw 42; } });
+
+// must not throw 42, might if push is used
+var arr = Intl.getCanonicalLocales(["en-US"]);
+
+assert(compareArray(arr, ["en-US"]));
```
leobalter: Rather than checking for access of Array.prototype[0], a test to check the returned object is an Array instance with the expected _own_ properties provide a more reliable check.

test/intl402/Intl/getCanonicalLocales/weird-cases.js
```diff
@@ -0,0 +1,24 @@
...
+---*/
+
+var weirdCases =
+  [
+   "x-u-foo",
+   "en-x-u-foo",
+   "en-a-bar-x-u-foo",
+   "en-x-u-foo-a-bar",
+   "en-a-bar-u-baz-x-u-foo",
+  ];
```
leobalter: can we approximate these cases to something that looks close to locales but are not?

Examples: `en-us-`, `-en-us`, `enus`, `en-us-en-us`.

I believe these examples act as negative cases for `IsStructurallyValidLanguageTag`, but please correct me if I'm wrong.

test/intl402/Intl/getCanonicalLocales/main.js
```diff
@@ -0,0 +1,48 @@
...
+
+assert(compareArray(gCL({'a': 0}), []));
+
+assert(compareArray(gCL({}), []));
+
+assert(compareArray(gCL(['ar-ma-u-ca-islamicc']), ['ar-MA-u-ca-islamicc']));
+
+assert(compareArray(gCL(['th-th-u-nu-thai']), ['th-TH-u-nu-thai']));
+
+assert(compareArray(gCL(NaN), []));
```
leobalter: this assertion is an important which deserves it's home, or just a test file. 

We can have a specific test file for ToObject operations if _locales_ is not a string

test/intl402/Intl/getCanonicalLocales/duplicates.js
zbraniecki: would it make sense to macro it into:
```
function assertArray(l, r) {
  assert(compareArray(l, r), r);
}
```
?

test/intl402/Intl/getCanonicalLocales/overriden-set.js
zbraniecki: how should a check like this look like?

test/intl402/Intl/getCanonicalLocales/weird-cases.js
zbraniecki: so, that would test for when canonicalization changes the result, right? Like, testing for `-en-us` will throw a `RangeError`.
Do you want a separate test for that?

leobalter:
It's looking much better! Now I was able to left some feedback on specific cases and found some missing coverage:

### 9.2.1 CanonicalizeLocaleList (locales)
- [ ] step 1: return an empty Array if _locales_ is _undefined_.
- [ ] step 4.a: _locales_ is coerced to an if it's not an string
  - [ ] throws a TypeError for _null_ values
  - [ ] returns an empty Array instance if _locales_ is `false`, `true`, or a Symbol. (Numbers and NaN are already covered in the main.js file)
- [ ] step 5:
  - [ ] check the `Get(O, "length")` operation, you may use an Array-like object patching a get accessor on the length property and returning a custom number.
  - [ ] return custom abrupt completion from this length property (the same get accessor can throw a `new Test262Error()`
  - [ ] values are coerced to an integer indexed length value on the ToLength (strings become numbers, float numbers become integers)
  - [ ] throws a TypeError if length value is a Symbol
  - [ ] length boundaries: -Infinity, negative values, +Infinity, values over 2**53-1, -0 becomes +0
  - Note: the PR 643 has some tests with these operations
- [ ] step 7.b:
  - [ ] abrupt completions from HasProperty(O, Pk) using array-like objects on locales (a Proxy helps a lot here)
- [ ] step 7.c.i:
  - [ ] abrupt completions from Get(O, Pk) (easy to catch throwing a Test262Error from a property accessor)
  - [ ] This is a tricky one: _locales_ size and values are not cached, so getting a property can modify the values for the next indexes.
- [ ] step 7.c.ii: If Type(kValue) is not String or Object, throw a TypeError exception. Numbers, including NaN (important to avoid it being cast as a String), a Symbol, and Boolean values
- [ ] step 7.c.iii:
  - [ ] abrupt completion from ToString(kValue)
  - [ ] ToString operations if kValue is an object (it can change the _locales_ size and values as well)
- [ ] step 7.c.iv: RangeError cases

It looks longer than it is, but that should do the coverage.

test/intl402/Intl/getCanonicalLocales/weird-cases.js
leobalter: yes! thanks! a separate test file is better.

test/intl402/Intl/getCanonicalLocales/overriden-set.js
leobalter:
```js
var locales = ["en-US"];
var result = Intl.getCanonicalLocales(["en-US"]);

assert.notSameValue(result, locales, "result is a new array instance");
assert.sameValue(result.length, 1, "result.length");
assert(result.hasOwnProperty("0"), "result an own property '0'");
assert.sameValue(result[0], "en-US", "result[0]");
```

test/intl402/Intl/getCanonicalLocales/duplicates.js
leobalter: sure thing!

zbraniecki:
Thanks a lot @leobalter ! It's my first test262 patch, but with the amount of new formatters coming in for 4th ed, I expect to have a lot of tests to write, so it's good to learn :)

I updated the tests to cover most of the things you listed, but I do have some troubles with the things from step 7 of your list. I'm not sure how to test `HasProperty(O, Pk)` using Proxy, this thing: `locales size and values are not cached, so getting a property can modify the values for the next indexes.` and `ToString operations if kValue is an object (it can change the locales size and values as well)`.

Can you help me with some mock code or pointers? Thanks!

test/intl402/Intl/getCanonicalLocales/locales-is-not-a-string.js
```diff
@@ -0,0 +1,29 @@
...
+assertArray(gCL(), []);
+assertArray(gCL(false), []);
+assertArray(gCL(true), []);
+assertArray(gCL(Symbol), []);
+assertArray(gCL(NaN), []);
+assertArray(gCL(1), []);
+
+Number.prototype[0] = "en-US";
+Number.prototype.length = 1;
+assertArray(gCL(NaN), ["en-US"]);
```
leobalter: nice one!

leobalter:
> I'm not sure how to test HasProperty(O, Pk) using Proxy

My first idea is around this:

```js
var locales = {
  '0': 'en-US',
  '1': 'pt-BR',
  length: 2
};

var p = new Proxy(locales, {
  has: function(_, prop) {
    if  (prop === '0') {
      throw new Test262Error();
    }
  }
});

  Intl.getCanonicalLocales(p);
});
```

Reviewing this PR reminded me I missed the "values are not cached" while testing `Array#includes`, so I made something similar there: https://github.com/tc39/test262/pull/643/commits/9e86f589452ef99d62cb0b1fc27612d0dec71086

for the ToString operations on values:

```js
var locales = {
  '0': { toString: function() { locales[1] = 'pt-BR'; return 'en-US'; },
  length: 2
};

... compareArray(Intl.getCanonicalLocales(locales), [ "en-US", "pt-BR" ])
```

zbraniecki:
@leobalter - thanks so much! I think I covered everything you put in your review. Can you look again?

test/intl402/Intl/getCanonicalLocales/to-string.js
```diff
@@ -0,0 +1,20 @@
...
+// This code is governed by the license found in the LICENSE file.
+
+/*---
+esid: sec-intl.getcanonicallocales
+description: Test Intl.getCanonicalLocales.name for step 7.c.iii 
+info: |
+  9.2.1 CanonicalizeLocaleList (locales)
+    7. Repeat, while k < len.
+      b. If kPresent is true, then
+        iii. Let tag be ? ToString(kValue).
```
leobalter: this is item 7.c.iii, right? just a typo, I believe

zbraniecki: right, fixed!

test/intl402/Intl/getCanonicalLocales/locales-is-not-a-string.js
jugglinmike: Agreed

test/intl402/Intl/getCanonicalLocales/has-property.js
```diff
@@ -0,0 +1,29 @@
...
+
+var locales = {
+  '0': 'en-US',
+  '1': 'pt-BR',
+  length: 2
+};
+
+var p = new Proxy(locales, {
+  has: function(_, prop) {
+    if  (prop === '0') {
```
jugglinmike: Superfluous whitespace here

test/intl402/Intl/getCanonicalLocales/overriden-set.js
```diff
@@ -0,0 +1,20 @@
+// Copyright 2016 Mozilla Corporation. All rights reserved.
+// This code is governed by the license found in the LICENSE file.
+
+/*---
+esid: sec-intl.getcanonicallocales
+description: Tests the getCanonicalLocales function for overridden set().
```
jugglinmike: This test's description and name do not match its contents.

jugglinmike:
Hi @zbraniecki! Leo asked me to take a look at your patch, and (besides a few minor comments shared inline), I have two requests:

We generally try to avoid using features from ES2015 because, as the most recent standard, it is possible that they are not available in the consumer's runtime. I think this is especially important for the Intl API because we would like these tests to be usable by shims like [intl.js](https://github.com/andyearnshaw/Intl.js), and those projects specifically target "legacy" runtimes. Could you remove the `for..of` loops, arrow functions, and `let` bindings?

This is not always avoidable, of course. Some tests require new features by their very nature (nice work on those ones, by the way). In those cases, we add [the `features` tag](https://github.com/tc39/test262/blob/30d4c04182a9c2be88a8bf3dc4a3fd54b44c5ebf/CONTRIBUTING.md#features) to help consumers skip tests in an automated way.

Could you add a test for the `length` property of the method? Here's an example [test/built-ins/TypedArray/of/length.js](https://github.com/tc39/test262/blob/master/test/built-ins/TypedArray/of/length.js).

`length.js` would be a good filename (since it fits with the `name.js` test you've already written, and the similar tests throughout this project). I realize that this patch already has a file with that name, but (if I'm reading it correctly), that test would fit in the file named `overriddent-arg-length.js`.

test/intl402/Intl/getCanonicalLocales/overriden-set.js
zbraniecki: fixed

test/intl402/Intl/getCanonicalLocales/has-property.js
zbraniecki: fixed

zbraniecki:
Thanks @jugglinmike !
I moved `for..of`, `let` and arrow functions. Not sure where do you need the featuers tags and which ones should I use (for example, there's one expotential operator, but I'm not sure what's the feature name for it).

I added the length.js.

jugglinmike:
We're primarily interested in annotating tests that rely on "new" language
features, where "new" is ES2015 (for now, anyway). For this patch, that means
adding tags to the following files:

- locales-is-not-a-string.js: `features: [Symbol]`
- overriden-arg-length.js: `features: [Symbol]`
- error-cases.js: `features: [Symbol]`
- has-property.js: `features: [Proxy]`

As for the exponentiation operator, it would be best to use `Math.pow(2, 53)` so older runtimes could benefit from your work there.

zbraniecki:
Applied your feedback! Thanks @jugglinmike :)

test/intl402/Intl/getCanonicalLocales/overriden-arg-length.js
```diff
@@ -0,0 +1,115 @@
...
+  '0': 'en-US',
+  '1': 'pt-BR',
+};
+
+Object.defineProperty(locales, "length", {
+  get: function() { return Math.pow(2, 53); }
+});
+
+assert(compareArray(Intl.getCanonicalLocales(locales), []),
+  "should return empty array if locales.length is over 2**53-1");
```
jugglinmike: I can't find the justification for this assertion. Can you show me which step interprets the length value in this way?

test/intl402/Intl/getCanonicalLocales/name.js
```diff
@@ -0,0 +1,21 @@
...
+description: Intl.getCanonicalLocales.name value and descriptor. 
+info: |
+  8.2.1 Intl.getCanonicalLocales (locales)
+  1. Let ll be ? CanonicalizeLocaleList(locales).
+  2. Return CreateArrayFromList(ll).
+includes: [propertyHelper.js]
+---*/
+
+assert.sameValue(
+  typeof Intl.getCanonicalLocales.name, 'getCanonicalLocales',
```
jugglinmike: We need to assert the property value (not the `typeof` the value)

test/intl402/Intl/getCanonicalLocales/name.js
```diff
@@ -0,0 +1,21 @@
...
+  'The value of `Intl.getCanonicalLocales.name` is `"getCanonicalLocales"`'
+);
+
+verifyNotEnumerable(Intl.getCanonicalLocales, 'name');
+verifyWritable(Intl.getCanonicalLocales, 'name');
```
jugglinmike: From [ECMA262 section 17](https://tc39.github.io/ecma262/#sec-ecmascript-standard-built-in-objects):

> Unless otherwise specified, the name property of a built-in Function object, if it exists, has the attributes { [[Writable]]: false, [[Enumerable]]: false, [[Configurable]]: true }.

test/intl402/Intl/getCanonicalLocales/overriden-arg-length.js
```diff
@@ -0,0 +1,115 @@
...
+  '0': 'en-US',
+  '1': 'pt-BR',
+};
+
+Object.defineProperty(locales, "length", {
+  get: function() { return +Infinity }
+});
+
+assert(compareArray(Intl.getCanonicalLocales(locales), []),
+  "should return empty array if locales.length is +Infinity");
```
jugglinmike: The same goes for this assertion. It seems as though the length should be interpreted as the finite value `2**53-1` in this case.

jugglinmike:
Hi again, Zibi,

I've found a few more details that need to be fixed--see above. We're almost there!

test/intl402/Intl/getCanonicalLocales/overriden-arg-length.js
zbraniecki: In a comment by @leobalter - https://github.com/tc39/test262/pull/642#issuecomment-220478113 he mentions:

```
step 5:

    check the Get(O, "length") operation, you may use an Array-like object patching a get accessor on the length property and returning a custom number.
    return custom abrupt completion from this length property (the same get accessor can throw a new Test262Error()
    values are coerced to an integer indexed length value on the ToLength (strings become numbers, float numbers become integers)
    throws a TypeError if length value is a Symbol
    length boundaries: -Infinity, negative values, +Infinity, values over 2**53-1, -0 becomes +0
```

test/intl402/Intl/getCanonicalLocales/overriden-arg-length.js
zbraniecki: You're right. How should I test it?

test/intl402/Intl/getCanonicalLocales/overriden-arg-length.js
zbraniecki: Oh, it should return `2**51-1` if the length is over. How to test it?

jugglinmike: Since `k` is always initialized to `0`, and since the value of `len` is not directly observable, I don't think there is a practical way to test that boundary condition.

If @leobalter can verify, then we can resolve this by just removing the assertions for those conditions.

zbraniecki:
Ok, I think I've done everything you asked for. Waiting for @leobalter confirmation that I can remove the two tests for over-limit length.

test/intl402/Intl/getCanonicalLocales/overriden-arg-length.js
leobalter: This assertion wrong. I believe it's reasonable enought to assume `len` will be `(2 ** 53 - 1)` after step 5: "_Let len be ? ToLength(? Get(O, "length"))._". 

It properly gets the length from the locales object and it iterates on step 7 "_Repeat, while k < len_". This loop is safe with step 7.b and it should get values from locales 0 and 1.

Maybe I badly formed my sentence where:

> length boundaries: -Infinity, negative values, +Infinity, values over 2**53-1, -0 becomes +0

should be read as:

> length boundaries: -Infinity; negative values; +Infinity; values over 2**53-1; (-0 becomes +0)

leobalter:
answered inline here https://github.com/tc39/test262/pull/642#discussion_r65621970

I believe these tests can be fixed by basically expecting `["en-US", "pt-BR"]` on both.

zbraniecki:
@leobalter yeah, but the tests takes forever if length is `Math.pow(2, 53) - 1` before it returns the correct value. It seems like it would make testing really slow :(

leobalter:
The hasProperty check is usually not slow on runtimes using this custom
object. That slowness would real if using an array that long.

zbraniecki:
I'm testing against SpiderMonkey with `Intl.getCanonicalLocales`. Up until `Math.pow(2, 20)` it works well, around `Math.pow(2, 30)` it starts taking a couple seconds. With `Math.pow(2,50)` it takes forever...

leobalter:
Ok, that's a fact now. Let's remove them.

zbraniecki:
Removed! ready to land? :)

jugglinmike:
Merged to `master` at commit aeb88953f0303d26a47886f9d60c23d00cacea9e. Thanks!

caridy:
awesome, great work guys!

PR #643, opened 2016-05-18
Add tests for Array.prototype.includes

leobalter:


leobalter:
I missed tests for non-number values of length and out of range values as well. I'll add them.

jugglinmike:
Alright; I'll hold off on reviewing until you're ready

leobalter:
It's done. But I've got [this change](1857b6c) I didn't like to avoid breaking runtimes around 2**53. Without a fromIndex I got issues with V8, SM, es6draft, etc.

test/built-ins/Array/prototype/includes/fromIndex-equal-or-greater-length-returns-false.js
```diff
@@ -0,0 +1,23 @@
...
+  produces the value 0.)
+  5. If n ≥ 0, then
+    a. Let k be n.
+  ...
+  7. Repeat, while k < len
+    ...
+  8. Return false.
+---*/
+
+var sample = new Array(42).map(function() { return 7; });
```
jugglinmike: This complexity doesn't really add much value, and it's a little inconsistent with the way you seed arrays in other tests. Would you be satisfied with something like `var sample = [7, 7, 7, 7];`?

test/built-ins/Array/prototype/includes/length-boundaries.js
```diff
@@ -0,0 +1,82 @@
...
+  [].includes.call(obj, "d", fromIndex),
+  false,
+  "2**53-1, not found value at 2**53-1"
+);
+
+obj.length = 9007199254740992;
+assert.sameValue(
+  [].includes.call(obj, "d", fromIndex),
+  true,
+  "2**53, found value at 2**53-1"
```
jugglinmike: I think there's a mistake in this test. `includes` will never report a positive match for an element at index `2**53 - 1` ("d" in this case) because [that is the maximum length value](https://tc39.github.io/ecma262/#sec-tolength), and the algorithm stops iterating when it reaches it ("Repeat, while k < len").

test/built-ins/Array/prototype/includes/length-zero-returns-false.js
```diff
@@ -0,0 +1,28 @@
...
+    throw new Test262Error();
+  }
+};
+
+var sample = [];
+assert.sameValue(sample.includes(0), false, "returns false");
+assert.sameValue(sample.includes(), false, "returns false - no arg");
+assert.sameValue(
+  sample.includes(0, fromIndex), false,
+  "length is checked before ToInteger(fromIndex)"
```
jugglinmike: This message will never be displayed to the consumer. I'd recommend either initializing the `Test262Error` with a description string, or (better still) getting rid of the `Test262Error` altogether and instead asserting the value of an invocation counter

test/built-ins/Array/prototype/includes/search-not-found-returns-false.js
```diff
@@ -0,0 +1,40 @@
...
+    a. Let k be n.
+  6. Else n < 0,
+    a. Let k be len + n.
+    b. If k < 0, let k be 0.
+  7. Repeat, while k < len
+    a. Let elementK be the result of ? Get(O, ! ToString(k)).
+    b. If SameValueZero(searchElement, elementK) is true, return true.
+    c. Increase k by 1.
+  8. Return false.
+---*/
```
jugglinmike: The frontmatter for this test should declare "features: [Symbol]"

test/built-ins/Array/prototype/includes/search-not-found-returns-false.js
```diff
@@ -0,0 +1,40 @@
...
+
+assert.sameValue([undefined, false, 0, 1].includes(null), false, "null");
+assert.sameValue([null].includes(undefined), false, "undefined");
+
+assert.sameValue([Symbol("1")].includes(Symbol("1")), false, "symbol");
+assert.sameValue([{}].includes({}), false, "object");
+assert.sameValue([[]].includes([]), false, "array");
+
+var sample = [42];
+assert.sameValue(sample.includes(sample), false, "this");
```
jugglinmike: Cool

test/built-ins/Array/prototype/includes/return-abrupt-tonumber-length.js
```diff
@@ -0,0 +1,37 @@
...
+var obj2 = {
+  length: {
+    toString: function() {
+      throw new Test262Error();
+    }
+  }
+};
+
+assert.throws(Test262Error, function() {
+  [].includes.call(obj1, 7);
```
jugglinmike: Here and in a few other places, an argument for the searchElement parameter isn't necessary

jugglinmike:
Just a few suggestions for this one Leo. We're almost done with arrays!

leobalter:
One extra test I missed: values are not cached

test/built-ins/Array/prototype/includes/length-boundaries.js
leobalter: yes, I stepped on this one

leobalter: off-by-one errors

leobalter:
fixes are up including the missing test

test/built-ins/Array/prototype/includes/values-are-not-cached.js
```diff
@@ -0,0 +1,52 @@
...
+
+var obj;
+obj = getCleanObj();
+assert.sameValue([].includes.call(obj, "tc39"), true, "'tc39' is true");
+
+obj = getCleanObj();
+assert.sameValue([].includes.call(obj, "ecma262"), true, "'ecma262' is true");
+
+obj = getCleanObj();
+assert.sameValue([].includes.call(obj, "cake"), false, "'cake' is false");
```
leobalter: the cake is a lie

test/built-ins/Array/prototype/includes/values-are-not-cached.js
```diff
@@ -0,0 +1,52 @@
...
+      return 2;
+    }
+  });
+
+  return obj;
+}
+
+var obj;
+obj = getCleanObj();
+assert.sameValue([].includes.call(obj, "tc39"), true, "'tc39' is true");
```
jugglinmike: This strategy is a little unorthodox, but I can roll with it. To really take advantage of your setup, though, you should `assert.sameValue(obj[1], undefined)` after this call to `includes`. This ensures that iteration halts once a match is found.

leobalter: The iteration halt on match is covered here: https://github.com/tc39/test262/pull/643/files#diff-e26af1e445b843aadb5c20d15bf12f39R41

leobalter: I added a fixup commit to simplify this file: 8373b77

jugglinmike: Right you are. One day goes by and I forget about everything I looked at before. Glad to see that simplification

test/built-ins/Array/prototype/includes/values-are-not-cached.js
```diff
@@ -20,16 +20,8 @@ function getCleanObj() {
...
-              return "ecma262";
-            }
-          });
-          Object.defineProperty(obj, "2", {
-            get: function() {
-              return "cake";
-            }
-          });
+          obj[1] = "ecma262";
+          obj[2] = "tc39";
```
jugglinmike: As nice as the simplification is, this change invalidated the test. The final assertion in this file expects `obj[2]` to have the value `"cake"`, but (after this change) it actually takes the value `"tc39"`.

jugglinmike:
The newest test is now invalid (my in-line comment is above but rendering as collapsed for some reason). If we address that problem, this patch should be ready to land.

leobalter:
oops. It looked as such an easy change I didn't test, I should never guess something is just a piece of cake.

jugglinmike:
I was just checking this patch for coverage parity with gh-95 and found that this version is missing a test for [the extension of `Array.prototype[@@unscopables]`](https://tc39.github.io/ecma262/#sec-array.prototype-@@unscopables). Could you add a test for that? Can you also check for additional tests that we may have missed in this version?

This makes me wonder (a little late, now that gh-641 has landed) if it would be better to use gh-95 as a "base" and work from there. Probably you considered this prior to starting. Can you say a bit about why you preferred a fresh start?

leobalter:
>  I was just checking this patch for coverage parity with gh-95 and found that this version is missing a test for the extension of Array.prototype[@@unscopables]

oops, I missed that.

> Could you add a test for that? Can you also check for additional tests that we may have missed in this version?

yes, yes

> This makes me wonder (a little late, now that gh-641 has landed) if it would be better to use gh-95 as a "base" and work from there. Probably you considered this prior to starting. Can you say a bit about why you preferred a fresh start?

I used it as a quick reference only, but I preferred using the other tests for typedArray methods (mostly indexOf) as the base for the tests on gh-641, and then I reused them for this one. I've done a coverage check in a piece of paper, as I've been doing to check coverage on other tests to find missing parts for the steps of Array#includes and quickly compared with what we have on gh-95, that's where I missed unscopables.

leobalter:
the test for unscopables is up

jugglinmike:
Sounds good. Merged to `master` at commit c7f9b12cc4daf26c2b82461f4c403c81952e4402. Thanks, Leo!

PR #644, opened 2016-05-19
Document host-defined realm/script API

jugglinmike:
@bterlson @domenic @littledan @caridy @dherman Between my fumbling with ECMA262
references, my (mis)use of [eshost's current API](https://github.com/bterlson/eshost#runtime-library), and my negligence of [the Realms proposal](https://github.com/caridy/proposal-realms), I expect I have stepped on *all* of your toes, collectively.  I'm submitting this naive draft to serve as a concrete basis for discussion.

Background: I am currently writing tests for cross-realm semantics, and either @leobalter or I will soon be writing tests for cross-script semantics as well. I'd like to define an API to enable this, and my goals are (1) ease of implementation (to encourage consumers to actually run those new tests), and (2) harmony with the Realms proposal (to limit friction when/if that is standardized).

From tips on improving the presentation (do you think we need full-blown ECMArkup for this?), to improving the language, to fixing the fundamental API design, I'm really interested in what you all have to say!

Commit message:

> Define the expected behavior of new host-defined utilities. These will
> facilitate forthcoming tests that concern cross-realm and cross-script
> semantics (which are currently untestable using standard ECMAScript
> alone).

### DIFF
--- a/INTERPRETING.md
+++ b/INTERPRETING.md
@@ -36,15 +36,15 @@ properties of the global scope prior to test execution.
   tests (via the `async` flag, described below).
 - **`$`** An ordinary object with the following properties:
   - **`createRealm`** - a function which creates a new [ECMAScript
-    Realm](http://www.ecma-international.org/ecma-262/6.0/#sec-code-realms),
+    Realm](https://tc39.github.io/ecma262/2016/#sec-code-realms),
     defines this API on the new realm's global `this` value, and returns the
     `$` property of the new realm's global `this` value
   - **`detachArrayBuffer`** - a function which implements [the
     DetachArrayBuffer abstract
-    operation](http://www.ecma-international.org/ecma-262/6.0/#sec-detacharraybuffer)
+    operation](https://tc39.github.io/ecma262/2016/#sec-detacharraybuffer)
   - **`evalScript`** - a function which accepts a string value as its first
     argument and performs
-    [ScriptEvaluation](http://www.ecma-international.org/ecma-262/6.0/#sec-runtime-semantics-scriptevaluation)
+    [ScriptEvaluation](https://tc39.github.io/ecma262/2016/#sec-runtime-semantics-scriptevaluation)
     on that value, using the current Realm as the argument.
   - **`global`** - a reference to the global `this` value on which `$` was
     initially defined
--- a/INTERPRETING.md
+++ b/INTERPRETING.md
@@ -43,9 +43,21 @@ properties of the global scope prior to test execution.
     DetachArrayBuffer abstract
     operation](https://tc39.github.io/ecma262/2016/#sec-detacharraybuffer)
   - **`evalScript`** - a function which accepts a string value as its first
-    argument and performs
-    [ScriptEvaluation](https://tc39.github.io/ecma262/2016/#sec-runtime-semantics-scriptevaluation)
-    on that value, using the current Realm as the argument.
+    argument and executes is as [an ECMAScript
+    script](https://tc39.github.io/ecma262/2016/#sec-scripts) according to the
+    following algorithm:
+
+        1. Let hostDefined be any host-defined values for the provided
+           sourceText (obtained in an implementation dependent manner)
+        2. Let realm be the current Realm Record.
+        3. Let s be ParseScript(sourceText, realm, hostDefined).
+        4. If s is a List of errors, then
+           a. Let error be the first element of s.
+           b. Return
+              Completion{[[Type]]: throw, [[Value]]: error, [[Target]]: empty}.
+        5. Let status be ScriptEvaluation(s).
+        6. Return Completion(status).
+
   - **`global`** - a reference to the global `this` value on which `$` was
     initially defined
 
--- a/INTERPRETING.md
+++ b/INTERPRETING.md
@@ -34,6 +34,20 @@ properties of the global scope prior to test execution.
 - **`print`** A function that exposes the string value of its first argument to
   the test runner. This is used as a communication mechanism for asynchronous
   tests (via the `async` flag, described below).
+- **`$`** An ordinary object with the following properties:
+  - **`createRealm`** - a function which creates a new [ECMAScript
+    Realm](http://www.ecma-international.org/ecma-262/6.0/#sec-code-realms),
+    defines this API on the new realm's global `this` value, and returns the
+    `$` property of the new realm's global `this` value
+  - **`detachArrayBuffer`** - a function which implements [the
+    DetachArrayBuffer abstract
+    operation](http://www.ecma-international.org/ecma-262/6.0/#sec-detacharraybuffer)
+  - **`evalScript`** - a function which accepts a string value as its first
+    argument and performs
+    [ScriptEvaluation](http://www.ecma-international.org/ecma-262/6.0/#sec-runtime-semantics-scriptevaluation)
+    on that value, using the current Realm as the argument.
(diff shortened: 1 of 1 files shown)


INTERPRETING.md
```diff
@@ -34,6 +34,20 @@ properties of the global scope prior to test execution.
 - **`print`** A function that exposes the string value of its first argument to
   the test runner. This is used as a communication mechanism for asynchronous
   tests (via the `async` flag, described below).
+- **`$`** An ordinary object with the following properties:
+  - **`createRealm`** - a function which creates a new [ECMAScript
+    Realm](http://www.ecma-international.org/ecma-262/6.0/#sec-code-realms),
```
domenic: Wrong spec, use https://tc39.github.io/ecma262/

INTERPRETING.md
```diff
@@ -34,6 +34,20 @@ properties of the global scope prior to test execution.
...
+    defines this API on the new realm's global `this` value, and returns the
+    `$` property of the new realm's global `this` value
+  - **`detachArrayBuffer`** - a function which implements [the
+    DetachArrayBuffer abstract
+    operation](http://www.ecma-international.org/ecma-262/6.0/#sec-detacharraybuffer)
+  - **`evalScript`** - a function which accepts a string value as its first
+    argument and performs
+    [ScriptEvaluation](http://www.ecma-international.org/ecma-262/6.0/#sec-runtime-semantics-scriptevaluation)
+    on that value, using the current Realm as the argument.
+  - **`global`** - a reference to the global `this` value on which `$` was
```
domenic: global and global this value are different; better to be explicit.

INTERPRETING.md
```diff
@@ -34,6 +34,20 @@ properties of the global scope prior to test execution.
...
   tests (via the `async` flag, described below).
+- **`$`** An ordinary object with the following properties:
+  - **`createRealm`** - a function which creates a new [ECMAScript
+    Realm](http://www.ecma-international.org/ecma-262/6.0/#sec-code-realms),
+    defines this API on the new realm's global `this` value, and returns the
+    `$` property of the new realm's global `this` value
+  - **`detachArrayBuffer`** - a function which implements [the
+    DetachArrayBuffer abstract
+    operation](http://www.ecma-international.org/ecma-262/6.0/#sec-detacharraybuffer)
+  - **`evalScript`** - a function which accepts a string value as its first
```
domenic: If this uses the current realm as the argument, then you can just use `eval`, no?

domenic: Well, actually, I'm not sure. But in general it's weird you expect this to use the current realm while `$.global` uses "the realm in which `$` was defined".

jugglinmike: There are some details (e.g. redefinition of global lexical bindings) that
can't be simulated via `eval`, so I think there's value in requiring a
dedicated API for script evaluation.

As for the active realm record: I actually intended for the same realm to be
used in both cases: `otherRealm.$` should be the global "this" value of
`otherRealm`, and `otherRealm.evalScript` should evaluate the provided code in
`otherRealm`.

In writing, "the current Realm", I was hoping to avoid the awkwardness of "the
realm in which `$` was defined", instead relying on [the formal definition of
"the current realm"](https://tc39.github.io/ecma262/#current-realm) and the
execution context created by invoking the `evalScript` function. But maybe my
definitions aren't rigorous enough to rely on those definitions--the intended
value of the function's [[Realm]] slot likely isn't clear at all.

Do you have any suggestions on how I could make this more clear?

domenic: Well, you have to pick which semantics you want. Consider the code running in realm B that does `realmA$.global` or `realmA$.evalScript`. Do you want that to get the global of realm A, or realm B? Do you want it to eval script in realm A, or in realm B?

jugglinmike: Here's what I have in mind:

```js
assert.sameValue($.global, this);
assert.sameValue($.evalScript('this'), this);

var realmB$ = $.createRealm();
assert.sameValue(realmB$.global, realmB$.global.eval('this'));
assert.sameValue(realmB$.evalScript('this'), realmB$.global.eval('this'));

realmB$.global.realmA$ = $;
assert.sameValue(realmB$.evalScript('realmA$.global'), this);
assert.sameValue(realmB$.evalScript('realmA$.evalScript("this")'), this);
```

jugglinmike:
I've re-written `$.evalScript` to be a little more formal. Hopefully this
wording avoids ambiguities.

jugglinmike:
I should add that I have written a number of tests for both realms and scripts
using the proposed API. Barring objection from anyone mentioned above, we're
planning on merging this documentation  and moving forward with those tests.

littledan:
Sounds good to me. Maybe clarify that it's a "same-origin" realm.

bterlson:
I think this is awesome! I will need to take a look at some of the tests you managed to write.

INTERPRETING.md
```diff
@@ -34,6 +34,32 @@ properties of the global scope prior to test execution.
...
+        3. Let s be ParseScript(sourceText, realm, hostDefined).
+        4. If s is a List of errors, then
+           a. Let error be the first element of s.
+           b. Return
+              Completion{[[Type]]: throw, [[Value]]: error, [[Target]]: empty}.
+        5. Let status be ScriptEvaluation(s).
+        6. Return Completion(status).
+
+  - **`global`** - a reference to the global `this` value on which `$` was
+    initially defined
```
leobalter: this is great to prevent misuses of `global` as there's a slight chance of `System.global` landing as `global` only.

domenic: Could we rename this globalThis?

bterlson: I think this is not "globalThis" but actually the global object.

domenic: Probably in the test runner environments these are all running in they are the same thing. I wonder which of the two the actual engines expose though. E.g. if I recall it's not possible to access the actual global via the V8 API.

bterlson: What's important that this prints 1: `$.evalScript('this.x = 1;print(x))`.

jugglinmike:
The "realms" and "scripts" tests are still blocked by a resolution to this
discussion. As far as I can see, the only reluctance here is related to
ambiguity in Test262 about the relationship between the global "this" value and
the global object.

I've submitted gh-668 in the hopes of resolving that ambiguity and subsequently
landing this patch.

jugglinmike:
This proposal consistently references "the global 'this' value," which is
problematic for two reasons:

1. It's imprecise. Implementors can provide whatever value they wish
2. It's redundant. As discussed in gh-568, tests are free to use the
   `this` keyword to access the global "this" value (although it's not clear
    why we would want to, given # 1 above)

Instead, `$.global` should be specified as "the global object". (I'm a bit slow on the uptake: @bterlson [recognized this a few days ago](https://github.com/tc39/test262/pull/644/files#r65402465) despite the contradictory text.) As @domenic pointed out in gh-668:

> Hosts could then, at their own risk, return the global this (even when that's
> not correct), but if they do so, test262 should not get blamed for test
> failures resulting from that.

jugglinmike:
I've updated this patch to consistently refer to "the global object" rather than "the global 'this' value." I would like to merge this in the next few days so that we can move on with the tests for realms and scripts. Please let me know if there are any further issues with this documentation.

leobalter:
Considering this last change have got not objections so far, I believe it's good to land. Updates and fixes may come on follow-up PRs.

PR #645, opened 2016-05-19
Improve test coverage for Symbol

jugglinmike:
I discovered some holes in the coverage for Symbol while working on some tests for realms. I think this addresses everything that was missing.


leobalter:
LGTM

PR #646, opened 2016-05-20
Add tests for object coercion of Reference base

jugglinmike:
These are kind of fundamental, and I expect that we implicitly test their concerns in other tests (well, GetValue, anyway). I'm working on related tests for realms, and it will be good to have some symmetry.

> Occurs that when a PutValue or GetValue operation is applied to a
> property reference with a primitive base, the base is first coerced to
> an Object.


leobalter:
💯

PR #648, opened 2016-05-23
Add test for NaN in OrdinaryDefineOwnProperty

jugglinmike:
These semantics are currently contested, but the test reflects the
intended behavior of the latest version of the specification.

Depending on the result of the following discussion, this test may need to be updated: https://github.com/tc39/ecma262/pull/353


leobalter:
The referenced issue has achieved consensus and it will be merged soon, I expect.

PR #651, opened 2016-05-26
Re-format destructuring assignment tests

jugglinmike:
Utilize the test generation tool to increase coverage of destructuring
assignment semantics. Previously, only destructuring assignment in the
AssignmentExpression position was tested. With this change applied, the
same tests will assert expected behavior for destructuring assignment in
`for..of` statements, as well.

A limited number of tests are applied to the `for..in` statement as
well, but due to the iteration protocol observed by that statement, many
destructuring tests are not relevant, and others cannot be automatically
generated from this format.

---

This may be surprising, but extending coverage in this way actually caches more
bugs in real engines!

A recent build of SpiderMonkey has this pass rate for the destructuring
assignment tests in `master` right now:

> Passed 327 tests (73.8%) / Failed 116 tests (26.2%)

...but that is limited to usage from the AssignmentExpression production.

This patch uses the test generation tool to create equivalent tests for
destructuring assignment from the head of a `for-of` statement. SpiderMonkey
has the following pass rate for those tests:

> Passed 307 tests (69.3%) / Failed 136 tests (30.7%)


leobalter:
LGTM

Issue #647, opened 2016-05-20
Definitive decision for Test262 console runner


jugglinmike:
This project includes a console runner to facilitate running tests from console
hosts. I understand that some folks at Mozilla rely on it in some capacity
(@tschneidereit may be able to give more detail), and the V8 team uses some of
its internals in their own Python-based runner.

TC-39 agreed to deprecate the runner in 2014. This decision wasn't
well-communicated, so there was some contention a few months later when
@domenic filed gh-128 to remove the runner. The discussion died off without a
change to TC-39's decision or to the runner itself.

While implementing new features for the test runner (first for the proposed
`async` flag, and later for executing module code), I came across a few of bugs
in the implementation (see gh-512). I submitted a bug fix (gh-514), but as an
alternative, I also suggested formal deprecation (gh-513) in light of the
long-standing TC-39 decision and because I was afraid I was devoting energy to
a rarely-used/updated tool.

As it stands, the runner is still defined on the project's `master` branch, but
its "deprecated" status is formally documented. As @littledan pointed out,
though:

> I don't really see the point in leaving something in the repository frozen
> and deprecated. Why don't we either let Till maintain it here for Mozilla's
> use, or remove it?
> 
> [...]
> 
> Let's bring this up at the next TC39 meeting to see if we should reverse the
> deprecation decision. To address the concern about wasted maintanence work:
> Whether or not the runner is deprecated, I don't think @jugglinmike is
> personally responsible for maintaining it. But it would definitely be
> unfortunate if Mozilla and V8 (for the inner parts Andreas mentioned) needed
> to maintain separate copies of the same code.

I am opening this issue to solicit feedback on this from any stakeholders. Some
ideas for moving forward:
- Simply remove the tool (expecting dependents to maintain their own versions)
- Formally re-instate the tool (expecting Test262 maintainers to continue
  maintenance responsibilities)
- Move the tool to a new repository, possibly within the TC39 GitHub
  organzation (allowing for shared maintenance of a single tool by some
  combination of Test262 maintainers and interested consumers)

@leobalter has plans to raise this subject at next week's TC-39 meeting, as
well.

tschneidereit:
Just as v8, we're currently using parts of the runner imported into our own repository. We were planning on using the upstream runner unmodified (with a small driver script), but when I learned about the deprecation I put that project on ice. While it'd obviously be nice for us to be able to use the runner, if we'd be the only ones to do so, I agree that removing it makes sense.

[...]

PR #654, opened 2016-05-28
[runner] Add support for "folding" block delimiter

jugglinmike:
We recently began using the `|` character as a delimiter for YAML blocks in test files. The content of these blocks can confuse the custom YAML parser, disrupting test runs.

[It looks like TC-39 has reached consensus](https://github.com/tc39/agendas/commit/a2fa5e1444552954c06e76d5cc6f120a5fd8d32c) on gh-647 ("Definitive decision for Test262 console runner"), but I'm not sure what that decision was.

In any case, the code is still in use, and this is a blocking bug (no pun intended--honest). The fix was simple enough; we can either merge it to `master` here, or allow consumers to apply the patch to their forks of the runner.

### DIFF
--- a/tools/packaging/monkeyYaml.py
+++ b/tools/packaging/monkeyYaml.py
@@ -43,7 +43,7 @@ def load(str):
     return dict
 
 def myReadValue(lines, value):
-    if value == ">":
+    if value == ">" or value == "|":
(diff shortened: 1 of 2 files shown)


leobalter:
We can merge this as a final last to the test runner. While it's still here, let it be fixed.

PR #655, opened 2016-05-28
Encode expected error "phase"

jugglinmike:
This patch implements the "negative reform" first proposed in gh-542. Please
refer to that issue for motivation and design discussion. I've incorporated all
the feedback shared there, and I've also applied the changes to the files that
have landed in `master` since that was submitted. This includes "static" test
files, the test generation tool, the new "INTERPRETING.md" documentation, and
the "case" files used as input for the generation tool.

Feedback of any sort is appreciated, but I suppose the change to
INTERPRETING.md is the only piece that will be interesting to those that have
been involved so far.

As before, I'm submitting this as a number of distinct commits in order to
facilitate review. I believe there is some value in merging these commits to
`master`, but I recommend doing so with a merge commit (since the test suite
passes through invalid states where only a subset of "negative" tests are
re-formatted).


leobalter:
I believe we are good to go on this, and the changes look good. 

@jugglinmike would you mind to rebase this branch? I can do this if you prefer (I would force push, preserving a copy in another remote branch).

jugglinmike:
@leobalter Given the amount of change in `master` over the past week or so, I
should probably do the rebasing. I will coordinate with you more closely when
I am ready so that we can avoid getting in to this situation again.

leobalter:
That's good, thanks

jugglinmike:
Superseded by gh-778.

Issue #588, opened 2016-04-19
Coverage: class extends null & constructor this value


bterlson:
https://github.com/tc39/ecma262/issues/543

I don't know if there exists tests for `this` being useless in such classes, but if there are, they should be deleted. In any case, new tests for the semantics in the fix to https://github.com/tc39/ecma262/issues/543 should be added.

PR #658, opened 2016-05-30
Update tests concerning null-extending classes

jugglinmike:
The latest revision of ECMA262 makes special provisions for classes
which extend the `null` value [1]. Update the relevant tests
accordingly.

[1] https://github.com/tc39/ecma262/issues/543

@bterlson I believe this fully addresses gh-588. As far as I can tell, there
are two observable difference for that change in the spec:

- There is no requirement for the constructor to return a value (ES2016
  required a return value in this case, and I've updated the relevant test)
- `this` may be referenced from within the constructor (ES2016 does not allow
  such a reference, but we previously had no test for that)

SuperCall continues to be invalid for the same reason. Despite the fact that
there was no change to that behavior, I added a test for good measure.

I'm reasonably confident that there are no lingering tests for the outdated
semantics. (My heuristic is manually inspecting all files that contain the
strings "extends" and "null" but not necessarily on the same line.)

    $ grep -E 'extends' test/ -r -l --null | xargs -0 grep null -l

This should resolve gh-588.

### DIFF
--- a/test/language/statements/class/subclass/class-definition-null-proto-super.js
+++ b/test/language/statements/class/subclass/class-definition-null-proto-super.js
@@ -1,4 +1,4 @@
-// Copyright (C) 2014 the V8 project authors. All rights reserved.
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
 // This code is governed by the BSD license found in the LICENSE file.
 /*---
 esid: sec-runtime-semantics-classdefinitionevaluation
--- a/test/language/statements/class/subclass/class-definition-null-proto-super.js
+++ b/test/language/statements/class/subclass/class-definition-null-proto-super.js
@@ -4,11 +4,8 @@
 esid: sec-runtime-semantics-classdefinitionevaluation
 description: >
   The `this` value of a null-extending class is automatically initialized,
-  obviating the need for an explicit return value in the constructor.
+  preventing the use of `super` from within the constructor.
 info: |
-  The behavior under test was introduced in the "ES2017" revision of the
-  specification and conflicts with prior editions.
-
   Runtime Semantics: ClassDefinitionEvaluation
 
   [...]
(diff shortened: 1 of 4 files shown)


test/language/statements/class/subclass/class-definition-null-proto-this.js
```diff
@@ -0,0 +1,46 @@
...
+     a. Let thisArgument be ? OrdinaryCreateFromConstructor(newTarget,
+        "%ObjectPrototype%").
+  [...]
+---*/
+
+var thisVal, instance;
+
+class C extends null {
+  constructor() {
+    thisVal = this;
```
leobalter: It's great to have this test and class-definition-null-proto-super.js

test/language/statements/class/subclass/class-definition-null-proto.js
```diff
@@ -1,19 +1,41 @@
 // Copyright (C) 2014 the V8 project authors. All rights reserved.
 // This code is governed by the BSD license found in the LICENSE file.
 /*---
-es6id: 14.5.14_S6.e
+esid: sec-runtime-semantics-classdefinitionevaluation
+es6id: 14.5.14
```
leobalter: this is a whole new test, I wonder if the copyright year should be 2016

test/language/statements/class/subclass/class-definition-null-proto-super.js
```diff
@@ -0,0 +1,67 @@
+// Copyright (C) 2014 the V8 project authors. All rights reserved.
```
leobalter: 2016

leobalter:
just a few descriptive content to fix, the tests LGTM.

test/language/statements/class/subclass/class-definition-null-proto.js
jugglinmike: Sure

test/language/statements/class/subclass/class-definition-null-proto-super.js
jugglinmike: Got it

jugglinmike:
All set!

PR #660, opened 2016-06-02
Add tests for default parameters

jugglinmike:
I've used the test generation tool for some of these tests, but other tests do
not generalize well. This includes:

- tests involving the `arguments` object (since arrow functions do not provide
  an `arguments` object)
- tests involving the `yield` token (since its interpretation is highly
  dependent on the function under test and the strictness of the surrounding
  code--which itself is informed by the function's presence in a class body)

Additionally, I decided not to use the tool for the tests for the `length`
property of function objects. Although this test *does* generalize well, it
would require a completely new set of test templates which would only be used
by a single `case` file. The conceptual overhead of the generation tool didn't
seem warranted for this case.

I found an opportunity to re-use prior work: the existing destructuring binding
tests are also valid when the binding takes place in the "default parameter"
position. This means we can add a few new template files for destructuring
binding and re-use all the existing tests there. For example, a destructuring
binding test like this:

    function f([x]) { /* assertions here */ } f([2]);

Can also be expressed as a default parameter test:

    function f([x] = [2]) { /* assertions here */ } f();

Finally, I've removed a handful of old tests for default parameters because
they did not offer consistent coverage (and they were not optimally organized).


src/dflt-params/duplicates.case
```diff
@@ -0,0 +1,12 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+desc: >
+  It is a Syntax Error if BoundNames of FormalParameters contains any duplicate
+  elements.
```
leobalter: nice one! If I'm not wrong, here is the ref: https://tc39.github.io/ecma262/#prod-ArrowFormalParameters

test/language/statements/static-method-length-dflt.js
```diff
@@ -0,0 +1,67 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
```
leobalter: is this intentionally in this folder?

test/language/expressions/arrow-function/length-dflt.js
```diff
@@ -0,0 +1,60 @@
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
```
leobalter: there's a few files with 2015, is this right?

src/dflt-params/duplicates.case
jugglinmike: Actually, I believe it's here https://tc39.github.io/ecma262/#sec-function-definitions-static-semantics-early-errors . I'll add an "info" field with that text

jugglinmike:
@leobalter I guess I forgot about this pull request, somehow. Anyway, I think I've addressed all your feedback.

leobalter:
I rebased and squashed the fixup commits. The original branch changes are kept on `dflt-params-bkp` on the same original fork.

PR #662, opened 2016-06-02
`Object.getOwnPropertyDescriptors`: add test to ensure undefined descriptors are not added

ljharb:
Per https://github.com/tc39/ecma262/pull/593

**NOTE**: do not merge yet, until TC39 consensus is reached and the above PR is also merged.

### DIFF
--- a/test/built-ins/Object/getOwnPropertyDescriptors/duplicate-keys.js
+++ b/test/built-ins/Object/getOwnPropertyDescriptors/duplicate-keys.js
@@ -27,5 +27,12 @@ var proxy = new Proxy({}, {
 
 var result = Object.getOwnPropertyDescriptors(proxy);
 assert.sameValue(result.hasOwnProperty('DUPLICATE'), true);
-assert.sameValue(result.DUPLICATE, undefined);
+
+var lastDescriptor = descriptors[descriptors.length - 1];
+assert.notSameValue(result.DUPLICATE, lastDescriptor);
+assert.sameValue(result.DUPLICATE.enumerable, lastDescriptor.enumerable);
+assert.sameValue(result.DUPLICATE.configurable, lastDescriptor.configurable);
+assert.sameValue(result.DUPLICATE.value, lastDescriptor.value);
+assert.sameValue(result.DUPLICATE.writable, lastDescriptor.writable);
+
 assert.sameValue(log, 'ownKeys|getOwnPropertyDescriptor:DUPLICATE|getOwnPropertyDescriptor:DUPLICATE|getOwnPropertyDescriptor:DUPLICATE|');
--- a/test/built-ins/Object/getOwnPropertyDescriptors/proxy-undefined-descriptor.js
+++ b/test/built-ins/Object/getOwnPropertyDescriptors/proxy-undefined-descriptor.js
@@ -1,0 +1,32 @@
+// Copyright (C) 2016 Jordan Harband. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Object.getOwnPropertyDescriptors should filter out undefined OwnPropertyDescriptors
+esid: sec-object.getownpropertydescriptors
+author: Jordan Harband
+features: [Proxy]
+includes: [proxyTrapsHelper.js]
+---*/
+
+var key = "a";
+var ownKeys = [key];
+var badProxyHandlers = allowProxyTraps({
+  getOwnPropertyDescriptor: function () {},
+  ownKeys: function () {
+	return ownKeys;
+  }
+});
+var proxy = new Proxy({}, badProxyHandlers);
+
+var keys = Reflect.ownKeys(proxy);
+assert.notSameValue(keys, ownKeys, 'Object.keys returns a new Array');
+assert.sameValue(Array.isArray(keys), true, 'Object.keys returns an Array');
+assert.sameValue(keys.length, ownKeys.length, 'keys and ownKeys have the same length');
+assert.sameValue(keys[0], ownKeys[0], 'keys and ownKeys have the same contents');
+
+var descriptor = Object.getOwnPropertyDescriptor(proxy, key);
+assert.sameValue(descriptor, undefined, "Descriptor matches result of [[GetOwnPropertyDescriptor]] trap");
+
+var result = Object.getOwnPropertyDescriptors(proxy);
+assert.sameValue(key in result, false, "key is not present in result");


leobalter:
Thank you!

caitp:
This would also require changes in built-ins/Object/getOwnPropertyDescriptors/duplicate-keys.js

ljharb:
@caitp thanks, updated

test/built-ins/Object/getOwnPropertyDescriptors/duplicate-keys.js
```diff
@@ -21,7 +21,7 @@ var proxy = new Proxy({}, {
   },
   getOwnPropertyDescriptor(t, name) {
     log += 'getOwnPropertyDescriptor:' + name + '|';
-    return descriptors[i++];
+    return descriptors[i >= descriptors.length ? descriptors.length - 1 : i++];
```
caitp: no, its the assertion on line  30 that is wrong here, it shouldn't be overwritten with undefined.

caitp: should be equal to the first returned descriptor, in this case

caitp: wait, I'm wrong, the second returned descriptor. But not the undefined one :)

ljharb: ah right, thanks

test/built-ins/Object/getOwnPropertyDescriptors/duplicate-keys.js
```diff
@@ -27,5 +27,5 @@ var proxy = new Proxy({}, {
 
 var result = Object.getOwnPropertyDescriptors(proxy);
 assert.sameValue(result.hasOwnProperty('DUPLICATE'), true);
-assert.sameValue(result.DUPLICATE, undefined);
+assert.sameValue(result.DUPLICATE, descriptors[descriptors.length - 1]);
 assert.sameValue(log, 'ownKeys|getOwnPropertyDescriptor:DUPLICATE|getOwnPropertyDescriptor:DUPLICATE|getOwnPropertyDescriptor:DUPLICATE|');
```
caitp: Last nit: the result from the trap is not the same value as the item added to the result object, even though in this case they should have the same keys and values.

https://tc39.github.io/ecma262/#sec-proxy-object-internal-methods-and-internal-slots-getownproperty-p

ljharb: aha, good call again :-) thanks, keep the nits coming!

test/built-ins/Object/getOwnPropertyDescriptors/proxy-undefined-descriptor.js
```diff
@@ -0,0 +1,29 @@
...
+var ownKeys = [key];
+var badProxyHandlers = allowProxyTraps({
+  getOwnPropertyDescriptor: function () {},
+  ownKeys: function () {
+	return ownKeys;
+  }
+});
+var proxy = new Proxy({}, badProxyHandlers);
+
+var keys = Object.keys(proxy);
```
caitp: I guess there's another nit --- Object.keys() does not simply return the result of `O.[[OwnPropertyKeys]]()`, so this will not be the same value.

Maybe test262 could benefit from a "deepEquals()" sort of assertion to simplify this sort of thing?

test/built-ins/Object/getOwnPropertyDescriptors/proxy-undefined-descriptor.js
```diff
@@ -0,0 +1,29 @@
...
+assert.sameValue(keys, ownKeys, 'Object.keys returns the result of the [[OwnKeys]] trap');
```
caitp: side note, I am a little confused about why this assertion is in a gOPDs test case, seems more appropriate for Proxy or Object.keys. ditto for the Object.getOwnPropertyDescriptor test below

jugglinmike: When it comes to a "deepEquals" helper, we have had arguments both for and against--see gh-149. For my part, I think that a complete "deepEquals" is generally much more complicated than what we need for Test262 tests. [`harness/compareArray.js`](https://github.com/tc39/test262/blob/fce8b5852d7d9d3839dea6922bae306934ac115a/harness/compareArray.js) is available for shallow comparison of array-like objects, though even that is generally more abstraction than I'm comfortable with.

For instance, this test compares two arrays of length 1, so it can be written with two assertions (no need for external functions, loops, or recursion of any kind):

```js
assert.sameValue(keys.length, 1);
assert.sameValue(keys[0], 'a');
```

Though to Caitlin's point, it's probably a good idea to also add in `assert.notSameValue(keys, ownKeys);`.

Beyond just being spartan for the sake of it, explicit tests like these are less prone to false positives resulting from comparing one computed value against another ([example here](https://github.com/tc39/test262/pull/602#discussion_r61342819))

ljharb: I'll update to explicit checks. (I also want to assert that things are an array rather than relying on an assumption)

caitp: Those assertions still seem out of place (probably duplicated in other parts of test262), but if maintainers are okay with that, I guess that's fine

ljharb:
This received consensus at the July 2016 TC39 meeting; please merge at your leisure.

test/built-ins/Object/getOwnPropertyDescriptors/proxy-undefined-descriptor.js
```diff
@@ -0,0 +1,32 @@
...
+  getOwnPropertyDescriptor: function () {},
+  ownKeys: function () {
+	return ownKeys;
+  }
+});
+var proxy = new Proxy({}, badProxyHandlers);
+
+var keys = Object.keys(proxy);
+assert.notSameValue(keys, ownKeys, 'Object.keys returns a new Array');
+assert.sameValue(Array.isArray(keys), true), 'Object.keys returns an Array');
```
evilpie: typo `true)`

test/built-ins/Object/getOwnPropertyDescriptors/proxy-undefined-descriptor.js
```diff
@@ -0,0 +1,32 @@
...
+assert.sameValue(keys.length, ownKeys.length, 'keys and ownKeys have the same length');
```
evilpie: I just ran this in Firefox and this is not true. https://tc39.github.io/ecma262/#sec-enumerableownproperties ignores undefined descriptors. Did you mean to use Reflect.ownKeys?

ljharb: whoops, yes, i did. will fix.

ljharb:
ping :-)

test/built-ins/Object/getOwnPropertyDescriptors/proxy-undefined-descriptor.js
```diff
@@ -0,0 +1,32 @@
+// Copyright (C) 2016 Jordan Harband. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Object.getOwnPropertyDescriptors should filter out undefined OwnPropertyDescriptors
+esid: sec-object.getownpropertydescriptors
+author: Jordan Harband
+featurewws: [Proxy]
```
leobalter: I believe your cat typed this. We do not support meow features yet.

ljharb: /me shakes fist at vim

ljharb: updated

leobalter:
Thanks!

Issue #579, opened 2016-04-14
Tests for CreateByteDataBlock


jugglinmike:
I believe we are missing tests that assert that the following step of [CreateByteDataBlock](https://tc39.github.io/ecma262/#sec-createbytedatablock) is honored:

> 3\. Set all of the bytes of db to 0.

e.g.

```js
var x = new Int8Array(4);
assert.sameValue(x[0], 0);
assert.sameValue(x[1], 0);
assert.sameValue(x[2], 0);
assert.sameValue(x[3], 0);
```

PR #663, opened 2016-06-03
Add tests for CreateByteDataBlock

jugglinmike:
Ensure that all bytes of newly-created Data Block are initialized to
zero.

Resolves gh-579.


leobalter:
I renamed this PR to reflect the method being tested. No farewells to Data Blocks here.

PR #661, opened 2016-06-02
Add tests for Number#toPrecision

leobalter:
I'm not happy with the lack of separation for the tests on return-values.js, but it sounded less confusing this way.

PR #664, opened 2016-06-03
Add tests for Number#toExponential

leobalter:


jugglinmike:
This patch duplicates assertions in the same was as gh-661, but I'm willing to
let that slide. Just a couple of other things:

- The file named `undefined-fractionDigits.js` should be re-named to use
  lower-case letters
- We're missing an interesting edge case: when the value of "n" is not uniquely
  determined by the algorithm. I had to work this out on a paper, so if you
  need a hand writing that test, I may be able to help.

jugglinmike:
This patch could still use coverage for that edge case.

leobalter:
I missed this one, I'll do it.

leobalter:
I've changed a few assertions to actually compare different results when fractionDigits is undefined and it may act different than being 0.

I don't know if that provides the behavior you asked but it was good to add.

test/built-ins/Number/prototype/toExponential/infinity.js
```diff
@@ -0,0 +1,29 @@
+// Copyright (C) 2016 The V8 Project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+es6id: 20.1.3.2
+esid: sec-number.prototype.toexponential
+description: >
+  Return "NaN" if this is NaN
```
tcare: Infinity?

leobalter: oops. fixed

tcare:
Looks good.

PR #667, opened 2016-06-06
Add tests for usage of IsArray

jugglinmike:
Interestingly, SpiderMonkey fails just one of these specific cases. The
following code is expected to throw a TypeError, but does not in SpiderMonkey:

    var h = Proxy.revocable({}, {});
    h.revoke();
    Object.prototype.toString.call(h.proxy);

### DIFF
--- a/test/built-ins/Array/isArray/proxy.js
+++ b/test/built-ins/Array/isArray/proxy.js
@@ -1,0 +1,30 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-array.isarray
+es6id: 22.1.2.2
+description: Proxy of an array is treated as an array
+info: |
+  1. Return IsArray(arg).
+
+  7.2.2 IsArray
+
+  [...]
+  3. If argument is a Proxy exotic object, then
+     a. If the value of the [[ProxyHandler]] internal slot of argument is null,
+        throw a TypeError exception.
+     b. Let target be the value of the [[ProxyTarget]] internal slot of
+        argument.
+     c. Return ? IsArray(target).
+features: [Proxy]
+---*/
+
+var objectProxy = new Proxy({}, {});
+var arrayProxy = new Proxy([], {});
+var arrayProxyProxy = new Proxy(arrayProxy, {});
+
+assert.sameValue(Array.isArray(objectProxy), false);
+assert.sameValue(Array.isArray(arrayProxy), true, 'proxy for array');
+assert.sameValue(
+  Array.isArray(arrayProxyProxy), true, 'proxy for proxy for array'
+);
--- a/test/built-ins/Array/prototype/concat/create-proxy.js
+++ b/test/built-ins/Array/prototype/concat/create-proxy.js
@@ -1,0 +1,40 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 22.1.3.1
+esid: sec-array.prototype.concat
+description: Species constructor of a Proxy object whose target is an array
+info: |
+    [...]
+    2. Let A be ? ArraySpeciesCreate(O, 0).
+    [...]
+    7. Return A.
+
+    9.4.2.3 ArraySpeciesCreate
+
+    [...]
+    3. Let isArray be ? IsArray(originalArray).
+
+    7.2.2 IsArray
+
+    [...]
+    3. If argument is a Proxy exotic object, then
+       a. If the value of the [[ProxyHandler]] internal slot of argument is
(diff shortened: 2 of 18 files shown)


test/built-ins/Array/isArray/proxy.js
```diff
@@ -0,0 +1,30 @@
...
+---*/
+
+var objectProxy = new Proxy({}, {});
+var arrayProxy = new Proxy([], {});
+var arrayProxyProxy = new Proxy(arrayProxy, {});
+
+assert.sameValue(Array.isArray(objectProxy), false);
+assert.sameValue(Array.isArray(arrayProxy), true, 'proxy for array');
+assert.sameValue(
+  Array.isArray(arrayProxyProxy), true, 'proxy for proxy for array'
```
leobalter: lol, the name is funny and it's great IsArray handling proxy targets.

test/built-ins/Array/prototype/concat/create-proxy.js
```diff
@@ -0,0 +1,40 @@
...
+var proxy = new Proxy(new Proxy(array, {}), {});
+var Ctor = function() {};
+var result;
+
+array.constructor = function() {};
+array.constructor[Symbol.species] = Ctor;
+
+result = Array.prototype.concat.call(proxy);
+
+assert.sameValue(Object.getPrototypeOf(result), Ctor.prototype);
```
leobalter: nice!

test/built-ins/Array/prototype/concat/is-concat-spreadable-proxy-revoked.js
```diff
@@ -0,0 +1,43 @@
...
+     c. Return ? IsArray(target).
+features: [Proxy]
+---*/
+
+var handle = Proxy.revocable([], {});
+
+handle.revoke();
+
+assert.throws(TypeError, function() {
+  [].concat(handle.proxy);
```
leobalter: this throws a TypeError but it does not reach the IsArray check. It returns the abrupt from IsConcatSpreadable step 2

The test and description are right, but the info can be adjusted.

leobalter: in anyway, we could have another test where the Proxy is revoked from handler.get trap which returns undefined.

test/built-ins/Array/prototype/concat/is-concat-spreadable-proxy-revoked.js
```diff
@@ -0,0 +1,43 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-isconcatspreadable
+es6id: 22.1.3.1.1
```
leobalter: This is minor, but I think the ids should refer to [].concat in 22.1.3.1 / sec-array.prototype.concat

test/built-ins/JSON/parse/revived-proxy-revoked.js
```diff
@@ -0,0 +1,49 @@
...
+
+assert.throws(TypeError, function() {
+  JSON.parse('[null, null]', function() {
+    callCount += 1;
+    this[1] = handle.proxy;
+    returnCount += 1;
+  });
+});
+
+assert.sameValue(callCount, 1, 'reviver is invoked exactly once');
```
leobalter: the callCount seems unnecessary as we should not expect any normal completion before the returnCount increment

test/built-ins/JSON/parse/revived-proxy.js
```diff
@@ -0,0 +1,82 @@
...
+
+arrayProxy.other = 0;
+
+injectProxy = function(name, val) {
+  if (name === 'other') {
+    visitedOther = true;
+  }
+  this[1] = objectProxy;
+  return val;
+};
```
leobalter: wow!

wow

leobalter:
Just a few parts mentioned above to fix.

test/built-ins/Array/prototype/concat/is-concat-spreadable-proxy-revoked.js
jugglinmike: Nice catch! The correct version of this test exposes a bug in V8

test/built-ins/JSON/parse/revived-proxy-revoked.js
jugglinmike: I suppose you're right. I think I was being a little too defensive here

test/built-ins/Array/prototype/concat/is-concat-spreadable-proxy-revoked.js
jugglinmike: Following that pattern would make the IDs largely redundant with the information encoded in the test file names.

PR #668, opened 2016-06-08
Document requirements for global `this` binding

jugglinmike:
In gh-644, @domenic reminded me that the global "this" value is distinct from
the global object. However, many tests in Test262 are written assuming that
they are the same value. Although our preference should be to avoid adding
requirements beyond those defined in ECMA262, I don't think it would be
possible to test the relevant semantics without making this assumption. That's
why I'm suggesting that we formally document it as a requirement for running
Test262.

Commit message:

> Document requirements for global `this` binding
> 
> ECMA262 states, "Hosts may provide any ECMAScript Object value [for the
> global 'this' value]." However, many tests in Test262 are written under
> the assumption that the global "this" value is the global object.
> Formally document this expectation so it is more easily discovered by
> consumers.

### DIFF
--- a/INTERPRETING.md
+++ b/INTERPRETING.md
@@ -26,10 +26,23 @@ global scope prior to test execution:
 1. `harness/assert.js`
 2. `harness/sta.js`
 
-### Host-Defined Functions
+### Host-Defined Behaviors
 
-The following values must be defined as writable, configurable, non-enumerable
-properties of the global scope prior to test execution.
+ECMA262 includes features which are open to definition/extension by the host
+environment. In order to consistently assert the semantics of operations built
+on these extension points, Test262 tests assumes the host conforms to the
+semantics defined in this section.
+
+**Global "this" Value** The "this" value of every global [Environment
+Record](https://tc39.github.io/ecma262/#sec-environment-records) must be the
+global object itself (as may be implemented by setting `thisValue` to
+`undefined` in the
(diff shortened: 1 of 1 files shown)


leobalter:
LGTM.

I'll wait for more comments before merging this.

domenic:
I'm a strong -1 on this. This means test262 is not applicable to web browsers (where the two are *always* different), which I think is not the intention.

jugglinmike:
Ah, that's a good point. Do you agree that documentation is called for here? If
so, can you think of a way to specify the semantics we need for testing
purposes?

My own attempt is weak enough to make me question the utility of any such
documentation:

> Test262 tests interact with the global "this" value as though it were the
> global object itself. Implementors are encouraged to define the global "this"
> value in this way, although proxied access may be sufficient to satisfy the
> test expectations.

domenic:
I think ideally test262 should not rely on them behaving the same. For example, maybe test262 should introduce a host-implemented function, `getGlobal()`, and use that when testing the global. Hosts could then, at their own risk, return the global `this` (even when that's not correct), but if they do so, test262 should not get blamed for test failures resulting from that.

jugglinmike:
That sounds like a good solution to me, although I think a value (rather than a
value-returning function) should be sufficient. We should be able to address
this with gh-644 itself, though with a modification. I'll close this and
elaborate there.

PR #669, opened 2016-06-09
Update tests for TypedArrays, DataView and ArrayBuffer

leobalter:
Ref tc39/ecma262#410


test/built-ins/DataView/prototype/setUint8/toindex-byteoffset.js
```diff
@@ -0,0 +1,109 @@
...
+sample.setUint8(1.1, 42);
+assert.sameValue(typedArray[1], 42, "1.1");
+
+sample.setUint8(0, 0);
+sample.setUint8(1.9, 42);
+assert.sameValue(typedArray[1], 42, "1.9");
+
+sample.setUint8(0, 0);
+sample.setUint8(-0.1, 42);
+assert.sameValue(typedArray[0], 42, "-0.1");
```
jugglinmike: This is a little surprising, but it's definitely to-spec.

test/built-ins/DataView/prototype/getFloat32/toindex-byteoffset.js
```diff
@@ -0,0 +1,64 @@
...
+sample.setUint8(2, 76);
+sample.setUint8(3, 76);
+sample.setUint8(4, 75);
+sample.setUint8(5, 75);
+sample.setUint8(6, 76);
+sample.setUint8(7, 76);
+
+var obj1 = {
+  valueOf: function() {
+    return 3;
```
jugglinmike: I like your idea to use distinct return values for `toString` and `valueOf`--is there any reason that you did not make this change for the other tests?

test/built-ins/DataView/prototype/getFloat32/toindex-byteoffset.js
```diff
@@ -0,0 +1,64 @@
...
+assert.sameValue(sample.getFloat32(NaN), 13323340, "NaN");
+assert.sameValue(sample.getFloat32(null), 13323340, "null");
+assert.sameValue(sample.getFloat32(0.1), 13323340, "0.1");
+assert.sameValue(sample.getFloat32(0.9), 13323340, "0.9");
+assert.sameValue(sample.getFloat32(1.1), 13388875, "1.1");
+assert.sameValue(sample.getFloat32(1.9), 13388875, "1.9");
+assert.sameValue(sample.getFloat32(-0.1), 13323340, "-0.1");
+assert.sameValue(sample.getFloat32(-0.99999), 13323340, "-0.99999");
+assert.sameValue(sample.getFloat32(undefined), 13323340, "undefined");
+assert.sameValue(sample.getFloat32(), 13323340, "no arg");
```
jugglinmike: Great to see all these assertions

jugglinmike:
Leo, hello!

This is looking good. I left a few comments in-line, but a couple others apply to multiple files:

- `DataView/prototype/*/negative-byteoffset-throws.js` the `info` field of
  these files still documents the old semantics (pretty minor, but it confused
  me when reviewing this change set)
- `DataView/prototype/*/toindex-byteoffset.js`
  These tests re-set the buffer value to `0` immediately before each assertion.
  Some of the "setup" steps introduced in this patch set the value at the wrong
  offset, making them susceptible to false positives. For example:

       sample.setInt32(0, 0);
       sample.setInt32(1.1, 42);
       assert.sameValue(sample.getInt32(1), 42, "1.1");

       sample.setInt32(0, 0);
       sample.setInt32(1.9, 42);
       assert.sameValue(sample.getInt32(1), 42, "1.9");

test/built-ins/DataView/prototype/getFloat32/toindex-byteoffset.js
leobalter: When it does return something > 1, I believe it's already specific enough, but a few cases (on DataView) are returning 1, I'll change these.

leobalter:
> DataView/prototype/*/negative-byteoffset-throws.js the info field of these files still documents the old semantics (pretty minor, but it confused me when reviewing this change set)

They look fine here, or I couldn't find anything. Can you point me to an example?

leobalter:
17e3115 fixes the resets and the different values for {}.toString and {}.valueOf

jugglinmike:
> They look fine here, or I couldn't find anything. Can you point me to an example? 

Looks like I referenced the wrong file name--sorry for the confusion. The files
I had in mind actually match the pattern
`DataView/prototype/*/detached-buffer-after-integer-byteoffset.js`. For
example:

```diff
--- a/test/built-ins/DataView/prototype/getFloat32/detached-buffer-after-integer-byteoffset.js
+++ b/test/built-ins/DataView/prototype/getFloat32/detached-buffer-after-integer-byteoffset.js
@@ -17,22 +17,19 @@ info: |
   24.2.1.1 GetViewValue ( view, requestIndex, isLittleEndian, type )
 
   ...
   6. If numberIndex ≠ getIndex or getIndex < 0, throw a RangeError exception.
   ...
   8. Let buffer be the value of view's [[ViewedArrayBuffer]] internal slot.
   9. If IsDetachedBuffer(buffer) is true, throw a TypeError exception.
   ...
 includes: [detachArrayBuffer.js]
 ---*/
 
 var buffer = new ArrayBuffer(6);
 var sample = new DataView(buffer, 0);
 
 $DETACHBUFFER(buffer);
-assert.throws(RangeError, function() {
-  sample.getFloat32(1.1);
-});
```

Here, the quoted steps in GetViewValue are no longer accurate. I don't think
this would be worth correcting if it were just about the step numbers, though
(that would border on busywork). In this case, the source of the expected
RangeError has changed, and this could confuse people reading the test.

jugglinmike:
Merged to `master` at commit cfc77c83e3ccbf0070968f01cd3ff7a904975b40. Nice going, Leo!

Issue #671, opened 2016-06-11
SpreadElement used in ArrayLiteral is lacking tests


avdg:
I've noticed only 26 indirect test related to testing spread in `ArrayLiteral`s, they are easy to pass if spread is followed only by an identifier.  A `assignmentExpression` is allowed after the spread punctuator in current grammar.

jugglinmike:
Thanks for the report! It's great to see parsers making use of Test262 and even better to see contributions coming back. (By the way, you may be interested in gh-196 and the proposed solution at gh-655.)

@leobalter and I recently began [a coverage audit for ES2015 features](https://github.com/bocoup/test262-v8-machinery/wiki/2016-audit), and "spread in array initializers" made our list, too.

I've submitted gh-672 to address this.

avdg:
Thanks! And yeah, I had some doubts about these negative tests. Thanks for clarifying that error will be checked soon.

jugglinmike:
Apologies: I'm a little confused about your comment. When you say, "some doubts
about these negative tests," are you referring to technically incorrect test
files (either in the `master` branch or in this pull request)? Or are you
referring to the difficulty of consuming tests that are expected to produce
runtime errors?

avdg:
Uh, I rather thought about the error checking involved into these negative tests.

avdg:
And sorry, this was a bit off topic

jugglinmike:
No problem--I just wanted to be sure that we don't have any invalid tests!

michaelficarra:
@avdg: You may also be interested in #559. I'd love to see UglifyJS start using those tests. I will get them published to https://github.com/tc39/test262-parser-tests ASAP.

avdg:
@michaelficarra Are there es5.1 specific tests too?

michaelficarra:
@avdg No. In my opinion, all of the tests should be kept up to date with the latest working draft. Anyway, the tests are now published at https://github.com/tc39/test262-parser-tests.

avdg:
Should now be resolved in #672

PR #672, opened 2016-06-13
Improve coverage for "spread" operator

jugglinmike:
As reported in gh-671, Test262's coverage of the spread operator is lacking in
the following respects:

- It does not include the use of the operator within array initializers
- It does not include the application of the operator to expressions other than
  Identifiers

This patch introduces coverage for both cases.

Existing coverage of the operator is provided by the test generation tool, and
I wanted to draw attention to that here. In order to re-use the same test
cases, the formulation of these new tests is a little indirect. For instance:

```js
(function() {
  assert.sameValue(arguments.length, 3);
  assert.sameValue(arguments[0], 3);
  assert.sameValue(arguments[1], 4);
  assert.sameValue(arguments[2], 5);
}.apply(null, [...[3, 4, 5]]));
```

Referencing the array instance would be more direct, but it would not
generalize for re-use with the existing tests for MemberExpression,
CallExpression, and SuperCall.

### DIFF
--- a/src/spread/sngl-expr.case
+++ b/src/spread/sngl-expr.case
@@ -1,0 +1,35 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+desc: Spread operator applied to AssignmentExpression as only element
+template: default
+info: |
+    12.3.6.1 Runtime Semantics: ArgumentListEvaluation
+
+    ArgumentList : ... AssignmentExpression
+
+    1. Let list be an empty List.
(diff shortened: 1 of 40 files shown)


src/spread/sngl-expr.case
```diff
@@ -0,0 +1,35 @@
...
+var source = [2, 3, 4];
+var target;
+//- args
+...target = source
+//- body
+assert.sameValue(arguments.length, 3);
+assert.sameValue(arguments[0], 2);
+assert.sameValue(arguments[1], 3);
+assert.sameValue(arguments[2], 4);
+assert.sameValue(target, source);
```
leobalter: nice!

leobalter:
LGTM

PR #673, opened 2016-06-14
Relax test for forbidden extension

jugglinmike:
According to ES2015, functions created using the FunctionDeclaration
syntax may define "own" properties named "caller" and "arguments" as
long as this is done using the AddRestrictedFunctionProperties abstract
operation.

Remove the assertions that prohibit this extension.

@caitp Does this look okay to you?


tcare:
Looks good.

PR #678, opened 2016-06-18
Switch to non-positional argument.

samccone:
Since the argument is required, we mark it as so. Using this approach
gives the user a much nicer error message, as compared to just the "not
enough args" message.


leobalter:
LGTM. @jugglinmike, any other thoughts?

tcare:
Looks good.

samccone:
cheers :+1:

PR #679, opened 2016-06-20
Remove precision-equivalency tests

leobalter:
Due to specs approximation values, some tests are invalid as the harness
helpers rely on another approximated values as Math.LN2.


jugglinmike:
Nice find!

Issue #665, opened 2016-06-06
Are there full tests for iterators/generators and yield* ?


littledan:
We recently found a bug in V8 where generators weren't wrapping the value for return in an iteration in a particular case https://bugs.chromium.org/p/v8/issues/detail?id=5057 . Does test262 have good tests for this sort of path? I don't think this is a "one implementation" sort of issue as there is a particular line in the spec indicating returning an iteration result that we didn't do. @jugglinmike

jugglinmike:
Test262's coverage for delegating yield is pretty minimal--more than a third are surface tests dedicated to newline parsing:

    $ grep -E 'yield\s*\*' test/ -rl
    test/language/statements/generators/yield-star-before-newline.js
    test/language/statements/class/definition/methods-gen-yield-star-before-newline.js
    test/language/statements/for-of/yield-star-from-finally.js
    test/language/statements/for-of/yield-star-from-catch.js
    test/language/statements/for-of/yield-star.js
    test/language/statements/for-of/yield-star-from-try.js
    test/language/expressions/yield/star-iterable.js
    test/language/expressions/yield/star-string.js
    test/language/expressions/yield/star-array.js
    test/language/expressions/generators/yield-star-before-newline.js
    test/language/expressions/object/method-definition/yield-star-before-newline.js

I've already started a coverage audit, but I'm still way back in section 8. I've added [Section 14.4.14](https://tc39.github.io/ecma262/#sec-generator-function-definitions-runtime-semantics-evaluation) to our list. We're going to try maintaining it as a public wiki page; you can find it here: https://github.com/bocoup/test262-v8-machinery/wiki/2016-audit

PR #681, opened 2016-06-21
Improve coverage for YieldExpression

jugglinmike:
This is broken up in to two commits: the first re-organizes some existing tests for `yield` (and removes duplicates). The second introduces brand new tests, largely focus on "delegating yield." Those tests themselves are mainly focused on how the expression behaves for various completion types, named accordingly: `star-rhs-iter-{nrml,rtrn,thrw}-*.js`.

This includes tests for the `yield*` semantics introduced for ES2016 discussed in https://github.com/tc39/ecma262/issues/293. (I have omitted `es6id` from the frontmatter of those test files.)

This should resolve gh-665.


tcare:
Looks good!

PR #684, opened 2016-06-22
Extend test for HTML comments

jugglinmike:
The V8 engine incorrectly requires a leading newline character for
MultiLineComments which contain the optional trailing HTMLCloseComment
[1]. Extend the current tests to fail when such an invalid restriction
is in place.

[1] https://bugs.chromium.org/p/v8/issues/detail?id=5142

### DIFF
--- a/test/annexB/language/comments/multi-line-html-close.js
+++ b/test/annexB/language/comments/multi-line-html-close.js
@@ -20,29 +20,30 @@ info: |
 negative: Test262Error
 ---*/
 
+
 var counter = 0;
-/*
+0/*
 */-->
 counter += 1;
 
-/*
+0/*
 */-->the comment extends to these characters
 counter += 1;
 
-/* optional FirstCommentLine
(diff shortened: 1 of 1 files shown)


tcare:
Do we have any tests for these cases without a prefix?

jugglinmike:
If I understand your question correctly, then the answer is, "we will not if this patch is applied in its current form." As surprising as the V8 behavior is, it seems even less likely that a bug would be localized to the use of this production *without* a preceding token. But who am I to say? I'll push up a commit that re-factors this to cover both cases simultaneously.

tcare:
Sorry I could have worded that better! Yes it seemed like the V8 bug was a little strange. Thanks for updating.

LGTM.

tcare:
BTW, that V8 bug referencing my change is correct that my interpretation of the spec was wrong for the single line delimited comment case. I've filed a bug at https://github.com/Microsoft/ChakraCore/issues/1176

PR #687, opened 2016-06-22
Improve coverage for "super"

jugglinmike:
This includes tests for both SuperProperty and SuperCall. I've organized them in `language/expressions/super/`, to reflect [the spec's treatment of Runtime semantics, where they are grouped together](https://tc39.github.io/ecma262/#sec-makesuperpropertyreference). I'd also be open to splitting them between `language/expressions/call/` and `langauge/expressions/member/` if anyone thinks that organizing according to the grammar would be better.


suwc:
LGTM

PR #688, opened 2016-06-22
Add tests for realm interactions

jugglinmike:
ECMA262 defines a "realm" as follows:

> Before it is evaluated, all ECMAScript code must be associated with a realm.
> Conceptually, a realm consists of a set of intrinsic objects, an ECMAScript
> global environment, all of the ECMAScript code that is loaded within the
> scope of that global environment, and other associated state and resources.

Although there is currently no standard mechanism for the creation of realms,
many hosts offer this functionality through custom APIs. In those hosts, the
semantics describing realm interactions are directly observable.

In order to guarantee conformance to the specification in advance of a
standardized API, Test262 now requires that hosts provide a `$.createRealm`
function whose behavior is defined in the projects "INTERPRETING.md" file.

Use this host-provided API to ensure that implementations correctly observe the
specification text that dictates realm interactions.

### DIFF
--- a/test/built-ins/TypedArrays/typedarray-arg-other-ctor-buffer-ctor-custom-species-proto-from-ctor-realm.js
+++ b/test/built-ins/TypedArrays/typedarray-arg-other-ctor-buffer-ctor-custom-species-proto-from-ctor-realm.js
@@ -1,0 +1,59 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-typedarray-typedarray
+description: >
+  Derive the ArrayBuffer prototype from the realm of the species constructor
+info: >
+  22.2.4.3 TypedArray ( typedArray )
+
+  This description applies only if the TypedArray function is called with at
+  least one argument and the Type of the first argument is Object and that
+  object has a [[TypedArrayName]] internal slot.
+
+  ...
+  18. Else,
+    a. Let bufferConstructor be ? SpeciesConstructor(srcData, %ArrayBuffer%).
+    b. Let data be ? AllocateArrayBuffer(bufferConstructor, byteLength).
+  ...
+
+  7.3.20 SpeciesConstructor ( O, defaultConstructor )
+
+  ...
+  5. Let S be ? Get(C, @@species).
+  6. If S is either undefined or null, return defaultConstructor.
+  7. If IsConstructor(S) is true, return S.
+  ...
+
+  9.1.14 GetPrototypeFromConstructor
+
+  ...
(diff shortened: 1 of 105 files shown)


jugglinmike:
As part of [an ongoing effort to improve coverage](https://github.com/bocoup/test262-v8-machinery/wiki/2016-audit), I recently identified another case where a specification-defined error may be created in the wrong realm:

```
new nonConstructorFromAnotherRealm();
```

Because the function is from another realm and because it is ostensibly being executed, runtimes may mistakenly interpret that function as "the active function object." In fact (as defined by [the EvaluateNew algorithm](https://tc39.github.io/ecma262/#sec-evaluatenew)), such a non-constructor function is never formally invoked, so the TypeError produced should be derived from the %TypeError% intrinsic of the *current* realm.

SpiderMonkey exhibits the incorrect behavior, while V8 has this detail right.

I've added a test to verify this via a "fixup!" commit.

tcare:
@bterlson would appreciate help reviewing this.

test/language/eval-code/indirect/realm.js
```diff
@@ -0,0 +1,16 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: TODO
```
jugglinmike: Whoops. Good catch

jugglinmike:
Should be all set now, @tcare. Thanks for the review!

test/built-ins/TypedArrays/typedarray-arg-other-ctor-buffer-ctor-custom-species-proto-from-ctor-realm.js
```diff
@@ -0,0 +1,59 @@
...
+C.prototype = null;
+
+
+testWithTypedArrayConstructors(function(TA) {
+  var sample = TA === Int8Array ? sample2 : sample1;
+  var ctor = {};
+
+  sample.buffer.constructor = ctor;
+
+  ctor[Symbol.species] = C;
```
littledan: Won't setting ArrayBuffer[Symbol.species] this way fail, since that is an accessor property with no setter?

littledan:
Ping, I left some review comments a few months ago; I'm not sure if this test is accurate.

test/built-ins/TypedArrays/typedarray-arg-other-ctor-buffer-ctor-custom-species-proto-from-ctor-realm.js
littledan: Ping, any thoughts here?

PR #689, opened 2016-06-22
Add tests for script interactions

jugglinmike:
ECMA262 allows for an arbitrary number of "ScriptJob"s to run in a given realm.
Although there is no standard mechanism for enqueuing these jobs, many
implementations offer this functionality through custom APIs. In those hosts,
the semantics describing script interactions are directly observable.

In order to guarantee conformance to the specification in advance of a
standardized API, Test262 now requires that hosts provide a `$.evalScript`
function whose behavior is defined in the project's "INTERPRETING.md" file.

Use this host-provided API to ensure that implementations correctly observe the
specification text that dictates script interactions.

(In writing these tests, I noticed some gaps in coverage that are observable
from a single script execution. This patch includes a dedicated commit for
these tests that do not require `$.evalScript`.)

### DIFF
--- a/test/language/global-code/decl-func-dup.js
+++ b/test/language/global-code/decl-func-dup.js
@@ -6,5 +6,6 @@ description: >
     redeclaration outermost:
     allowed to redeclare function declaration with function declaration
 ---*/
-function f() {} function f() {}
+function f() { return 1; } function f() { return 2; }
 
+assert.sameValue(f(), 2);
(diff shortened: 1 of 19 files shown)


jugglinmike:
I've pushed up one more test case at @ajklein's request: `let` declarations
successfully shadowing configurable properties of the global object.

tcare:
Looks good.

PR #690, opened 2016-06-22
Improve assertions comparing values to NaN

leobalter:
The global isNaN is not precise at all, and Number.isNaN is an ES6 feature that makes it preferrable to use assert's sameValue for NaN values, as it handles it internally using the comparison.

---

The changes are not complete for the `test/language` folder, but that requires an extra time I'd prefer expending later. The current changes on this branch are useful to follow up on the tests for built-ins.


test/built-ins/NaN/S15.1.1.1_A1.js
```diff
@@ -2,22 +2,8 @@
...
-// CHECK#2
-if (isNaN(NaN) !== true) {
-	$ERROR('#2: NaN === Not-a-Number. Actual: ' + (NaN)); 
-}
-
-// CHECK#3
-if (isFinite(NaN) !== false) {
-	$ERROR('#3: NaN === Not-a-Finite. Actual: ' + (NaN)); 
-}
+assert(NaN !== NaN);
```
suwc: Remind me why we are not testing typeof(NaN), isNaN(NaN), isFinite(NaN), or they are being tested elsewhere?

leobalter: Those belong to the tests for typeof, isNaN and isFinite.

suwc: Thanks. #2 and #3 are accounted for. Could you help me find #1 (typeof NaN === 'number')?
I only found 'typeof Number.NaN' and 'typeof "NaN"' in language/expressions/typeof

---
In reply to: [69065687](https://github.com/tc39/test262/pull/690#discussion_r69065687) [](ancestors = 69065687)

leobalter: As noticed in #713, we already have tests for Number.NaN which is already asserted as a NaN value. So it does not really matter checking typeof for NaN and Number.NaN.

In any case, I replaced the namespaced names with their equivalent globals.

PR #692, opened 2016-06-22
Improve coverage for Math

leobalter:


test/built-ins/Math/imul/results.js
```diff
@@ -0,0 +1,61 @@
+// Copyright (C) 2016 The V8 Project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+es6id: 20.2.2.19
+esid: sec-math.imul
+description: >
+  Return results
+info: |
+  Math.fround ( x )
```
jugglinmike: This field describes the wrong section

test/built-ins/Math/prop-desc.js
```diff
@@ -0,0 +1,42 @@
...
+
+var obj = Object.defineProperty({}, "length", {
+  get: function() {
+    throw new Test262Error(
+      "Should throw a TypeError if target has no [[Construct]]"
+    );
+  }
+});
+assert.throws(TypeError, function() {
+  Reflect.construct(Math, obj);
```
jugglinmike: This assertion belongs in a separate file. Also, the details about the second parameter are features of `Reflect.construct`, not `Number`, so they aren't necessary here. I'm not sure that Reflect is necessary, either--have you considered using `new Math`? That would make the test viable for ES5 environments.

leobalter: I was trying to be super specific and precise here, but I agree that ends being a coverage for Reflect.construct.

I'll replace this using the new operator.

leobalter:
fixes are up.

jugglinmike:
Squashed and merged to `master`:

- f3bfaa6 Expand tests for Math.log1p
- 080afac Update tests for Math.{sign,sin,trunc}
- e3e7a51 Expand tests for Math.imul
- 80c1c46 Update tests for Math.hypot
- f88c935 Expand tests for Math.fround
- d6eb13b Expand tests for Math.clz32
- 440211e Update tests for Math.acosh
- 561ac9e Extend coverage for Math.abs
- 0f29f57 Update Math functions to handle distinct -0 and +0
- 13e01e4 Add tests for Math.{max,min} handling zeros
- ec3a89e Add test for the Math object

Thanks, Leo!

PR #693, opened 2016-06-23
Add tests for assignment target validation of new ES2015 forms

jugglinmike:
This re-factors some existing Sputnik tests to be more targeted and to use a
pattern that can be generalized to other forms. We could test these all day,
but I've limited myself to forms introduced in ES2015, specifically
YieldExpression and new.target. Note that SpiderMonkey incorrectly throws a
SyntaxError for these.

I thoughtlessly wrote ReferenceError tests for `yield = 1` until I realized
that such productions are not actually recognized by the grammar, so the early
errors do not apply. Instead, I've added a negative syntax test for that case.


tcare:
Looks good.

PR #698, opened 2016-06-27
Remove invalid tests

leobalter:
ConversionError and RegExpError tests are methods that existed on
something that preceeded a IE9 Standard. They do not exist in the
specs, at least since ES5.

It's not test262 responsibility to maintain these tests as these
features can be freely implemented by any runtime.


jugglinmike:
Merged to `master` at commit 059b7cd4a3de6c0d214d723d8afcea39d4537191. Nice find!

Issue #685, opened 2016-06-22
Outdated tests: built-ins/DataView/prototype/set*/range-check-after-value-conversion.js


bakkot:
All of the following tests check setting a value in a DataView performs conversion of the value before checking any bounds. But step 4 of 24.2.1.2 has been changed from ToInteger to ToIndex, which checks that its input is within [0, 2^53-1]. 'Infinity' should be replaced with '2**53-1' or similar, removed outright, or possibly moved into the corresponding index-check-before-value-conversion tests.

Invalidated tests:
[1](https://github.com/tc39/test262/tree/master/test/built-ins/DataView/prototype/setFloat32/range-check-after-value-conversion.js), [2](https://github.com/tc39/test262/tree/master/test/built-ins/DataView/prototype/setFloat64/range-check-after-value-conversion.js), [3](https://github.com/tc39/test262/tree/master/test/built-ins/DataView/prototype/setInt16/range-check-after-value-conversion.js), [4](https://github.com/tc39/test262/tree/master/test/built-ins/DataView/prototype/setInt32/range-check-after-value-conversion.js), [5](https://github.com/tc39/test262/tree/master/test/built-ins/DataView/prototype/setInt8/range-check-after-value-conversion.js), [6](https://github.com/tc39/test262/tree/master/test/built-ins/DataView/prototype/setUint16/range-check-after-value-conversion.js), [7](https://github.com/tc39/test262/tree/master/test/built-ins/DataView/prototype/setUint32/range-check-after-value-conversion.js), [8](https://github.com/tc39/test262/tree/master/test/built-ins/DataView/prototype/setUint8/range-check-after-value-conversion.js).

Ping @leobalter, @anba.

leobalter:
> 'Infinity' should be replaced with '2**53-1' or similar

You're right. The assertion using an Infinity index position is not valid.

That position would throw a RangeError before evaluating the poisoned value.

The respective `index-check-before-value-conversion.js` tests should be updated as the files you mentioned can be completely removed.

bakkot:
I'm not sure they should be completely removed. The tests with '100' are valid; those tests check that an implementation hits the poisoned value before checking that 100 is in bounds for the array.

leobalter:
The assertion with `100` is valid, the others are not adding anything to the coverage on that file. There's a file already covering type conversions.

woess:
[...]

PR #699, opened 2016-06-27
Fix range check tests for DataView#set methods

leobalter:
Closes gh-685


tcare:
Pending discussion in #685

tcare:
Looks good.

PR #701, opened 2016-06-27
Extend coverage for Section 25 - Promises

jugglinmike:
Add new tests (and new assertions within existing tests) to more completely cover the specification text for Promises.

PR #702, opened 2016-06-27
Add tests ensuring iterator is not closed

jugglinmike:
A subtle aspect of the for-of iteration protocol concerns abrupt
completions that do *not* trigger iterator closing. Although this detail
is implicit in the current structure of the specification text, some
hosts may violate the protocol by closing the iterator because later steps
*do* specify that behavior.

The V8 engine is one such host--as of this writing, it incorrectly
closes the iterator when accessing the `value` property of the iterator
result produces an abrupt completion.

Add tests verifying that the iterator protocol is not violated in this
way for abrupt completions during the semantics of for-of evaluation.

(I discovered this while writing tests for the iteration protocol as specified
by Promise.all and Promise.race see gh-701; V8 fails a handful of tests there
for the same reason.)

### DIFF
--- a/test/language/statements/for-of/iterator-next-error.js
+++ b/test/language/statements/for-of/iterator-next-error.js
@@ -5,16 +5,25 @@ es6id: 13.6.4.13 S5.d
 description: >
     If `nextResult` is an abrupt completion as per IteratorStep (ES6 7.4.5),
     return the completion.
+info: |
+  [...]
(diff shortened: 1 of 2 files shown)


littledan:
cc @GeorgNeis Do these tests correspond to your understanding of the specification?

GeorgNeis:
Yes, looking good.

Issue #697, opened 2016-06-24
Coverage: B.3.3 with nested blocks


bakkot:
As far as I can tell, test262 isn't checking anything like

```js
(function () {
  {
    let f = 2;
    {
      function f() {}
    }
  }
  return typeof f == 'undefined'; // true
})()
```

In this case there is no hoisting of f, because replacing it with `var f` would cause an early error (in particular a conflict with the `let`).

V8 and JavaScriptCore both fail this test (or at least have failed it in the recent past).

littledan:
There are a bunch more tests available at https://codereview.chromium.org/2099623003/diff/40001/test/mjsunit/es6/block-sloppy-function.js . I'm not sure which of these are more like V8 regression tests and which are general, and some of these may overlap on things that test262 already tests for.

jugglinmike:
We do have some coverage for those semantics. For example, in [annexB/language/function-code/block-decl-func-skip-early-err.js](https://github.com/tc39/test262/blob/f3bfaa692d70a636c4a2e52ef302199ca71fd52c/test/annexB/language/function-code/block-decl-func-skip-early-err.js):

```js
var init, after;

(function() {
  let f = 123;
  init = f;

  {
    function f() {  }
  }

  after = f;
}());

assert.sameValue(init, 123, 'binding is not initialized to `undefined`');
assert.sameValue(after, 123, 'value is not updated following evaluation');
```

These tests are procedurally generated, so there are actually 40 files following this pattern (files whose name contain `-skip-early-err`). I would have thought these tests were sufficient, but (as evidenced by the current behavior of V8 and JavaScript core) I would have been wrong. Although the example you've shared demonstrates the same algorithm step in Annex B, it is dependent on a completely different early error ([the one defined for Block Statements](https://tc39.github.io/ecma262/#sec-block-static-semantics-early-errors), not [the one defined for Function Definitions](https://tc39.github.io/ecma262/#sec-function-definitions-static-semantics-early-errors)).

This feature is actually dependent on three (mostly) independent circumstances:

1. the position which contains the function declaration; one of: Block,
   IfStatement (x5), or SwitchStatement (x2)
2. the declaration algorithm defining the hoisting extension; one of:
   GlobalDeclarationInstantiation, FunctionDeclarationInstantiation, or
   EvalDeclarationInstantiation
3. the condition that defines an early error which prevents hoisting (maybe
   called, "the disqualifying context"); one of: ScriptBody, FunctionBody,
[...]

PR #704, opened 2016-06-28
Annex b skip fib

jugglinmike:
There's extra noise here because the test generation tool has changed since the
"annex B function-in-block" tests were first introduced. I've documented this
in a standalone commit, and I'm happy to submit that as a separate pull request
(blocking this one) if my reviewers would prefer that.

This should resolve gh-697.

> The "variable-like" function hoisting semantics defined in Annex B extension
> B.3.3 is only applied if "[...] replacing the FunctionDeclaration f with a
> VariableStatement that has F as a BindingIdentifier would not produce any Early
> Errors [...]". Test262 previously included tests for this condition when the
> disqualifying early error originated from the ScriptBody and FunctionBody
> productions.
>
> Add test cases to assert the behavior when it is disqualified by all other
> relevant early errors: Block statements, `for` statements, `for-of` statements,
> `for-in` statements, and Switch statements.

### DIFF
--- a/src/annex-b-fns/eval-func-no-skip-try.case
+++ b/src/annex-b-fns/eval-func-no-skip-try.case
@@ -13,6 +13,15 @@ info: |
         has F as a BindingIdentifier would not produce any Early Errors for
         body, then
     [...]
+
+    B.3.5 VariableStatements in Catch Blocks
+
+    [...]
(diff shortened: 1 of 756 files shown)


jugglinmike:
This patch now includes tests for TryStatement, as discussed in gh-697.

tcare:
Looks good.

Issue #703, opened 2016-06-28
Getter tests doesn't cover full PropertyName


avdg:
```
MethodDefinition[Yield]:
  PropertyName[?Yield] `(` StrictFormalParameters[~Yield]`)` `{` FunctionBody[~Yield] `}`
  GeneratorMethod[?Yield]
  `get` PropertyName[?Yield] `(` `)` `{` FunctionBody[~Yield] `}`
  `set` PropertyName[?Yield] `(` PropertySetParameterList `)` `{` FunctionBody[~Yield] `}`

PropertyName[Yield]:
  LiteralPropertyName
  ComputedPropertyName[?Yield]

LiteralPropertyName:
  IdentifierName
  StringLiteral
  NumericLiteral
```

There seems to be currently no coverage for `get` followed by StringLiteral or NumericLiteral. At least not in https://github.com/tc39/test262/tree/master/test/language/expressions/object

avdg:
(same could be applied for `set`)

jugglinmike:
It looks that way to me, too. We have some coverage for NumericLiteral in other places (e.g. [`built-ins/RegExp/prototype/Symbol.replace/result-get-matched-err.js`](https://github.com/tc39/test262/blob/f3bfaa692d70a636c4a2e52ef302199ca71fd52c/test/built-ins/RegExp/prototype/Symbol.replace/result-get-matched-err.js)), but those are neither direct nor exhaustive. I will have a patch for this ready by the end of today.

Thanks a lot for reporting coverage holes like this! It's encouraging to receive feedback from someone working on a parser :)

avdg:
Yeah, I just found some tests like https://github.com/tc39/test262/blob/master/test/built-ins/RegExp/prototype/Symbol.match/g-get-result-err.js

avdg:
Yeah, I found it a bit weird that chrome doesn't seem to support it, while at the same time firefox does have seem to have support for that.

avdg:
Chrome seems to have catched up by now.

PR #706, opened 2016-06-28
Add tests for accessor names in object initializer

jugglinmike:
This should resolve gh-703. What do you think, @avdg?


avdg:
Looking generally very good, although there are lots of 0 and 1 used in the numbers.

avdg:
Hmmm... classes should also be covered as well I guess

avdg:
And set seems to accept a BindingElement as parameter.

avdg:
(there are some tests though at https://github.com/tc39/test262/tree/master/test/language/computed-property-names/class/accessor)

jugglinmike:
Superseded by gh-843. Thanks for keeping me honest, @avdg.

avdg:
Np

PR #707, opened 2016-06-28
Add bit-precision tests for TypedArray#copyWithin

leobalter:
I'm not 100% convinced the nan bit encoding is preserved w/ copyWithin, but it's good to check.

@littledan would you mind reviewing this one?

### DIFF
--- a/test/built-ins/TypedArray/prototype/copyWithin/bit-precision.js
+++ b/test/built-ins/TypedArray/prototype/copyWithin/bit-precision.js
@@ -16,14 +16,13 @@ info: |
 includes: [nans.js, compareArray.js, testTypedArray.js]
 ---*/
 
-var distinctNaNsDouble = new Array(distinctNaNs.length * 2);
-
-distinctNaNs.forEach(function(v, i) {
-  distinctNaNsDouble[i] = v;
-});
-
 function body(FloatArray) {
-  var subject = new FloatArray(distinctNaNsDouble);
+  var subject = new FloatArray(distinctNaNs.length * 2);
+
+  distinctNaNs.forEach(function(v, i) {
+    subject[i] = v;
+  });
+
   var originalBytes, copiedBytes;
   var length = distinctNaNs.length * FloatArray.BYTES_PER_ELEMENT;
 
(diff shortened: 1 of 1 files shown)


littledan:
This test seems excessively complicated. Why not just allocate a TypedArray 2x the length of distinctNaNs, write them into the first n elements of the array, then copyWithin to the second half, and compare the bytes of the two lengths? I don't really know what having two TypedArrays is trying to get at; it seems to just conflate testing `TypedArray.prototype.copyWithin` with testing that scalar NaNs are preserved in Arrays or that SetValueInBuffer is stable.

leobalter:
> ... copyWithin to the second half, and compare the bytes of the two lengths?

The compare part would not complex as well instead of just using compareArray. It does not add clarity for the tests. I can tell this as I tried at least 4 different approaches for this test files.

I agree the preparation for this test is complicated too, but it was a subjective choice at that moment.

> I don't really know what having two TypedArrays is trying to get at

I use one instance (`originalBytes = new Uint8Array(originalValues.buffer);`) to find the original bit encodings of each NaN, they will be used as the values I expect for the result of copyWithin. Then, I use another TypedArray instance to find the copied bytes. (`copiedBytes = new Uint8Array( subject.buffer, ...`)

I'll try to rewrite these tests avoiding two FloatArrays.

leobalter:
d719088 is up reusing the same FloatArray instance.

test/built-ins/TypedArray/prototype/copyWithin/bit-precision.js
```diff
@@ -0,0 +1,45 @@
...
+---*/
+
+var distinctNaNsDouble = new Array(distinctNaNs.length * 2);
+
+distinctNaNs.forEach(function(v, i) {
+  distinctNaNsDouble[i] = v;
+});
+
+function body(FloatArray) {
+  var subject = new FloatArray(distinctNaNsDouble);
```
littledan: Possible further simplification: Rather than make a distinctNaNsDouble array, what if you allocated a FloatArray of the appropriate size and wrote each distinct NaN into the first half?

leobalter: Ok, I've changed this at 976ef41. That's one less object, thanks

jugglinmike:
Squashed and merged to `master` at commit d190e700f34cc2839f80da0beec33979154666eb. Thanks!

Issue #677, opened 2016-06-18
Outdated RegExp tests check for exceptions being thrown, when no exception is thrown


littledan:
The semantics were changed in https://github.com/tc39/ecma262/commit/1a2ca97c1949ce0d0ee64dbc40da01574f0b1cd2 . It'd be good to have tests that check for the `undefined` value for `RegExp.prototype` and throwing on other values.

```
built-ins/RegExp/prototype/source/15.10.7.1-1
built-ins/RegExp/prototype/global/15.10.7.2-1
built-ins/RegExp/prototype/ignoreCase/15.10.7.3-1
built-ins/RegExp/prototype/multiline/15.10.7.4-1
```

jugglinmike:
A correction is on the way!

PR #711, opened 2016-06-29
Improve coverage for section 21: RegExp

jugglinmike:
This changeset increases coverage for section 21, including addressing the
recent change to the specification reported in gh-677. It also improves test
structure.

Resolves gh-677.


littledan:
cc @tcare

tcare:
Looks good. No interference with tc39/ecma262#627

PR #712, opened 2016-06-29
Improve coverage for section 22: String

jugglinmike:
This changeset increases coverage for section 22, specifically "21.1 String
Objects".

### DIFF
--- a/test/built-ins/String/length.js
+++ b/test/built-ins/String/length.js
@@ -1,0 +1,29 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-properties-of-string-instances-length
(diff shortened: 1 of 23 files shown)


test/built-ins/String/length.js
```diff
@@ -0,0 +1,29 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-properties-of-string-instances-length
+es6id: 21.1.1.1
+description: The "length" property of String objects
```
tcare: I don't see any direct reference to length in 21.1.1.1. Perhaps we could reference:

21.1.4.1 length
or
9.4.3.3 StringCreate 

Which both refer to the property descriptor of length on a String instance.

tcare:
Looks good

PR #713, opened 2016-06-30
Use global values on typeof tests for Number values

leobalter:
This is a pure miscellaneous change to use global values of NaN and Infinity instead of their namespaced equivalents.

Ref https://github.com/tc39/test262/pull/690/files/9ae0567a50b4367cf0c692da947ed3613e2e35c3#r69049894

cc @suwc

### DIFF
--- a/test/language/expressions/typeof/number.js
+++ b/test/language/expressions/typeof/number.js
@@ -11,29 +11,29 @@ description: typeof (number value) === "number"
 assert.sameValue(
   typeof 1,
   "number",
-  '#1: typeof 1 === "number". Actual: ' + (typeof 1)
+  'typeof 1 === "number". Actual: ' + (typeof 1)
 );
 
 assert.sameValue(
-  typeof Number.NaN,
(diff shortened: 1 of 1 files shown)


jugglinmike:
This looks technically correct to me, but I have to admit that I don't
understand the motivation. Is this new version of the test preferable in some
way?

leobalter:
This was quick change I don't see as really necessary, but I opened this as it might solve the concerns raised.

While it does not improves coverage, the change makes the assertions slightly more consistent with their descriptions.

Maybe we should set a canonical name for these values, I also believe it's not worth changing the code everywhere unless in some free time if someone is available.

leobalter:
@tcare do you mind to review this one?

I'm not personally attached to this PR (or any other), if it's discarded the test will remain valid.

tcare:
Looks good to me. I don't feel strongly either way about going back and changing these. In this particular case I agree it makes more sense given the descriptions.

PR #716, opened 2016-07-01
Update tests for Proxy.[[SetPrototypeOf]]

leobalter:
Expand coverage for proxies' [[SetPrototypeOf]] and update tests to
verify that target's [[IsExtensible]] is not called anymore if trap
returns a falsy value.

Ref tc39/ecma262#331


test/built-ins/Proxy/setPrototypeOf/toboolean-trap-result-false.js
```diff
@@ -0,0 +1,65 @@
...
+assert.sameValue(result, false, "null");
+assert.sameValue(called, 0, "null - isExtensible is not called");
+
+result = Reflect.setPrototypeOf(p, { attr: undefined });
+assert.sameValue(result, false, "undefined");
+assert.sameValue(called, 0, "undefined - isExtensible is not called");
+
+result = Reflect.setPrototypeOf(p, { attr: false });
+assert.sameValue(result, false, "false");
+assert.sameValue(called, 0, "false - isExtensible is not called");
```
jugglinmike: I think this is unnecessary--the behavior for the `false` value has already been asserted above.

jugglinmike:
This patch is a little inconsistent with its use of `Object.setPrototypeOf` and `Reflect.setPrototypeOf`. Some tests use both methods. I think either one works for these tests, so we ought to choose one and use it consistently. Can you think of a reason why one might be more appropriate than the other?

test/built-ins/Proxy/setPrototypeOf/toboolean-trap-result-false.js
leobalter: oops, you're right. I got caught in a copy & paste scene.

jugglinmike:
The try-catch block shouldn't be necessary because we should always either
expect an exception (using `assert.throws`) or expect a normal completion
(allowing uncaught exceptions to trigger test failure). Otherwise that sounds
fine.

test/built-ins/Proxy/setPrototypeOf/internals-call-order.js
```diff
@@ -0,0 +1,48 @@
...
+Object.preventExtensions(target);
+
+var proxy = new Proxy(target, {
+  setPrototypeOf: function() {
+    calls.push("proxy.[[setPrototypeOf]]");
+    return true;
+  }
+});
+
+Reflect.setPrototypeOf(proxy, proto);
```
leobalter: Now I remember why I used Reflect instead of Object. The assert.throws using Object.setPrototypeOf might catch a false positive when the Proxy internals throws. See line 35 as necessary configuration step for the assertions which would have caused a TypeError if absent.

In anyway, I need to include an assertion for the result here.

leobalter: the `non-extensible-target-*` would catch any possible error cause a false positive here. Even though, I'd like to have these tests as atomic and not dependent on the results from other files.

leobalter:
fixes are up

tcare:
Looks good.

test/built-ins/Proxy/setPrototypeOf/internals-call-order.js
```diff
@@ -41,7 +41,7 @@ var proxy = new Proxy(target, {
   }
 });
 
-Reflect.setPrototypeOf(proxy, proto);
+assert.sameValue(Reflect.setPrototypeOf(proxy, proto), false);
```
hferreiro: shouldn't this return 'true' given that the prototype argument is the original one?

PR #717, opened 2016-07-01
Improve coverage for section 14: Functions and Classes

jugglinmike:
This changeset increases test coverage for the semantics defined in ECMA262 section 14.


tcare:
Looks good.

PR #718, opened 2016-07-01
Improve coverage for class "name" inference

jugglinmike:
This is technically part of section 14. I recently submitted a patch for
coverage in that section (see gh-717), but because this change involves the
generation tool, I thought it would be easier on my reviewers if it were
submitted separately.

As usual, this change set factors all the procedurally-generated changes into
a dedicated commit.


tcare:
Looks good.

PR #719, opened 2016-07-05
Improve coverage for section 15: Scripts and Modules

jugglinmike:
This changeset increases test coverage for the semantics defined in [ECMA262 section 15](https://tc39.github.io/ecma262/#sec-ecmascript-language-scripts-and-modules).


tcare:
Looks good.

PR #720, opened 2016-07-05
Add test for subclassing bound functions

jugglinmike:
The semantics under test have been incorrectly implemented by the SpiderMonkey engine.


tcare:
Looks good.

PR #721, opened 2016-07-05
Add tests for abrupt completions during JSON.parse

jugglinmike:
Due to the use of the IsArray abstract operation, observing many of the
targeted semantics requires the use of a Proxy exotic object. Mark the
tests that require this unrelated ES2015 feature using a dedicated
"features" flag.

### DIFF
--- a/test/built-ins/JSON/parse/reviver-call-err.js
+++ b/test/built-ins/JSON/parse/reviver-call-err.js
@@ -6,6 +6,9 @@ es6id: 24.3.1.1
 description: >
   Abrupt completion when reviver function returns an abrupt completion
 info: |
+  JSON.parse ( text [ , reviver ] )
+
+  [...]
   7. If IsCallable(reviver) is true, then
      [...]
      e. Return ? InternalizeJSONProperty(root, rootName).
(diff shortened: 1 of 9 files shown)


test/built-ins/JSON/parse/reviver-call-err.js
```diff
@@ -0,0 +1,23 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-internalizejsonproperty
+es6id: 24.3.1.1
+description: >
+  Abrupt completion when reviver function returns an abrupt completion
+info: |
+  7. If IsCallable(reviver) is true, then
```
tcare: Would you mind putting something along the lines of:
JSON.parse ( text [ , reviver ] )
Before the parts that quote that section? It wasn't immediately obvious to me that they were from the section above.

jugglinmike:
> Would you mind putting something along the lines of:  
> JSON.parse ( text [ , reviver ] )  
> Before the parts that quote that section? It wasn't immediately obvious to me that they were from the section above.

Sure! I just pushed up a "fixup" commit which does exactly that.

tcare:
Perfect! Thanks. LGTM

PR #709, opened 2016-06-29
Ensure U+180E is no longer considered whitespace

mathiasbynens:
Ref. https://hashseed.blogspot.com/2014/08/in-ecma-262-5.html
Ref. https://github.com/tc39/ecma262/pull/300#issuecomment-181376767
Ref. https://github.com/mathiasbynens/regexpu-core/commit/9b10d2a597d4e56b236cae0f9aa2f21a9c6d1122

PR #724, opened 2016-07-06
Json whitespace

jugglinmike:
@mathiasbynens I planned to re-factor (and rename) the test for `JSON.parse` as
a follow-up to gh-709. As I did so, I came to believe that it wasn't necessary
to remove U+180E from that test in the first place. The second commit
re-introduces that assertion and includes an explanation--can you let me know
what you think?


test/built-ins/JSON/parse/invalid-whitespace.js
```diff
@@ -0,0 +1,73 @@
+// Copyright (c) 2012 Ecma International.  All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+es5id: 15.12.1.1-0-7
```
mathiasbynens: `esid: sec-json.parse`

mathiasbynens:
You’re right! Sorry for incorrectly modifying that test.

From ECMA-404:

> [In JSON,] Insignificant whitespace is allowed before or after any token. The whitespace characters are: character tabulation (U+0009), line feed (U+000A), carriage return (U+000D), and space (U+0020)

jugglinmike:
No problem at all. Thanks for the review!

tcare:
Looks good.

PR #725, opened 2016-07-07
Add "how to run" on front page

smikes:
As a returning contributor, I see that the old test262.py runner is officially deprecated -- good!

But I don't know the "official" way to run the test suite, using my default installed node.

Here's a stab at adding that information to the front page README.md.  Please correct me if I'm wrong.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,3 +11,8 @@ Test262 tests conformance to the continually maintained draft future ECMAScript 
 ### Running Test262
 
 > See [INTERPRETING.md](./INTERPRETING.md)
+
+There are a number of volunteer-maintained projects that may be used to execute Test262 in various ECMAScript hosts:
+
+- https://github.com/bterlson/test262-harness (platform: Node.js)
+- https://github.com/test262-utils/test262-harness-py (platform: Python)


jugglinmike:
My understanding of TC-39's decision to remove the Python runner is that the
committee wishes to avoid taking on this responsibility altogether, regardless
of the platform on which it is implemented. In that sense, @bterlson's
Node.js-based runner is no more official or preferable than the Python runner
(which is now maintained in a standalone repository on GitHub:
[test262-utils/test262-harness-py](https://github.com/test262-utils/test262-harness-py)).

The `INTERPRETING.md` is not intended as supplemental material; it is our
attempt to document "how to run Test262" exhaustively and in a way that is in
keeping with TC-39's interests.

That said, I was not present at the most recent meeting where this was
discussed. I'd welcome corrections from @bterlson or @leobalter on this!

If that's accurate though, I think the information is still useful, but we'll
just want to present it slightly differently. We can offer a list of links and
avoid implementation-specific instructions, instead allowing the maintainers of
those projects to document their setup procedures. What do you think of
something like this:

```diff
 ### Running Test262

 > See [INTERPRETING.md](./INTERPRETING.md)
+
+ There are a number of volunteer-maintained projects that may be used to
+ execute Test262 in various ECMAScript hosts:
+
+ - https://github.com/bterlson/test262-harness (platform: Node.js)
+ - https://github.com/test262-utils/test262-harness-py (platform: Python)
```

leobalter:
there's no current official test runner. Both test262-harness and test262-harness-py are volunteeer maintained projects, as I see now Mike has already mentioned on his comment.

I believe it's useful to mention them in our __readme__ file, but not document it's usage, they should have an appropriate documentation.

The Interpreting documentation is useful to implement a runner. I believe someone that only wants to run the tests and see the results will be fine just knowing about the available ones.

tcare:
I agree with @leobalter and @jugglinmike - we can link to the runners for now.

smikes:
Looks good to me, will update the PR

leobalter:
LGTM

PR #726, opened 2016-07-07
Improve coverage for section 9

jugglinmike:
This changeset increases test coverage for the semantics defined in [ECMA262 section 9](https://tc39.github.io/ecma262/#sec-ordinary-and-exotic-objects-behaviours).


tcare:
@aneeshdk would you mind taking a look at: Assert creation of 'arguments' object … f96285b

tcare:
@suwc would you mind taking a look at: Add tests for NewTarget override of bound function 1488100

tcare:
Looks good.

PR #727, opened 2016-07-09
Improve coverage for section 20

jugglinmike:
This changeset increases test coverage for the semantics defined in [ECMA262
section 20](https://tc39.github.io/ecma262/#sec-numbers-and-dates).

There are additional methods of the Date.prototype methods that still need
coverage, but I felt this patch was getting a bit long in the tooth as it is.


tcare:
Date.UTC looks good. We'll do the no arguments case in a separate PR for easy linking with the spec related PRs.

leobalter:
LGTM

PR #730, opened 2016-07-17
Add Test262Error.name to check the kind of error easily by just looking up `error.name`

Constellation:
Add Test262Error.name property. This addition aligns Test262Error more to usual Error instance.
For example, JavaScriptCore engine[[1]] uses the uncaught error.name to check the expected error is thrown.

[1]: https://bugs.webkit.org/show_bug.cgi?id=159862

### DIFF
--- a/harness/sta.js
+++ b/harness/sta.js
@@ -6,11 +6,12 @@ var EarlyErrorRePat = "^((?!" + NotEarlyErrorString + ").)*$";
 var NotEarlyError = new Error(NotEarlyErrorString);
 
 function Test262Error(message) {
+    this.name = "Test262Error";
     this.message = message || "";
 }
 
 Test262Error.prototype.toString = function () {
-    return "Test262Error: " + this.message;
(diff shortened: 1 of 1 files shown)


jugglinmike:
I don't believe this change is necessary. From the project's [test interpretation guidelines](https://github.com/tc39/test262/blob/dbf251586e280cebc716b96f1a8a089ffa63a54f/INTERPRETING.md):

> `negative`
>
> These tests are expected to generate an uncaught exception. The value of this
> attribute is the name of the constructor of the expected error. If a test
> configured with the `negative` attribute completes without throwing an
> exception, or if the name of the thrown exception's constructor does not
> match the specified constructor name, the test must be interpreted as
> "failing."

It should be possible to use `errorInstance.constructor.name` today, without
any change in Test262. Would that approach suit your use case?

Constellation:
Aha! That sounds nice. Thanks :)

PR #733, opened 2016-07-26
Tests for trailing commas in function arg lists

jeffmo:
Adds tests for the proposal as described here:
http://jeffmo.github.io/es-trailing-function-commas/

cc @bterlson @ljharb

### DIFF
--- a/test/language/expressions/arrow-function/params-trailing-comma-length.js
+++ b/test/language/expressions/arrow-function/params-trailing-comma-length.js
@@ -1,0 +1,21 @@
+// Copyright (C) 2016 Jeff Morrison. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+description: >
+  Check that trailing commas don't affect the .length property of arrow
+  functions.
+info: http://jeffmo.github.io/es-trailing-function-commas/
+author: Jeff Morrison <lbljeffmo@gmail.com>
+---*/
+
+assert.sameValue(
+  ((a,) => {}).length,
+  1,
+  "Arrow function with 1 arg + trailing comma reports incorrect .length!"
+);
+
+assert.sameValue(
+  ((a,b,) => {}).length,
+  2,
+  "Arrow function with 2 args + trailing comma reports incorrect .length!"
+);
--- a/test/language/expressions/generators/params-trailing-comma-length.js
+++ b/test/language/expressions/generators/params-trailing-comma-length.js
@@ -1,0 +1,21 @@
+// Copyright (C) 2016 Jeff Morrison. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+description: >
+  Check that trailing commas don't affect the .length property of generator
+  function expressions.
+info: http://jeffmo.github.io/es-trailing-function-commas/
+author: Jeff Morrison <lbljeffmo@gmail.com>
+---*/
+
+assert.sameValue(
+  (function*(a,) {}).length,
+  1,
+  "Generator expression with 1 arg + trailing comma reports incorrect .length!"
+);
+
+assert.sameValue(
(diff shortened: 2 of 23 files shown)


caitp:
Looks like there are some good tests missing

- negative tests for `,` following rest parameters
- testing arrow functions, generators, methods, etc

it also might be a bit confusing to mix the Arguments trailing commas and FormalParameters trailing commas tests in the same file, but I don't have that much of an opinion on that

leobalter:
we also need to structure these tests under the existing folders for the functions forms including those listed by @caitp

jeffmo:
> Looks like there are some good tests missing

Good catch! Will add

> we also need to structure these tests under the existing folders for the functions forms

Can you point me to which directories this would be? I wasn't sure if there was a better place for these (or if I named the test file in an acceptable way?)

leobalter:
sure!

- `test/language/expressions/{function,generators,arrow-function}` 
- `test/language/rest-parameters`
- `test/language/statements/{function,generators}` 
- `test/language/expressions/object/method-definition`
- `test/language/statements/class/definition`
- `test/language/{expressions,statements}/async-function`  (see #479)
- `test/built-ins/{Function,GeneratorFunction,AsyncFunction}`

It looks like a lot of tests, but they might be easier if you use the test generation tool available within the repo.

tcare:
Would you mind adding some tests making sure the arguments object is unaffected?

test/language/expressions/generators/params-trailing-comma-length.js
```diff
@@ -0,0 +1,19 @@
+// Copyright (C) 2016 Jeff Morrison. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+description: Check that trailing commas are permitted in function argument lists
```
tcare: generator function*?

tcare: Similar for other descriptions, which should be a bit more descriptive about the individual test's high level purpose

jeffmo: Ok. I've updated the generator descriptions to include "generator".

What level of clarity are you looking for for other descriptions though? The current descriptions seem pretty clear to me, but happy to add additional context where necessary.

test/language/expressions/arrow-function/params-trailing-comma-length.js
```diff
@@ -0,0 +1,19 @@
...
+/*---
+description: Check that trailing commas are permitted in function argument lists
+info: http://jeffmo.github.io/es-trailing-function-commas/
+author: Jeff Morrison <lbljeffmo@gmail.com>
+---*/
+
+assert.sameValue(
+  ((a,) => {}).length,
+  1,
+  "Arrow function with 1 arg + trailing comma reports incorrect .length1"
```
jeffmo: "length1" (typo -- willfix)

test/language/expressions/generators/params-trailing-comma-length.js
tcare: Sure, it doesn't have to be as detailed as Mike's tests for example, but something like https://github.com/tc39/test262/pull/735 is a good example.

In this particular case documenting the type of function and/or why the test is interesting (e.g. rest) would be helpful.

jeffmo:
Ok, I've gone through the descriptions and tried to be more specific. Let me know what you think

tcare:
Looks good! Thanks for the update. Would @caitp or @leobalter mind doing a second signoff?

caitp:
Rubberstamp --- I think there are good tests here, which are not tested for all the variants. I am a fan of exhaustive tests, but it is a lot of effort I guess =)

tcare:
I'll merge this for now - @jeffmo feel free to add more tests if you agree with @caitp :)

PR #735, opened 2016-07-27
Add tests for for-in initializers

bakkot:
These were re-introduced in sloppy mode per
https://github.com/tc39/ecma262/pull/614

These could use a review for filenames, style, frontmatter, etc.

### DIFF
--- a/test/annexB/language/statements/for-in/bare-initializer.js
+++ b/test/annexB/language/statements/for-in/bare-initializer.js
@@ -1,0 +1,12 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-initializers-in-forin-statement-heads
+description: >
+    for-in heads prohibit AssignmentExpressions
+negative: SyntaxError
+---*/
+var a;
+throw NotEarlyError;
+for (a = 0 in {});
+
--- a/test/annexB/language/statements/for-in/nonstrict-initializer.js
+++ b/test/annexB/language/statements/for-in/nonstrict-initializer.js
@@ -1,0 +1,39 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-initializers-in-forin-statement-heads
+description: >
+    for-in initializers in nonstrict mode
+flags: [noStrict]
+---*/
+(function() {
+  var effects = 0;
+  for (var a = ++effects in {});
+  assert.sameValue(effects, 1);
+})();
+
+
+(function() {
+  var stored;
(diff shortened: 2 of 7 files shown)


tcare:
Looks good. @leobalter @jugglinmike for a second opinion?

test/annexB/language/statements/for-in/bare-initializer.js
```diff
@@ -0,0 +1,12 @@
+// Copyright (C) 2011 the V8 project authors. All rights reserved.
```
leobalter: 2016

leobalter:
I would like to wait for @jugglinmike's feedback here, as he's got more experience with the tests for for loops

test/annexB/language/statements/for-in/bare-initializer.js
bakkot: Thanks, fixed.

test/annexB/language/statements/for-in/nonstrict-initializer.js
```diff
@@ -0,0 +1,28 @@
...
+(function() {
+  var effects = 0;
+  for (var a = ++effects in {});
+  assert.sameValue(effects, 1);
+})();
+
+
+(function() {
+  var stored;
+  for (var a = 0 in (stored = a, {}));
```
jugglinmike: The For-In statement supports an Expression here, so the parenthesis aren't
required. Removing them might make this a little more useful as a syntax
test--ensuring that there aren't inconsistent grammar restrictions in the
presence of the initializer.

jugglinmike:
Would you mind adding a test where iteration actually occurs? That will let us
assert that initialization happens exactly once, prior to assignment during
body evaluation.

test/annexB/language/statements/for-in/nonstrict-initializer.js
bakkot: Done.

bakkot:
@jugglinmike, done; thanks for the feedback.

test/annexB/language/statements/for-in/nonstrict-initializer.js
```diff
@@ -0,0 +1,39 @@
...
+(function() {
+  var effects = 0;
+  var iterations = 0;
+  var stored;
+  for (var a = (++effects, -1) in stored = a, {a: 0, b: 1, c: 2}) {
+    ++iterations;
+  }
+  assert.sameValue(stored, -1);
+  assert.sameValue(effects, 1);
+  assert.sameValue(iterations, 3);
```
leobalter: I like these tests, but I believe we can improve the assertions providing a message to identify each one. This would be helpful if any fails.

leobalter:
just one comment above on assertion messages, besides that, it LGTM

bakkot:
@leobalter, added some messages, thanks.

leobalter:
Thanks!

PR #734, opened 2016-07-27
Add DateTimeFormat.prototype.formatToParts tests

zbraniecki:
This is a set of tests for https://github.com/tc39/ecma402/pull/64 which we'd like to land in the spec soon.

I'd like to get the test262 tests for this change ready to land as part of the getting the proposal to stage 4.

I'm not sure how much more testing is required since most of the feature is tested in `DateTimeFormat.prototype.format` tests, and this is just a different representation of the output.

I'd like to get the PR to be ready to land once we land the feature in the spec, so would like to get it reviewed now.

@leobalter, @jugglinmike - can you review this?

PR #736, opened 2016-07-28
NumberFormat.prototype.formatToParts tests

stasm:
This is similar to #734 but for `NumberFormat.prototype.formatToParts`.  I think I could use some help deciding which parts of the spec should be covered. Thanks!

### DIFF
--- a/test/intl402/NumberFormat/prototype/formatToParts/formatToParts.js
+++ b/test/intl402/NumberFormat/prototype/formatToParts/formatToParts.js
@@ -1,0 +1,17 @@
+// Copyright 2016 Mozilla Corporation. All rights reserved.
+// This code is governed by the license found in the LICENSE file.
+
+/*---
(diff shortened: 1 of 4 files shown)


leobalter:
would you mind providing a link to the proposal/spec?

stasm:
@leobalter https://github.com/tc39/ecma402/pull/79 is the best I have. I've seen your edits in #734 -- I'm assuming I need to do something similar here?

leobalter:
yes, please! I couldn't tell if they just require a repetition for the tests on DateTimeFormat so that's why I asked the specs first. It's good to have a check for every part we can observe, not only the successful cases.

zbraniecki:
Since @stasm is on PTO, I took over and updated the tests in https://github.com/tc39/test262/pull/744

leobalter:
I'm closing this in favor of #744

PR #739, opened 2016-07-28
add PropertyDefinitionList evaluation order test

michaelficarra:
This test is being added because the committee is considering changing this evaluation order (as discussed at the May 2016 Munich meeting). The consequences of this spec change will be more clear as a test change than a simple test addition.

/cc @leobalter

### DIFF
--- a/test/language/expressions/object/computed-property-evaluation-order.js
+++ b/test/language/expressions/object/computed-property-evaluation-order.js
@@ -1,0 +1,26 @@
+// Copyright (C) 2016 Michael Ficarra. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-object-initializer-runtime-semantics-propertydefinitionevaluation
+description: >
+    Evaluation of PropertyDefinitionList occurs in order, and each
+    PropertyDefinition's PropertyName is evaluated before its
+    AssignmentExpression.
+---*/
+
+var counter = 0;
+var o = {
+  [++counter]: ++counter,
+  [++counter]: ++counter,
+  [++counter]: ++counter,
+}
+
+var keys = Object.keys(o);
+
+assert.sameValue(keys.length, 3, '3 PropertyDefinitions should result in 3 properties');
+assert.sameValue(keys[0], '1');
+assert.sameValue(o[keys[0]], 2);
+assert.sameValue(keys[1], '3');
+assert.sameValue(o[keys[1]], 4);
+assert.sameValue(keys[2], '5');
+assert.sameValue(o[keys[2]], 6);


test/language/expressions/object/computed-property-evaluation-order.js
```diff
@@ -0,0 +1,24 @@
+/*---
```
leobalter: This is missing the copyright section.

some equivalent comment is necessary:

```js
// Copyright (C) 2016 Michael Ficarra. All rights reserved.
```

leobalter:
missing a small detail, the tests LGTM.

michaelficarra:
Added the copyright comment.

leobalter:
LGTM.

cc @tcare for double sign off

test/language/expressions/object/computed-property-evaluation-order.js
```diff
@@ -0,0 +1,26 @@
...
+---*/
+
+var counter = 0;
+var o = {
+  [++counter]: ++counter,
+  [++counter]: ++counter,
+  [++counter]: ++counter,
+}
+
+var keys = Object.keys(o).sort();
```
leobalter: I believe `.sort()` is not necessary here, this is a detail that won't break the tests and can be fixed during merge.

michaelficarra: I wanted it to work in an implementation that doesn't necessarily implement `Object.keys` ordering properly. This is not an `Object.keys` test.

leobalter: We have to assume Object.keys is acting according to the specs, which the order is defined.

If this built-in becomes a problem, I believe pure assertions over the expected properties will solve this in a better fashion:

```js
assert.sameValue(o[1], 2);
assert.sameValue(o[3], 4);
assert.sameValue(o[5], 6);
```

If necessary, Object.keys could still be used to compare the assigned keys, which I believe is overkill, but still a valid test:

```js
assert(compareArray(Object.keys(o), [1, 3, 5]));
```

michaelficarra:
@tcare What needs to be done here? I'm okay with either implementation. They're both correct, and I believe the current implementation is more readable because it is more explicit. Let me know if I need to change it though.

tcare:
I spoke with @leobalter and he'd prefer you avoid .sort() before merging, so I'll wait for his final signoff. Either of his two suggestions are fine with me.

leobalter:
As I also spoke with @jugglinmike and we agreed there's no need to use .sort here. In terms of implementation dependent definitions, .sort is far more susceptible to a different reordering. `Object.keys` API has a pretty defined spec, any different reordering would be an error which is not a responsibility for this test. This is indeed not a test for Object.keys.

michaelficarra:
I've removed the `.sort()` call.

leobalter:
Thank you.

PR #741, opened 2016-08-04
Fix incorrect value retrieving in module-code test

Constellation:
Fix language/module-code/namespace/internals/get-own-property-str-found-init.js.


leobalter:
Thanks!

Constellation:
Thanks for super quick response!

HomerOnSkype:
Good Show , Keep Up the great work !

PR #747, opened 2016-08-19
islamicc should be canonicalized to islamic-civil

### LINKED ISSUE #743: intl402: Intl/getCanonicalLocales/main.js - islamicc is deprecated and need to be canonicalized to ismalic-civil?
intl402/Intl/getCanonicalLocales/main.js has the following: 

```
assertArray(gCL(['ar-ma-u-ca-islamicc']), ['ar-MA-u-ca-islamicc']);
```

v8's implementation (that uses two ICU APIs in series: uloc_fromLanguage() and uloc_toLanguage()) replacens 'islamicc' (deprecated in LDML) with 'islamic-civil' (its preferred form).  See 
http://www.unicode.org/repos/cldr/tags/latest/common/bcp47/calendar.xml

 Ecma 402 6.2.3 (
https://tc39.github.io/ecma402/#sec-canonicalizelanguagetag ) has the following:

> It takes the steps specified in RFC 5646 section 4.5, or successor, to bring the language tag into 
> canonical form, and to regularize the case of the subtags, but does not take the steps to bring a 
> language tag into "extlang form" and to reorder variant subtags.

RFC 5646 section 4.5 has the following:

>  point 2.  Redundant or grandfathered tags are replaced by their 'Preferred-
>       Value', if there is one.
> 
>  point 3.  Subtags are replaced by their 'Preferred-Value', if there is one.
>    For extlangs, the original primary language subtag is also
>    replaced if there is a primary language subtag in the 'Preferred-
>    Value'.

RFC 6067 for the '-u-' extension does not say anything special about deprecated vs preferred. 

My interpretation is that '-u-ca-islamicc' has to be replaced by '-u-ca-islamic-civil'  (btw, this is spun off from https://bugs.chromium.org/p/v8/issues/detail?id=5012 )

@zbraniecki ,  what would you say?

jungshik:
Fix #743 

/cc @zbraniecki @littledan @caridy

### DIFF
--- a/test/intl402/Intl/getCanonicalLocales/main.js
+++ b/test/intl402/Intl/getCanonicalLocales/main.js
@@ -29,6 +29,4 @@ assertArray(gCL({'a': 0}), []);
 
 assertArray(gCL({}), []);
 
-assertArray(gCL(['ar-ma-u-ca-islamicc']), ['ar-MA-u-ca-islamicc']);
-
 assertArray(gCL(['th-th-u-nu-thai']), ['th-TH-u-nu-thai']);
--- a/test/intl402/Intl/getCanonicalLocales/preferred-subtag.js
+++ b/test/intl402/Intl/getCanonicalLocales/preferred-subtag.js
@@ -1,0 +1,24 @@
+// Copyright 2016 Mozilla Corporation. All rights reserved.
+// This code is governed by the license found in the LICENSE file.
+
+/*---
+esid: sec-canonicalizelanguagetag
+description: Test for the replacement of a deprecated subtag with its preferred subtag.
+info: |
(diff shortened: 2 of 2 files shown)


littledan:
LGTM

leobalter:
it LGTM, but this can be improved by moving this specific assertion to a new test file. If you prefer I can do that in a follow-up PR.

jungshik:
Oops. sorry for the late reply. I didn't press 'comment' button. @leobalter,  I pulled the test out of main.js and created a new test (preferred-subtag.js). Please, take another look.

jungshik:
@tcare, could you drop 'waiting-author' label now that I have responded? :-)  thanks.

tcare:
Looks good.

PR #748, opened 2016-08-20
Remove Python-based test runner

jugglinmike:
TC-39 agreed to remove this tool from the Test262 repository in May of
2016 [1]. It has since been re-published as a standalone project to
facilitate independent development by interested parties [2]. Remove it
from the Test262 repository.

[1] https://github.com/tc39/tc39-notes/blob/master/es7/2016-05/may-25.md
[2] https://github.com/test262-utils/test262-harness-py

This resolves gh-128 and gh-647. /cc @littledan @tschneidereit


littledan:
LGTM; this wouldn't block the next V8 test262 roll as we are using it from another location.

PR #752, opened 2016-08-28
Create return-values.js (Number#toFixed)

Yaffle:


test/built-ins/Number/prototype/toFixed/return-values.js
```diff
@@ -0,0 +1,12 @@
+// Copyright (C) 2016 The V8 Project authors. All rights reserved.
```
littledan: You probably want to out your own name here.

test/built-ins/Number/prototype/toFixed/return-values.js
```diff
@@ -0,0 +1,12 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+es6id: 20.1.3.3
+esid: sec-number.prototype.tofixed
+description: >
+  Return regular string values
+---*/
+
+// a bug in IE ? - Edge 14
```
leobalter: this is vague, but a link reference wouldn't help it as well. The bug case seems as a good reason for this addition but I would add it as a commit description, not to the code.

Do you mind to move this to the commit description providing more info if any?

test/built-ins/Number/prototype/toFixed/return-values.js
```diff
@@ -0,0 +1,12 @@
...
+assert.sameValue((1.255).toFixed(2), "1.25");
```
Yaffle: Probabyl I should use `1.25499999999999989342` here instead of `1.255`.

leobalter: I don't have Edge right now to verify, but if the current assertion is enough to flag the bug, I would say both examples would be fine.

test/built-ins/Number/prototype/toFixed/return-values.js
Yaffle: No, I don't

test/built-ins/Number/prototype/toFixed/return-values.js
Yaffle: @leobalter , who reads commit messages?
I have did the change.

test/built-ins/Number/prototype/toFixed/return-values.js
leobalter: @Yaffle this change is necessary if you're not on the V8 team, which I don't really know.

I need confirm but if you don't want to worry about it, the best default option would be setting the Copyright to ECMA International. cc @bterlson

bterlson: Does v8 own the copyright to this file? If not, the copyright line must change.

The proper thing to do is to sign the CLA and put your own name there.

test/built-ins/Number/prototype/toFixed/return-values.js
bterlson: 1.255 works too, so I would say use both!

test/built-ins/Number/prototype/toFixed/return-values.js
```diff
@@ -1,4 +1,4 @@
-// Copyright (C) 2016 The V8 Project authors. All rights reserved.
+// Copyright (C) 2016 Anonymous Contributor. All rights reserved.
```
leobalter: This still need to change. I can't merge it this way.

test/built-ins/Number/prototype/toFixed/return-values.js
```diff
@@ -8,5 +8,4 @@ description: >
   Return regular string values
 ---*/
 
-// a bug in IE ? - Edge 14
-assert.sameValue((1.255).toFixed(2), "1.25");
+assert.sameValue((1.25499999999999989342).toFixed(2), "1.25");
```
leobalter: can we keep both assertions?

test/built-ins/Number/prototype/toFixed/return-values.js
Yaffle: @bterlson , 1.25499999999999989342 === 1.255 for binary64 floating point

test/built-ins/Number/prototype/toFixed/return-values.js
Yaffle: @leobalter , 1.255 === 1.25499999999999989342 .
In case some engine will use binary256 arithmetic  `1.255` _could_ be closer to "1.26" , and so the test will be incorrect.

rwaldron:
This has been pending a license change for nearly 5 months

Yaffle:
The license wasn't the issue to add the test to the ChakraCore -  https://github.com/Microsoft/ChakraCore/issues/1511

leobalter:
That is not the same project and neither has the same governance.

We follow rules and convention and personally I even want to solve this by
not requiring licenses in new files since we have the CLA requirement. For
that I need to get consensus in the next meeting.

bterlson:
@leobalter I think you're saying not requiring a license header, but in case you mean no license at all, that will not work. CLA requires a BSD license.

For this test, I don't understand why we enforced the license requirement really. IANAL (obviously) but I would be surprised if it qualifies as a creative work under US copyright law. Also, de minimis non curat lex and all that. It might be interesting to find an expert and see if we can craft a concrete triviality bar below which we can accept any test regardless of license.

Yaffle:
Well... I cannot reproduce the bug in Edge 15.15014, so may be, this test is not interesting now.

PR #754, opened 2016-08-30
Fix operator precedence

shvaikalesh:
Unlike PHP, in JavaScript `!` has higher precedence than `instanceof`, thus `!smth instanceof TypeError` will never (unless `@@hasInstance` is defined) be `true`.


tcare:
Thanks. It looks like there is no other instanceof this in the repo. :)

PR #755, opened 2016-08-30
Update toString/proxy.js to test stringifing functions

kmiller68:


rwaldron:
All of this should be in a different file, maybe even more than one, but definitely not jammed into an existing test file

leobalter:
I agree with Rick. The current file could be renamed to specify what it is currently testing. Then, we can add an individual test file for the function definitions, arrow functions, and generators.

The files can be named as `proxy-array.js`, `proxy-fn.js`, `proxy-arr-fn.js`, `proxy-gen-fn.js` or anything better.

test/built-ins/Object/prototype/toString/proxy.js
```diff
@@ -38,3 +47,50 @@ assert.sameValue(
...
+assert.sameValue(
+  Object.prototype.toString.call(generatorProxy), '[object GeneratorFunction]', 'generator function proxy'
+);
+assert.sameValue(
+  Object.prototype.toString.call(generatorProxyProxy),
+  '[object GeneratorFunction]',
+  'proxy for generator function proxy'
+);
+
+delete generatorProxy.__proto__[Symbol.toStringTag];
```
leobalter: is this really necessary? If so, it's probably interesting to have one another test file for it and avoid using `__proto__`.

kmiller68: Are you referring to testing generators as a whole? Or are you more referring to testing generators' Symbol.toStringTag?

kmiller68: I could do `generatorProxy.constructor.prototype` or `(function*() {}).constructor.prototype`. Would that be preferable? I'm not sure how much value there is in assuming both the `constructor` and `prototype` properties work correctly over assuming `__proto__` works correctly, though.

leobalter: you already have `generatorProxy.constructor.prototype` so I think that would be the best option.

rwaldron:
@kmiller68 ping

leobalter:
@kmiller68 Please tell me if you need any support to fix this PR, technically or even with available time.

Issue #749, opened 2016-08-26
Test wanted: RegExp lastIndex semantics change


littledan:
A change to RegExp semantics are proposed in ES https://github.com/tc39/ecma262/pull/627 . This change is blocking on test262 tests being written (and possibly validation by V8 of web compatibility, which is in progress). Once that happens, both PRs should be merged at the same time. cc @schuay

leobalter:
fixed at #760

PR #760, opened 2016-09-19
Tests for new lastIndex semantics

schuay:
Add and update tests for the lastIndex semantic change introduced in
https://github.com/tc39/ecma262/pull/627. See also tc39/test262#749.

### DIFF
--- a/test/built-ins/RegExp/prototype/Symbol.match/builtin-failure-g-set-lastindex.js
+++ b/test/built-ins/RegExp/prototype/Symbol.match/builtin-failure-g-set-lastindex.js
@@ -30,8 +30,6 @@ info: >
 features: [Symbol.match]
 ---*/
 
-// TODO: Update the comment.
-
 var r = /a/g;
 r.lastIndex = 3;
 
--- a/test/built-ins/RegExp/prototype/Symbol.match/builtin-failure-g-set-lastindex.js
+++ b/test/built-ins/RegExp/prototype/Symbol.match/builtin-failure-g-set-lastindex.js
@@ -20,18 +20,19 @@ info: >
 
     [...]
     3. Let length be the number of code units in S.
-    4. Let lastIndex be ToLength(Get(R,"lastIndex")).
     [...]
-    14. Let matchSucceeded be false.
-    15. Repeat, while matchSucceeded is false
+    11. Let matchSucceeded be false.
+    12. Repeat, while matchSucceeded is false
         a. If lastIndex > length, then
-           i. Let setStatus be Set(R, "lastIndex", 0, true).
-           ii. ReturnIfAbrupt(setStatus).
-           iii. Return null.
+           i. If global is true or sticky is true, then
+              1. Perform ? Set(R, "lastIndex", 0, true).
+           ii. Return null.
 features: [Symbol.match]
 ---*/
 
-var r = /a/;
+// TODO: Update the comment.
+
+var r = /a/g;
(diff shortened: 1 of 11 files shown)


test/built-ins/RegExp/prototype/exec/success-g-lastindex-no-access.js
```diff
@@ -0,0 +1,32 @@
...
+    15. If global is true or sticky is true, then
+        a. Perform ? Set(R, "lastIndex", e, true).
+---*/
+
+var lastIndexReads = 0;
+
+var r = /./g;
+r.lastIndex = {
+  valueOf: function() {
+    lastIndexReads++;
```
leobalter: You could check the final value for `lastIndexReads`

test/built-ins/RegExp/prototype/Symbol.search/set-lastindex-restore.js
```diff
@@ -6,25 +6,29 @@ es6id: 21.2.5.9
...
 features: [Symbol.search]
 ---*/
 
-var latestValue;
+var latestValue = 86;
+var callCount = 0;
 var fakeRe = {
   get lastIndex() {
-    return 86;
+    return latestValue;
```
leobalter: I might be wrong, but by doing this you avoid confirming what the description says: "The `lastIndex` value is restored following match execution"

Now you can't tell `get/set lastIndex` were called

test/built-ins/RegExp/prototype/Symbol.match/builtin-failure-g-set-lastindex.js
```diff
@@ -20,18 +20,19 @@ info: >
...
-           ii. ReturnIfAbrupt(setStatus).
-           iii. Return null.
+           i. If global is true or sticky is true, then
+              1. Perform ? Set(R, "lastIndex", 0, true).
+           ii. Return null.
 features: [Symbol.match]
 ---*/
 
-var r = /a/;
+// TODO: Update the comment.
```
leobalter: which comment? This seems fine, the file renaming should be enough.

test/built-ins/RegExp/prototype/exec/success-g-lastindex-no-access.js
schuay: Thanks, good catch :+1: Fixed.

test/built-ins/RegExp/prototype/Symbol.match/builtin-failure-g-set-lastindex.js
schuay: The comment was already updated, I just forgot to remove the TODO. Fixed.

test/built-ins/RegExp/prototype/Symbol.search/set-lastindex-restore.js
schuay: I think the test is working as intended, but it's quite subtle:

1. Symbol.search sets lastIndex to 0 prior to the exec call (since previousLastIndex != 0).
2. exec sets lastIndex to null.
3. Symbol.search restores lastIndex to 86 (since previousLastIndex != currentLastIndex).

We check that exec was actually called, and that the value was restored to 86 after the its call.

schuay:
Thanks for reviewing, your comments should be addressed by the latest patch.

leobalter:
@tcare +1 to merge it.

PR #764, opened 2016-09-27
Add tests to handle revised template literals

disnet:
Add tests for the template literal revision [proposal](https://github.com/tc39/proposal-template-literal-revision) (the proposal is currently at stage 3). The proposal allows tagged template literals to use previously disallowed escape sequences.

### DIFF
--- a/test/language/expressions/tagged-template/invalid-escape-sequences.js
+++ b/test/language/expressions/tagged-template/invalid-escape-sequences.js
@@ -1,0 +1,70 @@
+// Copyright (C) 2016 Tim Disney. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+description: Invalid unicode escape sequence in tagged template literals are allowed
+esid: sec-template-literal-lexical-components
+---*/
+
+(strs => {
+  assert.sameValue(strs[0], undefined, 'Cooked template value should be undefined for illegal escape sequences');
+  assert.sameValue(strs.raw[0], '\\01');
+})`\01`;
+
+(strs => {
+  assert.sameValue(strs[0], undefined, 'Cooked template value should be undefined for illegal escape sequences');
+  assert.sameValue(strs.raw[0], '\\1');
+})`\1`;
+
+(strs => {
+  assert.sameValue(strs[0], undefined, 'Cooked template value should be undefined for illegal escape sequences');
+  assert.sameValue(strs.raw[0], '\\1');
+})`\xg`;
+
+(strs => {
+  assert.sameValue(strs[0], undefined, 'Cooked template value should be undefined for illegal escape sequences');
+  assert.sameValue(strs.raw[0], '\\1');
+})`\xAg`;
+
+(strs => {
+  assert.sameValue(strs[0], undefined, 'Cooked template value should be undefined for illegal escape sequences');
+  assert.sameValue(strs.raw[0], '\\u0');
+})`\u0`;
+
+(strs => {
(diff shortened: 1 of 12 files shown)


michaelficarra:
I think the syntax errors should be separate files.

test/language/expressions/template-literal/invalid-unicode-escape-sequence-truncated.js
```diff
@@ -9,4 +9,26 @@ info: >
...
+
+`\u000g`;
+
+`\u{g`;
+
+`\u{0`;
+
+`\u{10FFFFF}`;
+
+`\u{10FFFFF}${'inner'}right`;
```
leobalter: these cases should be split into different files for each case, otherwise you might find false positives for the pre-existing code.

test/language/expressions/tagged-template/invalid-escape-sequences.js
```diff
@@ -0,0 +1,70 @@
+// Copyright (C) 2016 Tim Disney. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+description: Invalid unicode escape sequence in tagged template literals are allowed
+es6id: 11.8.6
```
leobalter: there's no valid `es6id` here, this is after es6 and `esid` is the tag you need to apply with the respective section id in the specs.

test/language/expressions/tagged-template/invalid-escape-sequences.js
```diff
@@ -0,0 +1,70 @@
...
+---*/
+
+(strs => {
+  assert(strs[0] === undefined);
+  assert(strs.raw[0] === '\\01');
+})`\01`;
```
leobalter: can we use `assert.equal` and provide an assertion message? Otherwise failing tests would be very confusing in this file.

test/language/expressions/tagged-template/invalid-escape-sequences.js
michaelficarra: It's a fairly minor change to an existing ES6 feature. I don't think any new sections were even added. I may be wrong, though.

tcare: esid is the preferred way going forward, so we don't need to keep updating section numbers. The named spec section is used instead, e.g. esid: sec-template-literals

michaelficarra: :+1: Okay, that sounds great.

leobalter: es6id is a strict reference to the es6 release. Anything after ES6 should use esid tag only.

If the section is undefined, we conventionally use `esid: pending`

disnet:
Updated PR to address comments.

tcare:
Looks good, will wait for @leobalter to sign off.

test/language/expressions/template-literal/invalid-unicode-escape-sequence-1.js
```diff
@@ -0,0 +1,9 @@
+// Copyright (C) 2016 Tim Disney. All rights reserved.
```
leobalter: This is a [file renaming](test/language/expressions/template-literal/invalid-unicode-escape-sequence-truncated.js), the copyrights should be kept in this specific case.

disnet: Fixed

michaelficarra:
LGTM.

PR #766, opened 2016-09-29
Add tests for non-optional arguments of Date.UTC()

juandopazo:
Are these actually necessary? I'm tempted to add them just to be able to change them in the future (new default for month).

@bterlson

### DIFF
--- a/test/built-ins/Date/UTC/nans.js
+++ b/test/built-ins/Date/UTC/nans.js
@@ -17,6 +17,7 @@ info: |
   9. Return TimeClip(MakeDate(MakeDay(yr, m, dt), MakeTime(h, min, s, milli))).
 ---*/
 
+assert.sameValue(Date.UTC(NaN), NaN, 'year');
 assert.sameValue(Date.UTC(NaN, 0), NaN, 'year');
 assert.sameValue(Date.UTC(1970, NaN), NaN, 'month');
 assert.sameValue(Date.UTC(1970, 0, NaN), NaN, 'date');
--- a/test/built-ins/Date/UTC/nans.js
+++ b/test/built-ins/Date/UTC/nans.js
@@ -14,13 +14,16 @@ info: |
   7. If ms is supplied, let milli be ? ToNumber(ms); else let milli be 0.
   8. If y is not NaN and 0 ≤ ToInteger(y) ≤ 99, let yr be 1900+ToInteger(y);
      otherwise, let yr be y.
-  9. Return TimeClip(MakeDate(MakeDay(yr, m, dt), MakeTime(h, min, s, milli))). 
+  9. Return TimeClip(MakeDate(MakeDay(yr, m, dt), MakeTime(h, min, s, milli))).
 ---*/
 
-assert.sameValue(new Date(NaN, 0).getTime(), NaN, 'year');
-assert.sameValue(new Date(1970, NaN).getTime(), NaN, 'month');
-assert.sameValue(new Date(1970, 0, NaN).getTime(), NaN, 'date');
-assert.sameValue(new Date(1970, 0, 1, NaN).getTime(), NaN, 'hours');
-assert.sameValue(new Date(1970, 0, 1, 0, NaN).getTime(), NaN, 'minutes');
-assert.sameValue(new Date(1970, 0, 1, 0, 0, NaN).getTime(), NaN, 'seconds');
-assert.sameValue(new Date(1970, 0, 1, 0, 0, 0, NaN).getTime(), NaN, 'ms');
+assert.sameValue(Date.UTC(NaN, 0), NaN, 'year');
+assert.sameValue(Date.UTC(1970, NaN), NaN, 'month');
+assert.sameValue(Date.UTC(1970, 0, NaN), NaN, 'date');
+assert.sameValue(Date.UTC(1970, 0, 1, NaN), NaN, 'hours');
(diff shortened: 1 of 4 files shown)


bterlson:
Consensus is Date.UTC(year) should work (default other parameters to 0/1 as appropriate)

Also why do none of the previous assertions in this file actually use Date.UTC? Curious...

juandopazo:
Yeah I was surprised too. Should I remove the `Date.UTC(1970) === NaN` test?

bterlson:
@juandopazo yeah, and add a test for Date.UTC(1970) returns 0 :-P

juandopazo:
Done 😛

test/built-ins/Date/UTC/nans.js
```diff
@@ -14,13 +14,16 @@ info: |
...
+  9. Return TimeClip(MakeDate(MakeDay(yr, m, dt), MakeTime(h, min, s, milli))).
 ---*/
 
-assert.sameValue(new Date(NaN, 0).getTime(), NaN, 'year');
-assert.sameValue(new Date(1970, NaN).getTime(), NaN, 'month');
-assert.sameValue(new Date(1970, 0, NaN).getTime(), NaN, 'date');
-assert.sameValue(new Date(1970, 0, 1, NaN).getTime(), NaN, 'hours');
-assert.sameValue(new Date(1970, 0, 1, 0, NaN).getTime(), NaN, 'minutes');
-assert.sameValue(new Date(1970, 0, 1, 0, 0, NaN).getTime(), NaN, 'seconds');
-assert.sameValue(new Date(1970, 0, 1, 0, 0, 0, NaN).getTime(), NaN, 'ms');
```
leobalter: it's interesting these tests should have a place for the Date constructor.

test/built-ins/Date/UTC/nans.js
```diff
@@ -14,13 +14,16 @@ info: |
...
+assert.sameValue(Date.UTC(NaN, 0), NaN, 'year');
```
leobalter: as we are tests the new defined two forms of Date.UTC, with one or two arguments, can we __clone__ this assertion to use the single argument form?

```js
assert.sameValue(Date.UTC(NaN), NaN, 'year');
```

test/built-ins/Date/UTC/nans.js
```diff
@@ -14,13 +14,16 @@ info: |
...
+assert.sameValue(Date.UTC(1970, NaN), NaN, 'month');
+assert.sameValue(Date.UTC(1970, 0, NaN), NaN, 'date');
+assert.sameValue(Date.UTC(1970, 0, 1, NaN), NaN, 'hours');
+assert.sameValue(Date.UTC(1970, 0, 1, 0, NaN), NaN, 'minutes');
+assert.sameValue(Date.UTC(1970, 0, 1, 0, 0, NaN), NaN, 'seconds');
+assert.sameValue(Date.UTC(1970, 0, 1, 0, 0, 0, NaN), NaN, 'ms');
+
+// Tests for non optional arguments
+assert.sameValue(Date.UTC(), NaN, 'missing year');
```
leobalter: this test case is great, but I would leave this alone to an individual file called `no-arg.js`.

The reason is because the current tests are handling NaN as Date.UTC arguments, this case asserts calling Date.UTC with no args.

test/built-ins/Date/UTC/return-value.js
```diff
@@ -14,9 +14,11 @@ info: |
   7. If ms is supplied, let milli be ? ToNumber(ms); else let milli be 0.
   8. If y is not NaN and 0 ≤ ToInteger(y) ≤ 99, let yr be 1900+ToInteger(y);
      otherwise, let yr be y.
-  9. Return TimeClip(MakeDate(MakeDay(yr, m, dt), MakeTime(h, min, s, milli))). 
+  9. Return TimeClip(MakeDate(MakeDay(yr, m, dt), MakeTime(h, min, s, milli))).
 ---*/
 
+assert.sameValue(Date.UTC(1970), 0, '1970');
```
leobalter: We didn't have tests for the single arg cases before that now we could have some interesting cases.

Can we test some variation here? Instead of just asserting 1970, it would be great to include another year (2016, maybe?)

leobalter: Also, the `infinity-make-day.js` could include these cases:

```js
assert.sameValue(Date.UTC(Infinity), NaN, 'year: Infinity - single arg');
assert.sameValue(Date.UTC(-Infinity), NaN, 'year: -Infinity - single arg');
```

rwaldron:
@juandopazo I believe I can make the changes that @leobalter requested and add them to this PR, that way we can get it merged and resolved

test/built-ins/Date/UTC/return-value.js
rwaldron: Done

test/built-ins/Date/UTC/nans.js
rwaldron: Done

test/built-ins/Date/UTC/nans.js
rwaldron: Done

test/built-ins/Date/UTC/nans.js
rwaldron: Done

PR #769, opened 2016-10-05
Object computed property is an ES6 feature.

nadavrot:
@sebmarkbage

### DIFF
--- a/test/language/expressions/object/computed-property-evaluation-order.js
+++ b/test/language/expressions/object/computed-property-evaluation-order.js
@@ -1,7 +1,7 @@
 // Copyright (C) 2016 Michael Ficarra. All rights reserved.
 // This code is governed by the BSD license found in the LICENSE file.
 /*---
-esid: sec-object-initializer-runtime-semantics-propertydefinitionevaluation
+es6id: 12.2.6
 description: >
     Evaluation of PropertyDefinitionList occurs in order, and each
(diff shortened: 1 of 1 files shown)


tcare:
It may be an ES6 feature, but we should leave an esid in so that it maps to future spec changes regardless of the numbering.

nadavrot:
@tcare I am not sure I understand. The test case is not a valid ES5 program. Why do we want to mark it as esid and not es6id? Is it because of the label that follows?

leobalter:
esid is not meant for es5. For es5 features we used es5id, than we set es6id for es6. The problem is that it became unsustainable and unnecessary after the standards moved to GitHub and TC39 is using the per year releases model.

esid just means this is related to the last version of the specs (draft included). Test262 does not keep track of per specific editions of the specs. There's just one, new editions replace the previous.

In this PR esid could be kept, as it's a better mapping for the particular feature. You can still add the es6id if you want, it's not relevant and not necessary. These tags are just mappings of the features in the specs, not the version of the specs.

rwaldron:
@nadavrot ping

re: @leobalter's question: 

> In this PR esid could be kept, as it's a better mapping for the particular feature. You can still add the es6id if you want, it's not relevant and not necessary.

Can this be closed or are you going to make that change above?

nadavrot:
Thanks for the help guys. Closing.

Issue #770, opened 2016-10-07
Invalid YAML in test/built-ins/global/global-object.js and property-descriptor.js


dop251:
Description contains '`' which is not allowed (I am getting 'found character that cannot start any token while scanning for the next token at line 2').

leobalter:
@dop251 please take a look at #772

dop251:
Looks good, cheers

PR #772, opened 2016-10-07
Fix invalid yaml on tests for global

leobalter:
Closes gh-770


tcare:
Looks good. Something we can look at catching in a PR gate.

PR #774, opened 2016-10-13
Revert "islamicc should be canonicalized to islamic-civil (#747)"

leobalter:
This reverts commit da4f4385fdf88ff2c8acf036efaaa62f8cd6bd58.

cc @zbraniecki @tcare @jungshik


tcare:
Based on discussion in #743 and https://bugzilla.mozilla.org/show_bug.cgi?id=837961#c30 and https://tools.ietf.org/html/rfc6067#section-2.1.1 I think this is wrong but I would like for @jungshik to confirm. I'll give him a bit longer to confirm.

tcare:
Merging, @jungshik please open an issue if you think this is wrong.

littledan:
With this revert, it looks like test262 tests for it to *not* be canonicalized. However, [CanonicalizeLanguageTag](https://tc39.github.io/ecma402/#sec-canonicalizelanguagetag), says,

> Implementations are allowed, but not required, to apply these additional rules.

Seems like we should delete both the old and the new test. Also, seems like we should change the spec; it's not great to allow and not require something like this, in my opinion.

leobalter:
@littledan you're right. To address that "not required" phrase we could flag the tests but we don't have the appropriate tooling for that. Removing the test is the current solution.

> Also, seems like we should change the spec; it's not great to allow and not require something like this, in my opinion.

I'm +1 for that.

littledan:
@anba mentioned on IRC,

> AFAICT RFC-6067 canonicalization only handles reordering of attributes and keywords (https://tools.ietf.org/html/rfc6067#section-2.1.1), so "islamicc" should stay "islamicc"

@jungshik Do you know why our current implementation canonicalizes the name of the calendar?

jungshik:
Based on the discussion in #743, I'll make a new pull request to drop the test.

Issue #773, opened 2016-10-13
language/module-code/namespace/Symbol.iterator/values-binding-types_.js should be a FIXTURE


GeorgNeis:
language/module-code/namespace/Symbol.iterator/values-binding-types_.js should be called language/module-code/namespace/Symbol.iterator/values-binding-types_FIXTURE.js as it doesn't set the module flag.

jugglinmike:
Right you are! I've opened gh-775 to address this.

PR #775, opened 2016-10-13
Correct name of fixture file

jugglinmike:
The file previously named `values-binding-types_.js` is not intended to
be interpreted as a test. Therefor (in accordance with the project's
`INTERPETING.md` file), its name should include `_FIXTURE` as a suffix.

This resolves gh-773.


tcare:
Looks good

PR #776, opened 2016-10-14
Add test for inequality of RegExps created from the same source

bakkot:
We already have [a single test](https://github.com/bakkot/test262/blob/master/test/language/literals/regexp/S7.8.5_A4.2.js) (that I could find) which tests that two identical regular expression literals create distinct objects, but we don't test that the same literal source, evaluated at different times, creates distinct objects.


tcare:
Looks good, thanks.

PR #777, opened 2016-10-19
Update S15.4.4.11_A7.2.js

Yaffle:
`Array.prototype.sort` should be used instead of `Array.sort`

### DIFF
--- a/test/built-ins/Array/prototype/sort/S15.4.4.11_A7.2.js
+++ b/test/built-ins/Array/prototype/sort/S15.4.4.11_A7.2.js
@@ -9,7 +9,7 @@ description: Checking use hasOwnProperty, delete
 
 //CHECK#1
 if (Array.prototype.sort.hasOwnProperty('length') !== true) {
-  $ERROR('#1: Array.sort.prototype.hasOwnProperty(\'length\') === true. Actual: ' + (Array.sort.prototype.hasOwnProperty('length')));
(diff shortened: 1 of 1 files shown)


tcare:
Looks good, thanks!

test/built-ins/Array/prototype/sort/S15.4.4.11_A7.2.js
```diff
@@ -9,7 +9,7 @@ description: Checking use hasOwnProperty, delete
 
 //CHECK#1
 if (Array.prototype.sort.hasOwnProperty('length') !== true) {
-  $ERROR('#1: Array.sort.prototype.hasOwnProperty(\'length\') === true. Actual: ' + (Array.sort.prototype.hasOwnProperty('length')));
+  $ERROR('#1: Array.prototype.sort.hasOwnProperty(\'length\') === true. Actual: ' + (Array.prototype.sort.hasOwnProperty('length')));
 }
```
rwaldron: Would be great to get tests like these updated to match the modern approach:

- https://github.com/tc39/test262/blob/master/test/built-ins/Array/prototype/find/find.js
- https://github.com/tc39/test262/blob/master/test/built-ins/Array/prototype/find/length.js
- https://github.com/tc39/test262/blob/master/test/built-ins/Array/prototype/find/name.js

PR #782, opened 2016-10-25
Add Intl.PluralRules tests

zbraniecki:
This is not yet ready for landing, since the spec is not finalized,
but I'd like to get a preliminary feedback on the structure and list of tests to make it easier to land once we finalize the spec (there's just one issue remaining).

The spec is at https://github.com/tc39/proposal-intl-plural-rules

Can someone give me a general feedback on if this is the right way to structure the patch for the spec?

### DIFF
--- a/test/intl402/PluralRules/builtin.js
+++ b/test/intl402/PluralRules/builtin.js
@@ -2,6 +2,11 @@
 // This code is governed by the license found in the LICENSE file.
 
 /*---
+esid: sec-Intl.PluralRules
+description: >
+    Tests that Intl.PluralRules meets the requirements for
+    built-in objects defined by the introduction of chapter 17 of the
+    ECMAScript Language Specification.
 author: Zibi Braniecki
 includes: [testBuiltInObject.js]
 ---*/
--- a/test/intl402/PluralRules/builtin.js
+++ b/test/intl402/PluralRules/builtin.js
@@ -1,0 +1,9 @@
+// Copyright 2016 Mozilla Corporation. All rights reserved.
+// This code is governed by the license found in the LICENSE file.
+
+/*---
+author: Zibi Braniecki
+includes: [testBuiltInObject.js]
+---*/
+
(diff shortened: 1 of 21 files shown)


tcare:
Sorry for the delayed response, I wrote a post but didn't actually post it.

Naming looks good, but I think there could be some improvements to the header comments

I really like the way @jugglinmike does them, for example: https://github.com/tc39/test262/blob/c9569a7e7e648b95beb89957204254dc85c00525/test/language/expressions/yield/star-rhs-iter-rtrn-no-rtrn.js

In particular I like seeing:
- ESID and the section header name from the spec text
- A brief description of what the test is trying to achieve, and possibly why it's interesting
- Some of the spec text that's relevant if there's a large blob of it. This isn't really that important if it's subject to change though.

As for the test code itself, using asserts such as assert.sameValue is preferable to the explicit if blocks.

Hope that helps!

zbraniecki:
Yeah, I'm planning to add header comments to all tests, just waiting for two patches to land in the spec to avoid having to update them.

Ok, I'll look into asserts vs `if` blocks.

Does it seem like a fairly good initial coverage of the spec? In other words, once I fix the header comments and if blocks, is it landable or does it require more work?

tcare:
It's definitely in the spirit of what I would be testing, yes. However without description of the motivations of the individual tests it's a bit harder for me to be an authority on that - that's probably a question best answered by people authoring/reviewing the proposal.

zbraniecki:
@tcare - added the esid, description and info where appropriate.

Can you verify if this is good for landing?

tcare:
Looks good, thanks :)

zbraniecki:
If this looks good, can we merge it?

tcare:
Sorry I missed this. Merging :)

littledan:
It looks like these tests still leave out a bunch of functionality of PluralRules and focus on the surface. For example, I don't see anything testing the way that numbers are rounded according to the options. SpiderMonkey has more tests internally in `js/src/tests/Intl/PluralRules`; would it make sense to upstream these into test262?

PR #784, opened 2016-10-28
Test for new TypedArray iterator detach logic

littledan:
Tests new logic in https://github.com/tc39/ecma262/pull/724

Not ready to merge yet, pending discussion in https://github.com/tc39/ecma262/issues/713

### DIFF
--- a/test/built-ins/ArrayIteratorPrototype/next/detach-typedarray-in-progress.js
+++ b/test/built-ins/ArrayIteratorPrototype/next/detach-typedarray-in-progress.js
@@ -1,0 +1,25 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-%arrayiteratorprototype%.next
+description: If the underlying TypedArray is detached during iteration, throw
+info: >
+  %ArrayIteratorPrototype%.next( )
+
+  ...
+  8. If _a_ has a [[TypedArrayName]] internal slot, then
+    a. If IsDetachedBuffer(_a_.[[ViewedArrayBuffer]]) is *true*, throw a *TypeError* exception.
+includes: [testTypedArray.js, detachArrayBuffer.js]
+---*/
+
+testWithTypedArrayConstructors(TA => {
+  var typedArray = new TA(5);
+  var i = 0;
+  assert.throws(TypeError, () => {
+    for (let key of typedArray.keys()) {
+      $.detachArrayBuffer(typedArray.buffer);
+      i++;
+    }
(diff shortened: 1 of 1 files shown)


bterlson:
What's the plan for $DETACHBUFFER? Currently it's just a stub that throws, right?

caitp:
TC39 proposal for ArrayBuffer.detach(), just for fun?

bterlson:
I mean, there's some argument for it - if you want to test some functionality in the presence of detached buffers, it'd be nice to have a standard way to make one.

leobalter:
> I mean, there's some argument for it - if you want to test some functionality in the presence of detached buffers, it'd be nice to have a standard way to make one.

I mentioned this before. I'm also sad ArrayBuffer.transfer is no longer active too.

`$DETACHBUFFER` should be replaced by any non-es-standard way to detach a buffer, if that's not available, skip tests that includes `detachArrayBuffer.js`.

littledan:
Plan? I think it's in pretty good shape--the web platform exposes it with `postMessage`, and multiple harnesses have ways of calling into that same hook. I don't see a problem. What's our plan for the way that the module tests require harness integration?

littledan:
@leobalter I don't think harnesses which don't implement `$DETACHBUFFER` should get a free pass for not having it. Those should count as failures, if we want to have a fair comparison.

leobalter:
@littledan yes. that's what we are already doing, and `$DETACHBUFFER` does it nicely mentioning why it's failing.

Still, that allows anyone to monkey-patch files including `detachArrayBuffer.js` or simply ignore them, we can't control that.

bterlson:
> What's our plan for the way that the module tests require harness integration?

I'm going to teach eshost how to module. Once done it should paper over the differences in running modules in browsers vs. command line hosts. Shouldn't a similar thing be done for $DETACHBUFFER? Eg. $.detachBuffer in eshost maybe?

littledan:
`detachBuffer` in eshost sounds good to me. Do you anticpate anything in the test262 repository changing as a result?

bterlson:
@littledan presumably $DETACHBUFFER would simply call $.detachBuffer (or else tests could be updated to call this directly).

littledan:
Updated the test for $.detachArrayBuffer and minor spec text change.

tcare:
Looks good

littledan:
Well, this isn't quite correct until that spec change is merged. However, I hope it gets merged soon.

tcare:
Alright. Let me know if you want me to undo it.

bterlson:
I pulled in tc39/ecma262#724 so no revert needed :-P

PR #786, opened 2016-11-08
Integer indexed property descriptor must be non-configurable

hferreiro:


test/built-ins/TypedArrays/internals/DefineOwnProperty/key-is-lower-than-zero.js
```diff
@@ -24,7 +24,6 @@ testWithTypedArrayConstructors(function(TA) {
   assert.sameValue(
     Reflect.defineProperty(sample, "-1", {
       value: 42,
-      configurable: true,
```
caitp: I think I see the value in specifically testing the `return false` on step b.iii. --- but it would match the style of the test better to just change `configurable` to false, to explicitly match the spec requirements.

hferreiro: Agreed.

tcare:
Looks good

PR #787, opened 2016-11-10
fix incorrect tests for trailing commas

thejoshwolfe:
* Method declarations should not be followed by a comma.
* A comma after a spread parameter is actually allowed.

### DIFF
--- a/test/language/expressions/call/trailing-comma.js
+++ b/test/language/expressions/call/trailing-comma.js
@@ -2,13 +2,10 @@
 // This code is governed by the BSD license found in the LICENSE file.
 /*---
 description: >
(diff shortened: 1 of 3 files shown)


test/language/expressions/call/trailing-comma.js
```diff
@@ -6,9 +6,6 @@ description: >
   in a call expression.
```
caitp: I would update that description comment

thejoshwolfe: good catch. updated.

jeffmo:
Thanks, this look right -- and good catch!

tcare:
Looks good

PR #788, opened 2016-11-11
Fix 'negative' frontmatter key in yield tests

nwidger:
This commit corrects the 'negative' frontmatter key in
test/language/expressions/yield/in-iteration-stmt.js and
test/language/expressions/yield/star-in-iteration-stmt.js.


nwidger:
Nevermind, this is a duplicate or PR #785.

PR #789, opened 2016-11-14
Replace islamicc with islamic-umalqura in getCanonicalLocales/main.js (Intl402)

jungshik:
Per discussion in #743, it's implementation-dependent how ismalicc is
canonicalized. However, 'islamic-umalqura' should be treated the same
way by all implementations.

### DIFF
--- a/test/intl402/Intl/getCanonicalLocales/main.js
+++ b/test/intl402/Intl/getCanonicalLocales/main.js
@@ -29,6 +29,4 @@ assertArray(gCL({'a': 0}), []);
 
 assertArray(gCL({}), []);
 
-assertArray(gCL(['ar-ma-u-ca-islamic-umalqura']), ['ar-MA-u-ca-islamic-umalqura']);
-
 assertArray(gCL(['th-th-u-nu-thai']), ['th-TH-u-nu-thai']);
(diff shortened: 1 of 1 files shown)


jungshik:
@tcare @zbraniecki  @littledan 
Pls, take a look.

littledan:
Not sure what the point of this new test is. Wasn't the old test trying to get specifically at canonlicalizing or not canonicalizing? I think the ideal test here would make sure that one of the two legal outcomes is achieved with respect to islamicc.

jungshik:
> Wasn't the old test trying to get specifically at canonlicalizing or not canonicalizing? 

I'm not sure of the original intent of having both 'ar-ma-u-ca-islamicc' and 'th-th-u-nu-thai' in the test. 
@zbraniecki : Didn't you write the test? If so, what's your intent? 

If it's about 'islamicc', either we have to delete it (because it's implementation dependent at the moment; replace it with an implementation-independent one which also tests a legitimate value with two subtags) or test that the result is either A or B (as you wrote).

zbraniecki:
Yeah, it was about `islamicc`. I'm ok with removing it.

jungshik:
Sorry for the delay. I removed  'islamic-umalqura'. Put together, islmaicc is dropped from main.js. 
Could you take a look? Thanks

jungshik:
@tcare, could you merge? This PR removes ''ar-ma-u-ca-islamicc'' from main.js per @zbraniecki 
Thanks

tcare:
Looks good

PR #792, opened 2016-11-17
Ensure Proxy [[Get]] and [[Set]] pass correct receiver

shvaikalesh:
WebKit issue: https://bugs.webkit.org/show_bug.cgi?id=164849
Chai.js issue: https://github.com/chaijs/chai/issues/855


littledan:
Great find! I hope we can get tests like this for all the Proxy traps if they are missing. Might also be good to test some more complex cases (Proxy in a proto chain, for example).

shvaikalesh:
Adding test for `[[Set]]` and `Receiver` in this PR seems reasonable. Regarding other traps: it seems all traps except `construct` and `apply` have this covered. I will tackle function traps in a separate PR.

littledan:
LGTM, modulo naming nit. Seems like only the first of three tests in the file tests for an undefined trap. Maybe `trap-receiver.js`?

test/built-ins/Proxy/get/trap-is-undefined-receiver.js
```diff
@@ -0,0 +1,25 @@
...
+    get attr() {
+        return this;
+    }
+};
+
+var p = new Proxy(target, { get: undefined });
+assert.sameValue(p.attr, p);
+
+var pHerit = new Proxy(Object.create(target), { get: null });
+assert.sameValue(pHerit.attr, pHerit);
```
caitp: how is this assertion different from the first one? Seems like it would be more interesting to make pHerit inherit `p`, and assert that since the handler doesn't have a "get" trap, it calls `target`'s accessor with `pHerit` as the receiver.

```js
var p = new Proxy(target, { get: undefined });

var pHerit = Object.create(p);
assert.sameValue(pHerit.attr, pHerit);
```

test/built-ins/Proxy/get/trap-is-undefined-receiver.js
```diff
@@ -0,0 +1,25 @@
...
+
+var pParent = Object.create(new Proxy(target, {}));
+assert.sameValue(pParent.attr, pParent);
```
caitp: I guess you've covered that down here, nevermind. Still not clear what the 2nd assertion is checking for though

test/built-ins/Proxy/get/trap-is-undefined-receiver.js
```diff
@@ -0,0 +1,25 @@
+// Copyright (C) 2016 Aleksey Shvayka. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.8
+description: >
```
tcare: Generally we put raw spec under 'info' and use 'description' for a brief description of the test. Would you mind updating these?

test/built-ins/Proxy/get/trap-is-undefined-receiver.js
```diff
@@ -0,0 +1,25 @@
+// Copyright (C) 2016 Aleksey Shvayka. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.8
```
tcare: In addition to es6id could you please add esid: sec-proxy-object-internal-methods-and-internal-slots-get-p-receiver

test/built-ins/Proxy/set/trap-is-undefined-receiver.js
```diff
@@ -0,0 +1,29 @@
+// Copyright (C) 2016 Aleksey Shvayka. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 9.5.9
```
tcare: Same here for esid: sec-proxy-object-internal-methods-and-internal-slots-set-p-v-receiver

test/built-ins/Proxy/get/trap-is-undefined-receiver.js
shvaikalesh: Thanks for pointing this out. Indeed, second test is useless. Removed it.

tcare:
LGTM, @caitp?

shvaikalesh:
Is file naming OK?

tcare:
Yep seems fine to me

caitp:
still LGTM, fwiw

PR #794, opened 2016-11-19
Avoid writing to top-level vars named "top", for browser compat

bakkot:
The HTML spec requires browsers to define a non-configurable property of the global (window) object named "top". (See [attributes of the window object](https://html.spec.whatwg.org/multipage/browsers.html#the-window-object) and [the definition of "[Unforgeable]"](https://www.w3.org/TR/WebIDL/#Unforgeable).) This makes it impossible for a browser to successfully run a test in strict mode (as is [required](https://github.com/tc39/test262/blob/master/INTERPRETING.md#strict-mode)) if said test attempts to write to a global variable named "top".

This patch switches variable names around in two tests so that they don't try to do that. It also incidentally fixes an off-by-one error which was only not causing failure due to two variables having the same value, and modifies those variables to be distinct.

Having these tests be runnable in a browser isn't an explicit goal, but it still seems worth it for such an easy fix. Other tests also make assumptions about properties of the global object (e.g. that `length` [does not exist](https://github.com/tc39/test262/blob/bef7f988d2a954c44fb4c8f99e91d9122ef275e3/test/language/expressions/function/dstr-ary-ptrn-rest-obj-prop-id.js#L64-L66) or that setting `name` to a Symbol [will succeed](https://github.com/tc39/test262/blob/bef7f988d2a954c44fb4c8f99e91d9122ef275e3/test/language/expressions/object/method-definition/generator-name-prop-symbol.js#L13)) which are by default violated in a browser context, but generally those conditions be arranged by deleting the relevant properties of the global object. `top` is an exception in that it is explicitly non-configurable.

### DIFF
--- a/test/language/statements/function/S13.2.2_A5_T1.js
+++ b/test/language/statements/function/S13.2.2_A5_T1.js
@@ -11,26 +11,26 @@ es5id: 13.2.2_A5_T1
 description: Declaring a function with "function __FACTORY(arg1, arg2)"
 ---*/
 
-var __VOLUME, __RED, __ID, __TOP, __BOTTOM, __LEFT, color, top, left, __device;
+var __VOLUME, __RED, __ID, __BOTTOM, __TOP, __LEFT, color, bottom, left, __device;
 
 __VOLUME=8;
 __RED="red";
 __ID=12342;
-__TOP=1.1;
-__BOTTOM=0.0;
-__LEFT=0.0;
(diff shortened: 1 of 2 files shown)


leobalter:
>  It also incidentally fixes an off-by-one error which was only not causing failure due to two variables having the same value, and modifies those variables to be distinct.

👏 👏 👏

PR #795, opened 2016-11-19
Use $.detachArrayBuffer API instead of throwing

bakkot:
INTERRPETING.md [requires](https://github.com/tc39/test262/blob/master/INTERPRETING.md#host-defined-functions) hosts to expose a `$.detachArrayBuffer` method, but nothing is using it; relevant tests call `$DETACHBUFFER`, which unconditionally throws. This PR simply modifies that function to defer to `$.detachArrayBuffer`.

If hosts don't have a way to expose the relevant abstract operation, they can choose to define `$.detachArrayBuffer` such that it throws. We don't need to decided that for them.

(They could of course also just monkey-patch `$DETACHBUFFER`, but that goes against INTERRPETING.md. The less special knowledge required the better, I should think.)

ref #784; ping @leobalter and @bterlson in light of that discussion.

### DIFF
--- a/harness/detachArrayBuffer.js
+++ b/harness/detachArrayBuffer.js
@@ -1,3 +1,6 @@
 function $DETACHBUFFER(buffer) {
-  throw new Test262Error("No method available to detach an ArrayBuffer");
+  if (!$ || typeof $.detachArrayBuffer !== "function") {
+    throw new Test262Error("No method available to detach an ArrayBuffer");
+  }
+  $.detachArrayBuffer(buffer);
 }


littledan:
This is a nice idea for a cleanup. I noticed that the documentation mismatched a while ago, but didn't do anything about it, so thanks for actually following up, @bakkot.

Note that some hosts, like V8, overrode `$DETACHBUFFER` to do something other than unconditionally throwing. The tests have been very useful for V8 (though it'd be awesome if we could do something about the mismatch between the spec and reality across all implementations!).

I think including a default definition of `$.detachArrayBuffer` that does something to cause the tests to fail, as `$DETACHBUFFER` does, is the right thing. If the error gets swallowed, test262 becomes less valuable. Maybe other VM maintainers pay better attention than me, but I don't think I would notice that a particular test262 update forces me to do some plumbing to support `$.detachArrayBuffer` unless Im punished by new failures.

bakkot:
@littledan, do you think this would cause spurious passes for existing consumers? I don't think it's a worry for new ones, since if the consumer doesn't provide their own implementation of `$.detachArrayBuffer` calling `$DETACHBUFFER` will throw, just as it would have before.

It also probably won't cause new failures for existing consumers, since if you overrode `$DETACHBUFFER` before it'll continue to be overridden after. Might be a maintainability hit, but that's kind of inevitable when monkey-patching. It's unfortunate that monkey-patching is currently (without this patch) required, but doing away with that is the point of this patch.

littledan:
Oh sorry actually this patch should be fine. Was responding to your description, but the patch itself is good.

leobalter:
This defers to a [previous discussion](https://github.com/tc39/test262/pull/527#issuecomment-199421887) when the $DETACHBUFFER was implemented.

The problem with `$.detachArrayBuffer` is because it is non-spec and implementation specific, right? 

The best solution would be collecting feedback from other runtimes and find all the options we should defer `$DETACHBUFFER` to.

leobalter:
It is good to track this issue as well: https://github.com/bterlson/test262-harness/issues/57

leobalter:
I just re-read the interpreting file. My bad I missed that `$.detachArrayBuffer` part.

One thing that might help implementations: `$DETACHBUFFER` should still check if `$.detachArrayBuffer` is defined and throw if not. Otherwise the failing message would be that `$` is undefined or `detachArrayBuffer` is not a function.

bakkot:
@littledan, sorry for the confusion.

@leobalter, good idea; updated.

bterlson:
Yes! I will try to get this into eshost soon.

Issue #791, opened 2016-11-16
Meta: doneprintHandle.js naming inconsistency


bakkot:
[INTERPRETING.md](https://github.com/tc39/test262/blob/84e6ba81d0861c7324cb0f4202ce40acd861b1ef/INTERPRETING.md) refers to `harness/donePrintHandle.js`, but the actual file is named `harness/doneprintHandle.js`. They should be consistent.

Personally I think the actual file's name ought to be fixed, but this might not be worth the cost to consumers.

jugglinmike:
I agree that changing the file name would be preferable but that the cost to
consumers is hard to judge. For the short term, lets consider it a typo in the
documentation--see gh-799 for a fix along those lines.

PR #798, opened 2016-11-21
Ensure that tests calling $DONE have the async flag

bakkot:
`$DONE` is defined in [harness/doneprintHandle.js](https://github.com/tc39/test262/blob/bef7f988d2a954c44fb4c8f99e91d9122ef275e3/harness/doneprintHandle.js), and INTERPRETING.md only [requires](https://github.com/tc39/test262/blob/master/INTERPRETING.md#flags) that doneprintHandle.js (modulo [naming inconsistency](https://github.com/tc39/test262/issues/791)) be included when the `async` flag is set, but several tests referenced it without setting that flag. This PR adds that flag to those tests.


jugglinmike:
Merged to `master` at commit e6abf652d95e09e5d952890f3a66470d21021b76. Thanks!

PR #801, opened 2016-11-23
Add link to web-based runner

bakkot:
Page is [here](https://bakkot.github.io/test262-web-runner/), source is [here](https://github.com/bakkot/test262-web-runner).

This project is incidentally the origin of my recent handful of fixes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,3 +16,4 @@ There are a number of volunteer-maintained projects that may be used to execute 
 
 - https://github.com/bterlson/test262-harness (platform: Node.js)
 - https://github.com/test262-utils/test262-harness-py (platform: Python)
+- https://bakkot.github.io/test262-web-runner/ (platform: web)


littledan:
Have you tried this on various browsers? Does it run well to completion across browsers, or do you ever encounter issues with that?

bakkot:
Works fine on Safari and Chrome, although crashes on sufficiently old Safari. Firefox appears to hang on some of the Array.prototype.slice/splice tests (at a guess it's trying to allocate massive amounts of memory) but is otherwise fine. Don't have a copy of IE or Edge around for easy testing.

bakkot:
Also, results on Chrome pretty much look [as expected](https://github.com/v8/v8/blob/master/test/test262/test262.status), though I haven't verified every single result. I don't know of similar places to look for what to expect of other browsers.

--

I should also mention that the UI is a bit less smooth on Chrome (lagging by a few seconds while running tests instead of continuously updating) than FF or Safari, but all my attempts to fix it made it take twice as long to finish on Chrome. Didn't seem worth it.

hzoo:
This is awesome @bakkot! Was going to try running Babel with https://github.com/bterlson/test262-harness but this is cool too

PR #804, opened 2016-12-06
Add missing operator in test

GyrosGeier:
Automatic Semicolon Insertion hides an error here by transforming a long
ReturnStatement into a ReturnStatement followed by an ExpressionStatement
that is never reached. The conditions on the second line are thus never
tested.


littledan:
Wow, great catch, LGTM.

PR #805, opened 2016-12-08
Add test for setting a module namespace object's prototype to null

ajklein:
The behavior changed from returning false to true in
https://github.com/tc39/ecma262/commit/13906140a


tcare:
Looks good, thanks for the update.

PR #810, opened 2016-12-24
Remove tests for @@iterator on module namespace objects

bakkot:
It has been [removed](https://github.com/tc39/ecma262/pull/747).

Instead just test that it is not present.


bakkot:
Strictly speaking I guess implementations are permitted to define it anyway as an extension, so maybe we shouldn't include the test that it doesn't exist - but I worry that if we don't, implementations which have added it will forget to remove it.

tcare:
I tend to agree. Looks good.

PR #812, opened 2016-12-27
Remove last usage of arguments.caller

bakkot:
I missed this in https://github.com/tc39/test262/pull/797.

I've checked more thoroughly now; I think this was the only one overlooked.


littledan:
LGTM

PR #813, opened 2016-12-29
Fix test for instantiating a class extending null

ajklein:
The sameValue assertion should be that the [[Prototype]] is
equal to the class's prototype, not the constructor.

### DIFF
--- a/test/language/statements/class/subclass/class-definition-null-proto-missing-return-override.js
+++ b/test/language/statements/class/subclass/class-definition-null-proto-missing-return-override.js
@@ -38,4 +38,4 @@ class Foo extends null {
 
 var foo = new Foo();
 
-assert.sameValue(Object.getPrototypeOf(foo), Foo);
+assert.sameValue(Object.getPrototypeOf(foo), Foo.prototype);


test/language/statements/class/subclass/class-definition-null-proto-missing-return-override.js
```diff
@@ -38,4 +38,4 @@ class Foo extends null {
 
 var foo = new Foo();
```
tcare: Most implementations seem to throw on this right now, because 'this' is a use-before-decl in 'constructor () {}'. Perhaps because they treat the kind of the constructor as 'derived' and not 'base', as per #sec-ecmascript-function-objects-construct-argumentslist-newtarget step 5.

Can you confirm that this is an implementation gap and this isn't meant to throw here? Otherwise I agree the assert should be on Foo.prototype.

bakkot: See [step 15 of ClassDefinitionEvaluation](https://tc39.github.io/ecma262/#sec-runtime-semantics-classdefinitionevaluation), which is the only place in the spec that a constructor is set to `"derived"`, and which does not apply to classes which extend `null`.

bakkot: (Constructible functions have [[ConstructorKind]] `"base"` by default per [FunctionAllocate 9.a](https://tc39.github.io/ecma262/#sec-functionallocate).)

bakkot: See https://github.com/tc39/ecma262/issues/543#issuecomment-210191577 for the discussion.

tcare: Ah! That's what I missed.

Yes that is my interpretation too. Thanks for the confirmation.

tcare:
Looks good

test/language/statements/class/subclass/class-definition-null-proto-missing-return-override.js
ajklein: Should have mentioned in the PR description: the modified version of this test passes with a local version of V8 I have in which "extends null" has the newly-specced behavior.

Issue #814, opened 2016-12-29
Missing test for mapped arguments property descriptor


caiolima:
Arguments object mapping rules are being well covered on test262, however there is no test case covering arguments property descriptor behavior. They are important, since it is not uncommon implement the mapping rules and forget about descriptor changes. One example of case that it can catch this kind of problem is:

```
function foo(x) {
    Object.defineProperty(arguments, 0, {configurable: true, enumerable: false, writable:true, value:42});
    return [x, arguments[0], arguments]
}

var result = foo(1);
if (Object.getOwnPropertyDescriptor(result[2], 0).enumerable === true)
    throw new Error();
```
In this case, the arguments is still mapped and in this case and the ```enumerable``` property can be ignored. We identified this behavior on WebKit JavaScriptCore implementation and fixed on https://bugs.webkit.org/show_bug.cgi?id=159398.

bakkot:
See also #729.

leobalter:
this is fixed by #815

PR #815, opened 2016-12-29
Adding test cases to validate property descriptors on cases of mapped arguments

caiolima:
Implementing test cases for issue #814

### DIFF
--- a/test/language/arguments-object/mapped/nonconfigurable-nonenumerable-nonwritable-descriptors-basic.js
+++ b/test/language/arguments-object/mapped/nonconfigurable-nonenumerable-nonwritable-descriptors-basic.js
@@ -1,0 +1,27 @@
+// Copyright (C) 2017 Caio Lima. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Mapped arguments property descriptor change with non-configurable, non-enumerable and non-writable property
+info: >
+    Mapping stop working when property is set to non-writable. The
+    descriptor's enumerable property is the one set before the mapping removal.
+flags: [noStrict]
+esid: sec-arguments-exotic-objects-defineownproperty-p-desc
+includes: [propertyHelper.js]
+---*/
+
+function fn(a) {
+  Object.defineProperty(arguments, "0", {configurable: false, enumerable: false, writable: false});
+
+  // Postcondition: Arguments mapping is removed.
+  a = 2;
+
+  let propertyDescriptor = Object.getOwnPropertyDescriptor(arguments, "0");
+  assert.sameValue(propertyDescriptor.value, 1);
+  verifyNotEnumerable(arguments, "0");
+  verifyNotWritable(arguments, "0");
+  verifyNotConfigurable(arguments, "0");
+}
+fn(1);
+
--- a/test/language/arguments-object/mapped/nonconfigurable-nonenumerable-nonwritable-descriptors-set-by-param.js
+++ b/test/language/arguments-object/mapped/nonconfigurable-nonenumerable-nonwritable-descriptors-set-by-param.js
@@ -1,0 +1,36 @@
+// Copyright (C) 2017 Caio Lima. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Mapped arguments property descriptor change to non-configurable, non-enumerable and non-writable
+info: >
+    Change the descriptor using [[DefineOwnProperty]] to
+    {configurable: false, enumerable: false}, set a = 2 and then
+    change property descriptor to {writable: false}. The descriptor's enumerable
+    property need to be the one set before the mapping removal.
+flags: [noStrict]
+esid: sec-arguments-exotic-objects-defineownproperty-p-desc
+includes: [propertyHelper.js]
+---*/
+
+function fn(a) {
+  Object.defineProperty(arguments, "0", {configurable: false, enumerable: false});
+  a = 2;
+  Object.defineProperty(arguments, "0", {writable: false});
+
+  let propertyDescriptor = Object.getOwnPropertyDescriptor(arguments, "0");
+  assert.sameValue(propertyDescriptor.value, 2);
+  verifyNotEnumerable(arguments, "0");
+  verifyNotWritable(arguments, "0");
+  verifyNotConfigurable(arguments, "0");
+
+  // Postcondition: Arguments mapping is removed.
+  a = 3;
+  propertyDescriptor = Object.getOwnPropertyDescriptor(arguments, "0");
+  assert.sameValue(propertyDescriptor.value, 2);
+  verifyNotEnumerable(arguments, "0");
+  verifyNotWritable(arguments, "0");
+  verifyNotConfigurable(arguments, "0");
+}
+fn(1);
+
--- a/test/language/arguments-object/mapped/nonwritable-nonconfigurable-descriptors-basic.js
+++ b/test/language/arguments-object/mapped/nonwritable-nonconfigurable-descriptors-basic.js
@@ -1,0 +1,36 @@
+// Copyright (C) 2017 Caio Lima. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Mapped arguments property descriptor change to non-writable and non-configurable
+info: >
+    Mapping stop working when property is set to non-writable. Here we change the
+    descriptor using [[DefineOwnProperty]] to {writable: false} and then
+    change property descriptor to {configurable: false} in sequence.
+    The property descriptor need to be the one set before the property be
+    configured as {writable: false}.
+flags: [noStrict]
+esid: sec-arguments-exotic-objects-defineownproperty-p-desc
+includes: [propertyHelper.js]
+---*/
+
+function fn(a) {
+  Object.defineProperty(arguments, "0", {writable: false});
+  Object.defineProperty(arguments, "0", {configurable: false});
+  let propertyDescriptor = Object.getOwnPropertyDescriptor(arguments, "0");
+  assert.sameValue(propertyDescriptor.value, 1);
+  verifyEnumerable(arguments, "0");
+  verifyNotWritable(arguments, "0");
+  verifyNotConfigurable(arguments, "0");
+
+  // Postcondition: Arguments mapping is removed. Descriptors need to be the same
+  // as above.
+  a = 2;
+  propertyDescriptor = Object.getOwnPropertyDescriptor(arguments, "0");
+  assert.sameValue(propertyDescriptor.value, 1);
+  verifyEnumerable(arguments, "0");
+  verifyNotWritable(arguments, "0");
+  verifyNotConfigurable(arguments, "0");
+}
+fn(1);
+
--- a/test/language/arguments-object/mapped/nonwritable-nonconfigurable-descriptors-set-by-arguments.js
+++ b/test/language/arguments-object/mapped/nonwritable-nonconfigurable-descriptors-set-by-arguments.js
@@ -1,0 +1,37 @@
+// Copyright (C) 2017 Caio Lima. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Mapped arguments property descriptor change to non-writable and non-configurable
+info: >
+    Mapping stop working when property is set to non-writable. Change the
+    descriptor using [[DefineOwnProperty]] to {writable: false}, set argument[0] = 2 and then
+    change property descriptor to {configurable: false}.
+    The descriptor's value is the one set before the property be
+    configured as {writable: false} because mapping was removed.
+flags: [noStrict]
+esid: sec-arguments-exotic-objects-defineownproperty-p-desc
+includes: [propertyHelper.js]
+---*/
+
+function fn(a) {
+  Object.defineProperty(arguments, "0", {writable: false});
+  arguments[0] = 2;
+  Object.defineProperty(arguments, "0", {configurable: false});
+  let propertyDescriptor = Object.getOwnPropertyDescriptor(arguments, "0");
+  assert.sameValue(propertyDescriptor.value, 1);
+  assert.sameValue(a, 1);
+  verifyEnumerable(arguments, "0");
+  verifyNotWritable(arguments, "0");
+  verifyNotConfigurable(arguments, "0");
+
+  // Postcondition: Arguments mapping is removed.
+  a = 3;
+  propertyDescriptor = Object.getOwnPropertyDescriptor(arguments, "0");
+  assert.sameValue(propertyDescriptor.value, 1);
+  verifyEnumerable(arguments, "0");
+  verifyNotWritable(arguments, "0");
+  verifyNotConfigurable(arguments, "0");
+}
+fn(1);
+
--- a/test/language/arguments-object/mapped/nonwritable-nonconfigurable-descriptors-set-by-param.js
+++ b/test/language/arguments-object/mapped/nonwritable-nonconfigurable-descriptors-set-by-param.js
@@ -1,0 +1,28 @@
+// Copyright (C) 2017 Caio Lima. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Mapped arguments property descriptor change to non-writable and non-configurable
+info: >
+    Mapping just stop working when property is set to non-writable. Change the
+    descriptor using [[DefineOwnProperty]] to {writable: false}, set a = 2 and then
+    change property descriptor to {configurable: false}.
+    The descriptor's value is the one set before the property be
+    configured as {writable: false} because mapping was removed.
+flags: [noStrict]
+esid: sec-arguments-exotic-objects-defineownproperty-p-desc
+includes: [propertyHelper.js]
+---*/
+
+function fn(a) {
+  Object.defineProperty(arguments, "0", {writable: false});
+  a = 2;
+  Object.defineProperty(arguments, "0", {configurable: false});
+  let propertyDescriptor = Object.getOwnPropertyDescriptor(arguments, "0");
+  assert.sameValue(propertyDescriptor.value, 1);
+  verifyEnumerable(arguments, "0");
+  verifyNotWritable(arguments, "0");
+  verifyNotConfigurable(arguments, "0");
+}
+fn(1);
+
--- a/test/language/arguments-object/mapped/nonwritable-nonenumerable-nonconfigurable-descriptors-set-by-arguments.js
+++ b/test/language/arguments-object/mapped/nonwritable-nonenumerable-nonconfigurable-descriptors-set-by-arguments.js
@@ -1,0 +1,34 @@
+// Copyright (C) 2017 Caio Lima. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Mapped arguments property descriptor change to non-writable, non-enumerable and non-configurable
+info: >
+    Change the descriptor using [[DefineOwnProperty]] to {writable: false, enumerable: false},
+    set argument[0] = 2 and then change property descriptor to {configurable: false}.
+    The descriptor's enumerable property continues with its configured value.
+flags: [noStrict]
+esid: sec-arguments-exotic-objects-defineownproperty-p-desc
+includes: [propertyHelper.js]
+---*/
+
+function fn(a) {
+  Object.defineProperty(arguments, "0", {writable: false, enumerable: false});
+  arguments[0] = 2;
+  Object.defineProperty(arguments, "0", {configurable: false});
+
+  let propertyDescriptor = Object.getOwnPropertyDescriptor(arguments, "0");
+  assert.sameValue(propertyDescriptor.value, 1);
+  verifyNotEnumerable(arguments, "0");
+  verifyNotWritable(arguments, "0");
+  verifyNotConfigurable(arguments, "0");
+
+  // Postcondition: Arguments mapping is removed.
+  a = 3;
+  propertyDescriptor = Object.getOwnPropertyDescriptor(arguments, "0");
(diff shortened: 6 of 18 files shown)


test/language/arguments-object/mapped/mapped-arguments-nonconfigurable-descriptors-4.js
```diff
@@ -0,0 +1,23 @@
...
+    value is changed using arguments[i] where 'i' is the argument index.
+flags: [noStrict]
+---*/
+
+function argumentsAndSetByIndex(a) {
+  Object.defineProperty(arguments, "0", {configurable: false});
+
+  arguments[0] = 2;
+  let propertyDescriptor = Object.getOwnPropertyDescriptor(arguments, '0');
+  assert.sameValue(propertyDescriptor.value, 2);
```
bakkot: This should also do `assert.sameValue(a, 2)`, I should think.

test/language/arguments-object/mapped/mapped-arguments-nonconfigurable-nonenumerable-nonwritable-descriptors-4.js
```diff
@@ -0,0 +1,32 @@
...
+    The descriptor's enumerable property is the one set before the mapping removal.
+flags: [noStrict]
+---*/
+
+function argumentsNonConfigurableNonEnumerableThenNonWritableWithInterveningSetByIndex(a) {
+  Object.defineProperty(arguments, "0", {configurable: false, enumerable: false});
+  arguments[0] = 2;
+  Object.defineProperty(arguments, "0", {writable: false});
+  let propertyDescriptor = Object.getOwnPropertyDescriptor(arguments, '0');
+  assert.sameValue(propertyDescriptor.value, 2);
```
bakkot: I would add `assert.sameValue(a, 2)` here too.

bakkot:
I haven't read them all over carefully, but these all pass with latest Chrome (per [web runner](https://bakkot.github.io/test262-web-runner/)).

I think Chrome's behavior is correct, so I'm pretty confident these tests are good.

test/language/arguments-object/mapped/mapped-arguments-nonconfigurable-descriptors-4.js
caiolima: I was in doubt to add it, since it is already being tested on test/language/arguments-object/mapped/mapped-arguments-nonconfigurable-4.js

I'm also in doubt if these cases should be added as a new JS test cases or if we can just add descriptor asserts on current tests.

caiolima:
Updated commit based on @bakkot suggestions.

leobalter:
Hi! The tests are good, but I'd like to add some suggestions:

In cases you're checking the descriptors, there's a few [helper functions](https://github.com/tc39/test262/blob/master/harness/propertyHelper.js) we use to assert the descriptor matches reality. It would be good to replace these assertions with the existent helpers, example:

From:
```js
let propertyDescriptor = Object.getOwnPropertyDescriptor(arguments, '0');
assert.sameValue(propertyDescriptor.value, 1);
assert(!propertyDescriptor.configurable);
assert(propertyDescriptor.enumerable);
assert(propertyDescriptor.writable);
```

to:
```js
assert.sameValue(arguments['0'], 1);
verifyEnumerable(arguments, '0');
verifyWritable(arguments, '0');
verifyNotConfigurable(arguments, '0');
```

Still, I kind like when you assert the value from the descriptor, but that can be solved in a single file.

With these changes, we assert some extra behaviors as throwing a type error trying to delete a configurable property. Mapped `arguments` is exotic enough to be worth of this free extra.

leobalter:
One other thing the tests are missing is the [frontmatter `esid`](https://github.com/tc39/test262/blob/master/CONTRIBUTING.md#esid) tags, this should be a quick addition.

In the meanwhile, please [sign the CLA](http://tc39.github.io/test262-cla/).

I look forward on seeing the updates and merging this.

🎆 🎆

caiolima:
@leobalter updated PR following almost all of your comments. However, I didn't follow the case to assert ```arguments[0]``` value because I didn't understand your comment.
The reason that I'm not testing it in these new tests is because the current tests are verifying that already.

BTW, I've already signed the CLA.

leobalter:
There's a test case I'm missing:

```js
function fn(a, b) {
  Object.defineProperty(arguments, '0', { get() { return 42 } })
  console.log(a, arguments[0]) // should be 0, 42

  Object.defineProperty(arguments, '0', { value: 74 })
  console.log(a, arguments[0]) // should be 0, 74

  Object.defineProperty(arguments, '1', { value: 1 })
  console.log(b, arguments[1]) // should be 1, 1

  arguments[0] = 'foo';
  console.log(a, arguments[0]) // should be 0, 74 (get is preserved)

  arguments[1] = 'bar';
  console.log(b, arguments[1]) // should be 'bar', 'bar'
}

fn(0, 39)
```

Where after defining an Accessor Descriptor unmap that property only, preserving the other mapped values.

See [`[[DefineOwnProperty]]`](https://tc39.github.io/ecma262/#sec-arguments-exotic-objects-defineownproperty-p-desc) __8.a.i__

test/language/arguments-object/mapped/mapped-arguments-nonwritable-nonconfigurable-descriptors-1.js
```diff
@@ -0,0 +1,35 @@
...
+---*/
+
+function argumentsNonWritableThenNonConfigurableDescriptor(a) {
+  Object.defineProperty(arguments, "0", {writable: false});
+  Object.defineProperty(arguments, "0", {configurable: false});
+  let propertyDescriptor = Object.getOwnPropertyDescriptor(arguments, '0');
+  assert.sameValue(propertyDescriptor.value, 1);
+  verifyEnumerable(arguments, '0');
+  verifyNotWritable(arguments, '0');
+  verifyNotConfigurable(arguments, '0');
```
leobalter: lines 19 to 24 are unnecessary for this test case, just repeating what you should have in another test case (defining to non configurable).

There's a more interesting case here that you can assert the parameter is set to the value if you provide it along with the `writable: false` descriptor, and it becomes unmapped after that. So for example: 

```js
Object.defineProperty(arguments, "0", {writable: false, value: 42});
assert.sameValue(a, 42, 'value is set');

a = 39;
assert.sameValue(a, 39, 'it is still writable');
assert.sameValue(arguments[0], 42, 'and it is unmapped');

arguments[0] = 'foo';
assert.sameValue(a, 39, 'really unmapped');
assert.sameValue(arguments[0], 42, 'non-writable for sure');

...
```

test/language/arguments-object/mapped/mapped-arguments-nonwritable-nonconfigurable-descriptors-2.js
```diff
@@ -0,0 +1,36 @@
...
+
+  // Postcondition: Arguments mapping is removed.
+  a = 3;
+  propertyDescriptor = Object.getOwnPropertyDescriptor(arguments, '0');
+  assert.sameValue(propertyDescriptor.value, 1);
+  verifyEnumerable(arguments, '0');
+  verifyNotWritable(arguments, '0');
+  verifyNotConfigurable(arguments, '0');
+}
+argumentsNonWritableThenNonConfigurableWithInterveningSetMutableBindingDescriptor(1);
```
leobalter: this will become a duplicate of `arguments-object/mapped/mapped-arguments-nonwritable-nonconfigurable-descriptors-1.js` if you remove the configurable part.

test/language/arguments-object/mapped/mapped-arguments-nonwritable-nonenumerable-nonconfigurable-descriptors-4.js
```diff
@@ -0,0 +1,39 @@
...
+    The descriptor's enumerable property continues with its configured value.
+flags: [noStrict]
+esid: sec-arguments-exotic-objects-defineownproperty-p-desc
+includes: [propertyHelper.js]
+---*/
+
+function argumentsNonWritableNonEnumerableThenNonConfigurableWithInterveningDefineOwnPropertyDescriptor(a) {
+  Object.defineProperty(arguments, "0", {writable: false, enumerable: false});
+  Object.defineProperty(arguments, "0", {value: 2});
+  Object.defineProperty(arguments, "0", {configurable: false});
```
leobalter: this test, as others, can be simplified to get to the point you really want to test here.

The goal here is apparently to assert the final descriptor, the same mapping parts is already asserted in other test cases.

leobalter: this is also a duplicate of `mapped-arguments-nonwritable-nonenumerable-nonconfigurable-descriptors-3.js` where there you actually asserting for non-writable set, more than non-enumerable.

test/language/arguments-object/mapped/mapped-arguments-nonwritable-nonenumerable-nonconfigurable-descriptors-2.js
```diff
@@ -0,0 +1,35 @@
+// Copyright (C) 2017 Caio Lima. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+info: Mapped arguments property descriptor change to non-writable, non-enumerable and non-configurable
+description: >
+    Change the descriptor using [[DefineOwnProperty]] to
+    {writable: false, enumerable: false}, set a = 2 and then
+    change property descriptor to {configurable: false}.
+    The descriptor's enumerable property cotinue with its configured value.
```
leobalter: typo: cotinue

test/language/arguments-object/mapped/mapped-arguments-nonwritable-nonconfigurable-descriptors-4.js
```diff
@@ -0,0 +1,39 @@
...
+    change property descriptor to {configurable: false}.
+flags: [noStrict]
+esid: sec-arguments-exotic-objects-defineownproperty-p-desc
+includes: [propertyHelper.js]
+---*/
+
+function argumentsNonWritableThenNonConfigurableWithInterveningDefineOwnPropertyDescriptor(a) {
+  Object.defineProperty(arguments, "0", {writable: false});
+  Object.defineProperty(arguments, "0", {value: 2});
+  Object.defineProperty(arguments, "0", {configurable: false});
```
leobalter: for clarity, this case could be exposed in a more simple test for better clarity:

```js
function fn(a) {
   Object.defineProperty(arguments, "0", {writable: false, enumerable: false});
   Object.defineProperty(arguments, "0", {value: 2});

   assert.sameValue(arguments[0], 2, 'a value can be defined to a non-writable, configurable prop');
   
   arguments[0] = 42;

   console.log(arguments[0], 2, 'a new value cannot be set to a non-writable prop');
}

fn(1);
```

And also split to a new file:

```js
function fn(a) {
   Object.defineProperty(arguments, "0", {writable: false, configurable: false});
    Object.defineProperty(arguments, "0", {value: 2});
  });
}

fn(1);
```

leobalter: The problem here is where it's really asserting the Mapping behaviors. So I would say this is a bit out of scope.

caiolima: Hum...I think this test case is different from the sample that you advised. In this case I want to verify if arguments[0].descriptor.value is updated even if {writable: false}. Doing Object.defineProperty(arguments, "0", {writable: false, enumerable: false}); makes it not viable.

caiolima: Never mind. I got your comment now.

caiolima:
@leobalter and @rwaldron updated following your comments.

The test case proposed by @leobalter (I named it ```arguments-object/mapped/nonwritable-nonenumerable-descriptor-set-by-define-property.js```) is failing in my last test262-harness on line 24. If I'm not wrong, the test is correct, since there is no behavior described on [[[DefineOwnProperty]]](https://tc39.github.io/ecma262/#sec-arguments-exotic-objects-defineownproperty-p-desc) 8.a.i specifying that property is changed to ```{writable: true}``` when ```value``` is set. Could you confirm that?

bakkot:
@caiolima, the test is wrong in a really subtle way: `verifyConfigurable` calls `isConfigurable` which [just deletes the property](https://github.com/tc39/test262/blob/4ff7e4342e1a15ce7af893ba9af9bd55c7d95e51/harness/propertyHelper.js#L4), after which the write will succeed.

I see three possible fixes:
- make `isConfigurable` restore the state of the property, like `isWritable` does
- don't use `verifyConfigurable`
- save all the property descriptors and only validate them after all steps have been performed.

caiolima:
@bakkot Thank you for that explanation and nice catch!

Another alternative that I have in mind is to create a new helper functions named ```verifyConfigurableNonDestructible ``` that just checks```propertyDescriptor.configurable``` value. What do you think?

bakkot:
@caiolima - I think I'd prefer fixing `isConfigurable` somehow (as an alternative to setting the property to a data or accessor property, it could also e.g. toggle the property's enumerability, and could even only make use of these alternatives when the object is not extensible), but let's see what @leobalter thinks.

rwaldron:
cc @smikes w/r to @bakkot's question about `isConfigurable`. I went back and read https://github.com/tc39/test262/pull/72#issuecomment-51000784, but any additional insight would be appreciated :)

leobalter:
> I see three possible fixes:
> - make isConfigurable restore the state of the property, like isWritable does
> - don't use verifyConfigurable
> - save all the property descriptors and only validate them after all steps have been performed.

I wouldn't change isConfigurable to restore the state of the property, this would case extra logic we don't really need and might throw errors elsewhere.

We have a specific point where we want to assert a value change after checking the configurability.

The third option is closest of what I believe it's best, splitting the test in two groups. This is easy with arguments:

```js
function fn(a, b) {
  Object.defineProperty(arguments, '0', {writable: false, enumerable: false});
  Object.defineProperty(arguments, '0', {value: 42});

  assert.sameValue(arguments[0], 42);
  let propertyDescriptor = Object.getOwnPropertyDescriptor(arguments, '0');
  assert.sameValue(propertyDescriptor.value, 42);
  verifyNotEnumerable(arguments, '0');
  verifyNotWritable(arguments, '0');
  verifyConfigurable(arguments, '0');
  
  // Repeat the same config from above, now using another argument
  Object.defineProperty(arguments, '1', {writable: false, enumerable: false});
  Object.defineProperty(arguments, '1', {value: 39});
  arguments[1] = 42;

  propertyDescriptor = Object.getOwnPropertyDescriptor(arguments, '1');
  assert.sameValue(propertyDescriptor.value, 39);
  verifyNotEnumerable(arguments, '1');
  verifyNotWritable(arguments, '1');
  verifyConfigurable(arguments, '1');
}

fn(1, 2);
```

What is still not clear, is the main goal of this test. It looks like the whole part I tested for arguments[0] above is irrelevant to this current test.

leobalter:
> Another alternative that I have in mind is to create a new helper functions named verifyConfigurableNonDestructible that just checkspropertyDescriptor.configurable value. 

That's too loose when the problem is that we need to have a specific assertion goal. It's good to not only confirm the descriptor value but if it's in effect.

leobalter:
There are some mixed `"` and `'`, this one should be easy to solve doing a folder search to replace.

caiolima:
> What is still not clear, is the main goal of this test. It looks like the whole part I tested for arguments[0] above is irrelevant to this current test.

You are right. Just decided to remove this test case.

About one case missing mentioned by you above, I think it is good idea work on another PR. I've created the issue already (it is #817).

Sorry for the noise on this PR, I realized that I didn't remove the numbered tests =( Fixing it now and uploading a new version.

caiolima:
Ok guys, I updated and squashed test case following your comments and reviews.

I decided to assert ```descriptor.value``` in all cases because ```arguments[0]``` is already being asserted in other test cases.
Also, I decided to remove ```arguments-object/mapped/nonwritable-nonenumerable-descriptor-set-by-define-property.js``` since I couldn't explain its relevance.

caiolima:
@rwaldron @bakkot @leobalter updated PR based on comments.

smikes:
The isConfigurable test is deliberately destructive, per Brian's request 

> We need to always test that the descriptor is accurately reflected in the actions you can take on an object. One or two tests would make sense if implementations were guaranteed to always use the same codepaths for different types of objects, but this is not the case unfortunately - perf optimizations especially can create strange artifacts seemingly random places. Destruction doesn't seem to be too big of an issue to me.

I wasn't enthused about this decision, but Brian's probably right.

leobalter:
visually it looks good to me. For the peace of mind I'll review these tests running tomorrow (Monday) in a proper environment.

test/language/arguments-object/mapped/nonconfigurable-nonenumerable-nonwritable-descriptors-basic.js
```diff
@@ -0,0 +1,27 @@
...
+info: >
+    Mapping stop working when property is set to non-writable. The
+    descriptor's enumerable property is the one set before the mapping removal.
+flags: [noStrict]
+esid: sec-arguments-exotic-objects-defineownproperty-p-desc
+includes: [propertyHelper.js]
+---*/
+
+function fn(a) {
+  Object.defineProperty(arguments, "0", {configurable: false, enumerable:false, writable: false});
```
bakkot: Nit: space after `enumerable:`

test/language/arguments-object/mapped/nonconfigurable-nonenumerable-nonwritable-descriptors-set-by-param.js
```diff
@@ -0,0 +1,36 @@
+// Copyright (C) 2017 Caio Lima. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Mapped arguments property descriptor change to non-configurable, non-enumerable and non-writable
+info: >
+    Change the descriptor using [[DefineOwnProperty]] to
+    {configurable: false, enumerable: false}, set a = 2 and then
+    change property descriptor to {writable: false}. The descriptor'ss enumerable
```
bakkot: Nit: "descriptor'ss" -> "descriptor's"

test/language/arguments-object/mapped/nonwritable-nonconfigurable-descriptors-basic.js
```diff
@@ -0,0 +1,36 @@
...
+function fn(a) {
+  Object.defineProperty(arguments, "0", {writable: false});
+  Object.defineProperty(arguments, "0", {configurable: false});
+  let propertyDescriptor = Object.getOwnPropertyDescriptor(arguments, "0");
+  assert.sameValue(propertyDescriptor.value, 1);
+  verifyEnumerable(arguments, "0");
+  verifyNotWritable(arguments, "0");
+  verifyNotConfigurable(arguments, "0");
+
+  // Postcondition: Arguments mapping is removed. Descritors need to be the same
```
bakkot: Nit: "Descritors" -> "Descriptors"

test/language/arguments-object/mapped/nonwritable-nonconfigurable-descriptors-set-by-arguments.js
```diff
@@ -0,0 +1,37 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Mapped arguments property descriptor change to non-writable and non-configurable
+info: >
+    Mapping stop working when property is set to non-writable. Change the
+    descriptor using [[DefineOwnProperty]] to {writable: false}, set argument[0] = 2 and then
+    change property descriptor to {configurable: false}.
+    The descriptor's value is the one set before the property be
+    configured as {writable: false} beacuse mapping was removed.
```
bakkot: Nit: "beacuse" -> "because"

test/language/arguments-object/mapped/nonwritable-nonenumerable-nonconfigurable-descriptors-set-by-arguments.js
```diff
@@ -0,0 +1,34 @@
+// Copyright (C) 2017 Caio Lima. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Mapped arguments property descriptor change to non-writable, non-enumerable and non-configurable
+info: >
+    Change the descriptor using [[DefineOwnProperty]] to {writable: false, enumerable: false},
+    set argument[0] = 2 and then  change property descriptor to {configurable: false}.
```
bakkot: Nit: two spaces between "then" and "change"

test/language/arguments-object/mapped/nonwritable-nonconfigurable-descriptors-set-by-param.js
```diff
@@ -0,0 +1,28 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Mapped arguments property descriptor change to non-writable and non-configurable
+info: >
+    Mapping just stop working when property is set to non-writable. Change the
+    descriptor using [[DefineOwnProperty]] to {writable: false}, set a = 2 and then
+    change property descriptor to {configurable: false}.
+    The descriptor's value is the one set before the property be
+    configured as {writable: false} beacuse mapping was removed.
```
bakkot: Nit: "beacuse" -> "because"

leobalter:
@caiolima I'm merging it after the nit picks above are fixed.

caiolima:
@leobalter @rwaldron @bakkot updated to fix NITs. Thank you very much for the review!

caiolima:
@leobalter @rwaldron @bakkot do I need to do some other update?

leobalter:
It's all set now! Thanks!

caiolima:
Thank you very much!

PR #820, opened 2017-01-09
Function.prototype.toString no longer does line terminator normalisation

michaelficarra:
See https://github.com/tc39/Function-prototype-toString-revision/pull/22.


tcare:
Looks good

PR #836, opened 2017-01-20
Test for case-insensitivity in invalid BCP 47 usage

littledan:
This patch adds a regression test for a previous V8 bug reported originally at
https://bugs.chromium.org/p/v8/issues/detail?id=4215


tcare:
Looks good

PR #840, opened 2017-01-24
Fix various test issues

anba:
test/annexB/built-ins/Date/prototype/setYear/time-clip.js
test/built-ins/Date/prototype/setFullYear/new-value-time-clip.js
test/built-ins/Date/prototype/setMonth/new-value-time-clip.js
- Don't try to test time-clip at the end points, because this is near
impossible to get right (needs to consider time zone offset, dst, local
mean time because of Africa/Monrovia, etc.).

test/built-ins/DataView/prototype/setFloat64/detached-buffer-after-toindex-byteoffset.js
test/built-ins/DataView/prototype/setInt16/detached-buffer-after-toindex-byteoffset.js
- Wasn't update to expect RangeError

test/built-ins/Function/internals/Construct/derived-this-uninitialized-realm.js
- Change ClassDeclaration -> ClassExpression to get completion value

test/built-ins/Function/prototype/toString/AsyncFunction.js
- Add missing \n in expected string

test/built-ins/global/global-object.js
- Add 'var' to make test pass in strict-mode

test/language/block-scope/syntax/redeclaration-in-block/attempt-to-redeclare-function-declaration-with-function-declaration.js
- This is allowed in sloppy mode when Annex B is implemented

test/language/expressions/async-generators/expression-yield-as-statement.js
- Fix calls to then()

test/language/module-code/namespace/internals/own-property-keys-binding-types.js
test/language/module-code/namespace/internals/own-property-keys-sort.js
- Tests weren't updated after removal of @@iterator from module
namespace objects

test/language/module-code/namespace/internals/set-prototype-of-null.js
- Fix syntax error

test/language/statements/async-function/early-errors-no-async-generator.js
- No longer valid now that async iteration proposal is at stage 3


anba:
Two fixes are now duplicated by #842 and #847.

leobalter:
LGTM

leobalter:
This is great! Thanks, @anba!

littledan:
Thanks for this fix; I was about to report the second to last one.

PR #841, opened 2017-01-25
Add tests for known violation of invariants v2

### LINKED ISSUE #649: Are there tests against legacy magic properties that are nonconfigurable and nonwritable?
According to [6.1.7.3](https://tc39.github.io/ecma262/#sec-invariants-of-the-essential-internal-methods)

<blockquote>
     If a property P is described as a data property with Desc.[[Value]] equal to v and Desc.[[Writable]] and Desc.[[Configurable]] are both false, then the SameValue must be returned for the Desc.[[Value]] attribute of the property on all future calls to [[GetOwnProperty]] ( P ).
</blockquote>

However, as of today, many implementations define some legacy magic properties as nonconfigurable and nonwritable, but keep them changing their value over time, e.g.:
- `f.arguments` and `f.caller` for sloppy function `f` (bug observed in WebKit, V8 and Chakra);
- `RegExp.$1`, etc. (bug observed in WebKit).

Are there tests against these cases.

Suggested sample test:

``` js
function f() { return Object.getOwnPropertyDescriptor(f, 'arguments'); }

try { 
    Object.defineProperty(f, 'arguments', { writable: false, configurable: false });
}
catch (e) { }

var desc = Object.getOwnPropertyDescriptor(f, 'arguments');
if (desc && desc.configurable === false && desc.writable === false) {
    var desc2 = f();
    assert.sameValue(desc.value, desc2.value);
}
```

claudepache:
Subsumes #653 (issue with copyright/licence).
Resolves #649.
Also, add a test against a observed bug in Chakra: RegExp.$1 claims to successfully become non-writable, non-configurable, but remains writable.

### DIFF
--- a/test/built-ins/Object/internals/DefineOwnProperty/consistent-value-function-arguments.js
+++ b/test/built-ins/Object/internals/DefineOwnProperty/consistent-value-function-arguments.js
@@ -1,0 +1,31 @@
+// Copyright (C) 2017 Claude Pache. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-invariants-of-the-essential-internal-methods
+es6id: 6.1.7.3
+description: >
+  Value of non-writable, non-configurable data property must not change
+  ("arguments" property of a non-strict function)
+info: |
+  [[GetOwnProperty]] (P)
+  [...]
+  - If a property P is described as a data property with Desc.[[Value]] equal
+    to v and Desc.[[Writable]] and Desc.[[Configurable]] are both false, then
+    the SameValue must be returned for the Desc.[[Value]] attribute of the
+    property on all future calls to [[GetOwnProperty]] ( P ).
+  [...]
+  (This invariant was violated for the specific property under test by a number
+  of implementations as of January 2017.)
+---*/
+
+function f() {
+    return Reflect.getOwnPropertyDescriptor(f, 'arguments');
+}
+
+Reflect.defineProperty(f, 'arguments', { writable: false, configurable: false });
+
+var desc = Reflect.getOwnPropertyDescriptor(f, 'arguments');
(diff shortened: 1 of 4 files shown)


test/built-ins/Object/internals/DefineOwnProperty/consistent-value-function-arguments.js
```diff
@@ -0,0 +1,31 @@
...
+  [...]
+  (This invariant was violated for the specific property under test by a number
+  of implementations as of January 2017.)
+---*/
+
+function f() {
+    return Reflect.getOwnPropertyDescriptor(f, 'arguments');
+}
+
+Reflect.defineProperty(f, 'arguments', { writable: false, configurable: false });
```
littledan: Wouldn't it be valid for an implementation to throw here?

littledan: Never mind, I see that this is specified as configurable.

littledan:
These all seem valid to me, LGTM.

test/built-ins/Object/internals/DefineOwnProperty/consistent-value-function-arguments.js
claudepache: > Wouldn't it be valid for an implementation to throw here?

`Reflect.defineProperty` does not throw, but returns false when failing. In all these tests, an implementation may choose to refuse to make the property non-writable/non-configurable, which is a valid option as far as these tests are concerned.

leobalter:
The tests are fine, my only remaining concern is if the `test/built-ins/Object/internals/DefineOwnProperty` is the right folder for them and not `test/built-ins/Object/internals/GetOwnProperty`. 

This PR is open for so long I don't think it's good to keep holding it, so I'm merging this and we can follow up with a fix moving the files to the right folder if necessary.

PR #842, opened 2017-01-27
Fix test for yield in async generator

arai-a:
the test wasn't checking the resolved value.


test/language/expressions/async-generators/expression-yield-as-statement.js
```diff
@@ -19,7 +19,7 @@ iter1.next().then(function(result) {
   assert.sameValue(
     result.done, false, "Without right-hand-side: first result `done` flag");
 }).then(undefined, $DONE);
-iter1.next(function(result) {
+iter1.next().then(function(result) {
```
caitp: Oops, good catch. Lgtm

leobalter:
This is a duplicate from #840. Thanks for the PR.

PR #843, opened 2017-01-28
Add tests for accessor names

jugglinmike:
These tests for accessor names were originally submitted in gh-706, but that
change set only concerned their use within object initializers. At @avdg's
recommendation, I have expanded coverage to include class definitions. Because
this generally concerns four additional syntactic forms (instance accessors
within class expressions, static accessors within class expressions, instance
accessors within class declarations, and static accessors within class
declarations), I've made use of the test generation tool to promote consistent
coverage. Not all tests were suitable subjects for procedural generation,
however (there is only one interesting case for the use of the `in` keyword,
for example, making the "case/template" formatting needlessly indirect), so
this patch also includes some traditional non-generated test material. In all
cases, file meta-data and license information accurately describes the
provenance.

### DIFF
--- a/src/accessor-names/literal-string-double-quote.case
+++ b/src/accessor-names/literal-string-double-quote.case
@@ -1,0 +1,21 @@
+// Copyright (C) 2017 Mike Pennisi. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+desc: Computed values as accessor property names (string literal using double quotes)
+template: default
+info: |
+  12.2.6.7 Runtime Semantics: Evaluation
+
+  [...]
+
+  ComputedPropertyName : [ AssignmentExpression ]
+
+  1. Let exprValue be the result of evaluating AssignmentExpression.
+  2. Let propName be ? GetValue(exprValue).
+  3. Return ? ToPropertyKey(propName).
+---*/
+
+//- declareWith
+"singleQuote"
+//- referenceWith
+"singleQuote"
--- a/src/accessor-names/literal-string-double-quote.case
+++ b/src/accessor-names/literal-string-double-quote.case
@@ -16,6 +16,6 @@ info: |
 ---*/
 
 //- declareWith
-"singleQuote"
+"doubleQuote"
 //- referenceWith
-"singleQuote"
+"doubleQuote"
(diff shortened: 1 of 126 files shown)


src/accessor-names/literal-string-double-quote.case
```diff
@@ -0,0 +1,21 @@
...
+
+  ComputedPropertyName : [ AssignmentExpression ]
+
+  1. Let exprValue be the result of evaluating AssignmentExpression.
+  2. Let propName be ? GetValue(exprValue).
+  3. Return ? ToPropertyKey(propName).
+---*/
+
+//- declareWith
+"singleQuote"
```
gnarf: Is this just a beautiful troll? or did you mean `"doubleQuote"`

test/language/expressions/class/accessor-name-static-computed-yield-expr.js
```diff
@@ -0,0 +1,43 @@
...
+  2. Let propName be ? GetValue(exprValue).
+  3. Return ? ToPropertyKey(propName).
+features: [generators]
+---*/
+
+var yieldSet, C, iter;
+function* g() {
+  C = class {
+    static get [yield]() { return 'get yield'; }
+    static set [yield](param) { yieldSet = param; }
```
gnarf: I want to like frame this one in the next sideshow blogpost...

gnarf:
Overall 👍  they all pass locally on the node harness even!

test/language/expressions/class/accessor-name-static-computed-yield-expr.js
jugglinmike: :)

jugglinmike: :)

src/accessor-names/literal-string-double-quote.case
jugglinmike: Yup, that was an oversight on my part. Fixed

jugglinmike:
Thanks for the review, Corey!

littledan:
+1 happy to have this extra test coverage; as @bakkot can tell you, the cases related to contextual keywords and will only get worse with future proposals; @anba has found bugs in V8 related to using yield in weird places like this. So this is all great to have these systematic tests.

leobalter:
LGTM.

Merged at 228851fbe and 37d5f8e5e

leobalter:
I really loved using `yield` to create dynamic accessor names in classes. It should be the new factory pattern.

PR #844, opened 2017-01-30
Add some tests for execution order for yield* in async generator

arai-a:
I'm about to write tests for yield* in async generator.
Here's tests for some basic flow, about execution order.
can I get some feedback before start writing more tests for abrupt cases?

### DIFF
--- a/test/language/statements/async-generator/yield-star-async-next.js
+++ b/test/language/statements/async-generator/yield-star-async-next.js
@@ -1,0 +1,244 @@
+/*---
+ author: Tooru Fujisawa [:arai] <arai_a@mac.com>
+ esid: pending
+ description: execution order for yield* with async iterator and next()
+ info: >
+    YieldExpression: yield * AssignmentExpression
+
+    ...
+    2. Let value be ? GetValue(exprRef).
+    3. Let generatorKind be ! GetGeneratorKind().
+    4. Let iterator be ? GetIterator(value, generatorKind).
+    5. Let received be NormalCompletion(undefined).
+    6. Repeat
+      a. If received.[[Type]] is normal, then
+        i. Let innerResult be ? IteratorNext(iterator, received.[[Value]]).
+        ii. Let innerResult be ? Invoke(iterator, "next",
+            « received.[[Value]] »).
+        iii. If generatorKind is async, then set innerResult to
+             ? Await(innerResult).
+        ...
+        v. Let done be ? IteratorComplete(innerResult).
+        vi. If done is true, then
+           1. Return ? IteratorValue(innerResult).
+        vii. Let received be GeneratorYield(innerResult).
+      ...
+
+    GetIterator ( obj [ , hint ] )
+
+    ...
+    3. If hint is async,
+      a. Set method to ? GetMethod(obj, @@asyncIterator).
+      b. If method is undefined,
+        i. Let syncMethod be ? GetMethod(obj, @@iterator).
+        ii. Let syncIterator be ? Call(syncMethod, obj).
+        iii. Return ? CreateAsyncFromSyncIterator(syncIterator).
+    ...
+
+    GeneratorYield ( iterNextObj )
+
+    ...
+    10. If generatorKind is async,
+      a. Let value be IteratorValue(iterNextObj).
+      b. Let done be IteratorComplete(iterNextObj).
+      c. Return ! AsyncGeneratorResolve(generator, value, done).
+    ...
+
+ flags: [async]
+---*/
+
+var log = [];
+var iter = {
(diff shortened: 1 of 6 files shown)


test/language/statements/async-generator/yield-star-async-next.js
```diff
@@ -0,0 +1,196 @@
...
+
+    assert.sameValue(log[16].name, "after yield*");
+    assert.sameValue(log[16].value, "next-value-2");
+
+    assert.sameValue(v.value, "return-value");
+    assert.sameValue(v.done, true);
+
+    assert.sameValue(log.length, 17, "log.length");
+  }).then($DONE, $DONE);
+}).catch($DONE);
```
domenic: Maybe any rejections should cause the test to fail?

test/language/statements/async-generator/yield-star-sync-next.js
```diff
@@ -0,0 +1,171 @@
...
+              }
+            };
+          };
+        }
+      };
+    };
+  },
+  get [Symbol.asyncIterator]() {
+    log.push({ name: "get [Symbol.asyncIterator]" });
+    return null;
```
domenic: I'm probably missing something, but shouldn't this cause an error? Only undefined triggers the fallback to sync iterator, I thought?

caitp: The spec language agrees with you, this should be a TypeError trying to invoke `null`. V8 gets this wrong too, though. Maybe it would make sense to use `null` as a sentinel to use Async-from-Sync Iterator too?

domenic: Oh, we were reading the spec wrong. GetMethod censors `null` to `undefined`, so V8 and the test are right.

caitp: Ah, I see. Whew!

test/language/statements/async-generator/yield-star-async-next.js
arai-a: What does this comment mean?

domenic: Meaning, right now, if the promise rejects, the test will just be marked as done. Maybe the test should fail. E.g. maybe `.catch(e => assert(false, "Must not reject"))` is more appropriate?

caitp: `$DONE` should cause a test failure if the argument passed to it is truthy. (https://github.com/tc39/test262/issues/848)

domenic: Ah OK, and I suppose it's probably safe to assume this won't somehow reject with undefined; seems good.

test/language/statements/async-generator/yield-star-async-next.js
```diff
@@ -0,0 +1,196 @@
...
+              name: "call next",
+              args: args
+            });
+
+            nextCount++;
+            if (nextCount == 1) {
+              return {
+                get then() {
+                  log.push({ name: "get next then (1)" });
+                  return function(...args) {
```
caitp: For the sake of readability, how about `return function(resolve) { .... }`

and assign args with `args: [...arguments]`,

and call resolve as `resolve` rather than as `args[0]`?

test/language/statements/async-generator/yield-star-async-next.js
```diff
@@ -0,0 +1,196 @@
...
+      });
+      var nextCount = 0;
+      return {
+        get next() {
+          log.push({ name: "get next" });
+          return function(...args) {
+            log.push({
+              name: "call next",
+              args: args
+            });
```
arai-a: Do you think it's better checking thisValue for `next`, `return`, `throw`, and `then` ?
currently those functions' thisValue (object) are not global variable and cannot compare directly.
and putting them on global will make the test hard to follow, since the state (`nextCount` etc) needs to also be global.

would it be fine to define `name` property or something on those objects and compare `thisValue.name` against them?

arai-a: maybe, I just need to put expected this value to `log`.

domenic: I think the name solution is pretty good personally.

arai-a: added thisValue and thisValue.name check for each getter/method

test/language/statements/async-generator/yield-star-async-next.js
arai-a: fixed.

caitp:
Lgtm, thanks for these. Found a few bugs in the v8 impl!

PR #846, opened 2017-02-02
Add tests for try/finally in async functions

caitp:
Add tests to verify that try/finally defers execution of promise resolution for return statements and throw statements. This guards against a pattern which implementations may be tempted to use, such as rewriting return/throw statements at parse time like so:

From:
```js
return return_value;
```

To:
```js
return ResolvePromise(async_function_promise_, return_value), async_function_promise_;
```

v8 currently fails these tests, though a fix is in the pipeline.

### DIFF
--- a/test/language/expressions/async-arrow-function/try-return-finally-return.js
+++ b/test/language/expressions/async-arrow-function/try-return-finally-return.js
@@ -21,5 +21,5 @@ var f = async() => {
 };
 
 f().then(function(value) {
-    assert.sameValue(value, "override", "Return in finally block");
+  assert.sameValue(value, "override", "Return in finally block");
 }).then($DONE, $DONE);
--- a/test/language/expressions/async-arrow-function/try-return-finally-return.js
+++ b/test/language/expressions/async-arrow-function/try-return-finally-return.js
@@ -1,0 +1,25 @@
+// Copyright 2017 Caitlin Potter. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+author: Caitlin Potter <caitp@igalia.com>
+esid: pending
+description: >
+  Implementations must defer resolving an async function's Promise until after
+  all finally blocks have been evaluated.
+flags: [async]
+---*/
+
+var f = async() => {
+  try {
+    return "early-return";
+  } finally {
+    return await new Promise(function(resolve, reject) {
+      resolve("override");
+    });
+  }
+};
+
+f().then(function(value) {
+    assert.sameValue(value, "override", "Return in finally block");
+}).then($DONE, $DONE);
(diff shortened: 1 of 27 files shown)


test/language/expressions/async-function/try-throw-finally-return.js
```diff
@@ -0,0 +1,25 @@
...
+    throw "early-throw";
+  } finally {
+    return await new Promise(function(resolve, reject) {
+      resolve("override");
+    });
+  }
+};
+
+f().then($DONE, function(value) {
+  assert.sameValue("override", value);
```
jugglinmike: As above, this callback should be specified as the first argument to `then`, not the second.

test/language/statements/async-function/try-throw-finally-return.js
```diff
@@ -0,0 +1,25 @@
...
+    throw "early-throw";
+  } finally {
+    return await new Promise(function(resolve, reject) {
+      resolve("override");
+    });
+  }
+}
+
+f().then($DONE, function(value) {
+  assert.sameValue("override", value);
```
jugglinmike: As above, this callback should be specified as the first argument to `then`, not the second.

caitp: All of the *-return.js now do the assertion in the resolve callback

test/language/expressions/async-arrow-function/try-return-finally-return.js
```diff
@@ -0,0 +1,25 @@
...
+    return "early-return";
+  } finally {
+    return await new Promise(function(resolve, reject) {
+      resolve("override");
+    });
+  }
+};
+
+f().then(function(value) {
+    assert.sameValue("override", value, "Return in finally block");
```
caitp: I've switched them

caitp:
Ping, please take another look

test/language/statements/async-function/try-return-finally-reject.js
```diff
@@ -0,0 +1,25 @@
...
+    return "early-return";
+  } finally {
+    await new Promise(function(resolve, reject) {
+      reject("override");
+    });
+  }
+}
+
+f().then($DONE, function(value) {
+  assert.sameValue(value, "override", "Aaited rejection in finally block");
```
jugglinmike: Typo: "Aaited" -> "Awaiting"

jugglinmike:
I spotted a typo and a few indentation inconsistencies. Other than that, I
think this looks great!

Since these are pretty objective changes, and since I know you've been waiting
for a while, I pushed up a commit to resolve them myself. Hope that wasn't
presumptuous of me. If it looks good to you, let me know and I'll squash and
merge.

caitp:
The fixups are fine with me. On the v8 version of these tests, Dan had s digested tests to ensure that

```js
async function f() {
  try {
    return new Promise(function() {});
  } finally {
    return "foo";
  }
}
```

would not Await the `new Promise(...)`.

We could add those tests here before landing. I won't have time to add them until Monday, though. If you're up for it, or don't think it adds much, feel free to merge at will with or without those changes.

jugglinmike:
I'd be happy to review those extra tests whenever you can find the time. I suspect you've got plenty on your plate, though. These tests are great on their own, so I've merged them via 098f9ca3de5043c376e0cb68c14acf535ad8d66e. Thanks, Cait!

PR #849, opened 2017-02-06
Add tests for %TypedArray%.prototype.copyWithin() with target/start/end=Infinity

caitp:
ToInteger can result in the values +Infinity and -Infinity.


littledan:
LGTM.

For bonus points, does anyone want to backport these tests to Array.prototype.copyWithin? Could be in a different patch.

tcare:
Looks good

PR #853, opened 2017-02-09
Update RegExp tests for the semantics change

littledan:
Spec change at https://github.com/tc39/ecma262/pull/798

Tested on V8's implementation of the new semantics.

cc @schuay @bterlson @anba @claudepache

### DIFF
--- a/test/built-ins/RegExp/prototype/exec/failure-lastindex-access.js
+++ b/test/built-ins/RegExp/prototype/exec/failure-lastindex-access.js
@@ -2,13 +2,14 @@
 // This code is governed by the BSD license found in the LICENSE file.
 
 /*---
-description: lastIndex is not accessed when global and sticky are unset.
+description: lastIndex is read but not written when global and sticky are unset.
 es6id: 21.2.5.2.2
 info: >
     21.2.5.2.2 Runtime Semantics: RegExpBuiltinExec ( R, S )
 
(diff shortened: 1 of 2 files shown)


schuay:
Thanks for the quick turnaround, looks good to me.

littledan:
I couldn't stand to see you checking in failed test262 status file lines, @schuay !

test/built-ins/RegExp/prototype/exec/failure-lastindex-access.js
```diff
@@ -18,16 +19,18 @@ info: >
...
 
 var r = /a/;
-r.lastIndex = thrower;
+r.lastIndex = counter;
 
 var result = r.exec('nbc');
 assert.sameValue(result, null);
-assert.sameValue(r.lastIndex, thrower);
-
+assert.sameValue(r.lastIndex, counter);
```
leobalter: nice approach, I would try some get/set accessor operation in there but this works and it's simple.

leobalter:
LGTM

test/built-ins/RegExp/prototype/exec/failure-lastindex-access.js
littledan: That was my first idea, but I had forgotten that `lastIndex` is not configurable, so you can't make a getter or setter.

PR #857, opened 2017-02-11
fix for the wrong test of Intl.NumberFormat (ToRawPrecision)

Yaffle:
1) (123.44500) == 123.444999999999993179
2) (123.44500).toPrecision(5) === "123.44" gives correct value in Chrome and Firefox;


littledan:
Here, as with #856, I have some doubts about whether this is the behavior we should have, but either way, I'd still prefer if we kept the original test and changed the expectations (possibly adding a new test which could make it more clear or bring out some additional cases).

littledan:
Thanks for this update. As in https://github.com/tc39/test262/pull/856#issuecomment-280954832 , could you put a cross-reference to the 402 bug?

leobalter:
@littledan I'll merge this and follow up with the references you requested.

PR #860, opened 2017-02-14
Test for parameters with default expressions in Annex B FiB tests

syg:
@anba found this bug in Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1339123

This is worth testing because engines may implement the extra scope needed for function parameter expressions only if there are expressions.

### DIFF
--- a/test/annexB/language/function-code/block-decl-func-skip-param.js
+++ b/test/annexB/language/function-code/block-decl-func-skip-param.js
@@ -27,5 +27,18 @@ var init, after;
   after = f;
 }(123));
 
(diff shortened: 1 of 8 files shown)


jugglinmike:
This patch modifies files that are procedurally generated. Please update the relevant source files (as documented via in-line comments in each file) instead. You can find more information about working with the test generation tool in [the project's `CONTRIBUTING.md` file](https://github.com/tc39/test262/blob/master/CONTRIBUTING.md).

On a related note, if there is anything else we can do to make this distinction more clear, please let us know!

leobalter:
I'm trying to fix this on #891

PR #864, opened 2017-02-15
NumberFormat default minimumSignificantDigits

vabr-g:
Adds a test for NumberFormat default minimumSignificantDigits.

This originates from the V8 bugfix https://bugs.chromium.org/p/v8/issues/detail?id=5554.

### DIFF
--- a/test/intl402/NumberFormat/default-minimum-singificant-digits.js
+++ b/test/intl402/NumberFormat/default-minimum-singificant-digits.js
@@ -13,6 +13,9 @@ assert.throws(RangeError, function() {
   Intl.NumberFormat(undefined, {maximumSignificantDigits: 0});
 });
 
-// For similar reasons, the following statement is checking that
-// minimumSignificantDigits is at most 1.
-assert.notSameValue(Intl.NumberFormat(undefined, {maximumSignificantDigits: 1}), undefined);
+// If nothing is thrown, check that the options are resolved appropriately.
+var res = Intl.NumberFormat(undefined, {maximumSignificantDigits: 1})
+
+assert.sameValue(Object.getPrototypeOf(res), Intl.NumberFormat.prototype, 'result is an instance of NumberFormat')
+assert.sameValue(res.resolvedOptions().minimumSignificantDigits, 1, 'default minimumSignificantDigits')
+assert.sameValue(res.resolvedOptions().maximumSignificantDigits, 1, 'sets maximumSignificantDigits')
--- a/test/intl402/NumberFormat/default-minimum-singificant-digits.js
+++ b/test/intl402/NumberFormat/default-minimum-singificant-digits.js
@@ -1,0 +1,17 @@
+// Copyright (C) 2017 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Tests that the default value of minimumSignificantDigits is 1.
+---*/
+
+// maximumSignificantDigits needs to be in range from minimumSignificantDigits
+// to 21 (both inclusive). Setting maximumSignificantDigits to 0 will throw a
+// RangeError if the (default) minimumSignificantDigits is at least 1.
+assert.throws(RangeError, function() {
+  Intl.NumberFormat(undefined, {maximumSignificantDigits: 0});
+});
+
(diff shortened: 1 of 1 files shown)


littledan:
LGTM

vabr-g:
Thanks for the review!
I don't seem to be allowed to merge PRs in this repository, could you please merge it as well?
Thanks,
Vaclav

littledan:
I'm not allowed to merge them either! We'll have to wait for a real repository owner like @leobalter or @tcare

test/intl402/NumberFormat/default-minimum-singificant-digits.js
```diff
@@ -0,0 +1,17 @@
+// Copyright (C) 2017 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Tests that the default value of minimumSignificantDigits is 1.
+---*/
```
leobalter: it would be great to add the esid tag here so we have a proper reference to the spec.

test/intl402/NumberFormat/default-minimum-singificant-digits.js
```diff
@@ -0,0 +1,17 @@
...
+// maximumSignificantDigits needs to be in range from minimumSignificantDigits
+// to 21 (both inclusive). Setting maximumSignificantDigits to 0 will throw a
+// RangeError if the (default) minimumSignificantDigits is at least 1.
+assert.throws(RangeError, function() {
+  Intl.NumberFormat(undefined, {maximumSignificantDigits: 0});
+});
+
+// For similar reasons, the following statement is checking that
+// minimumSignificantDigits is at most 1.
+Intl.NumberFormat(undefined, {maximumSignificantDigits: 1});
```
leobalter: it would be good to assert the result of this. Explicit assertions are better than checking for API's that are not throwing anything.

test/intl402/NumberFormat/default-minimum-singificant-digits.js
vabr-g: Good idea, done.

test/intl402/NumberFormat/default-minimum-singificant-digits.js
vabr-g: Sure. I am just asserting that the result is not undefined, please let me know if you want more.

vabr-g:
Thanks @leobalter for the review (and @littledan for looping in the owners).
I tried to address both comments, please have a look again.

test/intl402/NumberFormat/default-minimum-singificant-digits.js
```diff
@@ -0,0 +1,18 @@
...
+// maximumSignificantDigits needs to be in range from minimumSignificantDigits
+// to 21 (both inclusive). Setting maximumSignificantDigits to 0 will throw a
+// RangeError if the (default) minimumSignificantDigits is at least 1.
+assert.throws(RangeError, function() {
+  Intl.NumberFormat(undefined, {maximumSignificantDigits: 0});
+});
+
+// For similar reasons, the following statement is checking that
+// minimumSignificantDigits is at most 1.
+assert.notSameValue(Intl.NumberFormat(undefined, {maximumSignificantDigits: 1}), undefined);
```
leobalter: I've had something else in mind. I'm used to test something I can have from the actual result, rather than just checking it's something (not undefined), otherwise, `42` would be a false positive.

What do you think about checking the result object with the expected set values? The assertions in the example below confirms the default and given values are properly set, right?

```js
var res = Intl.NumberFormat(undefined, {maximumSignificantDigits: 1})

assert.sameValue(Object.getPrototypeOf(res), Intl.NumberFormat.prototype, 'result is an instance of NumberFormat')
assert.sameValue(res.resolvedOptions().minimumSignificantDigits, 1, 'default minimumSignificantDigits')
assert.sameValue(res.resolvedOptions().maximumSignificantDigits, 1, 'sets maximumSignificantDigits')
```

vabr-g: Ah, thanks, this is much better! I put your proposal in the patch.

vabr-g:
Thanks again, @leobalter . The patch is now updated after your most recent comment and ready for re-review.

PR #867, opened 2017-02-20
Implementing initial set test cases for object rest deconstruction

caiolima:
This PR is an initial set for Object Properties Rest Deconstruction proposal. To run these tests is necessary generate test cases using make.py script. To run them, an agent harness is required. v8 is already implementing it behind ```--harmony_object_rest_spreas``` and a proposed Patch in JavaScriptCore is ongoing [1]. I ran these tests using JavaScriptCore's Patch applied.

Some tests around the syntax is also coming.

[1] - https://bugs.webkit.org/show_bug.cgi?id=167962

### DIFF
--- a/src/dstr-assignment/obj-rest-to-property-with-setter.case
+++ b/src/dstr-assignment/obj-rest-to-property-with-setter.case
@@ -1,0 +1,30 @@
+// Copyright (C) 2017 Caio Lima. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: >
+    When DestructuringAssignmentTarget is an object property setter, its value should be
+    binded as rest object.
+template: default
+esid: pending
+---*/
+
+//- setup
+var settedValue;
+var executedGetter = false;
+var src = {
+    get y() { executedGetter = true; },
+    set y(v) {
+        settedValue = v;
+    }
+}
+src.y = undefined;
+//- elems
+{...src.y}
+//- vals
+{ x: 1, y: 2}
+//- body
+assert.sameValue(settedValue.x, 1);
+assert.sameValue(settedValue.y, 2);
+assert(!executedGetter, "The property should not be accessed");
+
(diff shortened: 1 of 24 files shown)


leobalter:
@jugglinmike thank you, Mike!

@caiolima ping me anytime if you need any help or support from me.

caiolima:
@jugglinmike thanks for the review! I'm so sorry for late reply, but I was without internet connection this week. I'm working in your comments right now.

caiolima:
@jugglinmike and @leobalter Updated PR based in your comments. Also added cases suggested by Mike.

Ready for a second round.

src/dstr-assignment/obj-rest-to-property-with-setter.case
```diff
@@ -0,0 +1,30 @@
...
+    binded as rest object.
+template: default
+esid: pending
+---*/
+
+//- setup
+var settedValue;
+var executedGetter = false;
+var src = {
+    get y() { executedGetter = true; },
```
jugglinmike: Tough to see it on GitHub.com, but this line has trailing whitespace.

jugglinmike:
Don't worry about the delay, @caiolima. I wasn't planning on returning to this until the weekend, anyway.

Just one minor request, then we should be good to go!

caiolima:
@jugglinmike nice catch. I didn't see it in my diff locally. Now I think we are ready to go.

Thx for your time reviewing it!

leobalter:
🎉 🎉 🎉 🎉 🎉 

I'll add the generated the tests and merge it today. Thank you, @caiolima and @jugglinmike!

leobalter:
through a chat convo, we found a minor bug on cases assigning a var `obj` where the template already has an assignment for `obj`. I'm waiting for the fix before landing. 

+1 @caiolima

caiolima:
@leobalter updated

leobalter:
This is merged at 0200c63 with the generated tests! Way to go, @caiolima!

caiolima:
Thank you very much @leobalter and @jugglinmike!

PR #869, opened 2017-02-23
Test that @@unscopables is looked up once for inc/dec.

syg:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1341061

### DIFF
--- a/test/language/statements/with/unscopables-inc-dec.js
+++ b/test/language/statements/with/unscopables-inc-dec.js
@@ -1,0 +1,51 @@
+// Copyright (C) 2017 Mozilla Corporation. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-object-environment-records-hasbinding-n
+description: >
+  @@unscopables should be looked up exactly once for inc/dec.
+info: |
+  UpdateExpression : LeftHandSideExpression ++
+  1. Let lhs be the result of evaluating LeftHandSideExpression.
+
+  GetIdentifierReference ( lex, name, strict )
+  [...]
+  3. Let exists be ? envRec.HasBinding(name).
+
+  HasBinding ( N )
+  [...]
+  6. Let unscopables be ? Get(bindings, @@unscopables).
+flags: [noStrict]
+features: [Symbol.unscopables]
+---*/
+
+var unscopablesGetterCalled = 0;
+var a, b, flag = true;
+with (a = { x: 7 }) {
+  with (b = { x: 4, get [Symbol.unscopables]() {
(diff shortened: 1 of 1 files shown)


littledan:
I hate the world.

You could make this test marginally stronger by checking that flag is false after both test cases (which would verify that `@@unscopables` isn't read a second time, even if your test already already verifies that only the first read is used). You could also have a test case where it actually is reflected in `@@unscopables`, to verify that implementations do pay attention to unscopables getters (not sure if we have a separate test for that).

Either way, those are potential follow-ons, but this looks good to land already. LGTM.

syg:
@littledan What's meant by "it actually is reflected in` @@unscopables`"?

littledan:
@syg sorry, I meant, "where the `@@unscopables` value for the key was `true`, causing it to be excluded"

littledan:
New version LGTM, that other test possibility is sort of a tangent.

test/language/statements/with/unscopables-inc-dec.js
```diff
@@ -0,0 +1,49 @@
+// Copyright (C) 2017 Mozilla Corporation. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description:
```
leobalter: can you add a `>` at the end of this line to make it compatible with monkeyYaml?

`description: >`

Ref https://github.com/tc39/test262/pull/862

test/language/statements/with/unscopables-inc-dec.js
```diff
@@ -0,0 +1,49 @@
...
+  GetIdentifierReference ( lex, name, strict )
+  [...]
+  3. Let exists be ? envRec.HasBinding(name).
+
+  HasBinding ( N )
+  [...]
+  6. Let unscopables be ? Get(bindings, @@unscopables).
+flags: [noStrict]
+features: [Symbol.unscopables]
+---*/
```
leobalter: this is missing an `esid` tag

test/language/statements/with/unscopables-inc-dec.js
```diff
@@ -0,0 +1,49 @@
...
+                    } }) {
+    x++;
+  }
+}
+
+assert.sameValue(unscopablesGetterCalled, 1);
+assert.sameValue(a.x, 7);
+assert.sameValue(b.x, 5);
+
+flag = true;
```
leobalter: I like to set tests atomic, this means you could just reset the value of `unscopablesGetterCalled` to `0` here as well.

syg:
Comments addressed.

leobalter:
Thanks!

PR #870, opened 2017-02-23
Account for the use of ToIndex in Atomic operations

lars-t-hansen:
These test changes pertain to the spec change in https://github.com/tc39/ecma262/issues/807.

### DIFF
--- a/harness/testAtomics.js
+++ b/harness/testAtomics.js
@@ -13,13 +13,8 @@ function testWithAtomicsOutOfBoundsIndices(f) {
         (view) => -1,
         (view) => view.length,
         (view) => view.length*2,
-        (view) => undefined,
-        (view) => Number.NaN,
         (view) => Number.POSITIVE_INFINITY,
         (view) => Number.NEGATIVE_INFINITY,
-        (view) => '3.5',
-        (view) => 3.5,
-        (view) => { password: "qumquat" },
         (view) => ({ valueOf: () => 125 }),
         (view) => ({ toString: () => '125', valueOf: false }) // non-callable valueOf triggers invocation of toString
     ];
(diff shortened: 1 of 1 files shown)


harness/testAtomics.js
```diff
@@ -44,6 +39,11 @@ function testWithAtomicsInBoundsIndices(f) {
     var good_indices = [
         (view) => 0/-1, // -0
         (view) => '-0',
+        (view) => undefined, // 0
+        (view) => Number.NaN, // 0
+        (view) => 3.5,
+        (view) => '3.5',
```
leobalter: if the length is 3 or more this will fail.

`0.5` and `'0.5'` should avoid this

harness/testAtomics.js
```diff
@@ -44,6 +39,11 @@ function testWithAtomicsInBoundsIndices(f) {
...
+        (view) => ({ password: "qumquat" }),
         (view) => view.length - 1,
         (view) => ({ valueOf: () => 0 }),
         (view) => ({ toString: () => '0', valueOf: false }) // non-callable valueOf triggers invocation of toString
```
leobalter: you can include `false` to improve coverage for the ToIndex operation here.

`-0.9` should work as a good index as well, weird, but it works.

harness/testAtomics.js
```diff
@@ -44,6 +39,11 @@ function testWithAtomicsInBoundsIndices(f) {
     var good_indices = [
         (view) => 0/-1, // -0
         (view) => '-0',
+        (view) => undefined, // 0
+        (view) => Number.NaN, // 0
```
leobalter: the regular global `NaN` should do the trick as well, but it's not necessary to do change this.

harness/testAtomics.js
lars-t-hansen: Do you mean, if the length is 3 or less this will fail?  I'll change this to 0.5 which has the same effect for me, but may still fail if the length is zero.  I'll add a comment stipulating the minimum expected length.

harness/testAtomics.js
```diff
@@ -33,16 +33,20 @@ function testWithAtomicsOutOfBoundsIndices(f) {
  * Calls the provided function for each good index that should not throw when
  * passed to an Atomics method on a SAB-backed view.
  *
+ * The view must have length greater than zero.
```
leobalter: +1

PR #871, opened 2017-02-26
Async Generator flow control tests

caitp:
Here's a starter suite of tests for basic flow control in Async Generators.

This relates primarily to the behaviour of AsyncGeneratorResumeNext, AsyncGeneratorStart, AsyncGeneratorResolve and AsyncGeneratorReject, as well as some coverage of changes to GeneratorYield.

There is stuff missing, as I haven't exhaustively written tests for all of the variations that are necessary.

### DIFF
--- a/test/language/expressions/async-generators/return-suspendedStart-promise.js
+++ b/test/language/expressions/async-generators/return-suspendedStart-promise.js
@@ -1,0 +1,36 @@
+// Copyright 2017 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+author: Caitlin Potter <caitp@igalia.com>
+esid: pending
+description: >
+  AsyncGeneratorResumeNext:
+  If completion.[[Type]] is return, and generator.[[AsyncGeneratorState]] is
+  "suspendedStart", generator is closed without being resumed.
+
+  AsyncGeneratorResolve will unwrap Promise values (steps 6-10)
+flags: [async]
+---*/
+
+var g = async function*() {
+  throw new Test262Error('Generator must not be resumed.');
+};
+
+var it = g();
+var resolve;
+var promise = new Promise(function(resolver) {
+  resolve = resolver;
+});
+
+it.return(promise).then(function(ret) {
+  assert.sameValue(ret.value, 'unwrapped-value', 'AsyncGeneratorResolve(generator, completion.[[Value]], true)');
+  assert.sameValue(ret.done, true, 'AsyncGeneratorResolve(generator, completion.[[Value]], true)');
+
+  it.next().then(function(ret) {
+    assert.sameValue(ret.value, undefined, 'Generator is closed');
+    assert.sameValue(ret.done, true, 'Generator is closed');
+  }).then($DONE, $DONE);
+}).catch($DONE);
+
+resolve('unwrapped-value');
--- a/test/language/expressions/async-generators/return-suspendedYield-try-finally-throw.js
+++ b/test/language/expressions/async-generators/return-suspendedYield-try-finally-throw.js
@@ -1,0 +1,41 @@
+// Copyright 2017 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+author: Caitlin Potter <caitp@igalia.com>
+esid: pending
+description: >
+  AsyncGeneratorResumeNext:
+  If completion.[[Type]] is return, and generator.[[AsyncGeneratorState]] is
+  "suspendedYield", and generator is resumed within a try-block with an
+  associated finally block, resume execution within finally.
+flags: [async]
+---*/
+
+var error = new Error("boop");
+var g = async function*() {
+  try {
+    yield 1;
+    throw new Test262Error('Generator must be resumed in finally block.');
+  } finally {
+    throw error;
+    throw new Test262Error('Generator must not be resumed.');
+  }
+};
+
+var it = g();
+it.next().then(function(ret) {
+  assert.sameValue(ret.value, 1, 'Initial yield');
+  assert.sameValue(ret.done, false, 'Initial yield');
+
+  it.return('sent-value').then($DONE, function(err) {
+    assert.sameValue(err, error, 'AsyncGeneratorReject(generator, resultValue)');
+
+    it.next().then(function(ret) {
+      assert.sameValue(ret.value, undefined, 'Generator is closed');
+      assert.sameValue(ret.done, true, 'Generator is closed');
+    }).then($DONE, $DONE);
+    
+  }).catch($DONE);
+
+}).catch($DONE);
--- a/test/language/expressions/async-generators/return-suspendedYield-try-finally.js
+++ b/test/language/expressions/async-generators/return-suspendedYield-try-finally.js
@@ -1,0 +1,46 @@
+// Copyright 2017 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+author: Caitlin Potter <caitp@igalia.com>
+esid: pending
+description: >
+  AsyncGeneratorResumeNext:
+  If completion.[[Type]] is return, and generator.[[AsyncGeneratorState]] is
+  "suspendedYield", and generator is resumed within a try-block with an
+  associated finally block, resume execution within finally.
+flags: [async]
+---*/
+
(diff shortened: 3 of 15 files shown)


caitp:
Ping, it would be great to get a look at these and start filling in holes in coverage here

leobalter:
Hi, @caitp! I'm back to the regular work time on test262, although, there are other older PR I've had to look before this one.

I'll review this one first thing tomorrow.

Can you confirm if you're planning to add more tests in this PR or this is the current work that we should land and expect/start a follow PR for coverage completion?

caitp:
well, if I come up with more tests, they'll be either in this PR or a followup. I don't think there's any problem with incrementally landing more coverage.

leobalter:
It's not a problem, for sure. I just want to make the best time strategy for reviews and I really appreciate the work.

test/language/expressions/async-generators/return-suspendedYield-try-finally-throw.js
```diff
@@ -0,0 +1,44 @@
...
+
+var it = g();
+it.next().then(function(ret) {
+  // Initial yield
+  assert.sameValue(ret.value, 1);
+  assert.sameValue(ret.done, false);
+
+  it.return('sent-value').then($DONE, function(err) {
+    // Throw in finally block
+    assert(err instanceof Err);
```
leobalter: if you store the reference for `new Error` you can throw it in the finally block and assert `err` is the same reference, right?

It also saves the need for an Error subclass

test/language/expressions/async-generators/return-suspendedYield-try-finally.js
```diff
@@ -0,0 +1,50 @@
...
+      it.next().then(function(ret) {
+        assert.sameValue(ret.value, undefined);
+        assert.sameValue(ret.done, true);
+      }).then($DONE, $DONE);
+
+    }).catch($DONE);
+    
+  }).catch($DONE);
+
+}).catch($DONE);
```
leobalter: I wish we could have `Promise#finally` already

test/language/expressions/async-generators/throw-suspendedYield-try-finally.js
```diff
@@ -0,0 +1,50 @@
...
+  }
+};
+
+var it = g();
+it.next().then(function(ret) {
+  // Initial yield
+  assert.sameValue(ret.value, 1);
+  assert.sameValue(ret.done, false);
+
+  it.throw(new Err).then(function(ret) {
```
leobalter: this is the same case I mentioned above for storing an instance of `new Error`.

test/language/expressions/async-generators/throw-suspendedYield.js
```diff
@@ -0,0 +1,39 @@
...
+var g = async function*() {
+  yield 1;
+  $ERROR('Generator must not be resumed.');
+};
+
+var it = g();
+it.next().then(function(ret) {
+  // Initial yield
+  assert.sameValue(ret.value, 1);
+  assert.sameValue(ret.done, false);
```
leobalter: the comments are great to track the steps. If you don't mind, can you move them to become assertion messages?

This way we can have a powerful track of each assertion and, better, tells the implementors where each test is failing at.

Example for the case above:

```js
assert.sameValue(ret.value, 1, 'Initial yield');
assert.sameValue(ret.done, false, 'Initial yield');
```

Even if they are the same message, I can already identify each one due to the respective expected values.

leobalter:
To avoid noises, I left some suggestions that can be applied to multiple places, but in a single comment as an example.

Some cases are:

1. Moving inline comments to assertion messages.
2. throwing the error reference to check for the sameValue instead of err instanceof Err, avoiding the Error subclass
3. splitting the current frontmatter description content into description and info

etc.

Thank you for the tests, @caitp! I'm looking forward the next review round!

test/language/expressions/async-generators/throw-suspendedYield.js
caitp: I've done this --- but it gets complicated for the generator-closing operations. I've gone with using the spec step which leads to the promise capability resolve/reject call instead there, but that seems awkward too.

test/language/expressions/async-generators/expression-yield-promise.js
```diff
@@ -0,0 +1,48 @@
+// Copyright 2017 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+author: Caitlin Potter <caitp@igalia.com>
+esid: pending
+description: >
```
caitp: I'm open to suggestions on better description text, but otherwise I'm leaving these as-is, and the web runner can just deal with it

test/language/expressions/async-generators/return-suspendedStart-promise.js
```diff
@@ -0,0 +1,38 @@
...
+  AsyncGeneratorResumeNext:
+  If completion.[[Type]] is return, and generator.[[AsyncGeneratorState]] is
+  "suspendedStart", generator is closed without being resumed.
+
+  AsyncGeneratorResolve will unwrap Promise values (steps 6-10)
+flags: [async]
+---*/
+
+var g = async function*() {
+  $ERROR('Generator must not be resumed.');
```
caitp: Done

tcare:
Could you please keep spec text in an "info:" section and use "description:" as a short description of why the individual test is interesting?

leobalter:
Thanks for the updates, @caitp. 

As @tcare mentioned, the short description is something we need to keep the conventional conformance and a good point for a short reasoning on each test.

caitp:
The description lines are influenced by existing tests in the repository, which include spec citations. I would appreciate specific suggestions for individual tests.

test/language/expressions/async-generators/return-suspendedYield-try-finally.js
gnarf: Is there a specific reason these are nested instead of chained anyway?

```js
it.next()
.then(function(ret) {
  assert.sameValue(ret.value, 1, 'Initial yield');
  assert.sameValue(ret.done, false, 'Initial yield');

  return it.return('sent-value');
})
.then(function(ret) {
  assert.sameValue(ret.value, 2, 'Yield in finally block');
  assert.sameValue(ret.done, false, 'Yield in finally block');

  return it.next();
})
.then(function(ret) {
  assert.sameValue(ret.value, 'sent-value', 'AsyncGeneratorResolve(generator, resultValue, true)');
  assert.sameValue(ret.done, true, 'AsyncGeneratorResolve(generator, resultValue, true)');

  return it.next();
})
.then(function(ret) {
  assert.sameValue(ret.value, undefined, 'Generator is closed');
  assert.sameValue(ret.done, true, 'Generator is closed');
})
.then($DONE, $DONE);

caitp: harder to read, less obvious, etc

tcare:
Updating all the tests is preventatively expensive. Another example is the es6id there, which is now superseded by esid. We do drive-by cleanup and all new tests coming in should conform to the current standard.

caitp:
I'm happy to change the frontmatter to anything requested (within reason), but be please be specific about those requests.

tcare:
Sorry for the confusion.

A random example: https://github.com/tc39/test262/blob/82c2ca070976329e822a466e2d6bafa88b87bc4d/test/intl402/PluralRules/prototype/select/tainting.js

Generally, "info:" can be used to dump spec text, and "description:" is for the brief description of what the test is trying to achieve or why it is interesting. We try to avoid having spec text in "description:" so that it can be used as a quick reference.

The descriptions from your previous commit https://github.com/tc39/test262/commit/098f9ca3de5043c376e0cb68c14acf535ad8d66e are fine. Relevant spec text sections can be put into "info:".

test/language/expressions/async-generators/return-suspendedYield-try-finally.js
gnarf: I very much disagree, 4 level nested callbacks are much less readable than chaining promises, but I also don't care enough to block, it seems you've reached "annoyed" point on this PR and I don't want to add to it any more.   Thanks for the contribution!

leobalter:
@caitp You've done a lot of great work here, and I really appreciate your contribution. 

I will follow up with a new PR to address the remaining nit picky formatting things, mostly providing the frontmatter descriptions. Hopefully you may help me reviewing the new changes.

PR #872, opened 2017-02-28
Top-level `let undefined` is a runtime error, not an early error.

bakkot:
The error occurs in [GlobalDeclarationInstantiation](https://tc39.github.io/ecma262/#sec-globaldeclarationinstantiation), which is runtime semantics.

INTERPRETING.md is kind of unclear on what phase GlobalDeclarationInstantiation corresponds to, so I'm not entirely sure if this change is correct.

Ping @jugglinmike

### DIFF
--- a/test/language/global-code/decl-lex-restricted-global.js
+++ b/test/language/global-code/decl-lex-restricted-global.js
@@ -11,7 +11,7 @@ info: |
      c. Let hasRestrictedGlobal be ? envRec.HasRestrictedGlobalProperty(name).
      d. If hasRestrictedGlobal is true, throw a SyntaxError exception.
 negative:
-  phase: early
+  phase: runtime
   type: SyntaxError
 ---*/
 


anba:
The following tests are also marked with error phase=early, while actually throwing an error at runtime.

- test/language/module-code/eval-rqstd-abrupt.js
- test/language/module-code/instn-iee-err-ambiguous-as.js
- test/language/module-code/instn-iee-err-ambiguous.js
- test/language/module-code/instn-iee-err-circular-as.js
- test/language/module-code/instn-iee-err-circular.js
- test/language/module-code/instn-iee-err-dflt-thru-star-as.js
- test/language/module-code/instn-iee-err-dflt-thru-star.js
- test/language/module-code/instn-iee-err-not-found-as.js
- test/language/module-code/instn-iee-err-not-found.js
- test/language/module-code/instn-named-err-ambiguous-as.js
- test/language/module-code/instn-named-err-ambiguous.js
- test/language/module-code/instn-named-err-dflt-thru-star-as.js
- test/language/module-code/instn-named-err-dflt-thru-star-dflt.js
- test/language/module-code/instn-named-err-not-found-as.js
- test/language/module-code/instn-named-err-not-found-dflt.js
- test/language/module-code/instn-named-err-not-found.js
- test/language/module-code/instn-star-err-not-found.js

jugglinmike:
Your change is definitely in-line with the distinction we are trying to
observe. For instance, the error demonstrated in this file should not be
reported by ECMAScript parsers.

The documentation you referenced was tricky to write, and I understand why you
might be uncertain. If you have any ideas on how we might improve it, I'm all
ears!

littledan:
Hmm, if these are moved to a later phase, it might be good to somehow test that the error comes earlier than when the lines of code are executed. IIRC There was logic for this that @jugglinmike removed in the phase marking patch.

bakkot:
@littledan, good point. For non-raw "phase: early" tests I've just been prepending `throw new Test262Error` and asserting that the actual error is not a Test262Error, which I believe is what the old logic was.

I'd support either having an "phase: initialization" or adding `throw new Test262Error` as the first line of tests which throw as runtime semantics but before executing any line of code. I'm ambivalent between the two. Thoughts?

allenwb:
what really disturbing about this, is that it turns out that we allow non-global let/const/class to bind the name "undefined". Even in strict mode!

I think this is an oversight in the ES6 spec. Don't know if this is fixable at this point. But I'll file a spec. bug.

leobalter:
I opened #877 and #878 to follow up with the discussions here in open issues. This PR is getting merged and I don't want to lose the feedbacks.

leobalter:
@bakkot while we still need to discuss a new phase, I believe a `throw new Test262Error` would be good to identify the proper expected errors in places we might find have multiple observable steps.

For this test file, there's nothing else to throw an error before or after the one we expect, so it's not necessary. The changes are fine for me.

PR #873, opened 2017-03-01
Add tests for abrupt case in yield* in async generator - getting iterator

arai-a:
Here's first chunk of abrupt case for yield* in async generator.
from getting yield* operand to getting iterator from it.

### DIFF
--- a/test/language/statements/async-generator/yield-star-expr-abrupt.js
+++ b/test/language/statements/async-generator/yield-star-expr-abrupt.js
@@ -1,0 +1,49 @@
+/*---
+ author: Tooru Fujisawa [:arai] <arai_a@mac.com>
+ esid: pending
+ description: Abrupt completion while getting yield* operand
+ info: >
+    YieldExpression: yield * AssignmentExpression
+
+    1. Let exprRef be the result of evaluating AssignmentExpression.
+    2. Let value be ? GetValue(exprRef).
+    ...
+
+ flags: [async]
+---*/
+
+var log = [];
+var rejectReason = new Proxy({}, {
+  get() {
+    // No operation should be done on reject reason.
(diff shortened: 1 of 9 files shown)


test/language/statements/async-generator/yield-star-expr-abrupt.js
```diff
@@ -0,0 +1,49 @@
...
+var asyncIterator = (async function*() {
+  log.push({ name: "before yield*" });
+  yield* abrupt();
+  log.push({ name: "after yield*" });
+  return "return-value";
+})();
+
+assert.sameValue(log.length, 0, "log.length");
+
+asyncIterator.next("next-arg-1").then(() => {
```
leobalter: this argument is not necessary, right?

leobalter:
and the most important: Thank you for the PR! I'm looking forward for the next updates!

leobalter:
hi @arai-a!

Do you have interest to continue on this PR? At this point, I can go ahead and apply the changes from the feedback and create another PR on top. What do you think?

If you want to address it, I can go ahead and write more tests to cover the other parts for async generators.

arai-a:
sorry for the delay.

I've started rewriting with template and it would take some more time to figure out how
 to apply the template,
and I'm afraid now I'm busy.
I'm about to continue later this month or next month, but feel free to take this over if it's urgent.

thanks!

leobalter:
Thanks for the update! I'm writing some async-generator tests using the templates that won't cross with your current tests, so we can easily mix then later.

I'll ping you as soon as I have a formed PR so you can check the new templates/cases used on the test generation tool.

leobalter:
I'm working on this through the test generation tool, keeping the authorship as well. I'll open a PR soon.

arai-a:
thanks :)

PR #874, opened 2017-03-01
Add references to fixed tests for NumberFormat operations

leobalter:
cc @littledan 

Ref #856
Ref #857
Ref tc39/ecma402#128


littledan:
LGTM, these references seem really useful to me. Maybe it'd be better if they were full URLs, since we don't have space limitations and the abbreviated ones may be unclear to people without as much context.

PR #879, opened 2017-03-02
Missing Math desc tests

leobalter:
Fixes #827 
Fixes #828
Fixes #829
Fixes #830
Fixes #831

PR #880, opened 2017-03-03
Improve isEnumerable check with a for-in loop for string names

leobalter:
Ref https://github.com/tc39/test262/pull/879#discussion_r104128520

cc @anba 

For the piece of mind, I'm comparing tests results using a new compiled version of d8.

```
test262-harness --hostPath=`which d8` --hostType=d8 $(git grep --name-only --extended-regexp 'verify(Not)?Enumerable' test/built-ins)
```

### DIFF
--- a/harness/propertyHelper.js
+++ b/harness/propertyHelper.js
@@ -11,7 +11,22 @@ function isConfigurable(obj, name) {
 }
 
 function isEnumerable(obj, name) {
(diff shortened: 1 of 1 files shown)


leobalter:
```
Ran 2950 tests
2846 passed
104 failed
```

same failures on both. This change won't break anything and adds the extra coverage checking enumerable string names with for-in.

PR #883, opened 2017-03-04
Add tests for abrupt case in yield* in async generator - next

arai-a:
next part for https://github.com/tc39/test262/pull/873
tests for steps 6.a in https://tc39.github.io/proposal-async-iteration/#prod-YieldExpression

I skipped step 6.a.vii because of https://github.com/tc39/proposal-async-iteration/issues/86


leobalter:
similar to #938, I'm adding these cases to the test generation tool!

PR #884, opened 2017-03-04
Extend test generation tool

jugglinmike:
The test generation tool exhibited a bug that made it unable to produce valid
tests in some situations. I discovered this while working on a patch recently
and decided to manually duplicate tests as a short-term workaround. This patch
corrects the bug in the tool and re-formats those manually-duplicated tests to
a procedurally-generated form.


leobalter:
nice one! Thanks!

Issue #865, opened 2017-02-15
Add tests for Object rest/spread properties


leobalter:
@sebmarkbage do you already have someone working on the tests for this proposal? If not, we might take this task.

sebmarkbage:
I've started converting from the other Array rest/spread cases so that we cover the same scenarios. Unfortunately I've had to take a break on this until mid-March, so any help would be appreciated.

littledan:
@gsathya has checked in a number of tests into V8; those might be a good starting point

sebmarkbage:
Also see these from WebKit https://github.com/sebmarkbage/ecmascript-rest-spread/issues/32#issuecomment-279269912

caiolima:
@sebmarkbage If you are ok, I can submit a PR with WebKit's tests here.

sebmarkbage:
@caiolima That'd be great!

leobalter:
Caio, for test262 perspective, this would be great and we can coordinate
the tests migration.

I'll send you an email and see if we can connect in a video call. It's not
really necessary but great as we can talk about it in our native language.
:)

caiolima:
@leobalter sure! I'm planing to work on that tomorrow and during the weekend. I've seen that array Rest/spread tests are slightly different from the last PR I've submitted, since they are generated from sources into src/dstr-assignment, but I can figure out its pattern.

caiolima:
@leobalter and @sebmarkbage, I've opened the PR #867 with a initial set of tests for Object Rest deconstruction. But I have 2 questions before go further:

1. What Am I supposed to put in esid field?
2. I want to add some syntax tests, but I'm confused if I should add them and also where should I place them.

I'm planing to create tests for Object Spread this weekend.

leobalter:
#867 #890 and #892 should be covering Object Rest Spread for now. I'll check if we find more edge cases - and compare with the tests on v8. I'll follow up w/ new PRs if I find more coverages to fill.

credits to @caiolima who did a great work here!

leobalter:
I am reopening this because of a few extra items:

- [ ] #1050
- [ ] #1132

adrianheine:
I think the negative tests for complex rest properties are still missing.

leobalter:
Thanks for letting me know, I'll get this fixed today.

PR #890, opened 2017-03-06
Initial set for Object Spread feature

caiolima:
This is an initial set of tests for Object Spread proposal. It is fixing #865.

To run these tests it is necessary generate them using ```make.py``` script. I tested them using the following command:

```test262-harness --hostPath=`which d8` --hostType=d8 --hostArgs="--harmony_object_rest_spread"```

### DIFF
--- a/src/spread/mult-obj-ident.case
+++ b/src/spread/mult-obj-ident.case
@@ -32,6 +32,14 @@ assert.sameValue(obj.c, 3);
 assert.sameValue(obj.d, 4);
 assert.sameValue(Object.keys(obj).length, 4);
 
+verifyEnumerable(obj, "a");
+verifyWritable(obj, "a");
+verifyConfigurable(obj, "a");
+
+verifyEnumerable(obj, "b");
+verifyWritable(obj, "b");
+verifyConfigurable(obj, "b");
+
 verifyEnumerable(obj, "c");
 verifyWritable(obj, "c");
 verifyConfigurable(obj, "c");
--- a/src/spread/mult-obj-ident.case
+++ b/src/spread/mult-obj-ident.case
@@ -1,0 +1,41 @@
+// Copyright (C) 2017 Caio Lima. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+desc: Object Spread operator following other properties
+template: default
+esid: pending
+includes: [propertyHelper.js]
+info: |
+    Pending Runtime Semantics: PropertyDefinitionEvaluation
+
+    PropertyDefinition:...AssignmentExpression
+
+    1. Let exprValue be the result of evaluating AssignmentExpression.
+    2. Let fromValue be GetValue(exprValue).
+    3. ReturnIfAbrupt(fromValue).
+    4. Let excludedNames be a new empty List.
+    5. Return CopyDataProperties(object, fromValue, excludedNames).
+
+---*/
+
+//- setup
+let o = {c: 3, d: 4};
+
+//- args
+{a: 1, b: 2, ...o}
+//- params
+obj
+//- body
+assert.sameValue(obj.a, 1);
+assert.sameValue(obj.b, 2);
+assert.sameValue(obj.c, 3);
+assert.sameValue(obj.d, 4);
+assert.sameValue(Object.keys(obj).length, 4);
+
+verifyEnumerable(obj, "c");
+verifyWritable(obj, "c");
+verifyConfigurable(obj, "c");
+
+verifyEnumerable(obj, "d");
+verifyWritable(obj, "d");
+verifyConfigurable(obj, "d");
--- a/src/spread/mult-obj-null-empty-undefined.case
+++ b/src/spread/mult-obj-null-empty-undefined.case
@@ -4,6 +4,7 @@
 desc: Object Spread operator following other arguments with null, undefined and empty object
 template: default
 esid: pending
+includes: [compareArray.js]
 info: |
     Pending Runtime Semantics: PropertyDefinitionEvaluation
 
@@ -18,7 +19,7 @@ info: |
 ---*/
 
 //- args
-{a: 1, ...null, b: 2, ...undefined, c: 3, ...{}, d: 4}
+{a: 1, ...null, b: 2, ...undefined, c: 3, ...{}, ...{...{}}, d: 4}
 //- params
 obj
 //- body
@@ -26,4 +27,4 @@ assert.sameValue(obj.a, 1);
 assert.sameValue(obj.b, 2);
 assert.sameValue(obj.c, 3);
 assert.sameValue(obj.d, 4);
-assert.sameValue(Object.keys(obj).length, 4);
+assert(compareArray(Object.keys(obj), ['a', 'b', 'c', 'd']));
--- a/src/spread/mult-obj-null-empty-undefined.case
+++ b/src/spread/mult-obj-null-empty-undefined.case
@@ -27,4 +27,4 @@ assert.sameValue(obj.a, 1);
 assert.sameValue(obj.b, 2);
 assert.sameValue(obj.c, 3);
 assert.sameValue(obj.d, 4);
-assert(compareArray(Object.keys(obj), ['a', 'b', 'c', 'd']));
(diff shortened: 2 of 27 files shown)

### CODE REVIEW

File: src/spread/obj-skip-non-enumerable.case [hunk 1]
```diff
@@ -0,0 +1,20 @@
...
+
+let o = {};
+Object.defineProperty(o, "b", {value: 3, enumerable: false});
+
+//- args
+{...o}
+//- params
+obj
+//- body
+assert.sameValue(Object.getOwnPropertyDescriptor(obj, "b"), undefined)
```
leobalter: `obj.hasOwnProperty` should do the trick.

src/spread/obj-with-overrides.case
```diff
@@ -0,0 +1,31 @@
...
+    5. Return CopyDataProperties(object, fromValue, excludedNames).
+
+---*/
+
+//- setup
+
+let o = {a: 2, b: 3}
+
+//- args
+{...o, a: 1, b: 7}
```
leobalter: I like mixing some properties, maybe you can add `c: 4` to `o` and `d: 5` to this line.

src/spread/sngl-obj-ident.case
```diff
@@ -0,0 +1,39 @@
...
+    1. Let exprValue be the result of evaluating AssignmentExpression.
+    2. Let fromValue be GetValue(exprValue).
+    3. ReturnIfAbrupt(fromValue).
+    4. Let excludedNames be a new empty List.
+    5. Return CopyDataProperties(object, fromValue, excludedNames).
+
+---*/
+
+//- setup
+let o = {c: 3, d: 4};
```
leobalter: it would be nice to add some _false(ish)_ values: `e: undefined, f: null, g: false` and other non regular values as well: `h: -0, i: Symbol('tests')` and even a cross reference: `j: o`

leobalter:
I just had another idea in mind:

```js
function *gen() {
  var o = {...{get [yield]() { return 'get yield' }}};
  
  return o;
}

var iter = gen();

iter.next();
var item = iter.next('dynamic_property_lol');

item.value.dynamic_property_lol // should be 'get yield'
```

This is probably not something we'll get through the tests generation tool, and we can follow some models from `test/language/expressions/class/accessor-name-static-computed-yield-expr.js`

a `git grep 'get yield'` will get more examples.

caiolima:
@leobalter updated PR following your comments. Please double check ```src/spread/obj-symbol-property.case``` and ```src/spread/obj-spread-order.case``` because they are failing in my v8 version here.

Thanks for these new cases!

littledan:
cc @gsathya

src/spread/mult-obj-ident.case
```diff
@@ -0,0 +1,41 @@
...
+//- params
+obj
+//- body
+assert.sameValue(obj.a, 1);
+assert.sameValue(obj.b, 2);
+assert.sameValue(obj.c, 3);
+assert.sameValue(obj.d, 4);
+assert.sameValue(Object.keys(obj).length, 4);
+
+verifyEnumerable(obj, "c");
```
leobalter: I know we want to assert the new properties, but just to confirm, we should also check the `a` and `b` properties.

src/spread/mult-obj-null-empty-undefined.case
```diff
@@ -0,0 +1,29 @@
...
+//- args
+{a: 1, ...null, b: 2, ...undefined, c: 3, ...{}, d: 4}
+//- params
+obj
+//- body
+assert.sameValue(obj.a, 1);
+assert.sameValue(obj.b, 2);
+assert.sameValue(obj.c, 3);
+assert.sameValue(obj.d, 4);
+assert.sameValue(Object.keys(obj).length, 4);
```
leobalter: if you include the harness/compareArray.js, you can also assert the order they are added. To avoid false positives, resort the order you include a, b, c, and d.

src/spread/mult-obj-null-empty-undefined.case
```diff
@@ -0,0 +1,29 @@
...
+    1. Let exprValue be the result of evaluating AssignmentExpression.
+    2. Let fromValue be GetValue(exprValue).
+    3. ReturnIfAbrupt(fromValue).
+    4. Let excludedNames be a new empty List.
+    5. Return CopyDataProperties(object, fromValue, excludedNames).
+
+---*/
+
+//- args
+{a: 1, ...null, b: 2, ...undefined, c: 3, ...{}, d: 4}
```
leobalter: one more item to add here for a trolling test: `...{...{}}`

src/spread/obj-getter-descriptor.case
```diff
@@ -0,0 +1,36 @@
...
+    }
+};
+
+//- args
+{...o, c: 4, d: 5}
+//- params
+obj
+//- body
+
+assert.sameValue(Object.getOwnPropertyDescriptor(obj, "a").value, 42);
```
leobalter: it might be worth checking the other property desc props for `a` here.

src/spread/obj-overrides-prev-properties.case
```diff
@@ -0,0 +1,31 @@
...
+let o = {a: 2, b: 3}
+
+//- args
+{a: 1, b: 7, ...o}
+//- params
+obj
+//- body
+assert.sameValue(obj.a, 2);
+assert.sameValue(obj.b, 3);
+assert.sameValue(Object.keys(obj).length, 2);
```
leobalter: to make sure the values are assigned and not a reference to `o`, assert `o.a` and `o.b` are still `2` and `3`.

src/spread/obj-symbol-property.case
```diff
@@ -0,0 +1,33 @@
...
+let o = {};
+o[Symbol('foo')] = 1;
+
+//- args
+{...o, c: 4, d: 5}
+//- params
+obj
+//- body
+
+assert.sameValue(obj[Symbol('foo')], 1);
```
leobalter: this will fail. You should store this symbol in a variable or use Symbol.for. Everytime you call Symbol('foo') it creates a new symbol. Storing a symbol will preserve it or using Symbol.for will find that reference.

src/spread/obj-spread-order.case
```diff
@@ -0,0 +1,29 @@
...
+---*/
+
+//- setup
+var calls = []
+var o = { get z() { calls.push('z') }, get a() { calls.push('a') } };
+
+//- args
+{...o}
+//- body
+assert.sameValue(calls[0], 'a');
```
leobalter: this should be `z` and then `a`, no? My best guess it that the properties orders are preserved as they are assigned.

caiolima: I think not. [CopyDataProperties](http://sebmarkbage.github.io/ecmascript-rest-spread/#AbstractOperations-CopyDataProperties) uses ```from.[[OwnPropertyKeys]]()```. [OwnPropertyKeys](https://tc39.github.io/ecma262/#sec-ordinary-object-internal-methods-and-internal-slots-ownpropertykeys) has a defined order for its returning list.

src/spread/obj-symbol-property.case
caiolima: My bad. Fixed.

caiolima:
@leobalter fixed cases. Thank you for the tips. And the errors were false alarm.

leobalter:
Thank you, @caiolima! That's a great work!

src/spread/mult-obj-null.case
```diff
@@ -0,0 +1,27 @@
...
+    1. Let exprValue be the result of evaluating AssignmentExpression.
+    2. Let fromValue be GetValue(exprValue).
+    3. ReturnIfAbrupt(fromValue).
+    4. Let excludedNames be a new empty List.
+    5. Return CopyDataProperties(object, fromValue, excludedNames).
+
+---*/
+
+//- args
+{a: 1, b: 2, ...null}
```
caitp: Many of the tests in this PR seem to be wrong, based on what is in the spec. I pointed this out on https://github.com/tc39/ecma262/pull/1048#discussion_r200538519

I understand that there's been some expectation that it should behave like this, but it's not what's specified, due to the way CopyDataProperties is specified normatively.

I think we should either update the spec, or update these tests.

PR #891, opened 2017-03-06
Test for parameters with default expressions in Annex B FiB tests

leobalter:
Closes gh-860

cc @syg @jugglinmike 

This is my first try to fix #860 using the test generation tool.

As you can see in [the new code](https://github.com/tc39/test262/compare/master...leobalter:860-block-dft-params?expand=1#diff-44c8f167f3b6d4d7f5e745fa3727baaf), I used both the previous copyright from V8 and included an extra line for Mozilla for the new change.

I'm not sure this is the best way to handle this case (wrt copyrights) but I'm willing to fix it if necessary.


syg:
I don't know anything about the test generation tool but the copyright lines LGTM.

PR #892, opened 2017-03-07
Add tests for object spread on yield value

leobalter:
This PR adds a very small set of tests for Object spread. After [my comment](https://github.com/tc39/test262/pull/890#issuecomment-284600429) on #890, I realized it would end up being faster if I just write these test files.

I can also use the generation tool if request, this set is so small I thought it was ok doing it this way.

Ref #865 
Ref #890

cc @caiolima can you help me reviewing this one?

### DIFF
--- a/src/generators/default/class-method-definition.template
+++ b/src/generators/default/class-method-definition.template
@@ -1,0 +1,33 @@
+// Copyright (C) 2017 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+path: language/expressions/class/gen-method-
+name: Generator method as a ClassElement
+esid: prod-GeneratorMethod
+info: |
+  ClassElement[Yield, Await]:
+    MethodDefinition[?Yield, ?Await]
+
+  MethodDefinition[Yield, Await]:
+    GeneratorMethod[?Yield, ?Await]
+
+  14.4 Generator Function Definitions
+
+  GeneratorMethod[Yield, Await]:
+    * PropertyName[?Yield, ?Await] ( UniqueFormalParameters[+Yield, ~Await] ) { GeneratorBody }
+---*/
+
+var callCount = 0;
+
+class C { *gen() {
+    callCount += 1;
+    /*{ body }*/
+}}
+
+var gen = C.prototype.gen;
+
+var iter = gen();
+
+/*{ assertions }*/
+
+assert.sameValue(callCount, 1);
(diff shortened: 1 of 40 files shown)


caiolima:
LGTM with a few comments.

I think it should be a good idea test sloppy mode with an identifier named ```yield```. IMO, these cases are important because Object Spread changes JS syntax and consumer parsers could implement it wrongly.

Also, if there is no case like that for Array Spread, this is a good opportunity to add them.

test/language/expressions/generators/yield-as-obj-property-spread.js
```diff
@@ -0,0 +1,28 @@
...
+info: |
+  Spread Properties
+
+  PropertyDefinition[Yield]:
+    (...)
+    ...AssignmentExpression[In, ?Yield]
+---*/
+
+var gen = function *() {
+  return {...yield, y: 1, ...yield};
```
leobalter: feedback from @gnarf: replace the return with an yield.

leobalter:
I'm going to bring these new tests you suggested in a follow up PR to seek direct reviews on each. the idea you brought for testing identifier named yield is amazing but I'm creating some extra cases for it.

leobalter:
Now it's everything here, with the tests for Object Spread, Array Spread and yield identifiers in Object spread assignment expressions as you requested.

It's ready for a new round of reviews

gnarf:
Pretty boggling stuff here, yield {...yield yield} :+1: :)

Do you think it'd be worth adding a template for `var gen = (class Thing { *gen() {} }).prototype.gen`?  Or am I just inventing cases for no reason? I guess I wonder beyond a certain point what grammar around the messy yield spreads matters.

leobalter:
While it seems compelling to add that as well, this is a [form of method definition](https://tc39.github.io/ecma262/#prod-ClassElement) as well.

In theory, by spec text, it's not adding coverage, but I wonder if runtimes define them differently.

leobalter:
@gnarf this proposed template would conflict with the current cases containing [`yield` identifiers](https://github.com/tc39/test262/pull/892/files#diff-52309e6dfc9310c98186906faffa5794R17) as class bodies run in strict mode.

I'm trying to figure out how to avoid this conflict right now.

gnarf:
> I'm trying to figure out how to avoid this conflict right now.

don't let my half-baked idea stop you from landing some useful tests, we can work on this in a followup - maybe document the desire in an issue?

leobalter:
I did some sort of split in the templates. The trade off is that I'm literally cloning 3 out of 4 templates, but it generates what I need to be generated.

As you mentioned, I think this can be improved in a follow up PR, but if everything is current correct, it should be time to land.

leobalter:
Rebased and generated tests.

leobalter:
Thanks, @caiolima and @gnarf!

PR #895, opened 2017-03-10
Test for [no LineTerminator here] after 'async' for arrows and methods

syg:


test/language/expressions/async-arrow-function/early-errors-arrow-ident-lineterminator.js
```diff
@@ -0,0 +1,13 @@
+// Copyright (C) 2017 Mozilla Corporation. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: prod-AsyncArrowFunction
+description: async arrows cannot have a line terminator between "async" and the AsyncArrowBindingIdentifier
+negative:
+  phase: early
```
leobalter: isn't this a runtime phase? cc @bakkot

syg:
@leobalter Oops, you're right. The line terminator causes the 'async' to parse as an identifier with ASI.

leobalter:
It's alright. It's interesting as I thought I would already have tests for this here. I'll search the code again and if I'm correct, I'll also use these tests for other async fn cases.

leobalter:
I really like these tests. I probably missed posting my comment a while ago: I want to convert them to use the generation tool for the many async function forms.

I'll do it soon, in the meanwhile, this can land if I fix the phase.

PR #898, opened 2017-03-11
Add missing test cases for set __proto__

### LINKED ISSUE #887: Missing test cases for set __proto__
The ```test262/test/annexB/built-ins/Object/prototype/__proto__/set-non-object.js``` is testing https://tc39.github.io/ecma262/#sec-set-object.prototype.__proto__ step 2 return, however it is missing to assert if ```obj.__proto__``` was changed.

Zirak:
Fixes #887

Stolen from the sibling file `set-invalid-value.js`.


Zirak:
Hurray!

PR #907, opened 2017-03-13
Add tests for instances of Async/Generator Functions that are not con…

### LINKED ISSUE #779: Missing test for Functions constructed via GeneratorFunction() or AsyncFunction() are not constructors
In WebKit, the following test fails:

``` js
var GeneratorFunction = function*() {}.constructor;
try {
  new (GeneratorFunction());
  console.log("FAIL: did not throw");
} catch (e) {
  console.log("PASS: threw " + String(e));
}
```

I was surprised that there's no test coverage for this in test262, because the text of [CreateDynamicFunction](https://tc39.github.io/ecma262/#sec-createdynamicfunction) is fairly clear:

> ``` text
>  27. If kind is "generator", then
>     a. Let prototype be ObjectCreate(%GeneratorPrototype%).
>     b. Perform DefinePropertyOrThrow(F, "prototype",
>             PropertyDescriptor{[[Value]]: prototype,
>                                [[Writable]]: true,
>                                [[Enumerable]]: false,
>                                [[Configurable]]: false}).
>  28. Else, perform MakeConstructor(F).
> ```

and of course, this is also true for dynamically created async functions. (of course, the relevant part is really the parameters passed to FunctionAllocate(), but `MakeConstructor` speaks for itself).

leobalter:
…structors

Closes #779


test/built-ins/GeneratorFunction/instance-construct-throws.js
```diff
@@ -0,0 +1,29 @@
+// Copyright (C) Copyright 2017 the V8 project authors. All rights reserved.
```
leobalter: typo

leobalter:
I'm adding a review item to myself. yolo

leobalter:
fixed.

PR #908, opened 2017-03-13
Remove invalid String#indexOf test in a coercible Date instance

### LINKED ISSUE #793: Apparently invalid String.prototype.indexOf() test on Date(0)
* https://github.com/tc39/test262/blob/master/test/built-ins/String/prototype/indexOf/S15.5.4.7_A1_T11.js
* ES5: ch15/15.5/15.5.4/15.5.4.7/S15.5.4.7_A1_T11.js

This test takes a Date(0) object, sets its indexOf to that of String.prototype and expects it to find "31" or "01" (depending on getTimeZoneOffset()'s sign) at index 8.
If this was assuming use of the Date Time String Format, it would fall foul of "YYYY-MM-dd" having MM before the dd, so matching at index 5 (as "1970-01-01" has a "01" early).
It thus appears to be assuming something like "DDD MMM 01" (e.g. Thu Jan 01) as a prefix of the format Date(0) tacitly uses to .toString() itself as input to indexOf(), but either way it's clearly expecting a two-digit day number after 8 prior characters (nowhere in which "01" is apt to appear).

In ES5, Date.prototype.toString() says: "This function returns a String value. The contents of the String are implementation-dependent, but are intended to represent the Date in the current time zone in a convenient, human-readable form."

In ES7, Date(0) flows through the given algorithm to ToDateString, which should "Return an implementation-dependent String value that represents tv as a date and time in the current time zone using a convenient, human-readable form."

In either case, the String is implementation-dependent and there is nothing constraining it to use a two-digit day-of-month (so "Thu Jan 1, 1970, 0:0:0" would be valid), much less to position it at index 8.
This test thus appears to test beyond what it specified.

Relevant specifications:
* String.prototype.indexOf:
* ES7: http://www.ecma-international.org/ecma-262/7.0/index.html#sec-string.prototype.indexof
* ES5: http://www.ecma-international.org/ecma-262/5.1/#sec-15.5.4.7
* Date.prototype.toString:
* ES7: http://www.ecma-international.org/ecma-262/7.0/index.html#sec-date.prototype.tostring
* ES5: http://www.ecma-international.org/ecma-262/5.1/#sec-15.9.5.2
* ToDateString:
* ES7: http://www.ecma-international.org/ecma-262/7.0/index.html#sec-todatestring
* Date Time String Format (not actually relevant, as far as I can see):
* ES7: http://www.ecma-international.org/ecma-262/7.0/index.html#sec-date-time-string-format
* ES5: http://www.ecma-international.org/ecma-262/5.1/#sec-15.9.1.15

leobalter:
Closes #793

As described by @ediosyncratic:

Date(0) flows through the given algorithm to ToDateString, which should:

> Return an implementation-dependent String value that represents tv as
> a date and time in the current time zone using a convenient, human-readable form.

The string is implementation-dependent and there is nothing constraining it to use
a two-digit day-of-month (so "Thu Jan 1, 1970, 0:0:0" would be valid), much less to
position it at index 8.


leobalter:
s/Fix/Remove :)

rwaldron:
LGTM

PR #911, opened 2017-03-13
[DON'T SQUASH] Update params trailing comma tests

### LINKED ISSUE #822: Missing test language/statements/generators/params-trailing-comma-arguments.js
All the other things in the matrix seem to be there, and this is an unexpected hole. cc @jeffmo

leobalter:
Closes #822


leobalter:
cc @littledan

leobalter:
these tests are invalid. I'm fixing this PR soon


```js
function* f1() {
    length = arguments.length;
};
f1(1,).next();
```

this has no connection with the trailing commas on the parameters list, but the arguments object and how the length is formed to correspond the given arguments list.

Instead of adding this extra case, I'm removing all the `params-trailing-comma-arguments.js` files.

leobalter:
In the meanwhile, I need to search for tests asserting the well formed arguments.length

leobalter:
`test/language/expressions/call/trailing-comma.js` is the only other test for trailing comma in a call expression, but there's no tests for the arguments.length

leobalter:
I can add more cases but that would create more and more files and it's becoming way larger on every new addition.

I suggest we follow up with new PRs to have easier reviews.

leobalter:
these tests will open the way for me to add more stuff on async generators, early errors, etc. I'll have this merged but already assuming the responsibility to fix anything else we find wrong.

PR #914, opened 2017-03-14
Avoid top level name and length bindings

### LINKED ISSUE #850: Rename uses of global vars named `name` and `length`
There are tests that use `var name` at the top level. This ends up triggering the `name` setter on the WindowProxy in browsers. Ditto for `length`. We should rename these var names, as well disallow future tests from using them at the global level.

The web runner from I think does a regexp on the test source, and if the test contain uses of `name` or `length`, it actually deletes the properties from the window. @bakkot, is that right?

leobalter:
Closes #850 

cc @syg, this is the best I could do to fix. We need to add this as soon as we have some code style document (#297)

### DIFF
--- a/src/dstr-assignment/array-rest-nested-obj-undefined-hole.case
+++ b/src/dstr-assignment/array-rest-nested-obj-undefined-hole.case
@@ -12,7 +12,9 @@ es6id: 12.14.5.3
 
 //- setup
 var x = null;
-var length;
+// Use the the top-level lexical scope for 'length' to provide compatibility with browsers
+// where length and name are properties of WindowProxy
+let length;
 //- elems
 [...{ 0: x, length }]
 //- vals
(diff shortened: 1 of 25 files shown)


bakkot:
They'll need to be declared as `let` instead of `var` for the wrapping-in-a-block trick to work.

And I think just declaring them as `let` would be sufficient - IIRC, the top-level lexical scope is nested instead of the global scope, so if they were declared as `let` the block would be redundant.

leobalter:
> They'll need to be declared as let instead of var for the wrapping-in-a-block trick to work.

oops.

> And I think just declaring them as let would be sufficient - IIRC, the top-level lexical scope is nested instead of the global scope, so if they were declared as let the block would be redundant.

+1, thanks for the reminder.

leobalter:
this should be fixed now. @bakkot, would you mind verifying this?

PR #916, opened 2017-03-15
Add tests for computed __proto__ property keys

### LINKED ISSUE #904: Missing coverage for computed property key "__proto__"
based on https://github.com/tc39/test262/issues/423#issuecomment-133295358

We don't have tests for the `['__proto__']` variation in the example above, using a _ComputedPropertyName_.

```js
var a = {
  __proto__: { abc: 123 }
};
var b = {
  ['__proto__']: { abc: 123 }
};
a.abc; // 123
b.abc; // undefined
```

I'll write the tests based in the current `test/annexB/language/expressions/object/__proto__-value*` files we already have.

---

a pre-check on Chrome, Firefox and Safari shows the expected behavior as in the example.

leobalter:
Closes #904


leobalter:
cc @tuchida. Is this a valid fix or do you think I need to add another case?

tuchida:
LGTM

PR #917, opened 2017-03-15
Add case for obj rest name equal to a property

leobalter:
cc @caiolima

### DIFF
--- a/src/dstr-assignment/obj-rest-same-name.case
+++ b/src/dstr-assignment/obj-rest-same-name.case
@@ -1,0 +1,32 @@
+// Copyright (C) 2017 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: >
+    Proper setting in the values for rest name equal to a property name.
+template: default
+esid: pending
+---*/
+
+//- setup
+var o = {
+    x: 42,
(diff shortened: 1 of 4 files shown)


src/dstr-assignment/obj-rest-same-name.case
```diff
@@ -0,0 +1,33 @@
+// Copyright (C) 2017 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: >
+    Proper setting in the values for rest name equal to a property name.
+template: default
+esid: pending
+includes: [propertyHelper.js]
```
caiolima: I think you don't need this.

leobalter: you're right, thanks!

leobalter: There are more cases I repeat this, I'll fix all of them!

rwaldron:
LGTM

leobalter:
Somehow I forgot to push the fixes for this.

rwaldron:
I've actually just reviewed this twice (accidentally), once before a meeting and then again after the meeting.

LGTM.

leobalter:
Thank you! Thank you! :)

PR #918, opened 2017-03-16
Add async generator templates

leobalter:
reusing the generator templates and cases, these tests are mostly used to assert yield bindings and yield spread operations.

It also opens the way to create more tests for async generators using the templates.


leobalter:
cc @ajklein @domenic this might help the initial multiple shapes for async generators in v8. I look forward on review feedbacks.

test/language/statements/class/async-gen-method-static-yield-spread-arr-multiple.js
```diff
@@ -0,0 +1,51 @@
...
+    yield [...yield yield];
+}}
+
+var gen = C.gen;
+
+var iter = gen();
+
+iter.next(false);
+item = iter.next(['a', 'b', 'c']);
+item = iter.next(item.value);
```
caitp: I'm confused what `item.value` is supposed to refer to. Was there a spec change I missed?

leobalter: that's something wrong in the generation. I'll fix it. Thanks!

leobalter: The new commits should fix this.

leobalter: and thanks for catching this, this was definitely my error when generating this test.

test/language/expressions/async-generator/named-yield-identifier-spread-non-strict.js
```diff
@@ -0,0 +1,59 @@
...
+iter.next();
+iter.next();
+iter.next({ x: 10, a: 0, b: 0 });
+iter.next({ y: 20, a: 1, b: 1 });
+var item = iter.next({ z: 30, b: 2 });
+
+item.then(({ done, value }) => {
+  assert.sameValue(done, false);
+  assert.sameValue(value.x, 10);
+  assert.sameValue(value.y, 20);
```
caitp: maybe some of these should have a Symbol-keyed property, just for variety? I havent read all these so if it's already there, sorry for the noise

test/language/expressions/async-generator/named-yield-spread-arr-single.js
```diff
@@ -0,0 +1,42 @@
...
+  yield [...yield];
+};
+
+var iter = gen();
+
+iter.next(false);
+var item = iter.next(['a', 'b', 'c']);
+
+item.then(({ done, value }) => {
+  assert(compareArray(value, arr));
```
caitp: might be good to assert value !== the array sent to the generator, just for a little extra cov

domenic:
Thanks so much for this work!!

For such a large body of tests my preferred approach is to get an implementer or two to run them and see if any results are unexpected. This can even be done after landing as part of an iterative process. Basically I think it's more valuable to get these tests into the common repo than to make sure they're perfect beforehand.

That said @caitp appears to have gone over them in some detail and found some coverage improvement ideas just by inspection, so that's awesome too.

But in general, if they're up for it I'd like to defer to any of @caitp, @arai-a, and @gskachkov.

test/language/expressions/async-generator/named-yield-spread-arr-single.js
leobalter: Thanks for flagging this! I was already passing another array: `iter.next(['a', 'b', 'c'])` but I'm doing a cleanup fix to clarify it based on your suggestion.

leobalter:
@caitp I tried to address your feedback! The new commits are up! Thanks!

src/async-generators/default/async-class-decl-method.template
```diff
@@ -0,0 +1,34 @@
...
+  AsyncGeneratorMethod :
+    async [no LineTerminator here] * PropertyName ( UniqueFormalParameters ) { AsyncGeneratorBody }
+---*/
+
+var callCount = 0;
+
+class C { async *gen() {
+    callCount += 1;
+    /*{ body }*/
+}}
```
leobalter: I'm following the pattern here: https://github.com/tc39/test262/blob/master/src/generators/default/class-method-definition.template#L22

It confused me at first, but this is a good hack to reuse the same indentation from the cases as they are using four spaces for the other function forms.

It's a trade off, the generated tests become easier to read this way.

leobalter: e.g.: https://github.com/tc39/test262/pull/918/files#diff-dfa2c8a27fd52f9e3cef766b26e6ca86R27

leobalter: GitHub is giving me a hard time to show me this link, so here goes an example of a generated test:

```js
class C { static async *gen() {
    callCount += 1;
    yield [...yield yield];
}}
```

`    yield [...yield yield];`  comes from a case body.

caitp:
fwiw, I've just give these a run on v8, all of these tests are passing in v8 as of bc12053a6a0cc966ef3e4c6277343e33e0b119fa

leobalter:
Thank you, @caitp!

I have a v8 (d8) locally compiled using the `5.9.59` tag/branch and I can't get it to accept the syntax. 

I tried the `--harmony_async_iteration` and `--harmony` args but it's still not working for me. I'm probably missing something there I wonder if you can help me.

---

current calls are basically: 

```
test262-harness --hostPath=`which d8` --hostType=d8 --hostArgs='--harmony_async_iteration' $(git diff --cached --name-only) -t 4
```

```
test262-harness --hostPath=`which d8` --hostType=d8 --hostArgs='--harmony_async_iteration' $(git diff HEAD^1 --name-only) -t 4
```

caitp:
currently, only for-await-of in async functions, and the async-from-sync iterator, are upstream.

Async Generators can be used by applying code from https://chromium-review.googlesource.com/#/c/446961/

I am hoping to land the two CLs this week, but can't guarantee it.

leobalter:
@caitp the last couple commits add the `features: [async-iteration]` on each test file.

caitp:
I've taken a quick look at the feature flags for the generated tests, and I think they should work well for v8 (assuming @littledan or someone can edit the test runner to account for those feature flags).

Looks good to me

PR #920, opened 2017-03-16
Add async-gen templates for params in the test gen tool

leobalter:
Follow up for #918, this PR adds async-generator templates in the src/params folder

### DIFF
--- a/src/params/default/cls-decl-async-gen-meth-static.template
+++ b/src/params/default/cls-decl-async-gen-meth-static.template
@@ -1,0 +1,56 @@
+// Copyright (C) 2017 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+path: language/statements/class/params-async-gen-meth-static-
+name: static class expression generator method
+esid: sec-runtime-semantics-bindingclassdeclarationevaluation
+info: |
+    ClassDeclaration : class BindingIdentifier ClassTail
+
+    1. Let className be StringValue of BindingIdentifier.
+    2. Let value be the result of ClassDefinitionEvaluation of ClassTail with
+       argument className.
+    [...]
+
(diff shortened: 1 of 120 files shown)


src/params/default/cls-decl-async-gen-meth-static.template
```diff
@@ -0,0 +1,55 @@
...
+    /*{ body }*/
+    callCount = callCount + 1;
+  }
+}
+
+// Stores a reference `ref` for case evaluation
+var ref = C.method;
+
+ref(/*{ args }*/).next().then(() => {
+    assert.sameValue(callCount, 1, 'method invoked exactly once');
```
leobalter: I wonder if I even need these `.then` calls.

At the time I run `next()`, I believe `callCount` will get the +1 (line 46) already and this doesn't need to be an async test at all. Remember there's no yield neither await uses for the respective cases here.

In any way, it does not hurt having these promise calls, the tests are not broken because of it.

rwaldron: What was the original rationale for inclusion?

leobalter: I'm over protecting these tests w/ guarantee I'll observe callCount. That's the basic rationale.

Not being able yet to verify these tests locally makes me go overzealous. At least these are not making any test invalid.

caitp:
I've gone ahead and run these locally against https://chromium-review.googlesource.com/c/446961/, and they are all passing (I haven't looked at all of these closely, so those "passes" may not necessarily mean much in all cases)

leobalter:
as in #918, the last commits are also adding the features flag.

leobalter:
> I've gone ahead and run these locally against https://chromium-review.googlesource.com/c/446961/, and they are all passing

for now this is the best verification we can have. I'm willing to fix anything else we might find later.

PR #921, opened 2017-03-17
Add async generator templates for destructing binding

leobalter:
There's more on the way for these templates, but I'm trying to split them to help on reviews.


leobalter:
as in #918 and #920, this PR now provides the async-iteration features flag.

littledan:
As long as you're testing destructuring binding in async generators, it would be nice to make sure that the scopes of default arguments work (including, when there's a direct eval in a default argument, it receives the right scopes). I don't see eval in these tests.

leobalter:
This is definitely something we should add for more tests. Not restricted to async-generators only but directly related to dstr-binding, so that applies for the other function forms in the templates.

We have templates for dflt parameters in these templates, like:

```js
async function* f(/*{ elems }*/ = /*{ vals }*/) {
...

function f(/*{ elems }*/ = /*{ vals }*/) {
```

but we could investigate using stuff like:

```js
async function* f(/*{ elems }*/ = eval(/*{ vals }*/)) {
...

function f(/*{ elems }*/ = eval(/*{ vals }*/)) {
```

leobalter:
this is weird and nice:

```js
function f1([[x, y, z] = eval('[4, 5, 6]')]) {
  console.log(x,y,z);
}

f1([]);

f1([[7,8, 9]]);

function f2([[x, y, z] = eval('[4, 5, 6]')] = eval('[[7, 8]]')) {
  console.log(x,y,z);
}

f2(); / 7 8 undefined
```

@littledan are these attending your suggestion? Consider the function form can be extended and used in the many other forms from that template folder.

If it works for you, I suggest we open a new issue to keep track of this while I can't immediately write the tests right now.

PR #927, opened 2017-03-22
Test for TypedArray.prototype.fill semantics change

littledan:
The change is proposed in https://github.com/tc39/ecma262/pull/856
as a fix to https://github.com/tc39/ecma262/issues/855

Here, the ToNumber coercion is done only once, rather than on each
iteration. It does not appear that there were previously any
tests against repeated coercion for this parameter previously.

Tested this test against V8, which failed, as V8 implements the
current spec rather than the proposed one.

### DIFF
--- a/test/built-ins/TypedArray/prototype/fill/fill-values-conversion-once.js
+++ b/test/built-ins/TypedArray/prototype/fill/fill-values-conversion-once.js
@@ -2,7 +2,6 @@
 // This code is governed by the BSD license found in the LICENSE file.
 /*---
 esid: sec-%typedarray%.prototype.fill
-es6id: 22.2.3.8
 description: >
   Fills all the elements with non numeric values values.
 info: >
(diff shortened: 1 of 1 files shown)


test/built-ins/TypedArray/prototype/fill/fill-values-conversion-once.js
```diff
@@ -0,0 +1,27 @@
+// Copyright (C) 2017 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-%typedarray%.prototype.fill
+es6id: 22.2.3.8
```
leobalter: this is not a reflection of ES6 anymore, we should get rid of this es6id tag for this test.

littledan: Fixed

leobalter:
this reached consensus in March 22nd at the last TC39 meeting.

PR #933, opened 2017-03-24
Correct typo in annexB test file names and templates: exsting -> existing

dilijev:
* Renamed all affected test and template files
* Updated references to affected files from generated files

Commands used:

```sh
git ls-files | grep exsting     # used list to rename files
git grep -l exsting | xargs -I % sed -i % -e "s/\-exsting\-/-existing-/g"
```


dilijev:
Sorry, I should have squashed to a single commit as per the instructions. However, I think the two commits might still be useful for review, so here's the original set of commits: https://github.com/tc39/test262/compare/master...dilijev:typo-original

leobalter:
@dilijev Thanks for the PR!

I actually need the changes to in the source files and than applied to the generated tests through the automated process.

If we could have these changes in 2 separate commits it would be great and easy to review.

dilijev:
Looks like the script wants to write \r\n and backslashes for paths on Windows. I'm assuming that's not the intended behavior (especially given the unnecessary diffs that introduces and I'll re-run the script on Linux).

```
-// - src/annex-b-fns/eval-func-exsting-var-no-init.case
-// - src/annex-b-fns/eval-func/direct-if-decl-else-decl-a.template
+// - src\annex-b-fns\eval-func-existing-var-no-init.case
+// - src\annex-b-fns\eval-func\direct-if-decl-else-decl-a.template
```

dilijev:
Whoops, I accidentally deleted the remote branch which closed this PR. Opened #939

PR #934, opened 2017-03-27
Add feature flags for tests on object rest spread

leobalter:
cc @caitp 

I'm adding a frontmatter `features` flag for tests w/ Object rest/spread properties.

Next step is updating the async-iterator tests in the same way.


caitp:
I don't know if there are any omissions, but the changes look good. Thanks for doing this!

leobalter:
> I don't know if there are any omissions

Those files are all I've got from several searches. I hope this helps flagging the tests for new features, I'll keep this as a pattern.

PR #936, opened 2017-03-27
Fix frontmatter tags

leobalter:


rwaldron:
Glad we got this resolved.

PR #942, opened 2017-03-28
Add escape-above-astral.js to cover escaping code units as encoded from extended unicode escape.

dilijev:
This is more of an integration test because the semantics should be that the extended unicode escape is encoded in the string as two UTF-16 code units (surrogate pairs) which are then correctly decoded and escaped.

See also (should I include these in the test file itself in addition to the reference on escape?)

https://tc39.github.io/ecma262/#sec-static-semantics-sv

>The SV of UnicodeEscapeSequence::u{HexDigits} is the UTF16Encoding of the MV of HexDigits.

https://tc39.github.io/ecma262/#sec-utf16encoding


leobalter:
> See also (should I include these in the test file itself in addition to the reference on escape?)

The reference I got was enough for me to verify the test, I appreciate the well formatted test.

PR #944, opened 2017-04-03
Tests for using internal byteOffset in TypedArray.prototype.set

tunz:
Tests for checking byteOffset usage in TypedArray.prototype.set. It was violated by V8
(https://crbug.com/v8/6120).


test/built-ins/TypedArray/prototype/set/typedarray-arg-target-byteoffset-internal.js
```diff
@@ -0,0 +1,44 @@
+// Copyright 2017 the V8 project authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
```
leobalter: Why is this broken in an extra line?

tunz: I pushed a version of shorter license.

PR #946, opened 2017-04-05
Fix offset for detached TypedArray test

littledan:
Bug was reported by @anba at
https://github.com/tc39/ecma262/pull/852#issuecomment-291781031

Without this change, you'd expect a RangeError rather than a TypeError.


rwaldron:
LGTM cc @leobalter

PR #954, opened 2017-04-06
Test fixes

littledan:


src/dstr-binding/default/async-gen-func-named-expr-dflt.template
```diff
@@ -19,7 +19,7 @@ flags: [async]
 
 var callCount = 0;
 var f;
-f = async function* g(/*{ elems }*/ = /*{ vals }*/) {
+f = async function* h(/*{ elems }*/ = /*{ vals }*/) {
```
caitp: I'm unclear on what the point of this is. Is `g` used by `{ body }` or `{ vals }` or something?

littledan: Maybe this is to tickle a different area in parsing/scope resolution?

caitp: I was asking the rationale for changing the line. Is that the rationale? If so, it would be good to document that in the commit message

leobalter: I'm still trying to find any conflict this caused.

I'm happy you fixed it :)

leobalter: I've got it! There are cases defining a generator named `g`. 

e.g.: test/language/expressions/async-generator/dstr-named-ary-ptrn-elem-ary-elision-init.js

```js
var first = 0;
var second = 0;
function* g() {
  first += 1;
  yield;
  second += 1;
};


var callCount = 0;
var f;
f = async function* g([[,] = g()]) {
  assert.sameValue(first, 1);
  assert.sameValue(second, 0);
  callCount = callCount + 1;
};

f([]).next().then(() => {
    assert.sameValue(callCount, 1, 'invoked exactly once');
}).then($DONE, $DONE);
```

leobalter: lol for the search I used:

```
git grep 'g(' `git grep --name-only src/dstr-binding/default/async-gen-func-named-expr.template`
```

leobalter:
I updated the commit description and landed this manually.

Thanks, @littledan!

PR #955, opened 2017-04-07
Add Function.prototype.toString tests for async generator functions

anba:
The tests were derived from the existing Function.prototype.toString tests for async functions.


test/built-ins/Function/prototype/toString/AsyncGenerator.js
```diff
@@ -0,0 +1,16 @@
+// Copyright 2017 André Bargull. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: pending
+description: >
+  Function.prototype.toString on an async generator created with the
+  AsyncGenerator constructor.
+features: [async-iteration]
+---*/
```
leobalter: @anba would you mind to add an `info` tag with the respective spec reference?

This is helping us a lot to review and even with information after they land in the code base.

test/built-ins/Function/prototype/toString/async-generator-method.js
```diff
@@ -0,0 +1,14 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: pending
+description: Function.prototype.toString on an async generator method
+features: [async-iteration]
+---*/
+
+let f = { /* before */async /* a */ * /* b */ f /* c */ ( /* d */ ) /* e */ { /* f */ }/* after */ }.f;
+let g = { /* before */async /* a */ * /* b */ [ /* c */ "g" /* d */ ] /* e */ ( /* f */ ) /* g */ { /* h */ }/* after */ }.g;
```
leobalter: while this wrapped form is nice, it would be also nice to add a covered name with an expression to be evaluated.

example:

```js
let x = 'y';
let y = { /* before */async /* a */ * /* b */ [ /* c */ x /* d */ ] /* e */ ( /* f */ ) /* g */ { /* h */ }/* after */ }.y;
```

leobalter:
All the current tests are correct, the missing parts is something we can work as a follow up if you prefer, @anba.

test/built-ins/Function/prototype/toString/AsyncGenerator.js
anba: This could be problematic: https://github.com/tc39/Function-prototype-toString-revision doesn't cover async generator functions and https://github.com/tc39/proposal-async-iteration doesn't cover the Function.prototype.toString changes. 😕

anba:
I'll also update the async function tests to include the requested changes.

caitp:
Could you please add `features: [async-iteration]` to the front-matter for all of these which use features from that proposal, and ` [async-functions]` for tests that rely on async non-generators? This makes the tests easier to consume when features are behind a flag.

leobalter:
This update is amazing, @anba!

@caitp I'm glad to see the features tag is working. I'll write some documentation to cross reference features with their respective flag names, this way we secure consistent use for `async-functions`, `async-iteration`, etc.

test/built-ins/Function/prototype/toString/async-arrow-function.js
```diff
@@ -0,0 +1,15 @@
+// Copyright 2017 André Bargull. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-async-arrow-function-definitions-runtime-semantics-evaluation
+description: Function.prototype.toString on an async arrow function
+---*/
```
leobalter: I believe this is the only part here in this PR missing a `features: [async-functions]`.

PR #956, opened 2017-04-07
Add tests for the behavior of |let <newline> await| in normal and async functions

jswalden:
The behavior of

    let
    await 0;

differs depending whether it's in an async function or not.  In an async function, `LexicalDeclaration`'s binding name production excludes `await` from permissibility, so ASI can trigger to split this into two statements.  In a normal function, `await` isn't excluded, so `let await 0;` is observed and causes a syntax error.

Hat-tip to @anba for originally reporting this in [bug 1353690](https://bugzilla.mozilla.org/show_bug.cgi?id=1353690).

### DIFF
--- a/test/language/statements/async-function/let-newline-await-in-async-function.js
+++ b/test/language/statements/async-function/let-newline-await-in-async-function.js
@@ -1,0 +1,23 @@
+// Copyright (C) 2017 Mozilla Corporation. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+author: Jeff Walden <jwalden+code@mit.edu>
+esid: sec-let-and-const-declarations
+description: >
+  `await` must not be considered a permissible binding name in
+  LexicalDeclaration as used in async functions.
+info: >
+  LexicalDeclaration is parametrized to indicate whether `async` is permitted as
+  binding name.  In async functions `await` is excluded from LexicalDeclaration
+  as a binding name.  Therefore ASI *can* apply between `let` (where a
+  LexicalDeclaration is permitted) and `await`, so a subsequent `0` forms part
+  of an AwaitExpression and there is no syntax error.
+---*/
+
+async function f() {
+    let
+    await 0;
+}
+
+assert(f instanceof Function);
--- a/test/language/statements/let/syntax/let-newline-await-in-normal-function.js
+++ b/test/language/statements/let/syntax/let-newline-await-in-normal-function.js
@@ -1,0 +1,24 @@
+// Copyright (C) 2017 Mozilla Corporation. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+author: Jeff Walden <jwalden+code@mit.edu>
+esid: sec-let-and-const-declarations
+description: >
+  `await` must be considered a permissible binding name in LexicalDeclaration as
+  used in non-async functions.
+info: >
+  LexicalDeclaration is parametrized to indicate whether `async` is permitted as
+  binding name.  In non-async functions `await` is a perfectly cromulent binding
+  name.  Therefore ASI can't apply between `let` (where a LexicalDeclaration is
+  permitted) and `await`, so a subsequent `0` where `=` was expected is a syntax
+  error.
+negative:
+  phase: early
+  type: SyntaxError
+---*/
+
+function f() {
+    let
+    await 0;
+}


test/language/statements/async-function/let-newline-await-in-async-function.js
```diff
@@ -0,0 +1,18 @@
...
+author: Jeff Walden <jwalden+code@mit.edu>
+esid: sec-let-and-const-declarations
+description: >
+  |await| is excluded from LexicalDeclaration by grammar parameter, in
+  AsyncFunction.  Therefore |let| followed by |await| inside AsyncFunction is
+  an ASI opportunity, and this code must parse without error.
+---*/
+
+async function f() {
+    let
```
caitp: I am not sure how this is an ASI opportunity, since `let` by itself isn't a valid ExpressionStatement.

Is there something that makes this legal?

jswalden: Sure it is -- `ExpressionStatement` lookahead forbids `let [`, but it doesn't forbid `let` followed by some token other than `[`.  And plain old `let` is a perfectly cromulent `PrimaryExpression`.

caitp: Ah you're right. I was reading the spec on my phone and missed the `[`

test/language/statements/let/let-newline-await-in-normal-function.js
```diff
@@ -0,0 +1,20 @@
+// |reftest| error:SyntaxError
```
leobalter: this is for spidermonkey, right? I believe it will and should parse the negative flag already in the frontmatter

test/language/statements/let/let-newline-await-in-normal-function.js
```diff
@@ -0,0 +1,20 @@
...
+// Copyright (C) 2017 Mozilla Corporation. All rights reserved.
```
leobalter: this file can be placed in the `test/language/statements/let/syntax` folder where we keep the tests for syntax errors.

Btw, you already authored other tests in there :) see `test/language/statements/let/syntax/let-let-declaration-split-across-two-lines.js`

test/language/statements/async-function/let-newline-await-in-async-function.js
```diff
@@ -0,0 +1,18 @@
+// Copyright (C) 2017 Mozilla Corporation. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+author: Jeff Walden <jwalden+code@mit.edu>
+esid: sec-let-and-const-declarations
+description: >
+  |await| is excluded from LexicalDeclaration by grammar parameter, in
+  AsyncFunction.  Therefore |let| followed by |await| inside AsyncFunction is
+  an ASI opportunity, and this code must parse without error.
```
leobalter: can we add an info tag with the reference to the grammar?

rwaldron: While you're in there, change the pipes to ticks. (If I'm too late, then don't sweat it, we can change later)

test/language/statements/async-function/let-newline-await-in-async-function.js
```diff
@@ -0,0 +1,23 @@
...
+  LexicalDeclaration is permitted) and `await`, so a subsequent `0` forms part
+  of an AwaitExpression and there is no syntax error.
+---*/
+
+async function f() {
+    let
+    await 0;
+}
+
+reportCompare(true, f instanceof Function);
```
leobalter: what is reportCompare?

This should be ` assert(f instanceof Function);`

test/language/statements/async-function/let-newline-await-in-async-function.js
```diff
@@ -0,0 +1,23 @@
...
+  binding name.  In async functions `await` is excluded from LexicalDeclaration
+  as a binding name.  Therefore ASI *can* apply between `let` (where a
+  LexicalDeclaration is permitted) and `await`, so a subsequent `0` forms part
+  of an AwaitExpression and there is no syntax error.
+---*/
+
+async function f() {
+    let
+    await 0;
+}
```
littledan: This case is so bad that it makes me want to change the spec to instead have a static semantic rule prohibiting await and yield as binding identifiers, the way that strict mode and module cases are. Actually, it'd be good to add tests for those cases, to ensure that ASI is prohibited there.

rwaldron: > change the spec to instead have a static semantic rule prohibiting await and yield as binding identifiers, the way that strict mode and module cases are.

I support this. Let's drag @bterlson into the discussion.

leobalter: Same. That's why I pinged you both. 

The tests are valid, but I believe they reveal a syntax feature we might want to fix in the spec.

anba: So basically changing "BindingIdentifier[Yield, Await]: [~Yield] yield" to "BindingIdentifier[Yield, Await]: yield" and then add an early error restriction "It is a Syntax Error if this production has a [Yield] parameter."? And apply similar changes for the `await` case. (I haven't checked if this will cause any parsing issues.)

littledan: That's what I was thinking. Does that seem reasonable to you?

Tangential question: Why are BindingIdentifier, IdentifierReference and LabelIdentifier separate productions? Seems like they were all Identifier in ES5.1.

littledan: Any thoughts on this PR? https://github.com/tc39/ecma262/issues/885

isiahmeadows:
I'll note that the snippet would be invalid for other reasons either way. To approach it from the other side, `await 0` is a syntax error in normal functions regardless, so even if the engine accepts the sloppy `let`, it'll see the `await 0` and reject it outside async functions.

Adding the ASI restriction would only change how it's handled in sloppy async functions:

```js
// Currently in sloppy, this would print "Success", but with the
// proposed change, it would print "Fail".
var let;
async function foo() {
  let
  await 0;
}

foo().then(
  () => console.log("Success!"),
  () => console.log("Fail!"))
```

leobalter:
this is fixed through https://github.com/tc39/ecma262/pull/885

Issue #761, opened 2016-09-22
statements/generators/params-trailing-comma-length.js has incorrect generator syntax


JosephPecoraro:
Test: `test/statements/generators/params-trailing-comma-length.js` has incorrect syntax:

``` javascript
function one*(a,) {}
function two*(a,b,) {}
```

It should be:

``` javascript
function* one(a,) {}
function* two(a,b,) {}
```

rwaldron:
@JosephPecoraro please sign the CLA: http://tc39.github.io/test262-cla/ and ping back here when that's complete. Thanks!!

leobalter:
This is fixed on #762

PR #958, opened 2017-04-10
Correct test/built-ins/Function/prototype/bind/15.3.4.5-15-5.js

### LINKED ISSUE #957: 15.3.4.5-15-5.js does not appear to make sense
Test: test/built-ins/Function/prototype/bind/15.3.4.5-15-5.js

The test at a78650fde is currently:

```js
/*---
es5id: 15.3.4.5-15-5
description: >
    Function.prototype.bind - The [[Configurable]] attribute of length
    property in F set as false
---*/

        var canConfigurable = false;
        var hasProperty = false;
        function foo() { }
        var obj = foo.bind({});
        hasProperty = obj.hasOwnProperty("length");
        delete obj.caller;
        canConfigurable = !obj.hasOwnProperty("length");

assert(hasProperty, 'hasProperty !== true');
assert.sameValue(canConfigurable, false, 'canConfigurable');
```

There are a number of pieces that don't seem to make sense:

1. The comment says [[Configurable]] should be false. The spec says [[Configurable]] should be true:
https://tc39.github.io/ecma262/#sec-function.prototype.bind
    > 8. Perform ! DefinePropertyOrThrow(F, "length", PropertyDescriptor {[[Value]]: L, [[Writable]]: false, [[Enumerable]]: false, [[Configurable]]: true}).

2. The test is performing `delete obj.caller`, and not doing anything with the length property. Is caller somehow related to length? I suspect this should be `delete obj.length`.

JosephPecoraro:
The `length` property should be [[Configurable]]
https://tc39.github.io/ecma262/#sec-function.prototype.bind

Also the test was testing deleting the wrong property.

Fixes #957.

### DIFF
--- a/test/built-ins/Function/prototype/bind/15.3.4.5-15-5.js
+++ b/test/built-ins/Function/prototype/bind/15.3.4.5-15-5.js
@@ -5,7 +5,7 @@
 es5id: 15.3.4.5-15-5
 description: >
     Function.prototype.bind - The [[Configurable]] attribute of length
-    property in F set as false
+    property in F set as true
 ---*/
 
         var canConfigurable = false;
@@ -13,8 +13,8 @@ description: >
         function foo() { }
         var obj = foo.bind({});
         hasProperty = obj.hasOwnProperty("length");
-        delete obj.caller;
+        delete obj.length;
         canConfigurable = !obj.hasOwnProperty("length");
 
 assert(hasProperty, 'hasProperty !== true');
-assert.sameValue(canConfigurable, false, 'canConfigurable');
+assert(canConfigurable, 'canConfigurable !== true');


test/built-ins/Function/prototype/bind/15.3.4.5-15-5.js
```diff
@@ -5,16 +5,16 @@
...
+    property in F set as true
 ---*/
 
         var canConfigurable = false;
         var hasProperty = false;
         function foo() { }
         var obj = foo.bind({});
         hasProperty = obj.hasOwnProperty("length");
-        delete obj.caller;
+        delete obj.length;
```
leobalter: omg, why were we deleting obj.caller here? 

Thanks for finding this false positive.

This is also a reminder to convert these legacy tests to use the property helpers.

Btw, a reference for the length descriptor from [the specs](https://tc39.github.io/ecma262/#sec-function.prototype.bind)

```
8. Perform ! DefinePropertyOrThrow(F, "length", PropertyDescriptor {[[Value]]: L, [[Writable]]: false, [[Enumerable]]: false, [[Configurable]]: true}).
```

leobalter:
@JosephPecoraro can you sign the CLA, please? https://tc39.github.io/test262-cla/

If you already work for a org w/ a membership to ECMA, this is not necessary. Just confirm and I'll be able to merge this.

test/built-ins/Function/prototype/bind/15.3.4.5-15-5.js
rwaldron: @leobalter filed https://github.com/tc39/test262/issues/960

JosephPecoraro:
I had signed the agreement in the past for (https://github.com/tc39/test262/issues/761). You can clean it up and land.

PR #959, opened 2017-04-10
Remove invalid async-flag from syntax error tests

anba:


src/params/syntax/async-gen-named-func-expr.template
```diff
@@ -13,7 +13,6 @@ info: |
         7. Let closure be ! AsyncGeneratorFunctionCreate(Normal, FormalParameters,
            AsyncGeneratorBody, funcEnv, strict).
         [...]
-flags: [async]
```
bakkot: This one doesn't look like a syntax error test to me.

anba: It's currently only used for syntax error tests. The real issue is that the template generates tests with the "async" flag, but doesn't call $DONE.

leobalter: all the respective cases using templates from the `syntax` folder have the negative flag, i.e.:

```yaml
negative:
  phase: early
  type: SyntaxError
```

leobalter:
@anba if you don't mind, we try to split commits from the actual changes to the generated codes. This helps identifying the manual from generated changes.

I'm merging this as is. The changes here are small and not worth holding this PR open.

Thank you so much for finding these nit picks.

anba:
Okay, I'll try to remember splitting the commits next time.

PR #961, opened 2017-04-10
Update Async-from-Sync Iterator yield* tests

caitp:
Ordering of done and value property loads changed in
https://github.com/tc39/proposal-async-iteration/commit/395b2e3b2f5acb62f9fae11c5e189423d4af50e6


leobalter:
Thanks for the update!

Issue #960, opened 2017-04-10
Review of test files for cases that should be using propertyHelper.js


rwaldron:
I did some rudimentary grepping and found the following files, all of which contain some property test, but are not using propertyHelper.js

test/built-ins/Array/prototype/concat/15.4.4.4-5-b-iii-3-b-1.js
test/built-ins/Array/prototype/concat/15.4.4.4-5-c-i-1.js
test/built-ins/Array/prototype/splice/15.4.4.12-9-c-ii-1.js
test/built-ins/Function/prototype/bind/15.3.4.5-15-4.js
test/built-ins/Function/prototype/bind/15.3.4.5-15-5.js
test/built-ins/Object/create/15.2.3.5-4-206.js
test/built-ins/Object/create/15.2.3.5-4-212.js
test/built-ins/Object/create/15.2.3.5-4-213.js
test/built-ins/Object/create/15.2.3.5-4-215.js
test/built-ins/Object/create/15.2.3.5-4-216.js
test/built-ins/Object/create/15.2.3.5-4-217.js
test/built-ins/Object/create/15.2.3.5-4-218.js
test/built-ins/Object/create/15.2.3.5-4-219.js
test/built-ins/Object/create/15.2.3.5-4-220.js
test/built-ins/Object/create/15.2.3.5-4-221.js
test/built-ins/Object/create/15.2.3.5-4-222.js
test/built-ins/Object/create/15.2.3.5-4-223.js
test/built-ins/Object/create/15.2.3.5-4-224.js
test/built-ins/Object/create/15.2.3.5-4-225.js
test/built-ins/Object/create/15.2.3.5-4-226.js
test/built-ins/Object/create/15.2.3.5-4-228.js
test/built-ins/Object/create/15.2.3.5-4-229.js
test/built-ins/Object/create/15.2.3.5-4-230.js
test/built-ins/Object/create/15.2.3.5-4-266.js
test/built-ins/Object/create/15.2.3.5-4-268.js
test/built-ins/Object/create/15.2.3.5-4-269.js
test/built-ins/Object/create/15.2.3.5-4-270.js
test/built-ins/Object/create/15.2.3.5-4-271.js
test/built-ins/Object/create/15.2.3.5-4-272.js
test/built-ins/Object/create/15.2.3.5-4-273.js
test/built-ins/Object/create/15.2.3.5-4-274.js
test/built-ins/Object/create/15.2.3.5-4-275.js
[...]

PR #963, opened 2017-04-11
Cleanup prop desc tests in the Array folder

leobalter:
This is a partial fix for #960, actually it's only a kickstart.


leobalter:
cc @rwaldron

rwaldron:
This is excellent. Over the course of this year, we should try to update all built-ins.

PR #965, opened 2017-04-11
Update Travis to check for changes after regenerating tests

leobalter:


leobalter:
Thanks, @tcare!

jugglinmike:
Unless I am mistaken, this change was made with the expectation that
contributors manually generate tests and submit them as part of the review
process. This is not what I had in mind with gh-546. There, I proposed
deployment logic (which has since been moved to [a separate
script](https://github.com/tc39/test262/blob/1c1935e3639d639fb1507262d4703f755a42bd3b/tools/scripts/deploy.sh))
that would result in the automated introduction of generated files. Note that
the current contribution guidelines read:

> Tests expressed with this convention are built automatically following the
> source files' acceptance into the project. Patches should **not** include
> assets built from these sources.

So if we want to continue to require contributors to submit
procedurally-generated files, we'll need to change that text (in addition to
removing the deployment logic/key).

PR #966, opened 2017-04-11
Improve readability for ci scripts

leobalter:


leobalter:
@tcare, @gnarf the `deploy` tag accepts a single value or an array, instead of changing it back to `after_success`  I split the scripts into separate files. Let's see how it goes in Travis

leobalter:
it works for the PR, I need to check the stuff on master as well.

PR #967, opened 2017-04-11
Prohibit async methods named 'function'

littledan:
Addresses https://github.com/tc39/ecma262/issues/832


littledan:
We decided to not go with this spec change. At this point, we should go back and add tests to ensure that these async methods *are* allowed.

leobalter:
@littledan do you want to reuse this PR for the new tests? Otherwise, let's close this.

PR #968, opened 2017-04-11
Tests for PR to Prohibit ASI between 'let' and 'yield'/'await'

littledan:
Tests for the PR https://github.com/tc39/ecma262/pull/885


littledan:
These tests pass in V8, JSC, ChakraCore and SpiderMonkey.

littledan:
The spec change landed; should we land the tests?

PR #969, opened 2017-04-12
Tests for proposed Annex B semantics

littledan:


harness/propertyHelper.js
```diff
@@ -1,13 +1,17 @@
...
         delete obj[name];
     } catch (e) {
         if (!(e instanceof TypeError)) {
             $ERROR("Expected TypeError, got " + e);
         }
     }
-    return !Object.prototype.hasOwnProperty.call(obj, name);
+    var result = !Object.prototype.hasOwnProperty.call(obj, name);
+    // Restore property (no-op if not deleted)
+    Object.defineProperty(obj, name, descriptor);
```
leobalter: We've already discussed this before in other issues.

I don't think we should re-add the property. Once it's used, it should be consumed in the sense of atomic unit tests. To test anything in top of that, another set of object/properties would be recommended instead.

We can't avoid the side effect a configurable property will be deleted. This operation is not only restoring it, but triggering a `[[DefineOwnProperty]]` operation and this might cause other side effects in other test files. 

If I'm not wrong, @bterlson was in those other discussions about this.

src/annex-b-fns/global-existing-global-init.case
```diff
@@ -0,0 +1,45 @@
...
+
+//- setup
+Object.defineProperty(fnGlobalObject(), 'f', {
+  value: 'x',
+  enumerable: true,
+  writable: true,
+  configurable: false
+});
+
+$.evalScript(`
```
leobalter: `$262.evalScript`, but I wonder if we can't just use `eval`

littledan: Oh oops, will fix.

We need to be eval'ing another script; this has different behavior. I'm getting at the global case in this test, not the eval behavior. test262 is actually missing coverage of other properties of Annex B 3.3 extensions to GlobalDeclarationInstantiation.

harness/propertyHelper.js
littledan: OK, I guess that could come up if you are using a Proxy that there are other side effects, but for ordinary objects, there should be no side effect to deleting a property and re-adding it the same way.

Anyway, sure, we can leave a side effect in this one, but in that case, could we put something else in propertyHelper.js to test for a non-configurable property and not mutate the object? This would be useful to allow for the multiple observations of the same property on the same object (as in the tests in the following patch). Note that we can still not be sure that it won't be side-effecting in the case of Proxies, however.

src/annex-b-fns/global-existing-global-init.case
leobalter: thanks for the clarification.

harness/propertyHelper.js
leobalter: while it might not leave side effects beyond Proxies, it's beyond the responsibility for verifyConfigurable restore a state, as an extra step. Tests should follow the minimum possible steps to confirm an assertion. The best of all options would be removing the delete operation, but that's necessary as we already found parts where the descriptor was not matching the actual configurability of a property.

If the property is sent to be tested it should be consumed and that's it.

There's also the problem with non extensible objects:

```js 
Object.preventExtensions(Object);
delete Object.length;
Object.defineProperty(Object, 'length', { value: 42 }); // will throw
```

littledan: Oh right, you can delete properties from non-extensible objects. Sorry for my confusion.

Well, would you object to adding a function which just does the test and doesn't do the delete? I'd like to use a factored-out verifyConfiguable, and adding the property back in my test feels like a layering violation (who would think that a function with that name would delete a property? It feels like depending on an implementation detail).

leobalter: > Well, would you object to adding a function which just does the test and doesn't do the delete?

I agree with you this would be the proper approach. I really want to have some that only checks for `Object.getOwnPropertyDescriptor(obj, 'prop').configurable;`, but that's when we got the history of finding false positives from runtimes.

I have plans to create an easier to consume harness api for properties, where I would verify it like:

```js
verifyProperty(obj, 'prop', {
  configurable: false,
  enumerable: true,
  writable: false
});
```

So the tests won't need to worry about the order. e.g. `verifyWritable` being called after `verifyConfigurable`.

Using this same function, we can extend it with options, let's say:

```js
verifyProperty(obj, 'prop', {
  configurable: false,
  enumerable: true,
  writable: false
}, {
  descriptorOnly: true // avoid changing the actual values or deleting properties
});
```

or even: 

```js
verifyProperty(obj, 'prop', {
  configurable: false,
  enumerable: true,
  writable: false,
  __descriptorOnly: true
});
```

The function signature is only a bikeshed, but a new PR for it might be a good starting point to discuss the options.

Does it work for you?

littledan: No, because in this test I check property attributes of the same property multiple times from different places.

leobalter:
I was trying to find side effects from the new verifyConfigurable api and found this file only failing with the new approach. 

```
test/built-ins/TypedArrays/internals/DefineOwnProperty/non-extensible-redefine-key.js
```

src/annex-b-fns/global-existing-global-init.case
```diff
@@ -0,0 +1,45 @@
...
+assert.sameValue(f, 'x');
+verifyEnumerable(fnGlobalObject(), 'f');
+verifyWritable(fnGlobalObject(), 'f');
+verifyNotConfigurable(fnGlobalObject(), 'f');
+`)
```
rwaldron: Nit: `;`

src/annex-b-fns/global-existing-global-init.case
```diff
@@ -0,0 +1,45 @@
...
+return 'inner declaration';
+//- teardown
+`);
+
+$.evalScript(`
+assert.sameValue(f(), 'inner declaration');
+verifyEnumerable(fnGlobalObject(), 'f');
+verifyWritable(fnGlobalObject(), 'f');
+verifyNotConfigurable(fnGlobalObject(), 'f');
+`)
```
rwaldron: Nit: `;`

src/annex-b-fns/global-existing-non-enumerable-global-init.case
```diff
@@ -0,0 +1,45 @@
...
+  writable: true,
+  configurable: true
+});
+
+$.evalScript(`
+assert.sameValue(f, 'x');
+verifyNotEnumerable(fnGlobalObject(), 'f');
+verifyWritable(fnGlobalObject(), 'f');
+verifyConfigurable(fnGlobalObject(), 'f');
+`)
```
rwaldron: ;)

test/annexB/language/eval-code/direct/global-block-decl-eval-global-existing-non-enumerable-global-init.js
```diff
@@ -0,0 +1,38 @@
...
+});
+
+eval(
+  'var global = fnGlobalObject();\
+  assert.sameValue(f, "x", "binding is not reinitialized");\
+  \
+  verifyNotEnumerable(global, "f");\
+  verifyWritable(global, "f");\
+  verifyConfigurable(global, "f");{ function f() {  } }'
+);
```
rwaldron: For new multi-line strings, I recommend that we stick to template strings (I also recognize that this came from existing code)

test/annexB/language/eval-code/direct/global-if-decl-else-decl-a-eval-global-existing-non-enumerable-global-init.js
```diff
@@ -0,0 +1,47 @@
...
+  configurable: true
+});
+
+eval(
+  'var global = fnGlobalObject();\
+  assert.sameValue(f, "x", "binding is not reinitialized");\
+  \
+  verifyNotEnumerable(global, "f");\
+  verifyWritable(global, "f");\
+  verifyConfigurable(global, "f");if (true) function f() {  } else function _f() {}'
```
rwaldron: Same as above?

test/annexB/language/eval-code/direct/global-if-decl-else-decl-b-eval-global-existing-non-enumerable-global-init.js
```diff
@@ -0,0 +1,47 @@
...
+});
+
+eval(
+  'var global = fnGlobalObject();\
+  assert.sameValue(f, "x", "binding is not reinitialized");\
+  \
+  verifyNotEnumerable(global, "f");\
+  verifyWritable(global, "f");\
+  verifyConfigurable(global, "f");if (false) function _f() {} else function f() {  }'
+);
```
rwaldron: Same as above

test/annexB/language/eval-code/direct/global-if-decl-else-stmt-eval-global-existing-non-enumerable-global-init.js
```diff
@@ -0,0 +1,47 @@
...
+});
+
+eval(
+  'var global = fnGlobalObject();\
+  assert.sameValue(f, "x", "binding is not reinitialized");\
+  \
+  verifyNotEnumerable(global, "f");\
+  verifyWritable(global, "f");\
+  verifyConfigurable(global, "f");if (true) function f() {  } else ;'
+);
```
rwaldron: Same as above

test/annexB/language/eval-code/direct/global-if-decl-no-else-eval-global-existing-non-enumerable-global-init.js
```diff
@@ -0,0 +1,47 @@
...
+});
+
+eval(
+  'var global = fnGlobalObject();\
+  assert.sameValue(f, "x", "binding is not reinitialized");\
+  \
+  verifyNotEnumerable(global, "f");\
+  verifyWritable(global, "f");\
+  verifyConfigurable(global, "f");if (true) function f() {  }'
+);
```
rwaldron: Same as above

harness/propertyHelper.js
leobalter: I changed this to have a `restore` option instead, which is doing exactly what you're doing here, but now in the new API: https://github.com/tc39/test262/pull/979

leobalter:
@littledan do you have any plans to move the normative change further on ecma262? I can help with the fixes here when it's time for an update.

littledan:
Sorry for dropping the ball on the review feedback here. I believe this patch has consensus and is just waiting on fixing up the code review feedback to land.

leobalter:
no problem, I might be able to work on this tomorrow. I'll ping you back then.

littledan:
Thanks for your help!

rwaldron:
@littledan @leobalter sorry for letting this sit for so long. I've followed the chain of issues and PRs, all the back to the notes and verified that this does have consensus, so I will get this landed today.

rwaldron:
Landed

- 89a75a90f6df2bde0a7bee61b66a16f5cd248bee
- 8d7dd538f1aa575312922f9ef49353f0c26287f3

leobalter:
> Sorry for dropping the ball on the review feedback here. I believe this patch has consensus and is just waiting on fixing up the code review feedback to land.

@littledan can you help me pointing to the consensus part? I felt a bit lot trying to find it, maybe in the notes docs?

leobalter:
I found the link to the consensus but it's so weird the notes are not well captured...

```
12.i.k Sloppy function hoisting web reality tweak
(Daniel Ehrenberg)

Explainer
DE: (Presenting Explainer)

DE: Let's defer this.

Conclusion/Resolution
Consensus
```

Issue #950, opened 2017-04-06
[Question] Format for Unicode property escape test suite?


mathiasbynens:
As part of my work on [the Unicode property escapes in regular expressions proposal](https://github.com/tc39/proposal-regexp-unicode-property-escapes), I’m planning to contribute tests for this feature. (Ref. https://github.com/tc39/proposal-regexp-unicode-property-escapes/issues/4.)

[My plan was to create a script (hosted in its own, separate repo)](https://github.com/mathiasbynens/unicode-property-escapes-tests) that generates the tests based on a given Unicode version, so they could easily be updated + upstreamed to test262. (Note that the Unicode standard is updated every year, and thus, so should the tests.) I would be maintaining this script and submitting PRs with the updated tests to test262 every year.

Now, because these tests will be generated programmatically, they can be as exhaustive as we want them to be. We can do better than randomly testing a couple of properties and symbols, which doesn’t guarantee 100% spec compliance. Ideally, every Unicode symbol would be tested against every supported property escape. This would ensure 100% interoperability (assuming all major engines run tests262 tests). The downside would be that these tests would potentially take a long time to run. @syg, @tcare: thoughts?

Here’s an exhaustive list of the supported properties per this proposal (as per the current version): https://github.com/mathiasbynens/regexpu-core/blob/master/property-escapes.md

My questions:

1. Would such a test suite make sense as part of test262?
2. If so, what would the directory/file structure look like? Some options:
    - A separate `*.js` file for each individual property escape?
    - A separate file for every `0xFFFF` code points (i.e. every Unicode plane)?
    - Is there a better idea?

Any feedback is welcome.

cc @leobalter @jugglinmike @littledan

mathiasbynens:
whoops, hit submit too quickly… bear with me as I update the post

mathiasbynens:
As part of my work on [the Unicode property escapes in regular expressions proposal](https://github.com/tc39/proposal-regexp-unicode-property-escapes), I’m planning to contribute tests for this feature. (Ref. https://github.com/tc39/proposal-regexp-unicode-property-escapes/issues/4.)

My plan was to create a script (hosted in its own, separate repo) that generates the tests based on a given Unicode version, so they could easily be updated + upstreamed to test262. (Note that the Unicode standard is updated every year, and thus, so should the tests.) I would be maintaining this script and submitting PRs with the updated tests to test262 every year.

[...]

PR #971, opened 2017-04-12
Add exhaustive tests for RegExp Unicode property escapes

mathiasbynens:
Proposal: https://github.com/tc39/proposal-regexp-unicode-property-escapes

These tests have been generated by the script at https://github.com/mathiasbynens/unicode-property-escapes-tests. They check all the properties and values that should be supported by implementations against the symbols they’re supposed to match. False positives are detected as well.

Ref. #950.
Ref. https://github.com/tc39/proposal-regexp-unicode-property-escapes/issues/4.


anba:
Initial observations:
- There is a bug in the code generation for Noncharacter_Code_Point, it tests 10FFFE..10FFFF are not Noncharacter_Code_Point.
- Code point ranges must not be consecutive when surrogate ranges are involved, e.g. when testing General_Category=Private_Use, the string "\uDBFF\uDC00" is generated (U+DBFF is the last private-use high surrogate and U+DC00 is the first low surrogate).
- The `buildString` function should probably be optimized, otherwise debug builds will spent most of the time assembling the test string. For example when I changed it to this version
```js
const buildString = ({ loneCodePoints, ranges }) => {
    let result = String.fromCodePoint(...loneCodePoints);
    for (const [start, end] of ranges) {
        let codePoints = [];
        for (let i = 0, codePoint = start; codePoint <= end; codePoint++) {
            codePoints[i++] = codePoint;
            if (i === 10000) {
                result += String.fromCodePoint(...codePoints);
                codePoints.length = i = 0;
            }
        }
        result += String.fromCodePoint(...codePoints);
    }
    return result;
};
```
a v8-debug build created the string for ASCII_Hex_Digit in 750ms compared to 9750ms with the current `buildString` function.

mathiasbynens:
@anba Thanks for taking a look!

> There is a bug in the code generation for `Noncharacter_Code_Point`, it tests `10FFFE..10FFFF` [which] are not `Noncharacter_Code_Point`.

Did you mean [here](https://github.com/mathiasbynens/unicode-property-escapes-tests/blob/14729c23ea58b7cbc1ded155019c461c080ab24d/output/Noncharacter_Code_Point.js#L76)? That’s in `nonMatchSymbols`, i.e. the symbols that are supposed to not match — so that’s fine. Am I overlooking something?

I’ll address your other feedback (excellent points/suggestions!) tomorrow — if you don’t beat me to it with a PR, that is.

test/built-ins/RegExp/property-escapes/ASCII.js
```diff
@@ -0,0 +1,45 @@
...
+
+const buildString = ({ loneCodePoints, ranges }) => {
+	let result = String.fromCodePoint(...loneCodePoints);
+	for (const [start, end] of ranges) {
+		for (let codePoint = start; codePoint <= end; codePoint++) {
+			result += String.fromCodePoint(codePoint);
+		}
+	}
+	return result;
+};
```
leobalter: if buildString is the same among all tests, we should add a new js file in the `/harness/` folder and get it through the includes tag. i.e.:

```yaml
esid: sec-static-semantics-unicodematchproperty-p
features: [regexp-unicode-property-escapes]
includes: [buildString.js]
---*/
```

anba:
> Did you mean here? That’s in nonMatchSymbols, i.e. the symbols that are supposed to not match — so that’s fine. Am I overlooking something?

PropList.txt has:
> 10FFFE..10FFFF; Noncharacter_Code_Point # Cn   [2] \<noncharacter-10FFFE>..\<noncharacter-10FFFF>

So U+10FFFE and U+10FFFF should match `\p{Noncharacter_Code_Point}` instead of `\P{Noncharacter_Code_Point}`, shouldn't it?

littledan:
These tests are great, and they seem to test the feature in a pretty thorough way. A couple thoughts:
- The tests are somewhat slow. These tests took 5m30s to run on my machine for V8, whereas the rest of the test262 tests took 15m39s. That makes these tests not the slowest thing ever, but it could still be nice to cut down on them a little bit (and I wouldn't mind if more thought went into use of the test generation tool to ensure that the matrix of combinations really does add value...).
- Nit: "matches all proper symbols" is a funny error message. At least, it could read "failed to match certain symbols". The best would be if, in the failure case, you could go through the code points and find the errant one.
- FWIW V8 passes most (97%) of these if you turn the feature flag on, but fails some of these tests by disagreeing on the set of characters (different Unicode version?), and others because they are missing as property names: Emoji_Component, Prepended_Concatenation_Mark, and Composition_Exclusion. This is based on looking through all the properties in ICU 58; we're upgrading to 59, and I should be able to re-test afterwards. The first of these properties is new, but the second two are not. cc @hashseed @schuay @jungshik

leobalter:
> Nit: "matches all proper symbols" is a funny error message. At least, it could read "failed to match certain symbols". The best would be if, in the failure case, you could go through the code points and find the errant one.

I'm accountable for this message. We avoid to use messages as failure messages. They are actually a message of what we expect, a description of the assertion. 

While we want to improve "matches all proper symbols", it should be using some better description of each assertion. We also agree it should be a different message for each part so it helps finding the errant point.

test/built-ins/RegExp/property-escapes/ASCII.js
mathiasbynens: Done.

mathiasbynens:
> So U+10FFFE and U+10FFFF should match `\p{Noncharacter_Code_Point}` instead of `\P{Noncharacter_Code_Point}`, shouldn't it?

@anba You’re absolutely right! Fixed.

> The tests are somewhat slow. These tests took 5m30s to run on my machine for V8, whereas the rest of the test262 tests took 15m39s. That makes these tests not the slowest thing ever, but it could still be nice to cut down on them a little bit (and I wouldn't mind if more thought went into use of the test generation tool to ensure that the matrix of combinations really does add value...).

@littledan I wonder if they’re any faster now that `buildString` has been improved to use chunking.

> We also agree it should be a different message for each part so it helps finding the errant point.

@littledan & @leobalter The string currently isn’t checked in parts — it’s all in one go.

1. We could test the string in parts, i.e. create a separate string per `X` code points and test it, but that would also be slower than just checking once and would still not produce code-point-specific error messages.
2. We could test each code point individually, with code-point-specific error messages, but that would likely be very slow.

I’m tempted to leave it as-is: let’s optimize for the common case, where engines run these tests at build time long after implementing support for this feature. Nicer, more specific error messages only help engineers that are currently implementing the feature, but do so at a performance cost for everyone.

littledan:
@mathiasbynens They took 1m32s with the new version on my machine. A huge improvement, but that's still 10% as much time as the rest of test262--I'd say this is a bit borderline, and I'd be interested in input from others.

littledan:
For code point specific error messages, just a thought, you could search through the individual code points when one of the bigger strings finds a mismatch.

hashseed:
> because they are missing as property names: Emoji_Component, Prepended_Concatenation_Mark, and Composition_Exclusion

ICU 59 doesn't seem to include these properties either.

leobalter:
> I’m tempted to leave it as-is: let’s optimize for the common case, where engines run these tests at build time long after implementing support for this feature.

Yes, I can help with that in the source as a follow up.

> Nicer, more specific error messages only help engineers that are currently implementing the feature, but do so at a performance cost for everyone.

I believe we can fix it just by having less generic strings. I'll follow up with a PR in the source as well.

mathiasbynens:
https://github.com/mathiasbynens/unicode-property-escapes-tests/pull/6

leobalter:
Running on V8:

```
[~/dev/test262] (mathiasbynens-add-unicode-property-escape-tests)
$ time test262-harness --hostPath=`which d8` --hostType=d8 --hostArgs='--harmony_regexp_property' $(git diff --cached tc39/master --name-only --diff-filter=AM test/) -t 4
FAIL test/built-ins/RegExp/property-escapes/Assigned.js
  `\P{Assigned}` should match U+00DFFF (`���`)

FAIL test/built-ins/RegExp/property-escapes/Assigned.js
  `\P{Assigned}` should match U+00DFFF (`���`)

FAIL test/built-ins/RegExp/property-escapes/Composition_Exclusion.js
  Expected no error, got SyntaxError: Invalid regular expression: /^\p{Composition_Exclusion}+$/: Invalid property name

FAIL test/built-ins/RegExp/property-escapes/Composition_Exclusion.js
  Expected no error, got SyntaxError: Invalid regular expression: /^\p{Composition_Exclusion}+$/: Invalid property name

FAIL test/built-ins/RegExp/property-escapes/Emoji.js
  `\p{Emoji}` should match U+01F9E6 (`🧦`)

FAIL test/built-ins/RegExp/property-escapes/Emoji.js
  `\p{Emoji}` should match U+01F9E6 (`🧦`)

FAIL test/built-ins/RegExp/property-escapes/Emoji_Component.js
  Expected no error, got SyntaxError: Invalid regular expression: /^\p{Emoji_Component}+$/: Invalid property name

FAIL test/built-ins/RegExp/property-escapes/Emoji_Component.js
  Expected no error, got SyntaxError: Invalid regular expression: /^\p{Emoji_Component}+$/: Invalid property name

FAIL test/built-ins/RegExp/property-escapes/Emoji_Presentation.js
  `\p{Emoji_Presentation}` should match U+01F9E6 (`🧦`)

FAIL test/built-ins/RegExp/property-escapes/Emoji_Presentation.js
  `\p{Emoji_Presentation}` should match U+01F9E6 (`🧦`)

FAIL test/built-ins/RegExp/property-escapes/Emoji_Modifier_Base.js
  `\P{Emoji_Modifier_Base}` should match U+01F93C (`🤼`)

FAIL test/built-ins/RegExp/property-escapes/Emoji_Modifier_Base.js
  `\P{Emoji_Modifier_Base}` should match U+01F93C (`🤼`)

FAIL test/built-ins/RegExp/property-escapes/General_Category_-_Other.js
  `\P{C}` should match U+00DFFF (`���`)

FAIL test/built-ins/RegExp/property-escapes/General_Category_-_Other.js
  `\P{C}` should match U+00DFFF (`���`)

FAIL test/built-ins/RegExp/property-escapes/General_Category_-_Surrogate.js
  `\P{Cs}` should match U+00DFFF (`���`)

FAIL test/built-ins/RegExp/property-escapes/General_Category_-_Surrogate.js
  `\P{Cs}` should match U+00DFFF (`���`)

FAIL test/built-ins/RegExp/property-escapes/Prepended_Concatenation_Mark.js
  Expected no error, got SyntaxError: Invalid regular expression: /^\p{Prepended_Concatenation_Mark}+$/: Invalid property name

FAIL test/built-ins/RegExp/property-escapes/Prepended_Concatenation_Mark.js
  Expected no error, got SyntaxError: Invalid regular expression: /^\p{Prepended_Concatenation_Mark}+$/: Invalid property name

FAIL test/built-ins/RegExp/property-escapes/Script_Extensions_-_Inherited.js
  `\p{scx=Qaai}` should match U+00309A (`゚`)

FAIL test/built-ins/RegExp/property-escapes/Script_Extensions_-_Inherited.js
  `\p{scx=Qaai}` should match U+00309A (`゚`)

Ran 732 tests
712 passed
20 failed

real	1m4.783s
user	2m30.454s
sys	0m53.827s
```

I'm compiling SpiderMonkey to check these tests as well and the running time.

anba:
Assigned.js, General\_Category\_-\_Other.js, and General\_Category\_-\_Surrogate.js:
Has [0x00DC00, 0x00DFFF] in matchSymbols and nonMatchSymbols.

anba:
> I'm compiling SpiderMonkey to check these tests as well and the running time.

SpiderMonkey doesn't yet support Unicode property escapes.

leobalter:
> SpiderMonkey doesn't yet support Unicode property escapes.

Thanks for letting me know. I needed to build a new version of SM anyway. :)

mathiasbynens:
Looking into the unexpected failures in V8. (This is a problem with the tests — not V8’s implementation!)

mathiasbynens:
Fix for those three broken tests: #974

PR #972, opened 2017-04-12
Test for change to cache templates by site, not contents

littledan:
These tests are against a specification change based on discussion in
https://github.com/tc39/ecma262/issues/840

The tests here passed on SpiderMonkey but failed on other
implementations, which implement the current specification.

EDIT: They passed on a version of SpiderMonkey from February, and would likely fail on ToT.

### DIFF
--- a/test/language/expressions/tagged-template/cache-differing-expressions-eval.js
+++ b/test/language/expressions/tagged-template/cache-differing-expressions-eval.js
@@ -3,11 +3,9 @@
 /*---
 es6id: 12.2.8
 description: Template caching using distinct expressions within `eval`
-info: |
-    Previously-created template objects should be retrieved from the internal
-    template registry when their source is identical but their expressions
-    evaluate to different values and the tagged template is being evaluated in
-    an `eval` context.
+info: >
+    Templates are cached by source location; different locations will
+    have different template objects.
 ---*/
 function tag(templateObject) {
   previousObject = templateObject;
@@ -23,4 +21,4 @@ assert(firstObject !== null);
 previousObject = null;
 
 eval('tag`head${b}tail`');
-assert.sameValue(previousObject, firstObject);
+assert.notSameValue(previousObject, firstObject);
--- a/test/language/expressions/tagged-template/cache-differing-expressions-eval.js
+++ b/test/language/expressions/tagged-template/cache-differing-expressions-eval.js
@@ -1,11 +1,12 @@
 // Copyright (C) 2014 the V8 project authors. All rights reserved.
 // This code is governed by the BSD license found in the LICENSE file.
 /*---
-es6id: 12.2.8
+esid: sec-gettemplateobject
 description: Template caching using distinct expressions within `eval`
 info: >
-    Templates are cached by source location; different locations will
-    have different template objects.
+    1. For each element _e_ of _templateRegistry_, do
+      1. If _e_.[[Site]] is the same Parse Node as _templateLiteral_, then
+        1. Return _e_.[[Array]].
 ---*/
 function tag(templateObject) {
   previousObject = templateObject;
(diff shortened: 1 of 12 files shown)


test/language/expressions/tagged-template/cache-same-site.js
```diff
@@ -0,0 +1,36 @@
...
+var previousObject = null;
+
+function runTemplate() {
+  tag`head${a}tail`;
+}
+
+runTemplate();
+firstObject = previousObject;
+
+assert(firstObject !== null);
```
leobalter: assert(firstObject); should be fine, but you can actually get the proper value to test here, right?

```js
assert.sameValue(firstObject.length, 2);
assert.sameValue(firstObject[0], 'head');
assert.sameValue(firstObject[1], 'tail');
```

test/language/expressions/tagged-template/cache-differing-expressions-eval.js
```diff
@@ -4,10 +4,8 @@
 es6id: 12.2.8
```
leobalter: this behavior is not valid anymore for es6, these tags should be replaced by respective `esid` tags as in `esid: sec-gettemplateobject`

test/language/expressions/tagged-template/cache-same-site-top-level.js
```diff
@@ -0,0 +1,31 @@
...
+let a = 1;
+for (let i = 0; i < 2; i++) {
+  tag`head${a}tail`;
+}
+
+assert.sameValue(templates.length, 2);
+
+assert.sameValue(
+  templates[0],
+  templates[1],
```
caitp: I mentioned this on the normative issue, but I'm wondering how this is useful at all, for any reason.

I don't think we actually get any benefit at all from caching template objects, and this change would make it even less useful. I think we should just get rid of the caching entirely.

leobalter: It's good to have the perspective from you, @caitp. It feels weird or wrong that we can observe we cache these references.

littledan: JavaScript is a language that's based on objects with identity. If the system does any kind of caching of JS objects, it'll be observable exactly when it happens, for better or worse.

leobalter: I don't see this being any problem for me as a web developer. It looks like a curious fact we can find the same reference.

I'm neutral to this change, if it helps runtimes, I'm +1 for it.

littledan: For motivation, see https://github.com/tc39/ecma262/pull/890#issuecomment-293777826 and https://github.com/tc39/ecma262/pull/890#issuecomment-293924695

littledan:
I will need to add a test for multiple parse sites that are in distinct functions, e.g., multiple closures with the same literal syntax, to test that they are cached the same. (In V8 terms, this means the cache lives on the SharedFunctionInfo, not the JSFunction.)

test/language/expressions/tagged-template/cache-same-site.js
littledan: I was going to say, "this is covered by other tests". But actually I can't find any other tests that cover it. Either way, I think validating what the template actually is is a subject for another, focused test. The assertion here is just to validate that the tag() function ran.

test/language/expressions/tagged-template/cache-differing-expressions-eval.js
littledan: Done

leobalter:
@littledan this is here for a while. While I should give it a full review and resolve the merging conflicts, would you mind to review it as well?

leobalter:
seems good to land for now. We can work on a follow up if necessary.

littledan:
Sorry, I should've been more clear--the spec patch hasn't landed yet. I hope it should land soon!

leobalter:
I think it's fine to have the tests considering we reached consensus in the respective PR? Similar to Stage 3 proposals, we have at least a reference. Otherwise I can revert it if necessary.

littledan:
@leobalter Good analogy, that seems like a reasonable policy to me.

PR #973, opened 2017-04-13
Unify indentation and add .editorconfig

mathiasbynens:
Many different coding styles are being used across the various non-test files in this repository.

* For Python files, four-space indentation (per PEP 8) seemed to be the most common. However, some files were using two-space. I’ve changed them to use four spaces and codified this preference in `.editorconfig`.
* For JavaScript files, most of the tests, as well as most non-test files seemed to be using two-space indents. However, some files were using four-space indents, five-space indents (!), and a single file was using tabs. I’ve changed them to use two spaces and codified this preference in `.editorconfig`.

While going through the files, I noticed something odd. The indentation seems to imply that a brace is missing here: https://github.com/tc39/test262/blob/96aa8c77b33ef734d486fece1cf00fb373785e2c/harness/Date_constants.js#L46-L50 This patch adds that brace — let me know if it shouldn’t be there, in which case I’ll fix the indentation.

Diff sans whitespace changes: https://github.com/tc39/test262/pull/973/files?w=1

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,16 @@
+root = true
+
+[*]
+charset = utf-8
+indent_style = tab
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[{README.md,package.json,.travis.yml,*.sh,*.js}]
+indent_style = space
+indent_size = 2
+
+[{.jshintrc,*.py}]
+indent_style = space
+indent_size = 4
--- a/harness/Date_constants.js
+++ b/harness/Date_constants.js
@@ -35,49 +35,50 @@ var $LocalTZ,
     $DST_end_minutes;
 
 (function () {
-    /**
-      * Finds the first date, starting from |start|, where |predicate|
-      * holds.
-      */
-    var findNearestDateBefore = function(start, predicate) {
-        var current = start;
-        var month = 1000 * 60 * 60 * 24 * 30;
-        for (var step = month; step > 0; step = Math.floor(step / 3)) {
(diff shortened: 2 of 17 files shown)


leobalter:
This is great!

For more information, I asked @mathiasbynens to help with this and I believe applying a few whitespace style rules will be great.

anba:
> While going through the files, I noticed something odd. The indentation seems to imply that a brace is missing here: https://github.com/tc39/test262/blob/96aa8c77b33ef734d486fece1cf00fb373785e2c/harness/Date_constants.js#L46-L50 This patch adds that brace — let me know if it shouldn’t be there, in which case I’ll fix the indentation.

The brace was added at the wrong location, which results in an iloop when executing the file. But instead of fixing this, let's just remove everything from the file which isn't actually needed for test262: https://github.com/tc39/test262/blob/44b40e083ec851cd8448d5fe4755bdfc9f0dc14c/harness/Date_constants.js#L23-L84

mathiasbynens:
@anba Good catch. Could you please submit a PR for that change?

.editorconfig
```diff
@@ -0,0 +1,16 @@
...
+[*]
+charset = utf-8
+indent_style = tab
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[{README.md,package.json,.travis.yml,*.sh,*.js}]
+indent_style = space
+indent_size = 2
```
rwaldron: ❤️

rwaldron:
Makes me so happy.

PR #979, opened 2017-04-15
Add Verify Property harness

leobalter:
signature:

```js
verifyProperty(
  Object obj,
  String|Symbol name,
  Object|undefined {
    Boolean writable,
    Boolean configurable,
    Boolean enumerable
  },
  Object { Boolean restore }
);
```

You can set an expliecit `restore` option if you need to reuse the same property, works well for accessor descriptors too.

I've had to add a monkey patch to `$ERROR` in tests asserting negative effects, it is ugly, but it works.

cc @littledan;

### DIFF
--- a/harness/propertyHelper.js
+++ b/harness/propertyHelper.js
@@ -1,4 +1,74 @@
 
+function verifyProperty(obj, name, desc, options) {
+  assert(
+    arguments.length > 2,
+    'verifyProperty should receive at least 3 arguments: obj, name, and descriptor'
+  );
+
+  var originalDesc = Object.getOwnPropertyDescriptor(obj, name);
+  var nameStr = String(name);
+
+  // Allows checking for undefined descriptor if it's explicitly given.
+  if (desc === undefined) {
+    assert.sameValue(
+      originalDesc,
+      undefined,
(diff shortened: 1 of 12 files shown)


harness/propertyHelper.js
```diff
@@ -11,7 +81,7 @@ function isConfigurable(obj, name) {
 }
 
 function isEnumerable(obj, name) {
-  var stringCheck;
+  var stringCheck = false;
```
leobalter: this undefined value got me into a lot of trouble while writing the tests.

littledan:
Seems good to me. I could imagine more features, like `value` in the options. I also don't know the whole motivation for this approach as opposed to adding restore versions of the individual test functions. But this should be useful for my test.

leobalter:
> Seems good to me. I could imagine more features, like value in the options. 

this is for sure one of the goals. I had this in mind when I created this function and I'm planning to follow - in another PR - with support to verify values for data and accessor descriptors.

> I also don't know the whole motivation for this approach as opposed to adding restore versions of the individual test functions. But this should be useful for my test.

This solves a lot of other problems we've had with the other individual problems. The calling order for each function was one of the most annoying.

PR #981, opened 2017-04-15
RegExp property escapes: list emoji data version (UTR51)

mathiasbynens:
This is just as important as the Unicode database version.


rwaldron:
I'm not sure if it matters, but Ecma-262 doesn't have a normative reference for Emoji v5.0, but it does for Unicode Latest (presently 9.0.0). Maybe that point is irrelevant? @bterlson what do you know about it?

mathiasbynens:
That reference would be added as part of the proposal: https://tc39.github.io/proposal-regexp-unicode-property-escapes/#sec-bibliography

rwaldron:
> That reference would be added as part of the proposal:

Ah-ha! Question answered :)

PR #983, opened 2017-04-16
Reference `requirements.txt` file in CI

jugglinmike:
Installing dependencies using this file increases parity between the CI
and development environment. Because the file explicitly specifies a
version of the package to be installed, this approach also produces more
predictable results overall.


rwaldron:
LGTM

PR #987, opened 2017-04-18
Optimize decodeURI/encodeURI tests

anba:
These tests take a long time to complete in non-release builds, is almost like they were written to be slow. 😛 

I've prepared eight commits to clean-up and optimize the tests:
- 9e1cf7b: Only moves the decimalToHexString function into a new harness file to avoid duplicating it in multiple files
- 9efab94: Optimizes decimalToHexString and adds support to format numbers greater than 0xffff (needed for the next part)
- bcfad84: Removes the remaining, slightly different decimalToHexString functions from test files
- 09c0393: decimalToHexString was often called to format numbers in [0, 0xff], but as decimalToHexString always returns a zero-padded string, substring() was used to adjust the return value. Add decimalToHex2String to avoid these substring() calls.
- af1bf56: We can further improve 09c0393, because the two-digit hex-string is always needed as a percent string, so let's move the percent-sign concatenation into the helper function.

Further polishing:
- 29977e7: Replace two calls to String.fromCharCode with a single call
- a1dd4ff: Hoist string concatenations outside of loop bodies
- 1baa38b: It doesn't even make sense to handle Test262Error in the catch-handlers.

We could definitely perform more clean-up for these tests, e.g. remove the error collection code (which doesn't actually work because the first call to $ERROR will throw a Test262Error), but these changes should help implementers to run the test in non-release builds.

### DIFF
--- a/harness/decimalToHexString.js
+++ b/harness/decimalToHexString.js
@@ -14,3 +14,8 @@ function decimalToHexString(n) {
   }
   return s;
 }
+
+function decimalToHex2String(n) {
+  var hex = "0123456789ABCDEF";
+  return hex[(n >> 4) & 0xf] + hex[n & 0xf];
+}
--- a/harness/decimalToHexString.js
+++ b/harness/decimalToHexString.js
@@ -2,25 +2,15 @@
 // This code is governed by the BSD license found in the LICENSE file.
 
 function decimalToHexString(n) {
-  n = Number(n);
-  var h = "";
-  for (var i = 3; i >= 0; i--) {
-    if (n >= Math.pow(16, i)) {
-      var t = Math.floor(n / Math.pow(16, i));
-      n -= t * Math.pow(16, i);
-      if ( t >= 10 ) {
-        if ( t == 10 ) { h += "A"; }
-        if ( t == 11 ) { h += "B"; }
-        if ( t == 12 ) { h += "C"; }
-        if ( t == 13 ) { h += "D"; }
-        if ( t == 14 ) { h += "E"; }
-        if ( t == 15 ) { h += "F"; }
-      } else {
-        h += String(t);
-      }
-    } else {
-      h += "0";
-    }
+  var hex = "0123456789ABCDEF";
+  n >>>= 0;
+  var s = "";
+  while (n) {
+    s = hex[n & 0xf] + s;
+    n >>>= 4;
   }
-  return h;
+  while (s.length < 4) {
+    s = "0" + s;
+  }
+  return s;
 }
--- a/harness/decimalToHexString.js
+++ b/harness/decimalToHexString.js
@@ -1,0 +1,26 @@
+// Copyright (C) 2017 André Bargull. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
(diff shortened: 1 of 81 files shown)


harness/decimalToHexString.js
```diff
@@ -0,0 +1,21 @@
+// Copyright (C) 2017 André Bargull. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+function decimalToHexString(n) {
+  var hex = "0123456789ABCDEF";
+  n >>>= 0;
```
leobalter: a "really greater than" operator.

leobalter: This is actually the first time I used this assignment operator. Hang on while I read the specs, mdn, etc. :)

anba: It's basically just performing `n = ToUin32(n)`, but it actually doesn't really matter here. We could also use `n |= 0` to cast it into an signed int32, or even just omit the (u)int32 cast at all, because we _know_ all call sites will supply sensible values.

test/built-ins/decodeURI/S15.1.3.1_A2.2_T1.js
```diff
@@ -15,17 +16,16 @@ var uriReserved = [";", "/", "?", ":", "@", "&", "=", "+", "$", ","];
...
     var hex = String.fromCharCode(index);
     for (var indexC = 0; indexC < uriReserved.length; indexC++) {
       if (hex === uriReserved[indexC]) continue l;        
     } 
     if (hex === "#") continue l;
-    if (decodeURI("%" + hexB1.substring(2)) === hex) continue;
+    if (decodeURI(hexB1) === hex) continue;
   } catch (e) {
-    if (e instanceof Test262Error) throw e;
   }
```
leobalter: I believe we should remove the whole try/catch wrapping. If it throws, let the test fail.

test/built-ins/decodeURI/S15.1.3.1_A2.3_T1.js
```diff
@@ -15,15 +16,14 @@ var indexP;
...
     count++;
-    var hexB2 = decimalToHexString(indexB2);
+    var hexB1_B2 = hexB1 + decimalToPercentHexString(indexB2);
     var index = (indexB1 & 0x1F) * 0x40 + (indexB2 & 0x3F);  
     try {
-      if (decodeURI("%" + hexB1.substring(2) + "%" + hexB2.substring(2)) === String.fromCharCode(index)) continue;
-  } catch (e) {
-      if (e instanceof Test262Error) throw e;
+      if (decodeURI(hexB1_B2) === String.fromCharCode(index)) continue;
+    } catch (e) {
```
leobalter: same for the try/catch wrapping

leobalter:
There are other cases with the try/catch wrappings.

Thanks for the clarification on the `>>>=`, I'm always learning something. I can't quickly see any other issue in this patch.

leobalter:
The improvement can be seen both in spidermonkey and v8 running against the same set of files from this current and the master branch. I didn't check using other runtimes.

leobalter:
That's great, thanks!

Issue #964, opened 2017-04-11
Travis CI timeout waiting for a passphrase.


leobalter:
https://travis-ci.org/tc39/test262/ is failing in a timeout waiting for a passphrase.

I set the key without a passphrase as I expected it would be easier for scripting with Travis CI, it still requests a passphrase in there.

I don't know how to fix this properly.

leobalter:
Example: https://travis-ci.org/tc39/test262/builds/220849297#L189

This is happening during the `ssh-add github-deploy-key`.

edgemaster:
`ssh-add` seems to request a passphrase also if the provided keyfile is corrupted or malformed.

I also note that the `openssl` command outputs the error message `iv undefined`.

I've remade the ssh keys on my own fork, updated the various variable names that needed changing, and ran the CI, which seems to have got past this problem.

Speculation: Do the encryption environment variable names defined at https://travis-ci.org/tc39/test262/settings match those defined in `scripts/ci/deploy.sh` (`encrypted_7b3e5998334d_key` and `encrypted_7b3e5998334d_iv`)?

leobalter:
I'll try this.

leobalter:
Thanks! It's now fixed, see #996. I previously forgot to update the deploy script and to set travis to use this main repo instead of my fork.

I found it after you asked to check the encryption env variable names.

PR #992, opened 2017-04-20
Remove deploy scripts

leobalter:
This is a simplification of the current CI script, removing the deploy parts.

For now, every pull request or update into the master branch will check for manual changes in the generated tests, flagging an error if any is found.

This will also cancel the several emails from Travis errors in the master branch I'm receiving. Same might be happening to @tcare, I guess.


edgemaster:
Did you mean gh-546?

Indeed, it looks like the environment variable names in the deploy script weren't updated for this particular file encryption​ (and the step missed from the instructions too), so my speculation in gh-964 is likely to be correct and the deploy script just requires the correct names to function.

PR #851, opened 2017-02-07
Use root LICENSE file

leobalter:
Automatically attribute the root BSD license to allow test files without a copyright headers.

cc @bterlson @allenwb is this legally valid?

My motivation: I want to allow _new_ test files without the copyright header.

PR #994, opened 2017-04-23
Introduce automated validation for test format

jugglinmike:
This script is intended to identify common test file formatting errors
prior to their acceptance into the project. It is designed to support
future extensions for additional validation rules.

A good chunk of the new code here is for tests. I tried to make the format of
those tests as declarative as possible so that this would be easy to review and
maintain.


tools/lint/lib/checks/license.py
```diff
@@ -0,0 +1,43 @@
...
+    ID = 'LICENSE'
+
+    def run(self, name, meta, source):
+        if meta and 'flags' in meta and 'generated' in meta['flags']:
+            return
+
+        match = _LICENSE_PATTERN.search(source)
+
+        if not match:
+            return 'No license information found.'
```
leobalter: I'm positive #851 is on the way to reach final consensus soon, potentially in the next meeting. So I would avoid having this check. It's good to fix the current tests, but hopefully an unnecessary check for future test files.

tools/lint/lib/checks/license.py
```diff
@@ -0,0 +1,43 @@
+import re
+
+from ..check import Check
+
+_MIN_YEAR = 2009
+_MAX_YEAR = 2030
+
+_LICENSE_PATTERN = re.compile(
+    r'\/\/ Copyright( \([cC]\))? (\w+) .+\. {1,2}All rights reserved\.[\r\n]{1,2}' +
```
leobalter: regexp is hard, but would this work with files containing multiple copyright lines?

wrt the comment on #851, I'm not even sure this is worth investigating.

tools/scripts/ci_lint.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
+  paths=$(git diff --diff-filter ACMR --name-only $TRAVIS_BRANCH -- test/)
```
leobalter: `--diff-filter d` (lowercase d) is also a good option to avoid memorizing ACMR.

tools/lint/lib/checks/license.py
jugglinmike: Agreed.

tools/scripts/ci_lint.sh
jugglinmike: That sounds like a good shortcut for typing, but for a script, I would prefer to be explicit.

tools/lint/lib/checks/license.py
jugglinmike: How about we remove these two lines as part of gh-851?

jugglinmike:
@leobalter I just noticed that this patch was not passing its own tests. I've pushed up a commit that addresses the problem (it gets the tests to pass and includes the tests in the CI process).

leobalter:
I'm merging this and I'll follow up at #851 with the copyright parts.

lint.whitelist
```diff
@@ -0,0 +1,2 @@
+# This file documents authorship information and is not itself a test
```
cheerly: thanks

PR #995, opened 2017-04-24
Tests for precision of Math functions

littledan:
Tests for spec PR https://github.com/tc39/ecma262/pull/903


leobalter:
as a follow up - which doesn't need to be in this PR - we need to migrate `isEqual` and `getPrecision` harness to the assert object.

 isEqual(x, y) can become some sort of `assert.precision(x, y, 'message')` while getPrecision doesn't need to get exposed.

I would be +1 as well to add then to `harness/assert.js` and load this function as default.

@littledan I can do that and send a PR to your branch, if you agree with this change.

test/built-ins/Math/E/S15.8.1.1_A1.js
```diff
@@ -2,15 +2,12 @@
...
 description: Comparing Math.E with 2.7182818284590452354
-includes:
-    - math_precision.js
-    - math_isequal.js
 ---*/
 
 // CHECK#1
-if (!isEqual(Math.E, 2.7182818284590452354)) {
-  $ERROR('#1: \'Math.E is not approximately equal to 2.7182818284590452354\'');
+if (Math.E !== 2.718281828459045) {
```
leobalter: info and description are saying 2.7182818284590452354

This could use `assert.sameValue`

leobalter: further checks: info and description values are valid with the specs. Although, `2.718281828459045` is used in V8, SpiderMonkey, and Safari. I didn't check Edge.

Should we update the specs to reflect reality or fix these tests?

littledan: Well these extra digits just don't change the value of the number. In my spec PR, I omitted them for pi but forgot to for everything.

leobalter:
tc39/ecma262#903 is closed, how should we proceed here?

littledan:
Let's not pursue this.

PR #997, opened 2017-04-25
Tests for RegExp dotAll proposal

littledan:
Proposal from @mathiasbynens at

https://tc39.github.io/proposal-regexp-dotall-flag/

https://github.com/tc39/proposal-regexp-dotall-flag

These tests are mostly by @schuay . They pass on V8 with an experimental flag.


mathiasbynens:
Let’s update the commit message URL to https://github.com/mathiasbynens/es-regexp-dotall-flag so that it still works after the repo is moved (https://github.com/mathiasbynens/es-regexp-dotall-flag/issues/7). (I’ll try to do that later today.)

schuay:
Thanks alot for these!

anba:
Possible things we could also test:
- Position of "s" in string returned from RegExp.p.flags getter
- Duplicate "s" in regexp-literal and calls to RegExp and RegExp.p.compile throw SyntaxErrors
- `new RegExp("", "gimsuy")` (and `new RegExp().compile("", "gimsuy")`) works (https://chromium.googlesource.com/v8/v8.git/+/84dc8ed4c3e6c8c1e3005b2d2445c64328b139a4/src/objects.cc#16234 😉 )

schuay:
@anba nice catch, fixed here: https://codereview.chromium.org/2841633004/

littledan:
Added more tests that @anba suggested.

test/built-ins/RegExp/prototype/dotAll/this-val-invalid-obj.js
```diff
@@ -0,0 +1,29 @@
...
+assert.throws(TypeError, function() {
+  dotAll.call({});
+}, 'ordinary object');
+
+assert.throws(TypeError, function() {
+  dotAll.call([]);
+}, 'array exotic object');
+
+assert.throws(TypeError, function() {
+  dotAll.call(arguments);
```
leobalter: why arguments? This is already an Object. Maybe it would be better replacing it with `RegExp` itself or even better: [`RegExp.prototype`](https://tc39.github.io/ecma262/#sec-properties-of-the-regexp-prototype-object) that explicitly doesn't have any RegExp internals.

test/built-ins/RegExp/prototype/dotAll/this-val-non-obj.js
```diff
@@ -0,0 +1,40 @@
...
+    1. Let R be the this value.
+    2. If Type(R) is not Object, throw a TypeError exception.
+features: [Symbol, regexp-dotall]
+---*/
+
+var dotAll = Object.getOwnPropertyDescriptor(RegExp.prototype, 'dotAll').get;
+
+assert.throws(TypeError, function() {
+  dotAll.call(undefined);
+});
```
leobalter: as the `this-val-invalid-obj.js` file, this could use messages for each assertion

test/built-ins/RegExp/prototype/dotAll/this-val-regexp-prototype.js
```diff
@@ -0,0 +1,18 @@
...
+  1. Let R be the this value.
+  2. If Type(R) is not Object, throw a TypeError exception.
+  3. If R does not have an [[OriginalFlags]] internal slot, then
+     a. If SameValue(R, %RegExpPrototype%) is true, return undefined.
+features: [regexp-dotall]
+---*/
+
+var get = Object.getOwnPropertyDescriptor(RegExp.prototype, 'dotAll').get;
+
+assert.sameValue(get.call(RegExp.prototype), undefined);
```
leobalter: this test is invalid.

test/built-ins/RegExp/prototype/flags/order.js
```diff
@@ -0,0 +1,28 @@
+// Copyright (C) 2017 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 21.2.5.3
```
leobalter: we should use `esid`. We can use `es6id` as an extra but only if all the assertions in this file are valid for ES6. That's not the case.

test/built-ins/RegExp/dotall/with-dotall.js
```diff
@@ -0,0 +1,53 @@
...
+    3. Call CharacterSetMatcher(A, false) and return its Matcher result.
+
+esid: sec-atom
+features: [regexp-dotall]
+---*/
+
+// DotAll '.' behavior.
+{
+  let re = /^.$/s;
+  assert.sameValue(true, re.test("a"));
```
leobalter: this is weird. `true` should be the expected value as `assert.sameValue(actual, expected);` This format breaks convention.

test/built-ins/RegExp/dotall/with-dotall.js
```diff
@@ -0,0 +1,53 @@
...
+  assert.sameValue(true, re.test("3"));
+  assert.sameValue(true, re.test("π"));
+  assert.sameValue(true, re.test("\u2027"));
+  assert.sameValue(true, re.test("\u0085"));
+  assert.sameValue(true, re.test("\v"));
+  assert.sameValue(true, re.test("\f"));
+  assert.sameValue(true, re.test("\u180E"));
+  assert.sameValue(false, re.test("\u{10300}"));  // Supplementary plane.
```
leobalter: If it's work a line comment, it's probably work to be an assertion message.

test/built-ins/RegExp/dotall/without-dotall.js
```diff
@@ -0,0 +1,87 @@
...
+  assert.sameValue(true, re.test("\u180E"));
+  assert.sameValue(false, re.test("\u{10300}"));  // Supplementary plane.
+  assert.sameValue(false, re.test("\n"));
+  assert.sameValue(false, re.test("\r"));
+  assert.sameValue(false, re.test("\u2028"));
+  assert.sameValue(false, re.test("\u2029"));
+}
+
+// Default '.' behavior (unicode).
+{
```
leobalter: these blocks are again a reason we should split these cases into different test files.

mathiasbynens:
The proposal repository has moved. Please update the URLs accordingly:

```
https://github.com/mathiasbynens/es-regexp-dotall-flag
→
https://github.com/tc39/proposal-regexp-dotall-flag

https://mathiasbynens.github.io/es-regexp-dotall-flag/
→
https://tc39.github.io/proposal-regexp-dotall-flag/
```

test/built-ins/RegExp/prototype/dotAll/this-val-invalid-obj.js
littledan: I just copied these tests from the other flag tests (e.g., see `test/built-ins/RegExp/prototype/global/this-val-invalid-obj.js`). What if this patch let them go in as is, and a separate patch refactored all of them?

test/built-ins/RegExp/prototype/dotAll/this-val-non-obj.js
littledan: Also copied. Cleanup for another patch?

test/built-ins/RegExp/prototype/dotAll/this-val-regexp-prototype.js
littledan: What makes you think that? The behavior is provided by step 3.a. of https://tc39.github.io/proposal-regexp-dotall-flag/#sec-get-regexp.prototype.dotAll

test/built-ins/RegExp/prototype/flags/order.js
littledan: Right, thanks, will fix.

test/built-ins/RegExp/dotall/with-dotall.js
littledan: Right, I should just be using assert() here anyway. Fixed.

test/built-ins/RegExp/dotall/with-dotall.js
littledan: Done

test/built-ins/RegExp/dotall/without-dotall.js
littledan: Done

test/built-ins/RegExp/prototype/dotAll/this-val-regexp-prototype.js
leobalter: https://github.com/tc39/test262/pull/997/files/8720eaff38c7c9385c82d84a22a118a48cf64f38#diff-cbc9af224a51ffa4ffbc6a16826deca6R5

leobalter: I used the info tag there as the reference. Checking the specs, that info tag is wrong.

test/built-ins/RegExp/prototype/dotAll/this-val-invalid-obj.js
```diff
@@ -0,0 +1,29 @@
...
+
+/*---
+description: Invoked on an object without an [[OriginalFlags]] internal slot
+esid: pending
+info: >
+    get RegExp.prototype.dotAll
+
+    1. Let R be the this value.
+    2. If Type(R) is not Object, throw a TypeError exception.
+    3. If R does not have an [[OriginalFlags]] internal slot, throw a TypeError
```
leobalter: 3. If R does not have an [[OriginalFlags]] internal slot, then
  a. If SameValue(R, %RegExpPrototype%) is true, return undefined.
  b. Otherwise, throw a TypeError exception.

PR #998, opened 2017-04-25
RegExp named group tests

littledan:
Tests against the Stage 3 named capture groups proposal
https://tc39.github.io/proposal-regexp-named-groups

These tests are mostly by @schuay, from V8's tests for the feature


test/built-ins/RegExp/named-groups/functional-replace.js
```diff
@@ -0,0 +1,154 @@
...
+  third-to-last argument
+esid: pending
+features: [regexp-named-groups]
+---*/
+
+// --- Unicode mode ---
+
+// @@replace with a callable replacement argument (no named captures).
+{
+  let result = "abcd".replace(/(.)(.)/u, (match, fst, snd, offset, str) => {
```
leobalter: the `{}` wrappers are a perfect indication these tests should be split into multiple files. There's already the description for each part.

It would be great to provide a `info` tag containing the spec parts allowing each cases. It's a valuable reference to match the validity for each assertion.

test/built-ins/RegExp/named-groups/unicode-malformed.js
```diff
@@ -0,0 +1,27 @@
+// Copyright 2017 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Various syntax errors for Unicode RegExps containing named groups
+esid: pending
+features: [regexp-named-groups]
+---*/
+
+assert.throws(SyntaxError, () => eval("/(?<>a)/u"), "Empty name");
```
leobalter: +1 for each assertion message.

test/built-ins/RegExp/named-groups/non-unicode-property-names.js
```diff
@@ -0,0 +1,45 @@
...
+features: [regexp-named-groups]
+includes: [compareArray.js]
+---*/
+
+assert.sameValue("a", /(?<π>a)/.exec("bab").groups.π);
+assert.throws(SyntaxError, () => eval('/(?<\\u{03C0}>a)/'), "\\u{} escapes allowed only in Unicode mode");
+assert.sameValue("a", /(?<π>a)/.exec("bab").groups.\u03C0);
+assert.sameValue("a", /(?<$>a)/.exec("bab").groups.$);
+assert.sameValue("a", /(?<_>a)/.exec("bab").groups._);
+assert.sameValue("a", /(?<$𐒤>a)/.exec("bab").groups.$𐒤);
```
anba: This should be an early syntax error, because 𐒤 is parsed as two separate code units in non-Unicode mode. 

https://tc39.github.io/ecma262/#sec-pattern-semantics
> If a BMP pattern contains a non-BMP SourceCharacter the entire pattern is encoded using UTF-16 and the individual code units of that encoding are used as the elements of the List.

test/built-ins/RegExp/named-groups/functional-replace.js
littledan: OK, I broke up some tests further and added an info tag to some tests.

test/built-ins/RegExp/named-groups/non-unicode-property-names.js
littledan: @anba Good catch; do you think we should enforce this, or change RegExpIdentifierStart and RegExpIdentifierContinue to allow surrogate pairs?

anba: If we allow 𐒤 in BMP patterns, we probably also need to allow `\ud801\udca4` (and maybe even `\u{104a4}`) for consistency reasons, because it'd strange to accept a surrogate pair only if it's written in non-escaped form. (If we want to keep it simple, we can forbid 𐒤 for the time being, and if there's any user demand, we can later revisit this decision and allow 𐒤. This won't cause any compatibility issues, because we'd only change an error into a non-error.)

littledan: OK, fixed the test

PR #999, opened 2017-04-25
Tests for RegExp lookbehind

littledan:
Thanks @hashseed for writing these tests; this patch is a simple port.

### DIFF
--- a/harness/compareArray.js
+++ b/harness/compareArray.js
@@ -13,3 +13,17 @@ function compareArray(a, b) {
   return true;
 }
 
+assert.compareArray = function(actual, expected, message) {
+  if (compareArray(actual, expected)) return;
+
+
+  if (message === undefined) {
+    message = '';
+  } else {
+    message += ' ';
+  }
+
+  message += 'Expected SameValue(«' + String(actual) + '», «' + String(expected) + '») to be true';
+
+  $ERROR(`${message}${message === undefined ? '' : ' '}Expected the arrays [${actual}] to have the same contents as [${expected}]`);
+}
--- a/harness/compareArray.js
+++ b/harness/compareArray.js
@@ -14,16 +14,6 @@ function compareArray(a, b) {
 }
 
 assert.compareArray = function(actual, expected, message) {
-  if (compareArray(actual, expected)) return;
-
-
-  if (message === undefined) {
-    message = '';
-  } else {
-    message += ' ';
-  }
-
-  message += 'Expected SameValue(«' + String(actual) + '», «' + String(expected) + '») to be true';
-
-  $ERROR(`${message}${message === undefined ? '' : ' '}Expected the arrays [${actual}] to have the same contents as [${expected}]`);
+  assert(compareArray(actual, expected),
+         `Expected [${actual.join(", ")}] and [${expected.join(", ")}] to have the same contents. ${message}`);
 }
--- a/test/built-ins/RegExp/lookbehind.js
+++ b/test/built-ins/RegExp/lookbehind.js
@@ -1,0 +1,169 @@
+// Copyright (C) 2017 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-assertion
+description: Test RegExp lookbehind
+info: |
+    The production Assertion :: (?<=Disjunction) evaluates as follows:
+      1. Evaluate Disjunction with -1 as its direction argument to obtain a Matcher m.
+      2. Return an internal Matcher closure that takes two arguments, a State x and a Continuation
+         c, and performs the following steps:
+        a. Let d be a Continuation that always returns its State argument as a successful MatchResult.
+        b. Call m(x, d) and let r be its result.
+        c. If r is failure, return failure.
+        d. Let y be r's State.
+        e. Let cap be y's captures List.
+        f. Let xe be x's endIndex.
+        g. Let z be the State (xe, cap).
+        h. Call c(z) and return its result.
+features: [regexp-lookbehind]
+includes: [compareArray.js]
+---*/
+
+// Simple fixed-length matches.
+assert.compareArray(["a"], "a".match(/^.(?<=a)/));
+assert.sameValue(null, "b".match(/^.(?<=a)/));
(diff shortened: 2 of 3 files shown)


test/built-ins/RegExp/lookbehind.js
```diff
@@ -0,0 +1,169 @@
...
+assert.sameValue(null, "foooo".match(/"^foooo(?<=^o*)$/));
+assert.compareArray(["foo"], "foo".match(/^foo(?<=^fo+)$/));
+assert.compareArray(["foooo"], "foooo".match(/^foooo(?<=^fo*)/));
+assert.compareArray(["foo", "f"], "foo".match(/^(f)oo(?<=^\1o+)$/));
+assert.compareArray(["foo", "f"], "foo".match(/^(f)oo(?<=^\1o+)$/i));
+assert.compareArray(["foo\u1234", "f"], "foo\u1234".match(/^(f)oo(?<=^\1o+).$/i));
+assert.compareArray(["def"], "abcdefdef".match(/(?<=^\w+)def/));
+assert.compareArray(["def", "def"], "abcdefdef".match(/(?<=^\w+)def/g));
+
+// Word boundary matches.
```
leobalter: splitting these tests in multiple files help identifying the purpose of each one.

Plus, it helps identifying multiple aspects.

test/built-ins/RegExp/lookbehind.js
```diff
@@ -0,0 +1,169 @@
...
+assert.sameValue(null, "foo".match(/^foo(?<!foo)$/));
+assert.sameValue(null, "foo".match(/^f.o(?<!foo)$/));
+assert.compareArray(["fno"], "fno".match(/^f.o(?<!foo)$/));
+assert.compareArray(["foooo"], "foooo".match(/^foooo(?<=fo+)$/));
+assert.compareArray(["foooo"], "foooo".match(/^foooo(?<=fo*)$/));
+assert.compareArray(["abc", "abc"], /(abc\1)/.exec("abc"));
+assert.compareArray(["abc", "abc"], /(abc\1)/.exec("abc\u1234"));
+assert.compareArray(["abc", "abc"], /(abc\1)/i.exec("abc"));
+assert.compareArray(["abc", "abc"], /(abc\1)/i.exec("abc\u1234"));
+var oob_subject = "abcdefghijklmnabcdefghijklmn".substr(14);
```
leobalter: why not use `"abcdefghijklmn"`?

leobalter:
lookBehing is in stage 3, so this PR has blocker other than the review feedback.

test/built-ins/RegExp/lookbehind.js
littledan: Many JS implementations use a sort of virtual slice for situations like this, without copying. This test verifies that such implementations don't make the mistake of allowing lookbehind to go beyond the start of the string.

harness/compareArray.js
```diff
@@ -13,3 +13,17 @@ function compareArray(a, b) {
...
+
+  if (message === undefined) {
+    message = '';
+  } else {
+    message += ' ';
+  }
+
+  message += 'Expected SameValue(«' + String(actual) + '», «' + String(expected) + '») to be true';
+
+  $ERROR(`${message}${message === undefined ? '' : ' '}Expected the arrays [${actual}] to have the same contents as [${expected}]`);
```
littledan: Gah, I meant to delete those lines in the middle!

Changed as suggested.

test/built-ins/RegExp/lookbehind.js
littledan: Would splitting up the tests help more than the comments in the test file helps? The tests in a particular file are all getting at the same bit of spec text.

test/built-ins/RegExp/lookbehind.js
leobalter: If the idea is catching very specific behaviors, that's why I insist we should have split files.

At least a comment explaining this, otherwise the `substr` seems completely unnecessary for the purpose of testing the final value in `oob_subject`

rwaldron: @littledan ...

> Many JS implementations use a sort of virtual slice for situations like this, without copying. This test verifies that such implementations don't make the mistake of allowing lookbehind to go beyond the start of the string.

With respect, I have a hard time believing that the code you're referring to is actually doing what you claim it's doing.

hashseed: Since this test originates from V8, this targets V8's sliced strings, which are used for substrings above a length of 13 characters. I wrote this test for exactly the reason @littledan mentioned. That's why the variable name is called oob_subject. The underlying string backing store extends beyond the actual boundary of the sliced string.

leobalter: I only need some sort of description like a comment - it's better in a separate file - otherwise this can be removed in any further maintenance because it's implementation specific and I can't observe the behavior from the specs.

rwaldron: Ok, so it's a purely V8 implementation specific issue? If that's the case, and it's not testing anything that's actually observable per spec, then it falls in one of the following: 

1. Doesn't belong here.
3. Belongs in a special "shame" file.

rwaldron: Honestly, this is like saying that test262 should start supporting V8's testing macros.

leobalter:
I'll get this fixed

test/built-ins/RegExp/lookbehind.js
hashseed: I agree that that part should be removed.

I don't feel really strongly about this, and I think neither does @littledan. He simply adapted the tests I wrote while implementing lookbehind for V8 because that is more convenient than coming up with ones from scratch. It's not like we are trying to force them into test262.

littledan: I disagree that we should have such a high bar for getting test262 tests in, that we shouldn't have tests that target possible bug sources. If it is a test that's correct according to the spec, is it really creating such a big maintenance burden? I bet V8 isn't the only engine to treat strings like this either--this is likely useful for other engines too. Having such strict requirements for tests just discourages contributions and gives us all less test coverage.

PR #1000, opened 2017-04-25
Avoid using $ERROR in test files

leobalter:
while this is just a few changes, it's a headstart to avoid using $ERROR directly in test files and defer checks for the assertion api.

I'm doing this in my own free time and that's what my free time allowed me for now. I'll continue working in patches for the other files but I don't wanna hold this with the chance to forget this branch.


anba:
> I'm doing this in my own free time and that's what my free time allowed me for now. I'll continue working in patches for the other files but I don't wanna hold this with the chance to forget this branch.

I have somewhere a branch where I tried to perform these update automatically (similar to #417). I could try to resurrect that branch.

test/annexB/built-ins/RegExp/RegExp-decimal-escape-class-range.js
```diff
@@ -19,24 +19,11 @@ var __expected = ["1\n\n\n\n\nl"];
...
-}
+assert.sameValue(__executed.length, __expected.length, '.length');
+assert.sameValue(__executed.index, __expected.index, '.index');
+assert.sameValue(__executed.input, __expected.input, '.input');
 
 //CHECK#4
-for(var index=0; index<__expected.length; index++) {
-	if (__executed[index] !== __expected[index]) {
-		$ERROR('#4: __executed = /[\\d][\\12-\\14]{1,}[^\\d]/.exec("line1\\n\\n\\n\\n\\nline2"); __executed[' + index + '] === ' + __expected[index] + '. Actual: ' + __executed[index]);
-	}
```
rwaldron: lol @ all of these

leobalter:
@anba that would be amazing! I was going too deep with the optimization here but maybe even a simple conversion from these `if (x !== y)` cases to assert.sameValue is already a big win.

PR #1001, opened 2017-04-25
Remove duplicate

leobalter:
I found this duplicate case while working on some syntax error tests.

The deleted case is a duplicate of `src/dstr-assignment/array-elem-target-simple-strict.case`


src/dstr-assignment/array-elem-init-simple-strict.case
```diff
@@ -1,21 +0,0 @@
...
-es6id: 12.14.5.1
-flags: [onlyStrict]
-negative:
-  phase: early
-  type: SyntaxError
----*/
-
-//- elems
-[arguments]
-//- values
```
leobalter: It also had a wrong naming. `vals` should be the right thing here.

rwaldron:
Nice clean up. 

LGTM

PR #1002, opened 2017-04-25
Add early error cases for `async async` in property definitions

leobalter:
I've been working over [an original patch](https://bugzilla.mozilla.org/show_bug.cgi?id=1358246) from @anba to the SpiderMonkey tests suite, while I was trying to find a good way to port them here.

This approach is simple, fast, and allows a quick porting. 

The alternative is migrating each case to an individual file with the negative flag.

There are some trade offs for each option. The negative flag allows the purest form to verify an early error, nothing can beat it being the minimal unit part to assert.

In the meanwhile, this assert.throws.early api allows not only a quick port, but verifying the code being parsed internally as in `Function()`, `eval()`, and an indirect eval call.

I'm not silver into any of these forms, my preference is whatever communicates better with those contributing and consuming the tests from test262. For that, I'm requesting feedback.

### DIFF
--- a/harness/assert.js
+++ b/harness/assert.js
@@ -79,3 +79,10 @@ assert.throws = function (expectedErrorConstructor, func, message) {
   message += 'Expected a ' + expectedErrorConstructor.name + ' to be thrown but no exception was thrown at all';
   $ERROR(message);
 };
+
+assert.throws.early = function(err, code) {
+  let wrappedCode = `function wrapperFn() { ${code} }`;
+  let ieval = eval;
+
+  assert.throws(err, () => { Function(wrappedCode); }, `Function: ${code}`);
+};
--- a/test/harness/assert-throws-early-incorrect-ctor.js
+++ b/test/harness/assert-throws-early-incorrect-ctor.js
@@ -1,0 +1,20 @@
+// Copyright (C) 2017 Mozilla Corporation. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: >
+  Functions that throw values whose constructor does not match the specified
+  constructor do not satisfy the assertion.
+---*/
+
+// monkeypatch the API
+$ERROR = function $ERROR(message) {
+  throw new Test262Error(message);
+};
+
+assert.throws(Test262Error, () => {
+  assert.throws.early(SyntaxError, "1 = 1;");
+}, "'1=1' is a ReferenceError");
+assert.throws(Test262Error, () => {
+  assert.throws.early(ReferenceError, "var;");
+}, "'var;' is a SyntaxError");
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -159,7 +159,8 @@ $DONE(arg) | see Writing Asynchronous Tests, below
 assert(value, message) | throw a new Test262Error instance if the specified value is not strictly equal to the JavaScript `true` value; accepts an optional string message for use in creating the error
 assert.sameValue(actual, expected, message) | throw a new Test262Error instance if the first two arguments are not [the same value](https://tc39.github.io/ecma262/#sec-samevalue); accepts an optional string message for use in creating the error
 assert.notSameValue(actual, unexpected, message) | throw a new Test262Error instance if the first two arguments are [the same value](https://tc39.github.io/ecma262/#sec-samevalue); accepts an optional string message for use in creating the error
-assert.throws(expectedErrorConstructor, fn) | throw a new Test262Error instance if the provided function does not throw an error, or if the constructor of the value thrown does not match the provided constructor
+assert.throws(expectedErrorConstructor, fn, message) | throw a new Test262Error instance if the provided function does not throw an error, or if the constructor of the value thrown does not match the provided constructor
+assert.throws.early(expectedErrorConstructor, fn) | throw a new Test262Error instance if the provided function does not throw an early error, or if the constructor of the value thrown does not match the provided constructor. This assertion catches only errors that will be parsed through `Function(code)`.
 
 ```
 /// error class
--- a/test/harness/assert-throws-early-not-early.js
+++ b/test/harness/assert-throws-early-not-early.js
@@ -1,0 +1,16 @@
+// Copyright (C) 2017 Mozilla Corporation. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: >
+    The assertion fails when the code does not parse with an early error
+---*/
+
+// monkeypatch the API
+$ERROR = function $ERROR(message) {
+  throw new Test262Error(message);
+};
+
+assert.throws(Test262Error, () => {
+  assert.throws.early(ReferenceError, 'x = 1');
+});
--- a/test/harness/assert-throws-early-referenceerror.js
+++ b/test/harness/assert-throws-early-referenceerror.js
@@ -1,0 +1,9 @@
+// Copyright (C) 2017 Mozilla Corporation. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: >
+    The assertion pass when the code parses with an early ReferenceError
+---*/
+
+assert.throws.early(ReferenceError, '1 = 1;');
--- a/test/harness/assert-throws-early-syntaxerror.js
+++ b/test/harness/assert-throws-early-syntaxerror.js
@@ -1,0 +1,9 @@
+// Copyright (C) 2017 Mozilla Corporation. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: >
+    The assertion pass when the code parses with an early SyntaxError
+---*/
+
+assert.throws.early(SyntaxError, 'let let');
--- a/test/language/expressions/object/prop-def-invalid-async-prefix.js
+++ b/test/language/expressions/object/prop-def-invalid-async-prefix.js
@@ -1,0 +1,33 @@
+// Copyright (C) 2017 Mozilla Corporation. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: >
+  async is not a valid prefix of an identifier reference
+esid: sec-object-initializer
+info: |
+    PropertyDefinition:
+      IdentifierReference
+      CoverInitializedName
+      PropertyName : AssignmentExpression
+      MethodDefinition
+
+    MethodDefinition:
+      PropertyName ( UniqueFormalParameters ) { FunctionBody }
+      AsyncMethod
+
+    AsyncMethod:
+      async [no LineTerminator here] PropertyName ( UniqueFormalParameters ) { AsyncFunctionBody }
+      VariableDeclaration : BindingPattern Initializer
+
+      1. Let rhs be the result of evaluating Initializer.
+      2. Let rval be GetValue(rhs).
+      3. ReturnIfAbrupt(rval).
+      4. Return the result of performing BindingInitialization for
+         BindingPattern passing rval and undefined as arguments.
+negative:
+  phase: early
+  type: SyntaxError
+---*/
+
+({async async});


leobalter:
Reference for myself: using the _negative flag_ approach would require creating syntax error templates on `src/dstr-binding/syntax` similar to the forms in the `src/dstr-binding/error/`, it also requires some other templates for the object initializers.

tcare:
The only case that this won't work for is anything that has different behavior with early errors under eval. I think that's a fine limitation.

I prefer this method for anything that isn't affected by eval semantics. We can create separate files for any cases that need it.

leobalter:
Thanks for the review, @tcare. I'll wait a bit more as I also collected feedback against it in a private conversation. I would like to weight all in before moving ahead with this new harness extension.

Still, I'm leaning into this API as it seems to me similar to approaches used by other runtimes' own test suites.

bakkot:
I think I prefer the alternative (building out a file for each case) over this api, despite the upsides.

In particular, this is essentially testing syntax, but requires a full engine, and implements its own error-checking. If we had a file per test, the consumer could choose how to assert the error, and would be able to e.g. test just the parser, instead of having to pass it to the implementation and rely on it.

jugglinmike:
> I think I prefer the alternative (building out a file for each case) over
> this api, despite the upsides.
>
> In particular, this is essentially testing syntax, but requires a full
> engine, and implements its own error-checking. If we had a file per test, the
> consumer could choose how to assert the error, and would be able to e.g. test
> just the parser, instead of having to pass it to the implementation and rely
> on it.

Agreed. I've previously spoken to @littledan about using the negative syntax
tests to verify parity between V8's "pre-parse" and "parse" phases. Unless I'm
mistaken, expressing tests in this way would prevent that use. Is that right,
Dan?

For my own part, [I recently incorporated Test262 into a JavaScript linter's
test suite](https://github.com/jshint/jshint/pull/3098), and I've been using
the results to fix a bunch of previously-unreported bugs. I haven't even
started on new language features, but that's where I anticipate getting the
most value.

littledan:
I agree with what's said above about the incomplete testing for early errors. Some engines might invoke the preparser for the function constructor, but IIRC V8 typically eager-parses from that. If we want to put multiple things in the same file and test for early errors, maybe this could be done by embedder hooks.

leobalter:
I get that. While the current patch allows easier collaboration to test262, the negative tests provide better quality for the tests.

For now what I can do is use the test generation tool and add extra templates for the wrapped code to be evaluated in Function and eval.

littledan:
@leobalter What's the motivation for using the template tool this way? Do we suspect that we won't get early errors for some of these cases when passed into the Function constructor, when we would as a script? From the linked bug, it doesn't look like the fix had anything to do with the different entrypoints; why should this test be treated differently from other early error tests?

harness/assert.js
```diff
@@ -79,3 +79,12 @@ assert.throws = function (expectedErrorConstructor, func, message) {
...
 };
+
+assert.throws.early = function(err, code) {
+  var wrappedCode = `function wrapperFn() { ${code} }`;
+  var ieval = eval;
+
+  assert.throws(err, () => { Function(wrappedCode); }, `Function: ${code}`);
+  assert.throws(err, () => { eval(wrappedCode); }, `eval: ${code}`);
+  assert.throws(err, () => { ieval(wrappedCode); }, `indirect eval: ${code}`);
+};
```
mathiasbynens: This needs docs in `CONTRIBUTING.md` and tests in `test/harness`.

leobalter: sure thing. I didn't want to write docs and tests while I was only requesting feedback. This is a prototype.

If we are moving this forward, I'll provide docs and tests before merging.

leobalter:
> What's the motivation for using the template tool this way? 

I'm pushing new changes soon with templates covering parse and pre-parse phases for this case example.

leobalter:
Ok, this is now using the templates and generating code

test/language/statements/variable/async-async-identifier-assignment-parse.js
```diff
@@ -0,0 +1,31 @@
...
+      ...
+
+    ObjectBindingPattern:
+      {}
+      { BindingPropertyList }
+      { BindingPropertyList, }
+---*/
+
+
+assert.throws.early(SyntaxError, `var {async async = 0} = {};`);
```
leobalter: this is now generated code using a similar model for the early parse error.

If the pre-parse and parse are different, this approach must be reasonable, even if longer.

I tried to reuse the templates instead of duplicating each case file. For that I would need to define the negative flags there instead of each case. The test generation tool does not support that yet.

rwaldron:
> parity between V8's "pre-parse" and "parse" phases

...

> Some engines might invoke the preparser for the function constructor,

I'd like some further clarification about the nature of "pre-parse" and "parse" phases.

leobalter:
@mathiasbynens I've got some tests up, would you like to review them?

test/harness/assert-throws-early-incorrect-ctor.js
```diff
@@ -0,0 +1,20 @@
...
+  constructor do not satisfy the assertion.
+---*/
+
+// monkeypatch the API
+$ERROR = function $ERROR(message) {
+  throw new Test262Error(message);
+};
+
+assert.throws(Test262Error, () => {
+  assert.throws.early(SyntaxError, "1 = 1;");
```
mathiasbynens: `ReferenceError` instead of `SyntaxError`

mathiasbynens: Hmm… How did this test pass? 🤔

harness/assert.js
littledan: In V8, this assert.throws.early will never get at preparser behavior. The Function constructor currently always invokes the parser, rather than the preparser, with the heuristic reasoning being that calls to the Function constructor are less likely to be dead code. If you want to make sure there's as good coverage for this as there is for the frontmatter early error declarations, you'll need host integration.

Do you have any particular reason to believe that invoking these three calls is more likely to get at an early error than doing what harnesses do for early errors? In V8, I'll be surprised if find much of a difference among them, beyond the semantic differences (global vs in a function to itself vs directly in assert.throws.early).

test/harness/assert-throws-early-incorrect-ctor.js
littledan: I guess because of the outer negative expectation?

mathiasbynens: Ah, of course! Forget what I said 😊

littledan:
@rwaldron V8 has two modes that the parser can run in. The preparser is invoked only for certain cases when we heuristically predict that the code might not be run; all code that's executed is later parsed. There's also a third mode in V8, "lazy compilation", where we run the full parser but then throw away the result rather than compiling it, if it's a situation where we need something that comes out of the parser that the preparser doesn't support (e.g., scope-related information).

Preparsing and lazy compilation are expensive, but at the same time, there's a lot of dead JavaScript code on the web, so there are heuristics to skip the preparsing step and compile the first time around. There are many things that can trip the parser into action, skipping the preparser--for example, a function enclosed in parens is currently parsed eagerly in V8, and V8 recently added support for eagerly parsing a function preceded by `!`. The function constructor is always eagerly parsed. And if a function is larger than a particular size and looks like a collection of other functions, as discovered on-line by the preparser, V8 may abort pre-parsing and eagerly parse it. The logic here continues to evolve; V8 has been improving support for widely used packaging systems.

It's hard to get at all these states just by running JS code, so V8 has a test suite which explicitly invokes the preparser, even for cases where it wouldn't've run naturally. It's in `test/cctest/test-parsing.cc`, if you're curious. This test suite verifies that the parser and preparser accept and reject the same code, and it's based on templating (since there are so many cases to cover).

anba:
> From the linked bug, it doesn't look like the fix had anything to do with the different entrypoints; why should this test be treated differently from other early error tests?

I can confirm this. There was no particular reason to parse the source code using different functions, I only copied the `assertSyntaxError` function from other existing tests. In fact, SpiderMonkey will also use its full parser for all these cases, so we don't actually get broader coverage when direct eval + indirect eval + Function is used.

test/harness/assert-throws-early-incorrect-ctor.js
leobalter: yes :) I'm asserting the call will fail if I give the wrong constructor. It was tricky when I wrote it as well.

leobalter:
> In fact, SpiderMonkey will also use its full parser for all these cases, so we don't actually get broader coverage when direct eval + indirect eval + Function is used.

I can reduce the api to use Function only as it's enough for our goals.

> It's hard to get at all these states just by running JS code, so V8 has a test suite which explicitly invokes the preparser, even for cases where it wouldn't've run naturally. It's in test/cctest/test-parsing.cc, if you're curious.

This seems like runtimes running tests using `assert.throws.early(ErrorConstructor, codeString)` can patch this assertion to parse the contents of `codeString` invoking the preParser or anything else. Test harness already patch `$PRINT` and other values including the `$262` api.

In anyway, I still don't feel why early errors should be not parsed as early errors in different parsers.

I understand the logic for using negative tests but the more we discuss I believe the assertion api is much better for collaboration and even readability. I don't mind adding more test files to provide coverage, but I really want some good reason.

---

My current solution in this PR is using both approaches, negative + assertion api,  I want to know where to follow after this as I need to move on another tests as well.

leobalter:
After a conversation in a private channel, I'm confirming the purpose of these tests is to cover a regression in a specific runtime only.

`({ async async })` is a simple invalid syntax, without any match in the specs neither anything specific for early errors.

Considering test262 can't predict all the myriads of invalid syntax in JS, a regression test for a runtime sits better in the specific runtime if we can't find any explicit support from the specs.

I'm closing this for this reasons but I'll be happy to reuse `assert.throws.early` if necessary in any further case. (cc @mathiasbynens)

leobalter:
> I don't think regressions within a single runtime need to be kept out of test262, in general!

We cannot afford to cover regressions unless the respective behavior is explicitly specified in the specs. The case here is simply a non-match for any of the syntax grammar of ES. It might be probably a good addition to the parser tests repo.

The bug I linked already provided a patch fixing it, and AFAIK, the other runtimes are not failing.

In other hand, @rwaldron and I just found a runtime mismatch while discussing this on code that is defined. I'll push a new PR for it soon.

littledan:
What about adding a single test case for this, which is just a negative SyntaxError expectation for `({async async})`? It seems meaningful to have some sort of test; I don't know if it's necessary to go crazy with templates for every little syntax error test, but I also don't think we need to have such a high bar in general.

Last time we discussed whether regression tests are included, I thought we came to the conclusion that they shouldn't be excluded, that contributors can make their own decisions for priorities, and if a contributor wants to put in a valid test, then we don't have to reject it just because multiple implementations don't fail it.

rwaldron:
> What about adding a single test case for this, which is just a negative SyntaxError expectation for `({async async})`?

Can you cite the specific grammar or static semantics early error, or runtime syntax error that this tests?

> if a contributor wants to put in a valid test, then we don't have to reject it just because multiple implementations don't fail it.

As far as we can tell, this isn't a valid test. If you can provide a citation as I've requested, I will happily concede and consider it a learning opportunity.

syg:
Sorry I'm late to the party. I like @rwaldron's summary here for the citation-needed bar, but share @littledan's concerns about how high the bar should be. On the one hand, blind acceptance of these kind of narrower syntax tests might detract from the overall quality of tests. On the other hand, if the process for an implementer to determine if a test262 should be written and upstreamed is too time consuming (like testing all the other engines), I fear that'll de-incentivize participating in test262. On the third hand, offloading the burden of determining whether a test is worthy of inclusion entirely to test262 maintainers also sucks. What balance can we strike here?

I'm also unclear on how to interpret the citation-needed bar for most negative syntax tests. Outside of explicit early errors, wouldn't all negative syntax tests be because there is no specific grammar that matched the string? Some of things that don't parse would be worthwhile for inclusion, though. For instance, contextual keyword stuff.

leobalter:
I'll probably come back with something that balances in the complexity to fix this part here.

mathiasbynens:
Please land `assert.throws.early` (even if it’s a separate PR). I could really use it in https://github.com/mathiasbynens/unicode-property-escapes-tests/pull/8.

leobalter:
Done. I'm only merging one of the tests for now using the negative flag. If we want more cases, we should follow up in another PR.

PR #1003, opened 2017-04-26
Add tests for observable [[GetOwnProperty]] ops in EnumerableOwnProperties

caitp:


caitp:
Hi, this could probably use some improvements in frontmatter, and a similar test could be written for Object.values() and Objcet.entries().

test/built-ins/Object/keys/proxy-keys.js
```diff
@@ -0,0 +1,128 @@
+// Copyright (C) 2017 Caitlin Potter. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: >
+  Object.keys() observably performs [[GetOwnProperty]]
```
leobalter: +1 to merge after you add the `esid: sec-object.keys` plus the following:

```yaml
info: |
  19.1.2.16 Object.keys ( O )

  1. Let obj be ? ToObject(O).
  2. Let nameList be ? EnumerableOwnProperties(obj, "key").
  ...

  7.3.21 EnumerableOwnProperties ( O, kind )

  1. Assert: Type(O) is Object.
  2. Let ownKeys be ? O.[[OwnPropertyKeys]]().
  3. Let properties be a new empty List.
  4. For each element key of ownKeys in List order, do
    a. If Type(key) is String, then
      i. Let desc be ? O.[[GetOwnProperty]](key).
      ...
```

caitp: added verbatim

Issue #1008, opened 2017-04-28
Issues in new RegExp tests for named-capturing and dotAll


anba:
test/built-ins/RegExp/prototype/flags/order.js
- Uses `RegExp.prototype.compiles`, so it needs to be split into a normal and an Annex-B test.

test/built-ins/RegExp/named-groups/non-unicode-property-names.js
- Contains an unterminated string literal in line 16.
- `assert(RegExp("(?<a\u{104A4}>.)").test("a"), "Surrogate, ID_Continue");` in line 33 is invalid, BMP-patterns read `\u{104A4}` as a two-element character unit sequence, so in this case a SyntaxError should be expected.

test/built-ins/RegExp/named-groups/non-unicode-malformed.js
- Contains multiple tests which are only valid with Annex-B semantics, without Annex-B, a RegExp literal like `/\k/` should throw a SyntaxError. So these parts need to be moved into the test/annexB/RegExp folder.
- Contains tests which expect literals like `/(?<a>.)\k/` to throw a SyntaxError, but with the current named capturing groups spec, it should be parsed like `/(?<a>.)\\k/` (@littledan this looks like an error in the proposal spec? Or alternatively we need to apply the change from https://github.com/tc39/ecma262/pull/864#issuecomment-291926968, so ExtendedPatternCharacter doesn't include `\`.) 

test/built-ins/RegExp/named-groups/string-replace-nocaptures.js
- The test mustn't include the global flag.

leobalter:
thanks for catching these bugs. I'm fixing them.

leobalter:
@anba I'm having a hard time to find specs match for `/\k/`. It seems like `\k` is just a regular and valid source character.

I could not find anything on the Annex B and the rest of the specs.

leobalter:
https://tc39.github.io/proposal-regexp-named-groups/#prod-AtomEscape

I've got it. Sorry for the noise. I was looking at the actual specs

leobalter:
If I read the proposal well, the `\k` needs to be matched with a group name. Otherwise it is just a source character.

```
AtomEscape:
  ...
  k GroupName

GroupName:
  < RegExpIdentifierName >
```

So, cases like `/\k/`, `/\k<a/`, `/\ka/`, `/\ka>/`, are still valid matches where `\k` is only a source scape for `k`.

That's my understanding, I would need @littledan to confirm.

anba:
[...]

PR #1009, opened 2017-04-28
Update RegExp tests for named-capturing and dotAll

leobalter:
Ref #1008 

It does not address the `\k` parts as it's still being discussed there.


littledan:
LGTM, thanks for these fixes!

PR #1010, opened 2017-04-28
for-await-of: dstr-binding templates

rwaldron:
This is a work in progress, do not merge

Signed-off-by: Rick Waldron <waldron.rick@gmail.com>

### DIFF
--- a/src/dstr-binding/default/for-await-of-const.template
+++ b/src/dstr-binding/default/for-await-of-const.template
@@ -1,0 +1,51 @@
+// Copyright (C) 2017 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+path: language/statements/for-await-of/dstr-const-
+name: for-await-of statement
+esid: sec-for-in-and-for-of-statements-runtime-semantics-labelledevaluation
+features: [destructuring-binding, async-iteration]
+flags: [async]
+info: |
+    IterationStatement :
+        for await ( ForDeclaration of AssignmentExpression ) Statement
+
+    [...]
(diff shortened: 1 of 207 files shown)


src/dstr-binding/default/for-await-of-const.template
```diff
@@ -0,0 +1,51 @@
...
+               1. Assert: lhsKind is lexicalBinding.
+               2. Assert: lhs is a ForDeclaration.
+               3. Let status be the result of performing BindingInitialization
+                  for lhs passing nextValue and iterationEnv as arguments.
+          [...]
+---*/
+
+var iterCount = 0;
+
+async function fn() {
```
leobalter: clone these templates to support async generators as well.

file names can be like:

- `src/dstr-binding/default/for-await-async-func-const.template`
- `src/dstr-binding/default/for-await-async-gen-const.template`

and the `path` (frontmatter) be like:

- `path: language/statements/for-await-async-function/dstr-const-`
- `path: language/statements/for-await-async-generator/dstr-const-`

rwaldron:
After discussing offline, @leobalter agreed to let me do this: 

> clone these templates to support async generators as well.

...in a follow up PR

src/dstr-binding/default/for-await-of-const.template
leobalter: In a private chat, we decided to have this in a follow up PR, the tests are technical valid and the follow up fix will only add extra coverage.

leobalter:
Did you sign the CLA?

rwaldron:
🖕

PR #1011, opened 2017-04-28
for-await-of: dstr-binding, async func & async gen templates

rwaldron:
@leobalter baby steps! Please just review the templates themselves

Signed-off-by: Rick Waldron <waldron.rick@gmail.com>


src/dstr-binding/default/for-await-of-async-gen-const.template
```diff
@@ -0,0 +1,51 @@
+// Copyright (C) 2017 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+path: language/statements/for-await-of/dstr-const-
```
leobalter: this will generate files in the same path of the async function templates

leobalter: `language/statements/for-await-of/async-gen-dstr-const-`

leobalter:
good! time to generate the tests and merge this.

PR #1012, opened 2017-04-29
Add tests for completion value of function / class declaration statements

JosephPecoraro:
Check completion value of more declarations are empty:

```js
eval(`1; function f(){}`) === 1
eval(`1; function* f(){}`) === 1
eval(`1; async function f(){}`) === 1
eval(`1; class C {}`) === 1
```

Browsers:
* Firefox passes all
* Chrome fails for `class` - [bug](https://bugs.chromium.org/p/chromium/issues/detail?id=716715)
* WebKit fails for `class` - [bug](https://bugs.webkit.org/show_bug.cgi?id=171265)
* Edge fails for `class` - [bug](https://github.com/Microsoft/ChakraCore/issues/2890)

### DIFF
--- a/test/language/statements/async-function/cptn-decl.js
+++ b/test/language/statements/async-function/cptn-decl.js
@@ -1,0 +1,13 @@
+// Copyright (C) 2017 Apple Inc. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-async-function-definitions-runtime-semantics-evaluation
+description: Async function declaration completion value is empty.
+info: >
+    AsyncFunctionDeclaration : async [no LineTerminator here] function BindingIdentifier ( FormalParameters ) { AsyncFunctionBody }
+
+    1. Return NormalCompletion(empty).
+---*/
+
+assert.sameValue(eval('async function f() {}'), undefined);
+assert.sameValue(eval('1; async function f() {}'), 1);
--- a/test/language/statements/class/cptn-decl.js
+++ b/test/language/statements/class/cptn-decl.js
@@ -1,0 +1,14 @@
+// Copyright (C) 2017 Apple Inc. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-class-definitions-runtime-semantics-evaluation
+description: Class declaration completion value is empty.
+info: >
+    ClassDeclaration : class BindingIdentifier ClassTail
+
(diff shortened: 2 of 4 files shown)


littledan:
Thanks for the contribution! Keep them coming.

It looks from the specification that the completion value of class declarations is the constructor, rather than empty. Because of this, I think Firefox is the engine with a bug, rather than the other three.

anba:
> It looks from the specification that the completion value of class declarations is the constructor, rather than empty.

No, an empty completion value is correct -> https://tc39.github.io/ecma262/#sec-class-definitions-runtime-semantics-evaluation.

littledan:
Oh sorry, I was looking at ClassExpression.

JosephPecoraro:
> You may also add a test for async generator declaration (adding the features: [async-iteration] frontmatter tag, or we can work on that as a follow up.

I would rather someone address this in a follow-up. Its not in the current ES spec, and it looks like it does not behave like other declarations, and instead returns the function instead of an empty completion like all of these declarations. See https://tc39.github.io/proposal-async-iteration/#prod-AsyncGeneratorDeclaration

> AsyncGeneratorDeclaration : async [no LineTerminator here] function* BindingIdentifier ( FormalParameters ) { AsyncGeneratorBody }
>
> ...
> 3. Let F be ! AsyncGeneratorFunctionCreate(Normal, FormalParameters, AsyncGeneratorBody, scope, strict).
> ...
> 7. Return F.

leobalter:
> I would rather someone address this in a follow-up

This is fine.

> Its not in the current ES spec, ...

This is tricky. Test262 capture tests not only for the current spec draft, but also stage 3 proposals. The motivation is allowing runtimes to implement correct features before Stage 4.

Anyway, that doesn't make any requirement to land everything together, but it opens an reminder we should cover async generators as well. I'll create an issue to keep track.

Thanks for the tests!

littledan:
@JosephPecoraro I see that text, but it looks like it's under the heading Runtime Semantics: InstantiateFunctionObject. I actually don't see a Runtime Semantics: Evaluation for async generators (or ordinary generators for that matter), but maybe I'm missing something.

JosephPecoraro:
> Test262 capture tests not only for the current spec draft, but also stage 3 proposals.

I didn't realize that. Thanks for the clarification.

> If you look at the 8th bullet of this section, it mentions adding a clause for evaluation analogous to GeneratorDeclaration, which evaluates to NormalCompletion(empty). So I believe async generators should have an empty completion value just like other types of declarations.

Ahh, I see. Thanks.

PR #1014, opened 2017-05-01
Add regular RegExp Unicode property escape tests

mathiasbynens:
Ref. https://github.com/mathiasbynens/unicode-property-escapes-tests/pull/8.


test/built-ins/RegExp/property-escapes/character-class.js
```diff
@@ -0,0 +1,25 @@
...
+author: Mathias Bynens
+description: >
+  Unicode property escapes must be supported in character classes.
+esid: sec-static-semantics-unicodematchproperty-p
+features: [regexp-unicode-property-escapes]
+---*/
+
+/[\p{Hex}]/u;
+assert(
+  /[\p{Hex}-\uFFFF]/u.test('-'),
```
anba: This regular expression should result in a SyntaxError, because of step 1 in https://tc39.github.io/ecma262/#sec-runtime-semantics-characterrange-abstract-operation.

mathiasbynens: Good catch! Patch amended.

mathiasbynens: @anba Your commit in https://github.com/tc39/ecma262/commit/d898df271008762e73e9408a1da2887835709e86 replaced "If _A_ does not contain exactly one character or _B_ does not contain exactly one character, throw a *SyntaxError* exception." with "Assert: _A_ and _B_ each contain exactly one character.". Where is the error thrown now?

anba: It's now an early syntax error. 

https://tc39.github.io/ecma262/#sec-patterns-static-semantics-early-errors

NonemptyClassRanges :: ClassAtom - ClassAtomClassRanges
- It is a Syntax Error if IsCharacterClass of the first ClassAtom is true or IsCharacterClass of the second ClassAtom is true. 

A similar restriction is in place for " NonemptyClassRangesNoDash :: ClassAtomNoDash - ClassAtomClassRanges".

Annex B modifies the early syntax error for non-Unicode patterns in https://tc39.github.io/ecma262/#sec-patterns-static-semantics-early-errors-annexb.

mathiasbynens: Thanks for confirming. I'll update this test accordingly and upstream it to Test262.

PR #1017, opened 2017-05-02
Expression statement lookahead restriction and identifier tests

anba:
This started as a PR to create tests for https://bugzilla.mozilla.org/show_bug.cgi?id=1361143, it got a bit sidetracked, though.

- 75d153d has the tests for https://bugzilla.mozilla.org/show_bug.cgi?id=1361143
- ExpressionStatement also has a lookahead restriction for "let [" &rarr; 7a86bae
- Apropos "let", we don't seem to have tests for "let" with escapes (, or any other contextual keyword) &rarr;  3d8fe2a
- And also some tests for "yield" and "await" with escape sequences as labels &rarr; d887db8
- More "yield" and "await" with escape sequences identifier tests &rarr; 36a8672, 490b2dc, e0c2931
- And because I had to read the existing identifier tests while creating these commits, I also took the opportunity to rename the identifier tests and to remove a few duplicates. (053d63f, bd7c280, 4d95f88, 4494db1, 69735fd)


leobalter:
wow, this is amazing! I still need more time to review this but I really like everything so far.

PR #1020, opened 2017-05-03
for-await-of: dstr-binding, error, async func & async gen templates

rwaldron:


src/dstr-binding/error/for-await-of-async-func-var.template
```diff
@@ -0,0 +1,47 @@
...
+---*/
+
+async function fn() {
+  for await (var /*{ elems }*/ of [/*{ vals }*/]) {
+    return;
+  }
+}
+
+fn()
+  .then(_ => { throw new Test262Error("Expected async function to reject, but resolved."); }, ({ constructor }) => assert.sameValue(constructor, /*{ error }*/))
```
leobalter: something we can use in the next time is using `$DONE` instead of throwing a new error: 

```js
.then(_ => { $DONE("Expected async function to reject, but resolved."); }, /* ... */ );
```

This can be changed in a follow, I'm not blocking a PR because of this.

leobalter: because both ways are valid.

rwaldron: Let's do a follow up

test/language/statements/for-await-of/async-func-dstr-const-ary-ptrn-elision-step-err.js
```diff
@@ -0,0 +1,74 @@
...
+    return;
+  }
+}
+
+fn()
+  .then(_ => { throw new Test262Error("Expected async function to reject, but resolved."); }, ({ constructor }) => assert.sameValue(constructor, Test262Error))
+  .then($DONE, $DONE);
+
+
+iter.next();
```
leobalter: this seems like an issue from the case (not the templates you created in this PR).

You should not need to call `.next()` here to confirm the next assertion(s). The same goes for every other .js file you can find this `iter.next()`.

The iteration already happened above with iter, so asserting this specific iterator has closed early, is calling the next assertions directly. Some of them already secure the iterator runs.

We should have a follow up PR to remove these lines.

test/language/statements/for-await-of/async-gen-dstr-const-ary-ptrn-elision-step-err.js
```diff
@@ -0,0 +1,74 @@
...
+  }
+}
+
+gen().next()
+  .then(_ => { throw new Test262Error("Expected async function to reject, but resolved."); }, ({ constructor }) => assert.sameValue(constructor, Test262Error))
+  .then($DONE, $DONE);
+
+
+iter.next();
+assert.sameValue(following, 0, 'Iterator was properly closed.');
```
leobalter: this should follow the same pattern in `src/dstr-binding/ary-ptrn-rest-id-iter-step-err.case` and have a sameValue for `first` before throwing an error. Again, for a follow up PR. iter.next() is unnecessary.

test/language/statements/for-await-of/async-gen-dstr-const-ary-ptrn-rest-id-iter-step-err.js
```diff
@@ -0,0 +1,71 @@
...
+}();
+
+async function * gen() {
+  for await (const [...x] of [iter]) {
+    return;
+  }
+}
+
+gen().next()
+  .then(_ => { throw new Test262Error("Expected async function to reject, but resolved."); }, ({ constructor }) => assert.sameValue(constructor, Test262Error))
```
leobalter: this is really interesting. This test matches perfectly the expected behavior in https://tc39.github.io/proposal-async-iteration/#sec-asyncgeneratorstart.

SpiderMonkey and V8 are both failing on the same tests for this PR (I haven't checked with other runtimes yet). This promise is being resolved, instead. +1 @rwaldron adding the Test262Error in the resolve call.

If I read https://tc39.github.io/proposal-async-iteration/#await properly, the rejection should be projected to the async context of the async generator, right?

leobalter: The observed failure: the promise is being fulfilled, while the test expects it to be rejected.

anba: The test case doesn't take into account that there's an implicit Await in step 6.c of https://tc39.github.io/proposal-async-iteration/#sec-runtime-semantics-forin-div-ofbodyevaluation-lhs-stmt-iterator-lhskind-labelset and that the array iterator is wrapped in an Async-from-Sync iterator (https://tc39.github.io/proposal-async-iteration/#sec-async-from-sync-iterator-objects). So what we need to do, is to delay calling `iter.next()` by enqueuing some extra Promise jobs:

```js
Promise.resolve()
  .then()
  .then()
  .then(() => {
    iter.next();
    assert.sameValue(first, 1);
    assert.sameValue(second, 0, 'Iterator is closed following abrupt completion.');
  })
  .then($DONE, $DONE);
```

PR #1021, opened 2017-05-03
for-await-of: dstr-assignment, syntax template

rwaldron:


rwaldron:
This is trash, do not use.

leobalter:
:)

What comes next is the future! 

There's even a movie with this title.

In this case the future is async. I "promise".

PR #1022, opened 2017-05-03
Add minimal documentation for features flags

leobalter:


littledan:
Would it make sense to note that, when adding tests for a new proposal, that it's a good idea to add a new feature flag?

leobalter:
@littledan I believe there's a lot to improve in this document. My limited time allowed me only start it. 

> it's highly recommended to reuse the features flags for any matching case in new tests.

That's the part where I'm basically saying I'll ask for feature flags in new tests. My goal, but not for this PR, is to make it a requirement for any stage 3 proposal. I'm avoiding discussions that might block shipping the current text.

littledan:
Sure, this text is better than no text; I don't want to slow you down!

leobalter:
I'm adding this file as we should continue working on top of it.

>Would it make sense to note that, when adding tests for a new proposal, that it's a good idea to add a new feature flag? 

@littledan I want to make this a requirement, I'll open another PR this week to get feedback, but I believe we're gonna be fine on this.

PR #1025, opened 2017-05-05
Fix test for ES5 & add new test for ES6

tmysik:
The original test used `es5id` but verified `ES6` behavior. Added new test which verifies behavior on `ES6`.

There are more similar tests like this one that could be fixed this way but since I am not sure whether my understanding of `es5id` etc. is correct, creating PR for just one test for now :)

Thank you.

### DIFF
--- a/test/built-ins/Function/prototype/call/S15.3.4.4_A9.js
+++ b/test/built-ins/Function/prototype/call/S15.3.4.4_A9.js
@@ -17,11 +17,11 @@ if (!(Function.prototype.call.hasOwnProperty('length'))) {
 }
 
 //CHECK#1
-if (!delete Function.prototype.call.length) {
-  $ERROR('#1: The Function.prototype.call.length property does not have the attributes DontDelete');
+if (delete Function.prototype.call.length) {
+  $ERROR('#1: The Function.prototype.call.length property does have the attributes DontDelete');
 }
 
 //CHECK#2
-if (Function.prototype.call.hasOwnProperty('length')) {
-  $ERROR('#2: The Function.prototype.call.length property does not have the attributes DontDelete');
+if (!(Function.prototype.call.hasOwnProperty('length'))) {
+  $ERROR('#2: The Function.prototype.call has length property');
 }
--- a/test/built-ins/Function/prototype/call/length.js
+++ b/test/built-ins/Function/prototype/call/length.js
@@ -1,0 +1,26 @@
+// Copyright (C) 2017 Tomas Mysik. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+info: >
+    The Function.prototype.call.length property is configurable
+es6id: 19.2.3.3
+description: >
+    Checking if deleting the Function.prototype.call.length property
+    fails
+---*/
+
+//CHECK#0
+if (!(Function.prototype.call.hasOwnProperty('length'))) {
+  $ERROR('#0: the Function.prototype.call has length property');
+}
+
+//CHECK#1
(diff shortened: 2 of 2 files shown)


leobalter:
`es5id` is only a reference of the spec parts for the respective test.

Test262 does not maintain exclusive tests for ES5 or ES6, unless you rely on the git history, we can't afford the maintenance on multiple spec versions. The only target is the latest specs draft and stage 3+ proposals.

For ES2015, we used `es6id` and after this we started using `esid` to rely in organic references.

With that said, this patch should not try to cover different versions. The es5id tag can be fixed to use es6id or, even better, `esid`.

rwaldron:
In addition to @leobalter's notes... these tests should be replaced with tests that follow this convention: https://github.com/tc39/test262/commit/f5d8b1c1fb043e104ed36da92a01af35e4d74fbd

tmysik:
I see, thank you for you answer. So, is there a way to run tests for `ES5`? Using `es5-tests` branch from this repository maybe? I do not see any `tag` or `release` for `ES 5[.1]` (if it is part of this repository).

Feel free to close/cancel this PR, of course.

Thank you.
CC @leobalter @rwaldron

leobalter:
That branch is not currently maintained, it's there for a long time only. It should be still the best option to check ES5, but I assume it will cover ES5.1. 

Is there any reason to provide coverage for these older versions? The new editions replace the  previous editions, so ES5.1 is not even a recommendation anymore, it is there for historical reasons only.

edgemaster:
The es5-tests branch does cover ES5.1 behaviour. However various test semantics there are unclear and vary depending upon the original source of the test. A large effort went into cleaning up these cases for the ES6+ (master) test suite. There are [several incompatible differences](https://tc39.github.io/ecma262/2017/#sec-additions-and-changes-that-introduce-incompatibilities-with-prior-editions) between ES5 and 6 to take a note of as well.

I have previously used both the ES5 and ES6 versions of the test suite to test an incomplete-but-ES5-conformant implementation with filtering to remove invalid results. I found that it was far easier to use the master branch, and remove the invalid results, than to use the es5-tests branch and work with an untidier set of tests.

I also strongly recommend that current implementations are written to the current specification text, even if they opt not to implement sections of it (eg, for prototyping purposes). The majority of the core language features are unchanged between versions.

leobalter:
There's no ES5-conformant today. We can have something conforming to ECMAScript, the latest released edition is the only one currently valid as a standard.

> I also strongly recommend that current implementations are written to the current specification text, even if they opt not to implement sections of it (eg, for prototyping purposes).

+1

Parts of the specs can be literally detached in an implementation, like syntax grammar for arrow functions or API parts. That's probably the best approach for a modern subset conforming to parts of the specs as international standards.

One thing we are trying to fix on Test262 that might help on this is using the `features` flag more often in the frontmatter header of each file. This way those tests can be disabled, filtered or skipped by checking the flag values.

tmysik:
Guys, thank you for the explanation and your time.

edgemaster:
> One thing we are trying to fix on Test262 that might help on this is using the features flag more often in the frontmatter header of each file. This way those tests can be disabled, filtered or skipped by checking the flag values.
This would be of great help for testing incomplete implementations, one of the biggest issues we had was filtering out operations on built-ins we hadn't implemented yet, but were used in tests in other sections of the spec. I should take a look at my testing results database to see if I can quickly filter out anything for a PR to assist this task. Do you have an issue number for this?

leobalter:
I only added the FEATURES.md file to the project so far listing the current features flags we're using throughout the project. The project still needs improvement in the documentation.

PR #1027, opened 2017-05-10
Allow async failures

leobalter:
This is a long but very simple change where I'm adding a $DONE callback for promises rejection, catching failures for a proper report.

I found this bug while running the tests with Spidermonkey in a browser environment.

I'm not changing anything in the templates for for-await-of as @rwaldron is already working on that.


rwaldron:
> I found this bug while running the tests with Spidermonkey in a browser environment.

Can you link to the test that surfaced the issue? Thanks!

rwaldron:
After reviewing the failure that @leobalter mentions above, I'm not convinced that this change is necessary.

leobalter:
Same, it was my first thought, but I was wrong and at the same time convinced we just need to make sure that a chain of then calls will enqueue the rejection callbacks properly.

PR #1028, opened 2017-05-10
Add tests for Array.prototype methods when length property exceeds or is near integer limit

anba:
New tests when Array.prototype methods are called with array-like objects whose length property is near or exceeds the safe integer limit 2<sup>53</sup>-1.

And I've renamed slice/create-species-undef-invalid-len.js to slice/create-proxied-array-invalid-len.js and updated its description to reflect the actual test behaviour.

Ref: https://bugzilla.mozilla.org/show_bug.cgi?id=924058

### DIFF
--- a/test/built-ins/Array/prototype/pop/clamps-to-integer-limit.js
+++ b/test/built-ins/Array/prototype/pop/clamps-to-integer-limit.js
@@ -1,0 +1,35 @@
+// Copyright (C) 2017 André Bargull. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-array.prototype.pop
+description: >
+  Length values exceeding 2^53-1 are clamped to 2^53-1.
+info: |
+  1. ...
+  2. Let len be ? ToLength(? Get(O, "length")).
+  ...
+  4. Else len > 0,
+  a. Let newLen be len-1.
+  ...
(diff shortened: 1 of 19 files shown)


test/built-ins/Array/prototype/pop/clamps-to-integer-limit.js
```diff
@@ -0,0 +1,35 @@
...
+  ...
+  e. Perform ? Set(O, "length", newLen, true).
+  ...
+---*/
+
+var arrayLike = {};
+
+arrayLike.length = 2**53 - 1;
+Array.prototype.pop.call(arrayLike);
+assert.sameValue(arrayLike.length, 2**53 - 2, "Length is 2**53 - 1");
```
leobalter: If _Length values exceeding 2^53-1 are clamped to 2^53-1._,  the expected value should be `2 ** 53 - 1`, right?

but wait...

This is clamping everything to `2**53-2`! Is that intended? cc @bterlson 

We should remove the `a. Let newLen be len-1.` after using `ToLength`, IMHO.

rwaldron:
These are excellent!

test/built-ins/Array/prototype/pop/clamps-to-integer-limit.js
leobalter: omg, so much noice. Of course this is a .pop function.

PR #1030, opened 2017-05-10
Set a proper length reference to avoid conflict with browsers

leobalter:
cc @syg, @anba

### DIFF
--- a/src/dstr-binding/ary-ptrn-rest-obj-prop-id.case
+++ b/src/dstr-binding/ary-ptrn-rest-obj-prop-id.case
@@ -18,6 +18,8 @@ info: |
        [...]
 ---*/
 
+//- setup
+let length = "outer";
 //- elems
 [...{ 0: v, 1: w, 2: x, 3: y, length: z }]
 //- vals
@@ -29,6 +31,4 @@ assert.sameValue(x, 9);
 assert.sameValue(y, undefined);
 assert.sameValue(z, 3);
 
-assert.throws(ReferenceError, function() {
-  length;
-});
+assert.sameValue(length, "outer", "the length prop is not set as a binding name");
--- a/test/language/expressions/arrow-function/dstr-ary-ptrn-rest-obj-prop-id.js
+++ b/test/language/expressions/arrow-function/dstr-ary-ptrn-rest-obj-prop-id.js
@@ -50,6 +50,7 @@ info: |
              BindingPattern with A and environment as the arguments.
        [...]
 ---*/
+let length = "outer";
 
 var callCount = 0;
 var f;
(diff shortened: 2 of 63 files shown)


src/dstr-binding/ary-ptrn-rest-obj-prop-id.case
```diff
@@ -29,6 +31,4 @@ assert.sameValue(x, 9);
 assert.sameValue(y, undefined);
 assert.sameValue(z, 3);
 
-assert.throws(ReferenceError, function() {
-  length;
-});
+assert.sameValue(length, "outer", "the length prop is set as a binding name");
```
leobalter: typo... fixing it

rwaldron:
Can you include an explanation "avoid conflict with browsers", with links to bugs—thanks!

src/dstr-binding/ary-ptrn-rest-obj-prop-id.case
```diff
@@ -18,6 +18,8 @@ info: |
        [...]
 ---*/
 
+//- setup
+let length = "outer";
```
rwaldron: This might be a good use case for a Symbol, it would live in a harness file

leobalter:
Following the message from the commit:

> Set a proper length reference to avoid conflict with browsers
>
> In browsers length is a valid property of window, so they would have a normal
> completion instead of a ReferenceError

`length` is a known property of Window, returning the _number of child browsing contexts_.

https://www.w3.org/TR/html5/browsers.html#dom-length
https://developer.mozilla.org/en-US/docs/Web/API/Window/length

I notice this issue while working on https://bugzilla.mozilla.org/show_bug.cgi?id=1362169, where I noticed tests failing while running in a browser (Firefox) but not in the same respective runtime (Spidermonkey).

If we throw a ReferenceError when we try to evaluate a `length` binding name, the tests won't be valid if running in browser environments.

It's not any bug from outside, it's just a really bad choice of browsers shipping this property named length on the window object.

src/dstr-binding/ary-ptrn-rest-obj-prop-id.case
leobalter: I don't get what would be different using a Symbol value. 

```js
const s = Symbol("outer");
const length = s;

function scope(/* ... */) {
  assert.sameValue(length, s);
}

scope();
```

this would be like the same as checking if the value as an string.

One thing that is interesting in this case, is that we already protect it in the dstr pattern:

```js
[...{ 0: v, 1: w, 2: x, 3: y, length: z }]

/* ... */
assert.sameValue(z, 3);
```

When we assert `z` is 3, we could assume a wrong `length` would be 3, or close. Not `"outer"`.

leobalter:
Ok, after a private conversation, I agreed to clarify my comments:

When I say:

> If we throw a ReferenceError when we try to evaluate a `length` binding name, the tests won't be valid if running in browser environments.

A better wording should be:

__If we throw a ReferenceError when we try to evaluate a `length` binding name, which should not exist at all, the tests won't be valid if running in browser environments.__

That's because a global `length` does not exist in ECMAScript, as the reference above, it's part of the HTML5 API specs defined by W3C.

The current change makes the tests valid for both in the browser's environment and in an isolated runtime run.

PR #1034, opened 2017-05-15
Fix test involving a Unicode property escape in a character class range

mathiasbynens:
Thanks to @anba for spotting this: https://github.com/tc39/test262/pull/1014/files/4843f049fafb6cdd61e3be011e3701cc94a9fe46#r116528031

### DIFF
--- a/test/built-ins/RegExp/property-escapes/character-class.js
+++ b/test/built-ins/RegExp/property-escapes/character-class.js
@@ -10,9 +10,11 @@ features: [regexp-unicode-property-escapes]
 ---*/
 
 /[\p{Hex}]/u;
-assert(
-  /[\p{Hex}-\uFFFF]/u.test('-'),
-  'property escape in character class should not be parsed as the start of a range'
+assert.throws(
+  SyntaxError,
+  () => /[\p{Hex}-\uFFFF]/u,
+  // See step 1 of https://tc39.github.io/ecma262/#sec-runtime-semantics-characterrange-abstract-operation.
+  'property escape at start of character class range should throw if it expands to multiple characters'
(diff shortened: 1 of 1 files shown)


leobalter:
I just wanna make sure, this is a runtime error, right? Quickly checking the specs, I assume yes.

mathiasbynens:
Yeah, https://tc39.github.io/ecma262/#sec-runtime-semantics-characterrange-abstract-operation is listed under runtime semantics, not static semantics. (I checked before submitting this PR.)

anba:
It's complicated. 

ES 5.1 had:
> An implementation must treat any instance of the following kinds of errors as an early error:
> - [...]
> - Errors in regular expression literals that are not implementation-defined syntax extensions.

Somehow this was dropped in ES6 which leads to at least one contradiction, in https://tc39.github.io/ecma262/#sec-pattern, we've got:
> [...] The algorithms in 21.2.2 are designed so that compiling a pattern may throw a SyntaxError exception; on the other hand, once the pattern is successfully compiled, applying the resulting internal procedure to find a match in a String cannot throw an exception (except for any host-defined exceptions that can occur anywhere such as out-of-memory).

But a regular expression like `/a|(?![\d-\d])/u` may or may not evaluate the right-hand side disjunction (which contains an invalid range expression), it depends on the input string. 

So it should be listed as an early error in the spec, but no one spend time on correcting this issue...

mathiasbynens:
@anba Woah, good catch! Could you file this as a tc39/ecma262 bug?

leobalter:
We have https://github.com/tc39/ecma262/issues/918 for now.

PR #1037, opened 2017-05-15
Add templates for dstr-binding w/ for-await-of looping over async iterators

leobalter:
I've got some failures here on V8 6.0.0, while SpiderMonkey is running flawlessly here.


src/dstr-binding/default/for-await-of-async-func-const-async.template
```diff
@@ -0,0 +1,54 @@
...
+  yield* [/*{ vals }*/];
+})();
+
+async function fn() {
+  for await (const /*{ elems }*/ of asyncIter) {
+    /*{ body }*/
+
+    iterCount += 1;
+  }
+}
```
rwaldron: This test is awesome.

PR #1038, opened 2017-05-16
Fix features flags

leobalter:
I also double checked against duplicate `features` on the changed files.

Before the first commit:

```
git grep -c features $(git diff --cached --name-only)
```


rwaldron:
Nice

PR #1039, opened 2017-05-16
[WIP] for-await-of: dstr-assignment, templates & cases

rwaldron:


rwaldron:
Not ready for review

leobalter:
no problem, ping here when it's ready.

leobalter:
These are still missing the respective generated tests.

rwaldron:
> These are still missing the respective generated tests.

I decided to do this in a follow up PR that moved them to the correctly named directory—instead of generating tests, moving the source, deleting the tests and then generating the tests.

leobalter:
That's reasonable, thanks for the clarification.

PR #1040, opened 2017-05-17
Remove unnecessary tests

leobalter:
These tests are not adding any coverage to the respective API they
should be covering.

There are other files checking cases for array-like objects, and
playing this with the global might cause issues and unnexpected
behavior on specific environments like browsers.


rwaldron:
I went back to the fifth edition and confirmed that these tests are indeed pointless.

littledan:
I think it'd be great if we could figure out a way to test that these functions work well against the global object while being compatible with running tests in the browser. I don't think such tests are irrelevant, as I wrote about in this thread: https://github.com/tc39/test262/pull/1041

PR #1041, opened 2017-05-17
Remove more unnecessary tests

leobalter:
These tests are not helpful for their respective APIs being tested.

They don't add any coverage for the built-in apis and basically check
behaviors that might conflict with a implementation defined global.


littledan:
Could you explain a little bit more? It looks like they're testing something real. Are there other tests that these are redundant with? In many JS implementations, due to the surrounding web specs, the global object is implemented a bit specially, so I think it's worth it to have tests that the global object in particular acts correctly.

One small aspect: I don't think we should avoid having tests that do things with the global object out of fear for conflicting with things; we can just fix conflicts when they arise (like the $ -> $262 change).

leobalter:
I'm glad to explain. I'm sorry if it's too long.

My decision is after the fact Test262 is vendor independent. Even if all know browsers are the most used targets and - along with NodeJS - those implementations should always be accountable when we avoid conflicts or any surprise.

While we want to avoid conflicts, as the example you mentioned (`$` to `$262`), or avoiding verifying reference errors for global `length`, we can not - at the same time - assume what `this` is gonna behave or all of its intrinsics.

These tests removed here are testing unit parts of the `Object` built-in api. Let's see this example, testing `Object.defineProperty`:

```js
var obj = this;
 
Object.defineProperty(obj, "0", {
  value: 1002
});
```

This is assuming the implementation's global will always allow a `[[DefineOwnProperty]]` operation on it's global referenced as `this` and also setting the property `0` with the respective descriptor is fine. Ecma262 does not specify that. We could make something compatible with the HTML5's DOM specs or Node.js' global, but that's still an assumption.

Look at this description:

> [8.1 Lexical Environments](https://tc39.github.io/ecma262/#sec-lexical-environments)
> 
> A global environment is a Lexical Environment which does not have an outer environment. The global environment's outer environment reference is null. A global environment's EnvironmentRecord may be prepopulated with identifier bindings and includes an associated global object whose properties provide some of the global environment's identifier bindings. As ECMAScript code is executed, additional properties may be added to the global object and the initial properties may be modified.

The specs are open to descriptions where the properties __may__  be added or modified.

The other problem is: what exactly we want to test? Look at the description of the test:

> Object.defineProperty will update [[Value]] attribute of indexed property successfully when [[Configurable]] attribute is true and [[Writable]] attribute is false, 'O' is the global object (8.12.9 step Note)

This is verifying the api function, and the unit we want to verify is value being successfully modified.

__This can be  properly covered - and already is - with an ordinary object.__

If we are testing Ecma262's specific behaviors of the global object, we should have a direct reference for it, that would become the unit to be tested. The problem is that we don't have that unit referenced in Ecma262, so I'm afraid we don't have anything to test against.

---

This all doesn't mean it can't be tested, but I'm also afraid Test262 is not the proper place for it. Specific behavior of the browser's global object is described by the W3C HTML5's DOM API. We should hopefully have a coverage for that in the WPT. Any error is not Ecma262's runtime specific, but `window`'s implementation.

littledan:
The latitude that embedding environments have is so broad that I believe many many test262 tests are invalid. Many tests contain a global var declaration, which would be invalid if the embedder contains something of the same name in the global lexical contour. To take a random example, `annexB/built-ins/escape/to-string-observe.js` includes a global declaration `var obj, long;` which would throw in that case. In a test262 snapshot that I have locally, there are 33251 var declarations at the beginning of a line, which are probably mostly global var declarations.

I'm not sure about this notion that it's not worth it to test how things work applied to different things. If it were the case, there would be no point to have any of the template tests. But, in fact, those tests are useful and keep finding issues of interaction between different features. I don't understand what it means for something to be properly covered from a spec's perspective; I think we just have to be pragmatic here and look at what interactions seem likely to cause issues in realistic implementations. Since the global object is special in a number of ways, I think it falls into that case.

If tests262 is a bad place to hold tests like this, rather than just completely deleting the tests, I like your idea about moving these tests to web-platform-tests. Maybe that could be done around the same time tests are removed, to avoid a gap in coverage for web browsers that are continuously under development.

leobalter:
> The latitude that embedding environments have is so broad that I believe many many test262 tests are invalid. Many tests contain a global var declaration, which would be invalid if the embedder contains something of the same name in the global lexical contour. To take a random example, annexB/built-ins/escape/to-string-observe.js includes a global declaration var obj, long; which would throw in that case. In a test262 snapshot that I have locally, there are 33251 var declarations at the beginning of a line, which are probably mostly global var declarations.

Having custom names in the top scope is something we still need to improve with fine tunings and cleanups.

> I'm not sure about this notion that it's not worth it to test how things work applied to different things. If it were the case, there would be no point to have any of the template tests. But, in fact, those tests are useful and keep finding issues of interaction between different features. I don't understand what it means for something to be properly covered from a spec's perspective; I think we just have to be pragmatic here and look at what interactions seem likely to cause issues in realistic implementations. Since the global object is special in a number of ways, I think it falls into that case.

The cases I removed here are also causing other known compatibility issues defining specific properties in browser envs.

> If tests262 is a bad place to hold tests like this, rather than just completely deleting the tests, I like your idea about moving these tests to web-platform-tests. Maybe that could be done around the same time tests are removed, to avoid a gap in coverage for web browsers that are continuously under development.

For most things, Test262 is still a good place, whenever we can test parts of the specs. Due to its own legacy, there are still a lot of parts to be improved, but that demands time. 

For the WPT, I'm not sure if it already exists, but it would be great to add some specific ES integration tests there.

littledan:
> The cases I removed here are also causing other known compatibility issues defining specific properties in browser envs.

Oh, OK, what's the particular issue here?

littledan:
> Having custom names in the top scope is something we still need to improve with fine tunings and cleanups.

What's a way that this could be addressed? There are all kinds of special things about the global scope, so it's important to have at least some tests that run there. But for any particular name that's chosen, an embedding environment could exist which conflicts with it.

Issue #1043, opened 2017-05-19
Incorrect for-await tests because implicit Await and Async-from-Sync iterator behaviour is ignored


anba:
The following for-await tests are incorrect:

- test/language/statements/for-await-of/async-func-dstr-const-ary-ptrn-elision-step-err.js
- test/language/statements/for-await-of/async-func-dstr-const-ary-ptrn-rest-id-iter-step-err.js
- test/language/statements/for-await-of/async-func-dstr-let-ary-ptrn-elision-step-err.js
- test/language/statements/for-await-of/async-func-dstr-let-ary-ptrn-rest-id-iter-step-err.js
- test/language/statements/for-await-of/async-func-dstr-var-ary-ptrn-elision-step-err.js
- test/language/statements/for-await-of/async-func-dstr-var-ary-ptrn-rest-id-iter-step-err.js
- test/language/statements/for-await-of/async-gen-dstr-const-ary-ptrn-elision-step-err.js
- test/language/statements/for-await-of/async-gen-dstr-const-ary-ptrn-rest-id-iter-step-err.js
- test/language/statements/for-await-of/async-gen-dstr-let-ary-ptrn-elision-step-err.js
- test/language/statements/for-await-of/async-gen-dstr-let-ary-ptrn-rest-id-iter-step-err.js
- test/language/statements/for-await-of/async-gen-dstr-var-ary-ptrn-elision-step-err.js
- test/language/statements/for-await-of/async-gen-dstr-var-ary-ptrn-rest-id-iter-step-err.js

See https://github.com/tc39/test262/pull/1020#discussion_r116531160 for an explanation.

@syg, @caitp per discussion at http://logs.libuv.org/tc39/2017-05-19#01:10:48.834

leobalter:
I'm trying to analyze the files and I need to write it down here to see if I get the error:

On `test/language/statements/for-await-of/async-func-dstr-const-ary-ptrn-rest-id-iter-step-err.js`: 

```js
var first = 0;
var second = 0;
var iter = function*() {
  first += 1;
  throw new Test262Error();
  second += 1;
}();

async function fn() {
  for await (const [...x] of [iter]) {
    return;
  }
}

fn()
  .then(_ => { throw new Test262Error("Expected async function to reject, but resolved."); }, ({ constructor }) => assert.sameValue(constructor, Test262Error))
  .then($DONE, $DONE);


iter.next();
assert.sameValue(first, 1);
[...]

PR #1044, opened 2017-05-19
Move dstr-binding for-await-of cases and templates to dstr-binding-async-iteration. Fixes gh-1043

rwaldron:


src/dstr-binding-async-iteration/ary-ptrn-elision-step-err.case
```diff
@@ -0,0 +1,37 @@
...
+       b. If next is an abrupt completion, set iteratorRecord.[[done]] to true.
+       c. ReturnIfAbrupt(next).
+features: [generator]
+---*/
+
+//- setup
+var following = 0;
+var iter =function* () {
+  throw new Test262Error();
+  following += 1;
```
leobalter: this is a small issue now that we can't verify the value of `following` in the given error/* templates.

My suggestion is to change the templates to support assertions in the rejection body, so a file in the `error` folder, like `error/for-await-of-async-func-const.template` would end with:


```js
fn()
  .then(_ => {
    throw new Test262Error("Expected async function to reject, but resolved.");
  }, ({ constructor }) => {
    assert.sameValue(constructor, /*{ error }*/);
    /*{ rejectBody }*/
  })
  .then($DONE, $DONE);
```

and you can finally add to this case file:

```js
//- rejectBody
assert.sameValue(following, 0, "iterator is properly closed");
```

src/dstr-binding-async-iteration/ary-init-iter-close.case
```diff
@@ -0,0 +1,37 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
```
leobalter: 2017 (same for the other case files)

src/dstr-binding-async-iteration/ary-init-iter-close.case
```diff
@@ -0,0 +1,37 @@
...
+    4. If iteratorRecord.[[done]] is false, return ? IteratorClose(iterator,
+       result).
+    [...]
+features: [Symbol.iterator]
+---*/
+
+//- setup
+var doneCallCount = 0;
+var iter = {};
+iter[Symbol.iterator] = function() {
```
leobalter: we should add coverage for objects with a Symbol.asyncIterator. This can be fixed in a follow up PR.

src/dstr-binding-async-iteration/ary-init-iter-close.case
rwaldron: They are just copies...

src/dstr-binding-async-iteration/ary-ptrn-elision-step-err.case
rwaldron: I just went through all of the cases and this occurs one time only (right here).

src/dstr-binding-for-await/ary-ptrn-elision-step-err.case
```diff
@@ -0,0 +1,39 @@
...
+  following += 1;
+}();
+//- elems
+[,]
+//- vals
+iter
+//- error
+Test262Error
+//- rejectBody
+assert.sameValue(following, 0, "iterator is properly closed");
```
leobalter: 👍

PR #1045, opened 2017-05-20
Adding case to computed property

### LINKED ISSUE #1035: Computed property missing in Object Rest Destructuring tests
We are missing the following test case for Object Rest Destructuring:

```
let a = "foo";
let {[a]: b, ...rest} = { foo: 2, bar: 3};
```

This test case is broken in JSC implementation.

caiolima:
I'm adding a missing case using computed properties.

It closes #1035.

### DIFF
--- a/src/dstr-assignment/obj-rest-computed-property.case
+++ b/src/dstr-assignment/obj-rest-computed-property.case
@@ -1,0 +1,38 @@
+// Copyright (C) 2017 Caio Lima. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: >
+    Destructuring field can be a computed property, i.e it can be defined 
+    only at runtime. Rest operantion needs to skip these properties as well.
+template: default
+esid: pending
+includes: [propertyHelper.js]
+features: [object-rest]
+---*/
+
+//- setup
+var a = "foo";
+//- elems
+{[a]:b, ...rest}
+//- vals
+{ foo: 1, bar: 2, baz: 3 }
+//- body
+assert.sameValue(b, 1);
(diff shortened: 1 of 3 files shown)


src/dstr-assignment/obj-rest-computed-property.case
```diff
@@ -0,0 +1,32 @@
...
+//- vals
+{ foo: 1, bar: 2, baz: 3 }
+//- body
+assert.sameValue(b, 1);
+assert.sameValue(rest.bar, 2);
+assert.sameValue(rest.baz, 3);
+
+verifyEnumerable(rest, "bar");
+verifyWritable(rest, "bar");
+verifyConfigurable(rest, "bar");
```
leobalter: you can use the new `verifyProperty`:

```js
verifyProperty(rest, "bar", {
  enumerable: true,
  writable: true,
  configurable: true
});
```

It would also be nice to verify `rest` has not property `foo`.

Out of curiosity: do we have similar cases for dstr-binding too? A `let {[a]:b, ...rest} = vals` would be nice. I need to check this.

caiolima: Cool. That's fair. The problem with binding is the scope cases in the templates. One example of case that will be generated but it isn't valid is:

```
let a = "foo"

function b({[a]: c, ...r}) {
//...
}

b({foo: 2, bar: 3});
```
Do you have any workaround in mind? I could add "let" cases manually, BTW.

caiolima:
Done =)

caiolima:
I think the failure here is because of generated versions missing. I'm missing something, but don't know what =).

caiolima:
Thanks for the review! @leobalter.

PR #1059, opened 2017-06-10
{Array,%TypedArray%}.prototype.sort should throw on a non-undefined non-function

ljharb:
Per TC39 consensus on  https://github.com/tc39/ecma262/pull/785/


leobalter:
These tests are great.

Too bad the specs doesn't allow us to observe the order of TypeError if the `this` value is not Object.

We can also follow up (in another PR) with extra tests using a poisoned `get length` operation confirming the order of this TypeError:

```js
var obj = {
  get length() {
    throw new Test262Error('wrong verification order');
  }
};
var sort = Array.prototype.sort;

  sort.call(obj, null);
});
```

PR #1060, opened 2017-06-11
Lint test `features` tag

jugglinmike:
I noticed some recent changes related to correcting the "features" flag, so I
thought this would be a useful application for the linting tool. Hope it helps!

Commit message:

> A recent commit introduced a document that enumerated acceptable values
> for the test "features" metadata tag. However, this list was incomplete,
> and maintaining it placed extra burden on the project owners.
>
> Restructure the document into a machine-readable format. Add entries for
> all previously-omitted values. Add in-line documentation with
> recommendations for maintenance of the file. Extend the project's
> linting tool to validate tests according to the document's contents.


tools/lint/test/fixtures/features_unrecognized.js
```diff
@@ -0,0 +1,11 @@
...
+^ expected errors | v input
+// Copyright (C) 2017 Mike Pennisi. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-assignment-operators-static-semantics-early-errors
+description: Minimal test
+features: [not-a-valid-feature]
+---*/
+
+// empty
```
leobalter: you missed an opportunity to draw an ASCII art with semicolons here.

features.txt
```diff
@@ -0,0 +1,99 @@
...
+Map
+new.target
+Proxy
+Reflect
+Reflect.construct
+Reflect.set
+Reflect.setPrototypeOf
+Set
+String#endsWith
+String#includes
```
leobalter: I'm surprised we are using these values.

This is a nice reminder to fix and avoid values with `#`, or at least be consistent with the other `.prototype` properties.

jugglinmike:
Happy to help :)

PR #1061, opened 2017-06-12
Remove src/dstr-assignment/default/for-await-of.template (invalid)

rwaldron:
Will need to revisit with function and generator versions


rwaldron:
👍🏽 thanks @leobalter

PR #1063, opened 2017-06-19
Add test that noctal escapes are forbidden in strings in strict code

### LINKED ISSUE #1062: Add negative tests for legacy octal escape sequences in strings
[Annex B.1.2](https://tc39.github.io/ecma262/#sec-additional-syntax-string-literals) defines legacy octal escape sequences in string literals. From what I can tell, test262 doesn't have coverage of these, either for their acceptance in sloppy mode code nor for them being disallowed in strict mode code (as [Section 11.8.4](https://tc39.github.io/ecma262/#sec-literals-string-literals) requires).

Thanks to the reporter of [this v8 issue](https://bugs.chromium.org/p/v8/issues/detail?id=6504) for bringing this to my attention.

bakkot:
Fixes #1062.

### DIFF
--- a/test/language/literals/string/legacy-non-octal-escape-sequence-strict.js
+++ b/test/language/literals/string/legacy-non-octal-escape-sequence-strict.js
@@ -1,0 +1,36 @@
+// Copyright (C) 2017 Kevin Gibbons. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-additional-syntax-string-literals
+es6id: B1.2
+description: >
+    LegacyOctalEscapeSequence is not enabled in strict mode code (regardless of
+    the presence of Annex B)
+info: >
+    EscapeSequence ::
+      CharacterEscapeSequence
+      LegacyOctalEscapeSequence
+      HexEscapeSequence
+      UnicodeEscapeSequence
+
+    LegacyOctalEscapeSequence ::
+      OctalDigit [lookahead ∉ OctalDigit]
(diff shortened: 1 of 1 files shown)


test/language/literals/string/legacy-non-octal-escape-sequence-strict.js
```diff
@@ -0,0 +1,36 @@
...
+
+    This definition of EscapeSequence is not used in strict mode or when
+    parsing TemplateCharacter.
+flags: [onlyStrict]
+negative:
+  phase: early
+  type: SyntaxError
+---*/
+
+'\08';
```
littledan: This is a very particular case, banned because `\0` has a no decimal digit lookahead, in combination with how it would've been handled as a legacy octal literal because 8 isn't an octal digit (right?). Maybe it'd be worth adding another test that's a little more normal, such as `\12`.

bakkot: @littledan Right. But the more normal cases are already covered [here](https://github.com/tc39/test262/blob/0314c87b0816e78802533ba9b30d42fb30eebf19/test/language/literals/string/legacy-octal-escape-sequence-strict.js) and [here](https://github.com/tc39/test262/blob/0314c87b0816e78802533ba9b30d42fb30eebf19/test/language/literals/string/legacy-octal-escape-sequence-strict.js). Or was there something not covered by those two + this one you thought ought to be tested?

littledan: Oh sorry, LGTM.

PR #1064, opened 2017-06-20
Add initial set of class field tests (from V8's test-parsing.cc)

xanlpz:
At this point this requires Babel for testing, but most tests in this
commit won't pass until this is fixed:
https://github.com/babel/babel/issues/5056

### DIFF
--- a/src/class-fields/computed-name-empty-var-computed-name-empty-function.case
+++ b/src/class-fields/computed-name-empty-var-computed-name-empty-function.case
@@ -1,0 +1,29 @@
+// Copyright (C) 2016 the V8 project authors, 2017 Igalia S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: Computed name var not initialized, computed name function
+info: |
+  1.1 New Productions
+
+  [...]
+
+  FieldDefinitionList [Yield, Await]:
+    FieldDefinition [?Yield, ?Await]
+    FieldDefinitionList  [?Yield, ?Await], FieldDefinition [?Yield, ?Await]
+template: default
+includes: [propertyHelper.js]
+---*/
+
+//- body
+  ['a']; ['b'](){}
+
+//- assertions
+  assert.sameValue(this.a, undefined);
+  verifyEnumerable(this, "a");
+  verifyWritable(this, "a");
+  verifyConfigurable(this, "a");
+
+  assert.sameValue(typeof Object.getPrototypeOf(this).b, "function");
+  verifyNotEnumerable(Object.getPrototypeOf(this), "b");
+  verifyConfigurable(Object.getPrototypeOf(this), "b");
--- a/src/class-fields/computed-name-empty-var-empty-function.case
+++ b/src/class-fields/computed-name-empty-var-empty-function.case
@@ -1,0 +1,29 @@
+// Copyright (C) 2016 the V8 project authors, 2017 Igalia S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: Computed name var not initialized, empty function
+info: |
+  1.1 New Productions
+
+  [...]
+
+  FieldDefinitionList [Yield, Await]:
+    FieldDefinition [?Yield, ?Await]
+    FieldDefinitionList  [?Yield, ?Await], FieldDefinition [?Yield, ?Await]
+template: default
+includes: [propertyHelper.js]
+---*/
+
+//- body
+  ['a']; b(){}
+
+//- assertions
+  assert.sameValue(this.a, undefined);
+  verifyEnumerable(this, "a");
+  verifyWritable(this, "a");
+  verifyConfigurable(this, "a");
+
+  assert.sameValue(typeof Object.getPrototypeOf(this).b, "function");
+  verifyNotEnumerable(Object.getPrototypeOf(this), "b");
+  verifyConfigurable(Object.getPrototypeOf(this), "b");
--- a/src/class-fields/computed-name-empty-var-new-line-asi.case
+++ b/src/class-fields/computed-name-empty-var-new-line-asi.case
@@ -1,0 +1,25 @@
+// Copyright (C) 2016 the V8 project authors, 2017 Igalia S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: computed name empty var and newline (ASI)
+info: |
+  1.1 New Productions
+
+  [...]
+
+  FieldDefinitionList [Yield, Await]:
+    FieldDefinition [?Yield, ?Await]
+    FieldDefinitionList  [?Yield, ?Await], FieldDefinition [?Yield, ?Await]
+template: default
+includes: [propertyHelper.js]
+---*/
+
+//- body
+  ['a']
+
+//- assertions
+  assert.sameValue(this.a, undefined);
+  verifyEnumerable(this, "a");
+  verifyWritable(this, "a");
+  verifyConfigurable(this, "a");
--- a/src/class-fields/computed-name-empty-var-new-line-computed-name-empty-function-asi.case
+++ b/src/class-fields/computed-name-empty-var-new-line-computed-name-empty-function-asi.case
@@ -1,0 +1,29 @@
+// Copyright (C) 2016 the V8 project authors, 2017 Igalia S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: computed name empty var and newline, computed name empty function (ASI)
+info: |
+  1.1 New Productions
+
+  [...]
+
+  FieldDefinitionList [Yield, Await]:
+    FieldDefinition [?Yield, ?Await]
+    FieldDefinitionList  [?Yield, ?Await], FieldDefinition [?Yield, ?Await]
+template: default
+includes: [propertyHelper.js]
+---*/
+
+//- body
+  ['a']
+  ['b'](){}
+//- assertions
+  assert.sameValue(this.a, undefined);
+  verifyEnumerable(this, "a");
+  verifyWritable(this, "a");
+  verifyConfigurable(this, "a");
+
+  assert.sameValue(typeof Object.getPrototypeOf(this).b, "function");
+  verifyNotEnumerable(Object.getPrototypeOf(this), "b");
+  verifyConfigurable(Object.getPrototypeOf(this), "b");
--- a/src/class-fields/double-empty-var.case
+++ b/src/class-fields/double-empty-var.case
@@ -1,0 +1,30 @@
+// Copyright (C) 2016 the V8 project authors, 2017 Igalia S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: Empty var, empty var
+info: |
+  1.1 New Productions
+
+  [...]
+
+  FieldDefinitionList [Yield, Await]:
+    FieldDefinition [?Yield, ?Await]
+    FieldDefinitionList  [?Yield, ?Await], FieldDefinition [?Yield, ?Await]
+template: default
+includes: [propertyHelper.js]
+---*/
+
+//- body
+  a; b;
+
+//- assertions
+  assert.sameValue(this.a, undefined);
+  verifyEnumerable(this, "a");
+  verifyWritable(this, "a");
+  verifyConfigurable(this, "a");
+
+  assert.sameValue(this.b, undefined);
+  verifyEnumerable(this, "b");
+  verifyWritable(this, "b");
+  verifyConfigurable(this, "b");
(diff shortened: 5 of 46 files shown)


src/class-fields/empty-string-var.case
```diff
@@ -0,0 +1,25 @@
...
+
+  FieldDefinitionList [Yield, Await]:
+    FieldDefinition [?Yield, ?Await]
+    FieldDefinitionList  [?Yield, ?Await], FieldDefinition [?Yield, ?Await]
+template: default
+includes: [propertyHelper.js]
+---*/
+
+//- body
+  'a';
```
rwaldron: @littledan it occurred to me that this closes the door to directive prologues in class bodies (that might be intentional?)

src/class-fields/empty-var-computed-name-empty-function.case
```diff
@@ -0,0 +1,29 @@
...
+
+//- testfield
+  isEqualTo(this, "a", undefined);
+  verifyEnumerable(this, "a");
+  verifyWritable(this, "a");
+  verifyConfigurable(this, "a");
+
+  isEqualTo(this, "b", undefined);
+  verifyEnumerable(this, "b");
+  verifyConfigurable(this, "b");
```
rwaldron: L27-29 same issues as above

src/class-fields/empty-var-empty-function.case
```diff
@@ -0,0 +1,29 @@
...
+
+//- testfield
+  isEqualTo(this, "a", undefined);
+  verifyEnumerable(this, "a");
+  verifyWritable(this, "a");
+  verifyConfigurable(this, "a");
+
+  isEqualTo(this, "b", undefined);
+  verifyEnumerable(this, "b");
+  verifyConfigurable(this, "b");
```
rwaldron: L27-29 same issues as above

src/class-fields/empty-var-generator-empty-function.case
```diff
@@ -0,0 +1,29 @@
...
+
+//- testfield
+  isEqualTo(this, "a", undefined);
+  verifyEnumerable(this, "a");
+  verifyWritable(this, "a");
+  verifyConfigurable(this, "a");
+
+  isEqualTo(this, "b", undefined);
+  verifyEnumerable(this, "b");
+  verifyConfigurable(this, "b");
```
rwaldron: L27-29 same issues as above

src/class-fields/empty-var-new-line-computed-name-function-asi.case
```diff
@@ -0,0 +1,29 @@
...
+  ['b'](){}
+//- testfield
+  isEqualTo(this, "a", undefined);
+  verifyEnumerable(this, "a");
+  verifyWritable(this, "a");
+  verifyConfigurable(this, "a");
+
+  isEqualTo(this, "b", undefined);
+  verifyEnumerable(this, "b");
+  verifyConfigurable(this, "b");
```
rwaldron: L27-29 same issues as above

src/class-fields/empty-var-new-line-empty-function-asi.case
```diff
@@ -0,0 +1,29 @@
...
+  b(){}
+//- testfield
+  isEqualTo(this, "a", undefined);
+  verifyEnumerable(this, "a");
+  verifyWritable(this, "a");
+  verifyConfigurable(this, "a");
+
+  isEqualTo(this, "b", undefined);
+  verifyEnumerable(this, "b");
+  verifyConfigurable(this, "b");
```
rwaldron: L27-29 same issues as above

src/class-fields/empty-var-new-line-generator-empty-function-asi.case
```diff
@@ -0,0 +1,29 @@
...
+  *b(){}
+//- testfield
+  isEqualTo(this, "a", undefined);
+  verifyEnumerable(this, "a");
+  verifyWritable(this, "a");
+  verifyConfigurable(this, "a");
+
+  isEqualTo(this, "b", undefined);
+  verifyEnumerable(this, "b");
+  verifyConfigurable(this, "b");
```
rwaldron: L27-29 same issues as above

src/class-fields/yield-var.case
```diff
@@ -0,0 +1,25 @@
...
+
+  FieldDefinitionList [Yield, Await]:
+    FieldDefinition [?Yield, ?Await]
+    FieldDefinitionList  [?Yield, ?Await], FieldDefinition [?Yield, ?Await]
+template: default
+includes: [propertyHelper.js]
+---*/
+
+//- body
+  yield
```
rwaldron: The grammar has `[?Yield, ?Await]`, but doesn't appear to define what to do with those productions... I wonder if @littledan intends to define something similar to https://tc39.github.io/ecma262/#sec-identifiers-static-semantics-early-errors

src/class-fields/double-empty-var.case
```diff
@@ -0,0 +1,30 @@
...
+//- testfield
+  isEqualTo(this, "a", undefined);
+  verifyEnumerable(this, "a");
+  verifyWritable(this, "a");
+  verifyConfigurable(this, "a");
+
+  isEqualTo(this, "b", undefined);
+  verifyEnumerable(this, "b");
+  verifyWritable(this, "b");
+  verifyConfigurable(this, "b");
```
leobalter: yes, I have plans to improve verifyProperty for completeness. I'm lacking on time. For now, it's good for asserting enumerability, configurability and writability.

src/class-fields/computed-name-empty-var-new-line-computed-name-empty-function-asi.case
```diff
@@ -0,0 +1,29 @@
...
+  ['b'](){}
+//- testfield
+  isEqualTo(this, "a", undefined);
+  verifyEnumerable(this, "a");
+  verifyWritable(this, "a");
+  verifyConfigurable(this, "a");
+
+  isEqualTo(this, "b", undefined);
+  verifyEnumerable(this, "b");
+  verifyConfigurable(this, "b");
```
rwaldron: L27-29 same issues as above

src/class-fields/zero-initializer-var-computed-name-empty-function.case
```diff
@@ -0,0 +1,29 @@
...
+
+//- testfield
+  isEqualTo(this, "a", 0);
+  verifyEnumerable(this, "a");
+  verifyWritable(this, "a");
+  verifyConfigurable(this, "a");
+
+  isEqualTo(this, "b", undefined);
+  verifyEnumerable(this, "b");
+  verifyConfigurable(this, "b");
```
rwaldron: L27-29 same issues as above

src/class-fields/zero-initializer-var-empty-function.case
```diff
@@ -0,0 +1,29 @@
...
+
+//- testfield
+  isEqualTo(this, "a", 0);
+  verifyEnumerable(this, "a");
+  verifyWritable(this, "a");
+  verifyConfigurable(this, "a");
+
+  isEqualTo(this, "b", undefined);
+  verifyEnumerable(this, "b");
+  verifyConfigurable(this, "b");
```
rwaldron: L27-29 same issues as above

src/class-fields/zero-initializer-var-generator-empty-function.case
```diff
@@ -0,0 +1,29 @@
...
+
+//- testfield
+  isEqualTo(this, "a", 0);
+  verifyEnumerable(this, "a");
+  verifyWritable(this, "a");
+  verifyConfigurable(this, "a");
+
+  isEqualTo(this, "b", undefined);
+  verifyEnumerable(this, "b");
+  verifyConfigurable(this, "b");
```
rwaldron: L27-29 same issues as above

src/class-fields/zero-initializer-var-new-line-empty-function-asi.case
```diff
@@ -0,0 +1,30 @@
...
+
+//- testfield
+  isEqualTo(this, "a", 0);
+  verifyEnumerable(this, "a");
+  verifyWritable(this, "a");
+  verifyConfigurable(this, "a");
+
+  isEqualTo(this, "b", undefined);
+  verifyEnumerable(this, "b");
+  verifyConfigurable(this, "b");
```
rwaldron: L27-29 same issues as above

src/class-fields/computed-name-empty-var-empty-function.case
```diff
@@ -0,0 +1,29 @@
...
+//- body
+  ['a']; b(){}
+
+//- testfield
+  isEqualTo(this, "a", undefined);
+  verifyEnumerable(this, "a");
+  verifyWritable(this, "a");
+  verifyConfigurable(this, "a");
+
+  isEqualTo(this, "b", undefined);
```
leobalter: it would be nice to have before the body:

```js
let c = 'a';
```

and then check the computed property name based on the value of c:

```js
//- body
  [c]; b(){}
```

src/class-fields/computed-name-empty-var-computed-name-empty-function.case
```diff
@@ -0,0 +1,29 @@
...
+
+/*---
+desc: Computed name var not initialized, computed name function
+info: |
+  1.1 New Productions
+
+  [...]
+
+  FieldDefinitionList [Yield, Await]:
+    FieldDefinition [?Yield, ?Await]
```
leobalter: The complete grammar for FieldDefinition would be very helpful in these parts

src/class-fields/computed-name-empty-var-new-line-asi.case
```diff
@@ -0,0 +1,25 @@
...
+
+  FieldDefinitionList [Yield, Await]:
+    FieldDefinition [?Yield, ?Await]
+    FieldDefinitionList  [?Yield, ?Await], FieldDefinition [?Yield, ?Await]
+template: default
+includes: [propertyHelper.js]
+---*/
+
+//- body
+  ['a']
```
leobalter: I need to confirm here: https://github.com/tc39/proposal-class-fields/issues/7

If the semicolon is not optional, this - and a lot of other following cases - is a syntax error.

borela: It's not a syntax error, the test case is testing the ASI, look at the `-asi` suffix.

leobalter: yes, that's my bad. I've been working so long on nice written code that I've had a momentary chance to forget ASI exists.

borela: I understand, well formatted code makes semicolons completely unnecessary, makes me wonder why other languages don't have ASI too.

leobalter: I don't agree they are unnecessary but I believe it's not worth the debate here.

bakkot:
Could you add a test for

```js
class A {
  a = 0
  in
  b
}
```

parsing the same as

```js
class A {
  a = 0 in b;
}
```

and

```js
class A {
  a = 0;
  in
  b
}
```

parsing the same as

```js
class A {
  a = 0;
  in;
  b;
}
```
?

(I don't think this case is covered in v8's tests - I didn't put it in, at any rate.)

rwaldron:
@bakkot more tests can be added in future PRs—we don't expect full coverage of any feature in a single PR

xanlpz:
I have pushed a new version, but used --force by mistake! So I'll quickly list the changes:
* Using assert.sameValue everywhere instead of isEqualTo
* One indentation style only in templates (two spaces)
* Use "-assertions" for the assertion block
* Use the prototype of 'this' when verifying the properties of a method
* Check that the typeof methods is "function" instead of their value undefined
* Methods are NOT enumerable, so check that (this was not caught during review)
* Add one test where = undefined is added explicitly as "sentinel" that should work even with Babel right now.

Comments:
I have only added one "sentinel" to avoid a combinatorial explosion, not sure if I should add more.
Testing this with Babel I only get one type of failure related to https://github.com/babel/babel/issues/5056.

src/class-fields/empty-string-var.case
littledan: @rwaldron You mean that we can't have 'use strict' in a class body? I thought the idea was that we wouldn't add any more of those, with 1JS.

src/class-fields/yield-var.case
littledan: This example is totally fine; what those parameters control is what yield and await mean within a computed property name.

The `[?Yield, ?Await]` parameters are passed on through, where they eventually make it to an IdentifierReference. [here](https://tc39.github.io/ecma262/#prod-IdentifierReference), IdentifierReference includes `yield` and `await` only if the corresponding parameter is off.

src/class-fields/computed-name-empty-var-new-line-asi.case
littledan: Here, what's important is to have test coverage for both paths.

rwaldron:
@leobalter I'd like to land this to unblock additional tests I have written

leobalter:
no problem, I'll generate the files and merge this.

littledan:
These tests are at Stage 2; the PR was sent out to develop the tests, but it's not appropriate to merge them yet. Should we make a "Stage 2" label to mark these tests as not being mergeable yet? This comes up in the BigInt tests as well.

leobalter:
we need to restore this PR. Thanks for pulling the trigger too early.

Issue #1050, opened 2017-05-23
Remove tests that allow `...[` and `...{` in object destructuring


rwaldron:
per resolution of 16.i.h https://github.com/tc39/agendas/blob/master/2017/05.md

I'll work on this during my next cycle

leobalter:
by object destructuring, this means rest operators (not spread) directly in object patterns for destructuring will only be valid with name identifiers.

```js
let {a, ...[]} = {/*...*/}; // invalid
let {a, ...[b]} = {/*...*/}; // invalid
let {a, ...{}} = {/*...*/}; // invalid
let {a, ...{b}} = {/*...*/}; // invalid
let {a, ...b} = {/*...*/}; // valid
let {...[]} = {/*...*/}; // valid
```

This needs to wait the respective proposal to be fixed and match the references.

rwaldron:
> `let {...[]} = {/*...*/}; // valid`

I don't think that's correct. We need confirmation

> This needs to wait the respective proposal to be fixed and match the references.

Everything I've reviewed so far has `esid: pending`

leobalter:
> I don't think that's correct. We need confirmation 

right you are, my copy-pasta fault.

> Everything I've reviewed so far has esid: pending

pending means only we don't have the reference on what the link would be in the specs yet. Still, we require a minimum spec text to reference each case. Of course, providing this in the info tag helps a lot.

leobalter:
A few examples of test cases requiring a new review:

src/dstr-binding/obj-ptrn-rest-nested-obj.case
src/dstr-binding/obj-ptrn-rest-obj-nested-rest.case
src/dstr-binding/obj-ptrn-rest-obj-own-property.case

rwaldron:
> pending means only we don't have the reference on what the link would be in the specs yet

I know, but all the cases I've seen said "esid: pending", so there is no reason to...

> This needs to wait the respective proposal to be fixed and match the references.

...

> A few examples of test cases requiring a new review:
> src/dstr-binding/obj-ptrn-rest-nested-obj.case
> src/dstr-binding/obj-ptrn-rest-obj-nested-rest.case
> src/dstr-binding/obj-ptrn-rest-obj-own-property.case

Yes, already removed in a branch.

rwaldron:
I'll also include new negative syntax error tests.

gsathya:
FYI, these are the obsolete tests that I had to disable in V8:
[...]

PR #1072, opened 2017-06-23
Remove tests that allow `...{` in object destructuring. Ref gh-1050

rwaldron:


src/dstr-binding/obj-ptrn-rest-obj-own-property.case
```diff
@@ -1,24 +0,0 @@
-// Copyright (C) 2017 Caio Lima. All rights reserved.
-// This code is governed by the BSD license found in the LICENSE file.
-
-/*---
-desc: >
-    Rest object contains just soruce object's own properties
```
littledan: Is this property tested some other way? It's important to make sure that only own properties are included, and it would be easy to test this without using nested patterns like this (or templating).

src/dstr-binding/obj-ptrn-rest-nested-obj.case
```diff
@@ -1,25 +0,0 @@
-// Copyright (C) 2017 Caio Lima. All rights reserved.
-// This code is governed by the BSD license found in the LICENSE file.
-
-/*---
-desc: >
-    When DestructuringAssignmentTarget is an object literal, it should be parsed
-    parsed as a DestructuringAssignmentPattern and evaluated as a destructuring
-    assignment.
```
littledan: It's a little late for me to be asking now, but: Why is this desc talking about destructuring assignment? It doesn't look like it has to do with that; the test is all about destructuring binding.

leobalter:
pinging @caiolima wrt the questions.

the changes LGTM

Issue #878, opened 2017-03-02
Error phase: "initialization"


leobalter:
Following https://github.com/tc39/test262/pull/872#issuecomment-283405112

(...)

---

> @littledan: Hmm, if these are moved to a later phase, it might be good to somehow test that the error comes earlier than when the lines of code are executed. IIRC There was logic for this that @jugglinmike removed in the phase marking patch.

---

> @bakkot: @littledan, good point. For non-raw "phase: early" tests I've just been prepending throw new Test262Error and asserting that the actual error is not a Test262Error, which I believe is what the old logic was.

> I'd support either having an "phase: initialization" or adding throw new Test262Error as the first line of tests which throw as runtime semantics but before executing any line of code. I'm ambivalent between the two. Thoughts?

jugglinmike:
> @littledan: Hmm, if these are moved to a later phase, it might be good to
> somehow test that the error comes earlier than when the lines of code are
> executed. IIRC There was logic for this that @jugglinmike removed in the
> phase marking patch.

That was indeed a pattern prior to my patch, but it was applied very
inconsistently. With the meta-data in place, an automated test harness is able
to programatically make this transformation, much in the same way it is
currently expected to conditionally insert a global "use strict" directive,
just as @bakkot described.  From [the INTERPRETING.md
document](https://github.com/tc39/test262/blob/975e54de17ede3d8802dd7411181dbbc29147308/INTERPRETING.md):

> Consumers are free to assert the "early" phase as they see fit.
>
> For example, it is possible to insert a `throw` statement with a unique error
> type at the beginning of the test file. In this case, the statement should be
> inserted *after* the directive desribed in the section titled "Strict Mode"
> (where appropriate), though it must *not* be inserted for tests containing
> the "raw" flag.

littledan:
Well, I don't know if any of this matters from a test262 perspective, but I think 'initialization' and 'early' are a bit different from an implementation perspective, at least for how V8 works. 'early' errors can be checked for just by running the parser, and not the resulting compiler/bytecode generator. It's especially useful to mark them, since it lets us run both the parser and the preparser separately and ensure that they both reach the error. This is a stronger test than the previous idiom, so I'm happy that test262 has this metadata now. I believe most JavaScript implementations have this architecture, but I could be mistaken.

[...]

PR #778, opened 2016-10-19
Encode expected error "phase"

jugglinmike:
I have rebased the patch previously submitted at gh-655. Please refer to that
pull request for motivation, design discussion, and suggestions for merging
strategy. This version applies the patterns established there; it contains no
substantial changes other than updating the error format of recently-introduced
tests.

I have verified the validity of the change set by running the tests in a recent
build of V8, first on Test262's `master` branch and then on this branch. The
reported errors are almost identical; this factoring happens to catch a handful
of additional violations in that runtime:

- language/literals/regexp/invalid-braced-quantifier-exact.js
- language/literals/regexp/invalid-braced-quantifier-lower.js
- language/literals/regexp/invalid-braced-quantifier-range.js
- language/literals/regexp/u-dec-esc.js
- language/literals/regexp/u-invalid-class-escape.js
- language/literals/regexp/u-invalid-extended-pattern-char.js
- language/literals/regexp/u-invalid-identity-escape.js
- language/literals/regexp/u-invalid-legacy-octal-escape.js
- language/literals/regexp/u-invalid-non-empty-class-ranges-no-dash-a.js
- language/literals/regexp/u-invalid-non-empty-class-ranges-no-dash-ab.js
- language/literals/regexp/u-invalid-non-empty-class-ranges-no-dash-b.js
- language/literals/regexp/u-invalid-non-empty-class-ranges.js
- language/literals/regexp/u-invalid-oob-decimal-escape.js
- language/literals/regexp/u-invalid-quantifiable-assertion.js
- language/literals/regexp/u-unicode-esc-bounds.js
- language/literals/regexp/u-unicode-esc-non-hex.js

...all of these owe to the same basic flaw, which is well-known to the V8 team:
https://bugs.chromium.org/p/v8/issues/detail?id=896

@tcare @leobalter I would like to avoid having to rebase this work a third
time--if at all possible, can you please prioritize review? Thank you!

PR #1073, opened 2017-06-25
Require `throw` statement in tests for early errors

jugglinmike:
In gh-778, we began tracking tests for early errors with metadata. Prior to
that patch, some tests for early errors began with a `throw` statement. The
intent of this statement was to ensure that program code was not executed.

With the introduction of the metadata, we elected to remove that pattern in
favor of a recommendation for test consumers to dynamically insert such a
statement immediately prior to test execution [1].

However, this approach increases the burden for test consumers, and because it
makes the test material less literal, it also tends to obscure the expected
behavior (see the discussion at gh-878 for more information).

With the introduction of a static analysis tool [2], the project is now in a
position to automatically enforce a requirement for an explicit `throw`
statement. This change set introduces such a requirement and updates all tests
to satisfy it. In so doing, it makes expectations more transparent both for
contributors and consumers, at the expense of a small redundancy in test
content (i.e. tests for early errors must now include dedicated metadata and a
`throw` statement).

The linting heuristic is admittedly a little weak--it only verifies that the
source text contains text for a `throw` statement, not that it is the first
statement in the file (or even that it is a statement at all). Anything more
than this would significantly increase the linter's complexity, but I don't
think the relative benefit warrants this. For now, at least, it seems fair to
expect maintainers to recognize those extenuating circumstances that would fool
the heuristic.

[1] Commit ade6d2e384350055e1e23c35cfbf0d4dab03c82e
[2] Commit 74954bfa915136c7998f8a0d9ca40b885075d409


littledan:
This seems like a reasonable approach to me; great to see your attention to detail here.

rwaldron:
I wanted to just take @littledan's word for it, but I had to spend some time catching up on the cited patches to make sure the I fully follow this change.

CONTRIBUTING.md
```diff
@@ -187,7 +187,7 @@ function $DONE(arg) {
 
 ## Handling Errors and Negative Test Cases
 
-Expectations for **parsing errors** should be declared using [the `negative` frontmatter flag](#negative):
+Expectations for **parsing errors** should be declared using [the `negative` frontmatter flag](#negative). They must also include the exact `throw` statement specified in this example (in order to guarantee that implementations do not execute the code):
```
rwaldron: I wonder how we can call more attention to this—maybe some bold text? What does everyone else think? Maybe it's fine as it is?

leobalter:
I would say this should be responsibility of the test runner, prepending early error tests with a throw statement.

This change is already affecting 2000+ files and I wonder if this will raise the cost of reviews: "please add a throw statement before the failing code".

On the other side: there might a lot of test runners we don't have control and can't secure they will implement this approach. I would say this diplomatic communication would be easier.

This documentation can suggest test runners should prepend a throw statement on early error tests. 

What do you think, @jugglinmike, @rwaldron, @littledan?

cc'ing @anba and @bterlson wrt SpiderMonkey and Chakra as well.

rwaldron:
>  this will raise the cost of reviews: "please add a throw statement before the failing code".

The additional cost is minimal, because the work is done in CI. 

> there might a lot of test runners we don't have control and can't secure they will implement this approach.

This is the problem that @jugglinmike is solving directly.

tools/lint/lib/checks/negative.py
```diff
@@ -0,0 +1,27 @@
...
+            return '"negative" must be a dictionary with fields "type" and "phase"'
+
+        if not 'type' in negative:
+            return '"negative" must specify a "type" field'
+
+        if not 'phase' in negative:
+            return '"negative" must specify a "phase" field'
+
+        if negative["phase"] == "early" and not _THROW_STMT.search(source):
+            return 'Negative tests of type "early" must include a `throw` statement'
```
leobalter: oh, I finally got this! Thank you, @jugglinmike! 👑 👑 👑

CONTRIBUTING.md
leobalter: I think it's fine. What I think we should improve eventually a way for contributors to use the linter before the CI running over the PR. Something that puts it more visible or clear.

rwaldron:
👍

bterlson:
FWIW I also agree that this should be a harness requirement. There are a number of requirements that test runners have in order to fully test ECMAScript. This change feels to me like checking in separate files for strict, strict module, and non-strict mode.

leobalter:
@bterlson this PoV is important as I don't consume Test262 (as an implementor, etc). @jugglinmike uses it on JSHint, so does the other ppl I pinged here. 

I believe if we work on this requirement of runners prepending a throw statement, we would be able to get rid of this manual insertion. For now, I like this is at least checked through the linter.

PR #1075, opened 2017-06-27
Ensure `apply` and `construct` pass all parametes

shvaikalesh:
Follow-up of #792


test/built-ins/Proxy/apply/trap-is-undefined.js
```diff
@@ -2,20 +2,30 @@
...
+let handlers = [
+    {},
+    {apply: undefined},
+    {apply: null},
+];
+
+for (let handler of handlers) {
+    let p = new Proxy(target, handler);
+    let res = Reflect.apply(p, ctx, [1, 2]);
+    assert.sameValue(res, 3);
```
leobalter: I would rather have each test separated. If not on the same file, at least on different lines, not in a loop. DRY does not help much here, and an assertion failure can't tell what part is failing.

test/built-ins/Proxy/construct/call-parametes-new-target.js
```diff
@@ -0,0 +1,36 @@
...
+let Target = function() {};
+let NewTarget = function() {};
+
+let handler = {
+    construct: function(_Target, args, _NewTarget) {
+        assert.sameValue(this, handler, "trap context is the handler object");
+        assert.sameValue(_Target, Target, "first parameter is the target object");
+        assert.sameValue(args.length, 2, "arguments list contains all construct arguments");
+
+        let [a, b] = args;
```
leobalter: considering these are tests for ES6's Proxy, I'm not sure if we should use too many syntax from the same spec version.  `let` and `destructuring`, could be replaced with "classic" code to avoid a race condition for an implementation in this case.

maybe @rwaldron has a different opinion.

test/built-ins/Proxy/construct/call-parametes-new-target.js
```diff
@@ -0,0 +1,36 @@
...
+        assert.sameValue(a, 1, "arguments list has first construct argument");
+        assert.sameValue(b, 2, "arguments list has second construct argument");
+        assert.sameValue(_NewTarget, NewTarget, "newTarget is sent as the third parameter");
+
+        return {sum: a + b};
+    },
+};
+
+let P = new Proxy(Target, handler);
+let res = Reflect.construct(P, [1, 2], NewTarget);
```
leobalter: Can we add a `features: [Reflect.construct]` in the frontmatter?

test/built-ins/Proxy/construct/trap-is-undefined.js
```diff
@@ -2,20 +2,30 @@
...
+    {},
+    {construct: undefined},
+    {construct: null},
+];
 
-assert.sameValue((new P("foo")).attr, "foo");
+for (let handler of handlers) {
+    let P = new Proxy(Target, handler);
+    let res = Reflect.construct(P, [1, 2], NewTarget);
+    assert.sameValue(res.sum, 3);
```
leobalter: same issue with the for loop, plus we should have an assertion message on these.

test/built-ins/Proxy/apply/trap-is-undefined-no-property.js
```diff
@@ -1,19 +0,0 @@
...
-    7. If trap is undefined, then Return Call(target, thisArgument,
-    argumentsList).
----*/
-
-var target = function(a, b) {
-    return a + b;
-};
-var p = new Proxy(target, {});
-
-assert.sameValue(p(1, 2), 3);
```
rwaldron: Please restore this test file

rwaldron:
The file named "call-parametes-new-target.js" has a typo and should be "call-parameters-new-target.js"

test/built-ins/Proxy/construct/call-parametes-new-target.js
rwaldron: I'm of the opinion that ES6 features should be tested with <= ES5, ES7 with <= ES6, ES8 with <= ES7.

shvaikalesh:
Thank you guys for the thorough review. I hope you like the way `trap == null` tests are handled: after this is merged I would like to add `trap === null` tests for other traps as well.

test/built-ins/Proxy/construct/call-parametes-new-target.js
shvaikalesh: perhaps we should include it somewhere in `CONTRIBUTING.md`? some new tests are written in es6 and one may think it is because the files are new, not because they test newer features.

test/built-ins/Proxy/apply/trap-is-missing.js
```diff
@@ -2,18 +2,22 @@
...
     return a + b;
-};
-var p = new Proxy(target, {});
+}
 
-assert.sameValue(p(1, 2), 3);
+var ctx = {};
+var p = new Proxy(target, {});
+var res = p.call(ctx, 1, 2);
+assert.sameValue(res, 3, "`apply` trap is missing");
```
rwaldron: Instead of modifying the existing test file (which tests the implicit `apply` trap path), this should be added as a new test file.

test/built-ins/Proxy/apply/trap-is-null-or-undefined.js
```diff
@@ -0,0 +1,27 @@
...
+}
+
+var ctx = {};
+var p = new Proxy(target, {apply: null});
+var res = p.call(ctx, 1, 2);
+assert.sameValue(res, 3, "`apply` trap is `null`");
+
+p = new Proxy(target, {apply: undefined});
+res = p.call(ctx, 3, 4);
+assert.sameValue(res, 7, "`apply` trap is `undefined`");
```
rwaldron: The existing test keyed on a precise step in the spec; that file should be left as-is and an additional new file for the  "`apply` trap is `null`" test should be created.

test/built-ins/Proxy/construct/call-parameters-new-target.js
```diff
@@ -0,0 +1,38 @@
...
+
+    9. Let newObj be Call(trap, handler, «target, argArray, newTarget»).
+features: [Reflect.construct]
+---*/
+
+function Target() {}
+function NewTarget() {}
+
+var handler = {
+    construct: function(_Target, args, _NewTarget) {
```
rwaldron: nit: why not name these `target, argArray, newTarget`?

test/built-ins/Proxy/construct/trap-is-null-or-undefined.js
```diff
@@ -0,0 +1,28 @@
...
+    return {sum: a + b};
+}
+
+var P = new Proxy(Target, {construct: null});
+var obj = Reflect.construct(P, [3, 4], NewTarget);
+assert.sameValue(obj.sum, 7, "`construct` trap is `null`");
+
+P = new Proxy(Target, {construct: undefined});
+obj = Reflect.construct(P, [5, 6], NewTarget);
+assert.sameValue(obj.sum, 11, "`construct` trap is `undefined`");
```
rwaldron: Same request as above, restore the original test and add the "`construct` trap is `null`" test in a new file

rwaldron:
For added explanation of my latest requests: the existing test files have matching tests in every Proxy trap test directory (w/ exception of enumerate and revocable) and it would be ideal to keep that intact

leobalter:
@rwaldron are you ok if I just merge this and immediately work in a follow up addressing your review parts?

leobalter:
solved at #1213

PR #1077, opened 2017-06-27
update Array/From files ESID to sec-array.from

deathbearbrown:
As the first part of a project to create an interface for exploring tests, we are updating all tests so that they consistently map to an ESID.

https://tc39.github.io/ecma262/#sec-array.from


leobalter:
Merging after the CI

PR #1079, opened 2017-06-28
Add tests for proposal to make Intl prototypes not instances

littledan:
This proposal was started in https://github.com/tc39/ecma402/issues/122,
and specified in https://github.com/tc39/ecma402/pull/148


rwaldron:
@ericf ping

littledan:
The PR was merged, so if someone could review the tests, that would be great.

Issue #1082, opened 2017-06-28
Generated destructuring assignment rest test only valid in sloppy mode


littledan:
The test in test/language/statements/for-of/dstr-obj-rest-computed-property.js writes to an undeclared variable `b`. In strict mode, a ReferenceError will be thrown for this reason. The test needs to have an \[onlySloppy] tag, added to the appropriate template.

rwaldron:
I'll fix this asap

rwaldron:
Also, the `rest` variable.

PR #1083, opened 2017-06-28
Duplicate src/dstr-assignment/obj-rest-computed-property.case into regular and sloppy versions. Fixes gh-1082

rwaldron:


littledan:
LGTM, thanks for the quick fix.

PR #1084, opened 2017-06-28
Add ASI let/yield test

littledan:
I had previously intended to add this test in
https://github.com/tc39/test262/pull/968 ,
but it seems like I instead checked in a copy of another test. This
patch changes it to test what was initially intended, which is part of
the semantics for https://github.com/tc39/ecma262/pull/885


rwaldron:
@littledan ping for rebase

leobalter:
I just used GitHub's tool to resolve the conflict.

PR #1093, opened 2017-06-29
add esid to array/protoype

deathbearbrown:
Just one file!


test/built-ins/Array/prototype/Symbol.iterator.js
```diff
@@ -12,6 +12,7 @@ info: >
 includes: [propertyHelper.js]
 features: [Symbol.iterator]
 es6id: 22.1.3.30
+esid: sec-array.prototype.values
```
rwaldron: sec-array.prototype-@@iterator

rwaldron: sec-array.prototype-@@iterator

deathbearbrown:
@rwaldron  made that change

PR #1139, opened 2017-07-05
Update RegExp Unicode property escape tests per latest proposal & Unicode 10

mathiasbynens:
https://github.com/tc39/proposal-regexp-unicode-property-escapes/commit/c533fa914799f94b8e66b95d2fde65f2c5b9fcf2


rwaldron:
@mathiasbynens you're so prolific <3

@littledan thanks for reviewing this!

PR #1143, opened 2017-07-05
update pending esids in AsyncFunction tests

deathbearbrown:
@leobalter @rwaldron 
This is a case of where there was a "pending" esid that needed to be updated.  I tried to match as best I could to the docs. The only thing I couldn't match very well was for the test: "is not a global"


rwaldron:
Awesome!

rwaldron:
In another PR, you will need to also update `test/language/expressions/async-function/*.js`

PR #1145, opened 2017-07-10
Improve RegExp.prototype.source coverage

shvaikalesh:


test/built-ins/RegExp/prototype/source/prop-desc.js
```diff
@@ -2,16 +2,19 @@
...
-  
+var d = Object.getOwnPropertyDescriptor(RegExp.prototype, 'source');
 
 assert.sameValue(typeof d.get, 'function', 'typeof d.get');
 assert.sameValue(d.set, undefined, 'd.set');
-assert.sameValue(d.enumerable, false, 'd.enumerable');
-assert.sameValue(d.configurable, true, 'd.configurable');
+
+verifyNotEnumerable(RegExp.prototype, 'source');
+verifyConfigurable(RegExp.prototype, 'source');
```
leobalter: why not using `verifyProperty` here? For accessor descriptors, you can do this:

```js
verifyProperty(RegExp.prototype, 'source', {
  enumerable: false,
  configurable: true,
});
```

There's no support to compare `get` and `set` yet but the other parts can already use it.

test/built-ins/RegExp/prototype/source/this-val-regexp-prototype.js
```diff
@@ -3,15 +3,12 @@
...
   1. Let R be the this value.
   2. If Type(R) is not Object, throw a TypeError exception.
   3. If R does not have an [[OriginalFlags]] internal slot, then
      a. If SameValue(R, %RegExpPrototype%) is true, return "(?:)".
 ---*/
 
-var get = Object.getOwnPropertyDescriptor(RegExp.prototype, 'source').get;
-
-assert.sameValue(get.call(RegExp.prototype), '(?:)');
+assert.sameValue(RegExp.prototype.source, '(?:)');
```
leobalter: while I really like the new description, the current assertion change might provide a false positive if the descriptor returns a data descriptor, using `value` instead of a get call. I assume the current original test is doing an explicit get call for this purpose.

test/built-ins/RegExp/prototype/source/value-slash.js
```diff
@@ -0,0 +1,27 @@
...
+  formed by concatenating the Strings "/", S, "/", and F can be parsed
+  (in an appropriate lexical context) as a RegularExpressionLiteral that
+  behaves identically to the constructed regular expression.
+---*/
+
+var re = eval('/' + new RegExp('/').source + '/');
+
+assert(re.test('/'), 'input: "/"');
+assert(re.test('_/_'), 'input: "_/_"');
+assert(!re.test('\\'), 'input: "\\"');
```
leobalter: if the returned value is a string, this test is not relevant for source, as the assertions count only for the new created RegExp.

I just checked [other files](https://github.com/tc39/test262/blob/master/test/built-ins/RegExp/prototype/source/value.js) and to be honest I believe a simple comparison to the matching string would be enough.

I won't block this PR if you prefer to keep this file as it is. It's following consistency.

shvaikalesh: please correct me if misunderstood:

- we can just assert that `new RegExp('/').source` is `'\/'`

however, according to the [spec](http://www.ecma-international.org/ecma-262/#sec-escaperegexppattern):

> For example, if P is "/", then S could be "\/" or "\u002F", among other possibilities, but not "/"

leobalter: you are right! Thanks for being precise with the spec, while I missed it!

littledan: This is unfortunate! Four JS engines all escape as `\/`. It seems like a spec change is deserved here, to remove the implementation-defined behavior. (Ditto for line endings, where it's always `\r`, `\n`, and `\u2028`, `\u2029`, except in V8 which doesn't normalize line endings.)

PR #1148, opened 2017-07-11
add BigInt tests

cxielarko:
These are tests for the [BigInt proposal](tc39/proposal-bigint)

### DIFF
--- a/test/built-ins/Object/bigint.js
+++ b/test/built-ins/Object/bigint.js
@@ -1,0 +1,11 @@
+// Copyright (C) 2017 Robin Templeton. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Conversion of BigInt values to Objects
+esid: pending
+features: [BigInt]
+---*/
+
+assert(Object(0n) instanceof BigInt);
+assert.sameValue(Object(0n).valueOf(), 0n);
--- a/test/language/expressions/does-not-equals/bigint.js
+++ b/test/language/expressions/does-not-equals/bigint.js
@@ -1,0 +1,18 @@
+// Copyright (C) 2017 Robin Templeton. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: BigInt equality
+esid: pending
+features: [BigInt]
+---*/
+
+assert("x" != 0n);
+assert(0n != "x");
+assert(true != 0n);
+assert(false != 1n);
+assert(0n != true);
+assert(1n != false);
+assert(0n != NaN);
+assert(0n != Infinity);
+assert(0n != -Infinity);
--- a/test/language/expressions/greater-than-or-equal/bigint.js
+++ b/test/language/expressions/greater-than-or-equal/bigint.js
@@ -1,0 +1,13 @@
+// Copyright (C) 2017 Robin Templeton. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: BigInt comparison
+esid: pending
+features: [BigInt]
+---*/
+
+assert(1n >= 0);
+assert(1 >= 0n);
+assert(1n >= 1);
+assert(1 >= 1n);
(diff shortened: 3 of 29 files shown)


cxielarko:
This patch primarily tests integration with the rest of the language (syntax, type conversions, library methods, etc.) and doesn't include detailed arithmetic tests. Tests for the typed arrays and DataViews are also missing. I'm planning to submit more arithmetic tests in a separate patch.

leobalter:
it would help a lot to set a checklist - here on Test262 or in the proposal
issues - of parts that need coverage. I'll need to write it anyway if not
available when I find time to review so we can check what is missing to be
covered in follow up patches.

test/language/expressions/does-not-equals/bigint.js
```diff
@@ -0,0 +1,18 @@
...
+
+assert("x" != 0n);
+assert(0n != "x");
+assert(true != 0n);
+assert(false != 1n);
+assert(0n != true);
+assert(1n != false);
+assert(0n != NaN);
+assert(0n != Infinity);
+assert(0n != -Infinity);
```
littledan: - Rather than just `assert()`, use `assert.sameValue(true, ...)` to be completely sure that this outputs a boolean.
- Share the tests from the "equals" assertions, except asserting that the output is false
- Include a test which requires some calculation to compare the BigInt to the Number

test/language/expressions/greater-than-or-equal/bigint.js
```diff
@@ -0,0 +1,13 @@
...
+/*---
+description: BigInt comparison
+esid: pending
+features: [BigInt]
+---*/
+
+assert(1n >= 0);
+assert(1 >= 0n);
+assert(1n >= 1);
+assert(1 >= 1n);
```
littledan: Include some test which requires a little bit more calculation, and include a comparison between some non-numeric types and a BigInt.

test/language/expressions/logical-not/bigint.js
```diff
@@ -0,0 +1,11 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Conversion of BigInt values to booleans
+esid: pending
+features: [BigInt]
+---*/
+
+assert.sameValue(!!0n, false, "Expected ToBoolean(0n) to be false");
+assert.sameValue(!!1n, true, "Expected ToBoolean(1n) to be true");
```
littledan: Also test this through `if`, `||`, `&&`, `?:`.

cxielarko:
I used a checklist like the following, based on the proposal's TOC, to develop the initial tests. (It could probably be more detailed, e.g. listing which interactions between features are tested)

- BigInt literals
- Type Conversion
  - ToPrimitive
  - ToBoolean
  - ToNumber
  - ToString
  - ToObject
  - ToNumeric
  - RequireObjectCoercible
- Testing/comparison
- Expressions
  - increment/decrement
  - typeof
  - prefix +
  - negation
  - bitwise operators
  - shifts
  - arithmetic operators
- BigInt object
  - constructor
  - parseInt
  - asUintN
  - asIntN
  - prototype
    - constructor
    - toLocaleString
    - toString
    - valueOf
- JSON.stringify
- Number constructor
- TypedArrays/DataViews

leobalter:
I'm working on a follow up for this today.

test/language/expressions/does-not-equals/bigint.js
leobalter: there's a lot of observation parts here that I can't get an exact reference for all of these cases combined.

https://tc39.github.io/proposal-bigint/#sec-abstract-equality-comparison

test/built-ins/Object/bigint.js
```diff
@@ -0,0 +1,11 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Conversion of BigInt values to Objects
+esid: pending
+features: [BigInt]
+---*/
+
+assert(Object(0n) instanceof BigInt);
+assert.sameValue(Object(0n).valueOf(), 0n);
```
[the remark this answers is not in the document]
littledan:
```js
function wrap() { return this; }
assert.sameValue(typeof wrap.call(0n) === 'object');
```

PR #1149, opened 2017-07-13
Improve RegExp.prototype.flags coverage

shvaikalesh:


test/built-ins/RegExp/prototype/flags/coercion.js
```diff
@@ -0,0 +1,61 @@
...
+info: >
+  get RegExp.prototype.flags
+
+  [...]
+  4. Let global be ToBoolean(? Get(R, "global")).
+  6. Let ignoreCase be ToBoolean(? Get(R, "ignoreCase")).
+  8. Let multiline be ToBoolean(? Get(R, "multiline")).
+  10. Let dotAll be ToBoolean(? Get(R, "dotAll")).
+  12. Let unicode be ToBoolean(? Get(R, "unicode")).
+  14. Let sticky be ToBoolean(? Get(R, "sticky")).
```
rwaldron: Each step should be in its own file

test/built-ins/RegExp/prototype/flags/this-val-regexp.js
```diff
@@ -0,0 +1,20 @@
+// Copyright (C) 2017 Aleksey Shvayka. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-get-regexp.prototype.flags
+description: >
+  RegExp.prototype.flags returns RegExp flags as a string
+info: |
+  1. Let R be the this value.
+  2. If Type(R) is not Object, throw a TypeError exception.
```
rwaldron: These steps are not relevant in this test file.

test/built-ins/RegExp/prototype/flags/this-val-regexp-prototype.js
```diff
@@ -0,0 +1,16 @@
+// Copyright (C) 2016 Aleksey Shvayka. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-get-regexp.prototype.flags
+description: >
+  Return "" when the `this` value is the RegExp.prototype object
+info: |
+  1. Let R be the this value.
+  2. If Type(R) is not Object, throw a TypeError exception.
```
rwaldron: These steps are not relevant in this test file.

test/built-ins/RegExp/prototype/flags/rethrow.js
```diff
@@ -0,0 +1,68 @@
...
+  });
+}, 'unicode');
+
+assert.throws(Test262Error, function() {
+  get.call({
+    get sticky() {
+      throw new Test262Error();
+    },
+  });
+}, 'sticky');
```
rwaldron: It would be better if each of the assertion messages matched the step they were testing. For example, this would be `` `Let sticky be ToBoolean(? Get(R, "sticky"))` `` instead of just `'sticky'`

test/built-ins/RegExp/prototype/flags/this-val-regexp-prototype.js
leobalter: I agree, but considering the delay, I believe it's best to fix this with follow ups.

test/built-ins/RegExp/prototype/flags/coercion.js
leobalter: yes, I like to avoid loops and here we have the forEach on each one. It's better to just split them in different files and have a specific files for each case. If we find and errors or regression, the information will be more valuable and complete.

leobalter: I'm gonna work on a follow up for this as well.

leobalter:
I'm working in a follow up commit to apply your feedback, @rwaldron

PR #1150, opened 2017-07-13
Revert "Add initial set of class field tests (from V8's test-parsing.cc)"

littledan:
Reverts tc39/test262#1064

Class fields are at Stage 2, so it's not appropriate to land the tests in test262


rwaldron:
@littledan what's the plan here? Is there going to be a new PR opened for the patch that was landed?

littledan:
@rwaldron Created a reland patch at https://github.com/tc39/test262/pull/1151 .

PR #1151, opened 2017-07-13
Reland "Add initial set of class field tests (from V8's test-parsing.cc)"

littledan:
Reverts tc39/test262#1064

cc @xanlpz

### DIFF
--- a/src/class-fields/computed-name-empty-var-computed-name-empty-function.case
+++ b/src/class-fields/computed-name-empty-var-computed-name-empty-function.case
@@ -1,0 +1,29 @@
+// Copyright (C) 2016 the V8 project authors, 2017 Igalia S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: Computed name var not initialized, computed name function
+info: |
+  1.1 New Productions
+
+  [...]
+
+  FieldDefinitionList [Yield, Await]:
+    FieldDefinition [?Yield, ?Await]
+    FieldDefinitionList  [?Yield, ?Await], FieldDefinition [?Yield, ?Await]
+template: default
+includes: [propertyHelper.js]
+---*/
+
+//- body
+  ['a']; ['b'](){}
+
+//- assertions
+  assert.sameValue(this.a, undefined);
+  verifyEnumerable(this, "a");
+  verifyWritable(this, "a");
+  verifyConfigurable(this, "a");
+
+  assert.sameValue(typeof Object.getPrototypeOf(this).b, "function");
+  verifyNotEnumerable(Object.getPrototypeOf(this), "b");
+  verifyConfigurable(Object.getPrototypeOf(this), "b");
--- a/src/class-fields/computed-name-empty-var-empty-function.case
+++ b/src/class-fields/computed-name-empty-var-empty-function.case
@@ -1,0 +1,29 @@
+// Copyright (C) 2016 the V8 project authors, 2017 Igalia S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: Computed name var not initialized, empty function
+info: |
+  1.1 New Productions
+
+  [...]
+
+  FieldDefinitionList [Yield, Await]:
+    FieldDefinition [?Yield, ?Await]
+    FieldDefinitionList  [?Yield, ?Await], FieldDefinition [?Yield, ?Await]
+template: default
+includes: [propertyHelper.js]
+---*/
+
+//- body
+  ['a']; b(){}
+
+//- assertions
+  assert.sameValue(this.a, undefined);
+  verifyEnumerable(this, "a");
+  verifyWritable(this, "a");
+  verifyConfigurable(this, "a");
+
+  assert.sameValue(typeof Object.getPrototypeOf(this).b, "function");
+  verifyNotEnumerable(Object.getPrototypeOf(this), "b");
+  verifyConfigurable(Object.getPrototypeOf(this), "b");
--- a/src/class-fields/default/cls-decl-extends.template
+++ b/src/class-fields/default/cls-decl-extends.template
@@ -1,0 +1,18 @@
+// Copyright (C) 2016 the V8 project authors, 2017 Igalia S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+path: language/expressions/class/class-field-extends-
+name: class fields
+---*/
+
+class Base {}
+class C extends Base {
+  /*{ body }*/
+  constructor() {
+    super();
+    /*{ assertions }*/
+  }
+}
+
+const c = new C();
--- a/src/class-fields/zero-zero-initializer-var.case
+++ b/src/class-fields/zero-zero-initializer-var.case
@@ -1,0 +1,25 @@
+// Copyright (C) 2016 the V8 project authors, 2017 Igalia S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: zero initialized zero var
+info: |
+  1.1 New Productions
+
+  [...]
+
+  FieldDefinitionList [Yield, Await]:
+    FieldDefinition [?Yield, ?Await]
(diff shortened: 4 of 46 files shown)


src/class-fields/default/cls-decl-extends.template
```diff
@@ -0,0 +1,18 @@
+// Copyright (C) 2016 the V8 project authors, 2017 Igalia S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+path: language/expressions/class/class-field-extends-
+name: class fields
```
rwaldron: "class fields" => "FieldDefinitionList" (for both files)

src/class-fields/default/cls-decl-extends.template
```diff
@@ -0,0 +1,18 @@
+// Copyright (C) 2016 the V8 project authors, 2017 Igalia S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+path: language/expressions/class/class-field-extends-
```
rwaldron: "expressions" => "statements" (for both files)

These templates create class declarations, which should end up in test/language/statements/class/

src/class-fields/zero-zero-initializer-var.case
```diff
@@ -0,0 +1,25 @@
+// Copyright (C) 2016 the V8 project authors, 2017 Igalia S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: zero initialized zero var
```
rwaldron: This should be a single line, human readable sentence that provides a basic explanation of the test itself. (for all case files)

src/class-fields/zero-zero-initializer-var.case
```diff
@@ -0,0 +1,25 @@
...
+info: |
+  1.1 New Productions
```
rwaldron: Please remove this, as it won't be relevant when the proposal is merged into the spec

src/class-fields/computed-name-empty-var-computed-name-empty-function.case
```diff
@@ -0,0 +1,29 @@
...
+info: |
+  1.1 New Productions
+
+  [...]
+
+  FieldDefinitionList [Yield, Await]:
+    FieldDefinition [?Yield, ?Await]
+    FieldDefinitionList  [?Yield, ?Await], FieldDefinition [?Yield, ?Await]
+template: default
+includes: [propertyHelper.js]
```
rwaldron: Please add `esid: pending` to all case files

src/class-fields/computed-name-empty-var-empty-function.case
```diff
@@ -0,0 +1,29 @@
...
+
+//- assertions
+  assert.sameValue(this.a, undefined);
+  verifyEnumerable(this, "a");
+  verifyWritable(this, "a");
+  verifyConfigurable(this, "a");
+
+  assert.sameValue(typeof Object.getPrototypeOf(this).b, "function");
+  verifyNotEnumerable(Object.getPrototypeOf(this), "b");
+  verifyConfigurable(Object.getPrototypeOf(this), "b");
```
rwaldron: Using `Object.getPrototypeOf(this)` unnecessarily overcomplicates the tests, please use `this` (for all case files)

src/class-fields/default/cls-decl-extends.template
```diff
@@ -0,0 +1,18 @@
...
+name: class fields
+---*/
```
rwaldron: Add to front matter: 

```
features: [class-fields]
```

(for both files)

src/class-fields/computed-name-empty-var-empty-function.case
littledan: I'm confused by this comment; "b" is not a property of `this` but rather of `this.__proto__`.

rwaldron: This is the generated test: 

```js
class Base {}
class C extends Base {
  ['a']; b(){}

  constructor() {
    super();
    assert.sameValue(this.a, undefined);
      verifyEnumerable(this, "a");
      verifyWritable(this, "a");
      verifyConfigurable(this, "a");

      assert.sameValue(typeof Object.getPrototypeOf(this).b, "function");
      verifyNotEnumerable(Object.getPrototypeOf(this), "b");
      verifyConfigurable(Object.getPrototypeOf(this), "b");
  }
}

const c = new C();
```


I want it to read: 

```js
class Base {}
class C extends Base {
  ['a']; b(){}

  constructor() {
    super();
    assert.sameValue(this.a, undefined);
      verifyEnumerable(this, "a");
      verifyWritable(this, "a");
      verifyConfigurable(this, "a");

      assert.sameValue(typeof this.b, "function");
      verifyNotEnumerable(this, "b");
      verifyConfigurable(this, "b");
  }
}

const c = new C();
```

Because, as a follow up, I want to update the cases and templates with additions necessary to produce this: 

```js
class Base {}
class C extends Base {
  ['a']; b(){}

  constructor() {
    super();
    assert.sameValue(this.a, undefined);
    verifyEnumerable(this, "a");
    verifyWritable(this, "a");
    verifyConfigurable(this, "a");

    assert.sameValue(typeof this.b, "function");
    verifyNotEnumerable(this, "b");
    verifyConfigurable(this, "b");
  }
}

assert.sameValue(C.prototype.a, undefined);
assert.sameValue(typeof C.prototype.b, "function");

const c = new C();
```

(Or something similar to that.

leobalter: overcalling Object.getPrototypeOf(this) can really be improved here as Rick suggested.

It does not hurt checking `this` has no own property `a` or `b` and get the property from the prototype of the instance.

For this case, I suggest:

```
class Base {}
class C extends Base {
  ['a']; b(){ return 42; }

  constructor() {
    super();

    assert.sameValue(Object.hasOwnProperty.call(this, 'a'), true);
    assert.sameValue(Object.hasOwnProperty.call(this, 'b'), false); // redundant, already covered, but shrugs

    const proto = Object.getPrototypeOf(this);

    assert.sameValue(this.a, undefined);
    assert.sameValue(this.b(), 42);

    verifyProperty(this, 'a', {
      value: undefined,
      enumerable: true,
      writable: true,
      configurable: true,
    });
  }
}

new C();
```

Remember we are not testing the property values of method definitions here, these are covered in other files. So I wouldn't bother.

rwaldron: Correction to my earlier example: 

```js

const c = new C();
assert.sameValue(typeof C.prototype.b, "function");
verifyNotEnumerable(C.prototype, "b");
verifyConfigurable(C.prototype, "b");

```

(just moving `verifyConfigurable` to the end)

littledan: OK, it sounds like the test is valid, you're just making cleanup suggestions. Is that accurate? Or is this about test coverage as well?

For context, these tests are really aiming to get at whether the syntax is handled correctly. Verifying the properties is just meant as a smoke test that something vaguely correct happened. Tests for all the semantic details are expected to come separately. I'm not sure if that affects this review at all.

rwaldron: > OK, it sounds like the test is valid, you're just making cleanup suggestions. Is that accurate? 

100% accurate. I'm just looking for simplification. 

> For context, these tests are really aiming to get at whether the syntax is handled correctly. Verifying the properties is just meant as a smoke test that something vaguely correct happened. Tests for all the semantic details are expected to come separately. I'm not sure if that affects this review at all.

I understood that and it's exactly why I wanted to simplify the assertions. We're completely on the same page.

leobalter:
This can be merged as soon as it is fixed.

I volunteer myself to work on the fixes as I already messed with an early merge.

leobalter:
I was working on the cases here, and I believe I can refactor this to reuse more templates and less cases. There's a lot of combinations that only needs the proper placements for the cases. I'll see what I can rewrite and improve from it.

leobalter:
This is now solved by #1190.

Issue #1154, opened 2017-07-21
A number of incorrect async iteration tests


caitp:
There are a few classes of these errors:

1. These are wrong because the test templates expect variables to be accessed
while still in TDZ, but since there's an `await` before they're accessed, loading those
variables does not throw a ReferenceError (not in TDZ)

- https://github.com/tc39/test262/blob/master/test/language/statements/for-await-of/async-func-decl-dstr-array-elem-init-let.js
- https://github.com/tc39/test262/blob/master/test/language/statements/for-await-of/async-func-decl-dstr-array-elem-put-let.js
- https://github.com/tc39/test262/blob/master/test/language/statements/for-await-of/async-gen-decl-dstr-array-elem-init-let.js
- https://github.com/tc39/test262/blob/master/test/language/statements/for-await-of/async-gen-decl-dstr-array-elem-put-let.js

---

2. This test seems to expect an extra AsyncGeneratorRequest to be fulfilled, but
there's no `yield` in the generator to allow this.

- https://github.com/tc39/test262/blob/master/test/language/statements/for-await-of/async-gen-decl-dstr-array-elem-trlg-iter-elision-iter-abpt.js

---

3. These tests expects .return() on a suspended-at-start generator to reject the
AsyncGeneratorRequest promise, which can never happen.

- https://github.com/tc39/test262/blob/master/test/language/statements/for-await-of/async-gen-decl-dstr-array-elem-trlg-iter-get-err.js
- https://github.com/tc39/test262/blob/master/test/language/statements/for-await-of/async-gen-decl-dstr-array-elem-trlg-iter-list-nrml-close-null.js
- https://github.com/tc39/test262/blob/master/test/language/statements/for-await-of/async-gen-decl-dstr-array-elem-trlg-iter-list-rtrn-close-err.js
- https://github.com/tc39/test262/blob/master/test/language/statements/for-await-of/async-gen-decl-dstr-array-elem-trlg-iter-list-rtrn-close-null.js
- https://github.com/tc39/test262/blob/master/test/language/statements/for-await-of/async-gen-decl-dstr-array-elem-trlg-iter-rest-rtrn-close-null.js

Similarly, this one expects the generator body to have been evaluated after .return() when in
the suspended-at-start state.

- https://github.com/tc39/test262/blob/master/test/language/statements/for-await-of/async-gen-decl-dstr-array-elem-trlg-iter-list-rtrn-close.js

---

[...]

PR #1155, opened 2017-07-25
Remove selection of incorrect async iteration tests. Fixes gh-1154

rwaldron:


leobalter:
Note: I had the chance to discuss this a while ago with Rick and spent some time reading the changes last week while I was in the TC39 meeting. Silent merges sound silly but some of them holds a background conversation I might fail bringing any reason or updates here.

PR #1156, opened 2017-07-26
Promise.prototype.finally: add tests

ljharb:
(The first two commits can and should be cherry-picked in separately; i'll rebase this PR ASAP if that happens)

`Promise.prototype.finally` reached stage 3 today at TC39; here's my first crack at the tests.

See https://github.com/tc39/proposal-promise-finally/issues/18, https://tc39.github.io/proposal-promise-finally/, and https://github.com/es-shims/Promise.prototype.finally/blob/master/test/tests.js for more info.

### DIFF
--- a/test/built-ins/Promise/prototype/finally/invokes-then-with-non-function.js
+++ b/test/built-ins/Promise/prototype/finally/invokes-then-with-non-function.js
@@ -1,0 +1,49 @@
+// Copyright (C) 2017 Jordan Harband. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+author: Jordan Harband
+description: Promise.prototype.finally invokes `then` method
+esid: sec-promise.prototype.finally
+features: [Promise.prototype.finally]
+---*/
+
+var target = new Promise(function () {});
+var returnValue = {};
+var callCount = 0;
+var thisValue = null;
+var argCount = null;
+var firstArg = null;
+var secondArg = null;
+var result = null;
+
+target.then = function(a, b) {
+  callCount += 1;
+
+  thisValue = this;
+  argCount = arguments.length;
+  firstArg = a;
+  secondArg = b;
+
+  return returnValue;
+};
+
+result = Promise.prototype.finally.call(target, 1, 2, 3);
+
+assert.sameValue(callCount, 1, 'Invokes `then` method exactly once');
+assert.sameValue(
+  thisValue,
+  target,
+  'Invokes `then` method with the instance as the `this` value'
+);
+assert.sameValue(argCount, 2, 'Invokes `then` method with exactly two single arguments');
+assert.sameValue(
+  firstArg,
+  1,
+  'Invokes `then` method with the provided non-callable first argument'
+);
+assert.sameValue(
+  secondArg,
+  1,
+  'Invokes `then` method with the provided non-callable first argument'
+);
+assert.sameValue(result, returnValue, 'Returns the result of the invocation of `then`');
--- a/test/built-ins/Promise/prototype/finally/rejected-observable-then-calls.js
+++ b/test/built-ins/Promise/prototype/finally/rejected-observable-then-calls.js
@@ -1,0 +1,46 @@
+// Copyright (C) 2017 Jordan Harband. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+author: Jordan Harband
+description: finally observably calls .then
+esid: sec-promise.prototype.finally
+features: [Promise.prototype.finally]
+flags: [async]
+---*/
+
+var initialThenCount = 0;
+var noReason = {};
+var no = Promise.reject(noReason);
+no.then = function () {
+  initialThenCount += 1;
+  return Promise.prototype.then.apply(this, arguments);
+};
+
+var onFinallyThenCount = 0;
+var yesValue = {};
+var yes = Promise.resolve(yesValue);
+yes.then = function () {
+  onFinallyThenCount += 1;
+  return Promise.prototype.then.apply(this, arguments);
+};
+
+var finallyCalled = false;
+var catchCalled = false;
+
+no.catch(function (e) {
+  assert.sameValue(e, noReason);
+  throw e;
+}).finally(function () {
+  finallyCalled = true;
+  return yes;
+}).catch(function (e) {
+  catchCalled = true;
+  assert.sameValue(e, noReason);
+}).then(function () {
+  assert.sameValue(finallyCalled, true, 'initial finally was called');
+  assert.sameValue(initialThenCount, 1, 'initial finally invokes .then once');
+
+  assert.sameValue(catchCalled, true, 'catch was called');
+  assert.sameValue(onFinallyThenCount, 1, 'onFinally return promise has .then invoked once');
+  $DONE();
+}).catch($ERROR);
--- a/test/built-ins/Promise/prototype/finally/resolution-value-no-override.js
+++ b/test/built-ins/Promise/prototype/finally/resolution-value-no-override.js
@@ -1,0 +1,20 @@
+// Copyright (C) 2017 Jordan Harband. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+author: Jordan Harband
+description: finally on a fulfilled promise can not override the resolution value
+esid: sec-promise.prototype.finally
+features: [Promise.prototype.finally]
+flags: [async]
+---*/
+
+var obj = {};
+
+var p = Promise.resolve(obj);
+
+p.finally(function () {
+  assert.sameValue(arguments.length, 0, 'onFinally receives zero args');
+  return {};
+}).then(function (x) {
+  assert.sameValue(x, obj, 'onFinally can not override the resolution value');
+}).then($DONE).catch($ERROR);
--- a/test/built-ins/Promise/prototype/finally/this-value-non-object.js
+++ b/test/built-ins/Promise/prototype/finally/this-value-non-object.js
@@ -1,0 +1,17 @@
+// Copyright (C) 2017 Jordan Harband. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+author: Jordan Harband
+description: >
+  Promise.prototype.finally called with a non-object-coercible `this` value
+esid: sec-promise.prototype.finally
+features: [Promise.prototype.finally]
+---*/
+
+assert.throws(TypeError, function() {
+  Promise.prototype.finally.call(undefined);
+}, 'undefined');
+
+assert.throws(TypeError, function() {
+  Promise.prototype.finally.call(null);
+}, 'null');
--- a/test/built-ins/Promise/prototype/finally/this-value-then-not-callable.js
+++ b/test/built-ins/Promise/prototype/finally/this-value-then-not-callable.js
@@ -1,0 +1,51 @@
+// Copyright (C) 2017 Jordan Harband. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+author: Jordan Harband
+description: >
+  Promise.prototype.finally called with a `this` value that does not define a
+  callable `then` property
+esid: sec-promise.prototype.finally
+features: [Symbol, Promise.prototype.finally]
+---*/
+
+var symbol = Symbol();
+
+var thrower = function () { throw new Test262Error('this should never happen'); };
+
+var p = new Promise(function () {});
+
+p.then = undefined;
+assert.throws(TypeError, function() {
+  Promise.prototype.finally.call(p, thrower);
+}, 'undefined');
+
+p.then = null;
+assert.throws(TypeError, function() {
+  Promise.prototype.finally.call(p, thrower);
+}, 'null');
+
+p.then = 1;
+assert.throws(TypeError, function() {
+  Promise.prototype.finally.call(p, thrower);
+}, 'number');
+
+p.then = '';
+assert.throws(TypeError, function() {
+  Promise.prototype.finally.call(p, thrower);
+}, 'string');
+
+p.then = true;
+assert.throws(TypeError, function() {
+  Promise.prototype.finally.call(p, thrower);
+}, 'boolean');
+
+p.then = symbol;
+assert.throws(TypeError, function() {
+  Promise.prototype.finally.call(p, thrower);
+}, 'symbol');
+
+p.then = {};
+assert.throws(TypeError, function() {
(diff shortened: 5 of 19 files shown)


benjamingr:
> One interaction that I'm wondering about is the interaction with the HostPromiseRejectionTracker hook. Seems like the Promise there will be the result of finally, rather than the Promise used as input to it.

That's a good point, the promise caught should be the result of the `finally` and not the result of the  original promise. More specifically:

```js
let p = Promise.reject(new Error());
let p2 = p.finally(() => {});
```

The unhandled rejection should be p2 and not p.

littledan:
@benjamingr To be clear, I believe the semantics you're saying are the right ones are what the specification implies. I was just wondering about tests for this property. Thanks for spelling it out better.

ljharb:
If there are existing tests for this hook, then I should be able to utilize them to test it here - if not, I'd need some guidance.

benjamingr:
@ljharb Node tracks them with `process.on("unhandledRejection", (e, p) => {})` and the browser with `window.addEventListener("unhandledrejection", e => {... })`.

If you'd like help I can send you links to the PRs adding the hooks which should shed light on how one might test them.

ljharb:
@benjamingr i'm aware of how a specific implementation tests them; but how would test262 - which is supposed to be host-agnostic - test those? It's run on more engines than browsers and node, and the spec *permits* unhandled rejection tracking, but does not mandate it.

ljharb:
I don't see any mention of `HostPromiseRejectionTracker` in the entire codebase; it seems like testing this isn't yet a solved problem.

littledan:
Test262 would have to add a host hook to test this. I don't think it has a hook like this currently. If test262 maintainers don't like the idea of adding such a hook, another method of testing would be to add something to web-platform-tests.

test/built-ins/Promise/prototype/finally/this-value-then-poisoned.js
```diff
@@ -0,0 +1,20 @@
...
+---*/
+
+var poisonedThen = Object.defineProperty(new Promise(function () {}), 'then', {
+  get: function() {
+    throw new Test262Error();
+  }
+});
+
+assert.throws(Test262Error, function() {
+  Promise.prototype.finally.call(poisonedThen);
```
leobalter: this would like calling `poisonedThen.finally()`, right? We can add a new `assert.throws` in this file.

test/built-ins/Promise/prototype/finally/this-value-then-throws.js
```diff
@@ -0,0 +1,19 @@
...
+---*/
+
+var thrower = new Promise(function () {});
+thrower.then = function() {
+  throw new Test262Error();
+};
+
+assert.throws(Test262Error, function() {
+  Promise.prototype.finally.call(thrower);
+});
```
leobalter: this would like calling `thrower.finally()`, right? We can add a new `assert.throws` in this file.

test/built-ins/Promise/prototype/finally/this-value-then-not-callable.js
```diff
@@ -0,0 +1,49 @@
...
+features: [Symbol, Promise.prototype.finally, Function.prototype.call]
+---*/
+
+var symbol = Symbol();
+
+var p = new Promise(function () {});
+
+p.then = undefined;
+assert.throws(TypeError, function() {
+  Promise.prototype.finally.call(p);
```
leobalter: we could poison finally to make sure TypeError happens at the expected time:

Correct me if I'm not using the API properly:

```js
p.finally(function() { throw new Test262Error('this should not run'); });
```

test/built-ins/Promise/prototype/finally/this-value-non-object.js
```diff
@@ -0,0 +1,17 @@
...
+author: Jordan Harband
+description: >
+  Promise.prototype.finally called with a non-object-coercible `this` value
+esid: sec-promise.prototype.finally
+features: [Promise.prototype.finally, Function.prototype.call]
+---*/
+
+assert.throws(TypeError, function() {
+  Promise.prototype.finally.call(undefined);
+}, 'undefined');
```
leobalter: here you can poison a promise's `then`:

```js
const p = new Promise(function() {})

p.then = function() { throw new Test262Error('poison'); };

  p.finally.call(undefined);
}, 'undefined');
```

leobalter:
> Test262 would have to add a host hook to test this. I don't think it has a hook like this currently. If test262 maintainers don't like the idea of adding such a hook, another method of testing would be to add something to web-platform-tests.

I would need more time to review this, maybe @rwaldron is interested.

test/built-ins/Promise/prototype/finally/invokes-then-with-non-function.js
```diff
@@ -0,0 +1,49 @@
+// Copyright (C) 2017 Jordan Harband. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+author: Jordan Harband
+description: Promise.prototype.finally invokes `then` method
+esid: sec-promise.prototype.finally
+features: [Promise, Promise.prototype.finally]
```
ljharb: Would nobody want to run tests against an engine without `Promise`?

test/built-ins/Promise/prototype/finally/this-value-then-not-callable.js
ljharb: ah, interesting. sure, i can do that (no need to poison the finally method, but i can certainly pass a poisoned onFinally)

test/built-ins/Promise/prototype/finally/this-value-non-object.js
ljharb: I'm not sure what value this adds; it we use `.call` then the function has no more link to `p`, so `p.then` couldn't be reached even if i wanted it to.

ljharb:
Thanks, updated! I haven't added "info" since I'm not sure where to link to prior to actually merging with the spec.

gsathya:
Quick question: do these tests pass when running V8 with --harmony-promise-finally flag? 

FYI, there's a lot of V8 tests as well if you want to import them: https://cs.chromium.org/chromium/src/v8/test/mjsunit/harmony/promise-prototype-finally.js

ljharb:
@gsathya running test262 tests isn't easy; i can definitely give that a shot :-)

in general tho, wouldn't it make sense to never write v8-only tests, when something could go in test262? :-p

gsathya:
> @gsathya running test262 tests isn't easy; i can definitely give that a shot :-)

Sounds good. Thanks! Node 8 should have this feature behind the flag, no need to build d8.

> in general tho, wouldn't it make sense to never write v8-only tests, when something could go in test262? :-p

Until quite recently it wasn't possible to write test262 tests as part of a v8 patch -- we had to do out of band test262 code review dance and then roll that change into v8. @littledan has done some great work in making this possible now, so going forward we'll be writing more test262 tests.

ljharb:
Tests pass with both my `promise.prototype.finally` npm module, but I can't get test262-harness to run with the `--harmony-promise-finally` flag on node 8 latest (altho the flag works by itself) :-/

ljharb:
However, when I run `promise.prototype.finally`'s tests against `node --harmony-promise-finally`, there is one of my own tests that fails - specifically around observable then calls with a Promise subclass.

Needless to say we'll have to make sure both implementations as well as the spec and the test262 tests are all in line :-)

rwaldron:
> tests pass with both my promise.prototype.finally npm module, but I can't get test262-harness to run with the --harmony-promise-finally flag on node 8 latest (altho the flag works by itself) :-/

I'll dig into this.

rwaldron:
@ljharb the flag is `--harmony_promise_finally` (according to d8 built from source)

rwaldron:
I wouldn't bother using Node.js to test anything until the Node.js patch to eshost has landed. Also, I am experiencing the same problem with `--harmony_promise_finally` and it's because the order of arguments is wrong: 

(inside ConsoleAgent.js, which is in eshost)
```js
  createChildProcess(args) {
    args = args || [];
    args = args.concat(this.args);
    return cp.spawn(this.hostPath, args);
  }
```

Which results in the args coming out as (for example): 

```
[ '/var/folders/xz/k2lxgs3s43180_tjwn1v_kwc0000gn/T/117627-60511-189e99u.f98kg.js',
  '--harmony_promise_finally' ]
```

The flag needs to be before the file: 

```
rwaldron in ~/codez/javascript
$ node -v
v8.2.1
rwaldron in ~/codez/javascript
$ cat promise.prototype.finally.js
console.log(Promise.prototype.finally);
$ node --harmony_promise_finally promise.prototype.finally.js
[Function: finally]
rwaldron in ~/codez/javascript
$ node promise.prototype.finally.js --harmony_promise_finally
undefined
```

So I will make a PR to eshost that re-arranges the args for `NodeAgent`.

---------------------------------------------



Moving on to my local run results: 

## d8
```
$ test262-harness --hostType=d8 --hostPath=`which d8` --hostArgs='--harmony_promise_finally' test/built-ins/Promise/prototype/finally/**/*.js
Ran 32 tests
32 passed
0 failed
```

ljharb:
@rwaldron ah! thanks. with `node --harmony-promise-finally` it works in the repl, perhaps node normalizes dashes and the v8 api doesn't?

rwaldron:
Using my patch, I was able to confirm these tests on Node.js 8.2.1

Nice work @ljharb!

We can do the eshost dependent tests in a follow up PR.

ljharb:
figured it out; it's a bug in test262-harness. When patched, `node --harmony_promise_finally` passes this PR's tests!

rwaldron:
> perhaps node normalizes dashes and the v8 api doesn't?

Ah, that's probably very likely. There's no reason why NodeAgent couldn't/shouldn't try to do that as well.

rwaldron:
@ljharb 

>> Ah, that's probably very likely. 

I was wrong. Either dash or underscore work just fine, the issue was always _just_ the order of the args. Patch to eshost is coming.

test/built-ins/Promise/prototype/finally/rejected-observable-then-calls.js
```diff
@@ -0,0 +1,46 @@
...
+  onFinallyThenCount += 1;
+  return Promise.prototype.then.apply(this, arguments);
+};
+
+var finallyCalled = false;
+var catchCalled = false;
+
+no.catch(function (e) {
+  assert.sameValue(e, noReason);
+  throw e;
```
gskachkov: I would add cases with `thow e`, within fulfillment and rejection handlers i.e.:
```
Promise.resolve(1)
    .then(function () { throw new Error('');})
    .finally(/* Check if finally is reached */);

Promise.reject(1)
    .then(function () {}, function () { throw new Error('');})
    .finally(/* Check if finally is reached */)
```

gskachkov: It seems  that we need test that we raise TypeError when we use `finally` with `new` i.e.
```
const p = new Promise((resolve)=>{});

  const f = new p.finally();
}, 'no constructor');
```

ljharb: Hmm - I can add that test, but `p.finally()` returns a Promise, so I have no idea why it'd ever work with `new`, and `.catch()` lacks the same test. I'm pretty sure we don't test every returned object value to ensure it doesn't work with `new` - would checking that `typeof` wasn't "function" suffice?

gskachkov: Ohh, I see. You're right. 
So it will be enough to check if `finally` returns `Promise`.  Not sure that I can find such test, or I missed something.
If there is no such test we can use `Symbol.toStringTag`:
```
const p = Promise.resolve(1);
assert.sameValue(p.finally()[Symbol.toStringTag], 'Promise');
```

ljharb: I'd think chaining a `.then` or `.catch` on the end would be sufficient, and I think some of the tests already do that.

rwaldron:
The patch to eshost is in final stages now, so I don't consider this "blocked" anymore (not that it ever was, I just hoped to get that fixed first... you know how it is).

rwaldron:
`eshost` and `test262-harness` are both updated now and I ran these tests locally against `d8`: 

```
$ test262-harness --hostType d8 --hostPath `which d8` --hostArgs="--harmony_promise_finally" test/built-ins/Promise/prototype/finally/*.js
Ran 32 tests
32 passed
0 failed
```

Success!

test/built-ins/Promise/prototype/finally/resolution-value-no-override.js
```diff
@@ -0,0 +1,20 @@
...
+
+var obj = {};
+
+var p = Promise.resolve(obj);
+
+p.finally(function () {
+  assert.sameValue(arguments.length, 0, 'onFinally receives zero args');
+  return {};
+}).then(function (x) {
+  assert.sameValue(x, obj, 'onFinally can not override the resolution value');
```
gsathya: If the onFinally and the onFulfilled callbacks are skipped, the tests would still pass right?

gsathya: (I see other tests also that don't verify if the onFulfilled/onRejected/onFinally has actually run.)

gsathya:
We are missing tests that call the SpeciesConstructor. The following test fails in V8:

```js
// Flags: --allow-natives-syntax --harmony-promise-finally

var thenCalled = 0;

class FooPromise extends Promise {
  then(onFullfilled, onRejected) {
    thenCalled++;
    return super.then(onFullfilled, onRejected);
  }
}

var x = new FooPromise(r => r()).finally(() => {});
%RunMicrotasks();

// (1) Promise.prototype.finally
// 1. 7. Return ? Invoke(promise, "then", thenFinally, catchFinally).
// (2) ThenFinally function
// 1.1. 8. Return ? Invoke(promise, "then", valueThunk>
// (3) Promise Resolve Functions
// 25.4.1.3.2. 8. Let then be Get(resolution, "then").

print(thenCalled == 3);
```

test/built-ins/Promise/prototype/finally/resolution-value-no-override.js
ljharb: yes, that's true. i can make a followup PR that adds counters to ensure they're called; that's a good idea

ljharb:
Thanks, I'll write a followup PR this weekend that strengthens the tests!

leobalter:
+1 for follow up PRs.

ljharb:
#1398 is merged; but more followup is still ideal.

PR #1167, opened 2017-08-04
Add optional-catch-binding tests

### LINKED ISSUE #1166: Add coverage for Optional catch binding
Currently on stage 3

https://github.com/tc39/proposal-optional-catch-binding

azz:
First time contributing here, let me know if I've missed something!

Do we need tests for runtime semantics, given it is the same minus the runtime binding creation?

Closes #1166


bakkot:
Yeah, I think we definitely want some runtime semantics tests.

Maybe also some negative tests? That `try {} catch () {}` does not parse, for example?

leobalter:
Hi, Lucas!

Yes, while the current tests are valid - and well formatted - we need
coverage for the runtime semantics. So we can't close #1166 yet.

Are you interested in writing a follow up for the additional coverage?

leobalter:
Yes, negative tests are also interesting here as the detail on now allowing
empty parentheses are on the edge of both valid cases.

azz:
Thanks for the feedback.I'll add some negative cases too.

> Are you interested in writing a follow up for the additional coverage?
I am, I'll follow up shortly.

azz:
Looks like there are already negative cases covering this.

https://github.com/tc39/test262/blob/master/test/language/statements/try/S12.14_A16_T9.js


Also, what do we do with this test case?
https://github.com/tc39/test262/blob/master/test/language/statements/try/S12.14_A16_T4.js

```js
/*---
info: >
    TryStatement: "try Block Catch" or "try Block Finally" or "try Block
    Catch Finally"
es5id: 12.14_A16_T4
description: >
    Catch: "catch (Identifier ) Block". Checking if execution of
    "catch" that takes no arguments fails
negative:
  phase: early
  type: SyntaxError
---*/

// CHECK#1
try{}
catch{}
```

leobalter:
this second file should be removed, it's not valid anymore

test/language/statements/try/optional-catch-binding.js
```diff
@@ -0,0 +1,18 @@
...
+  Optional Catch Binding
+
+  Catch[Yield, Await, Return]:
+    (...)
+    catch Block[?Yield, ?Await, ?Return]
+---*/
+
+try {} catch {}
+
+try {} catch {} finally {}
```
rwaldron: This line should probably go in a separate file

leobalter: makes sense. They should both fail separately before implementation.

azz: Done.

rwaldron:
@azz great work on the updates. There is only one remaining issue to fix and it's just a one line removal and then this is good to go—thanks for your patience and diligence ❤️

test/language/statements/try/optional-catch-binding-parens.js
```diff
@@ -0,0 +1,21 @@
...
+  Catch[Yield, Await, Return]:
+    catch ( CatchParameter[?Yield, ?Await] ) Block[?Yield, ?Await, ?Return]
+negative:
+  phase: early
+  type: SyntaxError
+---*/
+
+throw "Test262: This statement should not be evaluated.";
+
+$ERROR('This code should not be executed.');
```
rwaldron: This line can be removed, it’s unreachable.

test/language/statements/try/optional-catch-binding-throws.js
```diff
@@ -0,0 +1,25 @@
...
+  Runtime Semantics: CatchClauseEvaluation
+
+  Catch : catch Block
+    (...)
+    Let B be the result of evaluating Block.
+    (...)
+    Return Completion(B).
+---*/
+
+assert.throws(Test262Error, function() {
```
michaelficarra: Why do we assert throwing for this? It's be just as easy to assert on a value

```js
(function() {
  try {
    throw 0;
    return false;
  } catch {
    return true;
  }
  return false;
})
```

leobalter: @michaelficarra, The description says it is checking the `catch` block won't hold new abrupt completions.

---

Anyway your code example it might be even more simple, but this is already being very well tested at test/language/statements/try/optional-catch-binding-lexical.js

```js
let x = false;

try {
  throw 0;
} catch {
  x = true
}

assert(x);
```

azz:
@rwaldron Are there any further changes required?

leobalter:
It's fine and good to go

PR #1168, opened 2017-08-04
Proxy, [[OwnPropertyKeys]]: If trapResult contains any duplicate entries, throw a TypeError

rwaldron:
Ref: https://github.com/tc39/ecma262/pull/833

### DIFF
--- a/test/built-ins/Proxy/ownKeys/return-duplicate-entries-throws.js
+++ b/test/built-ins/Proxy/ownKeys/return-duplicate-entries-throws.js
@@ -1,0 +1,22 @@
+// Copyright (C) 2017 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-proxy-object-internal-methods-and-internal-slots-ownpropertykeys
+description: >
+    The returned list must not contain any duplicate entries.
+info: >
+    [[OwnPropertyKeys]] ( )
+
+    ...
+    9. If trapResult contains any duplicate entries, throw a TypeError exception.
+---*/
+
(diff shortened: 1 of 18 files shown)


test/built-ins/Proxy/ownKeys/return-duplicate-entries-throws.js
```diff
@@ -0,0 +1,22 @@
...
+info: >
+    [[OwnPropertyKeys]] ( )
+
+    ...
+    9. If trapResult contains any duplicate entries, throw a TypeError exception.
+---*/
+
+var p = new Proxy({}, {
+  ownKeys() {
+    return ["a", "a"];
```
leobalter: Nice!

We also need to repeat this verification for duplicate symbols. A simple:

```
var s = Symbol();
var p = new Proxy({}, {
   ownKeys() {
     return [s, s];
...
```

should be enough.

test/built-ins/Proxy/ownKeys/return-duplicate-symbol-entries-throws.js
```diff
@@ -0,0 +1,23 @@
...
+var s = Symbol();
+var p = new Proxy({}, {
+  ownKeys() {
+    return [s, s];
+  }
+});
+
+assert.throws(TypeError, function() {
+  Object.keys(p);
+});
```
rwaldron: @leobalter ping

test/built-ins/Proxy/ownKeys/return-duplicate-entries-throws.js
rwaldron: This is completed

PR #1171, opened 2017-08-10
remove garbage: these try/catch tests don't actually test the section cited (or anything for that matter)

rwaldron:
Cited: https://www.ecma-international.org/ecma-262/5.1/index.html#sec-12.14.1

Present: https://tc39.github.io/ecma262/#sec-strict-mode-of-ecmascript

> It is a SyntaxError if a CatchParameter occurs within strict mode code and BoundNames of CatchParameter contains either eval or arguments (13.15.1).


littledan:
These have the smell of regression tests. It seems like the point is to make sure that those bindings are not incorrectly prohibited in sloppy mode, and now that tests are run in both strict and sloppy mode, it tests for strict mode as well.

PR #1190, opened 2017-08-21
Class fields

leobalter:
I'm working on on the class fields tests. The syntax has changed in the proposal and I'm also fixing this.

I'm also planning to add tests for private names but probably in a follow up PR, although, the templates might be good for them as well.

In this new template/case strategy I'm trying to stick with _cases_ for the class fields themselves, as the templates provide structural variations to hold them properly.

This should replace #1151

### DIFF
--- a/src/class-fields/computed-names.case
+++ b/src/class-fields/computed-names.case
@@ -1,0 +1,70 @@
+// Copyright (C) 2017 the V8 project authors, 2017 Igalia S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: prod-FieldDefinition
+desc: Computed property names
+info: |
+  ClassElement:
+    ...
+    FieldDefinition ;
+
+  FieldDefinition:
+    ClassElementName Initializer_opt
+
+  ClassElementName:
+    PropertyName
+template: default
+includes: [propertyHelper.js]
+features: [computed-property-names]
+---*/
+
+//- setup
+var x = "b";
+
+//- fields
+static ["a"] = 39; [x] = 42; [10] = "meep"; ["not initialized"]
+//- assertions
+assert.sameValue(Object.hasOwnProperty.call(C.prototype, "a"), false);
+assert.sameValue(Object.hasOwnProperty.call(c, "a"), false);
+
+verifyProperty(C, "a", {
+  value: 39,
+  enumerable: true,
+  writable: true,
+  configurable: true
+});
+
+assert.sameValue(Object.hasOwnProperty.call(C.prototype, "b"), false);
+assert.sameValue(Object.hasOwnProperty.call(C, "b"), false);
+
+verifyProperty(c, "b", {
+  value: 42,
+  enumerable: true,
+  writable: true,
+  configurable: true
+});
+
+assert.sameValue(Object.hasOwnProperty.call(C.prototype, "x"), false);
+assert.sameValue(Object.hasOwnProperty.call(C, "x"), false);
+assert.sameValue(Object.hasOwnProperty.call(c, "x"), false);
+
+assert.sameValue(Object.hasOwnProperty.call(C.prototype, "10"), false);
+assert.sameValue(Object.hasOwnProperty.call(C, "10"), false);
+
+verifyProperty(c, "10", {
+  value: "meep",
+  enumerable: true,
+  writable: true,
+  configurable: true
+});
+
+assert.sameValue(Object.hasOwnProperty.call(C.prototype, "not initialized"), false);
+assert.sameValue(Object.hasOwnProperty.call(C, "not initialized"), false);
+
+verifyProperty(c, "not initialized", {
+  value: "meep",
+  enumerable: true,
+  writable: true,
+  configurable: true
+});
--- a/src/class-fields/conventions.md
+++ b/src/class-fields/conventions.md
@@ -1,0 +1,8 @@
+# Conventions for the class fields templates and cases
+
+Templates should produce a class named `C` and instantiate it to an object named `c`.
+
+## Known template fields:
+
+* fields: it should contain the list of class fields, inserted in the class body. Please, avoid closing a field with a trailing semi-colon, this might prevent ASI checks properly.
+* assertions: it should contain the assertions to run after the class object is instantiated.
--- a/src/class-fields/default/cls-decl-wrapped-in-sc.template
+++ b/src/class-fields/default/cls-decl-wrapped-in-sc.template
@@ -1,0 +1,18 @@
+// Copyright (C) 2017 the V8 project authors, 2017 Igalia S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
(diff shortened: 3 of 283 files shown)


leobalter:
I believe I got enough coverage for a first round. I'm not touching on anything for private names or expected errors.

The last V8 version I built yesterday is not working here with the flag. 

```
test262-harness --hostType=d8 --hostPath=`which d8` --hostArgs="--harmony_class_fields" -t4 $(git diff master --name-only --diff-filter=d -- test/)

# or just

test262-harness --hostType=d8 --hostPath=`which d8` --hostArgs="--harmony_class_fields" test/language/statements/class/fields-*
```

For every test file I'm getting a `Test did not run to completion` error message.

leobalter:
Of course the CI is trolling.

On Travis CI, running `./tools/generation/generator.py create --out test src` is getting a difference and failing the process.

In my local machine I run `./make.py clean` and then `./make.py` and it doesn't show anything different to commit. Running the `./tools/generation/generator.py create --out test src` instead of `./make.py` has the same effect.

I'll defer this generation tool check to @rwaldron.

littledan:
I don't think you'll have much luck with V8 in running these tests. Much of the code was removed.

leobalter:
I should use babel to run the tests, in the meanwhile, it would be nice to have someone reviewing the templates and cases.

leobalter:
I found duplicate paths in the templates and the fix should solve the problem with the CI

src/class-fields/computed-names.case
```diff
@@ -0,0 +1,69 @@
...
+verifyProperty(c, "b", {
+  value: 42,
+  enumerable: true,
+  writable: true,
+  configurable: true
+});
+
+assert.sameValue(Object.hasOwnProperty.call(C.prototype, "x"), false);
+assert.sameValue(Object.hasOwnProperty.call(C, "x"), false);
+assert.sameValue(Object.hasOwnProperty.call(c, "x"), false);
```
rwaldron: Lines 47-49 should be covered in the existing computed property name tests; if not, then they should be—but not here.

src/class-fields/conventions.md
```diff
@@ -0,0 +1,8 @@
+# Conventions for the class fields templates and cases
+
+Templates should produce a class named `C` and instantiate it to an object named `c`.
+
+## Known template fields:
+
+* fields: it should contain the list of class fields, inserted in the class body. Please, avoid closing a field with a trailing semi-colon, this might prevent ASI checks properly.
+* assertions: it should contain the assertions to run after the class object is instantiated.
```
rwaldron: Excellent

src/class-fields/default/cls-decl-wrapped-in-sc.template
```diff
@@ -0,0 +1,18 @@
...
+/*---
+path: language/statements/class/fields-wrapped-in-sc-
+name: fields definition wrapped in semicolons
+features: [class-fields]
+---*/
+
+class C {
+  ;;;;
+  ;;;;;;/*{ fields }*/;;;;;;;
+  ;;;;
```
rwaldron: lol, so many empty statements

src/class-fields/static-computed-symbol-names.case
```diff
@@ -0,0 +1,55 @@
...
+  configurable: true
+});
+
+assert.sameValue(Object.hasOwnProperty.call(C.prototype, "x"), false);
+assert.sameValue(Object.hasOwnProperty.call(C, "x"), false);
+assert.sameValue(Object.hasOwnProperty.call(c, "x"), false);
+
+assert.sameValue(Object.hasOwnProperty.call(C.prototype, "y"), false);
+assert.sameValue(Object.hasOwnProperty.call(C, "y"), false);
+assert.sameValue(Object.hasOwnProperty.call(c, "y"), false);
```
rwaldron: Same issue as above, this Lines 49-55 belong in computed property names tests. I'm not going to note these anymore, but this applies to all tests this PR: if it's a test that asserts there is no property made with a name that is the string value of the IdentifierReference that was used in the computed property name.

leobalter:
WRT the current diff conflict, I'll rebase and merge the fixup commits right after the first review.

src/class-fields/computed-names.case
```diff
@@ -0,0 +1,69 @@
...
+/*---
+esid: prod-FieldDefinition
+desc: Computed property names
+info: |
+  ClassElement:
+    ...
+    FieldDefinition ;
+
+  FieldDefinition:
+    ClassElementName Initializer(opt)
```
leobalter: no specific reason, I'll fix it

src/class-fields/computed-names.case
```diff
@@ -0,0 +1,69 @@
...
+---*/
+
+//- setup
+var x = "b";
+
+//- fields
+static ["a"] = 39; [x] = 42; [10] = "meep"; ["not initialized"]
+//- assertions
+assert.sameValue(Object.hasOwnProperty.call(C.prototype, "a"), false);
+assert.sameValue(Object.hasOwnProperty.call(C, "a"), false);
```
leobalter: you found a bug :)

it should be the lowercase c here

src/class-fields/default/cls-decl-wrapped-in-sc.template
leobalter: :) still a valid syntax

src/class-fields/static-computed-symbol-names.case
leobalter: these are checking for strings, not symbols. I should add an assertion message to clarify. I'll check the case file anyway.

src/class-fields/computed-names.case
leobalter: but these are the computed property name tests, right? I'm verifying `[x] = 42` won't produce any property named `"x"` in `c`.

src/class-fields/computed-names.case
rwaldron: oh lol, not even what I thought :P

src/class-fields/computed-names.case
rwaldron: > but these are the computed property name tests, right? 

I disagree with what you think this implies, but I don't really want to argue about it either. Please add `features: [computed-property-names]` and also add `computed-property-names` to https://github.com/tc39/test262/blob/master/features.txt

rwaldron:
> I'll defer this generation tool check to @rwaldron.

Lol, thanks.

leobalter:
> Lol, thanks.

It's fixed now, but I reached a point I thought my machine was broken so I needed someone to help.

leobalter:
fixes are up

littledan:
I don't see any more bugs in this test or subsequent fixups.

The task list is not yet complete; it's just a start. It seems like the next step would be to write it all out. Does anybody want to do that?

cc @xanlpz

leobalter:
@littledan yes, there's still a missing coverage for syntax errors, other observable failures, private names, etc.

I'm planning to have this landed so we can use it as a base for follow ups.

@rwaldron can we :shipit: ?

PR #1191, opened 2017-08-21
Tests for BitInt.asIntN

thejoshwolfe:
I intend to make more BigInt tests. This is what I have so far. I wanted to submit this PR to get feedback on organization/style/coverage/etc before duplicating and modifying these for BigInt.asUintN.


test/built-ins/BigInt/asIntN/bigint-tobigint.js
```diff
@@ -0,0 +1,26 @@
+// Copyright (C) 2017 Josh Wolfe. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: pending
+description: BigInt.asIntN type coercion for bigint parameter
+info: >
+  BigInt.asIntN ( bits, bigint )
+
+  2. Let bigint ? ToBigInt(bigint).
+features: [Symbol]
```
leobalter: this should add the `arrow-function` feature as well, or at least use regular functions instead.

test/built-ins/BigInt/asIntN/bits-toindex.js
```diff
@@ -0,0 +1,32 @@
+// Copyright (C) 2017 Josh Wolfe. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: pending
+description: BigInt.asIntN type coercion for bits parameter
+info: >
+  BigInt.asIntN ( bits, bigint )
+
+  1. Let bits be ? ToIndex(bits).
+features: [Symbol]
```
leobalter: this should add the `arrow-function` feature as well, or at least use regular functions instead.

test/built-ins/BigInt/asIntN/bigint-tobigint.js
```diff
@@ -0,0 +1,26 @@
...
+assert.throws(MyError, () => BigInt.asIntN(0, {valueOf(){throw new MyError();}}));
+assert.throws(TypeError, () => BigInt.asIntN(0, undefined));
+assert.throws(TypeError, () => BigInt.asIntN(0, null));
+assert.sameValue(BigInt.asIntN(2, true), 1n);
+assert.sameValue(BigInt.asIntN(2, false), 0n);
+assert.throws(TypeError, () => BigInt.asIntN(0, 0));
+assert.throws(SyntaxError, () => BigInt.asIntN(0, "foo"));
+assert.sameValue(BigInt.asIntN(3, "10"), 2n);
+assert.sameValue(BigInt.asIntN(4, "12345678901234567890003"), 3n);
+assert.throws(TypeError, () => BigInt.asIntN(0, Symbol("1")));
```
leobalter: the mixed sameValue and throws calls makes it so hard to read. I know this is esthetic, but it would help a lot if they are sorted.

test/built-ins/BigInt/asIntN/bigint-tobigint.js
```diff
@@ -0,0 +1,26 @@
...
+info: >
+  BigInt.asIntN ( bits, bigint )
+
+  2. Let bigint ? ToBigInt(bigint).
+features: [Symbol]
+---*/
+
+function MyError() {}
+
+assert.sameValue(BigInt.asIntN(3, Object(10n)), 2n);
```
leobalter: this is basically the same as testing with 10n, there's no coercion here, AFAICT.

There should be other tests - not for this PR - to test calling Object with a bigint.

test/built-ins/BigInt/asIntN/bits-toindex.js
```diff
@@ -0,0 +1,32 @@
...
+assert.sameValue(BigInt.asIntN(undefined, 1n), 0n);
+assert.sameValue(BigInt.asIntN(null, 1n), 0n);
+assert.sameValue(BigInt.asIntN(true, 1n), -1n);
+assert.sameValue(BigInt.asIntN(false, 1n), 0n);
+assert.sameValue(BigInt.asIntN("foo", 1n), 0n);
+assert.sameValue(BigInt.asIntN("3", 10n), 2n);
+assert.throws(TypeError, () => BigInt.asIntN(Symbol(0), 0n));
+assert.throws(TypeError, () => BigInt.asIntN(1n, 0n));
+assert.sameValue(BigInt.asIntN(Object(3), 10n), 2n);
+assert.sameValue(BigInt.asIntN({valueOf:()=>3}, 10n), 2n);
```
leobalter: I don't think checking `Object(3)` is necessary as we already have this custom object here. Removing the previous line (with the Object call) should be fine, otherwise, it could be explicit using `new Number`.

test/built-ins/BigInt/asIntN/bigint-tobigint.js
```diff
@@ -0,0 +1,26 @@
...
+assert.sameValue(BigInt.asIntN(3, {valueOf:()=>10n}), 2n);
+assert.throws(MyError, () => BigInt.asIntN(0, {valueOf(){throw new MyError();}}));
```
leobalter: This is also not observing calls to `@@toPrimitive` or calling the toString method. I believe these are worth to be in a new separate file, splitting into one for objects and other for direct primitives coercion, at least.

test/built-ins/BigInt/asIntN/bits-toindex.js
```diff
@@ -0,0 +1,32 @@
...
+assert.throws(MyError, () => BigInt.asIntN({valueOf(){throw new MyError();}}, 0n));
```
leobalter: Same as mentioned above:

> This is also not observing calls to @@toPrimitive or calling the toString method. I believe these are worth to be in a new separate file, splitting into one for objects and other for direct primitives coercion, at least.

test/built-ins/BigInt/asIntN/length.js
```diff
@@ -0,0 +1,18 @@
...
+  17 ECMAScript Standard Built-in Objects
+
+includes: [propertyHelper.js]
+---*/
+
+assert.sameValue(BigInt.asIntN.length, 2);
+
+verifyNotEnumerable(BigInt.asIntN, 'length');
+verifyNotWritable(BigInt.asIntN, 'length');
+verifyConfigurable(BigInt.asIntN, 'length');
```
leobalter: we're trying to get rid of these separate verify helpers. `verifyProperty` should be fine:

```js
verifyProperty(BigInt.asIntN, 'length', {
  value: 2,
  enumerable: false,
  writable: false,
  configurable: true
});
```

The same applies for other test files in this PR

test/built-ins/BigInt/asIntN/arithmetic.js
```diff
@@ -0,0 +1,68 @@
+// Copyright (C) 2017 Josh Wolfe. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: pending
```
leobalter: we already have an esid: `sec-integer.asintn`

https://tc39.github.io/proposal-bigint/#sec-integer.asintn

test/built-ins/BigInt/asIntN/arithmetic.js
```diff
@@ -0,0 +1,68 @@
...
+description: BigInt.asIntN arithmetic test cases
+info: >
+  BigInt.asIntN ( bits, bigint )
+
+  3. Let mod be a BigInt representing bigint modulo 2**bits.
+  4. If mod ≥ 2**bits - 1, return mod - 2**bits; otherwise, return mod.
```
leobalter: all the test files should add a `features: [BigInt, BigInt.asIntN]`.

These features should be added to the features list as well: https://github.com/tc39/test262/blob/master/features.txt

leobalter: I'm not sure if BigInt.asIntN is necessary to add here, maybe BigInt only. 

cc @rwaldron

test/built-ins/BigInt/asIntN/bigint-tobigint.js
littledan: `Object(10n)` creates a BigInt object, wrapping the primitive.

test/built-ins/BigInt/asIntN/arithmetic.js
littledan: +1.

test/built-ins/BigInt/asIntN/arithmetic.js
littledan: What's the current policy on esid--do you add it for things that are at Stage 3/4, or only once it's integrated into the main spec?

test/built-ins/BigInt/asIntN/bigint-tobigint.js
littledan: I'm not sure what you're getting at.  Do you mean that the test doesn't sufficiently assert that it's being reached? This is converting to a numeric type, so I believe it would call `valueOf`.

test/built-ins/BigInt/asIntN/bits-toindex.js
littledan: Does it hurt to have tests like this? In some places in JS, Number wrappers are treated differently from other objects.

littledan: I don't really understand the preference for `new Number` of `Object`.

test/built-ins/BigInt/asIntN/bigint-tobigint.js
leobalter: We observe it calling valueOf only, we don't have any observations for Symbol.toPrimitive or .toString calls when valueOf is not available

test/built-ins/BigInt/asIntN/bigint-tobigint.js
leobalter: in this case, we should use `new BigInt`.

test/built-ins/BigInt/asIntN/bits-toindex.js
leobalter: `new Number` is explicit. If we want to be precise at the point we compare an ordinary object with a valueOf method vs another object with a primitive value, this all goes back to the fact we need tests for Symbol.toPrimitive calls or .toString calls after valueOf.

test/built-ins/BigInt/asIntN/arithmetic.js
leobalter: esid should be pending when an id is not available. In this case, we can have a prediction in the proposal. `esid: pending` should be the last resource, as it's becoming common to see tests left this way without any fix by the original contributors. The sooner we have it, better. Otherwise, this will cause an extra maintenance cost.

test/built-ins/BigInt/asIntN/arithmetic.js
leobalter: I agree, we should follow only with BigInt for now. This can be easily fixed in the future if anyone provides another feedback.

test/built-ins/BigInt/asIntN/bigint-tobigint.js
littledan: `new BigInt` throws, whereas `Object(10n)` creates a wrapper.

test/built-ins/BigInt/asIntN/bigint-tobigint.js
littledan: I thought you were opposed to too many redundant combinations of tests, as you've been removing them from the repository. This test is trying to get at some of what's involved in type conversions, but doesn't do all possible tests. I don't think the distinctions you're raising are extremely likely to cause bugs.

test/built-ins/BigInt/asIntN/arithmetic.js
littledan: Sounds good to me.

test/built-ins/BigInt/asIntN/bigint-tobigint.js
leobalter: > I don't think the distinctions you're raising are extremely likely to cause bugs.

That's a common phrase before tests. :)

> I thought you were opposed to too many redundant combinations of tests, as you've been removing them from the repository. This test is trying to get at some of what's involved in type conversions, but doesn't do all possible tests.

I don't think these would be redundant, but they would check other easy to observe units (getting and calling other methods). With that, we could also check the order of them being called, as valueOf throwing a custom error after a valid @@toPrimitive return, e.g.

test/built-ins/BigInt/asIntN/bigint-tobigint.js
leobalter: oh I see. Yes, `Object(10n)` seems like the only way to create an object with an internal `[[BigIntData]]`.

test/built-ins/BigInt/asIntN/bigint-tobigint.js
thejoshwolfe: I thought of that too. I organized it here in a depth-first traversal order of the spec for all this. It was easier for me to make sure I was getting some decent coverage with this organization, but I can easily sort it to make it clearer to read.

leobalter: I might be nitpicking on this, so just do as you prefer. It's ok in both ways.

test/built-ins/BigInt/asIntN/arithmetic.js
thejoshwolfe: So the section won't be renamed from `integer` to `bigint`?

leobalter: I guess yes, so `sec-bigint.asintn`. If it's still cloudy, let's keep `esid: pending`. I just need to include it into my future tasks as soon as it is defined.

test/built-ins/BigInt/asIntN/arithmetic.js
thejoshwolfe: The `features.txt` mentions `Symbol` as a "standard language feature" that's "largely maintained for historical reasons". How important is it to declare my dependency on that feature?

And for the `arrow-function` issue below, is there any value in me avoiding arrow functions to avoid depending on arrow functions? Is there any concern for allowing a partially-conformant implementation to test their support for BigInt without also needing support for arrow functions? If not, I can easily just throw `arrow-function` into the features declaration.

leobalter: We are trying to flag every feature that date from ES6+. We missed a bunch for some ES6, but we still have a reasonable amount of Symbol uses.

test/built-ins/BigInt/asIntN/bits-toindex.js
thejoshwolfe: I can add tests with `Symbol.toPrimitive` methods. I originally shied away from writing thorough tests for ToPrimitive for such a distant method as `BigInt.asIntN`, but if that's what we want, I can do that.

I figured at some point implementations are going to be using the same ToPrimitive implementation that they got test coverage for in another context. It seems like more work to violate the details of ToPrimitive for only BigInt.asIntN than to just use a common ToPrimitive implementation. So it seems not very important to test every corner case of every reachable abstract operation from every user-facing API, but I can see the appeal of that approach as well.

Should I throw in some `Symbol.toPrimitive` tests? If so, I think I agree that they should be sectioned out into a separate test file.

leobalter: Regardless this might look remote, the tests are covering a new primitive type and a test for the ToPrimitive operation should be one of the most important parts here.

littledan: I agree, it's good to have extensive coercion tests. You're right.

I don't think we need to check in tests for all combinations of possible coercion tests x operations which coerce to BigInt. However, this is the first place where we're testing coercions exhaustively, so it's not a bad place to put them.

test/built-ins/BigInt/asIntN/bigint-tobigint.js
rwaldron: @leo yes, similar to: 

```js
Object("string") instanceof String
```

Which will have: 

```
[[PrimitiveValue]]: "string"
```

test/built-ins/BigInt/asIntN/bigint-tobigint.js
rwaldron: > I might be nitpicking on this

I don't think you are, but we can always clean it up later.

test/built-ins/BigInt/asIntN/arithmetic.js
rwaldron: I'm +1 to `BigInt` only. `BigInt.asIntN` is part of the whole `BigInt`

test/built-ins/BigInt/asIntN/bigint-tobigint.js
leobalter: yes, I was nitpicking, requesting something like an explicit `new String` call instead of using `Object`. 

Anyway, that cannot happen with the BigInt ctor.

test/built-ins/BigInt/asIntN/bits-toindex.js
thejoshwolfe: So are you saying that `BigInt.asIntN` is going to be the call that exhaustively tests the type coercion for ToBigInt? And then `BigInt.asUintN` would not need as thorough tests? Is that what we're suggesting?

leobalter: other methods will still use similar tests, but after this part is done, anything next should be easier.

thejoshwolfe: I just noticed that `? GetMethod(input, @@toPrimitive)` can throw a TypeError if the `[Symbol.toPrimitive]` is not callable. I should probably add a test for that too.

I'm getting nervous about writing such thorough tests under the name "BigInt" for an aspect of the spec that has nothing to do with the BigInt proposal. This kind of thing makes #1197 appealing.

leobalter: in a user level, this might not seem relevant, but the purpose of this project is to confirm the observable steps in the spec, in the order they are defined. This is crucial to mitigate inconsistency from browsers and even detect issues like those in the web reality matters.

In the moment we say it's relevant for the spec to say a TypeError should be thrown if `@@toPrimitive` is not callable, a test for that is important, and that's a part of the BigInt api being covered here.

Although, it's not necessary to write a complete coverage in a single Pull Request, it can be fixed through follow up Pull Requests and I'm happy to volunteer writing those.

PR #1198, opened 2017-08-24
Basic tests for BigInt

leobalter:
This is only a refactor of tests from #1148. I'm not trying neither willing to add coverage in this PR.


leobalter:
Merging this will allow me to follow with more coverage in the same parts that are not complete yet.

test/language/expressions/greater-than-or-equal/bigint-and-number.js
```diff
@@ -0,0 +1,36 @@
...
+assert.sameValue(1 >= 0n, true);
+assert.sameValue(0.000000000001 >= 0n, true);
+assert.sameValue(1n >= 1, true);
+assert.sameValue(1 >= 1n, true);
+assert.sameValue(0n >= 1, false);
+assert.sameValue(0 >= 1n, false);
+assert.sameValue(0 >= 0n, true);
+assert.sameValue(0n >= 0, true);
+assert.sameValue(1n >= Number.MAX_VALUE, false);
+assert.sameValue(Number.MIN_VALUE >= 0n, false);
```
anba: I don't understand why the expected value for this case should be `false` (also applies to the other comparison operators tests with `Number.MIN_VALUE`). `Number.MIN_VALUE` is positive, non-zero number, so shouldn't this comparison return `true`?

leobalter: You're right, I missed this assertion when updating the assertions from the similar `<=` file.

PR #1199, opened 2017-08-25
Add tests for invalid BigInt MVs

leobalter:
I also have tests on hold for valid literals but I want to wait until a final decision for https://github.com/tc39/proposal-bigint/issues/77 before releasing them.

cc @littledan (this is an easy one to review)


littledan:
LGTM

littledan:
As for a decision on tc39/proposal-bigint#77 , what kind of confirmation are you looking for? The patch has been merged. Do you need it to be discussed at TC39, or do you want more feedback from implementers or users on that thread?

leobalter:
The issue is enough. Now I can write/push other tests with invalid MVs.

I need to do this when I'm back from my vacation. Right now I only got my phone available so I can't push anything.

Issue #1197, opened 2017-08-24
Reusable harness utilities to test parameter type coercion


thejoshwolfe:
As I'm working some tests through review here https://github.com/tc39/test262/pull/1191#discussion_r134826752 , the question came up of how thoroughly to test type coercion for API parameters. Specifically, the first parameter to `BigInt.asIntN` is passed into the abstract operation `? ToIndex(bits)`. To test that implementations are doing *everything* in `ToIndex` on that parameter, a pretty large number of test cases needs to be written. Not just testing the RangeErrors in `ToIndex`, but also the `floor(abs(number))` in `ToInteger`, the TypeErrors in `ToNumber`, the user-definable hooks in `ToPrimitive` and `ToOrdinaryPrimitive`, etc. And then after writing all these tests, copy-paste them all for `BigInt.toUintN`, which also calls `? ToIndex(bits)`.

As an idealistic software engineer, I of course tried to think of how to reduce this duplication. Here's my first draft of an idea (that for sure needs a lot of cleanup before it would be acceptable): https://gist.github.com/thejoshwolfe/bb00255524eeca6396269e1221c96a4a

My first feelings about this kind of approach is that it's too complicated to be worth the maintenance effort.

On the plus side though, this makes the test harness parallel the spec a bit more closely. You can write test cases for the abstract operation `ToInteger` in one place (the generic harness utility function), and then when an API calls `ToInteger`, the API test can call the test for `ToInteger`. This means that when `ToNumber` adds a row that `BigInt` should throw a TypeError, it would be theoretically possible to add that case to one place in the tests and get coverage through many APIs.

In general, this approach should be possible for `ToIndex`, `ToInteger`, `ToNumber`, and many other type coercion abstract operations.

bterlson:
I like any test helper that encodes a best-practice in a way that can easily be leveraged by test authors. But I wonder if this sort of thing would be better implemented as a generator that creates individual test files for each scenario as is the current practice? It does help with finding exactly where errors occurred.

littledan:
If you can do this as a file that's placed in the `harness/` directory, then this would be following the pattern of many other tests. We can discuss the details further in a code review.

leobalter:
a helper or a generation cases would both be great. There's an implied complexity here where the operations will happen and how they affect the result.

The other problem is the clarity. A single test file failing is not the best usage. Any failure will prevent any of the remaining code to execute, split test files provides a better overview of what exactly is failing and perhaps implementation progress.

[...]

PR #1200, opened 2017-08-28
Add generic type coercion utility functions

thejoshwolfe:
Add extremely thorough tests for the `? ToInteger(position)` type coercion in String.prototype.indexOf. This is done mostly through generic code.

See #1197

I wanted to submit a PR just to get early feedback on my approach here.

### DIFF
--- a/harness/typeCoercion.js
+++ b/harness/typeCoercion.js
@@ -6,249 +6,219 @@ description: |
     operations like ToNumber.
 ---*/
 
-function getValuesCoercibleToIntegerZero() {
-  var result = [];
+function testCoercibleToIntegerZero(test) {
+  function testPrimitiveValue(value) {
+    test(value);
+    // ToPrimitive
+    testPrimitiveWrappers(value, "number", test);
+  }
 
-  var primitiveValues = [
-    // ToNumber
-    null,
-    false,
-    0,
-    "0",
+  // ToNumber
+  testPrimitiveValue(null);
+  testPrimitiveValue(false);
+  testPrimitiveValue(0);
+  testPrimitiveValue("0");
 
-    // ToInteger: NaN -> +0
-    undefined,
-    NaN,
-    "",
-    "foo",
-    "true",
+  // ToInteger: NaN -> +0
+  testPrimitiveValue(undefined);
+  testPrimitiveValue(NaN);
+  testPrimitiveValue("");
+  testPrimitiveValue("foo");
+  testPrimitiveValue("true");
 
-    // ToInteger: floor(abs(number))
-    0.9,
-    -0,
-    -0.9,
-    "0.9",
-    "-0",
-    "-0.9",
-  ];
-
-  // ToPrimitive
-  primitiveValues.forEach(function(zero) {
-    result.push(zero);
-    result = result.concat(getPrimitiveWrappers(zero, "number"));
-  });
+  // ToInteger: floor(abs(number))
+  testPrimitiveValue(0.9);
+  testPrimitiveValue(-0);
+  testPrimitiveValue(-0.9);
+  testPrimitiveValue("0.9");
+  testPrimitiveValue("-0");
+  testPrimitiveValue("-0.9");
 
   // Non-primitive values that coerce to 0:
   // toString() returns a string that parses to NaN.
-  result = result.concat([
-    {},
-    [],
-  ]);
-
-  return result;
+  test({});
+  test([]);
 }
 
-function getValuesCoercibleToIntegerOne() {
-  var result = [];
+function testCoercibleToIntegerOne(test) {
+  function testPrimitiveValue(value) {
+    test(value);
+    // ToPrimitive
+    testPrimitiveWrappers(value, "number", test);
+  }
 
-  var primitiveValues = [
-    // ToNumber
-    true,
-    1,
-    "1",
+  // ToNumber
+  testPrimitiveValue(true);
+  testPrimitiveValue(1);
+  testPrimitiveValue("1");
 
-    // ToInteger: floor(abs(number))
-    1.9,
-    "1.9",
-  ];
-
-  // ToPrimitive
-  primitiveValues.forEach(function(value) {
-    result.push(value);
-    result = result.concat(getPrimitiveWrappers(value, "number"));
-  });
+  // ToInteger: floor(abs(number))
+  testPrimitiveValue(1.9);
+  testPrimitiveValue("1.9");
 
   // Non-primitive values that coerce to 1:
   // toString() returns a string that parses to 1.
-  result = result.concat([
-    [1],
-    ["1"],
-  ]);
-
-  return result;
+  test([1]);
+  test(["1"]);
 }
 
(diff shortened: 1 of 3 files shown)


harness/typeCoercion.js
```diff
@@ -0,0 +1,254 @@
...
+  var methodNames;
+  if (hint === "number") {
+    methodNames = ["valueOf", "toString"];
+  } else {
+    methodNames = ["toString", "valueOf"];
+  }
+  return [
+    // precedence order
+    {
+      [Symbol.toPrimitive]: method,
```
leobalter: So, we have a need - and a plan - to flag files using ES6+ features, but how do we do this using a harness file? Maybe using the features flag here in this file for Symbol.toPrimitive? cc @rwaldron 

We don't have this requirement for harness files, but it would be nice to add an info tag with the ToInteger and ToNumber operations. This can be done in a follow up for sure.

test/built-ins/String/prototype/indexOf/position-tointeger.js
```diff
@@ -0,0 +1,30 @@
...
+  String.prototype.indexOf ( searchString [ , position ] )
+
+  4. Let pos be ? ToInteger(position).
+
+includes: [typeCoercion.js]
+---*/
+
+getValuesCoercibleToIntegerZero().forEach(function(zero) {
+  assert.sameValue("aaaa".indexOf("aa", zero), 0, "with value " + zero);
+});
```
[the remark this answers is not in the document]
leobalter: As an alternative, calling `getValuesCoercibleToIntegerZero()` could replace the own forEach call. 

i.e.:

```js
valuesCoercibleToIntegerZero(function(zero) {
  assert.sameValue("aaaa".indexOf("aa", zero), 0, "with value " + zero);
});
```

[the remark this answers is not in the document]
thejoshwolfe: I like the idea of passing in a callback instead of using `forEach`. One problem I had while developing these tests is when there was a failure (due to me misreading the spec), it was difficult to determine which test case was causing the problem. It makes a lot of sense to design the test cases so that stack traces are as useful as possible. It was pretty amusing to try to debug print values like `{valueOf() { throw new MyError(); }}`, because every attempt at stringification would throw. Useful stack traces seem like a much more practical solution to the problem than stringifying the object.

I'm a huge fan of long and clear variable names, but I know I'm an outlier in that regard. One concern here is that the functions are exposed at global scope, so I want their full purpose to be in the variable name. I'm also a bit concerned with publishing the helper functions I wrote in the harness file. So perhaps a solution could be exposing an exports object where the name of the object gives context. Perhaps something like `CoercionUtils.ToIntegerZero`. (Now that I write that, I see it's just as long as the original name. Maybe this isn't the best idea.)

harness/typeCoercion.js
thejoshwolfe: Some ideas:

1. Just blaze ahead without flagging the features. This is done in this PR for `Symbol.toPrimitive` and in `harness/testAtomics.js`
2. Try to have the JavaScript behave differently if features are available at runtime. This is what I did with `BigInt` in this PR.
3. Harness files declare feature dependencies, and those feature dependencies are implicitly brought in by any test that includes the harness file.

harness/typeCoercion.js
```diff
@@ -0,0 +1,254 @@
...
+    "true",
+
+    // ToInteger: floor(abs(number))
+    0.9,
+    -0,
+    -0.9,
+    "0.9",
+    "-0",
+    "-0.9",
+  ];
```
littledan: I don't quite understand the phrasing here. A particular test which uses this function will then use the values for a particular parameter, which is cast in a particular way. Wouldn't it make more sense to have distinct functions for getting values that coerce to zero via ToInteger, via ToLength, etc? Seems like this function implements the ToInteger one (should that figure into the name?); for ToLength, for example, you could also include -10, and for ToNumber, you would remove many of these.

harness/typeCoercion.js
littledan: @leobalter I think the feature flags for ES6 features are most useful when adding the feature to the test suite initially. At this point, Symbol.toPrimitive is implemented in most engines, and it's difficult to add ever in the future in transpilers, so I'm not sure it's the highest-priority feature flag to get perfectly right.

rwaldron: > Just blaze ahead without flagging the features.

This is problematic for consumers of test262 that want to target their test runs while developing specific features. 

> Harness files declare feature dependencies, and those feature dependencies are implicitly brought in by any test that includes the harness file.

This is problematic for consumers of test262 that expect all test files to have the full meta information available in the test file itself. 


--------------------------


> Try to have the JavaScript behave differently if features are available at runtime. This is what I did with BigInt in this PR.

Can you elaborate on this?

harness/typeCoercion.js
```diff
@@ -0,0 +1,224 @@
...
+    // ToPrimitive
+    testPrimitiveWrappers(value, "number", function(value) {
+      test(TypeError, value);
+    });
+  }
+
+  // ToNumber: Symbol -> TypeError
+  testPrimitiveValue(Symbol("1"));
+
+  if (typeof BigInt !== "undefined") {
```
thejoshwolfe: @rwaldron this is what I mean by checking if the feature is enabled at runtime. Something similar could be done for `typeof Symbol !== "undefined"` and even `type Symbol.toPrimitive !== "undefined"` perhaps. And in the event they are undefined, then skip the parts of the test that use the features.

harness/typeCoercion.js
thejoshwolfe: Does the latest commit clarify the situation or make it more obscure? https://github.com/tc39/test262/pull/1200/commits/2bcee866c2f75796403a3024792ad493752ae2c0

thejoshwolfe:
I threw in a test for the first parameter of `String.prototype.indexOf` and added some generic functions for testing coercion `ToString`. I'm pretty happy with how little effort it took to implement that.

If this is deemed to be too much creep in one PR, it's pretty easy to chop that commit off into its own PR. But I think for the sake of the generic utility functions, it's a good demonstration of the flexibility to have different entry points both making use of `testCoercibleToPrimitiveWithMethod`, for example.

I could add more entry points as well, such as `ToIndex` and `ToLength`, which will be useful for BigInt testing. Should we do those later or in this PR?

thejoshwolfe:
I will also be on PTO for the next few business days.

leobalter:
I like this harness. Would you like to write some unit tests for the new API?

cc @rwaldron any thoughts?

thejoshwolfe:
> Would you like to write some unit tests for the new API?

Yes! I didn't know unit tests were even a thing. Looks like they go in `test/harness/`. I'll add some to this PR.

leobalter:
I didn't say anything before to allow an easier prototyping, but we maintain unit tests for the harness api as well.

rwaldron:
This is now included in https://github.com/tc39/test262/pull/1221

PR #1201, opened 2017-08-29
Delete identifier-let-allowed-as-lefthandside-expression-non-strict.js

### LINKED ISSUE #1192: Duplicate tests for `for (let in {})`
https://github.com/tc39/test262/blob/8422147c17aa3f4061e524a8f88c251a3b371ecf/test/language/statements/let/syntax/identifier-let-allowed-as-lefthandside-expression-non-strict.js#L10

https://github.com/tc39/test262/blob/8422147c17aa3f4061e524a8f88c251a3b371ecf/test/language/statements/for-in/head-lhs-let.js#L27

these two tests are for the same thing: `let` is allowed to be an identifier in a for-in loop in non-strict mode.

mroch:
Fixes #1192

This is already covered by https://github.com/tc39/test262/blob/8422147c17aa3f4061e524a8f88c251a3b371ecf/test/language/statements/for-in/head-lhs-let.js#L27


rwaldron:
The file being suggested for removal is one of a static semantics pair: 

- [test/language/statements/let/syntax/identifier-let-allowed-as-lefthandside-expression-non-strict.js](https://github.com/tc39/test262/blob/d9f62e4ccf11b2248c2045fb980bf4255621bab0/test/language/statements/let/syntax/identifier-let-allowed-as-lefthandside-expression-non-strict.js)
- [test/language/statements/let/syntax/identifier-let-allowed-as-lefthandside-expression-strict.js](https://github.com/tc39/test262/blob/d9f62e4ccf11b2248c2045fb980bf4255621bab0/test/language/statements/let/syntax/identifier-let-allowed-as-lefthandside-expression-strict.js)

I know that @leobalter agreed to removing this file, but I'm challenging that agreement. If I were authoring a new JS engine and wanted to test my `let` syntax support, I would assume that the following would be sufficient for me verify against: 

```js
test262-harness --hostType myjs --hostPath `which myjs` test/language/statements/let/syntax/*.js
```

Removal of this test would break that intuition, for (IMO) no real gain.

mroch:
@rwaldron a fair point :) totally up to you guys, feel free to close this if you don't want it

a feature tag is obviously another way to tag all the `let` tests across various directories. if that existed then could move the strict part of the pair into the for-in dir.

mroch:
oh wait, a `let` feature does exist! in that case, a `--features` flag in test262-harness would be cool

rwaldron:
> oh wait, a let feature does exist! in that case, a --features flag in test262-harness would be cool

Agreed, we need this

rwaldron:
@mroch I just discussed this with @leobalter and we agreed that instead of removal, this test needs to be relocated to https://github.com/tc39/test262/tree/master/test/language/statements/for-in

Also, the frontmatter needs to be updated to include:

```
esid: sec-iteration-statements
info: |
  for ( [ lookahead ∉ { let [ } ] LeftHandSideExpression [?Yield, ?Await] in 
    Expression[+In, ?Yield, ? Await]) Statement[?Yield, ?Await, ?Return]
```

leobalter:
I'll do this change today on top of this PR.

leobalter:
A bit late but #1271 is up.

PR #1207, opened 2017-09-02
Fixes for async iteration tests

anba:
Note: The dstr-assignment-for-await/array-elem-iter-rtrn-close-null.case tests may not succeed in all engines, because some engines call IteratorNext before resolving the destructuring assignment target.

### DIFF
--- a/src/async-generators/yield-promise-reject-next-yield-star-async-iterator.case
+++ b/src/async-generators/yield-promise-reject-next-yield-star-async-iterator.case
@@ -28,4 +28,4 @@ iter.next().then(() => {
     assert.sameValue(done, true, "The value of IteratorResult.done is `true`");
     assert.sameValue(value, undefined, "The value of IteratorResult.value is `undefined`");
   }).then($DONE, $DONE);
-}).then($DONE, $DONE);
+}, $DONE).catch($DONE);
--- a/src/dstr-assignment-for-await/array-elem-iter-rtrn-close-null.case
+++ b/src/dstr-assignment-for-await/array-elem-iter-rtrn-close-null.case
@@ -4,6 +4,12 @@
 desc: >
     IteratorClose throws a TypeError when `return` returns a non-Object value
 info: |
+    AssignmentElement : DestructuringAssignmentTarget Initializer
+    1. If DestructuringAssignmentTarget is neither an ObjectLiteral nor an ArrayLiteral, then
+      a. Let lref be the result of evaluating DestructuringAssignmentTarget.
+      b. ReturnIfAbrupt(lref).
+    [...]
+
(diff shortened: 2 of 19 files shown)


rwaldron:
Thanks for these!

littledan:
About that note: I'm pulling in the latest test262 tests to V8, and as you say, V8 calls IteratorNext before resolving the destructuring assignment target. I'm having trouble finding what part of the spec calls for evaluating the destructing assignment target first. Looking at [ForIn/OfBodyEvaluation](https://tc39.github.io/ecma262/#sec-runtime-semantics-forin-div-ofbodyevaluation-lhs-stmt-iterator-lhskind-labelset), I see IteratorStep in 5.a, and then DestructuringAssignmentEvaluation in 5.g.i. That looks like IteratorNext before resolving the destructuring assignment target to me. What am I missing here? cc @caitp

anba:
The tests contain this code `for await ([ {}[yield] ] of [iterable]) { ... }`, and when `iterable` is destructured into `[ {}[yield] ]`, [12.15.5.3 Runtime Semantics: IteratorDestructuringAssignmentEvaluation](https://tc39.github.io/ecma262/#sec-runtime-semantics-iteratordestructuringassignmentevaluation) for `AssignmentElement : DestructuringAssignmentTarget Initializer` is evaluated. `DestructuringAssignmentTarget` is `{}[yield]`, so we first need to execute the `yield` before calling IteratorStep in 12.15.5.3, step 2.a. 

Well, at least I misread `[iterable]` as `iterable` the first time I read this code, so I thought that the elements of `iterable` are iterated over in the for-await-of loop, instead of that the for-await-of loop only performs a single iteration over the array `[iterable]`...

anba:
Given that (assignment) destructuring evaluation order isn't really part of the test, we could change the test as follows:
1. Add a `next()` method to the iterator which performs `next() { return [undefined]; }`.
1. Change the loop header to 
```js
  let a; // or directly `let [ a = yield ]` in the loop.
  for await ([ a = yield ] of [iterable]) {
    ....
```

Then the test still checks for the return value when calling `return()` and the test will also pass in V8.

littledan:
@anba I see, thanks for the explanation. Seems like this is part of V8's issue https://bugs.chromium.org/p/v8/issues/detail?id=4951 and not related to async iteration. That test change sounds useful to improve coverage in engines like V8, though I guess it's not critical; V8 eventually may decide to fix the other bug anyway.

rwaldron:
> we could change the test as follows:

@anba can you make a PR for these changes?

anba:
> @anba can you make a PR for these changes?

Okay, will do this week.

PR #1208, opened 2017-09-05
BigInt.parseInt tests

cxielarko:
Additional tests for BigInt.parseInt. These are converted from the existing tests for the global parseInt function.


test/built-ins/BigInt/parseInt/S18.2.5-2-1.js
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) 2012 Ecma International.  All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-bigint-parseint-string-radix
+description: >
+    parseInt - 'S' is the empty string when inputString does not
+    contain any such characters
```
leobalter: so many things to start:

- please fix the file names, don't use the legacy format.
- please don't reuse legacy descriptions
- the info tag is missing or broken in tests. It needs to contain the respective spec text or other information that won't fit in the description line.
- the copyright headers are not valid
- missing the features flags with the valid features.
- If you use arrow functions, include it into the features flags as well

test/built-ins/BigInt/parseInt/S18.2.5_A2_T1.js
```diff
@@ -0,0 +1,15 @@
+// Copyright 2009 the Sputnik authors.  All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+info: Operator remove leading StrWhiteSpaceChar
+esid: sec-bigint-parseint-string-radix
+description: "StrWhiteSpaceChar :: TAB (U+0009)"
+---*/
+
+assert.sameValue(BigInt.parseInt("\u00091"), BigInt.parseInt("1"));
```
leobalter: why not assert the result against the actual expected value and not other parseInt call?

test/built-ins/BigInt/parseInt/S18.2.5_A1_T7.js
```diff
@@ -0,0 +1,32 @@
...
+assert.sameValue(BigInt.parseInt(object), 1n);
+
+var object = {toString: function() {return 1}};
+assert.sameValue(BigInt.parseInt(object), 1n);
+
+var object = {valueOf: function() {return {}}, toString: function() {return 1}};
+assert.sameValue(BigInt.parseInt(object), 1n);
+
+var object = {valueOf: function() {return 1}, toString: function() {throw "error"}};
+assert.throws("error", BigInt.parseInt(object));
```
anba: Two issues here:
- assert.throws() expects a constructor as its first argument
- and assert.throws() expects a function to execute as its second argument

test/built-ins/BigInt/parseInt/S18.2.5_A3.1_T7.js
```diff
@@ -0,0 +1,32 @@
...
+assert.sameValue(BigInt.parseInt("11", object), BigInt.parseInt("11", 2));
+
+var object = {toString: function() {return 2}};
+assert.sameValue(BigInt.parseInt("11", object), BigInt.parseInt("11", 2));
+
+var object = {valueOf: function() {return {}}, toString: function() {return 2}}
+assert.sameValue(BigInt.parseInt("11", object), BigInt.parseInt("11", 2));
+
+var object = {valueOf: function() {throw "error"}, toString: function() {return 2}};
+assert.throws("error", BigInt.parseInt("11", object));
```
anba: Same problem in this line.

rwaldron:
@cxielarko _nudge_

leobalter:
ping

leobalter:
I should apply the fixes

cxielarko:
I'm sorry for the delay on updating this. I've pushed a new version incorporating the suggested changes.

cxielarko:
Changes from the initial patch:

 * files have been renamed
 * description, info, and features tags updated
 * copyright headers updated
 * assert.throws calls fixed
 * unnecessary parseInt calls replaced with literals
 * other minor improvements, e.g. using propertyHelper.js

PR #1209, opened 2017-09-05
Fix assertion for Number.MIN_VALUE >= 0n

leobalter:
Ref https://github.com/tc39/test262/pull/1198#discussion_r137060514

cc @anba


rwaldron:
So fast. I was just reading through the Abstract Relational Comparison algorithm

anba:
Thanks!

The same issue is actually also present for the less-than, less-than-equals, and greater-than tests.

PR #1210, opened 2017-09-05
More fixes for bigint and MIN_VALUE comparisons

leobalter:
@anba this should end the issues with MIN_VALUE comparisons. The assertions are tricky to read, but I believe they are verified now.


anba:
The changes look good to me!

https://github.com/tc39/test262/blob/2f655036d4f969d9afd8eb36e7baedfbcd66f373/test/language/expressions/less-than/bigint-and-number.js#L34-L35 needs to be updated, too.

littledan:
I'm confused; why is this correct?

anba:
> I'm confused; why is this correct?

`Number.MIN_VALUE` is 5 &times; 10<sup>-324</sup>, so it should be greater than `0n`, shouldn't it?

littledan:
Oh sorry, of course, I was just confused about what Number.MIN_VALUE was.

PR #1212, opened 2017-09-06
Make 'wait' tests resilient on slow test systems

lars-t-hansen:
We have a couple of tests that check that when we wait in Atomics.wait and the wait is either broken by a wakeup or by a timeout, then the time waited is within a fairly small window of the waiting/timeout time.  On slow or overloaded tests systems these tests will fail spuriously because the actual time waited is greater.

These changes therefore make sure that the time waited is at least the expected time (minus epsilon), but there is no longer an upper bound.

### DIFF
--- a/test/built-ins/Atomics/wait/did-timeout.js
+++ b/test/built-ins/Atomics/wait/did-timeout.js
@@ -24,7 +24,7 @@ var ia = new Int32Array(new SharedArrayBuffer(Int32Array.BYTES_PER_ELEMENT));
 
 $262.agent.broadcast(ia.buffer);
 assert.sameValue(getReport(), "timed-out");
-assert.sameValue(Math.abs((getReport()|0) >= 500 - $ATOMICS_MAX_TIME_EPSILON, true);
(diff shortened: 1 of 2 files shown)


test/built-ins/Atomics/wait/did-timeout.js
```diff
@@ -24,11 +24,11 @@ var ia = new Int32Array(new SharedArrayBuffer(Int32Array.BYTES_PER_ELEMENT));
 
 $262.agent.broadcast(ia.buffer);
 assert.sameValue(getReport(), "timed-out");
-assert.sameValue(Math.abs((getReport()|0) - 500) < $ATOMICS_MAX_TIME_EPSILON, true);
+assert.sameValue(Math.abs((getReport()|0) >= 500 - $ATOMICS_MAX_TIME_EPSILON, true);
```
littledan: Missing a closing paren somewhere?

lars-t-hansen:
Oops.  It's not that the paren is missing, it's that there's half of a Math.abs call left in the code.  I've removed it.

PR #1220, opened 2017-09-08
intl: Tests for default options tweak proposal

littledan:
Tests for ECMA 402 PR https://github.com/tc39/ecma402/pull/170
The tests on Date/DateTimeFormat are valid without the PR.

### DIFF
--- a/test/intl402/Collator/default-options-object-prototype.js
+++ b/test/intl402/Collator/default-options-object-prototype.js
@@ -1,0 +1,20 @@
+// Copyright (C) 2017 Daniel Ehrenberg. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-initializecollator
+description: >
+  Monkey-patching Object.prototype does not change the default
+  options for Collator as a null prototype is used.
+info: >
+  InitializeCollator ( collator, locales, options )
+
+    1. If _options_ is *undefined*, then
+      1. Let _options_ be ObjectCreate(*null*).
+---*/
+
+if (new Intl.Collator("en").resolvedOptions().locale === "en") {
+  Object.prototype.sensitivity = "base";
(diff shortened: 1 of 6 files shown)


test/intl402/Collator/default-options-object-prototype.js
```diff
@@ -0,0 +1,20 @@
...
+  InitializeCollator ( collator, locales, options )
+
+    1. If _options_ is *undefined*, then
+      1. Let _options_ be ObjectCreate(*null*).
+---*/
+
+if (new Intl.Collator("en").resolvedOptions().locale === "en") {
+  Object.prototype.sensitivity = "base";
+  let collator = new Intl.Collator("en");
+  assert.sameValue(collator.resolvedOptions().caseFirst, "variant");
```
anba: Typo: caseFirst -> sensitivity

And I think I'd prefer this approach, because we won't skip the test on a hypothetical (!) implementation which doesn't support "en" and because I don't think the best-fit locale matcher is required to return the shortest possible locale ID (e.g. it could return "en-US" instead of "en").

```js
let defaultSensitivity = new Intl.Collator("en").resolvedOptions().sensitivity;

Object.prototype.sensitivity = "base";
let collator = new Intl.Collator("en");
assert.sameValue(collator.resolvedOptions().sensitivity, defaultSensitivity);
```

littledan: Thanks, switched to the suggested approach (the caseFirst/sensitivity thing was forgetting to amend the fix before uploading :/)

littledan:
Sorry, I forgot to add an "awaiting specification" label--these tests aren't valid without the ECMA 402 PR landing.

littledan:
The ECMA 402 PR has landed, so now it's valid that test262 has this patch landed.

Issue #1215, opened 2017-09-06
Remove obsolete feature flags


thejoshwolfe:
A number of feature flags are officially standard and are [largely maintained for historical reasons](https://github.com/tc39/test262/blob/be420f34f004fc014e9c2d701f421659a4d7434c/features.txt#L56) in this test suite. I propose we remove some/all standard flags from features.txt and from the `features:` declarations from test scripts.

## The value of old feature flags

Since "historical reasons" is pretty vague, let's discuss in more detail why old feature flags are valuable.

1. The flags already exist, and change is not free. It takes effort to remove the flags from test262, and it potentially takes effort for consumers to adapt to the change.
2. Consumers can filter which tests run while their implementation only supports a subset of the features. For example, someone starting a JavaScript implementation from scratch might not have arrow functions implemented yet, so they can skip all the tests that use arrow functions for the time being.

## The negative value of old feature flags

The presence of the old feature flags is not free.

1. Dragging around a bunch of obsolete feature flags puts a burden on contributors writing new tests. For example, if you test that passing a Symbol for a particular parameter throws a `TypeError`, you need to remember to note the dependency on the `Symbol` feature. As the number of features increases, this burden increases.
2. It's not always clear how to describe your dependence on particular features when harness files get involved. See [here](https://github.com/tc39/test262/pull/1200#discussion_r135683882). The fewer features exist, the smaller this problem becomes.

## Accountability and precedent

There are many examples where tests do not properly declare their feature dependencies. Here is one way to find tests that neglect to declare their dependency on the `Symbol` feature:

```
test262$ grep -PL 'features: \[.*\bSymbol\b.*\]| - Symbol$' $(grep -RPl '\bSymbol\(' test/) | wc -l
117
```

I may just be oblivious, but I don't think this negligence is causing any problems for consumers. We might infer from this that omitting the `Symbol` feature dependency is harmless, which lowers the resistance to removing that feature throughout the codebase.

## Git history is history

Depending on what "historical reasons" really means, we'll always have git history no matter what we do. If a consumer wants to try running this test suite and skip all the tests with a particular feature that has been removed, they might consider checking out an old git tag from before that feature was deleted.

## Immediate proposal

[...]

PR #1221, opened 2017-09-08
type coercion harness utilities + features flags + linting

rwaldron:
This builds on #1200 as described in https://github.com/tc39/test262/issues/1215#issuecomment-327900373

It's likely better to review this one commit at a time: 

1. https://github.com/tc39/test262/commit/65424be3ef1accc159c9a6b38f05493ab8ef8c8d
2. https://github.com/tc39/test262/commit/8a2ec342001b7fa11835f030d9da3342a8ac8f25
3. https://github.com/tc39/test262/commit/c7a5d21bee924f8f2ff7164b401e979fcc4e5348
4. https://github.com/tc39/test262/commit/fa610063130c7f421ecc7babeebffee83ff3d9e9
5. https://github.com/tc39/test262/commit/e3447b82001f36199dbd17f499bfad376d5e1a44


harness/assert.js
```diff
@@ -88,8 +88,8 @@ assert.throws = function (expectedErrorConstructor, func, message) {
 };
 
 assert.throws.early = function(err, code) {
-  let wrappedCode = `function wrapperFn() { ${code} }`;
+  let wrappedCode = 'function wrapperFn() { ' + code + ' }';
   let ieval = eval;
 
-  assert.throws(err, function() { Function(wrappedCode); }, `Function: ${code}`);
+  assert.throws(err, function() { Function(wrappedCode); }, 'Function: ' + code);
```
rwaldron: Sorry @leobalter ;)

harness/compareArray.js
```diff
@@ -20,5 +20,5 @@ function compareArray(a, b) {
 
 assert.compareArray = function(actual, expected, message) {
   assert(compareArray(actual, expected),
-         `Expected [${actual.join(", ")}] and [${expected.join(", ")}] to have the same contents. ${message}`);
+         'Expected [' + actual.join(', ') + '] and [' + expected.join(', ') + '] to have the same contents. ' + message);
```
rwaldron: Sorry again @leobalter ;)

harness/propertyHelper.js
```diff
@@ -29,47 +29,47 @@ function verifyProperty(obj, name, desc, options) {
...
+      failures.push('descriptor should ' + (desc.writable ? '' : 'not ') + 'be writable');
     }
   }
 
   if (Object.prototype.hasOwnProperty.call(desc, 'configurable')) {
     if (desc.configurable !== originalDesc.configurable ||
         desc.configurable !== isConfigurable(obj, name)) {
-      failures.push(`descriptor should ${desc.configurable ? '' : 'not '}be configurable`);
+      failures.push('descriptor should ' + (desc.configurable ? '' : 'not ') + 'be configurable');
     }
```
rwaldron: Ok,  @leobalter I'm not really sorry ;)

harness/compareArray.js
leobalter: Haha

features.txt
```diff
@@ -51,6 +51,7 @@ regexp-unicode-property-escapes
 
 # Shared Memory and atomics
 # https://github.com/tc39/ecmascript_sharedmem
+Atomics
```
littledan: Does it really make sense to have feature flags for parts of proposals? Atomics is part of the SharedArrayBuffer proposal. Do we expect implementations to do part of a language feature and separate it out this way?

leobalter: I'm +1 to use a single tag for each proposal, some exceptions can be considered, but in this case I believe it's ok to use the SharedArrayBuffer tag.

Wrt the exceptions, I guess this could possibly apply to private class fields, but we should postpone this discussion to a later time.

thejoshwolfe:
Is there a reason we're putting the `harness/` feature dependencies in an external file `features.yml` instead of in the harness files themselves? Couldn't the linter just parse the front matter of each dependent harness file for the transitive dependencies?

littledan:
@thejoshwolfe The harness files aren't test files, so they won't have the same frontmatter in general. Harness files are included in other files, so different logic is needed here. It might not be the prettiest, but it seems like the current system is a working solution.

features.txt
rwaldron: I added `Atomics` just to be thorough, since it's a new global. Is the consensus here that I should remove?

rwaldron:
> Is there a reason we're putting the harness/ feature dependencies in an external file features.yml instead of in the harness files themselves?

Dan's response is 100%, but I'll add some explanation re: the "frontmatter" that appears in the harness files: it's meaningless. I recently added that information just to keep track of what was in each file, I could've used jsdoc or similar—it wouldn't have made a difference.

features.txt
leobalter: I'm in favor of using a single name for features from the SharedArrayBuffer proposal. Even though, I believe we can change it as a follow up improvement.

If this is good to land, I don't want to block it, we can have a separate discussion to find an agreement for this specific part later.

leobalter:
@thejoshwolfe that's an interesting point, and Dan is correct. The harness files are conceptually replaceable, they only provide an API that is expected by the tests and those can run with anything similar. There are even examples of expected replacements, like `$ERROR` or API for Detached Array Buffers. The former needs to use the proper output call and Test262 can't use anything to really detach array buffers directly.

PR #1223, opened 2017-09-12
Add tests to verify valid object shorthand properties

### LINKED ISSUE #1202: Missing coverage: computed properties as shorthand props
The Flow parser incorrectly allowed object literals like `{0}` and `{[x]}` because it parsed the property's key and treated it as a shorthand definition when it didn't find a value.

The first case is covered already, the latter seems not to be. On one hand, it's not reasonable to test every bogus thing someone could write, but I think this one is worth testing because it's so similar to other property keys.

```js
({x})    // covered by language/expressions/object/prop-def-id-valid.js
({0})    // covered, indirectly, by while/if/do-while tests
({true}) // covered by language/reserved-words/ident-reference-false.js
({[x]})  // not covered
```

leobalter:
Fixes #1202


rwaldron:
Nice

PR #1224, opened 2017-09-13
BigInt.asUintN tests

thejoshwolfe:
* typeCoercion.js supports ToIndex
* typeCoercion.js supports ToBigInt
* updated BigInt.asIntN type coercion tests to use typeCoercion.js

I haven't run this through any implementation to test it yet, so there may be silly typos. I'll look into possibly doing that tomorrow.

### DIFF
--- a/test/built-ins/BigInt/asUintN/order-of-steps.js
+++ b/test/built-ins/BigInt/asUintN/order-of-steps.js
@@ -1,0 +1,32 @@
+// Copyright (C) 2017 Josh Wolfe. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: pending
+description: BigInt.asUintN order of parameter type coercion
+info: >
+  BigInt.asUintN ( bits, bigint )
+
+  1. Let bits be ? ToIndex(bits).
+  2. Let bigint ? ToBigInt(bigint).
+
+features: [BigInt]
+---*/
+
+var i = 0;
+var bits = {
+  valueOf() {
+    assert.sameValue(i, 0);
+    i++;
+    return 0;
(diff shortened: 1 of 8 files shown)


test/built-ins/BigInt/asUintN/order-of-steps.js
```diff
@@ -0,0 +1,32 @@
...
+};
+var bigint = {
+  valueOf() {
+    assert.sameValue(i, 1);
+    i++;
+    return 0n;
+  }
+};
+
+BigInt.asUintN(bits, bigint);
```
littledan: For good measure, you could assert that the value here is 0n (right?).

thejoshwolfe: Sure. I omitted that assert because it's not what this test is really about. I thought it would be clearer to a reader if they see that all the asserts have to do with execution order. But of course adding more asserts doesn't hurt.

I'd prefer to leave it the way it is, but I can change it if others think that would be better.

thejoshwolfe:
> I haven't run this through any implementation to test it yet, so there may be silly typos. I'll look into possibly doing that tomorrow.

I'm running into some trouble doing this. I don't think we should wait for this extra verification.

Can this be merged now?

anba:
> I'm running into some trouble doing this. I don't think we should wait for this extra verification.

I just ran the tests and only found one error. In built-ins/BigInt/asUintN/bits-toindex.js, the expected value for the `testCoercibleToIndexOne` test should be `1n` instead of `0n`.

thejoshwolfe:
> should be 1n instead of 0n

Thanks! fixed.

PR #1225, opened 2017-09-13
intl tests for proposed additional options in options bag

littledan:
This patch implements tests for the ECMA 402 PR at
https://github.com/tc39/ecma402/pull/175 . It should not land until that spec patch lands.

It is based on the test test/intl402/Collator/10.1.1_19_c.js

### DIFF
--- a/test/intl402/DateTimeFormat/numbering-system-calendar-options.js
+++ b/test/intl402/DateTimeFormat/numbering-system-calendar-options.js
@@ -1,0 +1,68 @@
+// Copyright 2012 Mozilla Corporation. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-initializedatetimeformat
+description: >
+    Tests that the options numberingSystem and calendar can be  set through
+    either the locale or the options.
+author: Norbert Lindenberg, Daniel Ehrenberg
+includes: [testIntl.js]
+---*/
+
+let defaultLocale = new Intl.DateTimeFormat().resolvedOptions().locale;
+
+let supportedNumberingSystems = ["latn", "arab"].filter(nu =>
+  new Intl.DateTimeFormat(defaultLocale + "-u-nu-" + nu)
+    .resolvedOptions().numberingSystem === nu
+);
+
+let supportedCalendars = ["gregory", "chinese"].filter(ca =>
+  new Intl.DateTimeFormat(defaultLocale + "-u-ca-" + ca)
+    .resolvedOptions().calendar === ca
+);
+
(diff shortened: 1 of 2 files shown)


leobalter:
Any updates here?

littledan:
Let's leave this open for another little bit; I need to update the patch and tests considering https://github.com/tc39/ecma402/pull/175#issuecomment-368948903

test/intl402/DateTimeFormat/numbering-system-calendar-options.js
```diff
@@ -0,0 +1,68 @@
+// Copyright 2012 Mozilla Corporation. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-initializedatetimeformat
+description: >
+    Tests that the options numberingSystem and calendar can be  set through
+    either the locale or the options.
+author: Norbert Lindenberg, Daniel Ehrenberg
+includes: [testIntl.js]
```
FrankYFTang: This PR need to tag with features:  ID since v8 is implementing this behind flag.

test/intl402/NumberFormat/numbering-system-options.js
```diff
@@ -0,0 +1,63 @@
+// Copyright 2012 Mozilla Corporation. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-initializenumberformat
+description: >
+    Tests that the options numberingSystem and calendar can be  set through
+    either the locale or the options.
+author: Norbert Lindenberg, Daniel Ehrenberg
+includes: [testIntl.js]
```
FrankYFTang: This PR need to tag with features:  ID since v8 is implementing this behind flag.

test/intl402/DateTimeFormat/numbering-system-calendar-options.js
littledan: I don't understand what you mean by "features: ID". You can also turn the flag on for these tests in test262.status.

FrankYFTang: ok got it. LGTM could you land this?

FrankYFTang:
@leobalter

littledan:
We've concluded in ECMA-402 meetings that https://github.com/tc39/ecma402/pull/175 is ready to land when the tests are ready, given that we have one implementation and repeated consensus. I think all of these are ready to land at this point.

FrankYFTang:
Could we land this so I can start testing my code in v8?

PR #1227, opened 2017-09-16
Split Annex-B parts of non-unicode-malformed.js into annexB folder

anba:
Moves the tests using Annex-B identity escapes into the annexB folder.


rwaldron:
LGTM.

PR #1235, opened 2017-09-20
additional BigInt JSON tests

cxielarko:
Additional JSON tests for BigInt, for JSON.stringify with a replacer function and the toJSON method.

### DIFF
--- a/test/built-ins/JSON/stringify/bigint-replacer.js
+++ b/test/built-ins/JSON/stringify/bigint-replacer.js
@@ -1,0 +1,24 @@
+// Copyright (C) 2017 Robin Templeton. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: JSON serialization of BigInt values with replacer
+esid: sec-serializejsonproperty
+info: >
+  Runtime Semantics: SerializeJSONProperty ( key, holder )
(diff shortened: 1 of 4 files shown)


littledan:
These tests look valid to me. While you're at it, it'd also be good to have a few tests that fully verify the ordering,
1. toJSON
2. replacer
3. unwrap
4. TypeError if it's a BigInt

(this was a recent spec change)

test/built-ins/JSON/stringify/bigint-replacer.js
```diff
@@ -0,0 +1,24 @@
...
+function replacer(k, v)
+{
+    if (typeof v === "bigint")
+        return "bigint";
+    else
+        return v;
+}
+
+assert.sameValue(JSON.stringify(0n, replacer), "\"bigint\"");
+assert.sameValue(JSON.stringify({x: 0n}, replacer), "{\"x\":\"bigint\"}");
```
rwaldron: Since there are updates to be made anyway (per @littledan's request), make all of these strings with single quotes so you can get rid of the escapes (both files)

cxielarko:
This now includes an ordering test for JSON.stringify, and also uses single quotes to avoid string escaping in some places.

PR #1236, opened 2017-09-22
Add cross-realm feature

leobalter:
This comes as a question as if we should flag tests requiring support for multiple realm operations. Adding this feature should be useful for filtering purposes.


rwaldron:
👍

PR #1238, opened 2017-09-25
Tag tests for U+180E and complete coverage for RegExp matching

leobalter:
Unicode 6.3.0 released a breaking change where U+180E (Mongolian Vowel Separator) is no longer considered a whitespace character.

While this is not fully fixed in known implementations, feature tagging this change is helpful to keep track on Test262 consumption.

I also added a test to cover this change on RegExp patterns for \s and \S.


leobalter:
oops, CI should fail. I need to list the feature as well.

test/built-ins/RegExp/S15.10.2.12_A1_T1.js
```diff
@@ -60,7 +60,7 @@ if (i5.replace(/\s+/g, "") !== o5) {
 var i6 = "";
 for (var j = 6144; j < 7168; j++)
```
anba: This loop also needs to be updated now that the expected string no longer contains U+180E.

leobalter: oh sure, fixing it right now

PR #1239, opened 2017-09-25
Follow up for tagging tests with U+180E

leobalter:
I just found these extras for U+180E, nothing else.


leobalter:
@anba I added a22ea7e to address the bug you found. Thanks for the accurate eyes!

PR #1240, opened 2017-09-25
Add test case for tc39/ecma262#1009

anba:
This is a test case for the normative change in tc39/ecma262#1009 (not yet approved or merged!).


rwaldron:
LGTM

@leobalter?

leobalter:
this can land after consensus is reached for tc39/ecma262#1009.

@anba @bterlson can we add it to agenda for the next meeting?

ljharb:
This can land now; per consensus in https://github.com/tc39/tc39-notes/blob/master/es8/2017-11/nov-30.md#8ia-arraybuffer-error-reporting-tweak-by-andre-bargull

PR #1241, opened 2017-09-25
BigInt ToBoolean type coercion test

thejoshwolfe:
Add ToBoolean support to typeCoercion.js harness, and test `0n` -> `false`, `1n` -> `true` among other tests.

### DIFF
--- a/harness/typeCoercion.js
+++ b/harness/typeCoercion.js
@@ -112,6 +112,46 @@ function testCoercibleToIntegerFromInteger(nominalInteger, test) {
   }
 }
 
+function testCoercibleToBooleanFalse(test) {
+  test(undefined);
+  test(null);
+  test(false);
+  test(0);
+  test(-0);
+  test(NaN);
+  test(0n);
+  test("");
+}
+
+function testCoercibleToBooleanTrue(test) {
+  test(true);
+  test(1);
+  test(-1);
+  test(Infinity);
+  test(-Infinity);
+  test(1n);
+  test("a");
+  test("true");
+  test("false");
+  test("0");
+  test(Symbol("1"));
+  test({});
+  test([]);
+
+  // ToBoolean does not call ToPrimitive.
+  // In all of these cases, it's just an object, which is truthy.
+  testPrimitiveWrappers(false, "number", test);
+  testPrimitiveWrappers(true, "number", test);
+  testPrimitiveWrappers(false, "string", test);
+  testPrimitiveWrappers(true, "string", test);
+
+  function notReallyAnError(error, value) {
+    test(value);
+  }
+  testNotCoercibleToPrimitive("number", notReallyAnError);
+  testNotCoercibleToPrimitive("string", notReallyAnError);
+}
+
 function testPrimitiveWrappers(primitiveValue, hint, test) {
   if (primitiveValue != null) {
     // null and undefined result in {} rather than a proper wrapper,
--- a/harness/typeCoercion.js
+++ b/harness/typeCoercion.js
@@ -231,10 +231,8 @@ function testNotCoercibleToNumber(test) {
   // ToNumber: Symbol -> TypeError
   testPrimitiveValue(Symbol("1"));
 
-  if (typeof BigInt !== "undefined") {
-    // ToNumber: BigInt -> TypeError
-    testPrimitiveValue(BigInt(0));
-  }
+  // ToNumber: BigInt -> TypeError
+  testPrimitiveValue(0n);
 
   // ToPrimitive
   testNotCoercibleToPrimitive("number", test);
@@ -293,10 +291,7 @@ function testCoercibleToString(test) {
   testPrimitiveValue(-Infinity, "-Infinity");
   testPrimitiveValue("", "");
   testPrimitiveValue("foo", "foo");
-
-  if (typeof BigInt !== "undefined") {
-    testPrimitiveValue(BigInt(0), "0");
-  }
+  testPrimitiveValue(0n, "0");
 
   // toString of a few objects
   test([], "");
--- a/harness/typeCoercion.js
+++ b/harness/typeCoercion.js
@@ -291,13 +291,10 @@ function testCoercibleToString(test) {
   testPrimitiveValue(-0, "0");
   testPrimitiveValue(Infinity, "Infinity");
   testPrimitiveValue(-Infinity, "-Infinity");
-  testPrimitiveValue(123.456, "123.456");
-  testPrimitiveValue(-123.456, "-123.456");
   testPrimitiveValue("", "");
   testPrimitiveValue("foo", "foo");
 
   if (typeof BigInt !== "undefined") {
-    // BigInt -> TypeError
     testPrimitiveValue(BigInt(0), "0");
   }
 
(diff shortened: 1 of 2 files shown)


harness/typeCoercion.js
```diff
@@ -112,6 +112,46 @@ function testCoercibleToIntegerFromInteger(nominalInteger, test) {
...
+function testCoercibleToBooleanFalse(test) {
+  test(undefined);
+  test(null);
+  test(false);
+  test(0);
+  test(-0);
+  test(NaN);
+  test(0n);
+  test("");
+}
```
rwaldron: I'm not sure that I agree with the benefit of the harness function here (which I recognize is similar to others that have been added for these coercion tests). It's not clear in the test file below what is actually being tested, without also seeking out this file and stepping through the layers that have been created. If the test file below contained the following, it would be both sufficient and more easily understood by those that would benefit from test262:

```js
assert.sameValue(Boolean(undefined), false);
assert.sameValue(Boolean(null), false);
assert.sameValue(Boolean(false), false);
assert.sameValue(Boolean(0), false);
assert.sameValue(Boolean(-0), false);
assert.sameValue(Boolean(NaN), false);
assert.sameValue(Boolean(0n), false);
assert.sameValue(Boolean(""), false);
```

harness/typeCoercion.js
```diff
@@ -112,6 +112,46 @@ function testCoercibleToIntegerFromInteger(nominalInteger, test) {
...
+  test(Infinity);
+  test(-Infinity);
+  test(1n);
+  test("a");
+  test("true");
+  test("false");
+  test("0");
+  test(Symbol("1"));
+  test({});
+  test([]);
```
rwaldron: Similar to my comment above, this would be sufficient in a single file: 

```js
assert.sameValue(Boolean(true), true);
assert.sameValue(Boolean(1), true);
assert.sameValue(Boolean(-1), true);
assert.sameValue(Boolean(Infinity), true);
assert.sameValue(Boolean(-Infinity), true);
assert.sameValue(Boolean(1n), true);
assert.sameValue(Boolean("a"), true);
assert.sameValue(Boolean("true"), true);
assert.sameValue(Boolean("false"), true);
assert.sameValue(Boolean("0"), true);
assert.sameValue(Boolean(Symbol("1")), true);
assert.sameValue(Boolean({}), true);
assert.sameValue(Boolean([]), true);
```

harness/typeCoercion.js
leobalter: the code being more verbose helps communicating what it is actually doing, as mentioned. I'd avoid adding too many extras to the harness file so it doesn't become more complex.

test/built-ins/Boolean/toboolean.js
```diff
@@ -0,0 +1,24 @@
...
+
+testCoercibleToBooleanFalse(function(falsy) {
+  assert.sameValue(Boolean(falsy), false);
+});
+
+testCoercibleToBooleanTrue(function(truthy) {
+  assert.sameValue(Boolean(truthy), true);
+});
+
+// `Boolean(value)` (without `new`) never throws.
```
leobalter: this line is not necessary

harness/typeCoercion.js
leobalter: to avoid ambiguity, I agree with @rwaldron

test/built-ins/Boolean/toboolean.js
```diff
@@ -0,0 +1,24 @@
...
+includes: [typeCoercion.js]
+---*/
+
+testCoercibleToBooleanFalse(function(falsy) {
+  assert.sameValue(Boolean(falsy), false);
+});
+
+testCoercibleToBooleanTrue(function(truthy) {
+  assert.sameValue(Boolean(truthy), true);
+});
```
leobalter: This should also be placed in 2 different files, at least. We are not trying to condense everything in the same test file. This is also not helpful if anything is failing.

harness/typeCoercion.js
```diff
@@ -231,10 +271,8 @@ function testNotCoercibleToNumber(test) {
   // ToNumber: Symbol -> TypeError
   testPrimitiveValue(Symbol("1"));
 
-  if (typeof BigInt !== "undefined") {
-    // ToNumber: BigInt -> TypeError
-    testPrimitiveValue(BigInt(0));
-  }
+  // ToNumber: BigInt -> TypeError
+  testPrimitiveValue(0n);
```
leobalter: are we just throwing this away and assuming this file will only be used with BigInt?

thejoshwolfe: We need to decide one way or another. In some places in this harness file, BigInt literals are used unconditionally. This check here was pointless, since the file would not parse without BigInt support already. Additionally, the `BigInt` feature is required for this harness file according `harness/features.yml`.

This was also a concern when my `String.prototype.indexOf` type coercion test required BigInt support back in #1200 and #1221. I'm open to discussing more versatility in this regard.

thejoshwolfe:
Really the main motivation for writing the `toboolean.js` test was to test that `Boolean(0n) === false` and `Boolean(1n) === true`. So maybe I should back up and just test that.

However, I'm a fan of the type coercion testing framework I've got going in `typeCoercion.js`; I think it's elegant, thorough, and clearly parallels the spec. Adding `ToBoolean` to the framework is a step in the direction I want it to go; eventually I'd like to see `typeCoercion.js` support all the `To[Type]` conversion functions.

I argue that what I've got in this PR is good organization. A test in the directory `test/built-ins/Boolean/` directory makes a reference to `ToBoolean` utility functions, which parallels the spec's definition of the `Boolean(value)` function making a reference to the `ToBoolean` abstract operation. Then in `harness/typeCoercion.js`, there's a list of test cases for `ToBoolean` conversion that parallels the spec's table of types listed in the definition of the `ToBoolean` abstract operation.

> It's not clear in the test file below what is actually being tested, without also seeking out this file and stepping through the layers that have been created.

To answer this directly, the test is testing that `Boolean(value)` is calling the abstract operation `ToBoolean(value)`. It's true that you can't tell what specific values are going into the `Boolean()` function, but that parallels the spec; it's not clear how the `Boolean(value)` function deals with specific types without stepping through the layers of abstraction in the spec. To me, this isn't a problem; this is good design both in the spec and in test262.

That's my argument for the elegance of this organization. You're also bringing up practical concerns, like having everything in one file means that if anything fails, then nothing else gets run. We can solve this problem at runtime without creating a combinatoric explosion of individual files. I'll make an independent PR with this idea.

thejoshwolfe:
> I'll make an independent PR with this idea.

#1245

littledan:
This test only uses Boolean right now, but there are a number of contexts that call ToBoolean. For example, `!!`, or `if` statements. Would checking in tests that do that justify using the coercion framework here?

leobalter:
> Would checking in tests that do that justify using the coercion framework here?

you're talking about a different thing, which doesn't apply here.

The current tests are checking the results for `Boolean` calls, which cannot be applied directly into the introduced harness functions.

What we have is:

1. a general check for different types that coerce to false or true
1. a harness helper and a file test - `test/built-ins/Boolean/toboolean.js` - that globs together all the units for type coercion and test Boolean at once, failing the general long-term purpose of Test262 to comprehend different units split in multiple files.
1. a check for a BigInt literal added to the harness helper, preventing Boolean itself to guarded on implementations that are not ready to implement it. Test262 would be demanding or forcing implementations to work on this or just fail for everything.

Rather than elegant, these tests are against the design we've been aiming for more than 2 years.  I'm not inclined to change the project structure design, neither if this starts with a single test file and not a more comprehensive structural reform.

PR #1242, opened 2017-09-26
Added test case to consider rest ordering field access

### LINKED ISSUE #1132: Missing test case for object rest restructuring order
We have the following test case for Object Spread to test property access ordering in spread operation https://github.com/tc39/test262/blob/master/src/spread/obj-spread-order.case, however we don't test it for rest restructuring. We should be doing that.

caiolima:
It closes #1132

We already have this test cases for Object Spread, and the same ordering needs to be tested on Object Rest Destructuring. I'm also sending generated tests.


rwaldron:
Awesome, thanks!

PR #1243, opened 2017-09-26
[bigint] ToNumeric type coercion tests via "-UnaryExpression"

thejoshwolfe:
With the BigInt spec changes, unary minus uses ToNumeric instead of just ToNumber. This test uses a unary minus expression to test the type coercion corner cases for ToNumeric.

This PR is based on #1241. Not sure I did all the paperwork properly to indicate and facilitate that.


thejoshwolfe:
Following up from the review feedback in #1241, this PR needs to take a different approach. I'll open a new PR for this issue.

PR #1248, opened 2017-09-27
Update iteration tests with respect to spec changes

caitp:
https://github.com/tc39/ecma262/pull/988 changes the iteration protocol
such that the "next" method is only loaded from the iterator object once
during the prologue of iteration, rather than during each step.

### DIFF
--- a/test/language/statements/for-of/iterator-next-reference.js
+++ b/test/language/statements/for-of/iterator-next-reference.js
@@ -1,41 +1,41 @@
 // Copyright (C) 2013 the V8 project authors. All rights reserved.
 // This code is governed by the BSD license found in the LICENSE file.
 /*---
-es6id: 13.6.4.13 S5.c
+esid: sec-getiterator
 description: >
-    The iterator's `next` method should be accessed with each iteration as per
-    the `IteratorStep` abstract operation (7.4.5).
-features: [Symbol.iterator]
+    The iterator's `next` method should be accessed only once with each
+    iteration as per the `GetIterator` abstract operation (7.4.1).
+features: [Symbol.iterator, for-of]
 ---*/
 
 var iterable = {};
 var iterator = {};
-var firstIterResult = { done: false };
-var iterationCount, invocationCount;
+var iterationCount = 0;
+var loadNextCount = 0;
 
 iterable[Symbol.iterator] = function() {
   return iterator;
 };
 
-iterator.next = function() { return { value: 45, done: false }; };
-iterationCount = 0;
-invocationCount = 0;
+function next() {
+  if (iterationCount) return { done: true };
+  return { value: 45, done: false };
+}
+Object.defineProperty(iterator, 'next', {
+  get() { loadNextCount++; return next; },
+  configurable: true
+});
+
 for (var x of iterable) {
   assert.sameValue(x, 45);
 
-  iterator.next = function() {
-    invocationCount++;
-
-    Object.defineProperty(iterator, 'next', {
-      get: function() {
-        $ERROR('Should not access the `next` method after iteration ' +
-          'is complete.');
-      }
-    });
-
-    return { value: null, done: true };
-  };
+  Object.defineProperty(iterator, 'next', {
+    get: function() {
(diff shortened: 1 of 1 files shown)


caitp:
picked out tests which need changing from https://chromium-review.googlesource.com/c/v8/v8/+/687997/4/test/test262/test262.status#544

I'll also add a PR for the async iteration tests which were affected seperately.

test/language/statements/for-of/iterator-next-reference.js
```diff
@@ -3,39 +3,33 @@
 /*---
 es6id: 13.6.4.13 S5.c
```
leobalter: can you replace this with the current esid? This test doesn't reflect the ES6 specs anymore.

test/language/statements/for-of/iterator-next-reference.js
```diff
@@ -3,39 +3,33 @@
...
-        $ERROR('Should not access the `next` method after iteration ' +
-          'is complete.');
-      }
-    });
-
-    return { value: null, done: true };
-  };
+  Object.defineProperty(iterator, 'next', {
+    get: function() {
+      $ERROR('Should not access the `next` method after the iteration ' +
```
leobalter: while you're here, would you mind to replace this `$ERROR` call to a `throw new Test262Error`?

This should not be a blocker.

test/language/statements/for-of/iterator-next-reference.js
```diff
@@ -3,39 +3,33 @@
...
+var iterationCount = 0;
 
 iterable[Symbol.iterator] = function() {
   return iterator;
 };
 
-iterator.next = function() { return { value: 45, done: false }; };
-iterationCount = 0;
-invocationCount = 0;
+iterator.next = function() {
```
leobalter: maybe in a follow up PR - which I can do - or here if you're up to: defining this over a ` Object.defineProperty(iterator, 'next',` get accessor should be better for catching it get's the method only once.

Again, this should not be a blocker.

test/language/statements/for-of/iterator-next-reference.js
caitp: ✔️

test/language/statements/for-of/iterator-next-reference.js
caitp: Is this in addition to the defineProperty below, or on its own?

test/language/statements/for-of/iterator-next-reference.js
caitp: ✔️

test/language/statements/for-of/iterator-next-reference.js
leobalter: the changes you did sounds great already.

littledan:
Is this the only test which changes due to the new iteration protocol? It might be a good follow-on to add some more tests, e.g., for spread, Array.from, Promise.all, and anyplace else that the iteration protocol is used. The parts of the spec updated in the patch could be a good guide.

leobalter:
> Is this the only test which changes due to the new iteration protocol?

If the tests are correct there should be no blocker to merge them.

> It might be a good follow-on to add some more tests, e.g., for spread, Array.from, Promise.all, and anyplace else that the iteration protocol is used.

We should probably find a list of all places this can also be observed. Are you interested to assemble this in a new issue?

Also, we have this extra coverage: https://github.com/tc39/test262/pull/1250

> The parts of the spec updated in the patch could be a good guide.

yes, fetch by GetIterator and IteratorNext calls are a good start

littledan:
> If the tests are correct there should be no blocker to merge them.

I'm not trying to block anything, just giving an area for future tests to be written.

> We should probably find a list of all places this can also be observed. Are you interested to assemble this in a new issue?

I'd recommend that whoever writes this patch goes through https://github.com/tc39/ecma262/pull/988 and writes a test for each bunch of code affected. It's great that there are async iteration tests, but that's not the only place. I don't think I'll be able to do this anytime soon, unfortunately.

PR #1250, opened 2017-09-28
Update async iteration tests with respect to pending spec change

caitp:
Update behaviour based on changes from
https://github.com/tc39/ecma262/pull/988. The actual spec change PR for
async iteration is not yet uploaded.

This does not include any changes to Async-from-Sync Iterator.

### DIFF
--- a/src/async-generators/yield-star-async-next.case
+++ b/src/async-generators/yield-star-async-next.case
@@ -208,35 +208,32 @@ iter.next("next-arg-1").then(v => {
   assert.sameValue(log.length, 9, "log.length");
 
   iter.next("next-arg-2").then(v => {
-    assert.sameValue(log[9].name, "get next");
-    assert.sameValue(log[9].thisValue.name, "asyncIterator", "get next thisValue");
+    assert.sameValue(log[9].name, "call next");
+    assert.sameValue(log[9].thisValue.name, "asyncIterator", "next thisValue");
+    assert.sameValue(log[9].args.length, 1, "next args.length");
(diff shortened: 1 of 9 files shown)


leobalter:
@caitp can we split out the helper from this PR into a new one? I'm ok to merge 
07dfe8c and ec32b6f, but I'm not sure about the helper itself.

The way Test262 is updated varies from each implementation and maintaining this tool here might be tricky for me as I'm not sure if we would have a standard target.

caitp:
Fine with me, or if you’d like to manually remove the last 2 commits thats fine too.

caitp:
Also, this should wait for the actual change to the proposal to make sure it’s correct

leobalter:
> Also, this should wait for the actual change to the proposal to make sure it’s correct

The respective PR already reached consensus, there's no need to wait as the test can be helpful if any implementor wants to run close to the contents release. This is a similar behavior of stage 3 proposals that are not part of the last specs draft.

cc @bterlson,  as the spec editor, do you have any objection to tests for "has consensus" PRs being merged before ecma262?

caitp:
@leobalter  the thing is that I am just guessing at the details of what will actually be changed :p I think it's very likely that this will end up being correct, but it's still unknown at this point. Maybe you guys that were actually sitting in the room discussing it can say for sure.

leobalter:
the label flagging consensus without anything else written in the PR reflects it was resolved without modifications. We should release the meeting notes at the end of the day tomorrow.

PR #1251, opened 2017-09-29
More thorough bigint relational comparison tests

thejoshwolfe:
Notably this introduces coverage for:

* bigint against `Infinity` and `NaN`, which are special cases in the new algorithm.
* `Number.MAX_VALUE` vs bigint values that are a little above and a little below it by a factor of 16.
* bigint against bigint, including values greater than `Number.MAX_SAFE_INTEGER` that only differ by `1n`.

Regarding `Number.MAX_VALUE`, my reading of the spec is that we should not test differing from that value by only `1`, since `Number.MAX_VALUE` is an approximate value. However, a factor of 16 above and below seems far enough way that any reasonable approximation should recognize the difference.

I'm trying a new approach to writing tests, where instead of using runtime abstraction to get thorough and consistent coverage, I'm using a python script to generate this code. I didn't include the generator script in this PR, because it seemed like a one-off thing that wouldn't be generally useful. If there is interest in including this generator in test262, [here](https://gist.github.com/thejoshwolfe/039e4102c76d251f089d4c8856aec31e)'s what I've got so far, and I already know the style would need cleaning up to fit test262's style (e.g. indentation width).

The generator approach made it very easy to separate tests into 16 separate files. I could separate these tests into even more files pretty easily if that'd be desired. For example, I could put `Infinity` and `NaN` into separate test files instead of the one file called `-non-finite`.


test/language/expressions/greater-than-or-equal/bigint-and-bigint.js
```diff
@@ -0,0 +1,37 @@
...
+description: Comparisons of BigInt and Number values
+esid: sec-abstract-relational-comparison
+info: |
+  ...
+  3. If both px and py are Strings, then
+    ...
+  4. Else,
+    a. Let nx be ? ToNumeric(px). Because px and py are primitive values evaluation order is not important.
+    b. Let ny be ? ToNumeric(py).
+    c. If Type(nx) is Type(ny), return ? Type(nx)::lessThan(nx, ny).
```
littledan: You're hitting this case, right? If so, maybe the below spec text doesn't need to be quoted. Also, you could quote BigInt::lessThan, which is always hit here.

test/language/expressions/greater-than-or-equal/bigint-and-bigint.js
```diff
@@ -0,0 +1,37 @@
...
+assert.sameValue(0n >= 1n, false);
+assert.sameValue(1n >= 0n, true);
+assert.sameValue(0n >= -1n, true);
+assert.sameValue(-1n >= 0n, false);
+assert.sameValue(1n >= -1n, true);
+assert.sameValue(-1n >= 1n, false);
+assert.sameValue(0x1fffffffffffff01n >= 0x1fffffffffffff02n, false);
+assert.sameValue(0x1fffffffffffff02n >= 0x1fffffffffffff01n, true);
+assert.sameValue(-0x1fffffffffffff01n >= -0x1fffffffffffff02n, true);
+assert.sameValue(-0x1fffffffffffff02n >= -0x1fffffffffffff01n, false);
```
littledan: Maybe test some more values, e.g., a combination of long and short BigInt values.

test/language/expressions/greater-than-or-equal/bigint-and-number-extremes.js
```diff
@@ -0,0 +1,38 @@
...
+    h. If the mathematical value of nx is less than the mathematical value of ny, return true, otherwise return false.
+features: [BigInt]
+---*/
+
+assert.sameValue(1n >= Number.MAX_VALUE, false);
+assert.sameValue(Number.MAX_VALUE >= 1n, true);
+assert.sameValue(1n >= -Number.MAX_VALUE, true);
+assert.sameValue(-Number.MAX_VALUE >= 1n, false);
+assert.sameValue(
+  0xfffffffffffff80000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000n >= Number.MAX_VALUE,
```
littledan: Would it make sense to have a test for the edge cases, namely the BigInt which represents one less than Number.MAX_VALUE, the BigInt which is Number.MAX_VALUE, and one greater than it? Also, for mixed Number/BigInt comparison tests, it would be great to have more tests of different magnitudes to get at what are likely to be different cases in implementations (e.g., test on a Number which is between 0 and 1, and therefore can't be represented faithfully as a BigInt).

thejoshwolfe: I addressed this in the OP of this PR, but it might be good to throw a comment in the file as well. What do you think?

leobalter:
> instead of using runtime abstraction to get thorough and consistent coverage, I'm using a python script to generate this code

Are you saying you can't achieve a thorough and consistent coverage this way?

test/language/expressions/greater-than-or-equal/bigint-and-number-extremes.js
littledan: Sorry, I missed the explanation there. I disagree with your reading of the spec--I think Number.MAX_VALUE is well-defined.

thejoshwolfe:
> Are you saying you can't achieve a thorough and consistent coverage this way?

Sorry. I'm saying that I am achieving the same level of runtime coverage. Instead of runtime logic, I'm using analogous logic in the code generator.

An example of this coverage is that the case of comparing `1n` against `0` is specified once in the generator, and 8 cases are output in the code: 4 different operators, and 2 different ways to order the operands.

test/language/expressions/greater-than-or-equal/bigint-and-number-extremes.js
thejoshwolfe: Alright, I'll tighten up the test.

test/language/expressions/greater-than-or-equal/bigint-and-bigint.js
thejoshwolfe: Good catch. The frontmatter needs to be specialized for the different files even more than this. I'll make that update.

rwaldron:
>  I'm using a python script to generate this code. I didn't include the generator script in this PR, because it seemed like a one-off thing that wouldn't be generally useful. 

Why not use the existing test generator tool?

rwaldron:
Please review the CI failures. The frontmatter is invalid.

```
Linting 16 files.
Linting complete. 4 errors found.
test/language/expressions/greater-than-or-equal/bigint-and-bigint.js: FRONTMATTER - No valid YAML-formatted frontmatter
test/language/expressions/greater-than/bigint-and-bigint.js: FRONTMATTER - No valid YAML-formatted frontmatter
test/language/expressions/less-than-or-equal/bigint-and-bigint.js: FRONTMATTER - No valid YAML-formatted frontmatter
test/language/expressions/less-than/bigint-and-bigint.js: FRONTMATTER - No valid YAML-formatted frontmatter
```

test/language/expressions/greater-than-or-equal/bigint-and-bigint.js
```diff
@@ -0,0 +1,56 @@
...
+  ...
+  3. If both px and py are Strings, then
+    ...
+  4. Else,
+    a. Let nx be ? ToNumeric(px). Because px and py are primitive values evaluation order is not important.
+    b. Let ny be ? ToNumeric(py).
+    c. If Type(nx) is Type(ny), return ? Type(nx)::lessThan(nx, ny).
+
+sec-numeric-types-bigint-lessThan
+BigInt::lessThan (x, y)
```
leobalter: missing spaces, this is causing the yaml parsing to fail.

thejoshwolfe:
> Please review the CI failures. The frontmatter is invalid.

Oops. Should have known that was invalid yaml.

thejoshwolfe:
> Why not use the existing test generator tool?

If you're talking about `tools/generator/generator.py`, then I don't think it's the right fit for this usecase. I don't think the template and case design can generate what I'm doing with switching the order of the operands and inverting the condition for the non-or-equal operators. I can look into trying to use the existing generator, but writing something quick from scratch seemed much easier.

PR #1253, opened 2017-09-30
Do not use BigInt literal without `features: [BigInt]`

### LINKED ISSUE #1252: typeCoercion.js includes BigInt literal without "features: [BigInt]"
New `typeCoercion.js` includes BigInt literal like `0n`. And it causes some test failures like `test262/test/built-ins/String/prototype/indexOf/position-tointeger.js` if VM does not support BigInt literal.
But `test262/test/built-ins/String/prototype/indexOf/position-tointeger.js` is essentially unrelated to BigInt.

Since BigInt is new feature, and since `typeCoercion.js` is not marked with `features: [BigInt]`, `BigInt(0)` is better way.

Constellation:
typeCoercion.js is included in several tests which are not marked as `features: [BigInt]`.
Since BigInt is new feature, we should not make the above unrelated tests failed due to
lack of BigInt syntax support.

Close #1252.

### DIFF
--- a/harness/typeCoercion.js
+++ b/harness/typeCoercion.js
@@ -330,8 +330,8 @@ function testCoercibleToBigIntZero(test) {
     testPrimitiveWrappers(value, "number", test);
   }
 
-  testCoercibleToBigIntFromBigInt(0n, test);
-  testPrimitiveValue(-0n);
+  testCoercibleToBigIntFromBigInt(BigInt(0), test);
+  testPrimitiveValue(-BigInt(0));
   testPrimitiveValue("-0");
   testPrimitiveValue(false);
   testPrimitiveValue("");
@@ -351,7 +351,7 @@ function testCoercibleToBigIntOne(test) {
     testPrimitiveWrappers(value, "number", test);
   }
 
-  testCoercibleToBigIntFromBigInt(1n, test);
+  testCoercibleToBigIntFromBigInt(BigInt(1), test);
   testPrimitiveValue(true);
 
   // toString() returns "1"


leobalter:
https://github.com/tc39/test262/issues/1252#issuecomment-333317786

Can you point any test including typeCoercion.js and not adding a feature tag for BigInt? This would be a linter problem we also need to solve.

I'm not sure replacing the literal value by the constructor is the best solution if neither are implemented. Anyway, I'm inclined to use this proposed change to at least avoid a parsing error. My only blocking question is: should we allow the test runner to fail at the end (using BigInt ctor) or early (using the literal form)?

Maybe the constructor allows a better handling for the unsupported feature and reading the features tags, but I'm afraid this might open a path for a bad practice.

I'll revisit this issue on the week days.

Constellation:
Oops, sorry, `test/built-ins/String/prototype/indexOf/position-tointeger.js` includes feature flag `BigInt` since it includes this file.

But I think this file should not require `BigInt` because this test is not using it.
Actually, this test is not marked as `BigInt` feature when this test is created at 29938e9525978c3e3601be3b30991bf1ddf5b832.

In addition, some places of typeCoercion.js take care of the environments missing `BigInt`. For example,
https://github.com/Constellation/test262/blob/d2724b9f438cbd79789bb2c41cdd219944e48217/harness/typeCoercion.js#L234-L237 is,

```javascript
  if (typeof BigInt !== "undefined") {
    // ToNumber: BigInt -> TypeError
    testPrimitiveValue(BigInt(0));
  }
```

Since `typeCoercion.js` is a set of utilities to test type coercion, I expect that we can replace existing type coercion tests with this file's utility. But I don't think it is good that making these tests failed if we do not support BigInt. This is because these tests are not for BigInt.

So, I think we need to take care of using new syntax feature (like BigInt literal) in a utitlty file.
I think either (1) using `typeof BigInt` check when using BigInt in `typeCoercion.js`, or (2) separating `BigInt` feature from `typeCoercion.js` is better.

leobalter:
This is accurate.

I’ll give a round check for more improvements.

We can start with a type checking for the constructor but if this trigger
other issues we should have a separate file, indeed.

Constellation:
Thank you :D

PR #1256, opened 2017-10-02
Improve tests for BigInt.prototype.valueOf

leobalter:


test/built-ins/BigInt/prototype/valueOf/return.js
```diff
@@ -3,16 +3,17 @@
...
 esid: sec-bigint.prototype.valueof
-description: BigInt.prototype.valueOf this value coercion
-info: >
+description: >
+  BigInt.prototype.valueOf returns the primitive BigInt value.
+info: |
   BigInt.prototype.valueOf ( )
 
   Return ? thisBigIntValue(this value). 
 includes: [typeCoercion.js]
```
leobalter: I need to remove the includes... That's why the linter triggered an error

leobalter:
@spectranaut the commits d729577 and b307b80 should fix the remaining parts.

test/built-ins/BigInt/prototype/valueOf/return.js
```diff
@@ -0,0 +1,18 @@
...
+  BigInt.prototype.valueOf ( )
+
+  Return ? thisBigIntValue(this value). 
+features: [BigInt]
+---*/
+
+var valueOf = BigInt.prototype.valueOf;
+
+assert.sameValue(valueOf.call(0n), 0n, "0n");
+assert.sameValue(valueOf.call(Object(0n)), 0n, "Object(0n)");
```
rwaldron: Thanks!

test/built-ins/BigInt/prototype/valueOf/this-value-invalid-primitive-throws.js
```diff
@@ -0,0 +1,63 @@
...
+}, "number (1)");
+
+assert.throws(TypeError, function() {
+  valueOf.call(NaN);
+}, "NaN");
+
+var s = Symbol();
+assert.throws(TypeError, function() {
+  valueOf.call(s);
+}, "symbol");
```
rwaldron: Very clear, glad to see the helper going away where possible. Thanks for taking the extra time to refine these.

littledan:
Note that although there was only one test checked in which used this infrastructure, there are more places in the spec where it would be applicable, namely in BigInt.prototype.toString and toLocaleString. cc @thejoshwolfe @cxielarko

leobalter:
There is a PR for toString already which I'm working to apply fixes at this very moment. We should have an update soon.

PR #1257, opened 2017-10-03
Equality comparison tests for bigint

thejoshwolfe:
This includes the same test cases for all four equality operators `==`, `!=`, `===`, `!==`. The test cases cover all new spec text in "Strict Equality Comparison" and "Abstract Equality Comparison".

It might be a little silly to have so many test cases for strict equality comparison with mismatched types. I simply generated the same test cases for both strict and non-strict comparison, so I can definitely remove some of the excess tests if that would be appropriate.

I did not do exhaustive `? ToPrimitive(value)` testing in this PR. I can submit that in a different PR if that is desired, but I omitted it here since it seems like it has little to do with bigint.


test/language/expressions/does-not-equals/bigint-and-incomparable-primitive.js
```diff
@@ -0,0 +1,25 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Non-strict inequality comparison of BigInt and miscellaneous primitive values
+esid: sec-abstract-equality-comparison
+info: |
+  ...
+  13. Return false.
+
+features: [BigInt]
```
leobalter: missing Symbol

test/language/expressions/does-not-equals/bigint-and-incomparable-primitive.js
```diff
@@ -0,0 +1,25 @@
+// Copyright (C) 2017 Josh Wolfe. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Non-strict inequality comparison of BigInt and miscellaneous primitive values
+esid: sec-abstract-equality-comparison
+info: |
+  ...
+  13. Return false.
```
leobalter: this is not useful

test/language/expressions/does-not-equals/bigint-and-object.js
```diff
@@ -0,0 +1,42 @@
...
+    a. Let n be StringToBigInt(y).
+    b. If n is NaN, return false.
+    c. Return the result of x == n.
+  7. If Type(x) is String and Type(y) is BigInt, return the result of y == x.
+
+features: [BigInt]
+---*/
+
+assert.sameValue(0n != {}, true);
+assert.sameValue({} != 0n, true);
```
leobalter: these tests can also use comparisons for objects with a `[[BigIntData]]` internal slot.

test/language/expressions/does-not-equals/bigint-and-bigint.js
```diff
@@ -0,0 +1,53 @@
...
+
+  sec-numeric-types-bigint-equal
+  BigInt::equal (x, y)
+
+    The abstract operation BigInt::equal with two arguments x and y of BigInt type returns true if x and y have the same mathematical integer value and false otherwise.
+
+features: [BigInt]
+---*/
+
+assert.sameValue(0n != 0n, false);
```
leobalter: It would be really useful to add a message argument in the assertions so we can keep track on what is happening if any of these fails. The message can just reflect the actual argument being tested, e.g.:

```js
assert.sameValue(0n != 0n, false, "0n != 0n");
```

Otherwise a failing assertion during an implementation phase might be hard to map.

test/language/expressions/does-not-equals/bigint-and-incomparable-primitive.js
spectranaut: It would probably be more accurate to go up one level in this test's frontmatter for these particular tests --  to do this you'd need two fixes:
(1) change the esid to: sec-equality-operators-runtime-semantics-evaluation
(2) change the info to: 
`
Runtime Semantics: Evaluation
  EqualityExpression : EqualityExpression != RelationalExpression 
  ...
    5. Let r be the result of performing Abstract Equality Comparison rval == lval.
    6. If r is true, return false. Otherwise, return true. 
`
Note that your tests actually tests the "!=" equality comparison, thus the change of the esid. And the info text is a bit more clear that you are testing the "false" results of the Abstract Equality Comparison.

test/language/expressions/equals/bigint-and-incomparable-primitive.js
```diff
@@ -0,0 +1,26 @@
+// Copyright (C) 2017 Josh Wolfe. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Non-strict equality comparison of BigInt and miscellaneous primitive values
+esid: sec-equality-operators-runtime-semantics-evaluation
+info: |
+  EqualityExpression : EqualityExpression == RelationalExpression
+    ...
+    5. Return the result of performing Abstract Equality Comparison rval == lval.
```
thejoshwolfe: This spec text doesn't seem very helpful to me. Is there something better to say here?

thejoshwolfe:
I added messages to all the asserts. It was pretty easy to augment my code generator to do that. Should I go back and do that for the relational comparison tests i submitted with #1251? I've still got that code generator sitting around.

Speaking of which, is there some place in this repo for these code generators? I don't want to distract anyone (including future developers) with unnecessary maintenance burdens, and the generators were written in a quick-and-dirty manner. But it still might be helpful to have them? I don't know.

leobalter:
We already have a code generation tool, which is also used to keep track of changes as it is integrated to the CI.

You can find it here: https://github.com/tc39/test262/tree/master/tools/generation

The src folder will have the templates and cases being used with this generation tool. It's also something created with inputs from other maintainers and users. I'd be happy to try making Test262 a fit for more options, as @mathiasbynens has another generation tool for his own cases.

test/language/expressions/does-not-equals/bigint-and-object.js
```diff
@@ -0,0 +1,56 @@
...
+  6. If Type(x) is BigInt and Type(y) is String,
+    a. Let n be StringToBigInt(y).
+    b. If n is NaN, return false.
+    c. Return the result of x == n.
+  7. If Type(x) is String and Type(y) is BigInt, return the result of y == x.
+
+features: [BigInt]
+---*/
+
+assert.sameValue(0n != Object(0n), false, "0n != Object(0n)");
```
leobalter: +1

test/language/expressions/equals/bigint-and-incomparable-primitive.js
```diff
@@ -0,0 +1,26 @@
...
+assert.sameValue(1n == undefined, false, "1n == undefined");
+assert.sameValue(undefined == 1n, false, "undefined == 1n");
+assert.sameValue(0n == null, false, "0n == null");
+assert.sameValue(null == 0n, false, "null == 0n");
+assert.sameValue(1n == null, false, "1n == null");
+assert.sameValue(null == 1n, false, "null == 1n");
+assert.sameValue(0n == Symbol("1"), false, "0n == Symbol(\"1\")");
+assert.sameValue(Symbol("1") == 0n, false, "Symbol(\"1\") == 0n");
+assert.sameValue(1n == Symbol("1"), false, "1n == Symbol(\"1\")");
+assert.sameValue(Symbol("1") == 1n, false, "Symbol(\"1\") == 1n");
```
leobalter: thanks for adding the assertion messages. Maybe in the next time we could mix single with double quotes to avoid the scapes. This doesn't need to block the current PR as it is.

test/language/expressions/equals/bigint-and-incomparable-primitive.js
spectranaut: Unfortunately, because you are testing false cases, I don't think there is more to say in the info. 

Thought experiment: 
(1) You could go an extreme route and copy the whole Abstract Equality Comparison to see that you are testing negative cases (cases that are not listed explicitly in the algorithm), but that potentially noisy/distracting. A curious tester could just look up Abstract Equaliity Comparison in the test themselves. 
(2) you could just include the "return false" line you had before, which is also a bit confusing (if the specs change -- specifically, if the line number changes and there is more than one "return false")

So, it seems like there is not much more to add to the info. Nonetheless your question made me realize the description of the "info" tag in the CONTRIBUTING.md file is out of date with the current best practices, so I'll fix that later!

Also this is just my opinion, and I only just started working on test262, other reviewers might have a different idea of what to include the info :)

test/language/expressions/equals/bigint-and-incomparable-primitive.js
thejoshwolfe: That's very easy to do. I actually did this style, because it's what the V8 infrastructure team prefers, but I agree that mixed quote styles looks a little better.

test/language/expressions/equals/bigint-and-incomparable-primitive.js
leobalter: (1), yes , the whole thing is certainly too much and distracting. 
(2), exactly, also the initial operation might be expanded or changed the Return false instruction might become even more unclear.

Thanks for volunteering on to improve the contributing, my bias with the project makes me assume the documentation is fine, fresh eyes are always great for improvements.

test/language/expressions/equals/bigint-and-incomparable-primitive.js
leobalter: seems easier to read, write and map.

PR #1260, opened 2017-10-04
Add a little helper for convenience in testing changes

caitp:
I split this out of #1250. I think it's a nice thing to have, as it makes it very easy
to test changes in VMs before submitting them to test262.

A more portable version would be nice, but this is tested and working well on OSX and
the windows linux subsystem, which is good enough for me.

### DIFF
--- a/tools/copy_tests.sh
+++ b/tools/copy_tests.sh
@@ -91,19 +91,19 @@ while [[ $# -gt 0 ]]; do
             ;;
         --head)
             # Add test files changed since the last commit
-            files+=( "$( git diff HEAD --name-only | grep "^test/" )" )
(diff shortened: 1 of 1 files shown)


tools/copy_tests.sh
```diff
@@ -0,0 +1,152 @@
...
+            shift
+            ;;
+        -n|--dryrun|--dry-run)
+            # Don't actually copy files, but instead list the files to copy
+            dry_run=1
+            shift
+            ;;
+        --head)
+            # Add test files changed since the last commit
+            files+=( "$( git diff HEAD --name-only | grep "^test/" )" )
```
azz: Do you want to add `--diff-filter=d` to exclude files deleted since the revision?

PR #1264, opened 2017-10-06
BigInt unary minus tests

thejoshwolfe:


thejoshwolfe:
On second thought, I think it might make more sense to split the non-primitive values into a different test file. In it's current form, the frontmatter looks awkward quoting two different sections in backward order with a double blank line between them. It would probably be better organization and granularity to split up this test.

PR #1265, opened 2017-10-06
Intl.DateTimeFormat hourCycle tests

zbraniecki:
Adds a test for https://github.com/tc39/ecma402/pull/135


littledan:
Tests seem correct to me.

Is there any way that the tests could format a particular time in a known numbering system (e.g., latn), in UTC, and check that they are wrapped around properly? All DateTimeFormats must contain an hour/minute format; you can format this way and check that the resulting string *contains* the properly wrapped hour.

PR #1267, opened 2017-10-10
CONTRIBUTING.md updates

spectranaut:
for @rwaldron and @leobalter and @jugglinmike !

I tried to update the contributing file a few places to reflect the advice I've gotten from all of you. After these edits I think the file is up-to-date, but it might be worth it for you all to take a quick read through the whole file.

One section I think is lacking is the "features" front matter tag. We should have some concrete recommendations about what necessitates the creation of a new feature (probably, for example, a new feature proposal to ECMA script will necessitate the creation of a new feature tag -- to be used in all the new tests for the new feature).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,11 +10,17 @@
 
 Test cases should be created in files that are named to identify the feature or API that's being tested.
 
-Take a look at these examples:
+There is no strict naming convention. The file names should be human readable, helpful and, ideally, consistent within a single directory. For examples:
 
 - `Math.fround` handling of `Infinity`: `test/built-ins/Math/fround/Math.fround_Infinity.js`
-- `Array.prototype.find` use with `Proxy`: `test/Array/prototype/find/Array.prototype.find_callable-Proxy-1.js`
+- `Array.prototype.find` use with `Proxy`: `test/built-ins/Array/prototype/find/Array.prototype.find_callable-Proxy-1.js`
 - `arguments` implements an `iterator` interface: `test/language/arguments-object/iterator-interface.js`
+
+See the following directory trees for further recommended examples:
+
+- [test/built-ins/Math/fround](./test/built-ins/Math/fround/)
+- [test/built-ins/WeakMap](./test/built-ins/WeakMap/)
+- [test/language/arguments-object](./test/language/arguments-object/)
 
 **Note** The project is currently transitioning from a naming system based on specification section numbers. There remains a substantial number of tests that conform to this outdated convention; contributors should ignore that approach when introducing new tests and instead encode this information using the [id](#id) frontmatter tag.
 
@@ -49,8 +55,6 @@ Test262 supports the following tags:
  - [**description**](#description) (required)
  - [**info**](#info)
  - [**negative**](#negative)
- - [**es5id**](#es5id)
- - [**es6id**](#es6id)
  - [**esid**](#esid) (required for new tests)
  - [**includes**](#includes)
  - [**timeout**](#timeout)
@@ -58,12 +62,15 @@ Test262 supports the following tags:
  - [**flags**](#flags)
  - [**features**](#features)
 
+The following tags are deprecated, but exist in old tests:
+
+ - [**es5id**](#es5id) (deprecated)
+ - [**es6id**](#es6id) (deprecated)
+
 #### description
 **description**: [string]
 
-This is one of two required frontmatter tags. It should be a short, one-line
-description of the purpose of this testcase.  This is the string displayed by
-the browser runnner.
+This is one of two required frontmatter tags. The description should be a short, one-line description of the purpose of this testcase. We suggested that the description be kept to less than 100 characters, but clarity is preferred over brevity.
 
 Eg: Insert &lt;LS&gt; between chunks of one string
 
@@ -105,18 +112,6 @@ For example:
     negative:
       phase: early
       type: ReferenceError
-
-#### es5id
-**es5id**: [es5-test-id]
-
-This tag identifies the section number from the portion of the ECMAScript 5.1 standard that is tested by this test.  It was automatically generated for tests that were originally written for the ES5 version of the test suite and are now part of the ES6 version.
-
-When writing a new test for ES6, it is only necessary to include this tag when the test covers a part of the ES5 spec that is incorporated into ES6. All other tests should specify the `es6id` (see below) instead.
-
-#### es6id
-**es6id**: [es6-test-id]
-
-This tag identifies the section number from the portion of the ECMAScript 6 standard that is tested by this test.
 
 #### esid
 **esid**: [spec-id]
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -49,21 +55,24 @@ Test262 supports the following tags:
...
+
 #### description
 **description**: [string]
 
-This is one of two required frontmatter tags. It should be a short, one-line
-description of the purpose of this testcase.  This is the string displayed by
-the browser runnner.
+This is one of two required frontmatter tags. It should be a short (less than
+100 characters), one-line description of the purpose of this testcase.  This 
+is the string displayed by the browser runnner.
```
leobalter: > This is the string displayed by the browser runnner.

We should remove this.

Maybe the 100 characters limit should be a relaxed as a convention, not a rule. What do you think?

CONTRIBUTING.md
```diff
@@ -168,36 +165,41 @@ This tag is for boolean properties associated with the test.
 
 Some tests require the use of language features that are not directly described by the test file's location in the directory structure. These features should be specified with this tag. See the `features.txt` file for a complete list of available values.
 
+#### es5id
+**es5id**: [es5-test-id]
+
+**Deprecated.**
+
+This tag identifies the section number from the portion of the ECMAScript 5.1 or ECMAScript 3 standard that is tested by this test. It was automatically generated for tests that were originally written for the ES5 (or earlier) version of the test suite and are now part of the ES6 version. You can use the es5id to discover the relevant portion of the ECMAScript standard by looking up the section number in previous publications of the specification:
```
leobalter: remove the `or ECMAScript 3`

CONTRIBUTING.md
```diff
@@ -168,36 +165,41 @@ This tag is for boolean properties associated with the test.
...
 assert.sameValue(actual, expected, message) | throw a new Test262Error instance if the first two arguments are not [the same value](https://tc39.github.io/ecma262/#sec-samevalue); accepts an optional string message for use in creating the error
 assert.notSameValue(actual, unexpected, message) | throw a new Test262Error instance if the first two arguments are [the same value](https://tc39.github.io/ecma262/#sec-samevalue); accepts an optional string message for use in creating the error
 assert.throws(expectedErrorConstructor, fn, message) | throw a new Test262Error instance if the provided function does not throw an error, or if the constructor of the value thrown does not match the provided constructor
 assert.throws.early(expectedErrorConstructor, code) | throw a new Test262Error instance if the provided code does not throw an early error, or if the constructor of the value thrown does not match the provided constructor. This assertion catches only errors that will be parsed through `Function(code)`.
+$ERROR(message) | construct a Test262Error object and throw it <br>**DEPRECATED** -- do not use new tests
```
leobalter: s/do not use new tests/do not use in new tests

or maybe tell to use one of the assertions or throw a `new Test262Error` instead?

CONTRIBUTING.md
```diff
@@ -239,15 +241,15 @@ var p = new Promise(function () { /* some test code */ });
 
 p.then(function checkAssertions(arg) {
   if (!expected_condition) {
-    $ERROR("failure message");
+    throw new Test262Error("failure message");
```
leobalter: `assert(expected_condition, "assertion message");` is even better than the whole if case.

CONTRIBUTING.md
```diff
@@ -239,15 +241,15 @@ var p = new Promise(function () { /* some test code */ });
...
   }
 
 }).then($DONE, $DONE);
 ```
 
 Function `checkAssertions` implicitly returns `undefined` if the expected condition is observed.  The return value of function `checkAssertions` is then used to asynchronously invoke the first function of the final `then` call, resulting in a call to `$DONE(undefined)`, which signals a passing test.
 
-If the expected condition is not observed, function `checkAssertions` throws a `Test262Error` via function $ERROR.  This is caught by the Promise and then used to asynchronously invoke the second function in the call -- which is also `$DONE` -- resulting in a call to `$DONE(error_object)`, which signals a failing test.
+If the expected condition is not observed, function `checkAssertions` throws a `Test262Error`.  This is caught by the Promise and then used to asynchronously invoke the second function in the call -- which is also `$DONE` -- resulting in a call to `$DONE(error_object)`, which signals a failing test.
```
leobalter: this description makes it compelling to leave the `throw new Test262Error` above, we can also consider a failing assertion will throw an exception, being caught by the Promise.

CONTRIBUTING.md
```diff
@@ -10,12 +10,18 @@
...
 
 - `Math.fround` handling of `Infinity`: `test/built-ins/Math/fround/Math.fround_Infinity.js`
-- `Array.prototype.find` use with `Proxy`: `test/Array/prototype/find/Array.prototype.find_callable-Proxy-1.js`
+- `Array.prototype.find` use with `Proxy`: `test/built-ins/Array/prototype/find/Array.prototype.find_callable-Proxy-1.js`
 - `arguments` implements an `iterator` interface: `test/language/arguments-object/iterator-interface.js`
 
+See the following directory trees for further recommended examples:
+
+- `test/built-ins/Math/fround`
+- `test/built-ins/weakmap`
```
leobalter: `WeakMap` for case sensitive file systems.

CONTRIBUTING.md
rwaldron: I think we can relax it to a "suggestion"?

CONTRIBUTING.md
rwaldron: @leobalter it's actually totally accurate.

CONTRIBUTING.md
```diff
@@ -10,12 +10,18 @@
...
+- `test/language/arguments-object`
```
rwaldron: As we discussed offline, these would make great links

CONTRIBUTING.md
rwaldron: Might be useful to link to the report https://github.com/tc39/test262/wiki/Test262-Technical-Rationale-Report,-October-2017#specification-reference-ids

CONTRIBUTING.md
```diff
@@ -297,14 +297,13 @@ Test cases and test templates specify meta-data using the same YAML frontmatter
 - test cases (`*.case`)
   - `template` - name of the sub-directory to locate templates for this test
   - `description` (see above)
+    // looks like this is actually "desc" !!! add how its made description with the appending of "name" in parens from the test template
```
leobalter: +1

CONTRIBUTING.md
```diff
@@ -297,14 +297,13 @@ Test cases and test templates specify meta-data using the same YAML frontmatter
...
   - `info` (see above)
   - `features` (see above; merged with value defined by test template)
 - test templates (`*.template`)
   - `path` - location within the published test hierarchy to output files created from this template
   - `name` - human-readable name of the syntactic form described by this template (used to generate the test file's `description` field)
   - `esid` (see above)
-  - `es5id` (see above)
-  - `es6id` (see above)
```
leobalter: 🎉

spectranaut:
I removed "WIP" from the title because these changes are ready to merge, I think! @leobalter @rwaldron

CONTRIBUTING.md
```diff
@@ -292,19 +295,23 @@ Some language features are expressed through a number of distinct syntactic form
...
+
+### test templates (`*.template`)
+Field | Description
+------|-------------
+`path` | location within the published test hierarchy to output files created from this template. This path will be ended with the name of the test case file. If path is "/test/language/template1-" and the test case is "cast1.js", the final location of the file will be "/test/language/template1-case1.js"
+`name` | human-readable name of the syntactic form described by this template. This text will be appended, in parentheses, to the end of the test cases `desc` field.
+`esid` | see the frontmatter definition of the "info" tag.
+`info` | see the frontmatter definition of the "info" tag. The generated test will have a have final "info" value which is this text concatenated at the end of the test case's "info" text.
+`features` | see the frontmatter definition of the "features" field. The generated test will have a final feature list in combination with the test case's feature field.
+any other valid frontmatter field | see the frontmatter definitions.
```
rwaldron: Tables!! I love it.

PR #1269, opened 2017-10-12
BigInt unary minus tests

thejoshwolfe:
obsoletes #1264

### DIFF
--- a/test/language/expressions/unary-minus/bigint-non-primitive.js
+++ b/test/language/expressions/unary-minus/bigint-non-primitive.js
@@ -1,0 +1,33 @@
+// Copyright (C) 2017 Josh Wolfe. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Unary minus for BigInt object wrappers
+esid: sec-unary-minus-operator-runtime-semantics-evaluation
+info: |
+  Runtime Semantics: Evaluation
+  UnaryExpression : - UnaryExpression
+
+  1. Let expr be the result of evaluating UnaryExpression.
+  2. Let oldValue be ? ToNumeric(? GetValue(expr)).
+  3. Let T be Type(oldValue).
+  4. Return ? T::unaryMinus(oldValue).
+
+features: [BigInt, Symbol.toPrimitive]
+---*/
+
+assert.sameValue(-Object(1n), -1n, "-Object(1n) === -1n");
+assert.notSameValue(-Object(1n), 1n, "-Object(1n) !== 1n");
+assert.notSameValue(-Object(1n), Object(-1n), "-Object(1n) !== Object(-1n)");
+assert.sameValue(-Object(-1n), 1n, "-Object(-1n) === 1n");
+assert.notSameValue(-Object(-1n), -1n, "-Object(-1n) !== -1n");
+assert.notSameValue(-Object(-1n), Object(1n), "-Object(-1n) !== Object(1n)");
+assert.sameValue(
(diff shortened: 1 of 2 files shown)


test/language/expressions/unary-minus/bigint-non-primitive.js
```diff
@@ -0,0 +1,33 @@
...
+---*/
+
+assert.sameValue(-Object(1n), -1n, "-Object(1n) === -1n");
+assert.notSameValue(-Object(1n), 1n, "-Object(1n) !== 1n");
+assert.notSameValue(-Object(1n), Object(-1n), "-Object(1n) !== Object(-1n)");
+assert.sameValue(-Object(-1n), 1n, "-Object(-1n) === 1n");
+assert.notSameValue(-Object(-1n), -1n, "-Object(-1n) !== -1n");
+assert.notSameValue(-Object(-1n), Object(1n), "-Object(-1n) !== Object(1n)");
+assert.sameValue(
+  -{[Symbol.toPrimitive]: function() { return 1n; }, valueOf: function() { $ERROR(); }, toString: function() { $ERROR(); }}, -1n,
```
leobalter: not necessary to change this here, but in future maybe abstracting these functions will help with less code:

```js
var err = function() { new Test262Error(); }
```

Plus, it's helpful to avoid $ERROR as it's a legacy we're trying to get rid of, but the project deserves better docs for that. Again, this is not a blocker.

thejoshwolfe: If I made that change, do you have an opinion on what the assert message should look like?

test/language/expressions/unary-minus/bigint-non-primitive.js
```diff
@@ -0,0 +1,33 @@
...
+  1. Let expr be the result of evaluating UnaryExpression.
+  2. Let oldValue be ? ToNumeric(? GetValue(expr)).
+  3. Let T be Type(oldValue).
+  4. Return ? T::unaryMinus(oldValue).
+
+features: [BigInt, Symbol.toPrimitive]
+---*/
+
+assert.sameValue(-Object(1n), -1n, "-Object(1n) === -1n");
+assert.notSameValue(-Object(1n), 1n, "-Object(1n) !== 1n");
```
leobalter: you already assert that `-Object(1n)` evaluates `-1n`. You don't need to assert any of the `nonSameValue`. The tests for `===` can capture the results for the primitives.

The same apply to the other `notSameValue` cases here.

test/language/expressions/unary-minus/bigint.js
```diff
@@ -0,0 +1,34 @@
...
+  BigInt::unaryMinus (x)
+
+  The abstract operation BigInt::unaryMinus with an argument x of BigInt type returns the result of negating x.
+
+  Note: There is only one 0n value; -0n is the same as 0n.
+
+features: [BigInt]
+---*/
+
+assert.sameValue(-0n, 0n, "-0n === 0n");
```
leobalter: I really like the info with the reminder `There is only one 0n value; -0n is the same as 0n.`. 

Nice test. 👍

leobalter:
those are not changes request, just comments.

test/language/expressions/unary-minus/bigint-non-primitive.js
leobalter: the actual method/trap?

```js
assert.sameValue(-{[Symbol.toPrimitive]: prim, valueOf: err, toString: err }, 'primitive from @@toPrimitive');
assert.sameValue(valueOf: prim, toString: err }, 'primitive from {}.valueOf');
assert.sameValue(toString: prim }, 'primitive from {}.toString');
```

PR #1270, opened 2017-10-12
Update RegExp named capture tests for spec change

littledan:
The RegExp named groups specification has changed to not throw
errors in certain cases. This patch updates the test262 tests to match
the new specification, and throws in an additional test that verifies
the interaction between named group syntax and other replacement.
The tests pass on a version of V8 which implements the new semantics.

https://github.com/tc39/proposal-regexp-named-groups/commit/92ceba518c2ab0d2811c2efa8248ed1b3f8b5506

### DIFF
--- a/test/built-ins/RegExp/named-groups/string-replace-nocaptures.js
+++ b/test/built-ins/RegExp/named-groups/string-replace-nocaptures.js
@@ -24,6 +24,7 @@ for (let flags of ["", "u"]) {
   assert.sameValue("bacd", "abcd".replace(re, "$2$1"));
   assert.sameValue("cd", "abcd".replace(re, "$3"));
   assert.sameValue("$<sndcd", "abcd".replace(re, "$<snd"));
+  assert.sameValue("$<sndacd", "abcd".replace(re, "$<snd$1"));
   assert.sameValue("$<42a>cd", "abcd".replace(re, "$<42$1>"));
   assert.sameValue("$<fth>cd", "abcd".replace(re, "$<fth>"));
   assert.sameValue("$<a>cd", "abcd".replace(re, "$<$1>"));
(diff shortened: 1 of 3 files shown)


rwaldron:
> The tests pass on a version of V8 which implements the new semantics.

Which version is that? I've reviewed the changes to spec, I'd like to see them in action.

test/built-ins/RegExp/named-groups/string-replace-nocaptures.js
```diff
@@ -24,6 +24,7 @@ for (let flags of ["", "u"]) {
   assert.sameValue("bacd", "abcd".replace(re, "$2$1"));
   assert.sameValue("cd", "abcd".replace(re, "$3"));
   assert.sameValue("$<sndcd", "abcd".replace(re, "$<snd"));
+  assert.sameValue("$<sndacd", "abcd".replace(re, "$<snd$1"));
```
leobalter: isn't this a case for `test/built-ins/RegExp/named-groups/string-replace-unclosed.js`? It seems like an unclosed name capture.

The test is not incorrect, thou. I wonder if it becomes hard to find it later if anything changes or navigating a test failure.

littledan: It's not really unclosed, since it was never opened :) In particular, the RegExp doesn't have any captures. That's why it's in the -nocaptures file.

littledan: About finding the test: I don't really know how you do it, but I have trouble finding tests by name; the directory is useful, but beyond that, I just find myself grepping around. I guess that informs how I divide up tests into files.

littledan:
@rwaldron I just compiled V8 at tip-of-tree.

rwaldron:
@littledan thanks

PR #1244, opened 2017-09-26
BigInt.prototype.toString

cxielarko:
This is an initial set of tests for BigInt.prototype.toString based on existing tests for Number.prototype.toString.

PR #1272, opened 2017-10-12
Updated tests for BigInt.prototype.toString

leobalter:
Replaces #1244

Applying fixes for #1244, the original commit is preserved.


leobalter:
Not covered in this PR:

1. `this` value capturing objects with an `[[BigIntData]]` internal. (e.g.: `BigInt.prototype.toString.call(Object(1n))`).
2. ToInteger coercions and abrupt completions from radix.
3. Observable variants for the same this value using different radix, as long as the implementation defined spec permits.

rwaldron:
> Not covered in this PR:

@leobalter Can you open a new issue with this list? Thanks!

leobalter:
done #1275

PR #1273, opened 2017-10-12
class fields early errors tests

spectranaut:


src/class-fields/initializer-error/cls-decl-fields-comp-name.template
```diff
@@ -0,0 +1,15 @@
+// Copyright (C) 2017 Valerie Young. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: static-semantics-early-errors
```
leobalter: sec-class-definitions-static-semantics-early-errors

This should be the final id for this section.

src/class-fields/initializer-error/cls-decl-fields-static-literal-name.template
```diff
@@ -0,0 +1,14 @@
...
+/*---
+esid: static-semantics-early-errors
+path: language/statements/class/fields-static-literal-
+name: static literal ClassElementName
+---*/
+
+throw "Test262: This statement should not be evaluated.";
+
+class C {
+  x = /*{ initializer }*/;
```
leobalter: not static :)

littledan:
Thanks for your help here!

Some especially tricky cases for implementations might be when super call or arguments are contained within eval or arrow functions. Would it make sense to add tests for those?

A lot of people are asking if `new.target` will be similarly banned. When you or someone else writes further tests for what is allowed, it will be good to have that included (it should always evaluate to undefined, just like in any method).

cc @xanlpz

leobalter:
@littledan this feedback is very important, I'm glad the test generation tool allow us to expand these cases easily. 

As a starter, we are trying to follow the explicit cases in the [static semantics](https://tc39.github.io/proposal-class-fields/#sec-static-semantics-early-errors) section, but the limits of expanding the grammar might end as subjective. The feedback on specific expansions as arrow functions and eval is fundamental to not miss the fundamental parts.

for the new.target and similar specific cases: would you mind filing separate issues for these? This way we guarantee we don't miss anything valuable and we can also have small but easy to review PRs. A single issue with a list for these specific cases is also a nice alternative, if it works better. I just want to make sure we keep track and don't miss this.

spectranaut:
Thanks @littledan for the suggestion! Added eval and arrow function tests.

leobalter:
these are great for now, we should expand the coverage in follow up PRs

PR #1280, opened 2017-10-16
typeof test for BigInt

cxielarko:


rwaldron:
Verified https://tc39.github.io/proposal-bigint/#sec-typeof-operator-runtime-semantics-evaluation

PR #1283, opened 2017-10-16
BigInt arithmetic tests

cxielarko:
This is an initial set of BigInt tests for addition, subtraction, multiplication, division, remainder and exponentiation, checking the results of operators with operands of various sizes.


littledan:
The way the tests are organized seems reasonable to me. I verified the tests by running them against V8 at tip of tree; they all pass except for exponentiation, due to exponentiation being unimplemented. Thanks for upstreaming these tests--I think they will be useful, especially as it turns out that at least some implementers are writing their own new BigInt libraries.

PR #1288, opened 2017-10-17
Make Atomics.wake test cases resilient on slow / overloaded systems

lars-t-hansen:
On slow and overloaded systems, it is possible for some of the fixed timeout limits in these tests to be insufficient because worker threads take a long time to become available and they can become available in weird orders.  I fixed some of these issues earlier; here are some more fixes.  The general theme is to always wait until all the worker threads have reported that they are available.  In addition I've cleaned up a little.

These tests are still not perfect, there is still a chance of timing out.  But making them even more robust than here makes it too likely that bugs will be hidden by the robustness mechanisms.

### DIFF
--- a/test/built-ins/Atomics/wake/wake-all-on-loc.js
+++ b/test/built-ins/Atomics/wake/wake-all-on-loc.js
@@ -8,12 +8,20 @@ description: >
   wake waiters on other locations.
 ---*/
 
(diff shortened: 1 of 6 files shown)


anba:
I've already reviewed the original patch at https://bugzilla.mozilla.org/show_bug.cgi?id=1356109#c11, so this should be good to go.

PR #1291, opened 2017-10-17
typeof fixups

rwaldron:
I started out adding cases to bigint, but decided to put in a bit of additional effort to fixup all of the tests (but not the one syntax test, which can be done in a follow up)

This also surfaced a bug in V8's BigInt implementation: https://bugs.chromium.org/p/v8/issues/detail?id=6950


test/language/expressions/typeof/bigint.js
```diff
@@ -3,13 +3,50 @@
...
+assert.sameValue(
+  typeof 0n,
+  "bigint",
+  "typeof 0n === 'bigint'"
+);
+assert.sameValue(
+  typeof BigInt(0n),
+  "bigint",
+  "typeof BigInt(0n) === 'bigint'"
+);
```
rwaldron: This one presently fails: 

```
 typeof BigInt(0n) === 'bigint' Expected SameValue(«undefined», «bigint») to be true
```

PR #1292, opened 2017-10-17
Fix Async-from-Sync Iterator tests

caitp:
This change updates test cases to assume that the "next" method is only
loaded from a synchronous iterator once, and is re-used for each call to
Async-from-Sync Iterator.next(), based on tc39/ecma262#988


rwaldron:
@caitp awesome :)

PR #1293, opened 2017-10-17
fix: esid in case file instead of template file

spectranaut:
@leobalter -- moved the esids for you!


rwaldron:
@spectranaut the next step for generated tests is to run `./make.py` and then make a second commit with some message like "Generate tests" (or whatever). Once you push that, CI should pass

PR #1296, opened 2017-10-17
CONTRIBUTING.md: More detail about "features"

rwaldron:


CONTRIBUTING.md
```diff
@@ -161,7 +161,7 @@ This tag is for boolean properties associated with the test.
 #### features
 **features**: [list]
 
-Some tests require the use of language features that are not directly described by the test file's location in the directory structure. These features should be specified with this tag. See the `features.txt` file for a complete list of available values.
+Some tests require the use of language features that are not directly described by the test file's location in the directory structure. These features should be specified with this tag. See the `features.txt` file for a complete list of available values. This tag is required for new tests written for new features, but contributions will not be "blocked" if the tag is missing from frontmatter. The committing maintainer is required to ensure that the tag is present and contains the correct feature names; this can be done in an follow up commit.
```
foolip: Which consumers of test262 use the features metadata? It says above that it's required, here it says it's not really required, and it is something that seems to come up fairly often in recent PRs I've looked at.

littledan: V8 uses some of the feature flags to switch off tests or switch on flags, expect the test to pass or fail, require/disable strict mode, or use modules; e.g., see https://github.com/v8/v8/blob/master/test/test262/testcfg.py#L43 . However, other parts of the front matter are purely documentation.

foolip:
@ajklein, is it important for V8 that feature flags are maintained early in the process? Is it just because fixing them is a lot of work now, and would it still be important with two-way sync?

Basically, I'm wondering if this metadata can be added by the people who need it when they need it, whether that's V8 or some other team.

ajklein:
@foolip What do you mean by "early in the process"? At the moment, test262 only accepts features that have reached Stage 3 at TC39, at which point they're relatively stable. But I do find it useful, even at that point, to have them feature-flagged, for the same reason @littledan mentions.

foolip:
I mean in code review, before the tests reach V8.

Now I'm intrigued: what do you do if you want to implement before tests are allowed into test262? Is that common/uncommon?

ajklein:
It's uncommon. Before stage 3, it's expected that a feature might change significantly, which generally makes engines reluctant to invest time in implementation. This is a place where Blink and V8's culture around new feature development is interestingly different.

CONTRIBUTING.md
rwaldron: > Which consumers of test262 use the features metadata?

Presently V8, SpiderMonkey, ChakraCore, Prepack and Flow. Babylon, Babel and JSHint have "works in progress" that use `features`. Features will be used for categorizing/filtering the dashboard. 

> It says above that it's required, here it says it's not really required

This is a very new compromise. There has been some feedback that typing those things is too much of a contributor burden, so Test262 is committing to taking on the burden when a contributor is unwilling to contribute according to the needs of the project. There is an effort underway to automate the process of tagging features. 

> is something that seems to come up fairly often in recent PRs I've looked at.

Yep, usually this is just a matter of asking for the addition.

rwaldron:
`features` are also important for new projects that are still implementing features, as it allows them to create filters which include or exclude tests based on features. The goal is that eventually all test files will be tagged with relevant `features` found in that file.

Issue #1286, opened 2017-10-16
Document how to install the lint dependencies (PyYAML)


foolip:
From "ImportError: No module named yaml" and reading .travis.yml I was able to work out that this is needed to get the lint running: `sudo pip install --requirement tools/generation/requirements.txt`.

It would be great if this were more obvious, either in CONTRIBUTING.md, or if the lint script itself points out what to do if the yaml module is missing.

PR #1298, opened 2017-10-18
tools/lint: install  PyYAML if not available

rwaldron:
cc @foolip WDYT?

### DIFF
--- a/tools/lint/lint.py
+++ b/tools/lint/lint.py
@@ -4,6 +4,23 @@
 
 import argparse
 import sys
+import pip
+import pip.req
+
+
+try:
+    __import__('yaml')
+except ImportError:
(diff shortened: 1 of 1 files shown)


leobalter:
works for me, but I'd like to make sure it works for @foolip before landing, so we can consider this fixes #1286.

tools/lint/lint.py
```diff
@@ -4,6 +4,23 @@
...
+    __import__('yaml')
+except ImportError:
+    for item in pip.req.parse_requirements("./tools/lint/requirements.txt", session="test262"):
+        if isinstance(item, pip.req.InstallRequirement):
+            requirement = item.name
+
+            if len(str(item.req.specifier)) > 0:
+                requirement = "{}{}".format(requirement, item.req.specifier)
+
+            # print(requirement)
```
leobalter: remove?

PR #1299, opened 2017-10-18
Add a test for GetIterator(obj, ~async~) attempting to call obj[@@asyncIterator] even if that value is an object with an [[IsHTMLDDA]] internal slot

jswalden:
Just a replay of the same thing as #1287, in a different scenario.  Also another bug SpiderMonkey has until I land a patch for it.


rwaldron:
👍

PR #1301, opened 2017-10-18
More class fields test of early errors

spectranaut:
This adds tests for private class fields (realized in this [comment](https://github.com/tc39/proposal-class-fields/issues/47))) and also! I misnamed a file in the previous commit.

@leobalter after you review I'll re-generate the test files?


spectranaut:
I added the template for equivalent static private fields, too.
thanks @leobalter @littledan !

leobalter:
there are some issues with the files generation, let me check.

Issue #1304, opened 2017-10-19
uncallableAndIsHTMLDDA should be callable


bakkot:
At least, as I understand it, it's important that `document.all` be able to fill the contract of `uncallableAndIsHTMLDDA`. (If nothing else because the spec says hosts should not create any objects with the `[[IsHTMLDDA]]` slot except for `document.all`.

But [per the WHATWG spec](https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#HTMLAllCollection-call), `document.all` when called with no arguments does *not* throw a type error - it just returns null. So it _can't_ fulfill the "for which Call(obj, any value, «») throws a TypeError." part of the contract of `uncallableAndIsHTMLDDA`.

This might require rewriting parts of #1287 and #1299, unfortunately. Sorry for not catching this before that landed. (Downsides of this sort of cross-spec test...)

cc @jswalden

jswalden:
Yeah.  I'm writing up a comment describing what we need for this stuff (possibly two separate functions :-( ), sec.

jswalden:
So the value returned by `$262.uncallableAndIsHTMLDDA` is currently specified to be something that, when called with no arguments, throws a `TypeError`.  That's not what the three tests that use it, actually need.

### [yield*](https://github.com/tc39/test262/blob/23bc183bb8e7c33a35d54a92794423bdf9cfe4ce/test/annexB/language/expressions/yield/star-iterable-return-emulates-undefined-throws-when-called.js) ###

The steps after the `return` property is gotten as `return` are:

1. Let `innerReturnResult` be `? Call(return, iterator, « received.[[Value]] »)`.
1. If `Type(innerReturnResult)` is not Object, throw a `TypeError` exception.

It's easy to make `document.all` invoked with one argument return `null` and thus throw that `TypeError`: make the argument `""`.  So we can make the argument `""` to fix the test to work with `document.all` there.

### [`IteratorClose`](https://tc39.github.io/ecma262/#sec-iteratorclose) ###

The steps after the `return` property is gotten as `return` are:

1. Let `innerResult` be `Call(return, iterator, « »)`.
1. If `completion.[[Type]]` is `throw`, return `Completion(completion)`.
1. If `innerResult.[[Type]]` is `throw`, return `Completion(innerResult)`.
1. If `Type(innerResult.[[Value]])` is not Object, throw a `TypeError` exception.

[...]

PR #1305, opened 2017-10-19
Modify $262.uncallableAndIsHTMLDDA() to $262.fakeDocumentAll, and narrowly/correctly prescribe its requirements consistent with `document.all`'s behavior in HTML

jswalden:
The current definitions of this stuff aren't consistent with how HTML defines `document.all`'s behavior, so we gotta fix that.  The three tests that use this functionality have very particular requirements for how this value behaves when called, so define the requirements of this object to narrowly encompass them.

@bakkot is mostly of the opinion that this should just be a browser-only sort of test and use `document.all` specifically.  In practice, however, those of us who work on JS engines get a ton of value out of not having to build the browser to run a test.  That value is well worth the narrowly-defined specificity of the requirements for this.

SpiderMonkey's `objectEmulatingUndefined` won't be adequate to mock this any more (it's not callable), but I can fix it up/rename it to have the necessary additional behaviors.  @bakkot has at least convinced me it's not worth exposing that unless it behaves as much like `document.all` as we need it to.  :-)

See also #1304, to which I don't know how to attach this branch-merge-request without opening an entirely fresh PR for it.

### DIFF
--- a/INTERPRETING.md
+++ b/INTERPRETING.md
@@ -64,16 +64,16 @@ properties of the global scope prior to test execution.
         6. Return Completion(status).
 
   - **`global`** - a reference to the global object on which `$262` was initially defined
-  - **`uncallableAndIsHTMLDDA`** - a function that returns an object *`obj`* for
-    which [Call](https://tc39.github.io/ecma262/#sec-call)(*`obj`*, *any value*, «»)
-    throws a `TypeError`.  (The value of [IsCallable]()(*`obj`*) is unspecified:
-    a callable *`obj`* that throws a `TypeError` or an uncallable *`obj`* works
-    equally well.)  In hosts supporting the
-    [IsHTMLDDA](https://tc39.github.io/ecma262/#sec-IsHTMLDDA-internal-slot)
-    internal slot, *`obj`* must also have such a slot.  (These highly specific
-    behaviors are entirely motivated by the very few tests that use this.  Read
-    them for an explanation.)  Tests that use this function should be marked as
-    using the `uncallableAndIsHTMLDDA` feature.
+  - **`IsHTMLDDA`** - (present only in implementations that can provide it) an
+    object that 1) has an [[IsHTMLDDA]] internal slot, and 2) when called with
+    no arguments or with the single argument `""` returns `null`.  Use this
+    property to test that ECMAScript algorithms aren't mis-implemented to treat
+    `document.all` as being `undefined` or of type Undefined (instead of
+    Object).  (The peculiar second requirement permits testing algorithms when
+    they also call `document.all` with such arguments, so that testing for
+    correct behavior requires knowing how the call behaves.  This is rarely
+    necessary.)  Tests using this function must be tagged with the `IsHTMLDDA`
+    feature so that only hosts supporting this property will run them.
   - **`agent`** - an ordinary object with the following properties:
     - **`start`** - a function that takes a script source string and runs
       the script in a concurrent agent.  Will block until that agent is
--- a/features.txt
+++ b/features.txt
@@ -124,4 +124,4 @@ WeakSet
 # language features, exposed through global-environment functions on the $262
 # object, go here.
 
-uncallableAndIsHTMLDDA
+IsHTMLDDA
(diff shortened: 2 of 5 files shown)


INTERPRETING.md
```diff
@@ -64,16 +64,15 @@ properties of the global scope prior to test execution.
...
-    using the `uncallableAndIsHTMLDDA` feature.
+  - **`fakeDocumentAll`** - an object that 1) optionally has an [[IsHTMLDDA]]
+    internal slot, and 2) when called with no arguments or with the single
+    argument `""` returns `null`.  Use this property to test that ECMAScript
+    algorithms aren't mis-implemented to treat `document.all` as being
+    `undefined` or of type Undefined (instead of Object).  (The peculiar second
+    requirement permits testing algorithms when they also call `document.all`
+    with such arguments, so that testing for correct behavior requires knowing
+    how the call behaves.  This is rarely necessary.)  Tests using this function
+    should be tagged with the `fakeDocumentAll` feature.
```
leobalter: while we are at it, I think just calling it `IsHTMLDDA` - instead of `fakeDocumentAll` - would be fine and it makes an easier reference for the ES specs.

This shouldn't be a big deal, but if we should also avoid rediscussion on this name here.

leobalter:
> is mostly of the opinion that this should just be a browser-only sort of test and use document.all specifically

Just having this collection of tests as a behavior of ES on top of a known exotic object (IsHTMLDDA). The host defined API solves the problem where they are implemented defined and the host can provide a stub or just filter the tests if necessary.

The changes in this PR reflects better the behaviors covered with the information we already have in the specs. It shouldn't be the first implementation defined thing we handle in the specs. We can list the Detached buffer and the realm creation API. 

Going deeper, this case might not be the best fit both for Test262 or the WPT, but the way it is formatted - even providing a features tag - allows some filtering I am willing to maintain and support other implementers with it if necessary. 

> @bakkot has at least convinced me it's not worth exposing that unless it behaves as much like document.all as we need it to. :-)

+1. I understand the concerns from @bakkot and they are all legit. I'm still accepting this for the lack of a better home for the tests.

INTERPRETING.md
```diff
@@ -64,16 +64,15 @@ properties of the global scope prior to test execution.
...
-    which [Call](https://tc39.github.io/ecma262/#sec-call)(*`obj`*, *any value*, «»)
-    throws a `TypeError`.  (The value of [IsCallable]()(*`obj`*) is unspecified:
-    a callable *`obj`* that throws a `TypeError` or an uncallable *`obj`* works
-    equally well.)  In hosts supporting the
-    [IsHTMLDDA](https://tc39.github.io/ecma262/#sec-IsHTMLDDA-internal-slot)
-    internal slot, *`obj`* must also have such a slot.  (These highly specific
-    behaviors are entirely motivated by the very few tests that use this.  Read
-    them for an explanation.)  Tests that use this function should be marked as
-    using the `uncallableAndIsHTMLDDA` feature.
+  - **`fakeDocumentAll`** - an object that 1) optionally has an [[IsHTMLDDA]]
```
bakkot: I would like to remove "optionally", personally. It should be on the hosts to decide whether or not they can provide this API - by analogy, not all hosts can provide "DetachArrayBuffer", but the hook still just says it is "a function which implements the DetachArrayBuffer abstract operation", no optionality about it.

And hosts which *can't* provide this API don't really need to run these tests anyway. From my perspective there's no point in allowing this API to be fulfilled by `{}`; it's just going to confuse test consumers about what is expected of them.

rwaldron:
Are we all in full agreement on this name? I will update my eshost PR.

leobalter:
> @rwaldron: Are we all in full agreement on this name? I will update my eshost PR.

not yet, please. Hang on while we don't have more feedback on the name. https://github.com/tc39/test262/pull/1305#discussion_r145743390.

leobalter:
> @bakkot: I would like to remove "optionally", personally. It should be on the hosts to decide whether or not they can provide this API - by analogy, not all hosts can provide "DetachArrayBuffer", but the hook still just says the detachArrayBuffer hook is "a function which implements the DetachArrayBuffer abstract operation", no optionality about it.

Agreed.

INTERPRETING.md
leobalter: I agree with this. @jswalden if you prefer I can fix/change this in a follow up.

jswalden: Yeah, that's fair.  Means SpiderMonkey's going to have to change before I can run these tests come next import, but that's easy enough to do.  I'm kinda more partial still to `fakeDocumentAll` for a name, but eh.  Changed the name as well in the latest update to the branch.

PR #1309, opened 2017-10-21
classfields: add tests for forbidden public/private propertynames

spectranaut:


src/class-fields/propname-constructor.case
```diff
@@ -0,0 +1,24 @@
+// Copyright (C) 2017 Valerie Young. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: Static class fields cannot have PropName 'prototype'
+info: |
+  // This test file tests the following early error:
```
spectranaut: this might look weird, but its to make the generated test file's info text a little easier to read. The case file info text is concatenated after the template file text, and it looks confusing without this clarification.

At least, I think it's less confusing this way ;)

test/language/expressions/class/fields-computed-name-propname-constructor.js
```diff
@@ -0,0 +1,34 @@
...
+    ClassElementName : PropertyName
+      Return PropName of PropertyName.
+
+    ClassElementName : PrivateName
+      Return empty.
+
+---*/
+
+
+x = constructor;
```
spectranaut: Should I have this positive tests?

src/class-fields/propname-error/cls-decl-computed-name.template
```diff
@@ -0,0 +1,18 @@
...
+path: language/statements/class/fields-computed-name-
+name: computed
+info: |
+  Static Semantics: PropName
+  ...
+  ComputedPropertyName : [ AssignmentExpression ]
+    Return empty.
+---*/
+
+x = /*{ propname }*/;
```
rwaldron: var, let or const?

src/class-fields/propname-error/cls-expr-computed-name.template
```diff
@@ -0,0 +1,18 @@
...
+path: language/expressions/class/fields-computed-name-
+name: computed
+info: |
+  Static Semantics: PropName
+  ...
+  ComputedPropertyName : [ AssignmentExpression ]
+    Return empty.
+---*/
+
+x = /*{ propname }*/;
```
rwaldron: var, let or const?

src/class-fields/propname-constructor.case
```diff
@@ -0,0 +1,24 @@
+// Copyright (C) 2017 Valerie Young. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: Static class fields cannot have PropName 'prototype'
```
rwaldron: 'prototype' => 'constructor'

src/class-fields/propname-error/cls-expr-literal-name.template
```diff
@@ -0,0 +1,21 @@
...
+  phase: early
+info: |
+  Static Semantics: PropName
+  LiteralPropertyName : IdentifierName
+    Return StringValue of IdentifierName.
+---*/
+
+throw "Test262: This statement should not be evaluated.";
+
+C = class {
```
rwaldron: var, let or const?

src/class-fields/propname-error/cls-expr-string-name.template
```diff
@@ -0,0 +1,22 @@
...
+info: |
+  Static Semantics: PropName
+  ...
+  LiteralPropertyName:StringLiteral
+    Return the String value whose code units are the SV of the StringLiteral.
+---*/
+
+throw "Test262: This statement should not be evaluated.";
+
+C = class {
```
rwaldron: var, let or const?

test/language/expressions/class/fields-computed-name-propname-constructor.js
rwaldron: Not sure, but I'm wondering if `constructor` is supposed to be a string `"constructor"`?

test/language/statements/class/fields-computed-name-propname-constructor.js
```diff
@@ -0,0 +1,34 @@
...
+
+    FieldDefinition : ClassElementNameInitializer
+      Return PropName of ClassElementName.
+
+    ClassElementName : PropertyName
+      Return PropName of PropertyName.
+
+    ClassElementName : PrivateName
+      Return empty.
+
```
rwaldron: Looks like the `negative:` frontmatter is gone? I'm not sure why that would happen, but @leobalter probably knows

src/class-fields/propname-error/cls-decl-computed-name.template
```diff
@@ -0,0 +1,18 @@
...
+/*---
+esid: sec-class-definitions-static-semantics-early-errors
+path: language/statements/class/fields-computed-name-
+name: computed
+info: |
+  Static Semantics: PropName
+  ...
+  ComputedPropertyName : [ AssignmentExpression ]
+    Return empty.
+---*/
```
leobalter: this template is missing the negative flags.

@rwaldron this should answer [your comment](https://github.com/tc39/test262/pull/1309#discussion_r147448551)

test/language/expressions/class/fields-computed-name-propname-constructor.js
leobalter: this can be further tested if you add a assert.throw when you construct a new instance of `C`.

In anyway this test file is valid already for static semantics, after the suggestions from Rick are applied in this line. (`var x = "constructor";`

It's really nice to have this test case.

Btw, out of curiosity:

```js
var constructor = "lol";

class C {
  [constructor];
}
var c = new C();
assert.sameValue(c.hasOwnProperty("lol"), true);
```

src/class-fields/propname-error/cls-expr-computed-name.template
leobalter: this line should be like `var x = "/*{ propname }*/";

test/language/expressions/class/fields-computed-name-propname-constructor.js
spectranaut: @leobalter I'm having troubling finding in the spec where/how it will error if there is a class field with the PropName "constructor" -- I think what happens is this:

You can see the making of the constructor from the constructor method starting at step 13 [here](https://tc39.github.io/proposal-class-fields/#runtime-semantics-class-definition-evaluation)
```
13. Else, let constructor be ConstructorMethod of ClassBody.
```
ConstructorMethod looks for a ClassElement with PropName "constructor" -- so the computer property name field won't surface here. The computed property name is evaluated in step 27.a.i, and then added to the instance while the instance is being constructed. 

So no error? I will check that it doesn't error when constructing a new instance, though!

test/language/statements/class/fields-computed-name-propname-constructor.js
spectranaut: @rwaldron @leobalter this file should not early error -- so it was not included in the template :)

spectranaut:
hey @leobalter @rwaldron -- I'm sorry I squashed too early! For some reason I thought github would preserve this history of comments XD it sucks because I left a couple detailed ones in reply to some questions.. but I did address all comments in the recent commit.

test/language/expressions/class/fields-computed-name-static-propname-constructor.js
```diff
@@ -0,0 +1,32 @@
...
+    // This test file tests the following early error:
+    Static Semantics: Early Errors
+
+      ClassElement : staticFieldDefinition;
+        It is a Syntax Error if PropName of FieldDefinition is "prototype" or "constructor".
+
+---*/
+
+
+var x = "constructor";
```
leobalter: `var constructor = "/* { name } */"`

using `constructor` for the variable name should fun (in the template)

The template may also have a better description why computed name does not produce an early error.

test/language/expressions/class/fields-computed-name-propname-constructor.js
```diff
@@ -2,7 +2,7 @@
 // - src/class-fields/propname-constructor.case
 // - src/class-fields/propname-error/cls-expr-computed-name.template
 /*---
-description: Static class fields cannot have PropName 'constructor' (computed)
+description: class fields forbid PropName 'constructor' (no early error -- PropName of ComputedPropertyName not forbidden value)
```
spectranaut: @leobalter I think this is a clearer explanation of why there is no error!

PR #1318, opened 2017-10-24
BigInt TypedArray tests (WIP)

cxielarko:


cxielarko:
This is a rough draft of BigInt TypedArray tests I'm using for development; does anyone have advice on how they could be improved for inclusion in test262?

Existing TypedArray tests are already parameterized by the constructor, and are run using testWithTypedArrayConstructors from testTypedArray.js. Numbers are used to represent array data, so the tests don't work as written for BigInt arrays. In this branch, the test functions take a data-conversion function as an additional argument, which is used to convert sample data to an appropriate type for the constructor being tested. The new argument is the identity function in most cases, but for Big(U)Int64Array it converts Numbers to BigInts and Number arrays to BigInt arrays.

This approach makes it easy to convert the existing tests, but there are some problems with it. BigInt TypedArrays are tested only when their constructors exist, so that the tests still work without BigInt. This means that the feature flags on the individual tests will be inaccurate in some conditions (whether they include BigInt or not). I don't know how much of a problem that is in practice, but it will be at least inconvenient for some users.

rwaldron:
I've read through the changes and while there might be other ways to do this, this way is as good as any. That said, I think @littledan and @leobalter should also put eyes on this, just to make sure I didn't miss anything (simply because there are a lot of files).

harness/testTypedArray.js
```diff
@@ -20,11 +20,22 @@ var typedArrayConstructors = [
   Uint8ClampedArray
 ];
 
+var numericTypedArrayConstructors = typedArrayConstructors.slice();
```
rwaldron: Once BigInt reaches Stage 4, Test262 will need to decide if it wants to start making implementations fail (where they previously passed) by adding BigInt64Array and BigUint64Array directly (unconditionally) to the `typedArrayConstructors` array.

harness/testTypedArray.js
```diff
@@ -20,11 +20,22 @@ var typedArrayConstructors = [
...
+  typedArrayConstructors.push(BigUint64Array);
+}
+
 /**
  * The %TypedArray% intrinsic constructor function.
  */
 var TypedArray = Object.getPrototypeOf(Int8Array);
 
+function convertToBigInt(x) {
+  return (x instanceof Array) ? x.map(convertToBigInt) : BigInt(x);
```
rwaldron: `(x instanceof Array)` => `Array.isArray(x)`

littledan:
I believe @thejoshwolfe is working on a version of these tests based on test generation rather than adding the extra level of calling, as this patch does. Since the work is already in progress, maybe we should wait for those tests rather than landing this one at this point.

rwaldron:
@littledan works for me

leobalter:
I used the waiting-author label to flag we need to wait for the change suggested by @littledan. We can replace it with another label using a better name if this cause any confusion.

jakobkummerow:
I have run these tests against my WIP implementation in V8, and found that the attached modifications were required (which I believe are all in conformance with the current spec).

I was a bit sloppy insofar as I simply assumed that BigInt64Array/BigUint64Array are available. For checking in those updates, one would probably have to add proper feature detection.

With those changes, the following tests still fail, which require more significant modifications to work with BigInts:

built-ins/TypedArray/prototype/includes/samevaluezero
built-ins/TypedArray/prototype/indexOf/strict-comparison
built-ins/TypedArray/prototype/lastIndexOf/strict-comparison
built-ins/TypedArray/prototype/set/typedarray-arg-set-values-same-buffer-other-type
built-ins/TypedArrays/object-arg-returns
built-ins/TypedArrays/typedarray-arg-returns-new-instance
built-ins/TypedArrays/typedarray-arg-use-custom-proto-if-object
built-ins/TypedArrays/typedarray-arg-use-default-proto-if-custom-proto-is-not-object

I'm a bit unclear on the status of this PR. Is it going to be abandoned? Is there another WIP PR I should be following for BigInt64Array tests?

[bigint-typedarray-updates.patch.txt](https://github.com/tc39/test262/files/1691159/bigint-typedarray-updates.patch.txt)

littledan:
Thanks for these fixes, @jakobkummerow !

I apologize for the delay in these tests. @thejoshwolfe has left Igalia, so I'll pick up the edits here.

cxielarko:
Thanks for the updates, @jakobkummerow. I added your changes and rebased the branch onto master. I also restricted failing tests to non-BigInt array types for now, and added feature checks to tests that refer directly to the BigInt array constructors.

leobalter:
I'm currently reviewing this PR, please ping me for any updates. I should get a proper feedback tomorrow (merge or comments).

harness/testTypedArray.js
leobalter: Yes, I'm reviewing this branch and checking the modified files. I'm gonna do some work here to split tests so we keep tests for TypedArrays going as ok while implementations for BigInt can run separately.

harness/testTypedArray.js
```diff
@@ -20,11 +20,22 @@ var typedArrayConstructors = [
   Uint8ClampedArray
 ];
 
+var numericTypedArrayConstructors = typedArrayConstructors.slice();
+
+if (typeof BigInt !== "undefined") {
+  typedArrayConstructors.push(BigInt64Array);
+  typedArrayConstructors.push(BigUint64Array);
+}
```
leobalter: this assumes by setting the constructor BigInt will also set the BigInt typed arrays.

leobalter: There is a strategy thing here, considering phenomenon the tests might all pass while `BigInt` is not implemented or used as a word.

At the point someone start implementing BigInt, just by setting any value to `BigInt` these tests will suddenly fail and not differentiate on what was already passing.

At some point, this might be an inheritance problem from the old TypeArray tests, not friendly to extensions, and I wrote them before I had much experience using the test generation tool.

In any case, they should not be failing as a whole.

leobalter: This would also require all the changed tests to flag BigInt in the feature flags, but it's weird to say they are also compatible if BigInt doesn't exist and we might be excluding all the respective typedArray tests if someone wants to filter out the BigInt feature.

I think it's time to clean up my mistake and generate individual tests for each typed array, otherwise we will have to deal with this in further extensions, if any.

ljharb: What about using `eval('1n')` to detect BigInt support?

leobalter:
The problem is not detecting BigInt, but making the whole set of typearray
tests fail if BigInt development is in progress.

I have tools to clone the test files for BigInt and avoid extra
abstractions as N().

This way we also avoid any BigInt filtering (feature flags) to remove all
the tests for TypedArrays.

I should have a prototype for this tomorrow.

On Feb 8, 2018 8:49 PM, "Jordan Harband" <notifications@github.com> wrote:

*@ljharb* commented on this pull request.
------------------------------

In harness/testTypedArray.js
<https://github.com/tc39/test262/pull/1318#discussion_r167121658>:

> @@ -20,11 +20,22 @@ var typedArrayConstructors = [
   Uint8ClampedArray
 ];

+var numericTypedArrayConstructors = typedArrayConstructors.slice()
;
+
+if (typeof BigInt !== "undefined") {
+  typedArrayConstructors.push(BigInt64Array);
+  typedArrayConstructors.push(BigUint64Array);
+}

What about using eval('1n') to detect BigInt support?

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/tc39/test262/pull/1318#discussion_r167121658>, or mute
the thread

PR #1320, opened 2017-10-26
BigInt: Add test for BigInt.prototype[Symbol.toStringTag]

littledan:
Tests the fix for https://github.com/tc39/proposal-bigint/issues/92


rwaldron:
Excellent!

PR #1321, opened 2017-10-26
BigInt: Test semantics changes for parseInt

littledan:
For proposed changes from https://github.com/tc39/proposal-bigint/issues/86
in the PR at https://github.com/tc39/proposal-bigint/pull/97

Please do not merge this patch until the specification change in https://github.com/tc39/proposal-bigint/pull/97 reaches consensus.


littledan:
We ended up dropping BigInt.parseInt, so dropping this PR.

PR #1322, opened 2017-10-26
Missing generators flags

leobalter:
I've found more tests without the necessary generators flags.

This shouldn't affect any test functionality but tooling only.

out of curiosity:

```
git grep -Le "generators" --or -e "async-iteration" -- $(git grep -lE "\*\s?[a-zA-Z_]+\s?\(" test/)
```


rwaldron:
Excellent.

PR #1326, opened 2017-10-27
Add missing parameter and reorder actual, expected assertion parameters

leobalter:
cc @anba


leobalter:
not a problem, thanks

PR #1327, opened 2017-10-28
Update 15.2.3.6-4-124.js

tomswedlund:
VerifyNotWriteable calls isWritable which tries to assign the parameter "value" to obj[name], where 'name' is the property name.  In this case, since no value is supplied, the value assigned is the string "unlikelyValue".  So we have array["length"] = "unlikelyValue", which in Chrome & Firefox result in a RangeError and isWriteable expects a TypeError to be thrown.  So implementations that validate the validate the assigned value before the property's writability will always fail this test.

### DIFF
--- a/test/built-ins/Object/defineProperty/15.2.3.6-4-124.js
+++ b/test/built-ins/Object/defineProperty/15.2.3.6-4-124.js
@@ -18,7 +18,7 @@ Object.defineProperty(arrObj, "length", {
 });
 verifyEqualTo(arrObj, "length", 0);
 
-verifyNotWritable(arrObj, "length");
+verifyNotWritable(arrObj, "length", undefined, 0);
 
 verifyNotEnumerable(arrObj, "length");
 


anba:
Before `verifyNotWritable` is called, the `length` property is [changed to non-writable](https://github.com/tc39/test262/blob/f8456c6dacefae23188830a33e300ed53b0eb67b/test/built-ins/Object/defineProperty/15.2.3.6-4-124.js#L16-L18). So any assignment like `array.length = "unlikelyValue"` will never end up in ArraySetLength, because [OrdinarySetWithOwnDescriptor](https://tc39.github.io/ecma262/#sec-ordinarysetwithowndescriptor) step 3.a already returns `false`. And that means it's not necessary to change this test case at all.

littledan:
Oh, I see. Then we should probably not make this change, so as to not mask the error in some JS implementations.

leobalter:
Following @anba comment, the only thing done by the array length in the test is setting its descriptor to non-writable. The value remains the same and the string doesn't need to be protected.

> So we have array["length"] = "unlikelyValue", which in Chrome & Firefox result in a RangeError and isWriteable expects a TypeError to be thrown.

Also, checking on Chrome & Firefox, setting both a string and a number value to `array.length` after it's non-writable on strict mode is triggering a TypeError, not a RangeError as described. e.g.: http://jsbin.com/gelojucuma/edit?js,console

```
19.1.2.4 Object.defineProperty ( O, P, Attributes )

...
4. Perform ? DefinePropertyOrThrow(O, key, desc).
...

7.3.7 DefinePropertyOrThrow ( O, P, desc )

...
3. Let success be ? O.[[DefineOwnProperty]](P, desc).
4. If success is false, throw a TypeError exception.
...

9.4.2.1 [[DefineOwnProperty]] ( P, Desc )

...
2. If P is "length", then
  a. Return ? ArraySetLength(A, Desc).
...

9.4.2.4 ArraySetLength ( A, Desc )

1. If Desc.[[Value]] is absent, then
  a. Return OrdinaryDefineOwnProperty(A, "length", Desc).
...

9.1.6.1 OrdinaryDefineOwnProperty ( O, P, Desc )

1. Let current be ? O.[[GetOwnProperty]](P).
2. Let extensible be O.[[Extensible]].
3. Return ValidateAndApplyPropertyDescriptor(O, P, extensible, Desc, current).

9.1.6.3 ValidateAndApplyPropertyDescriptor ( O, P, extensible, Desc, current )

...
9. If O is not undefined, then
  a. For each field of Desc that is present, set the corresponding attribute of the property named P of object O to the value of the field.
10. Return true.
```

rwaldron:
Thanks @tomswedlund—if you have any further questions about the responses from @anba and @leobalter, feel free to post here; in the meantime I'm going to close this.

PR #1331, opened 2017-10-30
classfields: add tests for privatename production and ASI

spectranaut:


src/class-fields/static-private-names.case
```diff
@@ -0,0 +1,50 @@
...
+  PrivateName:
+    #IdentifierName
+template: default
+---*/
+
+//- fields
+static #x; static #y
+//- privateinspectionfunctions
+  static x() {
+    this.#x = 42;
```
leobalter: this is a very interesting test case.

Maybe I'm being too precious, but I wonder if the function name should be different than the private name.

test/language/expressions/class/fields-asi-4.js
```diff
@@ -0,0 +1,17 @@
...
+/*---
+description: ASI test in field declarations -- error when generator interpreted as multiplication
+esid: sec-automatic-semicolon-insertion
+features: [class-fields]
+---*/
+
+var C = class {
+  x = 42
+  *gen() {}
+}
```
leobalter: I believe this is an early parse error. you should use the negative flags instead.

test/language/statements/class/fields-asi-1.js
```diff
@@ -0,0 +1,17 @@
...
+/*---
+description: ASI test in field declarations -- computed name interpreted as property
+esid: sec-automatic-semicolon-insertion
+features: [class-fields]
+---*/
+
+class C {
+  x = {}
+  ['lol'] = 42
+}
```
leobalter:
```js
var obj = {};
class C {
  x = o
  ['lol'] = 42
}
```

this way you can capture `o.lol` is also set to 42.

test/language/statements/class/fields-asi-3.js
```diff
@@ -0,0 +1,17 @@
...
+/*---
+description: ASI test in field declarations -- error when computed name interpreted as index
+esid: sec-automatic-semicolon-insertion
+features: [class-fields]
+---*/
+
+class C {
+  x = "string"
+  [0]() {}
+}
```
leobalter: this is also a parsing error (early).

test/language/statements/class/fields-asi-4.js
```diff
@@ -0,0 +1,17 @@
...
+
+/*---
+description: ASI test in field declarations -- error when generator interpreted as multiplication
+esid: sec-automatic-semicolon-insertion
+features: [class-fields]
+---*/
+
+class C {
+  x = 42
+  *gen() {}
```
leobalter: early parsing error too

test/language/statements/class/fields-asi-5.js
```diff
@@ -0,0 +1,18 @@
...
+features: [class-fields]
+---*/
+
+class C {
+  x
+  in
+  y
+}
+
+assert.throws(ReferenceError, function() {
```
leobalter: this is probably not an error, the construct should succeed. 

You might need to assert the new instance has own properties x, in, and y.

As we discussed in person, this can expand to the other case evaluating the right side expression:

```js
var x = 0;
var y = [42];
var z = 1;
class C {
  a = x
  in
  y
  b = z
  in
  y
}
```

src/class-fields/static-private-names.case
spectranaut: hey @leobalter -- hmmm well if a javascript interpreter confuses property "#x" with property "x", then that javascript interpreter really has a problem.

leobalter: that's a good point! I was afraid of returning a false positive, but I believe it was just some [FUD](https://en.wikipedia.org/wiki/Fear,_uncertainty_and_doubt).

spectranaut:
Hey @leobalter I submitted some fixes -- please review and then I'll squash the fix commit!

leobalter:
I rebased and fixed the commits with the missing generated tests.

PR #1325, opened 2017-10-27
 fix typos in class field templates

leobalter:
cc @anba

Issue #1313, opened 2017-10-23
built-ins/RegExp/incomplete_hex_unicode_escape.js belongs into annexB


anba:
`/\x/` and the other regular expressions in that file are Annex-B additions, so the file needs to be moved into `annexB/built-ins/RegExp`.

Issue #1312, opened 2017-10-23
string-is-code-units-of-decimal-digits-only.js can't be parsed properly by monkeyYaml


anba:
The [empty line](https://github.com/tc39/test262/blob/db05f2f51252c29bc80564f2eb379d7eb924eae1/test/built-ins/BigInt/prototype/toString/string-is-code-units-of-decimal-digits-only.js#L8) after `|` confuses monkeyYaml and it leads to not properly detecting the `features` tag.

Issue #1306, opened 2017-10-19
BigInt/prototype/toString/length.js expects the wrong length


anba:
The expected length should be `0` instead of `1`, because the radix is optional.

PR #1334, opened 2017-11-01
Fix multiple test issues

anba:
Fixes the issues noted in #1306, #1312, #1313, #1325, and two other issues I've noted some time ago.

Note:
The `eval-err-contains-arguments` tests are still not quite correct, because they expect a `ReferenceError` is thrown when `arguments` is referenced in indirect evals, but in some environments (d8) a global `arguments` property is present, which means these tests will fail there. Overall I don't think there's much value to test indirect evals in class-field initializers, because it doesn't provide more coverage compared to other contexts where indirect evals can be executed.


anba:
Sigh, `Uint16Array` is not a supported features tag...

leobalter:
@rwaldron can you fix it? I’m busy today with other tasks but the fixes here are all fine and priority to avoid bad test code.

leobalter:
it's working. 

Thanks for the patches, @anba!

PR #1336, opened 2017-11-02
Test setting/defining a typed array element to a value, when ToNumber(value) detaches the underlying ArrayBuffer

jswalden:
The historical failure of implementations to throw when setting a typed array element backed by a detached buffer rears its ugly head when I try these tests in SpiderMonkey.  Other implementations probably suffer similarly, but I didn't test anyone else.  Probably good for someone else to verify correctness here by a separate inspection, or in an implementation that handles detached-sets correctly.

### DIFF
--- a/test/built-ins/TypedArrays/internals/DefineOwnProperty/tonumber-value-detached-buffer.js
+++ b/test/built-ins/TypedArrays/internals/DefineOwnProperty/tonumber-value-detached-buffer.js
@@ -1,0 +1,53 @@
+// Copyright (C) 2017 Mozilla Corporation. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-integer-indexed-exotic-objects-defineownproperty-p-desc
+description: >
+    Defining a typed array element to a value that, when converted to the typed
+    array element type, detaches the typed array's underlying buffer, should
+    throw a TypeError and not modify the typed array.
+info: >
+  9.4.5.3 [[DefineOwnProperty]] ( P, Desc )
+
+  ...
+  3. If Type(P) is String, then
+    a. Let numericIndex be ! CanonicalNumericIndexString(P).
+    b. If numericIndex is not undefined, then
+      ...
+      x. If Desc has a [[Value]] field, then
+        1. Let value be Desc.[[Value]].
+        2. Return ? IntegerIndexedElementSet(O, numericIndex, value).
+  ...
+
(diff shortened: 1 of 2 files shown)


test/built-ins/TypedArrays/internals/DefineOwnProperty/tonumber-value-detached-buffer.js
```diff
@@ -0,0 +1,53 @@
...
+      ...
+      x. If Desc has a [[Value]] field, then
+        1. Let value be Desc.[[Value]].
+        2. Return ? IntegerIndexedElementSet(O, numericIndex, value).
+  ...
+
+  9.4.5.9 IntegerIndexedElementSet ( O, index, value )
+
+  ...
+  15. Perform SetValueInBuffer(buffer, indexedPosition, elementType, numValue).
```
leobalter: the operation won't reach this spec part, this will be caught in a previous step of the same operation:

```
3. Let numValue be ? ToNumber(value).
4. Let buffer be O.[[ViewedArrayBuffer]].
5. If IsDetachedBuffer(buffer) is true, throw a TypeError exception.
```

leobalter: in any case, the assertions are still valid. But you can take advantage of verifying the observable point even if the return of valueOf is not an integer value:

If you detach the buffer, you can't reach the following step:

``` 
6. If IsInteger(index) is false, return false.
```

test/built-ins/TypedArrays/internals/Set/tonumber-value-detached-buffer.js
```diff
@@ -0,0 +1,44 @@
...
+  2. If Type(P) is String, then
+    a. Let numericIndex be ! CanonicalNumericIndexString(P).
+    b. If numericIndex is not undefined, then
+      i. Return ? IntegerIndexedElementSet(O, numericIndex, V).
+  ...
+
+  9.4.5.9 IntegerIndexedElementSet ( O, index, value )
+
+  ...
+  15. Perform SetValueInBuffer(buffer, indexedPosition, elementType, numValue).
```
leobalter: the same thing happens here, the observable point happens earlier for this operation.

leobalter:
The tests are correct, the info description is not 100% accurate, but doesn't make the tests less correct, but they give an idea to follow up with more tests, observing the detached buffer check before returning false for non-integer values, otherwise they would prevent the detached buffer check.

Issue #1338, opened 2017-11-06
Various test case bugs


anba:
The following test cases are incorrect on c4e3d12597fae7e56fa475122aefb7bee3a16f84:

```
# Accessing `ta[0]` throws a TypeError.
test/built-ins/TypedArrays/internals/DefineOwnProperty/tonumber-value-detached-buffer.js
test/built-ins/TypedArrays/internals/Set/tonumber-value-detached-buffer.js

# array indices start at 0
test/language/expressions/class/fields-asi-5.js
test/language/statements/class/fields-asi-5.js

# static property tested on instance instead of class
test/language/expressions/class/fields-computed-name-static-propname-constructor.js
test/language/expressions/class/fields-computed-name-static-propname-prototype.js
test/language/expressions/class/fields-computed-variable-name-static-propname-constructor.js
test/language/expressions/class/fields-computed-variable-name-static-propname-prototype.js
test/language/statements/class/fields-computed-name-static-propname-constructor.js
test/language/statements/class/fields-computed-name-static-propname-prototype.js
test/language/statements/class/fields-computed-variable-name-static-propname-constructor.js
test/language/statements/class/fields-computed-variable-name-static-propname-prototype.js

# `assert.equal` is not defined
test/language/module-code/privatename-valid-no-earlyerr.js
test/language/statements/class/privatename-valid-no-earlyerr.js

# `verifyProperty` used, but harness file not included
test/language/expressions/class/fields-after-same-line-gen-private-names.js
test/language/expressions/class/fields-after-same-line-gen-static-private-names.js
test/language/expressions/class/fields-after-same-line-method-private-names.js
test/language/expressions/class/fields-after-same-line-method-static-private-names.js
test/language/expressions/class/fields-after-same-line-static-async-gen-private-names.js
test/language/expressions/class/fields-after-same-line-static-async-gen-static-private-names.js
test/language/expressions/class/fields-after-same-line-static-async-method-private-names.js
test/language/expressions/class/fields-after-same-line-static-async-method-static-private-names.js
test/language/expressions/class/fields-after-same-line-static-gen-private-names.js
test/language/expressions/class/fields-after-same-line-static-gen-static-private-names.js
test/language/expressions/class/fields-after-same-line-static-method-private-names.js
test/language/expressions/class/fields-after-same-line-static-method-static-private-names.js
test/language/expressions/class/fields-multiple-definitions-private-names.js
test/language/expressions/class/fields-multiple-definitions-static-private-names.js
[...]

PR #1339, opened 2017-11-07
Fix bugs reported in #1338

leobalter:
cc @anba


anba:
Looks good to me, down to only five failures:

- test/language/statements/class/fields-computed-name-static-propname-prototype.js
- test/language/expressions/class/fields-computed-name-static-propname-prototype.js

These two tests should expect a TypeError, because the tests attempt to redefine the non-configurable, non-writable `prototoype` property on a class. So basically it's similar to `class C { static ["prototype"](){} }`.

- test/language/statements/class/privatefieldget-success-5.js

The expected value should be the number `42`, not the string `'42'` in 88a2e077297d96d77ca188236ed0e001b68b4734.

- test/language/statements/class/privatefieldget-typeerror-3.js
- test/language/statements/class/privatefieldset-typeerror-3.js

&rarr; https://github.com/tc39/test262/issues/1338#issuecomment-342639706

leobalter:
> These two tests should expect a TypeError, because the tests attempt to redefine the non-configurable, non-writable prototoype property on a class. So basically it's similar to class C { static ["prototype"](){} }.

oops.

I need to take this off these cases off the generation tool. Not an issue. Thanks for catching it.

test/language/statements/class/privatefieldset-typeerror-3.js
```diff
@@ -40,16 +41,22 @@ class Outer {
...
 
-assert.throws(TypeError, function() {
-  // when f() is called, the private field will not be found in Inner's `this`
-  i.f();
-})
+assert.sameValue(outer.value(), 42);
+
+i.f();
+
+assert.sameValue(outer.value(), 1, "value is set from inner class instance");
```
leobalter: this feels so weird. I wonder if it is wrong or if it should be wrong.

anba:
d24e325e1a9657a9591cb8aa80276f0929307236 still needs some tweaking:
- The TypeError is already thrown during the ClassDeclaration/Expression evaluation, not when an instance of the class is instantiated.

7f94fffbafe54e9ac3f127a7b1b7fab227a1d286
- There's still something wrong here. The assertions should probably be:
```js
var outer = new Outer();
var Inner = outer.innerclass();
var i = new Inner();

assert.sameValue(outer.value(), 42);
assert.sameValue(i.value(), 42);

i.f();

assert.sameValue(outer.value(), 42, "value is set from inner class instance");
assert.sameValue(i.value(), 1, "value is set from inner class instance");
```

leobalter:
> The TypeError is already thrown during the ClassDeclaration/Expression evaluation, not when an instance of the class is instantiated.

https://tc39.github.io/ecma262/#sec-object-initializer-runtime-semantics-evaluation

Nice! This is the evaluation I was looking for.

---

```js
assert.sameValue(outer.value(), 42, "value is set from inner class instance");
assert.sameValue(i.value(), 1, "value is set from inner class instance");
```

I applied this fix, but I still think the inner class should only refer to `#x` through `super.#x`, not `this.#x`.

leobalter:
Thanks for the review, @anba! I hope we can get rid of these bugs for good. :)

leobalter:
Assuming these tests are certainly more correct than the ones with bugs, I'll land this and make myself available to fix anything else further.

PR #1340, opened 2017-11-09
Add test cases for tc39/ecma402#194

anba:
tc39/ecma402#194 is not yet merged and may need to be presented at tc39 (https://github.com/tc39/ecma402/pull/194#issuecomment-340605709).

### DIFF
--- a/harness/dateConstants.js
+++ b/harness/dateConstants.js
@@ -13,3 +13,6 @@ var date_1999_end = 946684799999;
 var date_2000_start = 946684800000;
 var date_2099_end = 4102444799999;
 var date_2100_start = 4102444800000;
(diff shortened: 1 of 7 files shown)


anba:
> I don't see any tests for what happens when you pass a string into this function. Given that we discussed and rejected calling the Date constructor here, maybe it'd be good to have a test that distinguishes the behavior here from that other behavior.

Added format[ToParts]/date-constructor-not-called.js to handle this case.

Also added format[ToParts]/time-clip-to-integer.js to test ToInteger is correctly applied for time values.

rwaldron:
@anba tc39/ecma402#194  is now merged

Issue #1308, opened 2017-10-20
Consider inlining and abridging the typeCoercion.js paradigm


thejoshwolfe:
Currently, #1282 is going through some review feedback where the use of the `harness/typeCoercion.js` is causing confusion. As the creator of `typeCoercion.js`, I have a proposal to make its use less confusing:

1. Remove `harness/typeCoercion.js`.
2. Add a python script to `tools/` that generates individual assertion cases to fill in the coverage that we were getting through `typeCoercion.js`.
3. Generate updated versions of the existing tests that use `typeCoercion.js`.

For background, `typeCoercion.js` has caused a non-trivial amount of trouble for test262 ( https://github.com/tc39/test262/pull/1241#discussion_r141140543 , https://github.com/tc39/test262/issues/1252 ).

I'm not sure the current test generation framework would be sufficient for this. Perhaps it would; I can look into it.

Currently, we're getting an awful lot of coverage from `typeCoercion.js` (see https://github.com/tc39/test262/pull/1282#discussion_r145216004 ), which is probably too much to list in individual assertions. We'll want to make sure we're generating useful assertions, and we don't want to waste effort testing too many things with a near 0% chance of helping anyone.

This issue thread is to discuss this possibility. Is this desirable?

rwaldron:
> 1. Remove harness/typeCoercion.js.
> 2. Add a python script to tools/ that generates individual assertion cases to fill in the coverage that we were getting through typeCoercion.js.
> 3. Generate updated versions of the existing tests that use typeCoercion.js.


I give "yes" to all three

thejoshwolfe:
Here's what I'm generating so far: https://github.com/thejoshwolfe/test262/commit/1adfab8e95e0b847423f5d8fd418e21598f28aa6

I emitted code that does everything the current framework is doing behind the scenes. This definitely needs to be cut down before this is ready to go. For example, the ToBigInt tests have over 600 lines of code, many of which are trivially duplicated.

I'd also like to separate tests into multiple files categorized by things such as feature usage.

And I'd like to generate meaningful messages in the assert calls.

thejoshwolfe:
Another status update: I think the output is good. The generator python code could use some polish, but I'd like feedback on the JavaScript code I'm outputting:

https://github.com/thejoshwolfe/test262/commit/994e716fb0431babf6e11b4ef39c101117248fa6

Things to note:

* Tests with the `BigInt` feature are separated into their own file (unless the API itself requires `BigInt`).
[...]

PR #1341, opened 2017-11-09
Generate type coercion tests with Python.

thejoshwolfe:
See #1308.

The generated files are not included in this PR. It seemed cleaner to make that a separate PR. Here's a preview: https://github.com/thejoshwolfe/test262/commit/fc8f65c7a095c1701b6c5477e2c8d6d5e4e54d15

The generated tests have somewhat long lines. It's proven to be difficult to get line wrapping to work and make sense, so there's no line wrapping in this PR.


littledan:
Is anyone interested in reviewing this patch? This was written in order to redo some BigInt tests to be more similar to the style that's been encouraged by test262 reviewers. @leobalter @rwaldron

CONTRIBUTING.md
```diff
@@ -291,19 +291,30 @@ In some cases, it may be necessary for a test to intentionally violate the rules
 
 ## Procedurally-generated tests
 
+There are two systems for procedurally generating tests. To generate/regenerate all generated tests, run these two commands:
+
+```
+./make.py
+./tools/generation/generate_type_coercion.py
+```
```
leobalter: in a follow up, this could be done as anchor links to each generation topic, so it's easier to understand why each one is running.

leobalter:
@littledan sure I am, the time available is just short.

One thing I'm concerned here is the work necessary to maintain the tests, not the generation tool.

The harness abstractions were hard to maintain because they are also hard to read. Generating split tests will probably generate files easier to read, but in case something needs a change, it seems that all the assertion parts are hardcoded in the generation script.

The current test generation tool is already considered a barrier by some people, and they are simple cases of templates and cases, with a similar frontmatter and JS comments for placement. 

I'd like to see an example in action for this generation tool so I could read it's usage, but I'm afraid the assertions being hardcoded will only transfer the complexity from the harness from a place to another, not minimize them.

This work is interesting, but we need to make sure this tool will be reused for more than a simple work task.

littledan:
@leobalter Would you prefer if the generation tool's results were checked in, rather than the generation tool itself? Checking in the tool is an attempt to follow a suggestion from @rwaldron .

thejoshwolfe:
Here's what the outputs look like from the sources that are in this PR: https://github.com/thejoshwolfe/test262/commit/fc8f65c7a095c1701b6c5477e2c8d6d5e4e54d15

> Generating split tests will probably generate files easier to read, but in case something needs a change, it seems that all the assertion parts are hardcoded in the generation script.

> I'm afraid the assertions being hardcoded will only transfer the complexity from the harness from a place to another, not minimize them.

Can you elaborate more on what you mean by this?

Just to bring up one example where I know this system breaks down: For `Number.prototype.toString`, the `radix` parameter has special handling when it's `undefined`, so the usual `undefined => NaN => 0` transformation doesn't happen. So either this infrastructure is inappropriate to use for that parameter (not ideal) or we need to add more complexity, e.g. an exception list that says skip the test for `undefined` coercion. And this special case happens before calling ToPrimitive, so `{valueOf: function() { return undefined; }}` should still result in `0`.

My idea is that we'd just keep tweaking the infrastructure with each little special thing like that that we'd need, but I can see how that kind of maintenance would be daunting for someone who didn't author the whole thing.

rwaldron:
@littledan 

> Is anyone interested in reviewing this patch? This was written in order to redo some BigInt tests to be more similar to the style that's been encouraged by test262 maintainers

In addition to any in-progress reviews, I've asked @jugglinmike to do a review.


> Checking in the tool is an attempt to follow a suggestion from @rwaldron .

Correct, my hope was that whatever test generation tools are created will be available for all test authors/maintainers/contributors (and so on), but that doesn't appear to be @leobalter's concern. My understanding of his feedback is that [this stuff](https://github.com/tc39/test262/pull/1341/files#diff-9b6585fbc41e2bf796abb4bbabfd3c6dR25) will be harder [than this stuff](https://github.com/tc39/test262/blob/master/src/function-forms/syntax/arrow-function.template) to maintain (a point that I agree with).

leobalter:
> @leobalter Would you prefer if the generation tool's results were checked in, rather than the generation tool itself? Checking in the tool is an attempt to follow a suggestion from @rwaldron .

Yes, I like the generation script being available in Test262, and the tool results as well.

In fact, the results (the generated tests) are the ones adding the value for Test262 consumption, while the script itself is only helpful for maintenance.


---

@thejoshwolfe:

Thanks for the generated files, you can push these as a separate commit in this PR if you want. 

> My idea is that we'd just keep tweaking the infrastructure with each little special thing like that that we'd need, but I can see how that kind of maintenance would be daunting for someone who didn't author the whole thing.

The problem today is that any tweaking requires someone to read and understand the script tool and that requires a lot of extra time for any maintenance. This generation tool seems very useful for any case of type coercion, as you also pointed some cases as `String#indexOf`. If the script could work on top of any sort of templates, this could be very helpful for someone to jump in and add more cases or manage the existing one.

---

In any case this generation script needs to be added the CI tooling as well.

littledan:
@leobalter Are you imagining that the script would work on top of the existing notion of templates, or a new one? I imagine with current templates, you'd have to actually generate the template files (sort of defeating the purpose). I wouldn't mind a new template format for this purpose, though.

littledan:
Should this test generation framework land before or after making the new template file format? I'm worried that if the project to check in the test generation tool that was used for existing tests keeps expanding in scope for more and more code cleanliness, it could take a while to finish. Ultimately, some maintainers will have to understand the Python generation tool too.

leobalter:
> Are you imagining that the script would work on top of the existing notion of templates, or a new one? I imagine with current templates, you'd have to actually generate the template files (sort of defeating the purpose). I wouldn't mind a new template format for this purpose, though.

Certainly not the current templates we have, otherwise we wouldn't need a new generation script tool.

It would be still helpful to have anything that just holds any sort of templates to abstract test contents away from the tooling.

> I wouldn't mind a new template format for this purpose, though.

Me neither.

> Should this test generation framework land before or after making the new template file format? I'm worried that if the project to check in the test generation tool that was used for existing tests keeps expanding in scope for more and more code cleanliness, it could take a while to finish. Ultimately, some maintainers will have to understand the Python generation tool too.

> I'm worried that if the project to check in the test generation tool that was used for existing tests keeps expanding in scope for more and more code cleanliness

This feels like FUD. Landing a test framework might leave us with a hard to maintain tool, consuming extra time. If this work is done for a specific case, what's the warranty the project will have a follow up feature to make it reusable?

I don't mind landing earlier the generated files from this tool - the ones in the examples seem just fine. The tool itself should be targeted for reusability before landing, even with extra updates to the generated files if necessary.

thejoshwolfe:
> Thanks for the generated files, you can push these as a separate commit in this PR if you want.

Done, and it looks like there's something about the existing generator I missed. I'll look into why the CI build is failing. My suspicion is that `./make.py clean` deletes all `flags: [generated]` tests.

rwaldron:
Before we commit to adding new tools (and bearing the burden of all that's implied by that), I'd like us to make the effort to prove or disprove whether or not the existing test generation tool is capable of creating these tests.

thejoshwolfe:
> ...prove or disprove whether or not the existing test generation tool is capable of creating these tests.

Here is my attempt.

The existing system provides an n-to-m combinatoric solution by using cases and templates. So for 3 cases and 3 templates, 9 combinations would be produced. I can see two basic approaches to doing type coercion testing this way:

* approach 1: the cases correspond to the APIs, and the templates correspond to the values.
* approach 2: the cases correspond to the values, and the templates correspond to the APIs.

Approach 1:

https://github.com/thejoshwolfe/test262/commit/1e4fc8fc0b7e4329d7bd13d789a886d044bf9e25

The problems with this strategy are:

* You have to define your asserts split between multiple macro values like [this](https://github.com/thejoshwolfe/test262/blob/1e4fc8fc0b7e4329d7bd13d789a886d044bf9e25/src/type-coercion/String.prototype.indexOf.case#L15-L20). This is ugly and error prone in many ways. Just one of the many ways this is error prone is that a misspelled macro name silently expands to the empty string (like with Makefiles and environment variables).
* The path to the generated test is primarily determined by the template not the case, which means the tests go in [a weird place](https://github.com/thejoshwolfe/test262/tree/1e4fc8fc0b7e4329d7bd13d789a886d044bf9e25/test/language/type-coercion/ToIndex).

Attempt 2:

https://github.com/thejoshwolfe/test262/commit/2a53ecd3dfe47eea208824c22325c06b01b6dd1c

This strategy doesn't work at all, because there's no mechanism to do a for-each in the macro expansion system. What I've done in the above commit is just list out the values like `throws_TypeError_0`, `throws_TypeError_1`, etc. This doesn't work because the template doesn't know how many to count up to, and in fact it's a different number for each case, so it's impossible to do it correctly.

It would work to make a specialized template for each case, but this largely defeats the purpose of a test generation framework. And even with this strategy, there's no mechanism in place to ensure that you got the number of macros correct, because omitting a macro is not an error and expanding a non-existent macro silently results in the empty string.

Conclusion:

Proving the impossibility of something is hard, but I do believe the current system is a poor fit for this problem. At best it's quite clumsy and extremely error prone. Perhaps there is a more clever way to use the existing system that isn't documented (yet?), but even after reading the source code, I don't think I missed anything that would make a significant difference.

littledan:
@thejoshwolfe Could the issues from the first approach be addressed by changes to the existing templating system?

thejoshwolfe:
> Could the issues from the first approach be addressed by changes to the existing templating system?

Here's what I would propose for that approach:

Allow cases OR templates to specify the path prefix, but not both. This solves the filename issue.

Enable macro functions. Then you can do this:

```js
// in the .case file:
//- zero(value, message)
assert.sameValue("a".indexOf("a", /*{value}*/), 0, /*{message}*/);
```

```js
// in the .template file:
/*{zero(*/0/*,*/undefined/*)}*/
/*{zero(*/NaN/*,*/"NaN => 0"/*)}*/
/*{zero(*/{valueOf: function() { return "foo"; }}/*,*/"valueOf => unparsable string => NaN => 0"/*)}*/
```

```js
// in the produced .js file:
assert.sameValue("a".indexOf("a", 0), 0, undefined);
assert.sameValue("a".indexOf("a", NaN), 0, "NaN => 0");
assert.sameValue("a".indexOf("a", {valueOf: function() { return "foo"; }}), 0, "valueOf => unparsable string => NaN => 0");
```

This solves the problem of splitting the test cases into syntactically unstructured source fragments, like `throws_0`, `throws_1`, `throws_2`.

Relative to this PR, there's the slight ugliness of passing `undefined` as a message to the assert functions sometimes. But this is a minor issue, and perhaps such trivial cases shouldn't even be tested anyway.

This still does not solve the line length problem.

The problem where misspelling a macro name is silently ignored is a separate issue, and can be its own discussion thread.

thejoshwolfe:
> The problem where misspelling a macro name is silently ignored is a separate issue, and can be its own discussion thread.

See #1360.

leobalter:
The tests generated are already merged. We should also not follow with this new set of tooling for now. Thanks

PR #1344, opened 2017-11-13
Class fields with the same name

leobalter:
cc @littledan, the test generation might be overkill for these but at least expand the same case for reuse on class declaration and expressions.

This also hints we should try some abrupt completion middle way and detecting behaviors as initializers expression changing values during the initialization.


leobalter:
I want to give an extra review on this - even it is my own work. I need to confirm the definition of computed prop name in the fields.

leobalter:
Ok, I should continue adding a few more tests for this. Including the execution order for the constructor method and fields initializers in an instance initialization and values changed for computed property names and new references in the initializers.

leobalter:
This should be good for a review.

cc @anba can you help me with some 👀  here?

leobalter:
Due to so many fixes, the split commits ended up as not much helpful. I rebased and squashed them.

littledan:
Definitely agree with your comment: Throwing an exception in the middle of initialization and observing half-initialized instances is important to test.

I took a look at a couple of the generated tests and they seem correct to me. Unfortunately I wouldn't be able to do a full review any time soon.

leobalter:
> I took a look at a couple of the generated tests and they seem correct to me. Unfortunately I wouldn't be able to do a full review any time soon.

I'll let this sit for a few more days, hoping I can get some extra eyes on it, before merging this. In any case, I want this to land before I move on the #1346 work to avoid a long extra rebasing.

leobalter:
Ok, I'll be prompt to fix anything else, but for now I'd like to open the path to finish the work on #1346.

PR #1348, opened 2017-11-16
Let paramVarEnv be NewDeclarativeEnvironment(originalEnv).

rwaldron:
Originally discovered by Cait Potter (@caitp), reported in irc.mozilla.org#jslang

@leobalter I need another critical look at this, I searched through _a lot_ of existing tests (thanks to detailed `info` frontmatter, I was able to search for tests with relevant spec prose!), but I'm afraid that I may have missed something—hopefully you can help me recheck. In the meantime, I will do another pass. It seems unlikely and would be very surprising that such a test doesn't already exist.

### DIFF
--- a/test/language/function-code/each-param-has-own-non-shared-eval-scope.js
+++ b/test/language/function-code/each-param-has-own-non-shared-eval-scope.js
@@ -3,7 +3,9 @@
 
 /*---
 esid: sec-function-definitions-runtime-semantics-iteratorbindinginitialization
-description: A new declarative environment is created, from the originalEnv, for each parameter.
+description: >
+  A new declarative environment is created, from the originalEnv, for each
+  parameter, and paramVarEnv is not shared.
 info: |
   Runtime Semantics: IteratorBindingInitialization
 
@@ -17,7 +19,7 @@ features: [arrow-function, default-parameters]
 ---*/
 
 var y;
-function f(a = eval("var x = 1; y = 42; x"), b = x) {}
+function f(a = eval("var x = 1; y = 42; x"), b = eval("x")) {}
 
 assert.throws(ReferenceError, () => {
   f();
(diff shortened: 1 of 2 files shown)


test/language/function-code/each-param-has-own-scope.js
```diff
@@ -0,0 +1,21 @@
...
+  Runtime Semantics: IteratorBindingInitialization
+
+  FormalParameter : BindingElement
+
+  ...
+  6. Let paramVarEnv be NewDeclarativeEnvironment(originalEnv).
+  ...
+---*/
+
+function f(a = eval("var x = 1; x"), b = x) {}
```
leobalter: maybe you should observe the eval execution by leaking an assignment to a name defined in the outer scope.

```js
var y;
function f(a = eval("var x = 1; y = 42; x"), b = x) {}

assert.throws(ReferenceError, () => {
  f();
});

assert.sameValue(y, 42); 
```

This prevents any false positive.

I also imagine the second parameter with an eval call to assure the `paramVarEnv` is not shared, but I'm not sure how to draft this. Maybe it's too much?

rwaldron: > This prevents any false positive.

+ 1 to that change

> I also imagine the second parameter with an eval call to assure the paramVarEnv is not shared, but I'm not sure how to draft this. Maybe it's too much?

Or maybe just another test altogether?

test/language/function-code/each-param-has-own-non-shared-eval-scope.js
```diff
@@ -0,0 +1,28 @@
...
+
+  ...
+  6. Let paramVarEnv be NewDeclarativeEnvironment(originalEnv).
+  ...
+
+features: [arrow-function, default-parameters]
+---*/
+
+var y;
+function f(a = eval("var x = 1; y = 42; x"), b = eval("x")) {}
```
leobalter: +1

PR #1350, opened 2017-11-18
Test change in PluralRules default options semantics

littledan:
The change is done in the following patch, where default options
have a null prototype. This matches what other Intl objects have.
https://github.com/tc39/proposal-intl-plural-rules/commit/1abe8af440fb47d888ace701e8a46559bdbe7b06

Note that this change has already been committed to the proposal repository, and it matches existing implementations, so the patch should be land-able from a spec perspective.

I looked at adding a feature flag, but it seems like there's no feature flag for PluralRules yet.

### DIFF
--- a/test/intl402/PluralRules/default-options-object-prototype.js
+++ b/test/intl402/PluralRules/default-options-object-prototype.js
@@ -1,0 +1,18 @@
+// Copyright (C) 2017 Igalia, S. L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-initializepluralrules
+description: >
+  Monkey-patching Object.prototype does not change the default
+  options for PluralRules as a null prototype is used.
+info: >
+  InitializePluralRules ( collator, locales, options )
+
+    1. If _options_ is *undefined*, then
+      1. Let _options_ be ObjectCreate(*null*).
+---*/
+
+Object.prototype.type = "ordinal";
+let pluralRules = new Intl.PluralRules("en");
+assert.sameValue(pluralRules.resolvedOptions().type, "cardinal");


test/intl402/PluralRules/default-options-object-prototype.js
```diff
@@ -0,0 +1,18 @@
...
+info: >
+  InitializePluralRules ( collator, locales, options )
+
+    1. If _options_ is *undefined*, then
+      1. Let _options_ be ObjectCreate(*null*).
+---*/
+
+Object.prototype.type = "ordinal";
+let pluralRules = new Intl.PluralRules("en");
+assert.sameValue(pluralRules.resolvedOptions().type, "cardinal");
```
leobalter: The resolveOptions still return an object which [prototype is %ObjectPrototype%](https://github.com/tc39/proposal-intl-plural-rules/commit/1abe8af440fb47d888ace701e8a46559bdbe7b06#diff-adc4f706d081d14e0e322335e03c7727R280).

I think it's valuable to check this is still effective, so one can benefit of this inheritance for the returned value (using toString, etc).

One other good thing for this test is: GetOption should not trigger any observable behaving on the Object.prototype for this new created `options`. You can verify it this using a `Object.defineProperty(Object.prototype, "type", { get() { throw new Test262Error() } } );`. This way you also assert the proper [`GetOption(options, type, ...`](https://github.com/tc39/proposal-intl-plural-rules/commit/1abe8af440fb47d888ace701e8a46559bdbe7b06#diff-adc4f706d081d14e0e322335e03c7727R20) operation.

littledan:
Those are good ideas for tests on all of the Intl options tests; I don't think any of them have tests like that right now. Let's pursue that test in a separate patch.

PR #1352, opened 2017-11-20
Improve Array#sort tests for verifying non callable comparefn

### LINKED ISSUE #1345: Array#sort with non-function argument: add tests for arrays with less than 2 elements
@ljharb In https://github.com/tc39/test262/blob/master/test/built-ins/Array/prototype/sort/comparefn-nonfunction-call-throws.js, the following case is tested:

```js
[1, 2, 3].sort(foo); // must throw if foo is a non-undefined non-function
```

However, the following should be tested, as [it fails on more engines](https://github.com/tc39/ecma262/pull/785#issuecomment-278101975):

```js
[1].sort(foo); // must throw if foo is a non-undefined non-function
```

leobalter:
Fix #1345 

cc @ljharb @claudepache

### DIFF
--- a/test/built-ins/Array/prototype/sort/comparefn-nonfunction-call-throws.js
+++ b/test/built-ins/Array/prototype/sort/comparefn-nonfunction-call-throws.js
@@ -15,40 +15,81 @@ info: >
 features: [Symbol]
 ---*/
 
-var sample = [1, 2, 3];
+var sample = [1];
+var poisoned = {
+  get length() {
+    throw new Test262Error("IsCallable(comparefn) should be observed before this.length");
+  }
+};
 
 assert.throws(TypeError, function() {
  sample.sort(null);
-});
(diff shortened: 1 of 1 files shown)


test/built-ins/Array/prototype/sort/comparefn-nonfunction-call-throws.js
```diff
@@ -15,40 +15,81 @@ info: >
...
+};
 
 assert.throws(TypeError, function() {
  sample.sort(null);
-});
+}, "sample.sort(null);");
+
+assert.throws(TypeError, function() {
+  [].sort.call(poisoned, null);
+}, "[].sort.call(poisoned, null);");
```
leobalter: I kept both tests using a regular array with one entry `[1]` and another one with an object containing a poisoned `length` property.

PR #1357, opened 2017-11-26
Fix several Decode tests

jridgewell:
The tests are explicitly for bad hex chars at certain indexes, and that
rogue `%` was giving false positives.

I'm not super in favor of the string concats, but it's not affecting anything...


leobalter:
this is another case of a poorly written test I'd love to improve (with a proper time for that).

Thanks for finding the rogue `%`.

PR #1359, opened 2017-11-28
Split class fields tags

### LINKED ISSUE #1346: Split class-fields features tag for public or private fields
We are tracking a single proposal, but I got a report from implementations this feature tag can help the development process splitting out between public and private fields.

leobalter:
Fixes #1346 

cc @rwaldron @gsathya

### DIFF
--- a/src/class-fields/call-expression-privatename.case
+++ b/src/class-fields/call-expression-privatename.case
@@ -4,6 +4,7 @@
 /*---
 desc: Syntax error if you call delete on call expressions . privatename
 template: delete-error
+features: [class, class-fields-private, class-fields-public]
 ---*/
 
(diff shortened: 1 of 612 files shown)


littledan:
Thanks for adding this useful metadata!

src/class-fields/call-expression-privatename.case
```diff
@@ -4,6 +4,7 @@
 /*---
 desc: Syntax error if you call delete on call expressions . privatename
 template: delete-error
+features: [class, class-fields-private, class-fields-public]
```
rwaldron: 🎉

rwaldron:
Dude... this is so awesome. Thanks!

Issue #1254, opened 2017-10-02
Extending metadata for "negative" tests


jugglinmike:
Many tests in Test262 describe programs that are expected to fail. For the past year or so, we've been maintaining metadata to differentiate between two kinds of failures: the "phase" attribute. It's set to "early" for tests that should fail before the code is executed, and it's set to "runtime" for tests that should fail only after execution has begun. This allows JavaScript parsers to appropriately interpret test results: since parsers don't execute the code, they should expect that tests expected to fail at "runtime" do *not* fail to parse.

This distinction is a little muddy when it comes to module code. If one JavaScript module depends on another module and the second module has an "early" error, then running the first should produce that same early error. The problem is that it doesn't necessarily make sense for a JavaScript parser to load external modules. This means that a negative test marked as "early" might not actually make parsers fail. Currently, the folks running the code have to inspect these mismatches and make them as "allowed" on a case-by-case basis.

To make this experience better we can differentiate between these kinds of failure with a third value for the test metadata. That way, parsers could correctly interpret tests that are expected to fail when modules are being loaded (i.e. they should parse without an issue).

A few of us discussed this change at last week's TC39 meeting in Boston, and I'm excited to write the patch. Unfortunately, we stopped short of defining the specific "phase" metadata values we should use. I'd like to start the ball rolling:

- "parse" for all early errors that originate in the current module or the
  current script
- "resolution" for all early errors that originate in external module code
- "runtime" for all errors that occur during evaluation

I'm suggesting that we no longer use the value "early" because that word can describe both of the "phases" that we are trying to differentiate.

adrianheine:
I think this can be closed now that #1366 is merged.

rwaldron:
Like @adrianheine said, this is done.

PR #1366, opened 2017-12-03
Restructure "negative" metadata

jugglinmike:
A handful of folks discussed this change at the September 2017 TC39 meeting,
and we've had gh-1254 to facilitate discussion since then. Now that I've
actually implemented the change, though, I'm starting to wonder if the change
is too disruptive to warrant the benefit. We're trying to improve our support
of consumers who are maintaining ECMAScript parsers, but this distinction is
only currently apparent in eight tests.

We could limit churn by keeping the "early" phase and simply introducing the
"resolution" phase. I'm not so hot on the idea of having "early" failures
alongside "resolution" failures since the latter is a subset of the former.
Such an imprecise taxonomy might make it hard for future contributors to
recognize when their new tests really should be "resolution" and not "early."
And while a lot of files are modified here, I can't imagine that responding to
this particular change would be onerous for any consumer.

I'm submitting this as three distinct commits so it's easier to see what it
might look like to keep the "early" phase as-is (basically by throwing out the
first commit).


rwaldron:
"early" is easier to map to a spec concept. I need to think about this more

rwaldron:
Resolved and Merged

jmdyck:
@rwaldron:
> "early" is easier to map to a spec concept. I need to think about this more

So it sounds like you were initially somewhat reluctant to eliminate "phase: early". For posterity, could you give your rationale for accepting it?

PR #1369, opened 2017-12-06
Remove BigInt.parseInt tests

littledan:
BigInt.parseInt will be removed in
https://github.com/tc39/proposal-bigint/pull/101

This patch removes the associated test262 tests.


littledan:
The associated spec patch has landed, so this test change is ready to land once reviewed.

littledan:
Just realized, I should probably add a test to assert that BigInt.parseInt isn't there at all.

littledan:
Added a test to verify that there is no parseInt method, so this is ready for review.

Issue #1356, opened 2017-11-23
Parser testing is incredibly inconvenient using Test262


isiahmeadows:
I was going through the tests to adapt them for parsers (since [this](https://github.com/tc39/test262-parser-tests) is hardly maintained, much less kept up to date), and I found a couple major issues:

**Some tests aren't actually testing what they're supposed to**

Here's a few examples:

- [This test](https://github.com/tc39/test262/blob/master/test/language/statements/function/S13_A14_T2.js) doesn't actually test functions with Unicode escapes in their name, because the string itself resolves it to literally `A`. So what's `eval`ed is `function __funcA(__arg){return __arg;}; __funcA`, not the expected `function __func\u0041(__arg){return __arg;}; __funcA`. It also has the below issue.
- These six tests ([1](https://github.com/tc39/test262/blob/master/test/language/statements/block/12.1-1.js), [2](https://github.com/tc39/test262/blob/master/test/language/statements/block/12.1-2.js), [3](https://github.com/tc39/test262/blob/master/test/language/statements/block/12.1-4.js), [4](https://github.com/tc39/test262/blob/master/test/language/statements/block/12.1-5.js), [5](https://github.com/tc39/test262/blob/master/test/language/statements/block/12.1-6.js), [6](https://github.com/tc39/test262/blob/master/test/language/statements/block/12.1-7.js)) have other grammatical errors that are irrelevant to what's being tested.

**A ton of tests are using `eval` unnecessarily**

I've found literally [over 300 instances](https://gist.github.com/isiahmeadows/ff0791a7dac10f06751006c6be4d6755#file-test262-overrides-json) where I had to replace a test with another because of `eval` getting in the way. And with few exceptions ([like](https://github.com/tc39/test262/blob/master/test/language/statements/continue/S12.7_A2.js) [these](https://github.com/tc39/test262/blob/master/test/language/statements/do-while/S12.6.1_A8.js) [three](https://github.com/tc39/test262/blob/master/test/language/statements/for-in/S12.6.4_A4.1.js)), the vast majority of them could be refactored to check for an early error without meaningfully changing the test. For some examples of how this could be achieved, I did find [two](https://github.com/tc39/test262/blob/master/test/language/white-space/S7.2_A1.5_T2.js) [good](https://github.com/tc39/test262/blob/master/test/language/white-space/S7.2_A1.2_T2.js) examples of what would be helpful for parsers. Those two, I don't have to preprocess and replace their contents, unlike every single other example here.

-----

[...]

PR #1375, opened 2017-12-10
Refactor AssignmentExpression tests for parsers

jugglinmike:
A number of tests for the parsing of the AssignmentExpression production
were expressed using `eval`. This made the tests more complex than
necessary, and also prevented the tests from providing value to
ECMAScript parsers.

expectations with literal source text. Remove superfluous "onlyStrict"
restriction from tests by declaring the probe binding prior to
assignment.

This is part of a larger effort to resolve gh-1356.


test/language/expressions/assignment/id-arguments-strict.js
```diff
@@ -0,0 +1,16 @@
...
+  the LeftHandSideExpression (PrimaryExpression) of simple assignment(=).
+negative:
+  phase: early
+  type: SyntaxError
+flags: [onlyStrict]
+---*/
+
+throw "Test262: This statement should not be evaluated.";
+
+(arguments) = 20;
```
rwaldron: Nice

PR #1381, opened 2017-12-21
Generate type coercion tests for DataView tests

thejoshwolfe:
This is the remainder of the uses of `harness/typeCoercion.js`. That file can be removed in a follow up PR.

(This is an iteration on the proposal in #1341 which does not include the generator code. For that, see https://github.com/thejoshwolfe/test262-generators .)

*I do not have access to an implementation of the apis that are being tested in this PR. I would appreciate if someone could run these to make sure I didn't make any typos.*

(This code was formatted with `js-beautify -n -w 100 -s 2` (after [patching](https://github.com/beautify-web/js-beautify/pull/1303) it to support BigInt literal notation).)


thejoshwolfe:
> I do not have access to an implementation of the apis that are being tested in this PR. I would appreciate if someone could run these to make sure I didn't make any typos.

Ok, I got an implementation, and these tests all pass. This is ready to merge in my opinion.

PR #1383, opened 2017-12-28
Add es[56]?id checks

adrianheine:
I can try to fix the malformed ids if these regular expressions seem agreeable. `es5id` is more complicated.

### DIFF
--- a/tools/lint/lib/checks/esid.py
+++ b/tools/lint/lib/checks/esid.py
@@ -1,0 +1,31 @@
+from ..check import Check
+import re
+
+class CheckEsid(Check):
+    '''Ensure tests specify only valid `es6id's'''
+    ID = 'ESID'
+
+    def __init__(self):
+        #self.es5idRegex = re.compile(r"^S?(B|\d+)(\.\d+)+(-(\d+|[a-z]|i+))*(_A\d(\.\d+)?(_T\d(\.\d+)?)?)?$")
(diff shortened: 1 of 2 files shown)


rwaldron:
I'm not sure I understand what this is trying to address? Many tests have `es5id` or `es6id`, those are still valid but eventually need to be updated to an `esid` (wherever applicable). New test contributions should only have `esid`.

adrianheine:
This is trying to validate the content of `es6id`, `es5id` and `esid` fields. There are some values that seem to be obviously invalid, like `static-semantics-hasproductionintailposition` for `esid` or `B3.5` for `es6id`. This catches them. I'm using the `es[56]?id` fields to group the tests acorn doesn't pass to get a better overview and mistakes like this messed that up a bit.

tools/lint/lib/checks/esid.py
```diff
@@ -0,0 +1,31 @@
...
+    def __init__(self):
+        #self.es5idRegex = re.compile(r"^S?(B|\d+)(\.\d+)+(-(\d+|[a-z]|i+))*(_A\d(\.\d+)?(_T\d(\.\d+)?)?)?$")
+        self.es6idRegex = re.compile(r"^(S?(B|\d+)(\.\d+)+(((_A\d\.\d)?_T?\d)|[ _]S\d+(\.[a-z])*)?(, |$))+")
+        self.esidRegex = re.compile(r"^(pending|(prod|sec)-[-_A-Za-z0-9.%@]+)$")
+
+    def run(self, name, meta, source):
+        if not meta:
+            return
+
+        # es5ids are a mess
```
rwaldron: _nods_ 

Because they are actually from _both_ ES3 and ES5 :(

PR #1388, opened 2018-01-05
flatten and flatMap tests

Shilpi3:


michaelficarra:
I've pushed up a commit to clean some things up. I do think we need a few more basic positive tests, and agree with @littledan that we need to use more non-arrays that are `IsConcatSpreadable` (`arguments` objects) and `Iterable` (strings) as the *contents* of the flattened/flatMapped arrays, not just as the targets.

rwaldron:
@michaelficarra your suggested additions can be done in follow up PRs, right?

test/built-ins/Array/prototype/flatMap/thisArg-argument.js
```diff
@@ -0,0 +1,20 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-array.prototype.flatMap
+description: >
+    Behavior when thisArg is provided
+    Array.prototype.flatMap ( mapperFunction [ , thisArg ] )
+includes: [compareArray.js]
+---*/
+
+"use strict";
```
rwaldron: Remove this line. If this test must only be executed in strict mode, then add the appropriate frontmatter: `flags: [onlyStrict]`

littledan:
Btw a new feature flag wouldn't hurt either, though these are easy to filter anyway as they are in one directory.

rwaldron:
@littledan yes, we need `Array.prototype.flatMap` and `Array.prototype.flatten` added to `features.txt` and added into all of these test files. I figured I could just add them after merging the tests.

michaelficarra:
I've pushed another few commits addressing the feedback, including adding two new feature flags.

@rwaldron Yes, those additions can be done in follow-up PRs if we want implementations to be able to start using these.

rwaldron:
> including adding two new feature flags.

I appreciate that, thank you. 

> Yes, those additions can be done in follow-up PRs if we want implementations to be able to start using these.

👍

PR #1390, opened 2018-01-15
Refactor DeleteExpression tests for parsers

jugglinmike:
A number of tests for the parsing of the DeleteExpression production
were expressed using `eval`. This made the tests more complex than
necessary, and also prevented the tests from providing value to
ECMAScript parsers.

expectations with literal source text. Remove superfluous tests which
only differed in the runtime semantics of source text that could not be
evaluated due to syntax errors.

This patch is in service of gh-1356.


rwaldron:
Nice work, thanks @jugglinmike!

PR #1391, opened 2018-01-17
Add test case for tc39/ecma402#204

anba:
Tests for the proposed change in tc39/ecma402#204


rwaldron:
tc39/ecma402#204 was merged!

PR #1401, opened 2018-01-31
Promise.prototype.finally: add more tests

ljharb:
- per https://github.com/tc39/ecma262/pull/1083#issuecomment-361775023

### DIFF
--- a/test/built-ins/Promise/prototype/finally/invokes-then-with-function.js
+++ b/test/built-ins/Promise/prototype/finally/invokes-then-with-function.js
@@ -27,6 +27,7 @@ target.then = function(a, b) {
 };
 
 var originalFinallyHandler = function () {};
+var anonName = Object(function () {}).name;
 
 var result = Promise.prototype.finally.call(target, originalFinallyHandler, 2, 3);
 
@@ -43,11 +44,16 @@ assert.sameValue(
   'Invokes `then` method with a function as the first argument'
 );
 assert.notSameValue(firstArg, originalFinallyHandler, 'Invokes `then` method with a different fulfillment handler');
+assert.sameValue(firstArg.length, 1, 'fulfillment handler has a length of 1');
+assert.sameValue(firstArg.name, anonName, 'fulfillment handler is anonymous');
+
 assert.sameValue(
   typeof secondArg,
   'function',
   'Invokes `then` method with a function as the second argument'
 );
-assert.notSameValue(secondArg, originalFinallyHandler, 'Invokes `then` method with a different fulfillment handler');
+assert.notSameValue(secondArg, originalFinallyHandler, 'Invokes `then` method with a different rejection handler');
+assert.sameValue(secondArg.length, 1, 'rejection handler has a length of 1');
+assert.sameValue(secondArg.name, anonName, 'rejection handler is anonymous');
 
 assert.sameValue(result, returnValue, 'Returns the result of the invocation of `then`');


test/built-ins/Promise/prototype/finally/invokes-then-with-function.js
```diff
@@ -27,6 +27,7 @@ target.then = function(a, b) {
 };
 
 var originalFinallyHandler = function () {};
+var anonName = Object(function () {}).name;
```
ljharb: The `Object()` call here is to avoid the function name inference pitfall; and pulling it off of a new function is to avoid having to hardcode the logic for an anonymous function name in this test (that should be tested elsewhere).

gsathya:
I've verified that v8 passes the test as well

test/built-ins/Promise/prototype/finally/invokes-then-with-function.js
leobalter: you're assigning the value of `.name` to `anonName`. I wonder if you can just use the empty string directly.

In any case, this won't make the test wrong.

test/built-ins/Promise/prototype/finally/invokes-then-with-function.js
```diff
@@ -43,11 +44,16 @@ assert.sameValue(
...
+assert.sameValue(firstArg.length, 1, 'fulfillment handler has a length of 1');
+assert.sameValue(firstArg.name, anonName, 'fulfillment handler is anonymous');
+
 assert.sameValue(
   typeof secondArg,
   'function',
   'Invokes `then` method with a function as the second argument'
 );
-assert.notSameValue(secondArg, originalFinallyHandler, 'Invokes `then` method with a different fulfillment handler');
+assert.notSameValue(secondArg, originalFinallyHandler, 'Invokes `then` method with a different rejection handler');
```
leobalter: nit: a different handler, used for rejection.

leobalter: while you're there, it might sound a bit silly but you could add an assertion confirming `firstArg !== secondArg`. 

It's also interesting to check that both are functions directly built from `Function.prototype` using `Object.getPrototypeOf` and are not from any subclass, which would give a false positive for typeof.

test/built-ins/Promise/prototype/finally/invokes-then-with-function.js
ljharb: There's some browsers where it's "anonymous"; I wanted to insulate this test against that unrelated failure.

test/built-ins/Promise/prototype/finally/invokes-then-with-function.js
ljharb: hm, subclasses can't give a false positive for `typeof`; typeof uses the internal [[Call]] slot

leobalter:
```js
class F extends Function {}
Object.getPrototypeOf(F) === Function.prototype; // false
typeof F; // "function"
```

ljharb: that's not a false positive for `typeof` tho; what `typeof` is checking is that you can do `F()`, which you can.

The prototype check, `assert.sameValue(Object.getPrototypeOf(F), Function.prototype)` - something that doesn't seem to be commonly done - checks something different.

PR #1402, opened 2018-02-01
Add missing closing parenthesis and remove invalid async flag

anba:
Array/prototype/flatMap/depth-always-one.js
- Someone forgot a closing parenthesis... 😇 

Promise/prototype/finally/subclass-species-constructor-resolve-count.js
- Was incorrectly marked as async.

### DIFF
--- a/test/built-ins/Promise/prototype/finally/subclass-species-constructor-resolve-count.js
+++ b/test/built-ins/Promise/prototype/finally/subclass-species-constructor-resolve-count.js
@@ -5,7 +5,6 @@ author: Jordan Harband
 description: finally on resolved Promise calls the SpeciesConstructor
 esid: sec-promise.prototype.finally
 features: [Promise.prototype.finally]
-flags: [async]
 ---*/
 
 class FooPromise extends Promise {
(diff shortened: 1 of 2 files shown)


test/built-ins/Promise/prototype/finally/subclass-species-constructor-resolve-count.js
```diff
@@ -5,7 +5,6 @@ author: Jordan Harband
 description: finally on resolved Promise calls the SpeciesConstructor
 esid: sec-promise.prototype.finally
 features: [Promise.prototype.finally]
-flags: [async]
```
ljharb: Why is this incorrect? it depends on a resolution that doesn’t happen synchronously.

gsathya: We don't test the resolution here. We only check the result of calling `.finally` .. which happens synchronously, so there's no need for async.

Also, from https://github.com/tc39/test262/blob/master/CONTRIBUTING.md#flags:
**async** - defer interpretation of test results until after the invocation of the global $DONE function

anba: Both assertions in the test only perform checks on the returned object from `Promise.prototype.finally`, which doesn't require any async functionality. I don't know why there's an arrow function argument for `Promise.prototype.finally`, because it is essentially unused for this test. And the similar test https://github.com/tc39/test262/blob/master/test/built-ins/Promise/prototype/finally/subclass-species-constructor-reject-count.js doesn't have the async flag. And if the async flag is correct, the test is still incomplete, because async tests are required to call `$DONE`.

PR #1404, opened 2018-02-04
Refactor function literal tests for parsers

jugglinmike:
A number of tests for the parsing of function literals were expressed
using `eval`. This made the tests more complex than necessary and also
prevented the tests from providing value to ECMAScript parsers.

expectations with literal source text.

---

While writing this patch, I re-discovered a potential conflict between
implementations and the specification text. I filed a patch to the spec here:
https://github.com/tc39/ecma262/pull/1091

I've tried to limit this patch to refactoring and avoid changing test behavior.
Since the issue referenced above is a long-standing one and unrelated to the
intent of this patch, I'd like to address that separately (pending the
resolution of the ECMA-262 issue).

This is in service of gh-1356


test/language/statements/function/invalid-function-body-3.js
```diff
@@ -0,0 +1,17 @@
...
+    Inserting elements that is different from SourceElements into the
+    FunctionBody
+negative:
+  phase: parse
+  type: SyntaxError
+---*/
+
+throw "Test262: This statement should not be evaluated.";
+
+function __func(){# ABC}
```
leobalter: lol

test/language/statements/function/line-terminator-non-strict.js
```diff
@@ -38,6 +38,12 @@ z
 
 z();
 
-eval("function\u0009\u2029w(\u000C)\u00A0{\u000D};");
+// The following function expression is expressed with literal unicode
+// characters so that parsers may benefit from this test. The included code
+// points are as follows:
```
leobalter: +1 for adding this comment.

PR #1407, opened 2018-02-08
Prefer spaces to tabs if it is not syntax related tests

Constellation:
This patch changes tabs to spaces for newly added tests.
The motivation is that SVN of WebKit tree rejects files including tabs.
If we annotate files with `allow-tabs` SVN property, we can land it.
But unless tabs are intentionally used for parser tests, using spaces is a safer choice.
It removes additional steps when updating test262 in WebKit tree, and WebKit tree can quickly update test262.

### DIFF
--- a/test/built-ins/Atomics/wake/wake-all-on-loc.js
+++ b/test/built-ins/Atomics/wake/wake-all-on-loc.js
@@ -70,8 +70,8 @@ function getReport() {
 function waitUntil(ia, k, value) {
     var i = 0;
     while (Atomics.load(ia, k) !== value && i < 15) {
-	$262.agent.sleep(100);
-	i++;
+        $262.agent.sleep(100);
+        i++;
     }
     assert.sameValue(Atomics.load(ia, k), value, "All agents are running");
 }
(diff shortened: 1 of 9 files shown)


rwaldron:
Incidentally, I'm the the process of converting all files away from tabs (as a result of https://bugs.webkit.org/show_bug.cgi?id=182468). I will land this and continue with the complete update.

leobalter:
@ljharb changes for project organization are welcome. The arbitrary restrictions exist but in this case it's because of the lack of consistency here.

After @rwaldron's work the idea is to provide more tools to avoid any mess like this to come up in the future.

Constellation:
Thanks, I think consistency is nice, and it would offer quick cycles of update & test in the third party engines, which is important to improve both test262 and the engines ;)

rwaldron:
Also, I think we can all agree that mixed tabs and spaces is a bummer ;)

ljharb:
Totally; not making an argument for one indentation style over the other, and everybody can agree that mixing is terribad :-)

PR #1409, opened 2018-02-08
js-beautify: make all indentation consistent (depth & character)

rwaldron:
This began as an attempt to safely replace all of the tab indents with spaces (no arguments about this will be acknowledged). 

Please review for correctness. This is just a pass with js-beautify, I won't respond to any subjective nit-picking.


test/built-ins/Array/prototype/every/15.4.4.16-3-17.js
```diff
@@ -17,7 +17,12 @@ function callbackfn2(val, idx, obj) {
   return val > 11;
 }
 
-var obj = { 0: 12, 1: 11, 2: 9, length: "0002.00" };
+var obj = {
+  0: 12,
+  1: 11,
+  2: 9,
+  length: "0002.00"
+};
```
leobalter: A lot of these changes are unnecessary, it creates a longer diff.

It's ok for now but I wonder if jsbeautify is the right tool for the next batches.

anba:
These changes are too large for GitHub, which makes it hard to review them online. Also I think it adds too much churn, like reformatting of object literals, one-liner functions are now spanning multiple lines, requiring two-space indent, etc.

PR #1410, opened 2018-02-09
Templates for TypedArrays

leobalter:
This is a first patch to migrate TypedArray tests to the test generation tool, creating individual tests for each constructor.

It's a migration, other details can be fixed in the process or in follow ups.


leobalter:
This is ridiculous. I will get the tests back to the function iterating over the ES2015 set of TypedArrays. I should split the cases into only this set and BigInt.

PR #1412, opened 2018-02-09
Revert "js-beautify: make all indentation consistent (depth & character)"

anba:
Reverts tc39/test262#1409

Breaks multiple tests, see for example https://github.com/tc39/test262/blob/master/test/built-ins/Function/prototype/toString/function-declaration-non-simple-parameter-list.js


anba:
Also breaks multiple BigInt and RegExp Unicode property name tests. So apparently tests which use new syntax which isn't supported by js-beautify.

leobalter:
thanks, @anba, I got tricked by GitHub and wrongly though it would be limited to the built-ins/Array folder. Turns out we need a better support for long diffs or I need to roll it locally.

rwaldron:
Thanks for spotting this. I wonder if I should break up this work by directory...?

PR #1419, opened 2018-02-12
built-ins/AsyncFunction/*: make all indentation consistent (depth & c…

rwaldron:
…haracter)

### DIFF
--- a/test/built-ins/AsyncFunction/is-not-a-global.js
+++ b/test/built-ins/AsyncFunction/is-not-a-global.js
@@ -8,6 +8,6 @@ description: >
   %AsyncFunction% is not exposed as a global
 ---*/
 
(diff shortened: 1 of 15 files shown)


test/built-ins/AsyncFunction/is-not-a-global.js
```diff
@@ -8,6 +8,6 @@ description: >
   %AsyncFunction% is not exposed as a global
 ---*/
 
-assert.throws(ReferenceError, function () {
+assert.throws(ReferenceError, function() {
```
ljharb: The thing i find most universally consistent is “function keyword, space, optional name, argument list”, but yo probably don’t want bikeshedding here :-)

rwaldron:
Thanks @ljharb

PR #1420, opened 2018-02-12
built-ins/Atomics/*: make all indentation consistent (depth & character)

rwaldron:


rwaldron:
Something is wrong with this commit, so I'm dropping it and will investigate locally

PR #1433, opened 2018-02-12
built-ins/Promise/*: make all indentation consistent (depth & character)

rwaldron:


test/built-ins/Promise/prototype/finally/this-value-non-promise.js
```diff
@@ -11,7 +11,9 @@ features: [Promise.prototype.finally]
 var called = false;
 var p = new Proxy(Promise.resolve(), {});
 var oldThen = Promise.prototype.then;
-Promise.prototype.then = () => { called = true; };
+Promise.prototype.then = () => {
```
ljharb: Oops, this probably shouldn’t have been an arrow function.

test/built-ins/Promise/prototype/then/rxn-handler-rejected-next.js
```diff
@@ -21,33 +21,33 @@ var promise = new Promise(function(_, reject) {
...
-        'Expected each "onFulfilled" handler to be invoked exactly once in series. ' +
-        'Expected: abc. Actual: ' + log
-      );
-      return;
-    }
+  $DONE('This promise should not be fulfilled.');
+}, function() {
+  if (log !== 'abc') {
+    $DONE(
+      'Expected each "onFulfilled" handler to be invoked exactly once in series. ' +
```
ljharb: This error message is incorrect; onFulfilled should be onRejected.

test/built-ins/Promise/prototype/finally/this-value-non-promise.js
leobalter: surely not, but also not really a thing to be fixed on this PR. After this work I should take a look for all the occurrences of arrow fns in the built-ins folder.

test/built-ins/Promise/prototype/then/rxn-handler-rejected-next.js
leobalter: another one to be fixed in a separate PR.

PR #1444, opened 2018-02-12
added coverage for built-in Map spec to ensure that a TypeError is thrown when Symbol.iterator is undefined

nomadtechie:


leobalter:
@ljharb yes, and probably other specific points in these and the respective Weak versions. These can be done in a follow up.

leobalter:
@nomadtechie are you interested in writing the same coverage for Set, WeakMap, and WeakSet?

PR #1445, opened 2018-02-14
Add tests for the behavior of INT32_MIN (-2147483648) as exponent

jswalden:
SpiderMonkey's implementation invokes signed integer overflow in this case.  It would have been nice if test262 had performed this operation and thus would have alerted us to the issue (even if, at least with the compilers we test with regularly, there might not have been an _observable_ deviation from the spec).

It's maybe a little silly having one test for operator, one for function, but this seems better for implementations that choose not to implement the behavior literally by shared function call as the spec does.

### DIFF
--- a/test/built-ins/Math/pow/int32_min-exponent.js
+++ b/test/built-ins/Math/pow/int32_min-exponent.js
@@ -1,0 +1,17 @@
+// Copyright (C) 2018 Mozilla Corporation. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-applying-the-exp-operator
+description: >
+    Using -(2**31) as exponent with Math.pow should behave as expected.
+---*/
+
+const INT32_MIN = -2147483648;
+
+assert.sameValue(Math.pow(2, INT32_MIN), +0.0,
+                 "Math.pow(2, -(gonzo huge exponent > 1074)) should be +0 " +
+                 "because 2**-1074 is the smallest positive IEEE-754 number");
+
+assert.sameValue(Math.pow(1, INT32_MIN), 1,
+                 "1**-(gonzo huge exponent > 1074) should be 1");
--- a/test/language/expressions/exponentiation/int32_min-exponent.js
+++ b/test/language/expressions/exponentiation/int32_min-exponent.js
@@ -1,0 +1,18 @@
+// Copyright (C) 2018 Mozilla Corporation. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-applying-the-exp-operator
+description: >
+    Using -(2**31) as exponent with the exponentiation operator should behave
+    as expected.
+---*/
+
+const INT32_MIN = -2147483648;
+
+assert.sameValue(2**INT32_MIN, +0.0,
+                 "2**-(gonzo huge exponent > 1074) should be +0 because " +
+                 "2**-1074 is the smallest positive IEEE-754 number");
+
+assert.sameValue(1**INT32_MIN, 1,
+                 "1**-(gonzo huge exponent > 1074) should be 1");


test/built-ins/Math/pow/int32_min-exponent.js
```diff
@@ -0,0 +1,17 @@
...
+
+/*---
+esid: sec-applying-the-exp-operator
+description: >
+    Using -(2**31) as exponent with Math.pow should behave as expected.
+---*/
+
+const INT32_MIN = -2147483648;
+
+assert.sameValue(Math.pow(2, INT_MIN), +0.0,
```
bnjbvr: nit: (twice in this file, and twice in the other file) `INT32_MIN`, not `INT_MIN`.

leobalter: Unfortunately, this can't be asserted in Test262. The result is implementation-dependent so the specs won't recommend over edge cases of the exponent value.
 
```
12.6.4 Applying the ** Operator

Returns an implementation-dependent approximation of the result of raising base to the power exponent.

...
```

I'd like to agree with obvious cases, e.g. `2**3`, `2**100`, but going something far from this is something we need to rely on the implementation.

My best suggestion is to propose some basic spec expectations for integer operations limited by IEEE 754 edge values.

jswalden:
I dunno how far the "implementation-dependent approximation" language should be pushed on in tests, but perhaps it's worth noting that existing tests already push on it, assuming behavior that isn't a specifically enumerated edge case.  Just skimming exponentiation tests, these all look over-written:

* https://github.com/tc39/test262/blob/8342824894d7c11788cb454ab388fced6d4f7831/test/language/expressions/exponentiation/exp-assignment-operator.js
* https://github.com/tc39/test262/blob/8342824894d7c11788cb454ab388fced6d4f7831/test/language/expressions/exponentiation/exp-operator-precedence-update-expression-semantics.js
* https://github.com/tc39/test262/blob/8342824894d7c11788cb454ab388fced6d4f7831/test/language/expressions/exponentiation/exp-operator-precedence-unary-expression-semantics.js
* https://github.com/tc39/test262/blob/8342824894d7c11788cb454ab388fced6d4f7831/test/language/expressions/exponentiation/exp-operator-precedence-update-expression-semantics.js

And those are just the exponentiation tests -- I'd bet there are similar dependencies in the `Math.pow` tests.

It seems like the particular tests here could be satisfied by saying integral exponents with base of 2 must be exactly computed, and fully specifying behavior for a base of 1 (going beyond the current specification merely for base 1 and only the exponents +∞ and -∞).  But probably tightening should extend past those narrow cases.  I'm not certain I have the background to propose more rigorous exponentiation behavior, to take this further immediately.  So I'll fix the very-much-not-a-nit `INT_MIN` issue in the PR, so it can be taken whenever the spec's tightened further, but otherwise set this aside for now.

leobalter:
ok, I've had a discussion in person with @rwaldron about this and I had to refresh myself on the IEEE 754 and edge limits for number representations in JS. Considering we'd have to find the minimum possible value for double precision numbers, I have to agree anything less than `2 ** -1074` - which is, btw, `Number.MIN_VALUE` - would not meet anything to be approximated to in JS but probably `0` or `Number.MIN_VALUE` itself.

```
20.1.2.9Number.MIN_VALUE
The value of Number.MIN_VALUE is the smallest positive value of the Number type, which is approximately 5 × 10-324.

In the IEEE 754-2008 double precision binary representation, the smallest possible value is a denormalized number. If an implementation does not support denormalized values, the value of Number.MIN_VALUE must be the smallest non-zero positive value that can actually be represented by the implementation.
```

Quickly checking in browsers, I can tell we confirm the edge from `2**-1074` to `2**-1075`, where the implementations use `+0` as the result.

So, these tests are valid considering we follow IEEE 754 and we defer the approximations to 0 (not MIN_VALUE), and now I wonder if we should also add tests for `2 ** -1075`.

My sincere apologize for initially rejecting this PR, I'm at least glad I did not close to hear back from your feedback which helped me to find where I was wrong.

PR #1448, opened 2018-02-16
Fix errors recently introduced.

leobalter:
Ref #1447.

Needs investigation for more errors, may come in follow up PRs.

PR #1452, opened 2018-02-21
Remove whitespace between Literal and BigIntLiteralSuffix

gsathya:


leobalter:
Thanks for catching this. This change would be fixed by #1448 but that's a larger set of fixes and yet requires a final review.

PR #1455, opened 2018-02-25
Add missing feature flag to TA + BigInt test

adrianheine:
Fixup for #1413.

### DIFF
--- a/test/built-ins/TypedArray/prototype/set/array-arg-src-tonumber-value-type-conversions.js
+++ b/test/built-ins/TypedArray/prototype/set/array-arg-src-tonumber-value-type-conversions.js
@@ -35,14 +35,7 @@ testWithTypedArrayConstructors(function(TA) {
       }
   };
 
-  // undefined and NaN covered on typedArrayConversions
-  var nullish;
-  try {
-    nullish = null;
-  } catch (e) {
-    nullish = 0n;
-  }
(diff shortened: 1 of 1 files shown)


test/built-ins/TypedArray/prototype/set/array-arg-src-tonumber-value-type-conversions.js
```diff
@@ -19,7 +19,7 @@ info: |
     kNumber).
   ...
 includes: [testTypedArray.js, compareArray.js]
-features: [TypedArray]
+features: [BigInt, TypedArray]
```
leobalter: this test is actually a bit wrong and should not have BigInt as a feature here.

```js
  var nullish;
  try {
    nullish = null;
  } catch (e) {
    nullish = 0n;
  }
  var arr = ["1", "", false, true, nullish, obj1, obj2, [], [1]];
```

We can remove lines 39 to 44 and replace line 45 with:

```js
var arr = ["1", "", false, true, null, obj1, obj2, [], [1]];
```

I have not captured this issue on #1448 because this test will still pass if the syntax for BigInt is implemented, but this is never testing any conversion with BigInt values. The catch block is dead code.

Do you want update this PR with this fix?

We still need to implement separate test files for type conversions in bigint TAs, but that's another task force.

adrianheine: Done.

PR #1456, opened 2018-02-26
Additional tests for QuantifiableAssertion

littledan:
- Tests that lookahead and lookbehind are not extended to
  QuantifiableAssertion, as in https://github.com/tc39/ecma262/pull/1102
- Additional tests verifying some more combinations of cases for
  QuantifiableAssertion being invalid in Unicode mode.

Based on the tests in https://chromium-review.googlesource.com/c/v8/v8/+/926102

These tests pass on V8 (if the throw for early errors is removed to
work around a V8 issue where RegExps don't have early errors).


littledan:
Thanks @hashseed for originally writing these tests and implementing the new semantics in V8!

littledan:
Oops, probably I was too eager to merge this, not sure what I was thinking... the PR itself doesn't explicitly have consensus.

PR #1460, opened 2018-02-26
async-iteration: AsyncFromSyncIteratorPrototype next/throw/return

spectranaut:


test/built-ins/AsyncFromSyncIteratorPrototype/next/iterator-result-poisoned-done.js
```diff
@@ -0,0 +1,52 @@
...
+var obj = {
+  [Symbol.iterator]() {
+    return {
+      next() {
+	return {
+	  get done() {
+	    throw thrownError;
+	  },
+	  value: 1
+	}
```
rwaldron: strange indentation

(occurs in other files as well)

PR #1464, opened 2018-02-28
numeric literal seperators: test change in CodePoint production

spectranaut:


test/language/literals/regexp/unicode-escape-nls-err.js
```diff
@@ -0,0 +1,73 @@
...
+
+features: [numeric-separator-literal]
+negative:
+  phase: parse
+  type: SyntaxError
+---*/
+
+throw "Test262: This statement should not be evaluated.";
+
+/\u{1F_639}/u;
```
spectranaut: this test fails when I run test harness for d8 -- but it does early error in chrome/firefox. Is there something obvious you see I'm missing, @leobalter ?

leobalter: I don't think you're missing anything here, this test should be fine. Maybe you just caught an implementation bug?

cc @mathiasbynens

mathiasbynens: V8 doesn’t throw *early* errors for invalid regular expression patterns. https://bugs.chromium.org/p/v8/issues/detail?id=896 It still throws at runtime, though.

leobalter: Thanks for the clarification and showing the already known bug, @mathiasbynens!

Issue #1467, opened 2018-03-02
Minor bugs in BigInt tests


jakobkummerow:
CC @rwaldron @leobalter @gsathya 

I've found two minor issues with BigInt tests:

**built-ins/BigInt/prototype/Symbol.toStringTag.js:**
"verifyProperty" is destructive unless {restore: true} is passed. This makes the subsequent assert fail.
Fix:

```
--- a/test/built-ins/BigInt/prototype/Symbol.toStringTag.js
+++ b/test/built-ins/BigInt/prototype/Symbol.toStringTag.js
@@ -19,7 +19,7 @@ verifyProperty(BigInt.prototype, Symbol.toStringTag, {
   writable: false,
   enumerable: false,
   configurable: true
-});
+}, {restore: true});
 
 assert.sameValue(Object.prototype.toString.call(3n), "[object BigInt]");
 assert.sameValue(Object.prototype.toString.call(Object(3n)), "[object BigInt]");
```

**built-ins/TypedArrays/ctors-bigint/typedarray-arg/typedarray-arg-other-ctor-returns-new-typedarray.js:**
The inputs must have length 7 in order for the newly constructed arrays to have length 7 as well.
Fix:

```
--- a/test/built-ins/TypedArrays/ctors-bigint/typedarray-arg/typedarray-arg-other-ctor-returns-new-typedarray.js
+++ b/test/built-ins/TypedArrays/ctors-bigint/typedarray-arg/typedarray-arg-other-ctor-returns-new-typedarray.js
@@ -14,8 +14,8 @@ includes: [testBigIntTypedArray.js]
 features: [BigInt, TypedArray]
 ---*/
 
-var sample1 = new BigInt64Array();
-var sample2 = new BigUint64Array();
+var sample1 = new BigInt64Array(7);
+var sample2 = new BigUint64Array(7);
 
 testWithBigIntTypedArrayConstructors(function(TA) {
   var sample = TA === BigInt64Array ? sample2 : sample1;
```

littledan:
cc @cxielarko

cxielarko:
These changes LGTM! (the second test fails in spidermonkey too; the first passes but probably shouldn't)

PR #1473, opened 2018-03-06
typed-arrays/bigint: fixes per recommendation. Closes gh-1467

rwaldron:


test/built-ins/BigInt/prototype/Symbol.toStringTag.js
```diff
@@ -19,7 +19,7 @@ verifyProperty(BigInt.prototype, Symbol.toStringTag, {
   writable: false,
   enumerable: false,
   configurable: true
-});
+}, {restore: true});
```
leobalter: restore is set here for a bad practice for tests. Tests should be atomic and assertions should not rely on other assertion behaviors or even order of execution in the matters of any other assertions.

The best practice here would be isolating the assertions if they can break or explode others. In Test262 case, we should split this into multiple files. It's fun that `restore` would become totally useless after that, because it does affect the assertion it's applied to.

leobalter: This should be applied to all cases relying on this restore, I don't see a good reason to use it otherwise.

test/built-ins/TypedArrays/ctors-bigint/typedarray-arg/other-ctor-returns-new-typedarray.js
```diff
@@ -14,8 +14,8 @@ includes: [testBigIntTypedArray.js]
 features: [BigInt, TypedArray]
 ---*/
 
-var sample1 = new BigInt64Array();
-var sample2 = new BigUint64Array();
+var sample1 = new BigInt64Array(7);
+var sample2 = new BigUint64Array(7);
```
leobalter: while this is great, the description message is useless here. Can you fix this one too?

test/built-ins/BigInt/prototype/Symbol.toStringTag.js
rwaldron: > This should be applied to all cases relying on this restore, I don't see a good reason to use it otherwise

Agreed.

rwaldron: > restore is set here for a bad practice for tests. Tests should be atomic and assertions should not rely on other assertion behaviors or even order of execution in the matters of any other assertions.

> The best practice here would be isolating the assertions if they can break or explode others. In Test262 case, we should split this into multiple files. It's fun that restore would become totally useless after that, because it does affect the assertion it's applied to.

Agreed. I will break these out.

PR #1477, opened 2018-03-07
additional Function.prototype.toString tests

michaelficarra:
This reflects some recent changes in the [Function.prototype.toString revision](https://github.com/tc39/Function-prototype-toString-revision) proposal. I recommend reviewing the commits individually. The final commit adds a test that reflects the current status of the proposal, but will not pass in any current implementations. But it's good to add it, as it will make whatever change we make in the future (assuming we do make one) clear.


rwaldron:
As this proposal is Stage 3, this is ok to land!

PR #1480, opened 2018-03-07
improve propertyHelper for failure messages

leobalter:
Example:
Before: _descriptor value should be 42 Expected SameValue(«1», «0») to be true_
After: _descriptor value should be 42_

---

Note that `Expected SameValue(«1», «0»)` is for the list of failures while checking the descriptor, as each error adds up to the list. The error message is confusing if the expected value is similar, e.g. 1:

> descriptor value should be 42 Expected SameValue(«1», «0») to be true

### DIFF
--- a/harness/propertyHelper.js
+++ b/harness/propertyHelper.js
@@ -73,7 +73,7 @@ function verifyProperty(obj, name, desc, options) {
     }
   }
 
-  assert.sameValue(failures.length, 0, failures.join('; '));
+  assert(!failures.length, failures.join('; '));
 
   if (options && options.restore) {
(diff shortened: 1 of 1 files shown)


leobalter:
@ljharb this is exactly the problem, the assertion was using internal actual and expected values, not anything from the user input.

This assertion was only telling if it found any failures while checking the descriptor.

Note that we already use the same [assertion message](https://github.com/tc39/test262/pull/1480/files#diff-c45e3dedeee3d3c6528ec3f417fccf0dR49) - through `failures.join('; ')` - to inform of a different value or other failures.

PR #1483, opened 2018-03-09
Default parameter value for catch binding

chicoxyzzy:
Tests for https://github.com/tc39/ecma262/pull/1126


leobalter:
well, this is just the tip for checking formal parameters. We thrive to maintain at least most of the variances we can observe for the syntax. That means checking the FormalParameter will observe the binding patterns (destructuring, etc). We also need to check for edge cases as early errors, naming conflicts, verifying it's a single formalparameter (and not a list), etc.

Also, it would be great to avoid using eval whenever is possible.

We have some templates and cases in the `src/` folder for destructuring that might be helpful for creating the new cases.

I'm interested to help with more information and support, just let me know.

chicoxyzzy:
@leobalter thank you for your feedback! As far as I understand, test for most of destructuring patterns and other cases are already implemented here https://github.com/tc39/test262/tree/master/test/language/statements/try

As for avoiding eval, I'm not sure how to rewrite my code. For example,
```js
try {
 throw undefined;
} catch (err = "foo") {
 assert.sameValue(err, 'foo');
}
```
will throw early syntax error on `=`. Should I use `!assert.throws` here?

Any help will be appreciated!


Also I noticed that `assert.throws.early` has unused local variable `ieval`.

leobalter:
> test for most of destructuring patterns and other cases are already implemented here https://github.com/tc39/test262/tree/master/test/language/statements/try

Yes, and some are generated. I'm experimenting here with a new template for catch initializers and I'll get you a feedback anytime soon.

> (...) will throw early syntax error on =. Should I use !assert.throws here?

The early error is intentional, and a good sign. You should let this throw a Syntax Error while it's not implemented. It should only pass when it's not anymore. It will get green when the feature is available.

leobalter:
@chicoxyzzy I just reused some templates for destructuring cases of the catch initializer here: https://github.com/tc39/test262/compare/master...leobalter:catch-initializer?expand=1

There are also cases to be used from `src/function-forms` that would apply to the catch initializer, but I think some of them could need some clarification in the produced specs, some others are just useful for checking edge cases like:

```js
try {
  throw undefined;
} catch (_ = (function() { throw new Test262Error(); }()) {}
```

or cases where falsy values won't trigger the initializer:

```js
throw null;
throw NaN;
throw 0;
throw false;
throw "";
throw {};
```

also consider expanding the tests with `throw void 0;`

I would also check this is being set to a single FormalParameter so rest parameters and trailing commas are still not allowed and they should continue throwing a SyntaxError

chicoxyzzy:
Thank you for such detailed answer! I'll return to this PR in next few days.

rwaldron:
This has Stage 4

chicoxyzzy:
Oops. I'll better create new PR

chicoxyzzy:
New PR https://github.com/tc39/test262/pull/2162

Issue #1476, opened 2018-03-06
Weaken tests about NaN canonicalization


littledan:
TC39 reached consensus on https://github.com/tc39/ecma262/pull/758 , which permits certain kinds of flexibility with NaN canonicalization. test262 includes tests that are invalid with that patch, some of which you can find listed in https://github.com/v8/v8/blob/master/test/test262/test262.status#L226 . If someone is interested in volunteering, it'd be great if we had a test PR to match the spec PR.

cc @ajklein

rwaldron:
cc @leobalter @spectranaut @nomadtechie

rwaldron:
@littledan @ajklein 

(noting the relevant change here)

> either the same encoding for each implementation distinguishable *NaN* value, or an implementation-defined canonical value.

IIUC, Test262 could satisfy the desired flexibility by just checking if the NaN bit pattern is in range? (Between 7FF0000000000000 and 7FFFFFFFFFFFFFFF, or FFF0000000000000 and FFFFFFFFFFFFFFFF)

littledan:
@rwaldron Well, if the NaN bit pattern is in range, it would be spec-compliant. The specification implies something a little more specific, but something which is pretty hard to write a test for as it's a hard to get at the particular shared implementation-dependent pattern.

Between the complexity of writing more specific tests, and @jfbastien's explanation that real implementations that don't do canonicalization are unlikely to satisfy even the newly more relaxed specification anyway, and the committee having repeatedly expressed being OK with V8's lack of NaN canonicalization, I'd be fine if we just had tests as you describe.

leobalter:
being in range doesn't mean we would observe canonicalization. The spec might become more relaxed but still requires some minimal matching for each production form of `NaN`. e.g.: `0/0` will always match the same bit pattern in Float arrays, canonicalized or not with other productions, like `Infinity/Infinity`.

rwaldron:
> being in range doesn't mean we would observe canonicalization

My interpretation is that since there are "options", that we can't depend on one or the other and must not explicitly test for matching NaN encodings, instead we'll just check that a NaN is a valid NaN and that's it.

littledan:
I agree with Leo that it would be possible to construct a test which is a little more specific--you could run several different scenarios and check that, each time, you rather get the same value out or a single canonical one shared between them. It could be tricky to get at the cases that might trip up a real implementation, though--to have full test coverage would require knowledge of details of a particular implementation.

rwaldron:
[...]

PR #1484, opened 2018-03-09
Weaken tests about NaN canonicalization. Fixes gh-1476

rwaldron:
This combines our ideas into relaxed "Is it a NaN?" tests.


rwaldron:
This combines our ideas into relaxed "Is it a NaN?" tests.

harness/nans.js
```diff
@@ -7,6 +7,17 @@ description: |
...
+  () => NaN * 0,
+  () => 0/0,
+  () => Infinity/Infinity,
+  () => -(0/0),
+  () => Math.pow(-1, 0.5),
+  () => -Math.pow(-1, 0.5),
+  () => Number("Not-a-Number"),
 ];
+
+var distinctNaNs = NaNs.map(n => n());
```
ljharb: In my experimenting with https://npmjs.com/get-nans , i saw a number of engines where something like this would cause it to get canonicalized, such that inlining it directly into the array was the only way to retain the ability to distinguish them.

rwaldron: Ah, good to know, I will adjust accordingly.

leobalter:
this is manually merged at 2975694, as I had to fix some conflicts introduced after this PR.

PR #1495, opened 2018-03-16
Add more tests for Atomics wait

leobalter:


test/built-ins/Atomics/wait/cannot-suspend-throws.js
```diff
@@ -0,0 +1,25 @@
+// Copyright (C) 2018 Amal Hussein. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
```
leobalter: @nomadtechie covers this part: 

> when [[CanBlock]] is false for an agent: call Atomics.wait on the main thread and expect that it will throw a TypeError since you can not block on the main thread

test/built-ins/Atomics/wait/was-woken-before-timeout.js
```diff
@@ -0,0 +1,55 @@
+// Copyright (C) 2018 Amal Hussein. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
```
leobalter: covers the item 21:

> If awoken is true, return the String "ok"
>  call Atomics.wait with a timeout, put an agent to sleep, and then call Atomics.wake before the timeout, and expect an output of "ok"

test/built-ins/Atomics/wait/not-an-object-throws.js
```diff
@@ -2,7 +2,7 @@
 // This code is governed by the BSD license found in the LICENSE file.
 /*---
 esid: sec-atomics.wait
-description:
+description: >
```
leobalter: those are for consistency

test/built-ins/Atomics/wait/was-woken-before-timeout.js
```diff
@@ -0,0 +1,55 @@
...
+var timeout = 20000;
+$262.agent.start(
+    `
+  $262.agent.receiveBroadcast(function (sab) { 
+    var int32Array = new Int32Array(sab);
+    $262.agent.report(Atomics.wait(int32Array, 0, 0, ${timeout}));
+    $262.agent.leaving();
+  })
+  `);
+
```
nomadtechie: I don't think `timeout` here shares the same scope -- you will need to declare it inside the agent.

Issue #1490, opened 2018-03-13
Coverage: Promise.all with non-iterable


bakkot:
A value can fail to be iterable in all kinds of exciting ways, including:

- Being `null` or `undefined`
- Missing a`Symbol.iterator` field or otherwise having it be non-callable
- Having a getter for `Symbol.iterator` which throws
- Having an `Symbol.iterator` method which returns non-objects

When provided any of these as its first argument, `Promise.all` is meant to return a rejected promise, rather than throwing. For at least the last of these, v8 throws, so I suspect there's a hole in the coverage.

ljharb:
https://tc39.github.io/ecma262/#sec-promise.all says that a nullish argument should definitely throw and not return a rejected promise. GetIterator can also throw synchronously, as can the IsConstructor check.

bakkot:
@ljharb, that's not my reading. Remember that abstract operations can't "throw" per se; they just return completion records, which may be abrupt completion records representing thrown exceptions. None of the steps past `NewPromiseCapability` are using [`ReturnIfAbrupt`](https://tc39.github.io/ecma262/#sec-returnifabrupt) or the `?` or `!` macros to unwrap a completion record; they instead use [`IfAbruptRejectPromise`](https://tc39.github.io/ecma262/#sec-ifabruptrejectpromise), which is like `ReturnIfAbrupt` except that it returns a rejected promise rather than throwing.

ljharb:
Step 2 explicitly says to throw a TypeError if it’s a primitive, at least.

On a reread, you’re correct about GetIterator check, but the IsConstructor check and Construct and IsCallable calls inside NewPromiseCapability have a ?, meaning they unpack an abrupt completion using ReturnIfAbrupt.

bakkot:
Step 2 says to throw a TypeError if the *receiver* is primitive.

Errors in NewPromiseCapability do cause synchronous throws, yes. This issue is just about the argument, though, not the receiver. Though I guess it would also be good to assert the synchronous behavior if that's not covered.

ljharb:
aha, thanks, that clarifies my misunderstanding :-) I now completely agree with your original issue, and apologies for the confusion.

My copy of Chrome seems to return a resolved promise for `Promise.all([{ [Symbol.iterator]() {return null} }])`, but it seems like it should return a rejected one?

Either way, we should definitely have tests that cover all of this behavior!

bakkot:
> apologies for the confusion

No worries, this stuff is inherently confusing.

[...]

PR #1496, opened 2018-03-19
Promise.all with non-iterable or invalid return from Symbol.iterator. Fixes gh-1490

rwaldron:


rwaldron:
@bakkot good suggestions, I'll do some review of existing test material and I'll get some updates pushed here (if necessary).

test/built-ins/Promise/all/iter-arg-is-false-reject.js
```diff
@@ -0,0 +1,33 @@
...
+    ...
+
+    #sec-getiterator
+    GetIterator ( obj [ , hint [ , method ] ] )
+
+    ...
+    Let iterator be ? Call(method, obj).
+    If Type(iterator) is not Object, throw a TypeError exception.
+    ...
+features: [Symbol.iterator]
```
leobalter: some of these tests are not relying on the Symbol.iterator at all but just the _iterator_ argument used on Promise.all and .race. I believe these won't need this feature tag, unless Symbol.iterator is explicitly used.

This can be fixed in a follow up.

ljharb: They do implicitly, because the iterator protocol that the methods use looks up Symbol.iterator - anything with the iteration protocol relies on Symbol.iterator.

leobalter:
This is wrong.

Features tags are to flag features fully necessary for the tests to be
compatible with. The spec could remove this feature entirely and the test
would be exactly the same. An implementation can fully skip Symbol.iterator
and the other spec parts are still valid for this code. I could also use an
object with length to simulate the iterator and it’s still valid.

ljharb:
The iterable can only be looked up via Symbol.iterator however - so without that value existing, nothing is iterable.

leobalter:
my fault to write this through my phone. I honestly thought the specs would try to fetch an iterator from any _arrayLike_ object with a .length property.

ljharb:
As far as I understand it, `Array.from` is the only thing in the spec that handles both iterables and arraylikes, so it's the only iterable-related thing that *wouldn't* require `Symbol.iterator` to exist to be usable.

leobalter:
TypedArray constructors receiving objects as arguments do it as well.

ljharb:
TIL, ty!

PR #1497, opened 2018-03-19
additional test coverage for atomics.wait + removing old tests which have been replaced + test cleanup and renaming test files for consistency

nomadtechie:


test/built-ins/Atomics/wait/wait-index-value-not-equal.js
```diff
@@ -0,0 +1,51 @@
...
+  `
+$262.agent.receiveBroadcast(function (sab) {
+  var int32Array = new Int32Array(sab);
+    
+  $262.agent.report(Atomics.wait(int32Array, 0, 44, 100));
+  
+  $262.agent.report(Atomics.wait(int32Array, 0, 251.4, 1000));
+    
+  $262.agent.report(Atomics.wait(int32Array, 0, Infinity, 1000));
+
```
nomadtechie: FYI @leobalter 
Regarding `$262.agent.report(Atomics.wait(int32Array, 0, Infinity, 1000));` the value here gets converted to `Int32` (Let v be ? ToInt32(value).)

Its currently timing out instead of returning `not-equal`

test/built-ins/Atomics/wait/did-timeout.js
```diff
@@ -23,9 +23,12 @@ $262.agent.receiveBroadcast(function (sab, id) {
 var ia = new Int32Array(new SharedArrayBuffer(Int32Array.BYTES_PER_ELEMENT));
 
 $262.agent.broadcast(ia.buffer);
+
 assert.sameValue(getReport(), "timed-out");
 assert.sameValue((getReport() | 0) >= 500 - $ATOMICS_MAX_TIME_EPSILON, true);
 
+assert.sameValue(Atomics.wake(ia, 0), 0);
```
leobalter: why do we ever need to call wake here?

test/built-ins/Atomics/wait/negative-timeout.js
```diff
@@ -27,3 +27,5 @@ function getReport() {
     $262.agent.sleep(100);
   return r;
 }
+
+assert.sameValue(Atomics.wake(ia, 0), 0);
```
leobalter: same question on calling .wake?

test/built-ins/Atomics/wait/non-int32-typedarray-throws.js
```diff
@@ -0,0 +1,58 @@
...
+
+assert.throws(TypeError, function() {
+  Atomics.wait(new Float64Array(), poisoned, poisoned, poisoned);
+}, 'Float64Array');
+
+assert.throws(TypeError, function() {
+  Atomics.wait(new Float32Array(), poisoned, poisoned, poisoned);
+}, 'Float32Array');
+
+
```
leobalter: double spacing.

test/built-ins/Atomics/wait/out-of-range-index-throws.js
```diff
@@ -0,0 +1,28 @@
...
+var int32Array = new Int32Array(new SharedArrayBuffer(1024));
+var poisoned = {
+  valueOf: function() {
+    throw new Test262Error("should not evaluate this code");
+  }
+};
+
+assert.throws(RangeError, () => Atomics.wait(int32Array, Infinity, poisoned, poisoned));
+assert.throws(RangeError, () => Atomics.wait(int32Array, int32Array.length + 1, poisoned, poisoned));
+assert.throws(RangeError, () => Atomics.wait(int32Array, int32Array.length + 100, poisoned, poisoned));
```
leobalter: static values are always better for clarify in the assertions. Also, avoid using arrow functions here.

test/built-ins/Atomics/wait/out-of-range-index-throws.js
```diff
@@ -0,0 +1,28 @@
...
+
+  2.Let i be ? ValidateAtomicAccess(typedArray, index).
+    ...
+    2.Let accessIndex be ? ToIndex(requestIndex).
+    ...
+    5. If accessIndex ≥ length, throw a RangeError exception.
+features: [ Atomics, SharedArrayBuffer, TypedArray ]
+---*/
+
+var int32Array = new Int32Array(new SharedArrayBuffer(1024));
```
leobalter: these SAB lengths are too long. A minimal value (4) can be used. That also makes it even easier for testing `length + 1` as the argument should be only `2`.

test/built-ins/Atomics/wait/poisoned-object-for-timeout-throws.js
```diff
@@ -34,28 +31,31 @@ function getReport() {
...
     toString: false
   };
-
-  var err;
   
   try {
     Atomics.wait(int32Array, 0, 0, poisoned);
   } catch(e) {
-    err = e.constructor;
+    $262.agent.report(e.name);
```
leobalter: that's a rollback. Keep err outside, don't report anything from inside the catch block. So you guarantee you always have a .report call

test/built-ins/Atomics/wait/poisoned-object-for-timeout-throws.js
```diff
@@ -34,28 +31,31 @@ function getReport() {
...
+var int32Array = new Int32Array(sab);
 
 $262.agent.broadcast(int32Array.buffer);
 
 $262.agent.sleep(150);
 
-assert.sameValue(getReport(), TypeError);
+assert.sameValue(getReport(), 'TypeError');
+
+assert.sameValue(Atomics.wake(int32Array, 0), 0);
```
leobalter: another rollback, why calling .wake here? It's already reported.

test/built-ins/Atomics/wait/shared-nonint-views.js
```diff
@@ -1,22 +0,0 @@
...
-
-var other_views = [Int8Array, Uint8Array, Int16Array, Uint16Array, Uint32Array,
-                   Uint8ClampedArray, Float32Array, Float64Array];
-
-testWithTypedArrayConstructors(function(View) {
-  var view = new View(sab);
-
-  // Even with timout zero this should fail
-  assert.throws(TypeError, (() => Atomics.wait(view, 0, 0, 0)));
-}, other_views);
```
leobalter: thanks for expanding this

test/built-ins/Atomics/wait/symbol-for-timeout-throws.js
```diff
@@ -25,23 +24,27 @@ function getReport() {
...
+$262.agent.receiveBroadcast(function (sab) {  
+
+  var int32Array = new Int32Array(sab);
   
   try {
-    Atomics.wait(int32Array, 0, 0, s);
+    Atomics.wait(int32Array, 0, 0, Symbol('foo'));
   } catch(e) {
-    err = e.constructor;
+   $262.agent.report(e.name);
```
leobalter: same issue with the report inside the catch block. Also, bad indentation

test/built-ins/Atomics/wait/symbol-for-timeout-throws.js
```diff
@@ -25,23 +24,27 @@ function getReport() {
...
 `);
 
-var int32Array = new Int32Array(new SharedArrayBuffer(Int32Array.BYTES_PER_ELEMENT));
+var sab = new SharedArrayBuffer(4);
+var int32Array = new Int32Array(sab);
 
 $262.agent.broadcast(int32Array.buffer);
 
-assert.sameValue(getReport(), TypeError);
+$262.agent.sleep(150);
```
leobalter: getReport will already do the trick calling sleep, it's not necessary to call this here.

test/built-ins/Atomics/wait/symbol-for-timeout-throws.js
```diff
@@ -25,23 +24,27 @@ function getReport() {
...
+
+assert.sameValue(getReport(), 'TypeError');
```
leobalter: Why rolling back from a string to the actual constructor? Unless we get a different evaluation because it's an object?

test/built-ins/Atomics/wait/true-for-timeout.js
```diff
@@ -38,8 +38,11 @@ var int32Array = new Int32Array(new SharedArrayBuffer(Int32Array.BYTES_PER_ELEME
 
 $262.agent.broadcast(int32Array.buffer);
 
-$262.agent.sleep(2);
+$262.agent.sleep(150);
```
leobalter: ?

test/built-ins/Atomics/wait/wait-index-value-not-equal.js
```diff
@@ -0,0 +1,51 @@
...
+  14.If v is not equal to w, then
+    a.Perform LeaveCriticalSection(WL).
+    b. Return the String "not-equal".
+
+features: [ Atomics, SharedArrayBuffer, TypedArray ]
+---*/
+
+function getReport() {
+  var r;
+  while ((r = $262.agent.getReport()) == null)
```
leobalter: please use a block here instead of a inline while.

test/built-ins/Atomics/wait/true-for-timeout.js
nomadtechie: Will revert -- the sleep in the while loop covers for the minimum required wait time

PR #1499, opened 2018-03-20
wip: atomics wake + additional wait tests

nomadtechie:


leobalter:
Tests seem fine for me. We probably might work on a tweak or another after merging but I'd need some extra feedback from engines running these.

PR #1501, opened 2018-03-22
Date.parse breathing test, time value maximum range test

apaprocki:
The specification directly calls out the parse identity of a `Date` `x` where `x.valueOf() === 0`. This adds a basic test for that base case and then adds an additional test for the maximum allowable *time value* range as stated in the spec.

The latter issue was raised in tc39/ecma262#1145, and currently Firefox (SpiderMonkey) is the only engine returning `NaN` for values outside of the *time value* range. This test is intended to bring the other engines in line with SM behavior.


rwaldron:
Pending outcome of https://github.com/tc39/ecma262/pull/1144

PR #1502, opened 2018-03-22
Fix various test bugs

### LINKED ISSUE #1492: Test failures at b07c546b8aa9ae6ded0599673e9e643a5ee4e7ea
4 test files with errors at b07c546b8aa9ae6ded0599673e9e643a5ee4e7ea

https://github.com/tc39/test262/blob/b07c546b8aa9ae6ded0599673e9e643a5ee4e7ea/test/built-ins/AsyncFromSyncIteratorPrototype/throw/iterator-result-unwrap-promise.js#L49
- `yield* obj;` needs to be `return yield* obj;`, otherwise the return value of the `throw()` handler will be ignored.

https://github.com/tc39/test262/blob/b07c546b8aa9ae6ded0599673e9e643a5ee4e7ea/test/built-ins/Atomics/wait/negative-index-throws.js#L27
- Doesn't throw a RangeError because `ToIndex(-0.999) == 0` 

https://github.com/tc39/test262/blob/b07c546b8aa9ae6ded0599673e9e643a5ee4e7ea/test/built-ins/Atomics/wait/negative-index-throws.js#L29
- Doesn't throw a RangeError because `ToIndex(-0.00000000000000001) == 0` 

https://github.com/tc39/test262/blob/b07c546b8aa9ae6ded0599673e9e643a5ee4e7ea/test/built-ins/Atomics/wait/negative-index-throws.js#L30
- Doesn't throw a RangeError because `ToIndex(NaN) == 0` 

https://github.com/tc39/test262/blob/b07c546b8aa9ae6ded0599673e9e643a5ee4e7ea/test/built-ins/TypedArray/prototype/map/return-new-typedarray-conversion-operation-consistent-nan.js#L53
- Missing function invocation -> `return NaNs[i++]();`.

https://github.com/tc39/test262/blob/b07c546b8aa9ae6ded0599673e9e643a5ee4e7ea/test/harness/verifyProperty-value-error.js#L32
- `«0»` is incorrect in the expected assertion message.

anba:
I've changed harness/testBigIntTypedArray.js so it can be loaded in even if `BigInt64Array` is not defined. This makes it easier to process this file for SpiderMonkey's test262 importer.

Fixes #1492

### DIFF
--- a/features.txt
+++ b/features.txt
@@ -146,3 +146,4 @@ WeakSet
 # object, go here.
 
 IsHTMLDDA
+CannotSuspendMainAgent
--- a/harness/features.yml
+++ b/harness/features.yml
@@ -1,4 +1,5 @@
 typeCoercion.js: [Symbol.toPrimitive, BigInt]
 testAtomics.js: [ArrayBuffer, Atomics, DataView, SharedArrayBuffer, arrow-function, let, for-of]
+testBigIntTypedArray.js: [BigInt, TypedArray]
 testTypedArray.js: [TypedArray]
 isConstructor.js: [Reflect.construct]
--- a/harness/testBigIntTypedArray.js
+++ b/harness/testBigIntTypedArray.js
@@ -8,7 +8,7 @@ description: |
 /**
  * The %TypedArray% intrinsic constructor function.
  */
-var TypedArray = Object.getPrototypeOf(BigInt64Array);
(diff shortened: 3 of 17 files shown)


features.txt
```diff
@@ -146,3 +146,4 @@ WeakSet
 # object, go here.
 
 IsHTMLDDA
+CannotSuspendMainAgent
```
leobalter: +1

PR #1503, opened 2018-03-22
Don't compute NaN values in functions

anba:
`function nan() { return /* expression to compute NaN */ }` may not return the same implementation-distinguishable NaN value before and after JIT compilation. As observed here: https://treeherder.mozilla.org/#/jobs?repo=try&revision=cb26ed14f1a42ae9a70b516ecc22046c3ce9c8fa

### DIFF
--- a/test/built-ins/Object/internals/DefineOwnProperty/nan-equivalence-define-own-property-reassign.js
+++ b/test/built-ins/Object/internals/DefineOwnProperty/nan-equivalence-define-own-property-reassign.js
@@ -57,12 +57,12 @@ for (var idx = 0; idx < len; ++idx) {
   for (var jdx = 0; jdx < len; ++jdx) {
     var a = {};
 
-    a.prop = NaNs[idx]();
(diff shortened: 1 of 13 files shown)


test/built-ins/Object/internals/DefineOwnProperty/nan-equivalence-define-own-property-reassign.js
```diff
@@ -57,12 +57,12 @@ for (var idx = 0; idx < len; ++idx) {
...
 
-    a.prop = NaNs[idx]();
-    a.prop = NaNs[jdx]();
+    a.prop = NaNs[idx];
+    a.prop = NaNs[jdx];
 
     assert(
       a.prop !== a.prop,
-      `Object property value reassigned to NaN produced by (${NaNs[idx].toString()}) results in a valid NaN`
+      `Object property value reassigned to NaN produced by (index=${idx}) results in a valid NaN`
```
leobalter: thanks for finding a way to show which form of NaN we are using.

PR #1506, opened 2018-04-06
Remove errant LF from CR line ending test.

rkirsling:
We've had a [bit](https://bugs.webkit.org/show_bug.cgi?id=184266) of a [runaround](https://bugs.webkit.org/show_bug.cgi?id=184341) getting test262 updated over in the WebKit repo, but ultimately, the cause of the issue seems to be that the test for CR line endings ends in—_gasp_—CRLF!

I think this should do the trick. :smile:


spectranaut:
hey @michaelficarra -- that LF was added in 9737f3704ff854464d74a7a933d1e0ab4f9d2f00 -- not intentional right?

michaelficarra:
No definitely not! Sorry about that. It's not easy working with such files.

spectranaut:
thanks @rkirsling ! :)

PR #1507, opened 2018-04-06
Add test for proposal-class-fields#92

bakkot:
This adds a test for the change in https://github.com/tc39/proposal-class-fields/pull/92. At a glance I don't believe the previous behavior was tested (it is _very_ edge-case-y), but the important part of the change for the old-vs-new behavior in this would look like

```diff
-  field = (thisDuringField = this, assert.throws(ReferenceError, probeCtorThis));
+  field = (thisDuringField = this, thisFromProbe = probeCtorThis());
```

cc @littledan

### DIFF
--- a/test/language/expressions/class/constructor-this-tdz-during-initializers.js
+++ b/test/language/expressions/class/constructor-this-tdz-during-initializers.js
@@ -1,0 +1,43 @@
+// Copyright (C) 2018 Kevin Gibbons. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-super-keyword-runtime-semantics-evaluation
+description: >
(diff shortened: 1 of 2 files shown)


gsathya:
Needs test that verifies that the super constructor runs twice on a double super call but the field initializers are run only once

bakkot:
Good catch, @gsathya. I hadn't noticed that the original semantics were to run the field initializers twice, and I think that would be pretty unintuitive, so I am now even more in favor of https://github.com/tc39/proposal-class-fields/pull/92.

Added a test.

rwaldron:
LGTM. I've searched through the existing tests to find coverage for the previous semantics and could not find anything.

littledan:
Thanks for your help in being so thorough, @rwaldron.

PR #1509, opened 2018-04-10
Corrections to sec-web-compat-evaldeclarationinstantiation and sec-web-compat-functiondeclarationinstantiation tests. Fixes gh-1364

### LINKED ISSUE #1364: Incorrect build for /language/function-code/switch-dflt-decl-strict.js
The file in [test/language/function-code/switch-dflt-decl-strict.js](https://github.com/tc39/test262/tree/master/test/language/function-code/switch-dflt-decl-strict.js) is missing half its headers and incorrect has `[onlyStrict]`, which leads to false positives. The source file [src/annex-b-fns/func/switch-dflt.template](https://github.com/tc39/test262/tree/master/src/annex-b-fns/func/switch-dflt.template) correctly has the `[noStrict]` tag and I'm not sure where it sheds this (I'm not a py dev).

If that is the wrong source file then I'm not sure which is but the output is wrong regardless as this is a func statement inside a default statement is a syntax error in strict mode.

I think there may be multiple such problems with this template. I didn't see a related issue filed when quickly going through the open list.

rwaldron:
(Some changes are only for consistency)

Fixes #1364

### DIFF
--- a/test/language/eval-code/direct/block-decl-eval-source-is-strict-onlystrict.js
+++ b/test/language/eval-code/direct/block-decl-eval-source-is-strict-onlystrict.js
@@ -1,26 +1,22 @@
 // Copyright (C) 2016 the V8 project authors. All rights reserved.
 // This code is governed by the BSD license found in the LICENSE file.
 /*---
-description: AnnexB extension not honored in strict mode
-es6id: B.3.3.3
+esid: sec-web-compat-evaldeclarationinstantiation
+description: >
+    AnnexB extension not honored in strict mode, Block statement
+    in eval code containing a function declaration
 info: |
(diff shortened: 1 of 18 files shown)


test/language/eval-code/direct/block-decl-eval-source-is-strict-onlystrict.js
```diff
@@ -0,0 +1,28 @@
...
+
+eval('"use strict";{ function f() {} }');
+
+try {
+  f;
+} catch (exception) {
+  err = exception;
+}
+
+assert.sameValue(err.constructor, ReferenceError);
```
leobalter: 👍

PR #1513, opened 2018-04-16
bigint comparison tests

cxielarko:
Add tests for comparisons between BigInt and string values. This commit checks for the string-to-bigint coercion changes in tc39/proposal-bigint@4768527d9dcf4c04b22a7382d41091e470d8a7fd.

### DIFF
--- a/test/language/expressions/greater-than-or-equal/bigint-and-incomparable-string.js
+++ b/test/language/expressions/greater-than-or-equal/bigint-and-incomparable-string.js
@@ -1,0 +1,30 @@
+// Copyright (C) 2018 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Relational comparison of BigInt and string values
+esid: sec-abstract-relational-comparison
+features: [BigInt]
+---*/
+
+assert.sameValue(1n >= "0n", false, "1n >= '0n'");
+assert.sameValue(1n >= "0.", false, "1n >= '0.'");
+assert.sameValue(1n >= ".0", false, "1n >= '.0'");
+assert.sameValue(1n >= "0/1", false, "1n >= '0/1'");
+assert.sameValue(1n >= "z0", false, "1n >= 'z0'");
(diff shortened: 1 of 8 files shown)


test/language/expressions/greater-than-or-equal/bigint-and-incomparable-string.js
```diff
@@ -0,0 +1,30 @@
+// Copyright (C) 2018 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Relational comparison of BigInt and string values
+esid: sec-abstract-relational-comparison
+features: [BigInt]
```
leobalter: The following info would be really helpful here:

```
7.2 StringToBigInt ( argument )

Apply the algorithm in 3.1.3.1 with the following changes:

- Replace the StrUnsignedDecimalLiteral production with DecimalDigits to not allow Infinity, decimal points, or exponents.
- If the MV is NaN, return NaN, otherwise return the BigInt which exactly corresponds to the MV, rather than rounding to a Number.
```

I can add that after merging this

test/language/expressions/greater-than-or-equal/bigint-and-string.js
```diff
@@ -0,0 +1,48 @@
...
+
+assert.sameValue(9007199254740993n >= "9007199254740992", true,
+                 "9007199254740993n >= '9007199254740992'");
+assert.sameValue("9007199254740993" >= 9007199254740992n, true,
+                 "'9007199254740993' >= 9007199254740992n");
+
+assert.sameValue(-9007199254740992n >= "-9007199254740993", true,
+                 "-9007199254740992n >= '-9007199254740993'");
+assert.sameValue("-9007199254740992" >= -9007199254740993n, true,
+                 "'-9007199254740992' >= -9007199254740993n");
```
leobalter: we should exaust the possible strings we can convert, this means including cases for binary, octal, and hex literals and whitespace (at least a common space and line terminator characters).

PR #1514, opened 2018-04-16
remove IsSafeInteger checks for BigInt

cxielarko:
Update tests for the BigInt constructor change in tc39/proposal-bigint@c80774327c9a223a2814f183c80c414574cca700 (changes the constructor tests, and removes an IsSafeInteger check in the DataView tests)


test/built-ins/BigInt/constructor-integer.js
```diff
@@ -0,0 +1,31 @@
...
+assert.sameValue(BigInt(-Number.MAX_SAFE_INTEGER), -9007199254740991n);
+
+var pos = Math.pow(2, 53);
+var neg = -pos;
+
+assert.sameValue(BigInt(pos), 9007199254740992n,
+                 "BigInt(2**53) === 9007199254740992n");
+
+assert.sameValue(BigInt(neg), -9007199254740992n,
+                 "BigInt(2**53) === -9007199254740992n");
```
rwaldron: This file's name is "constructor-integer.js" and the description says "BigInt constructor", but the tests are exercising `BigInt` called as a function.

test/built-ins/BigInt/constructor-integer.js
```diff
@@ -0,0 +1,31 @@
+// Copyright (C) 2018 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: BigInt constructor called with integer argument
+esid: sec-bigint-constructor
```
rwaldron: sec-bigint-constructor-number-value

test/built-ins/BigInt/constructor-integer.js
rwaldron: Side-note: I think the proposal specification prose should also be further refined.

rwaldron: https://github.com/tc39/proposal-bigint/pull/145

PR #1522, opened 2018-04-22
fix: missing includes in Array.from-descriptor.js

teppeis:
It uses `verifyWritable` but does not include `propertyHelper.js`.

### DIFF
--- a/test/built-ins/Array/from/Array.from-descriptor.js
+++ b/test/built-ins/Array/from/Array.from-descriptor.js
@@ -7,6 +7,7 @@ includes:
     - propertyHelper.js
 esid: sec-array.from
 es6id: 22.1.2.1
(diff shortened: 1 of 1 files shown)


jugglinmike:
Thanks for the patch! The modified file already contains the appropriate `includes` data, albeit in a different format:

```js
/*---
description: Testing descriptor property of Array.from
includes:
    - propertyHelper.js
esid: sec-array.from
es6id: 22.1.2.1
---*/
```

https://github.com/tc39/test262/blob/c55d2ab7c344e35a3ceb93cf1d4d30019584db82/test/built-ins/Array/from/Array.from-descriptor.js#L6-L7

My initial guess is that maybe the YAML parser you are using does not recognize the other list syntax.

Issue #1472, opened 2018-03-05
Test Coverage Audit & Tracking for Atomics.wake 


nomadtechie:
Below is a checklist for the observable items in the Atomics.wake spec: https://tc39.github.io/ecma262/#sec-atomics.wake.

I've mapped and noted the coverage of the current 17 test files and will be adding coverage for the remaining observables. I'll also be refactoring a few of the current tests because they cover too many spec observables in one test file, or partially cover the spec. 

Some of the observables I've listed below may not be directly testable via test262. I'll keep this list updated as I find unobservable abstractions and will mark them as NOT DIRECTLY OBSERVABLE

### Atomics.wake Static Property
- [x] Check Name
  - test/built-ins/Atomics/wake/name.js
- [x] Check length
  - test/built-ins/Atomics/wake/length.js
- [x] Check descriptor
  - test/built-ins/Atomics/wake/descriptor.js

### Atomics.wake( typedArray, index, count )
1. Let buffer be ? ValidateSharedIntegerTypedArray(typedArray, true)
  - [x] test for abrupt return if ValidateSharedIntegerTypedArray throws an error
    - test/built-ins/Atomics/wake/not-an-object-throws.js
    - test/built-ins/Atomics/wake/non-int32-typedarray-throws.js
    - test/built-ins/Atomics/wake/null-bufferdata-throws.js
    - test/built-ins/Atomics/wake/non-shared-bufferdata-throws.js
    - test/built-ins/Atomics/wake/not-a-typedarray-throws.js
  - **Observables from  *ValidateSharedIntegerTypedArray*:**
    - [x] if Type(typedArray) is not Object, throw a TypeError exception
      - test/built-ins/Atomics/wake/not-an-object-throws.js
      - **Covered but test cases cover multiple spec lines. Requires rework to isolate test cases:** test/built-ins/Atomics/wake/non-views.js
    - [x] If typedArray does not have a [[TypedArrayName]] internal slot, throw a TypeError exception
      - test/built-ins/Atomics/wake/not-a-typedarray-throws.js
    - [x] If typeName is not "Int32Array", throw a TypeError exception.
      - test/built-ins/Atomics/wake/non-int32-typedarray-throws.js
      - test/built-ins/Atomics/wake/shared-nonint-views.js
    - If IsSharedArrayBuffer(buffer) is false, throw a TypeError exception
      - test/built-ins/Atomics/wake/non-shared-int-views.js
    - **Observables from  *IsSharedArrayBuffer*:**
[...]

PR #1523, opened 2018-04-23
Atomics.wake: additional coverage. Completes gh-1472

rwaldron:


test/built-ins/Atomics/wake/symbol-count-throws.js
```diff
@@ -0,0 +1,25 @@
+// Copyright (C) 2018 Rick Waldron.  All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-atomics.wake
+description: >
+  Allowed boundary cases of the third 'count' argument to Atomics.wake
```
spectranaut: I realize this is nitpicking -- but this description makes not sense to me. Write "Return abrupt when symbol passed as 'count' agrument to Atomics.wake"

rwaldron: Agreed. I will fix.

PR #1525, opened 2018-04-24
Fix various test bugs

anba:
- "CannotSuspendMainAgent" feature was changed to "CanBlockIsFalse" flag
- Move annex-b tests into annex-b directory
- Update variable names in nonshared-int-views.js tests
- Move getReport() call in nan-for-timeout.js to avoid iloop
- Update BigInt constructor to match new semantics (tc39/proposal-bigint#138)


rwaldron:
Thanks for the quick repairs

PR #1537, opened 2018-04-28
Remove superfluous code from negative syntax tests

jugglinmike:
@rwaldon gh-1534 got me thinking about other places where `$ERROR` was being used unnecessarily. I'm sure there are plenty more overly-complicated "negative syntax" tests, but we'll leave those for another day.

---

Because these files contain syntax errors, the code they contain is not
intended to be executed, and the runtime semantics are therefore
irrelevant. Simplify the files by removing the unnecessary code.


rwaldron:
Thanks @jugglinmike!

PR #1543, opened 2018-05-07
Add JSON ⊂ ECMAScript tests

mathiasbynens:
Proposal: https://github.com/tc39/proposal-json-superset


mathiasbynens:
cc @gibson042

test/language/line-terminators/line-separator-string-literal.js
```diff
@@ -0,0 +1,11 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-literals-string-literals
+description: >
+  LINE SEPARATOR (U+2028) is allowed within string literals.
+features: [json-superset]
+---*/
+
+eval("'foo\u2028bar\u2028baz\u2028qux'");
```
leobalter: this should go a bit beyond here. We don't need eval for these new tests, that's only legacy we don't need to import from the removed file.

At the same time, it's also a good opportunity to duplicate these tests with the literal cases using unescaped and escaped forms in each test file. I can work in a follow up if it's such a burn to add more here, let me know and I'd be glad to help.

features.txt
```diff
@@ -77,6 +77,10 @@ numeric-separator-literal
 String.prototype.matchAll
 Symbol.matchAll
 
+# JSON ⊂ ECMAScript
```
michaelficarra: This is probably better written as "ECMAScript ⊃ JSON" given the proposal name.

test/language/line-terminators/line-separator-string-literal.js
mathiasbynens: I used `eval` to avoid having to put the raw U+2028 and U+2029 characters in the source code, which is hard to review and would be easy to accidentally break in the future. This approach seemed more maintainable.

mathiasbynens:
Thanks for the feedback. Patch updated.

test/language/line-terminators/line-separator-string-literal.js
bakkot: `eval` is annoying because it makes these tests not really consumable by parsers and other consumers which don't implement `eval`. (See #1356 etc.) I think it's worth having the literal character in the source.

gibson042:
Oops, I opened #1544 without seeing this. My mistake.

mathiasbynens:
I like the approach in #1544. Let’s go with that PR instead.

PR #1544, opened 2018-05-07
Tests for proposal-json-superset

gibson042:
https://github.com/tc39/proposal-json-superset/issues/9

Allow non-ASCII _LineTerminator_ characters (U+2028 LINE SEPARATOR and U+2029 PARAGRAPH SEPARATOR) in string literals.


ljharb:
Duplicate of #1543; is there anything here that could be added there?

mathiasbynens:
I’m fine with either #1543 or this patch.

`features.txt` needs to be updated:

```
ECMAScript ⊃ JSON
# https://github.com/tc39/proposal-json-superset
json-superset
```

And then the tests need the `json-superset` tag.

rwaldron:
@gibson042 this is good to go with or without the update to features.txt and addition of `json-superset` (I can take care of it if you don't have time).

leobalter:
Bikeshedding some ideas to follow up in a new patch:

- single quote strings
- the `.length` property being affected by these characters
- the .codePointAt returns
- syntax being constructed through eval where a line separator is not allowed... e.g.: `eval('async \u2029 function fn() {}')` and `eval('async \u2028 function fn() {}')`

rwaldron:
👍

PR #1546, opened 2018-05-08
Added tests for relational comparison among BigInt and Symbol

### LINKED ISSUE #1542: Missing tests for relational comparison operators among BigInt and Symbol
I couldn't find tests for BigInt - Symbol comparisons. According to Spec, it throws an exception because ```ToNumber``` will be called to convert symbol to a number.

caiolima:
These tests are important too check if evaluation ordering are being respected when some of the operands are BigInt. It closes #1542

### DIFF
--- a/test/language/expressions/greater-than/bigint-and-symbol.js
+++ b/test/language/expressions/greater-than/bigint-and-symbol.js
@@ -1,0 +1,17 @@
+// Copyright (C) 2018 Caio Lima. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Relational comparison of BigInt and Symbol values
+esid: sec-abstract-relational-comparison
+features: [BigInt, Symbol]
+---*/
+
+assert.throws(TypeError, function() {
+  3n > Symbol("2");
+}, "ToNumeric(rhs) throws.");
+
+assert.throws(TypeError, function() {
+  Symbol("2") > 3n;
+}, "ToNumeric(rhs) throws.");
+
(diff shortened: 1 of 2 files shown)


test/language/expressions/greater-than/bigint-and-symbol.js
```diff
@@ -0,0 +1,17 @@
...
+features: [BigInt, Symbol]
+---*/
+
+assert.throws(TypeError, function() {
+  3n > Symbol("2");
+}, "ToNumeric(rhs) throws.");
+
+assert.throws(TypeError, function() {
+  Symbol("2") > 3n;
+}, "ToNumeric(rhs) throws.");
```
rwaldron: Would you mind adding something to either of the comments that differentiates them? For example: `"ToNumeric(BigInt) on RHS throws."`

test/language/expressions/less-than/bigint-and-symbol.js
```diff
@@ -0,0 +1,19 @@
...
+
+function MyError() {}
+
+assert.throws(TypeError, function() {
+  3n < Symbol("2");
+}, "ToNumeric(rhs) throws.");
+
+assert.throws(TypeError, function() {
+  Symbol("2") < 3n;
+}, "ToNumeric(rhs) throws.");
```
rwaldron: Same request as above

test/language/expressions/greater-than/bigint-and-symbol.js
caiolima: That's really good.

Issue #1047, opened 2017-05-21
I want a CLA bot.


leobalter:
@bterlson there's a lot of nice CLA bots around, can we use any here? I believe the same is good for Ecma262 and other repos.

@kborchers worked in the one currently used for the @JSFoudation, hopefully we can join forces.

bterlson:
@leobalter I am supportive of a CLA Bot. Want to get one set up here and I'll incorporate into ECMA262?

kborchers:
@leobter @bterlson I definitely recommend CLA Assistant which is what we use for JSF projects. One shortcoming is that it doesn't support corporate CLAs so if you need that it would need to be added. The JSF would also be happy to provide a server for the bot. In fact, we could probably work it out so you can use our bot as it can assign different CLAs to different GitHub orgs or repos. We'd need to talk through the details but I think it would be doable. Then you don't even need to stand up your own. Happy to chat this week if any of this sounds interesting.

bterlson:
It sounds interesting for sure, though I'm not sure how everything works so unsure how to proceed. I don't have a problem sharing bots (our IRC logbot is a similar situation).

leobalter:
Cool, I'll talk to @bterlson during this meeting and see if I can draft something with the CLA Assistant.

rwaldron:
Since no one has stepped in to do this and I've had no issues checking the spreadsheet, I'm going to close this issue. @leobalter you're welcome to re-open if you find the time, or can assign someone to do it.

PR #1550, opened 2018-05-09
Update CONTRIBUTING.md

hotsphink:
typo in constructor Test252Error


jugglinmike:
Holding off on merging because I'm not able to verify CLA signatures.

@leobalter Can you send me instruction for how to do that? Or maybe we can just focus on gh-1047 instead.

rwaldron:
@jugglinmike I can confirm that @hotsphink is Steve Fink from Mozilla :)

leobalter:
@jugglinmike we should still focus on gh-1047, but at this point we need volunteers to bring a bot. I believe filling a google docs / form should be enough. @littledan might confirm this.

jugglinmike:
@rwaldron @leobalter I've realized that I didn't make the issue clear: I don't have access to the shared document where the signatures are stored. Can you share access with me?

PR #1552, opened 2018-05-11
Move the empty calendar option to the 'valid' test.

Ms2ger:
I misread the ABNF syntax.

### DIFF
--- a/test/intl402/Locale/constructor-options-calendar-invalid.js
+++ b/test/intl402/Locale/constructor-options-calendar-invalid.js
@@ -20,10 +20,9 @@ features: [Intl.Locale]
 
 /*
  alphanum = (ALPHA / DIGIT)     ; letters and numbers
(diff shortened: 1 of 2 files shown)


Ms2ger:
@littledan please confirm this is what you intended the spec to say; the inline ABNF is kinda hard to read.

rwaldron:
@littledan I've reviewed this and am ready to land, pending your response to @Ms2ger's question

rwaldron:
Status?

Ms2ger:
We need to make a call on the spec issue, and then probably close this PR as no-longer-needed.

rwaldron:
@Ms2ger thanks for the update!

littledan:
We made the call that this case actually *should* be invalid. See https://github.com/tc39/proposal-intl-locale/pull/23#issuecomment-389355014

PR #1560, opened 2018-05-18
Intl: Update language subtag data to "IANA language subtag registry, file date 2018-04-23"

rwaldron:
https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry


mathiasbynens:
Is this process automated? If not, we could make a build script that takes care of this.

rwaldron:
@mathiasbynens I wrote a throw-away script to generate the lists

Issue #1515, opened 2018-04-16
Feature flag for more Intl tests using particular data?


littledan:
It's currently difficult to write many completely "correct" tests against ECMA-402 (Intl). ECMA-402 defines the shape of internationalization data in general, but it doesn't define what the data is. And it shouldn't define what the data is--over time, conventions change, data quality improves, and the same code outputs different results, without either of the two results being "wrong".

However: This state presents a problem for testing. It's not possible to write a completely correct test262 test which checks that the string output is any particular string--the only tests that are completely correct test more superficial things about the form. So far, JS implementations have been maintaining non-upstreamed individual tests for this purpose.

I was wondering, what if we made a feature flag behind which we could share Intl tests which assume a little bit about locale data? For example, it could be a feature flag for en-US data, and we could use our subjective judgment to keep the tests generally within what is unlikely to change over time. This would enable more tests to be written and shared between implementations in test262. We could update the standard harness to skip these tests by default, and implementations could specifically opt into them, to minimize confusion.

@anba has found a ton of observable differences in behavior between ECMA-402 implementations which don't actually have to do with differences in underlying locale data, but rather how things are hooked up on top of that. I think the lack of test262 test coverage in this area is a cause, and this category of tests is an important component.

@ms2ger is beginning to write tests for various Stage 3 Intl proposals; in this context, it would be great to figure out if we could add a test category like this.

rwaldron:
@littledan I don't think this is a valid use of "feature" flags. "feature" flags are tagging a test file to indicate that a specific language feature is being used or tested in that file. It's also not clear to me how a "feature" flag would do anything special to change some observable aspect of a test. Could you write an example test that includes the sort of "en-US data" you're thinking of?

littledan:
As an example, we could test, "if NumberFormat has the en-US locale, then 2.3 is formatted as "2.3"'.

Without some kind of assumptions on what data looks like, it's hard to write many tests about Intl. These tests are useful even if they are technically assuming more than what the spec says.

We discussed this kind of testing in the April 2018 Intl call and a bunch of browser vendors and specification authors would like to share tests here. It would be the easiest to use test262, to have all the tests in one place and since everyone has harnesses set up in the right place already.

Maybe feature is the wrong idea, I was just suggesting someplace that we could place a flag. I'd be fine representing the metadata some other way. The higher level question is whether this is permissible within test262, and if so, how.

[...]

PR #1561, opened 2018-05-18
Frontmatter: Adds top level "locale" attribute, w/ "tags" attribute to declare language tags, subtags. Fixes gh-1515

rwaldron:
cc @littledan

### DIFF
--- a/INTERPRETING.md
+++ b/INTERPRETING.md
@@ -353,3 +353,27 @@ following strings:
 - **`generated`** The test file was created procedurally using the project's
   test generation tool. This flag is specified for informational purposes only
   and has no bearing on how the test should be interpreted.
+
+### `locale`
+
+The `locale` attribute allows tests to declare explicit information regarding locale specificity. Its value is an array of one or more valid language tags or subtags.
+
+  *Example*
+
+  ```js
+  /*---
+  locale: [en, en-US, ar]
+  ---*/
+  
+  var en = new Intl.PluralRules('en');
+  assert.sameValue(en.select(1), 'one', 'en.select(1) returns "one"');
+  assert.sameValue(en.select(2), 'other', 'en.select(2) returns "other"');  
+
+  var enUS = new Intl.PluralRules('en-US');
+  assert.sameValue(enUS.select(1), 'one', 'enUS.select(1) returns "one"');
+  assert.sameValue(enUS.select(2), 'other', 'enUS.select(2) returns "other"');
+
+  var ar = new Intl.PluralRules('ar');
+  assert.sameValue(ar.select(1), 'one', 'ar.select(1) returns "one"');
+  assert.sameValue(ar.select(2), 'other', 'ar.select(2) returns "two"');
+  ```
--- a/tools/lint/lib/checks/frontmatter.py
+++ b/tools/lint/lib/checks/frontmatter.py
@@ -3,7 +3,7 @@ from ..check import Check
 _REQUIRED_FIELDS = set(['description'])
 _OPTIONAL_FIELDS = set([
     'author', 'es5id', 'es6id', 'esid', 'features', 'flags', 'includes',
-    'info', 'negative', 'timeout'
+    'info', 'locale', 'negative', 'timeout'
 ])
 _VALID_FIELDS = _REQUIRED_FIELDS | _OPTIONAL_FIELDS
 


INTERPRETING.md
```diff
@@ -352,3 +352,30 @@ following strings:
...
+The `locale` attribute allows tests to declare explicit information regarding locale specificity.
+
+- **`tags`** A list of valid language tags or subtags found in this test. Use IANA Language Subtag Registry to validate.
+
+  *Example*
+
+  ```js
+  /*---
+  locale: 
+    tags: [en, en-US, ar]
```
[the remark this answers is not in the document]
rwaldron: That's what I was hoping we'd learn from @littledan.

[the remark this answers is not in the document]
littledan: If we're being pedantic about things, there's all sorts of details that could be included here--which Intl constructors are actually supported on the locale, what all the locale data is, etc are not directly implied by the locale tag itself. I like the idea in this patch since it's *not* pedantically self-contained.

I like Leo's idea of using something like `locale: [en]` for this purpose. I can imagine lots of hypothetical extensions, but I don't have a particular use case for any in mind; this is already more elaborate than what I was initially picturing.

Whether or not we make that syntax change, let's make sure the INTERPRETING.md explains the sloppiness here, for example, adding the disclaimer:

> Tests with `locale` specified make assumptions about the specific locale data supported which go beyond what is implied by any specification. The information included is determined pragmatically based on common data for locales, e.g., from the CLDR. Implementations which use significantly different data may conform to the specification precisely but fail these tests.

test/intl402/Locale/extensions-grandfathered.js
```diff
@@ -31,6 +31,8 @@ info: |
     ...
 
 features: [Intl.Locale]
+locale:
```
littledan: Intl.Locale supports all locales, regardless of what's supported on a given system. I think this frontmatter should only be used on tests which are *not* required to work, e.g., in your PluralRules example above.

test/intl402/Locale/extensions-private.js
```diff
@@ -31,6 +31,8 @@ info: |
     ...
 
 features: [Intl.Locale]
+locale:
+    tags: [x-default]
```
littledan: Ditto, I'm not sure we need a tag here.

littledan:
Thanks for following up with the patch here, and apologies for my delay on the review.

jugglinmike:
> Should this bump the version in the package.json?

Neither @rwaldron's patch nor @littledan's recommended modification describe any alteration to the way the test should be interpreted. If there really is no change, then this is semantically equivalent to introducing a new feature flag.  In other words, no need for a new version of Test262.

INTERPRETING.md
```diff
@@ -352,3 +352,30 @@ following strings:
 - **`generated`** The test file was created procedurally using the project's
   test generation tool. This flag is specified for informational purposes only
   and has no bearing on how the test should be interpreted.
+
+### `locale`
+
+The `locale` attribute allows tests to declare explicit information regarding locale specificity.
+
+- **`tags`** A list of valid language tags or subtags found in this test. Use IANA Language Subtag Registry to validate.
```
jugglinmike: "Use IANA Language Subtag Registry to validate." That seems more like instruction to test contributors--do you think it would fit better in CONTRIBUTING.md?

INTERPRETING.md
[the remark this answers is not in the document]
rwaldron: I'll be honest, your answer seems to say "I like both". 

I am trying to address this: 

> I was wondering, what if we made a feature flag behind which we could share Intl tests which assume a little bit about locale data? For example, it could be a feature flag for en-US data, and we could use our subjective judgment to keep the tests generally within what is unlikely to change over time


I don't think this information should be stored as a "feature" because it's not a new "feature" of the language in the same sense that e.g. SharedArrayBuffer is a new feature. 

I will change it to @leobalter's suggestion and then you can fill in the frontmatter of the relevant tests

littledan:
This is all looking pretty great to me. Ready to merge? Merging this patch will enable us to make tests for other Intl proposals, e.g., https://github.com/tc39/ecma402/pull/232#issuecomment-396301233

One nit: You might want to squash the patches and remove the commit message " Intl.Locale: two examples using "locale" frontmatter " since you took out these examples.

INTERPRETING.md
leobalter: I prefer having all the tags presented in the same place, and this phrase is too small to go to contributing as it is. Maybe we should follow up to add a paragraph with more context in that other doc? For that I would not block this current PR.

leobalter:
> One nit: You might want to squash the patches and remove the commit message " Intl.Locale: two examples using "locale" frontmatter " since you took out these examples.

I always prefer to squash commits, unless they have generated files as I prefer to keep these in separated commits.

> This is all looking pretty great to me. Ready to merge?

It looks good to me, but I'd like a +1 from @jugglinmike before going ahead. See https://github.com/tc39/test262/pull/1561#discussion_r194519205

Ms2ger:
LGTM

PR #1564, opened 2018-05-20
Adding prefixed string to compare with BigInt

### LINKED ISSUE #1562: Missing comparisons between String with Hex, binary and octal prefix and BigInt
We are missing following test cases:

```
"0x10" < 17
"0b10" > 1
"0o10" < 9
```

caiolima:
It closes #1562


caiolima:
Thank you for the review!

PR #1572, opened 2018-05-24
SuperProperty evaluation order

rwaldron:
super[super()] should evaluate and resolve super() first

### DIFF
--- a/test/language/statements/class/super/in-constructor-superproperty-evaluation.js
+++ b/test/language/statements/class/super/in-constructor-superproperty-evaluation.js
@@ -3,14 +3,15 @@
 /*---
 esid: sec-makesuperpropertyreference
 description: >
-  SuperProperty evaluation order: super() thisBinding initialization occurs first.
+  SuperProperty evaluation order: super() thisBinding initialization must occur first.
 ---*/
 class Derived extends Object {
   constructor() {
     super[super()];
+    throw new Test262Error();
   }
 }
 
-var derived = new Derived();
-assert.sameValue(derived instanceof Derived, true);
-assert.sameValue(derived instanceof Object, true);
+assert.throws(ReferenceError, function() {
+  new Derived();
+}, '`super[super()]` via `new Derived()` throws a ReferenceError');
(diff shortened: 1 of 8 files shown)


rwaldron:
@leobalter we also need @jridgewell to ✅

test/language/statements/class/super/in-constructor-superproperty-evaluation.js
```diff
@@ -0,0 +1,16 @@
+// Copyright (C) 2018 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-makesuperpropertyreference
+description: >
+  SuperProperty evaluation order: super() thisBinding initialization occurs first.
+---*/
+class Derived extends Object {
+  constructor() {
+    super[super()];
```
jridgewell: Do we want to test the return value?

rwaldron: If you think it's necessary, then sure. I reduced it to this because I was able to observe passing/failing in the runtimes you listed. We can also do both. Let me know!

jridgewell: This is fine with me.

ljharb:
Spec PR: https://github.com/tc39/ecma262/pull/1207#issuecomment-391888816

jridgewell:
I should have said this before, but this is the pre-change behavior. Was that intentional?

Post https://github.com/tc39/ecma262/pull/1207, this should throw a reference error.

rwaldron:
> Was that intentional?

Sort of. My thinking was this: let's have a test of the existing behavior (since none exists) and then update it to the new behavior.

test/language/statements/class/super/in-constructor-superproperty-evaluation.js
```diff
@@ -0,0 +1,17 @@
...
+    throw new Test262Error();
+  }
+}
+
+assert.throws(ReferenceError, function() {
+  new Derived();
+}, '`super[super()]` via `new Derived()` throws a ReferenceError');
```
rwaldron: @jridgewell ping

jridgewell: 👍

leobalter:
Merged both commits. This is already addressing the new behavior.

PR #1574, opened 2018-05-29
Add tests for the names of getters on Intl.Locale.prototype.

Ms2ger:


test/intl402/Locale/prototype/caseFirst/name.js
```diff
@@ -0,0 +1,25 @@
...
+const propdesc = Object.getOwnPropertyDescriptor(Intl.Locale.prototype, "caseFirst");
+if (propdesc !== undefined) {
+  const getter = propdesc.get;
+  verifyProperty(getter, "name", {
+    value: "get caseFirst",
+    writable: false,
+    enumerable: false,
+    configurable: true,
+  });
+}
```
rwaldron: What if `propdesc === undefined`?

Ms2ger: Then there isn't really anything to test, is there? The spec explicitly allows for it to be omitted ("This property only exists if %Locale%.[[RelevantExtensionKeys]] contains "kf".")

PR #1576, opened 2018-05-29
Add a test for pluralrules.resolvedOptions().pluralCategories.

Ms2ger:


Ms2ger:
See: https://github.com/tc39/ecma402/pull/233.

PR #1581, opened 2018-06-02
GetTemplateObject with a unique template object

WebReflection:
While Firefox never behaved as such, the specifications are clear and Chrome, Safari, and Edge always respected specifications [here](https://www.ecma-international.org/ecma-262/8.0/index.html#sec-gettemplateobject):

> Each TemplateLiteral in the program code of a realm is associated with a unique template object that is used in the evaluation of tagged Templates (12.2.9.5). The template objects are frozen and the same template object is used each time a specific tagged Template is evaluated.

However, since there are no tests around that little details that has a lot of libraries built around such assumption, it's easy to break the web like [this Chrome 67 bug](https://bugs.chromium.org/p/chromium/issues/detail?id=848999) demonstrates.

Please note that behavior is so trusted that even Babel transpiler would return true after executing this code:
```js
(function(s){ return s; })`a` ===
(function(s){ return s; })`a`; // true
```

Please add this test ASAP so that NodeJS 10, as well as future versions of Chrome, can go back to be aligned with what has worked as such for at least the last 2 years.

**Thank You**.

### DIFF
--- a/test/language/expressions/tagged-template/template-object.js
+++ b/test/language/expressions/tagged-template/template-object.js
@@ -8,11 +8,17 @@ info: |
     defined by the GetTemplateObject abstract operation.
 includes: [propertyHelper.js]
 ---*/
-var templateObject;
+var templateObject, sameObject;
 
-(function(parameter) {
+function sameSite() {
+  tag`${Math.random()}`;
+}
+
+function tag(parameter) {
   templateObject = parameter;
-})`${1}`;
+}
+
+tag`${1}`;
 
 assert(Array.isArray(templateObject.raw), 'The template object is an array');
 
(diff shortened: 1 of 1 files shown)


ljharb:
Which libraries depended on this?

WebReflection:
hyperHTML was born over this standard behaviour and never used a WeakMap to store a template literal. It has 1.7K and now everyone on Chrome 67 and Node 10, since it has a SSR counterpart called viperHTML, will have infinitely growing memory leaks unless they switch to the latest emergency patch releases I've just published instead of enjoying jsconfeu.

hyperHTML has 1.7K star but also lit-html might have used the same pattern , and also every other library based on template literals and their relation to any parsed once content.

As result, once all engines will change this long standing cross engine behaviour, every library that relates template objects to some parsing once action will be slower for no reason due the need to create a dictionary that will trap those templates forever regardless.

TL;DR changing this after 3 years and without warning won't fix the memory leak problem and will curse libraries and users that simply read and trusted ES specifications that shouldn't backfire after 3 years of adoption in every engine (but Firefox, that could just catch up).

ljharb:
Thanks; the PR here is premature prior to any change in https://github.com/tc39/ecma262/pull/890#issuecomment-394095516, however.

WebReflection:
FYI I have actually updated the test to reflect changes after this PR https://github.com/tc39/ecma262/pull/890 so that in the future, **maybe**, all engines will behave the same about template literals objects.

PR #1587, opened 2018-06-05
Update tests for String.prototype.matchAll

peterwmwong:
As per spec changes (https://github.com/tc39/proposal-string-matchall/pull/35), removed tests related to the removed IsRegExp call.

Also fix a misreference in `features` frontmatter.

/cc @ljharb


peterwmwong:
@ljharb There's no rush, but I'm unclear what I should do next to get this merged.  Are there others I need to add to review?

Thanks!

ljharb:
@peterwmwong altho i have the commit bit, I don't think I should be the one to merge this :-) maybe @rwaldron or @leobalter could take a look?

leobalter:
I wonder if we should expect a new behavior for the calls in the tests that are now removed here. In any case, I'd like to preserve the tests matching the same or modified behavior matching the current proposal spec. Consider we have implementations already observing these tests, now Test262 should also tell what's new to consider, right?

ljharb:
@peterwmwong you could keep the test that .calls with an object literal, but assert that its toString is called?

peterwmwong:
@ljharb, are you thinking something like these tests below?

[string-tostring.js](https://github.com/tc39/test262/blob/master/test/built-ins/RegExp/prototype/Symbol.matchAll/string-tostring.js)

```js
// Copyright (C) 2018 Peter Wong. All rights reserved.
/*---
esid: pending
description: String coercion of `string` argument
info: |
  RegExp.prototype [ @@matchAll ] ( string )
    [...]
    3. Return ? MatchAllIterator(R, string).
  MatchAllIterator ( R, O )
    1. Let S be ? ToString(O).
features: [Symbol.matchAll]
includes: [compareArray.js, compareIterator.js, regExpUtils.js]
---*/

var str = 'a*b';
var obj = {
  toString() {
    return str;
  }
};
var regexp = /\w/g;

assert.compareIterator(regexp[Symbol.matchAll](obj), [
  matchValidator(['a'], 0, str),
  matchValidator(['b'], 2, str)
]);
```

[string-tostring-throws.js](https://github.com/tc39/test262/blob/master/test/built-ins/RegExp/prototype/Symbol.matchAll/string-tostring-throws.js)

```js
// Copyright (C) 2018 Jordan Harband. All rights reserved.
/*---
esid: pending
description: String coercion of string parameter
info: |
  RegExp.prototype [ @@matchAll ] ( string )
    [...]
    3. Return ? MatchAllIterator(R, string).
  MatchAllIterator ( R, O )
    1. Let S be ? ToString(O).
features: [Symbol.matchAll]
---*/

var obj = {
  valueOf() {
    $ERROR('This method should not be invoked.');
  },
  toString() {
    throw new Test262Error('toString invoked');
  }
};

  /toString value/[Symbol.matchAll](obj);
});
```

ljharb:
Yup, both of those look great :-)

peterwmwong:
@ljharb @leobalter Huzzah! I have added a new test that will distinguish between implementations that have the spec update vs not: 

https://github.com/peterwmwong/test262/blob/ad6bcdf9d14401bc8278eabc0da0617a9ddf100d/test/built-ins/RegExp/prototype/Symbol.matchAll/isregexp-called-once.js

```js
// Copyright (C) 2018 Peter Wong. All rights reserved.
/*---
esid: pending
description: IsRegExp should only be called once
info: |
  RegExp.prototype [ @@matchAll ] ( string )
    [...]
    3. Return ? MatchAllIterator(R, string).
  MatchAllIterator ( R, O )
    [...]
    2. If ? IsRegExp(R) is true, then
      [...]
    3. Else,
      a. Let flags be "g".
      b. Let matcher be ? RegExpCreate(R, flags).
features: [Symbol.matchAll]
---*/

var internalCount = 0;
Object.defineProperty(RegExp.prototype, Symbol.match, {
  get: function() {
    ++internalCount;
    return true;
  }
});

var count = 0;
var o = {
  get [Symbol.match]() {
    ++count;
    return false;
  }
};

RegExp.prototype[Symbol.matchAll].call(o, '1');

assert.sameValue(0, internalCount);
assert.sameValue(1, count);
```

I verified the test fails without my latest CL to V8: https://chromium-review.googlesource.com/c/v8/v8/+/1086419

I have preemptively updated my commit with this new test.
What do you think?

leobalter:
Thanks for the follow ups, @peterwmwong!

PR #1590, opened 2018-06-07
Add tests for Symbol.prototype.description

joyeecheung:
Refs: https://github.com/tc39/proposal-Symbol-description/issues/5
V8 implementation in https://chromium-review.googlesource.com/c/v8/v8/+/1088871

### DIFF
--- a/test/built-ins/Symbol/prototype/description/descriptor.js
+++ b/test/built-ins/Symbol/prototype/description/descriptor.js
@@ -1,0 +1,19 @@
+// Copyright 2018 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-symbol.prototype.description
+description: >
+    Test the descriptor of Symbol.prototype.description.
+info: |
+    `Symbol.prototype.description` is an accessor property whose
+    set accessor function is undefined.
+features: [Symbol.prototype.description]
+---*/
+
+const desc = Object.getOwnPropertyDescriptor(Symbol.prototype, 'description');
+assert.sameValue(typeof desc.get, 'function');
+assert.sameValue(desc.set, undefined);
+assert.sameValue(desc.writable, undefined);
+assert.sameValue(desc.enumerable, false);
+assert.sameValue(desc.configurable, true);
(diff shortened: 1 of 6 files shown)


joyeecheung:
cc @michaelficarra @gsathya @littledan

test/built-ins/Symbol/prototype/description/descriptor.js
```diff
@@ -0,0 +1,19 @@
...
+info: |
+    `Symbol.prototype.description` is an accessor property whose
+    set accessor function is undefined.
+features: [Symbol.prototype.description]
+---*/
+
+const desc = Object.getOwnPropertyDescriptor(Symbol.prototype, 'description');
+assert.sameValue(typeof desc.get, 'function');
+assert.sameValue(desc.set, undefined);
+assert.sameValue(desc.writable, undefined);
```
ljharb: There’s a propertyHelper that could be used here

test/built-ins/Symbol/prototype/description/get.js
```diff
@@ -0,0 +1,21 @@
...
+    2. Let sym be ? thisSymbolValue(s).
+    3. Return sym.[[Description]].
+features: [Symbol.prototype.description]
+---*/
+
+const symbol = Symbol('test');
+assert.sameValue(symbol.description, 'test');
+assert.sameValue(symbol.hasOwnProperty('description'), false);
+
+const empty = Symbol();
```
ljharb: Could you also test `Symbol(undefined)` and `Symbol(‘’)`?

test/built-ins/Symbol/prototype/description/this-val-non-symbol.js
```diff
@@ -0,0 +1,42 @@
...
+    slot.
+info: |
+    1. Let s be the this value.
+    2. Let sym be ? thisSymbolValue(s).
+    3. Return sym.[[Description]].
+features: [Symbol.prototype.description]
+---*/
+
+assert.throws(TypeError, function() {
+  Symbol.prototype.description.call(null);
```
michaelficarra: This fails, but for the wrong reason. The getter is being called on `Symbol.prototype`. You should instead extract the getter with `Object.getOwnPropertyDescriptor`.

joyeecheung:
@ljharb @michaelficarra Thanks for the reviews. I have updated the PR, PTAL.

joyeecheung:
> LGTM - we may also want to add some tests that .call-ing the getter on real symbols works too?

@ljharb Sure! I've updated the PR, PTAL.

PR #1596, opened 2018-06-14
Add tests for the shape of the Intl.RelativeTimeFormat API.

Ms2ger:


Ms2ger:
Review ping? @rwaldron ?

littledan:
Cc @zbraniecki

harness/testIntl.js
```diff
@@ -771,6 +771,65 @@ function isCanonicalizedStructurallyValidLanguageTag(locale) {
...
+    [{ get length() { throw new CustomError(); } }, CustomError],
+
+    // fail ToLength
+    [{ length: Symbol.toPrimitive }, TypeError],
+    [{ length: { get [Symbol.toPrimitive]() { throw new CustomError(); } } }, CustomError],
+    [{ length: { [Symbol.toPrimitive]() { throw new CustomError(); } } }, CustomError],
+    [{ length: { get valueOf() { throw new CustomError(); } } }, CustomError],
+    [{ length: { valueOf() { throw new CustomError(); } } }, CustomError],
+    [{ length: { get toString() { throw new CustomError(); } } }, CustomError],
+    [{ length: { toString() { throw new CustomError(); } } }, CustomError],
```
leobalter: For custom errors you may also use Test262Error, no need to create a new function. That's optional.

test/intl402/RelativeTimeFormat/constructor/constructor/options-order.js
```diff
@@ -0,0 +1,55 @@
...
+});
+
+assert.compareArray(callOrder, [
+  "localeMatcher",
+  "localeMatcher toString",
+  "style",
+  "style toString",
+  "numeric",
+  "numeric toString",
+]);
```
leobalter: +1

test/intl402/RelativeTimeFormat/constructor/constructor/options-order.js
```diff
@@ -0,0 +1,55 @@
...
+
+/*---
+esid: sec-Intl.RelativeTimeFormat
+description: Checks the order of operations on the options argument to the RelativeTimeFormat constructor.
+info: |
+    InitializeRelativeTimeFormat (relativeTimeFormat, locales, options)
+    7. Let matcher be ? GetOption(options, "localeMatcher", "string", «"lookup", "best fit"», "best fit").
+    14. Let s be ? GetOption(options, "style", "string", «"long", "short", "narrow"», "long").
+    16. Let numeric be ? GetOption(options, "numeric", "string", «"always", "auto"», "always").
+includes: [testIntl.js, compareArray.js]
```
leobalter: is testIntl.js necessary here?

test/intl402/RelativeTimeFormat/constructor/constructor/options-invalid.js
```diff
@@ -0,0 +1,15 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-Intl.RelativeTimeFormat
+description: Checks handling of a null options argument to the RelativeTimeFormat constructor.
+info: |
+    InitializeRelativeTimeFormat (relativeTimeFormat, locales, options)
+    5. Else
+        a. Let options be ? ToObject(options).
+includes: [testIntl.js, compareArray.js]
```
leobalter: I'm not sure you need any of these includes

test/intl402/RelativeTimeFormat/constructor/constructor/locales-invalid.js
```diff
@@ -0,0 +1,16 @@
...
+description: Checks error cases for the locales argument to the RelativeTimeFormat constructor.
+info: |
+    InitializeRelativeTimeFormat (relativeTimeFormat, locales, options)
+    3. Let _requestedLocales_ be ? CanonicalizeLocaleList(_locales_).
+includes: [testIntl.js]
+features: [Intl.RelativeTimeFormat]
+---*/
+
+for (const [locales, expectedError] of getInvalidLocaleArguments()) {
+    assert.throws(expectedError, function() { new Intl.RelativeTimeFormat(locales) })
```
leobalter: I wonder if `getInvalidLocaleArguments` could return a third item to identify as the message. `assert.throws` accepts a third arg for the message. 

In the case of using loops, any failure could be way easier to be found given a description.

Just providing the `locales` won't work as this would throw in the given objects with a symbol inside.

Probably too much of a burden.

test/intl402/RelativeTimeFormat/constructor/constructor/newtarget-undefined.js
```diff
@@ -0,0 +1,21 @@
...
+    1. If NewTarget is undefined, throw a TypeError exception.
+features: [Intl.RelativeTimeFormat]
+---*/
+
+assert.throws(TypeError, function() {
+  Intl.RelativeTimeFormat();
+});
+
+assert.throws(TypeError, function() {
+  Intl.RelativeTimeFormat("en");
```
leobalter: it would be nice adding a `Intl.RelativeTimeFormat("not-valid-tag");` to make sure it still throws a TypeError.

test/intl402/RelativeTimeFormat/prototype/format/branding.js
```diff
@@ -0,0 +1,29 @@
...
+  "",
+  Symbol(),
+  1,
+  {},
+  Intl.RelativeTimeFormat,
+  Intl.RelativeTimeFormat.prototype,
+];
+
+for (const invalidValue of invalidValues) {
+  assert.throws(TypeError, () => fn.call(invalidValue));
```
leobalter: same problem with the messages here. For this case, everything is simple, it would be better to just write static assertions rather than using a loop.

test/intl402/RelativeTimeFormat/prototype/formatToParts/branding.js
```diff
@@ -0,0 +1,29 @@
...
+  "",
+  Symbol(),
+  1,
+  {},
+  Intl.RelativeTimeFormat,
+  Intl.RelativeTimeFormat.prototype,
+];
+
+for (const invalidValue of invalidValues) {
+  assert.throws(TypeError, () => fn.call(invalidValue));
```
leobalter: it would be better to just write static assertions rather than using a loop, adding a message to identify each one.

test/intl402/RelativeTimeFormat/prototype/resolvedOptions/branding.js
```diff
@@ -0,0 +1,29 @@
...
+  "",
+  Symbol(),
+  1,
+  {},
+  Intl.RelativeTimeFormat,
+  Intl.RelativeTimeFormat.prototype,
+];
+
+for (const invalidValue of invalidValues) {
+  assert.throws(TypeError, () => fn.call(invalidValue));
```
leobalter: it would be better to just write static assertions rather than using a loop, adding a message to identify each one.

harness/testIntl.js
rwaldron: I previously allowed a PR that used this `CustomError`, with the expectation that I'd fix it to use `Test262Error` and I didn't get to it, sorry. We can allow this here and then I'll fix them all today.

leobalter: > That's optional.

as I said. It doesn't hurt the tests in anyway. Consistency would be the only achievement from this change.

Ms2ger: The docs say

> Test262Error(message) | constructor for an error object that indicates a test failure

which makes it seem inappropriate for my purpose. If you think it should be used for cases like this, can you update the docs as well?

PR #1599, opened 2018-06-18
Update RegExp tests for CharacterClassEscape \S

leobalter:


leobalter:
@mathiasbynens would you mind to review this one, please?

test/built-ins/RegExp/character-class-escape-non-whitespace-u180e.js
```diff
@@ -0,0 +1,22 @@
...
+    The production CharacterClassEscape :: S evaluates by returning
+    the set of all characters not included in the set returned by
+    CharacterClassEscape :: s
+
+    The Mongolian Vowel Separator (u180e) became a non whitespace character
+    since Unicode 6.3.0
+features: [u180e]
+---*/
+
+var str = String.fromCharCode(0x180E);
```
mathiasbynens: Or `'\u180E'`

test/built-ins/RegExp/character-class-escape-non-whitespace.js
```diff
@@ -0,0 +1,55 @@
...
+  0x2006,
+  0x2007,
+  0x2008,
+  0x2009,
+  0x200A,
+  0x2028,
+  0x2029,
+  0x202F,
+  0x205F,
+  0x3000,
```
mathiasbynens: `0xFEFF` is the only whitespace character missing from this list. If you intentionally want to skip it, add a comment here.

test/built-ins/RegExp/character-class-escape-non-whitespace.js
```diff
@@ -0,0 +1,55 @@
...
+];
+
+for (j = 0x0000; j < 0x10000; j++) {
+  if (j === 0x180E) { continue; } // Skip 0x180E, current test in a separate file
+  if (j === 0xFEFF) { continue; } // Ignore BOM
```
mathiasbynens: Why?

test/built-ins/RegExp/character-class-escape-non-whitespace.js
```diff
@@ -0,0 +1,55 @@
...
+  str = String.fromCharCode(j);
+  res = str.replace(/\S+/g, "test262");
```
mathiasbynens: no need for the `+`

test/built-ins/RegExp/character-class-escape-non-whitespace-u180e.js
```diff
@@ -0,0 +1,22 @@
...
+assert.sameValue(
+  str.replace(/\S+/g, "test262"),
```
mathiasbynens: no need for the `+`

test/built-ins/RegExp/character-class-escape-non-whitespace-u180e.js
mathiasbynens: Or just `'\u180E'`

test/built-ins/RegExp/character-class-escape-non-whitespace.js
```diff
@@ -0,0 +1,55 @@
...
+  0x2029,
+  0x202F,
+  0x205F,
+  0x3000,
+];
+
+for (j = 0x0000; j < 0x10000; j++) {
+  if (j === 0x180E) { continue; } // Skip 0x180E, current test in a separate file
+  if (j === 0xFEFF) { continue; } // Ignore BOM
+  str = String.fromCharCode(j);
```
mathiasbynens: This won't work from U+010000 to U+10FFFF. Use `String.fromCodePoint` instead

test/built-ins/RegExp/character-class-escape-non-whitespace.js
mathiasbynens: If you use `String.fromCodePoint`, then you'll also have to use the `u` flag here. Otherwise, `\S` only matches characters from U+0000 to U+FFFF.

test/built-ins/RegExp/character-class-escape-non-whitespace.js
```diff
@@ -0,0 +1,55 @@
...
+  0x2009,
+  0x200A,
+  0x2028,
+  0x2029,
+  0x202F,
+  0x205F,
+  0x3000,
+];
+
+for (j = 0x0000; j < 0x10000; j++) {
```
mathiasbynens: You want to include U+10000, right? In that case: `j <= 0x10000`

leobalter: For this case, I don't I'm trying to reflect the coverage from the removed file. I'll check coverage for U+10000 to U+10FFFF in the other files

test/built-ins/RegExp/character-class-escape-non-whitespace-u180e.js
leobalter: sure thing!

test/built-ins/RegExp/character-class-escape-non-whitespace-u180e.js
leobalter: gotcha, just an unnecessary flag from the previous test file, where the strings were being stacked before the regexp test

test/built-ins/RegExp/character-class-escape-non-whitespace.js
leobalter: I think it's better to avoid using the `u` flag here and keep the check for the pure `\S`. We can have a separate test for the following range.

test/built-ins/RegExp/character-class-escape-non-whitespace.js
leobalter: legacy or just another case of turtle above the three. Maybe I should not skip this one.

test/built-ins/RegExp/character-class-escape-non-whitespace.js
leobalter: Not intentional, thanks for catching this one up.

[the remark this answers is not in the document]
leobalter: they are described in the Unicode specs and easy to find through many different websites around. I'm not sure I want to favor any of these websites and set a preference on Test262. e.g. this link does not say if a character is considered a whitespace, while other links will say it.

I could also experiment using https://codepoints.net/U+180E but I would need to spend more time reviewing their terms of use.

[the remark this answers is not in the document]
rwaldron: Sorry, I didn't mean that you **must** use the site I suggested, or even a site at all, just _something_ explanatory for each of the whitespace characters.

mathiasbynens:
> If @mathiasbynens approves, then I approve ;)

Huh? I didn't approve, and actually requested some changes...

leobalter:
I’ll work on the fixes and keep you posted, Mathias. If you prefer I can
revert this while I don’t push the follow up.

mathiasbynens:
Thanks, Leo. A follow-up patch would be fine!

anba:
Hmm, I'm not really happy about performing a single `replace` call on each code unit, because it increases the time needed to complete the test significantly when testing in configurations without an optimizing JIT compiler. For example the test now needs almost five times as long when running with `--no-ion` in SpiderMonkey (test duration increased from 70ms to 330ms) and more than ten times as long when running with `--no-ion --no-baseline` (test duration increased from 120ms to 1450ms). And yes, we run test262 in CI with only an interpreter enabled, so this does affect Firefox CI runs. (And it's probably even worse when running opt-debug builds on Android simulator or other slow configurations.)

leobalter:
@anba https://github.com/tc39/test262/pull/1599/files#diff-404cf35ba20e47ecd3d6b31b40262100L17

I used replace before. The tests are taking longer because they are complete - at least in #1611. 

I'll check and see how to reduce the execution time here.

anba:
Added a comment here https://github.com/tc39/test262/pull/1605#issuecomment-401456172 to show one possible approach to reduce the test duration for these types of tests.

PR #1603, opened 2018-06-19
Improve assertion messages, remove non ASCII char

leobalter:


jugglinmike:
Using French quotes was an intentional design decision, so the commit should probably include some detail regarding the motivation

https://github.com/tc39/test262/issues/174

leobalter:
Thanks for the reminder, @jugglinmike. I was responding to someone else's feedback and I forgot this one. We should be fine either way.

jugglinmike:
No problem

PR #1605, opened 2018-06-21
Update tests for RegExp non whitespace class escapes

leobalter:
Apply feedback from #1599

cc @mathiasbynens

PR #1606, opened 2018-06-21
Adds String.fromCodePoint to the features list

leobalter:
Supports the changes from #1605 and fix its CI build


rwaldron:
Easy 👍

PR #1612, opened 2018-06-27
Add a test covering the behavior of SingleLineHTMLCloseComment that uses either Unicode separator

jswalden:
SpiderMonkey tip has a bug in this regard, might as well have everyone be correct about this inanity.

Might be to properly review this you need to hexdump the relevant parts, seeing as GitHub seems to fail to display the relevant separator code points, somewhat unsurprisingly.

### DIFF
--- a/test/annexB/language/comments/single-line-html-close-unicode-separators.js
+++ b/test/annexB/language/comments/single-line-html-close-unicode-separators.js
@@ -1,0 +1,51 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-html-like-comments
+description: SingleLineHTMLCloseComment
+info: |
+    Comment ::
+      MultiLineComment
+      SingleLineComment
+      SingleLineHTMLOpenComment
+      SingleLineHTMLCloseComment
+      SingleLineDelimitedComment
+
+    SingleLineHTMLCloseComment ::
+      LineTerminatorSequenceHTMLCloseComment
+
+    HTMLCloseComment ::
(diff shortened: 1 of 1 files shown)


test/annexB/language/comments/single-line-html-close-unicode-separators.js
```diff
@@ -0,0 +1,51 @@
...
+      SingleLineDelimitedComment
+
+    SingleLineHTMLCloseComment ::
+      LineTerminatorSequenceHTMLCloseComment
+
+    HTMLCloseComment ::
+      WhiteSpaceSequence[opt] SingleLineDelimitedCommentSequence[opt] --> SingleLineCommentChars[opt]
+negative:
+  phase: runtime
+  type: Test262Error
```
leobalter: I don't think we need the negative tags here.

In this case you may remove this tag and replace the `if` at the end with an `assert.sameValue(counter, 2);`

The assertion api will throw a Test262Error for any failure.

jswalden:
The various features you comment on are a consequence of this file starting life as a copy of the sibling `single-line-html-close.js` file.  I don't have opinions on it, except that I mildly trust an approach already taken over one simply posed.

I don't honestly know whether the copyright thing should change; probably safest just leaving it as-is, I would guess.

leobalter:
In an irc convo with @jswalden I just learned this file follows the same format as the sibling [single-line-html-close.js](https://github.com/jswalden/test262/blob/058db04b29af28adacb81df1b21093cec2a8df7a/test/annexB/language/comments/single-line-html-close.js). For the same reason, it matches the respective copyright.

The negative format makes more sense now.

Thanks @jswalden for observing the suggested format and preventing me from removing it.

PR #1616, opened 2018-06-29
Fix multiple test bugs and remove incomplete or no longer tests

anba:


test/language/module-code/instn-star-as-props-dflt-skip.js
```diff
@@ -40,18 +40,19 @@ info: |
...
-import named from './instn-star-props-dflt-skip-star-as-named_FIXTURE.js';
-import production from './instn-star-props-dflt-skip-star-as-prod_FIXTURE.js';
+import {named} from './instn-star-props-dflt-skip-star-as-named_FIXTURE.js';
+import {production} from './instn-star-props-dflt-skip-star-as-prod_FIXTURE.js';
 
 assert('namedOther' in named);
 assert.sameValue(
-  'default' in named, false, 'default specified via identifier'
+  'default' in named, true, 'default specified via identifier'
 );
```
jdalton: ☝️ Where is `default` coming from. The `'./instn-star-props-dflt-skip-star-as-named_FIXTURE.js'` exports `named` which is an `export * as named from "itself"`.

Issue #1566, opened 2018-05-21
Atomics/wake/{count-defaults-to-infinity-missing,wake-all,wake-in-order}.js fail intermittently 


anba:
Mozilla recently updated test262 which resulted in intermittent failures for three `Atomics.wake` tests:
- https://bugzilla.mozilla.org/show_bug.cgi?id=1463110
- https://bugzilla.mozilla.org/show_bug.cgi?id=1462950
- https://bugzilla.mozilla.org/show_bug.cgi?id=1462975
- https://bugzilla.mozilla.org/show_bug.cgi?id=1412233#c14

All three test files have relatively short timeout values in common, for example:

https://github.com/tc39/test262/blob/c440a325a2677be17f2fd1376cf295ea40c1cd9d/test/built-ins/Atomics/wake/count-defaults-to-infinity-missing.js#L65

https://github.com/tc39/test262/blob/c440a325a2677be17f2fd1376cf295ea40c1cd9d/test/built-ins/Atomics/wake/wake-all.js#L49

https://github.com/tc39/test262/blob/c440a325a2677be17f2fd1376cf295ea40c1cd9d/test/built-ins/Atomics/wake/wake-in-order.js#L44

I haven't investigated this in detail, but (too) short timeouts are often problematic in CI environments which have different resource constraints when compared to local PCs. 

Also see #1288.

leobalter:
this should be fixed by #1617

Issue #1530, opened 2018-04-25
Atomics/wait/waiterlist-order-of-operations-is-fifo.js fails intermittently


anba:
Let's assume the scheduler preempts execution between these two lines, then the test case will fail. 

https://github.com/tc39/test262/blob/42ca7bbf9353194105120c2f3863478f1b9a4b14/test/built-ins/Atomics/wait/waiterlist-order-of-operations-is-fifo.js#L35-L36

To make it more likely to trigger a context switch for local testing, you could add `for (var i = 0; i < 10000; ++i) {}` between the two lines.

rwaldron:
I'm going to do another pass on the Atomics tests this week, I'm going to also be removing all uses of `Date.now()` in favor of `$262.agent.monotonicNow()`

leobalter:
@anba can you confirm if the problem persists? I believe it's fixed by #1533

anba:
Yes, it still persists &rarr; https://treeherder.mozilla.org/#/jobs?repo=try&revision=8c0d9506ca4c9b697f1a3048c6e12644c3242b36&selectedJob=185592613

We're currently marking the following tests as having intermittent failures in SpiderMonkey:
- built-ins/Atomics/wait/waiterlist-block-indexedposition-wake.js
- built-ins/Atomics/wait/waiterlist-order-of-operations-is-fifo.js
- built-ins/Atomics/wake/count-defaults-to-infinity-missing.js
- built-ins/Atomics/wake/count-defaults-to-infinity-undefined.js
- built-ins/Atomics/wake/wake-all-on-loc.js
- built-ins/Atomics/wake/wake-all.js
- built-ins/Atomics/wake/wake-in-order.js
- built-ins/Atomics/wake/undefined-index-defaults-to-zero.js

We're still at a8f7012587250b32ffb43a3cbd8da8a9f9d1565e, so when we update to a more recent test262 version additional Atomics test may need to marked as intermittent for us.

leobalter:
Thanks @anba! This info is valuable!

I need to get some updates after 56091a0c5fc0bbc172d039056685bc6992a6528b, I'll see how to check it from here.

leobalter:
this should be fixed by #1617

PR #1617, opened 2018-07-02
Make Atomics tests more resilient against intermittent failures

### LINKED ISSUE #1529: Atomics/wake/undefined-index-defaults-to-zero.js fails intermittently
https://github.com/tc39/test262/blob/master/test/built-ins/Atomics/wake/undefined-index-defaults-to-zero.js fails intermittently when the test system is under pressure, for example I can reliably reproduce the failure when I run the test in 16 tmux panes simultaneously.

anba:
Also fixes multiple issues and bugs in various Atomics tests.

CONTRIBUTING.md and INTERPRETING.md:
- Add missing description for `CanBlockIsFalse` flag and also introduce `CanBlockIsTrue` flag which is needed for some tests (see below).

harness/atomicsHelper.js:
- Updated `$262.agent.waitUntil` to clarify it can also be called with BigInt64Array objects.
- Added `$262.agent.timeouts` constants to unify the various, inconsistently used timeouts in the Atomics tests. Each timeout constant has a brief description and a usage example to clarify when it should or should not be used. The default values were tested in SpiderMonkey (locally and CI) under various system load levels and should hopefully also be valid for other engines. 
- Added `$262.agent.tryYield` to yield control from the main thread. The default implementation simply calls `$262.agent.sleep`, but test262 hosts could theoretically provide a different implementation.
- Added `$262.agent.trySleep` to replace direct calls to `$262.agent.sleep` from the main thread. Motivation for this function: Some test262 hosts (like browsers) may be able to pause and sleep on the main thread, so they could provide their implementation for `trySleep` which performs a busy-wait or something do nothing.

harness/testAtomics.js:
- Replace `let`, `arrow-functions`, and `for-of` loops to avoid creating extremely long `features` lists in tests using this helper. Removed `Map`, `WeakMap`, `Set`, and `WeakSet` (these weren't even listed in features.yaml) for the same reason and added the missing `Symbol` entry for this file to features.yaml.
- Updated all files including "harness/testAtomics.js" to match the new feature requirements.

test/built-ins/Atomics/store/good-views.js and test/built-ins/Atomics/store/bigint/good-views.js:
- Replace `arrow-functions` and `for-of` loops with ES5 alternatives.

test/built-ins/Atomics/wait/bigint/*.js and test/built-ins/Atomics/wake/bigint/*.js
- Some tests were using `BigInt64Array.BYTES_PER_ELEMENT * 8` for the SharedArrayBuffer length, but their non-BigInt counterparts are using `Int32Array.BYTES_PER_ELEMENT * 4`. For consistency and to make it easier to compare the BigInt against the non-BigInt versions, I've changed it to `BigInt64Array.BYTES_PER_ELEMENT * 4`.
- Also aligned formatting and statement placement when they differed between the non-BigInt and the BigInt version of a test file. (I've diffed some of the non-BigInt and BigInt files against each other and different formatting was a nuisance.)

Test files using `$262.agent.monotonicNow()`:
- Moved `$262.agent.report()` calls outside of the block of code measured `$262.agent.monotonicNow()` to avoid measuring how long it takes to execute `$262.agent.report()`. 
- Without this change some tests failed intermittently in certain test configurations in SpiderMonkey. For example with the flags `--ion-eager -- ion-offthread-compile=off` which forces early Ion compilation on the main thread. The `$262.agent.report()` implementation in the SpiderMonkey test262 host embedding uses a for-loop which was forcefully Ion compiled under these settings. And because Ion compilation can take some time, the test case ran longer than `$262.agent.MAX_TIME_EPSILON` which lead to intermittent failures.

Test files using `CanBlockIsFalse` / `CanBlockIsTrue`:
- Some of these tests actually expected that the main thread can wait and [[CanBlock]] is `true` for the agent record executing the test. Therefore I've added a new `CanBlockIsTrue` flag and replaced the flags where needed.

test/built-ins/Atomics/wait/**/*.js and test/built-ins/Atomics/wake/**/*.js:
- Use an atomic counter `RUNNING` in more tests to have better control when a worker agent was actually started.
- Replace the various `$262.agent.sleep(/* Sleep X ms to ensure worker actually sleeps */)` calls with the new `$262.agent.tryYield()` function. This `X` was sometimes as low as 10 milliseconds, which is definitely too short for CI systems under heavy load (observed by intermittent CI failures for SpiderMonkey) and sometimes as high as 500 milliseconds, which is probably much longer than needed even when the system is under heavy load. 
- Removed duplicate strings in assertion messages, presumably from copy-pasting the messages between different files.
- Removed extra empty lines at the end of multiple files.

test/built-ins/Atomics/wake/bad-range.js, test/built-ins/Atomics/wake/bigint/bad-range.js, and test/built-ins/Atomics/wait/bigint/non-bigint64-typedarray-throws.js:
- Removed unnecessary `features` and `includes` from this file.

test/built-ins/Atomics/wait/waiterlist-order-of-operations-is-fifo.js and test/built-ins/Atomics/wait/bigint/waiterlist-order-of-operations-is-fifo.js:
- The test was actually broken and didn't test what it said it does. This probably explains #1530.
- The test wants to ensure the waiterlist is implemented as a FIFO structure. This requires that the waiting agents all wait on the same index position, because the waiterlists are defined by each index. But if the agents wait on different indices, each agent is inserted into a different FIFO structure and therefore we can't observe any FIFO ordering between the agents when they're woken up.
- All this requires a bit of synchronization between the main agent and the waiting agent, I hope the added comments help to review these changes.

test/built-ins/Atomics/wait/good-views.js:
- The agent sends multiple reports to the main agent, but only the first one was read.
- This error was introduced during previous refactorings. I've changed it to back to use a while-loop as used in the first version of this file.

test/built-ins/Atomics/wait/no-spurious-wakeup-no-operation.js:
- Only the BigInt version of this test was present, copied it so we also get code coverage for the non-BigInt case.

test/built-ins/Atomics/wait/waiterlist-block-indexedposition-wake.js:
- Added extra while loops to avoid intermittent failures when the agent worker haven't started to wait.
- This should help to avoid some of the intermittent failures we saw for SpiderMonkey.

test/built-ins/Atomics/wake/wake-all-on-loc.js and test/built-ins/Atomics/wake/bigint/wake-all-on-loc.js:
- This test was also no longer after previous refactoring sessions.
- The "B" agent only waited for 10 milliseconds, which made it likely that it already timed out before the main agent was able to call `Atomics.wake`, which in turn rendered the test useless, because the test case wants to ensure that `Atomics.wake` cannot wake "B". But if "B" was already timed out, it can trivially not be woken by `Atomics.wake`.
- Added some safety measure to catch the case when "B" timed out before `Atomics.wake` was called and made it a test error if that happens.

test/built-ins/Atomics/wake/count-defaults-to-infinity-missing.js, test/built-ins/Atomics/wake/count-defaults-to-infinity-undefined.js, and test/built-ins/Atomics/wake/undefined-index-defaults-to-zero.js:
- Changed the `$262.agent.start()` calls to use a for-loop to avoid code duplication.
- (Forgot to undo the code formatting around `assert.sameValue`, still need to change it back.)
- Also more while-loops around `Atomics.wake`.
- These changes should fix #1529 and #1566.

test/built-ins/Atomics/wake/wake-all.js:
- Removed "B" worker agent.
- Without this removal the test case would be exactly equal to test/built-ins/Atomics/wake/wake-all-on-loc.js.

test/built-ins/Atomics/wake/wake-in-order-one-time.js:
- Add for-loops to avoid code duplication and make the test more readable.
- Make the `Atomics.wake` assertion messages unique by adding the current loop counter.
- Add `$262.agent.tryYield()` to give the worker agents enough time to actually start waiting.

test/built-ins/Atomics/wake/wake-in-order.js:
- Removed the outer loop `attempt < 10` because it uses `$262.agent` in a way currently not required to work. And which actually also doesn't work in SpiderMonkey's implementation of `$262.agent`.
- According to INTERPRETING.md `$262.agent.broadcast()` broadcasts its message to all agents ever started by `$262.agent.start()` and then blocks until all agents received the broadcast. It is not required that the agents started by the first `$262.agent.broadcast()` call will all be disabled/destroyed/whatever when the second `$262.agent.broadcast()` call occurs, which then means the second `$262.agent.broadcast()` call still tries to reach the agents started in the first loop iteration, but these may no longer accept broadcasts and therefore won't acknowledge they've received the broadcast. Which then means the second `$262.agent.broadcast()` call waits forever.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -157,6 +157,8 @@ This tag is for boolean properties associated with the test.
   created procedurally using the project's test generation tool; refer to the
   section titled "Procedurally-generated tests" for more information on this
   process
+- **`CanBlockIsFalse`** - only run the test when the [[CanBlock]] property of the [Agent Record](https://tc39.github.io/ecma262/#sec-agents) executing the test file is `false`
+- **`CanBlockIsTrue`** - only run the test when the [[CanBlock]] property of the [Agent Record](https://tc39.github.io/ecma262/#sec-agents) executing the test file is `true`
 
 #### features
 **features**: [list]
@@ -170,7 +172,7 @@ Some tests require the use of language features that are not directly described 
 
 This tag identifies the section number from the portion of the ECMAScript 5.1 or ECMAScript 3 standard that is tested by this test. It was automatically generated for tests that were originally written for the ES5 (or earlier) version of the test suite and are now part of the ES6 version. You can use the es5id to discover the relevant portion of the ECMAScript standard by looking up the section number in [previous publications of the specification](https://www.ecma-international.org/publications/standards/Ecma-262-arch.htm). Unfortunately, there is no way to identify which version of ECMAScript (specifically, 3 or 5.1) without looking up the section number and deciding whether it covers the observable in the test.
 
-Read the (Test262 Technical Rationale Report)[https://github.com/tc39/test262/wiki/Test262-Technical-Rationale-Report,-October-2017#specification-reference-ids] for reasoning behind deprecation.
+Read the [Test262 Technical Rationale Report](https://github.com/tc39/test262/wiki/Test262-Technical-Rationale-Report,-October-2017#specification-reference-ids) for reasoning behind deprecation.
 
 #### es6id
 **es6id**: [es6-test-id]
@@ -179,7 +181,7 @@ Read the (Test262 Technical Rationale Report)[https://github.com/tc39/test262/wi
 
 This tag identifies the section number from the portion of the ES6 standard that is tested by this test _at the time the test was written_. The es6ids might not correspond to the correction section numbers in the ES6 (or later) specification because routine edits to the specification will change section numbering. For this reason, only the esid is required for new tests.
 
-Read the (Test262 Technical Rationale Report)[https://github.com/tc39/test262/wiki/Test262-Technical-Rationale-Report,-October-2017#specification-reference-ids] for reasoning behind deprecation.
+Read the [Test262 Technical Rationale Report](https://github.com/tc39/test262/wiki/Test262-Technical-Rationale-Report,-October-2017#specification-reference-ids) for reasoning behind deprecation.
 
 ## Test Environment
 
--- a/INTERPRETING.md
+++ b/INTERPRETING.md
@@ -354,6 +354,30 @@ following strings:
   test generation tool. This flag is specified for informational purposes only
   and has no bearing on how the test should be interpreted.
 
+- **`CanBlockIsFalse`** The test file should only be run when the [[CanBlock]] property of the [Agent Record](https://tc39.github.io/ecma262/#sec-agents) executing the file is `false`.
+
+ *Example*
+
+  ```js
+  /*---
+  flags: [CanBlockIsFalse]
+  ---*/
+  var i32 = new Int32Array(new SharedArrayBuffer(Int32Array.BYTES_PER_ELEMENT));
+  assert.throws(TypeError, function() { Atomics.wait(i32, 0, 0, 1000); });
+  ```
+
+- **`CanBlockIsTrue`** The test file should only be run when the [[CanBlock]] property of the [Agent Record](https://tc39.github.io/ecma262/#sec-agents) executing the file is `true`.
+
+ *Example*
+
+  ```js
+  /*---
+  flags: [CanBlockIsTrue]
+  ---*/
+  var i32 = new Int32Array(new SharedArrayBuffer(Int32Array.BYTES_PER_ELEMENT));
+  Atomics.wait(i32, 0, 0, 1000); // Sleep for one second.
+  ```
+
 ### `locale`
 
 The `locale` attribute allows tests to declare explicit information regarding locale specificity. Its value is an array of one or more valid language tags or subtags.
--- a/harness/atomicsHelper.js
+++ b/harness/atomicsHelper.js
@@ -36,19 +36,204 @@ $262.agent.MAX_TIME_EPSILON = 100;
   };
 }
 /**
- * With a given Int32Array, wait until the expected number of agents have reported themselves by
- * calling:
+ * With a given Int32Array or BigInt64Array, wait until the expected number of agents have
+ * reported themselves by calling:
  *
- *    Atomics.add(i32a, index, 1);
+ *    Atomics.add(typedArray, index, 1);
  *
- * @param {Int32Array} i32a An Int32Array with a SharedArrayBuffer
- * @param {Number} index    The index of which all agents will report.
- * @param {Number} expected The number of agents that are expected to report as active.
+ * @param {(Int32Array|BigInt64Array)} typedArray An Int32Array or BigInt64Array with a SharedArrayBuffer
+ * @param {number} index    The index of which all agents will report.
+ * @param {number} expected The number of agents that are expected to report as active.
  */
-$262.agent.waitUntil = function(i32a, index, expected) {
-  let agents = 0;
-  while ((agents = Atomics.load(i32a, index)) !== expected) {
+$262.agent.waitUntil = function(typedArray, index, expected) {
+  var agents = 0;
+  while ((agents = Atomics.load(typedArray, index)) !== expected) {
     /* nothing */
   }
-  assert.sameValue(agents, expected, `Reporting number of 'agents' equals the value of 'expected' (${expected})`);
+  assert.sameValue(agents, expected, "Reporting number of 'agents' equals the value of 'expected'");
+};
+
+/**
+ * Timeout values used throughout the Atomics tests. All timeouts are specified in milliseconds.
+ *
+ * @property {number} yield Used for `$262.agent.tryYield`. Must not be used in other functions.
+ * @property {number} small Used when agents will always timeout and `Atomics.wake` is not part
+ *                          of the test semantics. Must be larger than `$262.agent.timeouts.yield`.
+ * @property {number} long  Used when some agents may timeout and `Atomics.wake` is called on some
+ *                          agents. The agents are required to wait and this needs to be observable
+ *                          by the main thread.
+ * @property {number} huge  Used when `Atomics.wake` is called on all waiting agents. The waiting
+ *                          must not timeout. The agents are required to wait and this needs to be
+ *                          observable by the main thread. All waiting agents must be woken by the
+ *                          main thread.
+ *
+ * Usage for `$262.agent.timeouts.small`:
+ *   const WAIT_INDEX = 0;
+ *   const RUNNING = 1;
+ *   const TIMEOUT = $262.agent.timeouts.small;
+ *   const i32a = new Int32Array(new SharedArrayBuffer(Int32Array.BYTES_PER_ELEMENT * 2));
+ *
+ *   $262.agent.start(`
+ *     $262.agent.receiveBroadcast(function(sab) {
+ *       const i32a = new Int32Array(sab);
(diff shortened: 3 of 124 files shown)


CONTRIBUTING.md
```diff
@@ -157,6 +157,8 @@ This tag is for boolean properties associated with the test.
   created procedurally using the project's test generation tool; refer to the
   section titled "Procedurally-generated tests" for more information on this
   process
+- **`CanBlockIsFalse`** - only run the test when the [[CanBlock]] property of the [Agent Record](https://tc39.github.io/ecma262/#sec-agents) executing the test file is `false`
+- **`CanBlockIsTrue`** - only run the test when the [[CanBlock]] property of the [Agent Record](https://tc39.github.io/ecma262/#sec-agents) executing the test file is `true`
```
leobalter: Thanks for this. While this is correct for the current model, I believe it would be easier to move this flag to the API on $262.agent. Perhaps, this improves communicating with the engine running the test. While the flag would add more separation in the test runners.

What do you think? We can have it fixed here or in a follow up, if you're positive with this new idea. Otherwise, the current changes here are the best.

harness/atomicsHelper.js
```diff
@@ -36,19 +36,204 @@ $262.agent.MAX_TIME_EPSILON = 100;
...
+ *   //   wokenCount += woken;
+ *   // }
+ *
+ *   // All agents were woken and none timeout.
+ *   for (var i = 0; i < NUMAGENT; i++) {
+ *     assert($262.agent.getReport(), "ok");
+ *   }
+ */
+$262.agent.timeouts = {
+  yield: 100,
```
leobalter: if this is not meant to be used outside of tryYield, I wonder if it should be exposed this way.

In any case, a change here is not worth blocking this PR.

harness/testAtomics.js
```diff
@@ -15,16 +15,17 @@ description: |
...
-    (view) => Number.NEGATIVE_INFINITY,
-    (view) => ({ valueOf: () => 125 }),
-    (view) => ({ toString: () => '125', valueOf: false }) // non-callable valueOf triggers invocation of toString
+    function(view) { return -1; },
+    function(view) { return view.length; },
+    function(view) { return view.length * 2; },
+    function(view) { return Number.POSITIVE_INFINITY; },
+    function(view) { return Number.NEGATIVE_INFINITY; },
+    function(view) { return { valueOf: function() { return 125; } }; },
+    function(view) { return { toString: function() { return '125'; }, valueOf: false }; }, // non-callable valueOf triggers invocation of toString
```
leobalter: thanks!

CONTRIBUTING.md
anba: I guess it depends on the test runner. For example for SpiderMonkey/Firefox, a separate flag is easy to implement and allows us to reuse existing machinery to differentiate between shell-only ([[CanBlock]] is true) and browser-only ([[CanBlock]] is false) tests. That being said, we don't yet run Atomics tests in the browser, because our `$test262.agent` implementation uses shell-only features. Therefore it's a bit difficult to say which approach works best (flags vs. `$262.agent` API), at least from the perspective of SpiderMonkey/Firefox.

harness/atomicsHelper.js
anba: I was wondering if some test262 consumers may want to have the ability to redefine `$262.agent.timeouts`, so it better fits their needs. Therefore I made it an additional property on the `timeouts` object. (For example for local testing, it could useful to decrease the various timeout values to make the tests complete faster.)

anba:
Thanks for merging! :smile:

harness/atomicsHelper.js
leobalter: sure! It also faster to just tweak the timeouts object instead of rewriting the whole tryYield function.

CONTRIBUTING.md
leobalter: The flag is already set, so I'm thinking on trying that with runners anyway. We can reconsider using the agent later if we find any arguments.

INTERPRETING.md
```diff
@@ -354,6 +354,30 @@ following strings:
...
+ *Example*
+
+  ```js
+  /*---
+  flags: [CanBlockIsTrue]
+  ---*/
+  var i32 = new Int32Array(new SharedArrayBuffer(Int32Array.BYTES_PER_ELEMENT));
+  Atomics.wait(i32, 0, 0, 1000); // Sleep for one second.
+  ```
+
```
rwaldron: This is great, thank you.

harness/atomicsHelper.js
rwaldron: This works for me

test/built-ins/Atomics/wait/bigint/false-for-timeout-agent.js
```diff
@@ -31,21 +33,31 @@ $262.agent.start(`
...
-  new SharedArrayBuffer(BigInt64Array.BYTES_PER_ELEMENT * 8)
+  new SharedArrayBuffer(BigInt64Array.BYTES_PER_ELEMENT * 4)
 );
 
 $262.agent.broadcast(i64a.buffer);
-$262.agent.sleep(100);
+$262.agent.waitUntil(i64a, RUNNING, 1n);
+
+// Try to yield control to ensure the agent actually started to wait.
+$262.agent.tryYield();
```
rwaldron: @anba I had worked to eliminate as much reliance on "sleep" as possible, by making codifying `waitUntil` and implementing it wherever it was necessary to wait for all agents to "report" that they are moving into a waiting state. `tryYield` is just reintroducing the "sleep" _in addition_ to the "wait until" mechanism. Can you help me understand why that extra step is necessary? If I had to guess, I _think_ that you're sleeping the main thread in order to allow the agents' execution to reach the end of calling `Atomics.wait(...)`? If that's true, then `262.agent.tryYield();` serves as essentially 1/2 of what `waitUntil` is meant to provide and I think that the implementation of `$262.agent.waitUntil` should itself call `$262.agent.tryYield();`, instead of putting the burden of remembering to _also_ call `$262.agent.tryYield();` on test authors (and future maintainers). 

If you agree with me, I will do the work of updating the test material. WDYT?

anba: Yes, `tryYield` is necessary for the case when the agent thread was interrupted between its call to `Atomics.add(i32a, ${RUNNING}, 1);` and the call to `Atomics.wait(i32a, 0, ...)`. This case happens all the time when the system is under heavy load, so not performing `tryYield` is sadly not an option. (We also want to have `waitUntil`, because that allows us to reduce the sleep time in `tryYield`.) 

But if a specific test262 host has better a mean to implement `tryYield` than just sleeping for some time, I'd currently recommend that this host provides their own `tryYield` implementation. (I haven't yet investigated what is necessary to enable the Atomics tests in a browser environment, but the patch in https://bugzilla.mozilla.org/show_bug.cgi?id=1349863 gives the impression that quite a few changes for `$test262.agent` are necessary. That's why I don't yet want to make `tryYield` a part of the requirement host binding for `$test262.agent`, because browsers may need a complete different approach than `tryYield`.)

Most, but not all tests, call `tryYield` directly after `waitUntil`, therefore I haven't moved `tryYield` into `waitUntil`. For example "waiterlist-block-indexedposition-wake.js" (`waitUntil` not followed by `tryYield`) or "waiterlist-order-of-operations-is-fifo.js" (`tryYield` not directly preceded by `waitUntil`).

And I also agree that reducing direct calls to `$test262.agent.sleep` in test cases is a useful goal. In fact my PR actually removed all direct calls to `$test262.agent.sleep`, so the tests are now only calling it indirectly. :smile:

PR #1618, opened 2018-07-02
added support for jsc curation log files

nomadtechie:


implementation-contributed/curation_logs/javascriptcore.json
```diff
@@ -0,0 +1,7 @@
+{
+	"targetRevisionAtLastExport": "master",
```
leobalter: this looks correct.

PR #1627, opened 2018-07-06
Add tests for objects created in the RelativeTimeFormat constructor.

Ms2ger:
The specification was changed in https://github.com/tc39/proposal-intl-relative-time/pull/79.


test/intl402/RelativeTimeFormat/constructor/constructor/options-proto.js
```diff
@@ -0,0 +1,88 @@
...
+    },
+  },
+
+  // PluralRules
+  "type": {
+    "get": function() {
+      throw new Test262Error("Should not call getter on Object.prototype: type");
+    },
+  },
+});
```
rwaldron: In the future, none of these keys needs to be quoted like this.

PR #1628, opened 2018-07-06
Update the test for negative zero in NumberFormat.

Ms2ger:
The specification was updated in https://github.com/tc39/ecma402/pull/232.


iamstolis:
This PR updated `format-negative-numbers.js` to ensure that -0 is formatted as negative number but it did not update `format-significant-digits.js` (namely https://github.com/tc39/test262/blob/master/test/intl402/NumberFormat/prototype/format/format-significant-digits.js#L26) in the same way. Is this an oversight or I am missing something?

Ms2ger:
Oversight, I'll fix at some point if nobody beats me to it.

Ms2ger:
#1653

PR #1630, opened 2018-07-08
Refactor string literal tests for parsers

jugglinmike:


rwaldron:
This is excellent. Thank you for the detailed commit messages

PR #1631, opened 2018-07-08
Fix the ranges of non-HexDigits in decodeURI and decodeURIComponent tests

sainaen:
The ranges `0x2A`–`0x2F` (characters `* + , - . /`) and `0x3A`–`0x3F` (`: ; < = > ?`) were missing from the non-HexDigit intervals in tests for decoding of illegal percent-encoded sequences.

(I forgot that I've already singed CLA and did it again. Sorry, if such a “renewal” causes any problems! :sweat_smile:)


rwaldron:
@leobalter @mathiasbynens can either of you take a quick look at this?

leobalter:
I have no quick answer rather than stopping and re-checking the specs. I can take a look at this next week unless @mathiasbynens comes with positive feedback.

rwaldron:
Thanks anyone!

PR #1636, opened 2018-07-12
Intl.RelativeTimeFormat: Add tests for invalid units in format/formatToParts.

Ms2ger:


test/intl402/RelativeTimeFormat/prototype/format/unit-invalid.js
```diff
@@ -0,0 +1,43 @@
...
+    SingularRelativeTimeUnit ( unit )
+
+    10. If unit is not one of "second", "minute", "hour", "day", "week", "month", "quarter", "year", throw a RangeError exception.
+
+features: [Intl.RelativeTimeFormat]
+---*/
+
+const rtf = new Intl.RelativeTimeFormat("en-US");
+
+assert.sameValue(typeof rtf.format, "function");
```
rwaldron: This is irrelevant to the semantics of `SingularRelativeTimeUnit(...)`. There is already a test file ([prop-desc.js](https://github.com/tc39/test262/blob/master/test/intl402/RelativeTimeFormat/prototype/format/prop-desc.js)) that verifies `Intl.RelativeTimeFormat.prototype.format` is function.

test/intl402/RelativeTimeFormat/prototype/formatToParts/unit-invalid.js
```diff
@@ -0,0 +1,43 @@
...
+    SingularRelativeTimeUnit ( unit )
+
+    10. If unit is not one of "second", "minute", "hour", "day", "week", "month", "quarter", "year", throw a RangeError exception.
+
+features: [Intl.RelativeTimeFormat]
+---*/
+
+const rtf = new Intl.RelativeTimeFormat("en-US");
+
+assert.sameValue(typeof rtf.formatToParts, "function");
```
rwaldron: Same as above, but different existing file: [prop-desc.js](https://github.com/tc39/test262/blob/master/test/intl402/RelativeTimeFormat/prototype/formatToParts/prop-desc.js)

Ms2ger: I added these because I kept forgetting that SpiderMonkey doesn't implement formatToParts, so I was wasting time trying to figure out why calling undefined threw a TypeError. I can remove if you prefer.

PR #1642, opened 2018-07-18
arguments / eval with prefix / postfix in strict code should raise early ReferenceError instead of SyntaxError

Constellation:
According to the latest spec[1,2], we should throw early ReferenceError if IsValidSimpleAssignmentTarget for `expr++`'s and `++expr`'s `expr` is false.
However, our tests expect SyntaxError now. This patch fixes it to align these tests to the latest spec.

1: https://tc39.github.io/ecma262/#sec-update-expressions-static-semantics-early-errors
2: https://tc39.github.io/ecma262/#sec-identifiers-static-semantics-isvalidsimpleassignmenttarget

### DIFF
--- a/test/language/expressions/postfix-decrement/11.3.2-2-1-s.js
+++ b/test/language/expressions/postfix-decrement/11.3.2-2-1-s.js
@@ -4,14 +4,14 @@
 /*---
 es5id: 11.3.2-2-1-s
 description: >
-    Strict Mode - SyntaxError is thrown if the identifier 'arguments'
+    Strict Mode - ReferenceError is thrown if the identifier 'arguments'
     appear as a PostfixExpression(arguments--)
 flags: [onlyStrict]
 ---*/
 
 function testcase() {
         var blah = arguments;
-        assert.throws(SyntaxError, function() {
+        assert.throws(ReferenceError, function() {
             eval("arguments--;");
         });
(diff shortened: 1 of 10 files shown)


Constellation:
See also: https://github.com/tc39/ecma262/issues/1267

IgnoredAmbience:
A conflicting PR is open at tc39/ecma262#1132 to fix the spec bug.

rwaldron:
The referenced PR has been merged. I can rebase this and merge it.

rwaldron:
> This PR can simply be closed, because the spec now requires to throw a SyntaxError - which is exactly what the current tests already checked for.

There are two outcomes: 

- Non Strict: Throw early Reference Error
- Strict Mode: Throw early Syntax Error


The existing tests that are called "non-simple.js" don't reflect the semantics as they are currently defined, ie. "It is an early Reference Error if IsValidSimpleAssignmentTarget of LeftHandSideExpression is false." is now "It is an early Reference Error if AssignmentTargetType of LeftHandSideExpression is invalid.". "IsValidSimpleAssignmentTarget" no longer exists. They also refer to grammar productions that no longer exist (all four directories could use some updating, but I don't have time to do that much work). 

I've created a set of changes that reflect all of the new specification language changes that are relevant to these tests.

PR #1643, opened 2018-07-19
Add a missing Reflect feature flag

IgnoredAmbience:
`built-ins/Proxy/deleteProperty/boolean-trap-result-boolean-true.js` was missing a feature flag for the Reflect library.

### DIFF
--- a/test/built-ins/Proxy/deleteProperty/boolean-trap-result-boolean-true.js
+++ b/test/built-ins/Proxy/deleteProperty/boolean-trap-result-boolean-true.js
@@ -6,6 +6,7 @@ description: >
     [[Delete]] (P)
 
     The result is a Boolean value.
+features: [Reflect]
 ---*/
 
 var p = new Proxy({}, {


test/built-ins/Proxy/deleteProperty/boolean-trap-result-boolean-true.js
```diff
@@ -6,6 +6,7 @@ description: >
     [[Delete]] (P)
 
     The result is a Boolean value.
+features: [Reflect]
```
ljharb: also Proxy, no?

IgnoredAmbience: Ah yes, good point. It didn't trip my test skips as I was testing Proxies :)

IgnoredAmbience: That said, no other Proxy test has Proxy in the features array, has it been assumed to be inferred from being in the Proxy directory?

rwaldron: > That said, no other Proxy test has Proxy in the features array, 

That's only because "features" was introduced _after_ most of those tests were written and technical debt is a real thing ;)

Anyway, I'd love it if you added these, but you don't have to.

PR #1644, opened 2018-07-19
Make Proxy/ownKeys tests ES5-parsable

IgnoredAmbience:
All other Proxy tests use the syntax `attr: function() {...}` for defining traps, ownKeys was unique in using the shorthand syntax `attr() {..}`. Change to longhand syntax for back-compat for partial implementations.

### DIFF
--- a/test/built-ins/Proxy/ownKeys/return-duplicate-entries-throws.js
+++ b/test/built-ins/Proxy/ownKeys/return-duplicate-entries-throws.js
@@ -12,7 +12,7 @@ info: |
 ---*/
 
 var p = new Proxy({}, {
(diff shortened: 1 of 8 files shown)


ljharb:
An engine that has implemented Proxy, but not the rest of ES6, shouldn't fail the Proxy tests.

leobalter:
I'm merging for the consistency with the other tests. I also would like to - slowly - start deprecating the feature tags for ES6, but that's a topic for another thread...

rwaldron:
>  I also would like to - slowly - start deprecating the feature tags for ES6

I'm strongly opposed to this. We can discuss this further at some later point.

PR #1645, opened 2018-07-20
Added a legal part types test

phistuck:
Chrome uses dayperiod while the specification says dayPeriod. :(
https://crbug.com/865351
This should prevent such mistakes in the future.

I am surprised such a basic test was not added until now... Is there a reason, or am I simply bad at searching?


phistuck:
Note - if it is not accepted due to Contributor License Agreement issues, anyone may feel free to take the code without any credit and submit a pull request themselves.

Issue #1343, opened 2017-11-10
Add coverage for Private methods and accessors


leobalter:
Already on Stage 3

littledan:
I'm working on a test plan with @mkubilayk @Neuroboy23 @jbhoosreddy @tim-mc @cameron-mcateer @molisani @rricard from Bloomberg ; we plan to collaborate on the private methods/accessors tests.

littledan:
A test plan--if you have other ideas, please post them further down on the thread:

* Early errors
   * Grammar
       * No space permitted between # and name (either in definition or usage)
       * Space permitted between receiver and private method access
       * "shorthand" not permitted -- a receiver is required
       * Private methods not permitted in object literals
       * Super private access (`super.#x()`) is not permitted
       * Private methods may not be deleted
   * Referencing a missing private name, including
       * Cases where the name is defined in a more inner context, as allowed in Java
       * In a subclass (this is private, not protected)
       * Indirect eval doesn't see any private names at all
       * Private names are not visible inside of the extends clause
   * Duplicate private names
       * Multiple private methods inside of a class
       * A private method and a getter/setter
       * Multiple get or multiple set
       * A private method and a private field conflicting
       * Static and instance declarations with the same name conflicts
       * Shadowing in a nested class is OK
* Semantics
    * Private method constructs
        * Private ordinary methods
        * Private accessors
        * Private generators, async functions, async generators
        * static variants of all three of the above
    * Usage
        * Calling a private method
        * Reading a private method separately from calling it
            * All instances of the class share the same function identity for the private method
        * Private methods can do super property access of public methods/properties, and this access uses the appropriate home object
        * The `.name` property of the private method is `"#foo"`
        * The `.toString()` of the private method is the exact string of its declaration (depends on the `Function.prototype.toString` Stage 3 feature flag as well) -- include comments and various declaration forms in the test
        * It's OK at parse-time to reference a static private method from an instance method, or vice versa, and you can make use of this at runtime with creative application of `Function.prototype.call`.
    * Complex contexts where private names are visible (use an assignment as the left operand of a comma expression as an easy mechanism to "leak" a function to access the private name)
        * Direct eval sees the same set of private names as the outer scope
        * Inside of computed property names
        * Inside of a nested function, block, class, arrow function
    * Error on missing method
        * The type check is done on reading the method, not on calling it--potentially, the method could be called with a different receiver
        * Execution order -- private methods are added before any field initializer is run, even if they appear textually later
[...]

PR #1651, opened 2018-07-26
[WIP] Early Errors > Grammar #1343

jbhoosreddy:
### Summary of changes:
- [ ] No space permitted between # and name (either in definition or usage) 🏃❓ 
- [ ] Space permitted between receiver and private method access
- [ ] "shorthand" not permitted -- a receiver is required
- [ ] Private methods not permitted in object literals
- [ ] Super private access (super.#x()) is not permitted
- [ ] Private methods may not be deleted

#### Legend
- 🏃 means it's in progress
- ❓ needinfo

### DIFF
--- a/src/class-fields/class-fields-early-errors/cls-decl.template
+++ b/src/class-fields/class-fields-early-errors/cls-decl.template
@@ -1,0 +1,16 @@
+// Copyright (C) 2018 Jaideep Bhoosreddy. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+path: language/statements/class/class-fields-early-errors-
+name: early errors in class declaration
+features: [class, class-fields-private]
+esid: sec-static-semantics-early-errors
+---*/
+
+throw "Test262: This statement should not be evaluated.";
+
+class C {
+  /*{ fields }*/
+  /*{ methods }*/
+}
--- a/src/class-fields/class-fields-early-errors/cls-expr.template
+++ b/src/class-fields/class-fields-early-errors/cls-expr.template
@@ -1,0 +1,16 @@
+// Copyright (C) 2018 Jaideep Bhoosreddy. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+path: language/expressions/class/class-fields-early-errors
+name: early errors in class declaration
+features: [class, class-fields-private]
+esid: sec-static-semantics-early-errors
+---*/
+
+throw "Test262: This statement should not be evaluated.";
+
+var C = class {
+  /*{ fields }*/
+  /*{ methods }*/
+}
--- a/src/class-fields/private-names-no-whitespace-after-sigil.case
+++ b/src/class-fields/private-names-no-whitespace-after-sigil.case
@@ -1,0 +1,19 @@
+// Copyright (C) 2018 Jaideep Bhoosreddy. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: Syntax error if whitespace after '#' sigil
+template: class-fields-early-errors
+features: [class-fields-private]
+negative:
+  phase: parse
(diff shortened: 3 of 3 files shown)


jbhoosreddy:
Question re/ early errors: When I try to write test cases for early errors, I run into babel parse errors which bail test262-harness. How can this be fixed? cc @littledan or others maintainers

mkubilayk:
I'm also working on a set of early errors for private methods and hitting the same issue as @jbhoosreddy with the babel parser.

mkubilayk:
I suspect test harness uses a version of the parser that doesn't support private methods syntax properly. 

```js
class C {
    #foo() {}
}
```

I was able to parse the code above using `@babel/parser@7.0.0-beta.55`. It didn't work with `babylon@6.18.0`, which is the version installed for `test262-harness`.

```js
const code = `
class C {
    #foo() {}
}
`;

// parses
const ast = require('@babel/parser').parse(code, {
    sourceType: 'module',
    plugins: [
        'classPrivateMethods'
    ]
});

// throws
const babylonAst = require('babylon').parse(code, {
    sourceType: 'module',
    plugins: [
        'classPrivateMethods'
    ]
});
```

It looks like `classPrivateMethods` is not a plugin available in [here](https://github.com/babel/babylon/blob/6.x/README.md#plugins).

rwaldron:
The Babel option test262-harness is basically the worst feature ever added to that tool and shouldn't be used at all.

rwaldron:
Also, we have tons of Class private field tests that need to be restored from git history.

jbhoosreddy:
@rwaldron does test262 not have a way to skip tests? I'm curious why some tests need to be restored from git history? Maybe @mkubilayk and I can PR to add that feature also.

littledan:
Great to see more progress on class field tests both in this PR as well as in https://github.com/tc39/test262/pull/1660 . I guess once #1660 lands, we can rebase these fixes on top if there is any more work to do.

I agree with the statements in the checkboxes at the top of this PR; could you explain more about what info you need about the first item, @jbhoosreddy ?

For tests against early errors with Babel, it's probably better to use @jugglinmike 's direct integration of Babylon with test262 at https://github.com/babel/babylon/pull/654 . Maybe this will be able to test the private methods grammar even if the transforms aren't ready yet (see https://github.com/babel/proposals/issues/22 -- the transform is listed as "in progress", even for Babel 7).

Thanks for your fixes to test262-harness's --babelPresets flag, @leobalter and @rwaldron . It definitely wasn't perfect on the first try, and has room for improvement still. I recommended that @jbhoosreddy develop tests for private methods using this mode, but imagined that improving the infrastructure would be part of the project as well. If this recommendation was a poor one, let's figure out the right development mode is. I saw that you marked this flag as deprecated, @rwaldron . I'm wondering, what would you recommend for running test262 tests against Babel transforms?

rwaldron:
> I'm curious why some tests need to be restored from git history? 

Because they were removed when the spec was split and part of the features were demoted to State 2.

littledan:
> I'd recommend using Babel's test runner for testing Babel transforms: https://github.com/babel/babel/blob/master/Makefile#L99-L100

Does this test the transforms, or [just the parser](https://github.com/babel/babel/blob/master/scripts/tests/test262/run_babel_parser_test262_utils.js#L59)?

rwaldron:
@jbhoosreddy I didn't previously recognize that this was meant to be focused on private methods—I thought it was for **all** new class field and class private (instance and static) features. So far we've been focused only on these: 

- https://tc39.github.io/proposal-class-fields/
- https://tc39.github.io/proposal-static-class-features/

Which means there is presently no coverage for: 

- https://tc39.github.io/proposal-private-methods/

rwaldron:
@jbhoosreddy please add `class-methods-private` to features.txt and make sure that the feature appears in all relevant tests 👍

jbhoosreddy:
I'm taking this up in #1709

PR #1659, opened 2018-07-30
Remove folders without unit tests for ES

leobalter:
I observed each file in each of these folders and they don't contain any test matching for specific parts of ECMAScript, without a specific place for Test262.

Ref https://github.com/bocoup/test262-automation/pull/76


leobalter:
I wonder if I need to add anything to the curation_logs json file, in the "curatedFiles" entry.

nomadtechie:
@leobalter `test262-automation`  scripts will log that test files were curated and add them to ` curation_logs json` as `DELETED_IN_TARGET` in the next export PR. You do not need to manage the `curation_logs json` manually.

leobalter:
That's great, thanks!

PR #1660, opened 2018-07-30
Restore class field tests: public, private (+static)

rwaldron:
Most of this was removed previously, but can now be restored and updated.

**This is a work in progress**

I've tagged @leobalter and @spectranaut for review, but I'm not actually ready for review yet.


rwaldron:
@anba If you have any opportunity to review, I always appreciate your time and efforts :)

rwaldron:
This now contains tests for grammar that's blocked by https://github.com/tc39/proposal-static-class-features/issues/37

I just assume that the issue will be resolved and the syntax will match what is shown in the readme of the proposal.

anba:
> @anba If you have any opportunity to review, I always appreciate your time and efforts :)

I haven't read any of the test files, but when executing them, I got the following errors. :-)

test/language/expressions/class/fields-after-same-line-gen-rs-static-method-privatename-identifier-alt.js
etc.
- Assignment to `#stored`, but `#stored` is never defined.


test/language/expressions/class/fields-after-same-line-gen-static-private-methods.js
etc.
- Declares `static #y(){}`, but then assigns to `#y`, which is not allowed, because private name methods are non-writable.
- This means the test always throws a TypeError when executing `this.#y = value;`.


test/language/expressions/class/fields-computed-name-propname-constructor.js
test/language/expressions/class/fields-computed-name-static-propname-constructor.js
etc.
- Invalid tests, `PropName(ComputedPropertyName)` is always `empty`.
- https://tc39.github.io/ecma262/#sec-object-initializer-static-semantics-propname


test/language/expressions/class/static-private-fields-proxy-default-handler-throws.js
- Needs to invoke the method: `C.x` -> `C.x()`.
- And for the proxy: `P.x` -> `P.x()`


test/language/expressions/class/syntax-error-grammar-field-def-has-initializer-no-sc.js
etc.
- Test expects `x = []` is an early syntax error.
- I assume "no-sc" means "no-super-call" (?), but no super-call is present in the test files.


test/language/expressions/class/syntax-error-grammar-field-no-initializer-with-method.js
etc.
- Test expects `x <newline> m(){}` is an early syntax error.
- No idea where this restriction is supposed to come from, unless ASI shouldn't apply at all for class fields?


test/language/expressions/class/syntax-error-grammar-fields.js
test/language/expressions/class/syntax-error-grammar-privatename.js
test/language/expressions/class/syntax-error-grammar-privatenames.js
- Missing error flags in YAML block.


Complete errors with stacktraces -> https://gist.github.com/anba/9045b8ce023eabd9cb90920c4427923b

rwaldron:
@anba Thank! I'll check on these today :)

rwaldron:
> - Assignment to #stored, but #stored is never defined.

This was corrected in a recent push to the WIP branch.

> - Declares static #y(){}, but then assigns to #y, which is not allowed, because private name methods are non-writable.
> - This means the test always throws a TypeError when executing this.#y = value;.

This was corrected in a recent push to the WIP branch.

> test/language/expressions/class/fields-computed-name-propname-constructor.js
test/language/expressions/class/fields-computed-name-static-propname-constructor.js
etc.
> 
> - Invalid tests, PropName(ComputedPropertyName) is always empty.
> - https://tc39.github.io/ecma262/#sec-object-initializer-static-semantics-propname

Dropped

> test/language/expressions/class/static-private-fields-proxy-default-handler-throws.js
>
> - Needs to invoke the method: C.x -> C.x().
> - And for the proxy: P.x -> P.x()

Will be fixed in next push

> test/language/expressions/class/syntax-error-grammar-field-def-has-initializer-no-sc.js
etc.
>
> - Test expects x = [] is an early syntax error.
> - I assume "no-sc" means "no-super-call" (?), but no super-call is present in the test files.

"no semi-colon" ;)

...But I'm seeing now that some of these are missing the offending _ClassElement_ or _FieldDefinition_ necessary for the syntax error exception. I'm going to work on those as well. 


> test/language/expressions/class/syntax-error-grammar-field-no-initializer-with-method.js
etc.
> 
> - Test expects x <newline> m(){} is an early syntax error.
> - No idea where this restriction is supposed to come from, unless ASI shouldn't apply at all for class fields?


This was corrected in a recent push to the WIP branch.




> test/language/expressions/class/syntax-error-grammar-fields.js
> test/language/expressions/class/syntax-error-grammar-privatename.js
> test/language/expressions/class/syntax-error-grammar-privatenames.js
> 
> Missing error flags in YAML block.


These don't exist anymore, so I guess "problem solved!". I will check to make sure the missing YAML didn't surface in some other files.

src/class-fields/grammar-field-classelementname-initializer-alt.case
```diff
@@ -0,0 +1,68 @@
...
+features: [class-fields-public]
+---*/
+
+//- fields
+$ = 1;
+_ = 2;
+\u{6F} = 3;
+℘ = 4;      // DO NOT CHANGE THE NAME OF THIS FIELD
+ZW_‌_NJ = 5; // DO NOT CHANGE THE NAME OF THIS FIELD
+ZW_‍_J = 6;  // DO NOT CHANGE THE NAME OF THIS FIELD
```
leobalter: the most lovable variable names in the world.

src/class-fields/grammar-fields-same-line-error.case
```diff
@@ -0,0 +1,26 @@
+// Copyright (C) 2018 Rick Waldron. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: SyntaxError
```
leobalter: I wish the description could be a bit longer here

src/class-fields/grammar-privatename-error.case
```diff
@@ -0,0 +1,29 @@
+// Copyright (C) 2018 Rick Waldron. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: No space allowed between sigial and IdentifierName
```
leobalter: typo, sigial

src/class-fields/initializer-error/cls-decl-fields-comp-name.template
```diff
@@ -0,0 +1,16 @@
...
+name: computed ClassElementName
+features: [class, class-fields-public]
+---*/
+
+throw "Test262: This statement should not be evaluated.";
+
+var x = "string";
+class C {
+  [x] = /*{ initializer }*/;
+}
```
leobalter: thanks for fixing this

src/class-fields/rs-static-async-generator-method-privatename-identifier-alt.case
```diff
@@ -0,0 +1,117 @@
...
+  "ID_Continue" include, respectively, the code points with Unicode
+  properties "Other_ID_Start" and "Other_ID_Continue".
+
+template: productions
+flags: [async]
+features: [class-static-methods-private]
+---*/
+
+//- fields
+static async #$(value) {
```
leobalter: It's so weird that we are matching proposals to have static private methods, but in any case, bring it on.

src/class-fields/rs-static-async-method-privatename-identifier.case
```diff
@@ -0,0 +1,118 @@
...
+
+
+  NOTE 3
+  The sets of code points with Unicode properties "ID_Start" and
+  "ID_Continue" include, respectively, the code points with Unicode
+  properties "Other_ID_Start" and "Other_ID_Continue".
+
+template: productions
+flags: [async]
+features: [class-static-methods-private]
```
leobalter: I'm glad we have proper flags for these

anba:
Updated list of test failures: https://gist.github.com/anba/42051c5232bec1ac5bfde560995ac586

src/class-fields/rs-static-async-generator-method-privatename-identifier-alt.case
- Uses `async`, but instead needs to use `async*` to declare async generator methods.

test/language/expressions/class/fields-after-same-line-gen-rs-static-generator-method-privatename-identifier-alt.js (multiple files with the same error)
- Needs to use `yield*` if delegating yield semantics are intended.
- (The test assumes `(function*() { yield function*(){ yield 0; }()})() === 1` is `true`, which is obviously incorrect.)

test/language/expressions/class/fields-after-same-line-gen-static-private-methods.js (multiple files with the same error)
- Has assignment to private name method, which is not allowed and throws a TypeError at runtime.

test/language/expressions/class/fields-after-same-line-static-async-gen-rs-static-async-generator-method-privatename-identifier.js (multiple files with the same error)
- Calls `$DONE()` multiple times.

test/language/statements/class/syntax-invalid-grammar-field-def-has-initializer-no-sc-error.js (multiple files with the same error)
- Test incorrectly assumes `x = [] <line-break> y;` is a SyntaxError.

leobalter:
> Test incorrectly assumes x = [] <line-break> y; is a SyntaxError.

That's a reading from @rwaldron of the proposed syntax:

``` 
  ClassElement :
    MethodDefinition
    static MethodDefinition
    FieldDefinition ;
    ;
``` 

where this reads the sc char in `FieldDefinition ;` as not optional. While I understand his PoV, I agree the grammar is beat by other parts like the do-while statement which has a sc in a point known for ASI.

That grammar formatting might need work, but this would be a side-tracking for this PR.

If you want to stick to the bikeshedding, we could think on possible alternatives for the trailing `;` that is a clear signal for separating fields with `;` if they are in the same line. The alternative I have in mind is ridiculously verbose, thou:

```
ClassElement :
  MethodDefinition
  static MethodDefinition
  FieldDefinitionList
  ;

FieldDeclarationList :
  FieldDeclarationOthers_opt FieldDeclaration

FieldDeclarationOthers :
  FieldDeclarationList [no LineTerminator here ] ; [no LineTerminator here ] FieldDeclaration
```

I'm not sure if that really works as the idea is to prevent multiple fields on the same line without a separator, but that's also kind weird as we can stack multiple methods with a field in the same line: `class C { m1() {} m2() {} }`.

As I stated before, I wish the sc could have avoided as an element of ClassElement, but that's a battle lost since ES2015.

leobalter:
@anba I tried to fix all the parts you mentioned. Can you help me there with another review, please?

leobalter:
If it helps, I ran the tests using es6draft and found no errors on the tests for this PR.

leobalter:
Travis CI is getting me some weird results as it appears to be comparing with another commit. I'm rebasing this locally and rolling a force push to see if anything changes for Travis.

leobalter:
The last commits should fix the problem. I had templates using the same path value. Thanks @jugglinmike for helping me finding the bug.

anba:
> If it helps, I ran the tests using es6draft and found no errors on the tests for this PR.

Hmm, the current version on GH is probably missing some fixes. With my local development version, I still get the following failures: https://gist.github.com/anba/25a51ad514b91870bdcb89f3ad478ed4

Most of the failures happen because `rs-static-async-generator-method-privatename-identifier.case` passes the value `1` to `yield*` (there's `yield * await value` with `value = 1`).

leobalter:
@anba I think 7ea8043 should fix this. Please let me know.

anba:
There seems to be a similar issue in `rs-static-generator-method-privatename-identifier-alt.case` and `rs-static-generator-method-privatename-identifier.case` (maybe more, I've stopped after those two files).


Additionally there are still some invalid property name tests:
- test/language/expressions/class/fields-computed-name-static-propname-prototype.js
- test/language/statements/class/fields-computed-name-propname-constructor.js
- test/language/statements/class/fields-computed-name-static-computed-var-propname-constructor.js
- test/language/statements/class/fields-computed-name-static-computed-var-propname-prototype.js
- test/language/statements/class/fields-computed-name-static-propname-constructor.js
- test/language/statements/class/fields-computed-name-static-propname-prototype.js
- test/language/statements/class/privatefields-string-name-propname-constructor.js

anba:
I see only a single remaining failure (`test/language/statements/class/privatefields-string-name-propname-constructor.js` incorrectly assume `["#constructor"]` is not valid). :+1:

leobalter:
> I see only a single remaining failure (test/language/statements/class/privatefields-string-name-propname-constructor.js incorrectly assume ["#constructor"] is not valid).

Thanks! I was still working on it, but I wanted to understand what that file was doing. In the process I even found a duplicate for checking privatenames of `#constructor`

leobalter:
@anba I believe this will be good to go for a final review. Thank you so much for all the support.

anba:
Zero failures with the latest commit!

leobalter:
I'm getting a pretty positive feedback from parts I can test from other engines, so I'm honestly assuming this is good to ship for now. I'd be happy to address any eventual bug if we found something.

PR #1669, opened 2018-08-10
Intl.RelativeTimeFormat: Add some tests for different styles in Polish.

Ms2ger:
This is interesting because Polish has different output for the "short" and
"narrow" styles.

### DIFF
--- a/test/intl402/RelativeTimeFormat/prototype/format/pl-pl-style-long.js
+++ b/test/intl402/RelativeTimeFormat/prototype/format/pl-pl-style-long.js
@@ -1,0 +1,69 @@
+// Copyright 2018 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-Intl.RelativeTimeFormat.prototype.format
+description: Checks the behavior of Intl.RelativeTimeFormat.prototype.format() in Polish.
+features: [Intl.RelativeTimeFormat]
+locale: [pl-PL]
+---*/
+
+function regular(s) {
+  return {
+    "many": s,
+    "few": s + "y",
+    "one": s + "ę",
+  }
+}
+
+// https://www.unicode.org/cldr/charts/33/summary/pl.html#1419
+const units = {
+  "second": regular("sekund"),
+  "minute": regular("minut"),
+  "hour": regular("godzin"),
+  "day": {
+    "many": "dni",
+    "few": "dni",
+    "one": "dzień",
+  },
+  "week": {
+    "many": "tygodni",
+    "few": "tygodnie",
+    "one": "tydzień",
+  },
+  "month": {
+    1000: "miesięcy",
+    "many": "miesięcy",
+    "few": "miesiące",
+    "one": "miesiąc",
+  },
+  "quarter": {
+    "many": "kwartałów",
+    "few": "kwartały",
+    "one": "kwartał",
+  },
+  "year": {
+    "many": "lat",
+    "few": "lata",
+    "one": "rok",
+  },
+};
+
+const rtf = new Intl.RelativeTimeFormat("pl-PL", {
+  "style": "long",
+});
+
+assert.sameValue(typeof rtf.format, "function", "format should be supported");
+
+for (const [unitArgument, expected] of Object.entries(units)) {
+  assert.sameValue(rtf.format(1000, unitArgument), `za 1\u00a0000 ${expected.many}`);
+  assert.sameValue(rtf.format(10, unitArgument), `za 10 ${expected.many}`);
+  assert.sameValue(rtf.format(2, unitArgument), `za 2 ${expected.few}`);
+  assert.sameValue(rtf.format(1, unitArgument), `za 1 ${expected.one}`);
+  assert.sameValue(rtf.format(0, unitArgument), `za 0 ${expected.many}`);
+  assert.sameValue(rtf.format(-0, unitArgument), `0 ${expected.many} temu`);
+  assert.sameValue(rtf.format(-1, unitArgument), `1 ${expected.one} temu`);
+  assert.sameValue(rtf.format(-2, unitArgument), `2 ${expected.few} temu`);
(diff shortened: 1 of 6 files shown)


rkirsling:
Are these tests valid with the thousands separator?

I see that V8 and SM pass them, but this seems wrong, as [CLDR](https://github.com/unicode-org/icu/blob/master/icu4c/source/data/locales/pl.txt#L283) has `minimumGroupingDigits{"2"}` for Polish and `ureldatefmt_format[Numeric]` respects that in ICU 64.

V8 and SM implementations of `RelativeTimeFormat` seem to ignore minimumGroupingDigits for other locales like `es` too, but I'm not seeing a reason for this in ECMA-402 or ICU's commit history.

mathiasbynens:
@FrankYFTang

FrankYFTang:
But filed in https://unicode-org.atlassian.net/browse/ICU-21086

rkirsling:
I don't think this is an ICU bug unless it was introduced after 64.2 though?

Like I was saying, in 64.2, making a URelativeDateTimeFormatter with `ureldatefmt_open` (using the default UNumberFormat and locale `pl`) and then calling `ureldatefmt_format` does respect minimumGroupingDigits in [JSC](https://github.com/WebKit/webkit/blob/master/Source/JavaScriptCore/runtime/IntlRelativeTimeFormat.cpp#L233-L267). That's why I'm so confused by the behavior of V8 and SM.

FrankYFTang:
> I don't think this is an ICU bug unless it was introduced after 64.2 though?

It is totally possible. 

> 
> Like I was saying, in 64.2, making a URelativeDateTimeFormatter with `ureldatefmt_open` (using the default UNumberFormat and locale `pl`) and then calling `ureldatefmt_format` does respect minimumGroupingDigits. That's what we're [doing in JSC](https://github.com/WebKit/webkit/blob/master/Source/JavaScriptCore/runtime/IntlRelativeTimeFormat.cpp#L233-L267) and it's why I'm so confused by the behavior of V8 and SM.

So... in JSC what is the output of the following code?
```
es = new Intl.RelativeTimeFormat("es")
es.format(1234, "day")
es.format(12345, "day")
pl = new Intl.RelativeTimeFormat("pl")
pl.format(1234, "day")
pl.format(12345, "day")
```
Currently, in v8, the output is INCORRECT as
```
"dentro de 1.234 días"
"dentro de 12.345 días"
"za 1 234 dni"
"za 12 345 dni"
```

The correct output should be
```
"dentro de 1234 días"
"dentro de 12.345 días"
"za 1234 dni"
"za 12 345 dni"
```
since both es and pl has minimumGroupingDigits{"2"}

rkirsling:
JSC does get the expected output:
```
"dentro de 1234 días"
"dentro de 12.345 días"
"za 1234 dni"
"za 12 345 dni"
```
...although I'm not sure if this would change if we build against ICU 65+.

FrankYFTang:
I don't know how JSC manage to do that in ICU64. I checkout maint/maint-64 and try the test patch in https://unicode-org.atlassian.net/browse/ICU-21086

FrankYFTang:
could you point to a url which showing that exact spot of JSC source code of calling ureldatefmt_open  that I can take a look at . Thanks But I agree w/ you that the test in this PR is WRONG. 

\u00a0 should be removed from the test. It will make v8 produce error but that reflect what it should be.

rkirsling:
> could you point to a url which showing that exact spot of JSC source code of calling ureldatefmt_open that I can take a look at . Thanks But I agree w/ you that the test in this PR is WRONG.

https://github.com/WebKit/webkit/blob/master/Source/JavaScriptCore/runtime/IntlRelativeTimeFormat.cpp#L157-L168

It's actually possible that Apple's doing something peculiar with their system ICU?
https://opensource.apple.com/source/ICU/ICU-64243.0.1/

> \u00a0 should be removed from the test. It will make v8 produce error but that reflect what it should be.

Cool! I really appreciate the confirmation.

FrankYFTang:
> > could you point to a url which showing that exact spot of JSC source code of calling ureldatefmt_open that I can take a look at . Thanks But I agree w/ you that the test in this PR is WRONG.
> 
> https://github.com/WebKit/webkit/blob/master/Source/JavaScriptCore/runtime/IntlRelativeTimeFormat.cpp#L157-L168
> 
> It's actually possible that Apple's doing something peculiar with their system ICU?
> https://opensource.apple.com/source/ICU/ICU-64243.0.1/
> 
> > \u00a0 should be removed from the test. It will make v8 produce error but that reflect what it should be.
> 
> Cool! I really appreciate the confirmation.

I think I know why, the code you point to is calling ureldatefmt_open with a supplied UNumberFormat in the second argument (notice the m_numberFormat.get()). 
According to https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/ureldatefmt_8h.html#ad321349fc3e3f6da7c0b1a542dc49ffb

nfToAdopt | A number formatter to set for this URelativeDateTimeFormatter object (instead of the default decimal formatter). Ownership of this UNumberFormat object will pass to the URelativeDateTimeFormatter object (the URelativeDateTimeFormatter adopts the UNumberFormat), which becomes responsible for closing it. If the caller wishes to retain ownership of the UNumberFormat object, the caller must clone it (with unum_clone) and pass the clone to ureldatefmt_open. May be NULL to use the default decimal formatter.
-- | --

so it is not letting the URelativeDateTimeFormatter to use the default decimal formatter.

rkirsling:
Hmm, I initially was using the default formatter though (since ownership of UNumberFormat is only needed for `formatToParts` and not for `format`) and the behavior was the same for this purpose.

FrankYFTang:
It is possible that the NumberFormat (not NumberFormatter) on Mac is fixed. I know we still have issue in ICU that NumberFormatter respect minimumGroupingDigits{"2"} but not NumberFormat. It is possible Apple has a local fix for that.

FrankYFTang:
any way I will dig into that in ICU https://unicode-org.atlassian.net/browse/ICU-21086 and talk to Peter Edberg @ Apple about that.

PR #1677, opened 2018-08-13
Object.fromEntries: use verifyProperty; add specification details

rwaldron:
https://tc39.github.io/proposal-object-from-entries/


test/built-ins/Object/fromEntries/iterator-closed-for-throwing-entry-value-tostring.js
```diff
@@ -0,0 +1,68 @@
+// Copyright (C) 2018 Kevin Gibbons. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-object.fromentries
+description: Closes iterators when toString on a value throws.
```
bakkot: This one isn't valid; only the keys get toString'd (or, to be precise, toPropertyKey'd). The values are passed through without being touched.

test/built-ins/Object/fromEntries/iterator-closed-for-null-entry.js
```diff
@@ -19,7 +36,7 @@ var iterable = {
         advanced = true;
         return {
           done: false,
-          value: 'null',
+          value: null,
```
bakkot: 👍

test/built-ins/Object/fromEntries/iterator-closed-for-throwing-entry-value-accessor.js
```diff
@@ -0,0 +1,64 @@
+// Copyright (C) 2018 Kevin Gibbons. All rights reserved.
```
bakkot: Presumably this (and other new tests) should be (C) someone else?

test/built-ins/Object/fromEntries/iterator-closed-for-throwing-entry-value-tostring.js
rwaldron: Whoops, yes, that would make sense. I will drop this.

rwaldron:
> Thanks for the followup; sorry for my typos.

You're very welcome and don't worry about the typos :)

test/built-ins/Object/fromEntries/iterator-closed-for-throwing-entry-value-accessor.js
rwaldron: Well, I didn't really write a majority of the test, it's your work with only a slight modification

PR #1678, opened 2018-08-14
Intl.RelativeTimeFormat: Add some tests for non-object options arguments to the constructor.

Ms2ger:


test/intl402/RelativeTimeFormat/constructor/constructor/options-toobject-prototype.js
```diff
@@ -0,0 +1,35 @@
+// Copyright 2018 Igalia, S.L. All rights reserved.
```
leobalter: this `constructor/constructor/` path was not introduced in this PR but it's confusing. From the common idea of Test262 files structure, mostly from the built-ins folder, these are tests that could live directly on `test/intl402/RelativeTimeFormat/`.

In any case, after reviewing these tests, they look like they belong to the `test/intl402/RelativeTimeFormat/prototype/resolvedOptions` folder instead. We are checking results of calling it, even if the differences come from when the initial object is created.

test/intl402/RelativeTimeFormat/constructor/constructor/options-toobject-prototype.js
```diff
@@ -0,0 +1,35 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-Intl.RelativeTimeFormat
```
leobalter: a reference to the [`resolvedoptions`](http://tc39.github.io/proposal-intl-relative-time/#sec-intl.relativetimeformat.prototype.resolvedoptions) method would help a lot here or at least anything in the info part.

test/intl402/RelativeTimeFormat/constructor/constructor/options-toobject-prototype.js
```diff
@@ -0,0 +1,35 @@
...
+  true,
+  "test",
+  7,
+  Symbol(),
+];
+
+for (const options of optionsArguments) {
+  const rtf = new Intl.RelativeTimeFormat([], options);
+  const resolvedOptions = rtf.resolvedOptions();
+  assert.sameValue(resolvedOptions.style, "short",
```
leobalter:
```
1.1.1 InitializeRelativeTimeFormat ( relativeTimeFormat, locales, options )

...
12. Let s be ? GetOption(options, "style", "string", «"long", "short", "narrow"», "long").
13. Set relativeTimeFormat.[[Style]] to s.
14. Let numeric be ? GetOption(options, "numeric", "string", «"always", "auto"», "always").
15. Set relativeTimeFormat.[[Numeric]] to numeric.
```

```
1.4.5 Intl.RelativeTimeFormat.prototype.resolvedOptions ()

...
5. For each row of Table 1, except the header row, do
  a. Let p be the Property value of the current row.
  b. Let v be the value of pr's internal slot whose name is the Internal Slot value of the current row.
  c. Assert: v is not undefined.
  d. Perform ! CreateDataPropertyOrThrow(options, p, v).
```

The proposal spec gives some ideas that we could give more strength to the tests:

- Considering [ToObject](https://tc39.github.io/ecma262/#sec-toobject), set the properties directly on the respective constructors prototype, not Object.prototype.
- Use accessor descriptors to store when the value is capture - when the new instance is created - and check again that `resolvedOptions` won't verify the values again. ~~An alternative - but weaker - would be changing the options values before calling resolvedOptions~~.
- Verify the descriptor of DataProperty, not only the Get operation.

test/intl402/RelativeTimeFormat/constructor/constructor/options-undefined.js
```diff
@@ -0,0 +1,38 @@
...
+  "numeric": {
+    get() {
+      throw new Error("Should not call numeric getter");
+    }
+  },
+})
+
+const optionsArguments = [
+  [],
+  [[]],
```
leobalter: this not really much different than the previous item for what concerns for the options...

test/intl402/RelativeTimeFormat/constructor/constructor/options-undefined.js
```diff
@@ -0,0 +1,38 @@
...
+  [[], undefined],
+];
+
+for (const args of optionsArguments) {
+  const rtf = new Intl.RelativeTimeFormat(...args);
+  const resolvedOptions = rtf.resolvedOptions();
```
leobalter: If the other tests capture that getting the option values are done doing the instance creation - `new Intl.RelativeTimeFormat(...args)` then this call of resolvedOptions is sort of testing another thing, the default values  are another separate thing being tested here. Not much to be changed here.

Ideally it would be nice to have these traps being triggered when the options arg is not undefined.

test/intl402/RelativeTimeFormat/constructor/constructor/options-undefined.js
```diff
@@ -0,0 +1,38 @@
...
+  assert.sameValue(resolvedOptions.style, "long",
+    `Calling with ${args.length} empty arguments should yield the correct value for "style"`);
```
leobalter: these are not exactly empty arguments, but `Intl.RelativeTimeFormat` handling `options` with an undefined value.

leobalter:
In any case, all of my feedback can be done in a follow up work. I don't want to request any sweating. Let me know wdyt.

PR #1680, opened 2018-08-14
Add test for private accessor methods

ktbee:
This PR adds templates  and generated tests for the [proposed private methods and getter/setters for JavaScript classes](https://github.com/tc39/proposal-private-methods). One issue I ran into is that single quotes are being included in generated accessor names, like in this example in `private-accessor-name-inst-computed.js`:

```
getPrivateReference() {
    return this[#'string'];
  }

setPrivateReference(value) {
    this[#'string'] = value;
}
```

One solution I was picturing was adding new variables in the case files for something like `privateDeclareWith` and `privateReferenceWith` and add the hash sign to the accessor name string there. Does that make sense or should I change this elsewhere?

### DIFF
--- a/src/accessor-names/default/cls-private-decl-inst.template
+++ b/src/accessor-names/default/cls-private-decl-inst.template
@@ -1,0 +1,26 @@
+// Copyright (C) 2017 Mike Pennisi. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+path: language/statements/class/accessor-name-inst-
+name: Class declaration, instance method
+esid: sec-runtime-semantics-classdefinitionevaluation
+es6id: 14.5.14
+info: |
+  [...]
+  21. For each ClassElement m in order from methods
+      a. If IsStatic of m is false, then
+         i. Let status be the result of performing PropertyDefinitionEvaluation
+            for m with arguments proto and false.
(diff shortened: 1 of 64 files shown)


ljharb:
I’m confused; private fields are not allowed to be computed - ie, no bracket access whatsoever.

leobalter:
Jordan, that’s my own fault. I was exercising Test262 templates and test
generation with Katie and asked her to open the PR. I didn’t have a proper
time to roll through the proposal yet and I missed this.

I also had some meetings until the EOD and couldn’t review the PR in time.

I’ll be back to this work and related templates on Thursday.

Thanks for reporting the issue.

ljharb:
No worries, just wanted to make sure :-) I’ll wait to do a more thorough review.

leobalter:
this is indeed invalid. After reviewing #1660 I'll check other parts we could reuse the templates here

leobalter:
We can still verify getters and setters together for the same private fields, we only need to remove the computed property names out of the way.

rwaldron:
> we only need to remove the computed property names out of the way.

Can you push the additional commits for this? Thanks!

leobalter:
Yes I will

leobalter:
Commits added on #1922

PR #1686, opened 2018-08-19
Array.prototype.{flat,flatMap} property access count

rhuanjl:
Tests for the number of has/get property accesses that occur when using Array.prototype.flat/flatMap.

@ljharb asked me to submit these after my initial implementation of flat and flatMap for chakracore was wrong on this point.

(I haven't signed a CLA here yet so assume I'll need to do that before this is merged)
EDIT: I have now completed a CLA

### DIFF
--- a/test/built-ins/Array/prototype/flat/proxy-access-count.js
+++ b/test/built-ins/Array/prototype/flat/proxy-access-count.js
@@ -1,0 +1,24 @@
+// Copyright (C) 2018 Richard Lawrence. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-array.prototype.flat
+description: >
+    properties are accessed correct number of times by .flat
+features: [Array.prototype.flat]
+includes: [compareArray.js]
+---*/
+
+
+const getCalls = [], hasCalls = [];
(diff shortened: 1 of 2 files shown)


test/built-ins/Array/prototype/flat/proxy-access-count.js
```diff
@@ -0,0 +1,23 @@
...
+  get : function (t, p, r) { ++getCount; return Reflect.get(t, p, r); },
+  has : function (t, p, r) { ++hasCount; return Reflect.has(t, p, r); }
+}
+
+const tier2 = new Proxy([4, 3], handler);
+const tier1 = new Proxy([2, [3, [4, 2], 2], 5, tier2, 6], handler);
+
+Array.prototype.flat.call(tier1, 3);
+
+assert.sameValue(getCount, 10, 'getProperty by .flat should occur exactly once per property and once for length and constructor');
```
michaelficarra: Instead of just counting, you can keep a trace of all of the arguments lists from the proxy trap and make this assertion much stronger.

rhuanjl: revised

PR #1687, opened 2018-08-20
some quick fixes for Array.prototype.{flat,flatMap} tests

leobalter:
some quick changes here with some cleanup included. The tests are basically the same.

cc @michaelficarra


leobalter:
> I’d rather see renamed vars than reassignment but LGTM

Same thought, but I just wanted something initially fast to apply, avoiding multiple var declarations for the same names.

PR #1694, opened 2018-08-27
TypedArray.prototype.sort: check result of compareFn is immediately converted ToNumber

khyperia:
This was originally a [bug in spidermonkey](https://bugzilla.mozilla.org/show_bug.cgi?id=1456973).

In it, the result of comparefn was only ran through `ToNumber` upon comparing it to -1/0/1, which has an observable difference in behavior (as this test indicates).

### DIFF
--- a/test/built-ins/TypedArray/prototype/sort/sort-tonumber.js
+++ b/test/built-ins/TypedArray/prototype/sort/sort-tonumber.js
@@ -12,7 +12,7 @@ info: |
     b. If IsDetachedBuffer(buffer) is true, throw a TypeError exception.
     ...
   ...
-includes: [testTypedArray.js]
+includes: [testTypedArray.js, detachArrayBuffer.js]
 features: [TypedArray]
 ---*/
 
@@ -21,19 +21,17 @@ testWithTypedArrayConstructors(function(TA) {
   var ab = ta.buffer;
 
   var called = false;
-  try {
+  assert.throws(TypeError, function() {
     ta.sort(function(a, b) {
       // IsDetachedBuffer is checked right after calling comparefn.
       // So, detach the ArrayBuffer to cause sort to throw, to make sure we're actually calling ToNumber immediately (as spec'd)
       // (a possible bug is to wait until the result is inspected to call ToNumber, rather than immediately)
-      detachArrayBuffer(ab);
+      $DETACHBUFFER(ab);
       return {
         [Symbol.toPrimitive]() { called = true; }
       };
     });
-  } catch (e) { }
+  });
 
   assert.sameValue(true, called);
-})
-
-reportCompare(0, 0);
+});
--- a/test/built-ins/TypedArray/prototype/sort/sort-tonumber.js
+++ b/test/built-ins/TypedArray/prototype/sort/sort-tonumber.js
@@ -1,0 +1,39 @@
+// Copyright (C) 2018 Mozilla Corporation. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-%typedarray%.prototype.sort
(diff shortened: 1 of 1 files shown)


test/built-ins/TypedArray/prototype/sort/sort-tonumber.js
```diff
@@ -0,0 +1,39 @@
...
+    ta.sort(function(a, b) {
+      // IsDetachedBuffer is checked right after calling comparefn.
+      // So, detach the ArrayBuffer to cause sort to throw, to make sure we're actually calling ToNumber immediately (as spec'd)
+      // (a possible bug is to wait until the result is inspected to call ToNumber, rather than immediately)
+      detachArrayBuffer(ab);
+      return {
+        [Symbol.toPrimitive]() { called = true; }
+      };
+    });
+  } catch (e) { }
```
leobalter: this try catch block will hide other effects. If `detachArrayBuffer` throws an error - e.g. it's not implemented in a specific engine - the next assertion for `called === true`  will fail.

Removing the try catch will tell what exactly is failing.

test/built-ins/TypedArray/prototype/sort/sort-tonumber.js
```diff
@@ -0,0 +1,39 @@
...
+
+  assert.sameValue(true, called);
+})
+
+reportCompare(0, 0);
```
leobalter: this is spidermonkey's exclusive.

test/built-ins/TypedArray/prototype/sort/sort-tonumber.js
```diff
@@ -0,0 +1,39 @@
...
+  var ta = new TA(4);
+  var ab = ta.buffer;
+
+  var called = false;
+  try {
+    ta.sort(function(a, b) {
+      // IsDetachedBuffer is checked right after calling comparefn.
+      // So, detach the ArrayBuffer to cause sort to throw, to make sure we're actually calling ToNumber immediately (as spec'd)
+      // (a possible bug is to wait until the result is inspected to call ToNumber, rather than immediately)
+      detachArrayBuffer(ab);
```
leobalter: we don't have a global detachArrayBuffer method, but we can use the one available from the helpers, just adding it to the includes option and calling `$DETACHBUFFER`

test/built-ins/TypedArray/prototype/sort/sort-tonumber.js
```diff
@@ -0,0 +1,39 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-%typedarray%.prototype.sort
+description: The result of compareFn is immediately passed through ToNumber
+info: |
+  22.2.3.26 %TypedArray%.prototype.sort ( comparefn )
+
+  ...
+  2. If comparefn is not undefined, then
+    a. Let v be ? ToNumber(? Call(comparefn, undefined, « x, y »)).
```
leobalter: I totally missed this change. Now I finally found this https://github.com/tc39/ecma262/pull/1000

I wish it could be easier to find these changes or make sure they are observed in Test262.

test/built-ins/TypedArray/prototype/sort/sort-tonumber.js
leobalter: I forgot to mention, if the case is to prevent the TypeError from the detached buffer, `ta.sort` can be called under an `assert.throws(TypeError, function() { /* ... */ })` instead.

khyperia:
Wasn't sure if the policy here is to squash or add another commit - I think github allows squash-on-merge, so I just added another commit and you can squash if you'd like.

leobalter:
not a problem. There isn't a policy but I prefer new commits so it's easier to see the new changes. Thanks!

PR #1699, opened 2018-08-30
test: Add private generator method tests (#1343)

jbhoosreddy:
Summary of changes:
1. Add tests for private generator methods

### DIFF
--- a/features.txt
+++ b/features.txt
@@ -20,11 +20,15 @@ BigInt
 class-fields-public
 class-fields-private
 
-# Static Class Fields & Methods: 
+# Static Class Fields & Methods:
 # https://github.com/tc39/proposal-static-class-features/
 class-static-fields-public
 class-static-fields-private
 class-static-methods-private
+
+# Private methods and getter/setters
+# https://github.com/tc39/proposal-private-methods
+class-methods-private
 
 # Promise.prototype.finally
 # https://github.com/tc39/proposal-promise-finally
--- a/src/async-generators/default/async-class-decl-static-private-method.template
+++ b/src/async-generators/default/async-class-decl-static-private-method.template
@@ -1,4 +1,4 @@
-// Copyright (C) 2018 Jaideep Bhoosreddy. All rights reserved.
+// Copyright (C) 2018 Bloomberg LP. All rights reserved.
 // This code is governed by the BSD license found in the LICENSE file.
 
 /*---
--- a/src/async-generators/default/async-class-decl-static-private-method.template
+++ b/src/async-generators/default/async-class-decl-static-private-method.template
@@ -16,7 +16,7 @@ info: |
 
   AsyncGeneratorMethod :
     async [no LineTerminator here] * PropertyName ( UniqueFormalParameters ) { AsyncGeneratorBody }
-features: [async-iteration, class-methods-private]
+features: [async-iteration, class-static-methods-private]
 ---*/
 
 var callCount = 0;
@@ -26,15 +26,12 @@ class C {
         callCount += 1;
         /*{ body }*/
     }
-    static gen() {
-        return C.#gen();
-    }
+    static get gen() { return this.#gen; }
 }
 
 // Test the private fields do not appear as properties before set to value
-assert.sameValue(Object.hasOwnProperty.call(C.prototype, "#gen"), false, "test 1");
-assert.sameValue(Object.hasOwnProperty.call(C, "#gen"), false, "test 2");
-assert.sameValue(Object.hasOwnProperty.call(c, "#gen"), false, "test 3");
+assert.sameValue(Object.hasOwnProperty.call(C.prototype, "#gen"), false, 'Object.hasOwnProperty.call(C.prototype, "#gen")');
+assert.sameValue(Object.hasOwnProperty.call(C, "#gen"), false, 'Object.hasOwnProperty.call(C, "#gen")');
 
 var iter = C.gen();
 
@@ -43,6 +40,5 @@ var iter = C.gen();
 assert.sameValue(callCount, 1);
 
 // Test the private fields do not appear as properties after set to value
-assert.sameValue(Object.hasOwnProperty.call(C.prototype, "#gen"), false, "test 1");
-assert.sameValue(Object.hasOwnProperty.call(C, "#gen"), false, "test 2");
-assert.sameValue(Object.hasOwnProperty.call(c, "#gen"), false, "test 3");
+assert.sameValue(Object.hasOwnProperty.call(C.prototype, "#gen"), false, 'Object.hasOwnProperty.call(C.prototype, "#gen")');
+assert.sameValue(Object.hasOwnProperty.call(C, "#gen"), false, 'Object.hasOwnProperty.call(C, "#gen")');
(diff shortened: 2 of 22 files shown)


src/async-generators/default/async-class-decl-static-private-method.template
```diff
@@ -0,0 +1,48 @@
...
+    static PrivateMethodDefinition
+
+  MethodDefinition :
+    AsyncGeneratorMethod
+
+  Async Generator Function Definitions
+
+  AsyncGeneratorMethod :
+    async [no LineTerminator here] * PropertyName ( UniqueFormalParameters ) { AsyncGeneratorBody }
+features: [async-iteration, class-methods-private]
```
leobalter: this should be using the class-static-methods-private

src/async-generators/default/async-class-decl-static-private-method.template
```diff
@@ -0,0 +1,48 @@
...
+    }
+    static gen() {
+        return C.#gen();
+    }
+}
+
+// Test the private fields do not appear as properties before set to value
+assert.sameValue(Object.hasOwnProperty.call(C.prototype, "#gen"), false, "test 1");
+assert.sameValue(Object.hasOwnProperty.call(C, "#gen"), false, "test 2");
+assert.sameValue(Object.hasOwnProperty.call(c, "#gen"), false, "test 3");
```
leobalter: there is no instance. `c` doesn't exist.

src/async-generators/default/async-class-decl-static-private-method.template
leobalter: this applies to the other files here with static methods too.

src/async-generators/default/async-class-decl-static-private-method.template
leobalter: this applies to other files here using static methods as well

src/generators/default/class-decl-method.template
```diff
@@ -2,7 +2,7 @@
 // This code is governed by the BSD license found in the LICENSE file.
 /*---
 path: language/statements/class/gen-method-
-name: Geenerator method as a ClassDeclaration element
+name: Generator method as a ClassDeclaration element
```
leobalter: thanks!

src/generators/default/class-decl-private-method.template
```diff
@@ -0,0 +1,50 @@
...
+    PrivateMethodDefinition
+
+  MethodDefinition :
+    GeneratorMethod
+
+  14.4 Generator Function Definitions
+
+  GeneratorMethod :
+    * PropertyName ( UniqueFormalParameters ) { GeneratorBody }
+features: [generators]
```
leobalter: this is a duplicate, add this to the list on line 6, please.

src/generators/default/class-decl-static-private-method.template
```diff
@@ -0,0 +1,48 @@
...
+
+var iter = gen();
+
+/*{ assertions }*/
+
+assert.sameValue(callCount, 1);
+
+// Test the private fields do not appear as properties before set to value
+assert.sameValue(Object.hasOwnProperty.call(C.prototype, "#gen"), false, "test 1");
+assert.sameValue(Object.hasOwnProperty.call(C, "#gen"), false, "test 2");
```
leobalter: for consistency, missing a line at the end of the file.

src/generators/default/class-expr-static-private-method.template
```diff
@@ -0,0 +1,48 @@
...
+
+  GeneratorMethod :
+    * PropertyName ( UniqueFormalParameters ) { GeneratorBody }
+features: [generators]
+---*/
+
+var callCount = 0;
+
+var C = class {
+    static *gen() {
```
leobalter: there is no private method here.

src/generators/default/class-expr-static-private-method.template
```diff
@@ -0,0 +1,48 @@
...
+        callCount += 1;
+        /*{ body }*/
+    }
+    static gen() {
+        return C.gen();
```
leobalter: and this also needs to get the private name, preferably from `this`.

src/generators/default/class-expr-static-private-method.template
```diff
@@ -0,0 +1,48 @@
...
+    }
+}
+
+var gen = C.gen;
```
leobalter: It should be more adequate if you just use `var iter = C.gen()` (on line 40).

Something else that might simplify is an accessor method:

```js
var C = class {
  static *#gen() {
    /* ... */
  }

  static get gen() { return this.#gen; }
}

var iter = C.gen();
```

src/generators/default/class-expr-private-method.template
```diff
@@ -0,0 +1,50 @@
...
+
+  GeneratorMethod :
+    * PropertyName ( UniqueFormalParameters ) { GeneratorBody }
+features: [generators]
+---*/
+
+var callCount = 0;
+
+var C = class {
+    *gen() {
```
leobalter: not a private method as well...

src/generators/default/class-expr-static-private-method.template
```diff
@@ -0,0 +1,48 @@
...
+assert.sameValue(Object.hasOwnProperty.call(C, "#gen"), false, "test 2");
+
+var iter = gen();
+
+/*{ assertions }*/
+
+assert.sameValue(callCount, 1);
+
+// Test the private fields do not appear as properties before set to value
+assert.sameValue(Object.hasOwnProperty.call(C.prototype, "#gen"), false, "test 1");
```
leobalter: the message can be improved in these assertions, rather than a number, a copy of the actual call `Object.hasOwnProperty.call(C.prototype, "#gen")` can help.

If iterating over a assert number, please consider the previous assertions.

src/generators/default/class-decl-static-private-method.template
```diff
@@ -0,0 +1,48 @@
...
+
+  GeneratorMethod :
+    * PropertyName ( UniqueFormalParameters ) { GeneratorBody }
+features: [generators]
+---*/
+
+var callCount = 0;
+
+class C {
+    static #*gen() {
```
leobalter: I believe it's `* #gen`

src/generators/default/class-decl-static-private-method.template
```diff
@@ -0,0 +1,48 @@
...
+        callCount += 1;
+        /*{ body }*/
+    }
+    static gen() {
+        return C.gen();
```
leobalter: needs to get the private name, may also use `this` instead of `C`.

jbhoosreddy:
Thanks @leobalter. I guess I was waiting for @mkubilay's review before I pinged the maintainers. But I can fix up based on comments.

leobalter:
well, if the PR is open and the title is not prefixed with WIP, I'll always assume it's ready for general review as well :)

I'm looking forward to the fixes. Thanks!

jbhoosreddy:
@leobalter I think I'm close to landing this PR. Can you take a look over the front matter? It maybe inaccurate.

leobalter:
Everything is technically correct here, there are just a few small things I can get fixed quicker than a review process, like including a features tags and a single trailing space.

Thanks for this work, it's really nice!

features.txt
```diff
@@ -20,12 +20,16 @@ BigInt
...
 
-# Static Class Fields & Methods: 
+# Static Class Fields & Methods:
 # https://github.com/tc39/proposal-static-class-features/
 class-static-fields-public
 class-static-fields-private
 class-static-methods-private
 
+# Private methods and getter/setters
+# https://github.com/tc39/proposal-private-methods
```
chicoxyzzy: Shouldn't it be https://github.com/tc39/proposal-class-fields instead?

PR #1700, opened 2018-08-31
Unified Intl.NumberFormat: Some initial tests.

Ms2ger:


leobalter:
AFAICT this is [still stage 2](https://github.com/sffc/proposal-unified-intl-numberformat) right? Would you mind pinging back when it's ready on Stage 3?

Ms2ger:
@jswalden @zbraniecki: are tests for this still a priority for y'all if they're going to sit in a PR indefinitely?

jswalden:
My instinct is to say they're not a _high_ priority if we're not to that stage yet, but that's just a gut feel not informed by past conventions of interaction.  I'll leave @zbraniecki to agree or disagree with me.  :-)

rwaldron:
@jswalden thanks for the follow up. 

For clarification here, Test262 will not merge tests for features that have not yet reached Stage 3.

FrankYFTang:
https://github.com/tc39-transfer/proposal-unified-intl-numberformat
is promoted to stage 3 today @ TC39 meeting (per https://docs.google.com/document/d/1CjucI0A_56xvuDT-Zo6CEBlrq4_TXfP3FK16LFNyX7o/edit )

FrankYFTang:
> @FrankYFTang please don't post links to the meetings notes.

@rwaldron - sorry, I removed that comment. I am new to TC39 so I may accidentally violate the social protocol here about such. Could you please educate me about why? Is TC39 meeting note considered private to TC39 members only? I may wrongfully assume those are in the public view.

leobalter:
>  Is TC39 meeting note considered private to TC39 members only?

They are private to TC39 until review and approval that occurs in at least 7 days after each respective meeting. There might be security concerns that should not be publicized and the delegates should be given proper time to review these.

Also, the link is open for anyone to edit, so this could jeopardize the note taking process of this on going meeting if any troll gets the link.

After review and approval, the notes are sent to Ecma and released on the respective repo.

PR #1711, opened 2018-09-05
some specific class elements cases

leobalter:
The main motivation for this PR is to tackle a coverage gap for cases where special methods can't be named `constructor` in a class body.

See ~~b2d4673~~ 9e740fb

This uses the generated tests to enforce cases in class expressions as well.

At some point, the src/class-fields folder is already checking for class elements, including methods. So this PR renames it for a more appropriate notation.


leobalter:
This PR covers cases found after analyzing the [regression test](https://github.com/tc39/test262/pull/1710/files#diff-bd1fe45b2b4885e11ccea9b0519f23e9R1) that can now be safely removed (56dfee3)

rwaldron:
This is incredible work. It's nice to see that the implementation contribution curation process is really paying off with large scale coverage changes being surfaced.

leobalter:
Thanks, @anba! I'll have a patch fixing these issues this afternoon.

PR #1712, opened 2018-09-06
Assert extensibility of dataview objects

leobalter:
This adds assertions to confirm extensibility of dataview objects, based on the contributed test from JSC.

See implementation-contributed/javascriptcore/stress/regress-189186.js


rwaldron:
This is excellent.

PR #1760, opened 2018-09-13
test: Add test for private method in object literal

jbhoosreddy:
Tracking Issue: #1343

### Summary of changes:
- Add tests for private method in object literal

### DIFF
--- a/test/language/expressions/object/method-definition/private-method-async-generator-object-literal.js
+++ b/test/language/expressions/object/method-definition/private-method-async-generator-object-literal.js
@@ -1,4 +1,4 @@
-// Copyright (C) 2018 Bloomberg LP. All rights reserved.
+// Copyright (C) 2018 Jaideep Bhoosreddy (Bloomberg LP). All rights reserved.
 // This code is governed by the BSD license found in the LICENSE file.
 
 /*---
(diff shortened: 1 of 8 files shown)


jbhoosreddy:
@rwaldron @leobalter for review

not really sure if that's the right place to even put these tests. thoughts?

rwaldron:
These need to be moved: 

`test/language/literals/object/` => `test/language/expressions/object/method-definition/`

jbhoosreddy:
@rwaldron for review

rwaldron:
I want to think more about this changeset. I have two concerns: 

1. I want to welcome your contribution, because you put in the work and that should be recognized. However...
2. It occurred to me that tests like this are generally avoided because it's hard to tell where the "coverage boundary" begins and ends, ie. what part of the spec is this actually testing? The obvious answer is that it's testing to ensure that _PrivateName_ is not allowed inside _PropertyDefinitionList_—but I'd counter that a lot of things aren't allowed inside _PropertyDefinitionList_... for example, Python code (that's a bit extreme, but I think it illustrates well).

Let's think about this a bit more.

jbhoosreddy:
@rwaldron I don't mind scrapping this PR. If we decide we don't want to cover these, do you mind updating the test plan linked to the PR?

littledan:
We previously considered such "private name shorthand", e.g., in https://github.com/littledan/proposal-private-shorthand , and decided to break it out into a separate, follow-on proposal. I think it's worth it to have a test that it is not supported, given that history. (Related: https://github.com/tc39/test262/pull/1782#issuecomment-425483531)

jbhoosreddy:
similar to #1782. closing

littledan:
Note, the lack of these tests  led to a delay in discovering a real issue in the V8 implementation of private fields and methods: https://chromium-review.googlesource.com/c/v8/v8/+/1461161 . I'd ask you to reconsider whether such tests are OK. Note that the recently committed `super.#x` tests are a similarly specified case of the grammar not having a particular production for it, not an early error. cc @gsathya

PR #1764, opened 2018-09-14
updating sourceRevisionAtLastExport for v8 curation with oldest commit to prime test262-automation for a first run

nomadtechie:


leobalter:
thanks!

PR #1765, opened 2018-09-14
Partial curation for the jsc/es6 tests

leobalter:


leobalter:
edited the title, this work can be continued in another branch if merged now. I'll have some interval before I'm back to the curation process.

PR #1770, opened 2018-09-16
tools: update lint.py to work with pip 10+

ryzokuken:
Update the lint.py script to work with pip 10+ as pip.req was moved to
pip._internal.req in version 10 onwards and the existing code only works
on versions of pip under or equal to 9.0.3


ryzokuken:
Thanks, @rwaldron!

PR #1772, opened 2018-09-17
intl: improve coverage for Intl.NumberFormat

ryzokuken:
Add Intl.NumberFormat.prototype.formatToParts to the test
this-value-not-numberformat.js to check if a TypeError is thrown if the
"this" value for the function call isn't a valid NumberFormat object.

/cc @gsathya @littledan @anba

### DIFF
--- a/test/intl402/NumberFormat/prototype/this-value-not-numberformat.js
+++ b/test/intl402/NumberFormat/prototype/this-value-not-numberformat.js
@@ -12,6 +12,7 @@ author: Norbert Lindenberg
 
 var functions = {
     "format getter": Object.getOwnPropertyDescriptor(Intl.NumberFormat.prototype, "format").get,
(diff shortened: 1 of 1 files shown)


test/intl402/NumberFormat/prototype/this-value-not-numberformat.js
```diff
@@ -12,6 +12,7 @@ author: Norbert Lindenberg
 
 var functions = {
     "format getter": Object.getOwnPropertyDescriptor(Intl.NumberFormat.prototype, "format").get,
+    formatToParts: Intl.NumberFormat.prototype.formatToParts,
     resolvedOptions: Intl.NumberFormat.prototype.resolvedOptions
 };
 var invalidTargets = [undefined, null, true, 0, "NumberFormat", [], {}];
```
gsathya: Add Symbol to this list

PR #1774, opened 2018-09-18
Intl: Add tests for throwing getters in options objects.

Ms2ger:


gsathya:
cc @ryzokuken

gsathya:
@Ms2ger it'd be great to have a similar test that checks that the order in which these getters are accessed is correct :)

Ms2ger:
@gsathya There should be a couple of those around already, I think. Feel free to file an issue if there's any missing.

PR #1777, opened 2018-09-19
Fix stray 'bigint' feature flags

bakkot:
First was introduced [here](https://github.com/tc39/test262/pull/1479/files#diff-1e9f6c0fc0c996566e881503eb23ef29L24).

Second was introduced [here](https://github.com/tc39/test262/commit/7f3a1c1be970b62defe0bc4a542c6efb9d0295bb#diff-4107580a3fb39f9dd84fc8336aacc891L21).


rwaldron:
Thanks!!

PR #1778, opened 2018-09-19
intl: increase coverage for NumberFormat

ryzokuken:
Increase coverage for Intl.NumberFormat by adding a test for format
checking it's default value.

### DIFF
--- a/test/intl402/NumberFormat/prototype/format/default-value.js
+++ b/test/intl402/NumberFormat/prototype/format/default-value.js
@@ -1,0 +1,43 @@
+// Copyright (C) 2018 Ujjwal Sharma. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-number-format-functions
+description: >
+  Tests that the default value for the argument of
+  Intl.NumberFormat.prototype.format (value) is undefined.
+info: |
+  11.1.4Number Format Functions
+
+  3. If value is not provided, let value be undefined.
+---*/
+
+const localesList = [
+  undefined,
+  ['de'],
+  ['th-u-nu-thai'],
+  ['en'],
+  ['ja-u-nu-jpanfin'],
+  ['ar-u-nu-arab']
+];
+const optionsList = [
+  undefined,
+  { style: 'percent' },
+  { style: 'currency', currency: 'EUR', currencyDisplay: 'symbol' },
+  { style: 'currency', currency: 'IQD', currencyDisplay: 'symbol' },
+  { style: 'currency', currency: 'KMF', currencyDisplay: 'symbol' },
+  { style: 'currency', currency: 'CLF', currencyDisplay: 'symbol' },
+  {
+    useGrouping: false,
+    minimumIntegerDigits: 3,
+    minimumFractionDigits: 1,
+    maximumFractionDigits: 3
+  }
+];
+
+localesList.forEach(locales => {
+  optionsList.forEach(options => {
+    const nf = Intl.NumberFormat(locales, options);
+    assert.sameValue(nf.format(), nf.format(undefined));
+  });
+});


test/intl402/NumberFormat/prototype/format/default-value.js
```diff
@@ -0,0 +1,43 @@
...
+    minimumIntegerDigits: 3,
+    minimumFractionDigits: 1,
+    maximumFractionDigits: 3
+  }
+];
+
+localesList.forEach(locales => {
+  optionsList.forEach(options => {
+    const nf = Intl.NumberFormat(locales, options);
+    assert.sameValue(nf.format(), nf.format(undefined));
```
gsathya: Correct me if I'm wrong, but isn't this testing that "a function call without an argument, gets passed undefined"? This isn't Intl.NumberFormat specific test, it's just testing JavaScript. In other words, this could just be `function t(arg) { assert.samevalue(arg, undefined); } t();` right?

test/intl402/NumberFormat/prototype/format/default-value.js
```diff
@@ -0,0 +1,43 @@
...
+/*---
+esid: sec-number-format-functions
+description: >
+  Tests that the default value for the argument of
+  Intl.NumberFormat.prototype.format (value) is undefined.
+info: |
+  11.1.4Number Format Functions
+
+  3. If value is not provided, let value be undefined.
+---*/
```
rwaldron: In the future you'll want to make sure that you include `features: [...]`, but even then, as now, it's ok if they aren't present (I'm working on a tool to automate that, so I like when I get new test material to process with that tool)

test/intl402/NumberFormat/prototype/format/default-value.js
littledan: Note that some JS standard library functions *do* treat a missing argument and passing `undefined` differently. It seems like a reasonable idea to test things here, especially if it ends up corresponding to a different implementation path (as it did in PluralRules where an early implementation had an error).

gsathya: Ah, that's fair. Thanks

rwaldron: @gsathya built-in functions _may_ vary in the specified handling of "what to do when the the arg is not present or undefined". Specific examples include most of the `Date.prototype.set*` methods, eg:

- https://tc39.github.io/ecma262/#sec-date.prototype.setfullyear
  - If month is not present, let m be MonthFromTime(t); otherwise, let m be ? ToNumber(month).

```js
new Date().setFullYear(2019, undefined);
// NaN

new Date().setFullYear(2019);
// 1568904766538
```

Because Intl.NumberFormat.prototype.format => Number Format Functions explicitly specifies what must happen, then I think it's worth having a test for. 

As for the test itself, I think this is overly complex. It can be simplified to something like: 

```js
const nf = new Intl.NumberFormat();
assert.sameValue(nf.format(), "NaN");
assert.sameValue(nf.format(undefined), "NaN");
```

ryzokuken: @gsathya Exactly. I saw this explicitly mentioned in the spec and decided to ensure there's a test for it, in case some implementor ends up using some alternate default values. (I realize that an argument is `undefined` when not passed, but it's not uncommon to see specific default values?)

@rwaldron I avoided something like that to ensure that the value returned by format when called without args is the same as when called with `undefined`, for a large variety of locales and options, thus avoiding hardcoding it to `NaN`. That way, the value associated with `format(undefined)` could change but it won't affect the test unless this particular clause (default value of `undefined`) is changed (at which point, this test would need to be updated anyway).

I do agree that I should try keeping things simple though, so will change the test. That said, I guess your example should have `NaN` instead of `"NaN"`, right?

ryzokuken:
@rwaldron PTAL.

> In the future you'll want to make sure that you include `features: [...]`

Does that apply to this change?

PR #1786, opened 2018-09-23
Refactor increment/decrement tests for parsers

jugglinmike:
The tests for the parsing of postfix increment, postfix decrement,
prefix increment, and prefix decrement were expressed using `eval`.
This made the tests more complex than necessary and also prevented the
tests from providing value to ECMAScript parsers.

Remove the use of `eval` and instead express the expectations with
literal source text.

---

Hey @leobalter; [a bug in eshost prevents it from running these tests in Node.js](https://github.com/bterlson/eshost/pull/66), so you'll have to use a different engine/runner to check my work.


leobalter:
After reviewing all the changes, I used es6draft to confirm the changes. The results are LGTM.

jugglinmike:

PR #1788, opened 2018-09-24
Fix and update tests

anba:
d0cec12299afc96b26f89f32315840d0df270c0b
- Updates matchAll tests per latest spec proposal

9c704c1cc819716c711ce4ae0a6382c8e660e65c
- Explicitly call `String(...)` to convert symbol value to a string, because the implicit ToString conversion throws a TypeError.
- And fix the NaN test to expect a string, which may not necessarily be `"NaN"`.

75e4feddcbeb6e914c815da6717d566956037223
- Split the tests which require the extra precision support into separate files, so that implementations don't have to skip the complete file.
- This also helps to spot the negative zero sign issue in d51d6d3ddd71a2e6259c69cbea6723b9546981ea.

d51d6d3ddd71a2e6259c69cbea6723b9546981ea
- Expect leading sign for negative zero per latest spec changes in ECMA-402.


leobalter:
LGTM, @ljharb do you want to confirm the tests verify the intended changes in the proposal for matchAll??

ljharb:
@leobalter yes, please, that'd be great. i'm hoping to have that done by this evening; after i've stamped then feel free to merge <3

test/built-ins/RegExp/prototype/Symbol.matchAll/isregexp-called-once.js
```diff
@@ -31,7 +30,8 @@ var o = {
   get [Symbol.match]() {
     ++count;
     return false;
-  }
+  },
+  flags: "",
```
ljharb: should this be `get flags() { }` and increment a `flagsCount`, to ensure this is looked up?

leobalter: this can be added as a follow up if you don't mind.

ljharb: certainly

PR #1795, opened 2018-09-26
assert calls order for Symbol.match on matchAll usage

### LINKED ISSUE #1794: Follow-ups for #1788
See: https://github.com/tc39/test262/pull/1788#discussion_r220247358

leobalter:
Fix #1794

### DIFF
--- a/test/built-ins/RegExp/prototype/Symbol.matchAll/isregexp-called-once.js
+++ b/test/built-ins/RegExp/prototype/Symbol.matchAll/isregexp-called-once.js
@@ -25,16 +25,32 @@ Object.defineProperty(RegExp.prototype, Symbol.match, {
   }
 });
 
-var count = 0;
+var calls = [];
 var o = {
   get [Symbol.match]() {
-    ++count;
+    calls.push('get @@match');
     return false;
   },
-  flags: "",
+  get flags() {
+    calls.push('get flags');
+    return {
+      toString() {
+        calls.push('flags toString');
+      }
+    };
+  },
 };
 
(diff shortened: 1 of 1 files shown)


test/built-ins/RegExp/prototype/Symbol.matchAll/isregexp-called-once.js
```diff
@@ -25,16 +25,32 @@ Object.defineProperty(RegExp.prototype, Symbol.match, {
...
+RegExp.prototype[Symbol.matchAll].call(o, {
+  toString() {
+    calls.push('arg toString')
+  }
+});
 
 assert.sameValue(0, internalCount);
-assert.sameValue(1, count);
+
+assert.sameValue(calls.length, 4);
```
ljharb: is there no "assert.deepEqual" that would allow comparing the entire array at a time? that way, failures would be more clear about what's missing.

leobalter: there is a `compareArray` helper but I honestly prefer going this way making it clear I'm expecting the given order and items in sequence.

In any case, there is no easy implementation of a `deepEqual` equivalent, I'd be happy to expand this discussion somewhere else.

leobalter: also, compareArray won't show at what point anything is not equal, just a true or false for comparing array like objects.

ljharb: i'd assume that the error message could be something like "got [a, c] expected [a, b, c]"

PR #1799, opened 2018-09-26
intl: add assertion to test that the result for default parameter

ryzokuken:
Add an assertion to the test for default parameters in Intl.NumberFormat's
method formatToParts checking that both the values are equal to the correct
result.

/cc @rwaldron @gsathya


rwaldron:
Thanks @ryzokuken!

Issue #1634, opened 2018-07-09
Move throw "..." for negative parsing errors to a helper function


mathiasbynens:
Re: #1073: @schuay had the idea of moving the `throw '...'` to a helper function like `failfNoEarlyErrorThrown()` or just `failIfCodeGetsExecuted()`. Has that been considered? It would make it possible for runners to override the function. For example, in V8 we'd want to make it a no-op for certain tests marked with a special flag like `FAIL_PHASE_ONLY`.

mathiasbynens:
To clarify, I'm asking if we could turn the following:

```js
/*---
negative:
  phase: parse
  type: SyntaxError
---*/

throw "Test262: This statement should not be evaluated.";

/?/;
```

...into something like:

```js
/*---
negative:
  phase: parse
  type: SyntaxError
---*/

failIfCodeGetsExecuted();

/?/;
```

jugglinmike:
I never considered that. If `failfNoEarlyErrorThrown` was implemented as a no-op, wouldn't that mean V8 would pass the test without actually following the specification?

mathiasbynens:
These tests verify three things:

A. an exception is thrown
B. the exception is of the expected type
C. the exception is thrown at the correct phase

V8 has known bugs where it throws errors at runtime instead of early. In such cases, I’d like to mark the Test262 test with a special (new) flag in our status file e.g. `FAIL_PHASE_ONLY`, which would then make `failfNoEarlyErrorThrown` a no-op. That way, even if we know we fail C, we can still verify A and B.

OTOH, if we mark the entire test as `FAIL` or skip the test, we don’t even verify `A` and `B` anymore.

jugglinmike:
In spec terms, I don't think an exception is thrown. The text for these errors read, "It is a Syntax Error if [...]" and "It is an early Reference Error if [...]". That suggests that the expected behavior can't be described as two distinctly observable conditions, "A" and "C".

From a developer's perspective, it doesn't seem useful to award "partial credit" for behavior like V8's. The distinction of "execute and throw" versus "fail to parse" is very significant, and only under rarefied circumstances would the difference result in interoperability.

mathiasbynens:
> In spec terms, I don't think an exception is thrown. The text for these errors read, "It is a Syntax Error if [...]" and "It is an early Reference Error if [...]". That suggests that the expected behavior can't be described as two distinctly observable conditions, "A" and "C".

[...]

PR #1801, opened 2018-09-26
Move `throw "…"` for negative parsing errors to a helper function

mathiasbynens:
This PR illustrates what I had in mind for #1634. It’s a nice cleanup on the Test262 side.


rwaldron:
Unrelated to any review or discussion about the utility or acceptance of this changeset: Please undo all changes to all files in the `/implementation-contributed/` directory.

mathiasbynens:
@rwaldron Done.

leobalter:
I find little value in adding this as the original line only prevents a test from executing anything in runtime, which captures well the idea of parse failures.

Although, I think we can work in a way to get this shipping.

One specific thing that I don't like here is not having any value for the called method. So my suggestion  is to extend the `harness/sta.js` file defining the function preventing the runtime evaluation.

The advantage here is to requiring more setup for those importing Test262. This change would also require no effort for other consumers of Test262 without requiring any immediate action to adapt to the new format.

```js
function failIfCodeGetsEvaluated() {
}
```

This way we could guarantee the throw is always being called by default and other test suites importing Test262 can modify the imported harness file.

WDYT?

An alternative is setting this in the `harness/assert.js` extending the assert object. I prefer the first option, thou.

---

Another thing I'd like to suggest is finding a function name that represents a shorter phrase. `$DONOTEVALUATE` might do the trick.

mathiasbynens:
@leobalter I completely agree, and I actually meant to do exactly that but forgot 😅

I've applied your suggestion in `harness/sta.js` and updated `test/harness/sta.js` accordingly. Thanks!

mathiasbynens:
I have no preference at all when it comes to the function name. So far, we've had the following suggestions:

- `failIfCodeGetsExecuted`
- `failIfCodeGetsEvaluated`
- `failfNoEarlyErrorThrown`
- `$DONOTEVALUATE`

Does anyone have a strong preference? Happy to make the change.

mathiasbynens:
cc @jugglinmike who might have an opinion

jugglinmike:
I oppose this change because it's designed to allow implementers to relax test expectations beyond a point which has practical significance.

The expected behavior of failing to parse is fundamentally different than parsing, executing, and throwing a runtime error. For a programmer trying to write portable code, there is no way to accommodate both behaviors. From that perspective, parsing, executing and *not* throwing an error is equally incorrect. That's why I don't see the value in helping implementers make a distinction.

In other words, [the test annotation `FAIL_PARSE_ONLY`](https://github.com/tc39/test262/issues/1634) seems no better than a more generic annotation of `FAIL`. By endorsing this use case, Test262 will be helping folks mask conformance issues and possibly de-prioritize work along those lines.

mathiasbynens:
> I oppose this change because it's designed to allow implementers to relax test expectations beyond a point which has practical significance.

Interestingly, the same could be said about [any harness function](https://github.com/tc39/test262/blob/master/CONTRIBUTING.md#test-environment). One could override `assert` to be a no-op, for example. So why is your line of thinking not applied elsewhere? We use harness functions as opposed to hardcoded expressions for every other common thing in Test262; what makes this the exception?

> The expected behavior of failing to parse is fundamentally different than parsing, executing, and throwing a runtime error.

We're in agreement.

> For a programmer trying to write portable code, there is no way to accommodate both behaviors.

I don't think programmers are trying to accommodate both behaviors.

> From that perspective, parsing, executing and not throwing an error is equally incorrect. That's why I don't see the value in helping implementers make a distinction.

Imagine the following. A developer makes a mistake, and writes some code that per spec results in an early error.

1. In a bug-free engine, the code throws an early error.
1. In V8 it throws a runtime error instead (known issue).
1. In another engine with a different bug, no exception is thrown at all. Instead, the broken code has _some_ non-standard behavior.

The developer then tests their code in one of those engines to verify its behavior. Here's what happens in each scenario:

1. The early error warns the developer that something is wrong. They go back and fix their code. Yay!
1. The runtime error warns the developer that something is wrong. They go back and fix their code. Yay!
1. The developer doesn't see any exceptions, assumes that the non-standard behavior they're witnessing is shared by other engines, and deploys the code to production. They later get bug reports saying their app is broken in other engines.

We're in agreement that the engines in scenarios 2 and 3 are both wrong, per spec. However, I *strongly* disagree with your assessment that "parsing, executing and not throwing an error is equally incorrect", which implies that scenario 2 and 3 are just as bad. That's a very binary way of thinking about it.

I think there is a *major* difference between scenario 2 and 3. Scenario 3 presents a much bigger compatibility risk.

As such, I believe there is value in detecting engine regressions from 2 to 3 in addition to regressions from 1 to 2. But _that_ particular discussion is outside the scope of this PR, or even this repository.

(By the way, scenario 3 is not a hypothetical situation. Here's an example of such a bug: https://github.com/Microsoft/ChakraCore/issues/4368.)

-----------

> In other words, the test annotation `FAIL_PARSE_ONLY` seems no better than a more generic annotation of `FAIL`.

It is more precise, in that it enables distinguishing between situation 2 and 3, and thereby clones negative tests that we currently fail into regression tests. It seems weird to oppose improved V8 test coverage.

But again, we're talking about a V8-specific flag here that doesn't concern the Test262 project. This is off-topic, and orthogonal to this PR.

> By endorsing this use case, Test262 will be helping folks mask conformance issues and possibly de-prioritize work along those lines.

This is false. No one is trying to "mask" conformance issues. I'm trying to double the usefulness of every single negative test in Test262 in order to avoid further regressions from the status quo in V8.

The decision made in this PR also won't affect any prioritization for fixing the early error issues. I've stated before that these are unlikely to get fixed anytime soon, if ever. Merging or rejecting this PR won't change that.

jugglinmike:
> So why is your line of thinking not applied elsewhere?

Consumers could also remove all the text from a test prior to evaluating :P The
difference between arbitrary alterations and testing modes we support comes
down to policy, documented via the interpreting guidelines. If we decide to
move forward with this patch, I'll insist that we extend that document with
instructions for replacing the `$DONOTEVALUATE` with a no-op function.

This isn't a misguided attempt to force anyone's hand. I want to be sure that
when we change a policy (in this case, refactoring a pattern with an
intentional design), we do so with good reason. Here are the advantages of a
literal `throw` statement:

- Clarity for readers
- Allow negative syntax tests to be interpreted without injecting any code,
  e.g. `sta.js` (this is not yet endorsed by INTERPRETING.md, but I would like
  to do that to help folks optimize)
- Potential to catch faulty dead code removal (ensure consumers like Prepack or
  Uglify validate the code which follows the `throw` statement and report a
  parsing failure rather than aggressively remove it)

I won't claim that any of these are giant benefits. The reason I'm pushing back
is that I see *some* value there, and I am still unconvinced that there is
greater value in V8's current behavior.

> 3. The developer doesn't see any exceptions, assumes that the non-standard
>    behavior they're witnessing is shared by other engines, and deploys the
>    code to production. They later get bug reports saying their app is broken
>    in other engines.

I agree that this result is the most concerning. To me, the mapping from
scenario to result is not so cut-and-dry. Here are a few other paths to that
result:

2. In V8:
  a. The literal exists outside of a hot code path, so the runtime error is not
     thrown during the developer's testing
  b. The runtime error is caught by the application code and handled in some
     inappropriate way

These are the conditions that lead me to consider "fail to parse" and "execute
and throw" fundamentally different.

mathiasbynens:
> Here are the advantages of a literal `throw` statement:
>
> - Clarity for readers

IMHO a descriptive and easily searchable (and documented) name like `$DONOTEVALUATE` is just as clear.

> - Allow negative syntax tests to be interpreted without injecting any code, e.g. sta.js (this is not yet endorsed by `INTERPRETING.md`, but I would like to do that to help folks optimize)

No other Test262 tests can run without a harness. It doesn't seem particularly valuable to make negative tests the exception here.

> I am still unconvinced that there is greater value in V8's current behavior.

No one is making that argument. I'm arguing that there is value in increasing V8's test coverage to ensure we don't deviate further from the spec.

> These are the conditions that lead me to consider "fail to parse" and "execute and throw" fundamentally different.

No one is arguing that they aren't different. I'm confused as to why this keeps coming up.

----

> If we decide to move forward with this patch, I'll insist that we extend that document with instructions for replacing the `$DONOTEVALUATE` with a no-op function.

[Done](https://github.com/tc39/test262/pull/1801/commits/c4a46754e73451e5256c0e912d52590b1f75975e).

mathiasbynens:
Regardless of the outcome, it would be great to get a decision on this sooner rather than later. Rebasing this large patch is quite painful and I've been through it a couple of times already :dizzy_face:

ajklein:
For what it's worth, as one of the folks responsible for prioritizing (or not) V8 work on spec compliance, I'll echo the fact that this behavior in V8 isn't likely to change anytime soon, whether this PR is merged or not. 

So for me the question comes down to whether there's really zero difference in spec (non-)compliance between V8's behavior and an implementation that happily executed a function with, e.g., a malformed regexp (what would such a regexp even do, I wonder?). My take is that there's non-zero value in having V8 run these tests, to keep us more compliant. And stylistically this seems like a very small price to pay (literal `throw` statements vs `$DONOTEVALUATE`).

rwaldron:
@bakkot how does this change impact your use cases?

bakkot:
@rwaldron It doesn't much affect them either way, I think. For parser tests it's irrelevant (since they don't evaluate the files), and for the web runner I can [prepend my own throw statement](https://github.com/bakkot/test262-web-runner/blob/ae0694ff0501f4c600a0565b260039fc95904f38/main.js#L351).

Any test runners which can modify test files before evaluation can trivially adapt to this either way. As such I think it only really affects test runners which drive engines which read files themselves (like engine CLIs).

mathiasbynens:
Follow-up patch that makes this change for tests that were added since this patch was created: https://github.com/tc39/test262/pull/1957

PR #1805, opened 2018-09-28

ryzokuken:
Increase coverage for the Intl.NumberFormat class contructor by adding
tests for checking that it calls ToObject on the options argument.

### DIFF
--- a/test/intl402/NumberFormat/constructor-options-toobject.js
+++ b/test/intl402/NumberFormat/constructor-options-toobject.js
@@ -15,7 +15,9 @@ info: |
 const toObjectResults = [
   [true, new Boolean(true)],
   [42, new Number(42)],
-  ['foo', new String('foo')]
+  ['foo', new String('foo')],
+  [{}, {}],
+  [Symbol(), Object(Symbol())]
 ];
 
 // Test if ToObject is used to convert primitives to Objects.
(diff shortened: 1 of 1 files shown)


test/intl402/NumberFormat/constructor-options-toobject.js
```diff
@@ -0,0 +1,32 @@
...
+  11.1.2 InitializeNumberFormat
+
+  3.a. Let options be ? ToObject(options).
+---*/
+
+const toObjectResults = [
+  [true, new Boolean(true)],
+  [42, new Number(42)],
+  ['foo', new String('foo')]
+];
```
gsathya: What about Symbol? Object?

ryzokuken: Objects are returned as-is by `ToObject`, so I didn't think that'd create much of a difference. Symbols don't have a primitive (right?) so it'll just be making a new Symbol like the one that was passed (plus also no `new`), so I thought that won't make a huge difference as well.

If you had suggestions about how to test for them in a useful way, I'd love to add things to this but I thought it would be pointless to compare results between (a) `{}` and `{}` (b) `Symbol('foo')` and `Symbol(Symbol('foo'))`

ryzokuken:
@gsathya Done! PTAL.

PR #1806, opened 2018-09-28

ryzokuken:
Increase coverage for the Intl.NumberFormat class contructor by adding
tests for checking that it calls ToObject on the locales argument.


test/intl402/NumberFormat/constructor-locales-tooptions.js
```diff
@@ -0,0 +1,28 @@
...
+description: >
+  Tests that Intl.NumberFormat contructor converts the locales argument
+  to an object using `ToObject` (7.1.13).
+info: |
+  9.2.1 CanonicalizeLocaleList
+
+  4.a. Let O be ? ToObject(locales).
+---*/
+
+const toObjectResults = [[true, new Boolean(true)], [42, new Number(42)]];
```
gsathya: What about Symbol? Object?

ryzokuken: Objects are returned as-is by `ToObject`, so I didn't think that'd create much of a difference. Symbols don't have a primitive (right?) so it'll just be making a new Symbol like the one that was passed (plus also no `new`), so I thought that won't make a huge difference as well.

If you had suggestions about how to test for them in a useful way, I'd love to add things to this but I thought it would be pointless to compare results between (a) `{}` and `{}` (b) `Symbol('foo')` and `Symbol(Symbol('foo'))`

gsathya: I don't think it's pointless at all. Each of these primitives take different code paths in the engine.

ryzokuken: @gsathya in that case, I'll add the two cases!

ryzokuken:
@gsathya Done! PTAL.

PR #1810, opened 2018-09-29

ryzokuken:
Increase coverage for the Intl.NumberFormat constructor by adding tests
checking if appropriate default values are used.


gsathya:
As discussed offline, I don't think this is very useful as it's not really testing if `ObjectCreate(null)` or `[]` is being used here. Anyway I'll defer to the test maintainers (@rwaldron) to decide if this is worthwhile

PR #1812, opened 2018-10-01
intl: increase coverage from the NumberFormat constructor

ryzokuken:
Improve coverage for the Intl.NumberFormat constructor by testing that
it accepts Array-like objects and handles them properly.

/cc @gsathya @rwaldron @littledan @anba

### DIFF
--- a/test/intl402/NumberFormat/constructor-locales-arraylike.js
+++ b/test/intl402/NumberFormat/constructor-locales-arraylike.js
@@ -11,7 +11,13 @@ description: >
 const actual = Intl.NumberFormat({
   length: 1,
   1: 'en-US'
-});
-const expected = Intl.NumberFormat(['en-US']);
+}).resolvedOptions();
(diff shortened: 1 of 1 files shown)


test/intl402/NumberFormat/constructor-locales-arraylike.js
```diff
@@ -0,0 +1,23 @@
...
+/*---
+esid: sec-initializenumberformat
+description: >
+  Tests that the Intl.NumberFormat constructor accepts Array-like values for the
+  locales argument and treats them well.
+---*/
+
+const actual = Intl.NumberFormat({
+  length: 1,
+  1: 'en-US'
```
littledan: I'm surprised that this is 1 rather than 0. I wonder if we're actually falling back to the system locale when running this test (so it might fail on some computers).

ryzokuken: @littledan this does seem to be a silly mistype. I'll make a PR fixing this ASAP.

PR #1813, opened 2018-10-03
remove "invalid" from invalid date of "co" and "nu"

FrankYFTang:
I see nowhere in https://www.ecma-international.org/ecma-402/ mention "cu" and "nu" cannot use "invalid" as value. I think we should remove this from the test. "invalid" should be treated as a valid value for the -u-co- and -u-nu- extension in locale. This is different from "standard" and "search" for "co" and "native", "traditio", or "finance" for "nu" because those are explicitly mentioned in the spec.

### DIFF
--- a/test/intl402/Collator/ignore-invalid-unicode-ext-values.js
+++ b/test/intl402/Collator/ignore-invalid-unicode-ext-values.js
@@ -24,7 +24,7 @@ var defaultLocale = defaultOptions.locale;
 var defaultSortedArray = testArray.slice(0).sort(defaultCollator.compare);
 
 var keyValues = {
-    "co": ["standard", "search", "invalid"],
+    "co": ["standard", "search"],
     "ka": ["noignore", "shifted", "invalid"],
     "kb": ["true", "false", "invalid"],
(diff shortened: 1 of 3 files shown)


FrankYFTang:
@jungshik  - could you help me to get the right person to review this? Thanks

gsathya:
cc @littledan @zbraniecki @anba

rwaldron:
> "cu" and "nu" cannot use "invalid" as value. 

This appears to be removing it for "co": https://github.com/tc39/test262/pull/1813/files#diff-2e65889a85405368247c0d3c52c8791eR27

Can you cite the section of https://www.ecma-international.org/ecma-402/ to which you're referring? I'm unable to find it.

FrankYFTang:
sorry for my typo. I indeed mean "nu" and "co". "cu" is a different beast.

littledan:
For future reference, please use https://tc39.github.io/ecma402/ to access the Intl spec.

I think we're talking about a somewhat vague part of the specification, which I'd be more than happy to tighten up. See https://github.com/tc39/ecma402/issues/113 for previous discussion.

PR #1816, opened 2018-10-03
Fix "undefiend" typo in unreserved-words.js

amccreight:
I'm not an expert on this, but I'm pretty sure this is supposed to be "undefined" not "undefiend".

### DIFF
--- a/test/language/reserved-words/unreserved-words.js
+++ b/test/language/reserved-words/unreserved-words.js
@@ -99,7 +99,7 @@ var typeid = 1;
 // u
 var uint = 1;
 var unchecked = 1;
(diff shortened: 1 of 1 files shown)


mathiasbynens:
This was not a typo. In strict mode, assigning to `undefined` is supposed to throw a `TypeError`. Please revert this asap.

mathiasbynens:
PR: https://github.com/tc39/test262/pull/1818

leobalter:
@mathiasbynens I did a full revert as I want to preserve the original content rather than just removing the line.

Maybe we should add some inline comment that typo is intentional?

PR #1820, opened 2018-10-03
intl: increase coverage for the Intl.NumberFormat constructor

ryzokuken:
Increase coverge for the Intl.NumberFormat constructor by testing that
Get(O, P) and ToString(arg) are properly called.

/cc @gsathya @rwaldron @littledan @anba

### DIFF
--- a/test/intl402/NumberFormat/constructor-locales-get-tostring.js
+++ b/test/intl402/NumberFormat/constructor-locales-get-tostring.js
@@ -1,0 +1,41 @@
+// Copyright (C) 2018 Ujjwal Sharma. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-initializenumberformat
+description: >
+  Tests that Get(O, P) and ToString(arg) are properly called within the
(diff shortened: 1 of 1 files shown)


test/intl402/NumberFormat/constructor-locales-get-tostring.js
```diff
@@ -0,0 +1,41 @@
...
+    actualLookups.push(prop);
+    return Reflect.get(...arguments);
+  }
+};
+
+const proxyLocales = new Proxy(locales, handlers);
+
+const nf = new Intl.NumberFormat(proxyLocales);
+
+actualLookups.forEach(lookup => assert(expectedLookups.indexOf(lookup) != -1));
```
gsathya: This should be the other way around. You should map over expectedLookups, not actualLookups because if the proxy trap didn't get called the test would still pass as it would map over an empty actualLookups array.

ryzokuken: Oh, I missed that! I'll make a PR making amends. Thanks.

Issue #1817, opened 2018-10-03
Correct `features` syntax in frontmatter


mathiasbynens:
I was doing a Test262 roll for V8 and had lots of new failing tests because of this issue.

https://github.com/tc39/test262/blob/7e65999731cb6cd6fab10a92e8b7bd39a6b0793c/test/language/expressions/class/dstr-private-gen-meth-dflt-ary-init-iter-no-close.js#L7-L8

This `features` value is spread across multiple lines, which https://github.com/tc39/test262/blame/master/tools/packaging/parseTestRecord.py doesn’t seem to handle very well. It seems to ignore the `features` altogether in this case.

If you need to spread it across multiple lines, keep in mind that YAML offers this syntax:

```
features: 
  - Symbol.iterator
  - class
  - class-methods-private
  - generators
  - destructuring-binding
  - default-parameters
```

There are three options here:

1. stick to the `features: [foo, bar]` syntax on a single line
1. update `tools/packaging/parseTestRecord.py` to handle the `features: [foo, bar]` syntax spread across multiple lines
1. use the multiline syntax (and verify that `tools/packaging/parseTestRecord.py` supports it)

Ref. https://github.com/tc39/test262/pull/1708 and others.

mathiasbynens:
These tests are affected:

```
test/language/statements/class/dstr-private-gen-meth-dflt-ary-init-iter-close.js test/language/statements/class/dstr-private-gen-meth-dflt-ary-init-iter-no-close.js test/language/statements/class/dstr-private-gen-meth-dflt-ary-ptrn-rest-id-exhausted.js test/language/statements/class/dstr-private-gen-meth-static-dflt-ary-init-iter-close.js test/language/statements/class/dstr-private-gen-meth-static-dflt-ary-init-iter-no-close.js test/language/statements/class/dstr-private-gen-meth-static-dflt-ary-ptrn-rest-id.js test/language/statements/class/dstr-private-gen-meth-static-dflt-obj-ptrn-rest-getter.js test/language/statements/class/dstr-private-gen-meth-static-dflt-obj-ptrn-rest-skip-nonmerable.js test/language/statements/class/dstr-private-gen-meth-static-dflt-obj-ptrn-rest-val-obj.js
```

leobalter:
> If you need to spread it across multiple lines, keep in mind that YAML offers this syntax (...)

[...]

PR #1821, opened 2018-10-03
Generation: fix yaml frontmatter writer to eliminate errant newlines from yaml.dump(). Fixes gh-1817

rwaldron:


test/language/statements/class/dstr-private-gen-meth-static-dflt-obj-ptrn-rest-val-obj.js
```diff
@@ -4,8 +4,7 @@
 /*---
 description: Rest object contains just unextracted data (private static class expression generator method (default parameter))
 esid: sec-runtime-semantics-bindingclassdeclarationevaluation
-features: [object-rest, class, class-static-methods-private, generators, destructuring-binding,
-  default-parameters]
+features: [object-rest, class, class-static-methods-private, generators, destructuring-binding,  default-parameters]
```
mathiasbynens: Nit: this leaves double spaces (before `default-parameters`). Maybe adjust the regex to trim `\n` followed by `\s*` into a single space instead?

tools/generation/lib/template.py
```diff
@@ -154,7 +154,7 @@ def _frontmatter(self, case_filename, case_values):
         features += self.attribs['meta'].get('features', [])
         features = list(OrderedDict.fromkeys(features))
         if len(features):
-            lines += ['features: ' + yaml.dump(features).strip()]
+            lines += ['features: ' + re.sub('\n|\r', '', yaml.dump(features).strip())]
```
mathiasbynens: My earlier suggestion would be `sub('\n\s*', ' ')`.

Rationale:

- Lone CR line endings are not really a thing although CRLF is. That is, I don’t think you want to match lone `\r`. 
- `.editorconfig` already enforces LF line endings. https://github.com/tc39/test262/blob/ff8b10c448ab0b884deac4e6e5818ed84ab8ecd9/.editorconfig#L6

rwaldron:
@mathiasbynens as soon as you sign off on this, we'll get it landed

leobalter:
I pushed a quick fixup for other missing tests to generate. This should get the CI green.

mathiasbynens:
> I pushed a quick fixup for other missing tests to generate. This should get the CI green.

It worked!

tools/generation/lib/template.py
```diff
@@ -154,7 +154,7 @@ def _frontmatter(self, case_filename, case_values):
         features += self.attribs['meta'].get('features', [])
         features = list(OrderedDict.fromkeys(features))
         if len(features):
-            lines += ['features: ' + yaml.dump(features).strip()]
+            lines += ['features: ' + re.sub('\n\s*', ' ', yaml.dump(features).strip())]
```
mathiasbynens: As a side note, we should probably do the same thing for all the other stuff that outputs `foo: [bar, baz]`, i.e. `flags` and `includes` (just a few lines below) and possibly others.

PR #1822, opened 2018-10-03
Fix duplicate paths

leobalter:


src/generators/default/class-decl-private-method.template
```diff
@@ -1,7 +1,7 @@
 // Copyright (C) 2018 Bloomberg LP. All rights reserved.
 // This code is governed by the BSD license found in the LICENSE file.
 /*---
-path: language/statements/class/gen-method-
+path: language/statements/class/gen-private-method-
```
leobalter: the previous path was a duplicate for another template file.

PR #1828, opened 2018-10-04
intl: increase coverage for the NumberFormat constructor

ryzokuken:
Increase the coverage for the constructor for Intl.NumberFormat by
ensuring that HasProperty(O, Pk) is properly called.

/cc @gsathya @rwaldron

### DIFF
--- a/test/intl402/NumberFormat/constructor-locales-hasproperty.js
+++ b/test/intl402/NumberFormat/constructor-locales-hasproperty.js
@@ -34,6 +34,6 @@ const proxyLocales = new Proxy(locales, handlers);
 const nf = new Intl.NumberFormat(proxyLocales);
 
 assert.sameValue(actualLookups.length, locales.length);
-for (index in actualLookups) {
+for (let index in actualLookups) {
   assert.sameValue(actualLookups[index], String(index));
 }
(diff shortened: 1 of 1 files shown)


test/intl402/NumberFormat/constructor-locales-hasproperty.js
```diff
@@ -0,0 +1,39 @@
...
+    return Reflect.has(...arguments);
+  }
+};
+
+const proxyLocales = new Proxy(locales, handlers);
+
+const nf = new Intl.NumberFormat(proxyLocales);
+
+assert.sameValue(actualLookups.length, locales.length);
+for (index in actualLookups) {
```
rwaldron: `for (index in actualLookups) {` => `for (let index in actualLookups) {` (or it will fail in strict mode code)

ryzokuken:
@rwaldron done!

test/intl402/NumberFormat/constructor-locales-hasproperty.js
mathiasbynens: should be `const` to match the style in the rest of the file

ryzokuken: @rwaldron @mathiasbynens should I make another PR to change it, or let it be?

PR #1798, opened 2018-09-26
Intl.Locale: constructor-options-*.js fixups

rwaldron:
Improves assertion messages.

A sample: 

```diff
FAIL test/intl402/Locale/constructor-options-language-grandfathered.js (default)
-  ...
+  new Intl.Locale("nb", {language: "no-bok"}) throws RangeError  ...

FAIL test/intl402/Locale/constructor-options-language-invalid.js (default)
-   is an invalid language option value ...
+  new Intl.Locale("en", {language: ""}) throws RangeError  ...

FAIL test/intl402/Locale/constructor-options-region-invalid.js (default)
-   is an invalid region option value ...
+  new Intl.Locale("en", {region: ""}) throws RangeError  ...

FAIL test/intl402/Locale/constructor-options-region-valid.js (default)
-  new Intl.Locale('en', options).toString() equals the value of en-FR...
+  new Intl.Locale('en', {region: "FR"}).toString() returns "en-FR" ...

FAIL test/intl402/Locale/constructor-options-script-invalid.js (default)
-   is an invalid script option value ...
+  new Intl.Locale("en", {script: ""}) throws RangeError ...

FAIL test/intl402/Locale/constructor-options-script-valid.js (default)
-  new Intl.Locale('en', options).toString() equals the value of "en-Null"...
+  new Intl.Locale("en", {script: "null"}).toString() returns "en-Null" ...

FAIL test/intl402/Locale/constructor-options-throwing-getters.js (default)
-  Exception from language getter should be propagated...
+  new Intl.Locale("en", {get language() {throw new CustomError();}}) throws CustomError ...
```

PR #1835, opened 2018-10-06
compare "numeric" as boolean instead of string

FrankYFTang:
https://tc39.github.io/proposal-intl-locale/#sec-Intl.Locale
36. If relevantExtensionKeys contains "kn", then
 a. Set locale.[[Numeric]] to ! SameValue(r.[[kn]], "true").
and then
https://tc39.github.io/proposal-intl-locale/#sec-Intl.Locale.prototype.numeric
Intl.Locale.prototype.numeric 
3. Return loc.[[Numeric]].

so it should be a boolean, not a string of "true" or "false"


FrankYFTang:
This was broke by https://github.com/tc39/test262/pull/1798/ 
@leobalter @rwaldron @Ms2ger

FrankYFTang:
Also notice https://github.com/tc39/test262/blob/master/test/intl402/Locale/getters.js it is correctly comparing with true and false not "true" and "false".

FrankYFTang:
The change is needed to reflect the change in the spec https://github.com/tc39/proposal-intl-locale/commit/cf84ee38db6af3c02feda5010ad32735bb558e99#diff-3540caefa502006d8a33cb1385720803

FrankYFTang:
this is because the test currently reflect the old spec, without changing with https://github.com/tc39/proposal-intl-locale/issues/55

littledan:
Thanks for the fix, Frank. Has anyone followed up with the remaining changes? @Ms2ger?

PR #1840, opened 2018-10-07
Refactor variable declaration tests for parsers

jugglinmike:
The tests for the parsing of variable declarations were expressed using
eval. This made the tests more complex than necessary and also prevented
the tests from providing value to ECMAScript parsers.

---

Organizing these tests called for a bunch of different operations. I've
tried to make my process/reasoning visible via the commit history.


test/language/statements/variable/arguments-strict-list-final-init.js
```diff
@@ -3,13 +3,16 @@
...
+  phase: parse
+  type: SyntaxError
 ---*/
 
+throw "Test262: This statement should not be evaluated.";
 
-assert.throws(SyntaxError, function() {
-    eval('function foo() { var a, arguments = 42;}');
-});
+var a, arguments = 42;
```
leobalter: these changes are great, but in this case I'd keep a test with the function declarations. I wonder if we could double these as a test file with and another without the function. WDYT?

rwaldron: @jugglinmike ping ^^

PR #1842, opened 2018-10-07
Refactor compound assignment tests for parsers

jugglinmike:
The tests for the parsing of compound assignment expressions were
expressed using eval. This made the tests more complex than necessary
and also prevented the tests from providing value to ECMAScript parsers.

Remove the use of eval and instead express the expectations with
literal source text.


jugglinmike:
See also https://github.com/bterlson/eshost/pull/70

jugglinmike:

PR #1848, opened 2018-10-09
test: syntax error if superproperty private name access

jbhoosreddy:
Tracking issues #1343 

Summary of changes:
1. Add test for super property private name early syntax error

### DIFF
--- a/src/class-elements/super-private-access-invalid.case
+++ b/src/class-elements/super-private-access-invalid.case
@@ -13,8 +13,8 @@ info: |
   PrivateName ::
     # IdentifierName
 
-  SuperProperty[Yield, Await]:
-    super[Expression[+In, ?Yield, ?Await]]
+  SuperProperty:
+    super[Expression]
     super.IdentifierName
 
 template: syntax/invalid
--- a/src/class-elements/super-private-access-invalid.case
+++ b/src/class-elements/super-private-access-invalid.case
@@ -1,0 +1,34 @@
+// Copyright (C) 2018 Jaideep Bhoosreddy (Bloomberg LP). All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: It is syntax error if PrivateName IdentifierName is accessed on SuperProperty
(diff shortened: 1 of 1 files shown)


jbhoosreddy:
@leobalter @rwaldron @littledan @mkubilayk for review

jbhoosreddy:
@leobalter @rwaldron for review?

leobalter:
Rick is off this week due to family obligations. I don't have any time assigned this week to review PRs and I'm pretty busy with other things. We should be able to review this one next week.

rwaldron:
When I generate these tests, I get the following: 

```js

class C extends B
{
  method() {
  super.#x();
  }
}
```

Which _will_ pass by throwing a `SyntaxError` exception, but I think it's a false positive: 

```
Uncaught SyntaxError: Invalid private field '#x'
```

The reason I believe this is a false positive is due this being the same exception that occurs for this: 

```js
class C extends B
{
  method() {
  this.#x();
  }
}
```

```
Uncaught SyntaxError: Invalid private field '#x'
```

So, when I add a private property `#x`, this code should throw a `SyntaxError` exception regarding the occurrence of `super.#x`: 

```js
class C extends B
{
  #x = 1;
  method() {
    super.#x;
  }
}
```

However, it does not! Take a look: 

```
$ v8 --harmony_private_fields
V8 version 7.2.9
d8> throw "Test262: This statement should not be evaluated.";class C extends B {#x = 1;method() {super.#x;}}
(d8):1: Test262: This statement should not be evaluated.
throw "Test262: This statement should not be evaluated.";class C extends B {#x = 1;method() {super.#x;}}
^
```

rwaldron:
I actually don't think this: 

```js
class B {
#x() {}
}
```

...is even necessary. Since the test is just "expect `super.#x()` to result in a SyntaxError", there doesn't even need to be a class declaration for `B`. 

I no longer believe that the failure is a false positive.

jbhoosreddy:
@rwaldron based on your last comment is there something I needed to fix in my PR?

PR #1849, opened 2018-10-09
test: private method usage: calling private method

jbhoosreddy:
Tracking issue: #1343 

Summary of changes:
1. Add test for private method usage

> `Calling a private method`

### DIFF
--- a/src/class-elements/private-field-usage.case
+++ b/src/class-elements/private-field-usage.case
@@ -1,0 +1,26 @@
+// Copyright (C) 2018 Jaideep Bhoosreddy (Bloomberg LP). All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: PrivateName CallExpression usage (private field)
+info: |
+  Updated Productions
+
+  CallExpression[Yield, Await]:
+    CoverCallExpressionAndAsyncArrowHead[?Yield, ?Await]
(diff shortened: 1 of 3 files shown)


jbhoosreddy:
@leobalter @rwaldron This one needs more work on both frontmatter and coverage. But I wanted to get some initial 👀 on it.

1. Can't find the normative spec text for the description in the spec.
2. Also should coverage include all variants of private methods?

jbhoosreddy:
Also, wanted to use this space to discuss my next PR
> Reading a private method separately from calling it
>  + All instances of the class share the same function identity for the private method

Thoughts on testing function identity? I think I know it but just wanted to verify. cc @mkubilayk we can catch up about this tomorrow.

rwaldron:
If these are just valid syntax tests then there shouldn't be any assertions (there may be older tests that do this, but those just need to be split into "syntax only" and "runtime semantics only"). If you want to do runtime semantics tests that exercise all of the grammar variants, then you'll need to use `src/class-elements/productions` (`template: productions`).

leobalter:
I'm working on a fix for these cases.

The current cases are definitely not calling the presented assertions. They are also a simplified version of the rs-private-getter*.case / rs-private-method*.case files in the same folder.

leobalter:
fixed on #1919

PR #1850, opened 2018-10-09
test: Add tests for private methods and accessors productions

mkubilayk:
Adds coverage for semantics of private methods and private accessors according to the test plan posted in #1343.

> - Semantics > Private method constructs
>   - Private ordinary methods
>   - Private accessors

I used private fields as an intermediate storage to test the behaviour of private methods. Let me know if that makes the tests too tightly coupled. 

---
cc: @littledan @jbhoosreddy

### DIFF
--- a/src/class-elements/rs-private-getter-alt.case
+++ b/src/class-elements/rs-private-getter-alt.case
@@ -54,6 +54,7 @@ features: [class-methods-private]
 ---*/
 
 //- fields
+#$_; #__; #\u{6F}_; #℘_; #ZW_‌_NJ_; #ZW_‍_J_;
 get #$() {
   return this.#$_;
 }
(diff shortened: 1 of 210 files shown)


littledan:
IMO it's not too tightly coupled, as the private methods proposal builds on the private fields proposal (explicitly, the specification says this).

src/class-elements/rs-private-getter-alt.case
```diff
@@ -0,0 +1,108 @@
...
+get #ZW_‌_NJ() {
+  return this.#ZW_‌_NJ_;
+}
+get #ZW_‍_J() {
+  return this.#ZW_‍_J_;
+}
+
+//- privateinspectionfunctions
+$(value) {
+  this.#$_ = value;
```
rwaldron: No such private field exists. You'll need to this: 

```
#$_; #__; #\u{6F}_; #℘_; #ZW_‌_NJ_; #ZW_‍_J_;
```
...under the `//- fields` label.

Or you can just `return 1;` from all of the getters and remove these assignments. 

Note that this applies to all of the case files in this changeset. I recommend running `./make.py build` to verify the resulting test material is valid.

mkubilayk:
Added missing field declarations.

mkubilayk:
@leobalter: Sure, no problem! Thanks for taking the time to review this. I pushed the generated tests.

Issue #1164, opened 2017-08-03
Add coverage for dynamic import


leobalter:
currently on stage 3.

https://github.com/tc39/proposal-dynamic-import

# Coverage Checklist for `import()`

## Default Templates

_general places it can be found as a CallExpression_

- [x] top level
- [x] consise body of an ArrowFunction
- [x] function body of an ArrowFunction
- [x] consise body of an __Async__ ArrowFunction
- [x] function body of an __Async__ ArrowFunction
- [x] Used in async generators
- [x] await position of an async function
- [x] return await position of an async function
- [x] nested in an async function
- [x] labeled blocks
- [x] blocks
- [x] do while
- [x] else braceless
- [x] else block
- [x] if braceless
- [x] if block
- [x] while block
- [x] inside `with`

### Syntax

- [x] assignment expression is not optional `assignment-expr-not-optional.case`
- [x] empty string is a valid assignment expression (not a syntax error) `empty-str-is-valid-assign-expr.case`
- [x] _Forbidden Extensions: ImportCall must not be extended._ (Only one assignment expression allowed)
    - [x] no trailing comma, it's not an arguments list (#1832)
    - [x] no other expressions (#1832)

### Common usage

- [x] indirect update default binding `eval-gtbndng-indirect-update-dflt.case`
- [x] indirect update named binding `eval-gtbndng-indirect-update-dflt.case`
- [x] Assignment Expression evaluations. Most can be used as syntax tests only, e.g. exprs evaluating to a boolean value) or error catching
    - [x] ConditionalExpression
        - [x] Primary Expressions on the import Assign. Expr. position (?)
        - [x] Cover Parenthesis expressions
        - [x] ...
    - [x] YieldExpression
        - [x] Check yield usage
    - [x] ArrowFunction
    - [x] AsyncArrowFunction
    - [x] LeftHandSideExpression = AssignmentExpression
    - [x] LeftHandSideExpression AssignmentOperator AssignmentExpression
- [x] ToString on the AssignmentExpression value `Let specifierString be ToString(specifier).`
    - [ ] `If AbruptRejectPromise(specifierString, promiseCapability).`
- [ ] nest `import()` in CallExpressions
    - [x] `import(import(...))` (#1832)
    - [ ] ... (other CallExpressions here)
- [x] Goal target can be a script code (not module code)
    - [x] in this case, it should not export anything (and it's not a SyntaxError like in static imports)
    - [ ] indirect updates
[...]

PR #1853, opened 2018-10-09
Add more cases for dynamic imports 

leobalter:
Ref #1164


leobalter:
cc @jdalton, one more round, please :)

jdalton:
I'm at the Node+JS interactive conf so my review will be slowed _(should still tackle it by Saturday)_.

leobalter:
sure thing! Enjoy the event!

test/language/module-code/dynamic-import/catch/top-level-import-catch-eval-script-code-target.js
```diff
@@ -0,0 +1,38 @@
...
+    Static Semantics: Early Errors
+
+      ModuleBody : ModuleItemList
+      - It is a Syntax Error if the LexicallyDeclaredNames of ModuleItemList containsany duplicate entries.
+      - It is a Syntax Error if any element of the LexicallyDeclaredNames of ModuleItemList also occurs in the VarDeclaredNames of ModuleItemList.
+
+---*/
+
+import('./script-code_FIXTURE.js').catch(error => {
+
```
jdalton: How does an implementation or interpretation detect that fixtures should be treated as script or module goals?

test/language/module-code/dynamic-import/catch/nested-if-import-catch-specifier-tostring-abrupt-rejects.js
```diff
@@ -0,0 +1,53 @@
...
+    8. Perform ! HostImportModuleDynamically(referencingScriptOrModule, specifierString, promiseCapability).
+    9. Return promiseCapability.[[Promise]].
+
+
+    Import Calls
+
+    Runtime Semantics: Evaluation
+
+    ImportCall : import(AssignmentExpression)
+
```
jdalton: This test means the `toString()` doesn't happen synchronously, handled at some point in another tick. Correct?

leobalter: yes! and this differentiates from a possible abrupt completion from `4. Let specifier be ? GetValue(argRef).` that will return instead of rejecting the promise.

leobalter: The rejection happens here: `7. IfAbruptRejectPromise(specifierString, promiseCapability).`

test/language/module-code/dynamic-import/catch/top-level-import-catch-eval-script-code-target.js
jdalton: The `script-code_FIXTURE` currently only evaluates _(without error in the script goal)_. Test files gave metadata to indicate `module` or script (lack of `module` specified)` goals, but I don't see FIXTURE scripts having such indicators.

jdalton: Ahhh, on further testing I get what you're saying. Sorry for the mixup.

jdalton: So script code is treated as module?
Is the limitation for `await` keyword, html  comments, and strict mode in place too?

leobalter: that's a weird thing. The script code is not exactly treated as module but it needs to create a Module Record. You can't have duplicate lexical bindings in a Module Record and that's where we get the error.

I'll check out a verification with the html comments.

leobalter:
I got more cases here improving mostly the coverage for the returned module namespace object.

src/dynamic-import/ns-define-own-property.case
```diff
@@ -0,0 +1,114 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
```
jdalton: Should the ns-xyz files be in the `/ns/` folder?

jdalton:
Other than the filename nit things look good!

Keep in mind I'm testing in the context of Node which doesn't have a script goal for scripts imported with dynamic `import()` so I'm skipping those. In Node, CJS modules imported with dynamic `import()` are done with an ESM facade around their `module.exports`.

jdalton:
> If this plan looks good to you, I'd be happy to land these as I'm already working on other parts for this feature. :)

:shipit: !

PR #1854, opened 2018-10-09
intl: fix mistyped test for array-like objects

ryzokuken:
/cc @littledan @rwaldron @gsathya


ryzokuken:
@littledan I'm close to being done with `NumberFormat` here, so I'll do a complete run-through for all the tests (including those I added) before jumping onto the next builtin. Will take care to look for mistypes. Thanks.

PR #1855, opened 2018-10-10
Intl.ListFormat: Add some tests for the localeMatcher constructor option.

Ms2ger:
This was added in https://github.com/tc39/proposal-intl-list-format/pull/25.

I don't know how to test that the option has any effect, so this just checks
that it is read and verified.


FrankYFTang:
+1

littledan:
@rwaldron @leobalter Would you mind if I merged this PR?

PR #1866, opened 2018-10-16
Intl: Add tests for table iteration order.

Ms2ger:
The order was defined in <https://github.com/tc39/ecma402/pull/279>.

### DIFF
--- a/test/intl402/Collator/prototype/resolvedOptions/order.js
+++ b/test/intl402/Collator/prototype/resolvedOptions/order.js
@@ -1,0 +1,31 @@
+// Copyright 2018 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-intl.collator.prototype.resolvedoptions
+description: Verifies the property order for the object returned by resolvedOptions().
+includes: [compareArray.js]
+---*/
+
+const options = new Intl.Collator([], {
+  "numeric": true,
+  "caseFirst": "upper",
+}).resolvedOptions();
+
+const expected = [
+  "locale",
+  "usage",
+  "sensitivity",
+  "ignorePunctuation",
+  "collation",
+];
+
+if ("numeric" in options) {
+  expected.push("numeric");
+}
+
+if ("caseFirst" in options) {
+  expected.push("caseFirst");
+}
+
+assert.compareArray(Object.getOwnPropertyNames(options), expected);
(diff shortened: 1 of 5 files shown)


test/intl402/Collator/prototype/resolvedOptions/order.js
```diff
@@ -0,0 +1,31 @@
...
+  "collation",
+];
+
+if ("numeric" in options) {
+  expected.push("numeric");
+}
+
+if ("caseFirst" in options) {
+  expected.push("caseFirst");
+}
```
rwaldron: Based on this: 

> Intl.Collator instances also have the following internal slots if the key corresponding to the name of the internal slot in Table 2 is included in the [[RelevantExtensionKeys]] internal slot of Intl.Collator:
>
> - [[Numeric]] is a Boolean value, specifying whether numeric sorting is used.
> - [[CaseFirst]] is one of the String values "upper", "lower", or "false".


There should be one test file that tests that these properties exist when the conditions are met, and one test file for when the condition is not met.

Ms2ger: That would be four tests, you mean? How would you distinguish them?

littledan: Note, the conditions here are choices an implementation can make, not different runtime conditions achievable in a single implementation.

rwaldron: > the conditions here are choices an implementation can make, 

After further reviewing the section "In Collator:" under [Annex A Implementation Dependent Behaviour
](https://tc39.github.io/ecma402/#annex-implementation-dependent-behaviour), I recognize that this test is valid, however I think a specification that allows implementations to vary in this way is a poor design. All properties listed in each "Resolved Options" table should be present on the object returned by `resolvedOptions()` with a value of `undefined` if no useful value can be determined.

littledan: I agree that the optionality is a sort of spec smell. I proposed that we make support mandatory in https://github.com/tc39/ecma402/issues/113 but I don't think I did a great job explaining what the change would really mean. Maybe I should try to push for it again.

littledan: I agree that the optionality is a sort of spec smell. I proposed that we make support mandatory in https://github.com/tc39/ecma402/issues/113 but I don't think I did a great job explaining what the change would really mean. Maybe I should try to push for it again.

PR #1874, opened 2018-10-18
Add missing includes for await tests

anba:


leobalter:
thanks for the fix, @anba

PR #1879, opened 2018-10-19
Some fixes to dynamic import + a few cosmetic changes

leobalter:
Some cosmetic changes include commenting specifiers resolution when it's not immediately set with direct strings.

Other small fixes consider runs in strict modes and removes unnecessary async generator productions.

The tests are also passing just fine on V8. I had to include the `--harmony` flag to include the `export * as ns from 'mod'` productions.

```
test262-harness --hostType=d8 --hostPath=`which v8` --hostArgs='--harmony'  'test/language/expressions/dynamic-import/**/*.js'
Ran 882 tests
882 passed
0 failed
```

### DIFF
--- a/src/dynamic-import/is-call-expression-square-brackets.case
+++ b/src/dynamic-import/is-call-expression-square-brackets.case
@@ -5,6 +5,9 @@ desc: ImportCall is a CallExpression, it can be followed by square brackets
 template: default
 ---*/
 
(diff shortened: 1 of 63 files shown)


leobalter:
Updating V8 runs everything just fine without the harmony flag.

cc @gsathya and @bmeck

PR #1884, opened 2018-10-22
Add more edgy cases for import()

leobalter:
Also rename test/language/expressions/dynamic-import/assignment-expression/


test/language/expressions/dynamic-import/for-await-resolution-and-error-agen-yield.js
```diff
@@ -0,0 +1,69 @@
...
+    assert.sameValue((await e).value.x, 39, 'e');
+
+    error = null;
+    try {
+        await f;
+    } catch (err) {
+        error = err;
+    }
+
+    assert.sameValue(error, 'foo', 'f');
```
leobalter: somehow JSC is failing on this assertion, V8 is passing. I need to check what is happening here but work time is over for today.

jdalton:
Thanks @leobalter!

I'm playing catch-up on getting `export-ns-from` tests to pass in my setup so haven't had time to do a validation run on these just yet.

test/language/expressions/dynamic-import/for-await-resolution-and-error-agen-yield.js
rwaldron: Have you resolved this?

leobalter: It seems like an error on JSC, and I don't think the test itself is wrong. I only want to isolate this case into another test to confirm my thoughts (maybe a test that imports() a failing module twice).

PR #1888, opened 2018-10-23
Add initial tests for import.meta

### LINKED ISSUE #1342: Add coverage for `import.meta`
Already on Stage 3

anba:
Fixes #1342 

For the most part syntax tests, because the properties, prototype, and extensibility state of the `import.meta` object is implementation-dependent.

### DIFF
--- a/test/language/expressions/import.meta/syntax/invalid-assignment-target-update-expr.js
+++ b/test/language/expressions/import.meta/syntax/invalid-assignment-target-update-expr.js
@@ -1,0 +1,32 @@
+// Copyright (C) 2018 André Bargull. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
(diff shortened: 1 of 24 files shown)


test/language/expressions/import.meta/syntax/invalid-assignment-target-update-expr.js
```diff
@@ -0,0 +1,32 @@
...
+esid: sec-static-semantics-static-semantics-assignmenttargettype
+description: >
+  import.meta is not a valid assignment target.
+info: |
+  Static Semantics: AssignmentTargetType
+
+    ImportMeta:
+      import.meta
+
+    Return invalid.
```
leobalter: I'm glad you're already using up to date spec text, while the spec proposal still has the [old format](https://tc39.github.io/proposal-import-meta/#sec-static-semantics-static-semantics-isvalidsimpleassignmenttarget).

PR #1892, opened 2018-10-24
Intl.Segmenter: Submit batch of V8 tests for segment().

Ms2ger:


rwaldron:
Curation of files collected by the automated process needs to be documented, ideally with a detailed commit message. 

Additionally, all of these files need the actual test source migrated: 

-  `assertEquals(expected, actual)` => `assert.sameValue(actual, expected)`
-  `assertTrue(expr)` => `assert(expr)`

Ms2ger:
Done, sorry.

Would you mind pointing me to the documentation for anything else I missed?

PR #1907, opened 2018-10-26
Fix typos; generate tests

rwaldron:


leobalter:
Needs to fix the test generation but it should be fine to land after that

PR #1912, opened 2018-10-30
Assert error when invoking special %RegExpPrototype% getters on cross-realm RegExp.prototype

gibson042:
RegExp.prototype is not a RegExp, and the algorithm only special-cases (same-realm) %RegExpPrototype%.

https://tc39.github.io/ecma262/#sec-get-regexp.prototype.source
> <code>RegExp.prototype.source</code> is an <emu-xref href="#sec-object-type" id="_ref_6226"><a href="https://tc39.github.io/ecma262/#sec-object-type">accessor property</a></emu-xref> whose set accessor function is **<emu-val>undefined</emu-val>**. Its get accessor function performs the following steps:
> 1. Let _<var>R</var>_ be the **<emu-val>this</emu-val>** value.
> 1. If <emu-xref aoid="Type" id="_ref_6227"><a href="https://tc39.github.io/ecma262/#sec-ecmascript-data-types-and-values">Type</a></emu-xref>(_<var>R</var>_) is not Object, throw a **<emu-val>TypeError</emu-val>** exception.
> 1. If _<var>R</var>_ does not have an [[OriginalSource]] internal slot, then
>    1. If <emu-xref aoid="SameValue" id="_ref_6228"><a href="https://tc39.github.io/ecma262/#sec-samevalue">SameValue</a></emu-xref>(_<var>R</var>_, <emu-xref href="#sec-properties-of-the-regexp-prototype-object" id="_ref_6229"><a href="https://tc39.github.io/ecma262/#sec-properties-of-the-regexp-prototype-object">%RegExpPrototype%</a></emu-xref>) is <emu-val>true</emu-val>, return <code>"(?:)"</code>.
>    1. Otherwise, throw a **<emu-val>TypeError</emu-val>** exception.
> 1. …

### DIFF
--- a/test/built-ins/RegExp/prototype/dotAll/this-val-invalid-obj.js
+++ b/test/built-ins/RegExp/prototype/dotAll/this-val-invalid-obj.js
@@ -16,6 +16,9 @@ features: [regexp-dotall]
 ---*/
 
 var dotAll = Object.getOwnPropertyDescriptor(RegExp.prototype, 'dotAll').get;
+var other = $262.createRealm().global;
+var otherRegExpProto = other.RegExp.prototype;
+var otherRegExpGetter = Object.getOwnPropertyDescriptor(otherRegExpProto, 'dotAll').get;
 
 assert.throws(TypeError, function() {
   dotAll.call({});
@@ -28,3 +31,11 @@ assert.throws(TypeError, function() {
 assert.throws(TypeError, function() {
   dotAll.call(arguments);
 }, 'arguments object');
+
+assert.throws(TypeError, function() {
+  dotAll.call(otherRegExpProto);
+}, 'cross-realm RegExp.prototype');
+
+assert.throws(other.TypeError, function() {
+  otherRegExpGetter.call(RegExp.prototype);
+}, 'cross-realm RegExp.prototype getter method against primary realm RegExp.prototype');
--- a/test/built-ins/RegExp/prototype/source/this-val-invalid-obj.js
+++ b/test/built-ins/RegExp/prototype/source/this-val-invalid-obj.js
@@ -13,6 +13,8 @@ info: |
 ---*/
 
 var get = Object.getOwnPropertyDescriptor(RegExp.prototype, 'source').get;
+var otherRealm = $262.createRealm().global;
+var otherRealmRegExpProto = otherRealm.RegExp.prototype;
 
 assert.throws(TypeError, function() {
   get.call({});
@@ -25,3 +27,7 @@ assert.throws(TypeError, function() {
 assert.throws(TypeError, function() {
   get.call(arguments);
 }, 'arguments object');
+
+assert.throws(TypeError, function() {
+  get.call(otherRealmRegExpProto);
+}, 'cross-realm RegExp.prototype');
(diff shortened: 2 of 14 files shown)


test/built-ins/RegExp/prototype/source/this-val-invalid-obj.js
```diff
@@ -25,3 +27,7 @@ assert.throws(TypeError, function() {
 assert.throws(TypeError, function() {
   get.call(arguments);
 }, 'arguments object');
+
+assert.throws(TypeError, function() {
+  get.call(otherRealmRegExpProto);
+}, 'cross-realm RegExp.prototype');
```
ljharb: Should we also assert that `otherRealmRegExpProto.get.call(RegExp.prototype)` throws, just to be thorough?

gibson042: I don't think so... none of the other cross-realm testing seems to do so, and a failure would be more an indictment of realm construction than of an issue with the `RegExp.prototype.source` getter anyway.

gibson042: Fair enough; I've updated the test and done the same for other special `RegExp.prototype` getters.

littledan:
BTW if all the implementations fail this test in the same way, maybe we should change the spec (I wrote this part; I won't be offended). Do you have a good understanding of the deviations from spec'd behavior here?

gibson042:
It looks like V8 passes, SpiderMonkey throws primary-realm exceptions instead of subrealm, EdgeHTML doesn't even expose getters for the magic properties (i.e., fails even the current tests), and I couldn't test JavaScriptCore.

As I mentioned [on Twitter](https://twitter.com/gibson042/status/1055265958326099968), I would like to see `RegExp.prototype`, `Date.prototype`, and every other built-in prototype predating ES 2015 restored to be instances of their respective classes as they were in ES 5.1 (albeit updated to have immutable internal state). But assuming that's off the table, we might as well keep the specified behavior of getters that privilege same-realm built-in prototypes to fake just enough of the interface provided by respective instances.

anba:
> SpiderMonkey throws primary-realm exceptions instead of subrealm

This is a general issue in SpiderMonkey when calling functions with cross-realm objects: https://bugzilla.mozilla.org/show_bug.cgi?id=1197096

test/built-ins/RegExp/prototype/dotAll/this-val-invalid-obj.js
```diff
@@ -16,6 +16,9 @@ features: [regexp-dotall]
 ---*/
 
 var dotAll = Object.getOwnPropertyDescriptor(RegExp.prototype, 'dotAll').get;
+var other = $262.createRealm().global;
+var otherRegExpProto = other.RegExp.prototype;
+var otherRegExpGetter = Object.getOwnPropertyDescriptor(otherRegExpProto, 'dotAll').get;
```
rwaldron: Please put all of the newly added cross-realm test code into a new test file. Here are helpful examples to follow: https://github.com/tc39/test262/search?p=1&q=cross-realm&unscoped_q=cross-realm


(Apply this same review feedback to all files in this PR)

gibson042: Done.

rwaldron:
@gibson042 this is great, thank you for the updates!

PR #1913, opened 2018-10-31
Fix Intl.Segmenter branding.js

rwaldron:
Based on: https://github.com/tc39/test262/blob/44fc8165d54de9c9f0128f70eda56bdcb5dde82c/implementation-contributed/v8/intl/segmenter/segment-iterator-ownPropertyDescriptor.js


FrankYFTang:
LGTM

FrankYFTang:
Is there way I can apply the right for reviewer?

FrankYFTang:
@leobalter @Ms2ger @rwaldron

PR #1916, opened 2018-11-02
In en/ko/fr 'loose' may break more than normal in new CLDR version

FrankYFTang:
@jungshik @Ms2ger @littledan  This test is based on my work in v8 but my original expectation is too restrict to icu62 and not general enough for newer version of CLDR pull int by icu63


FrankYFTang:
The fixing of https://unicode-org.atlassian.net/browse/ICU-13770 caused the different behavior between icu62 and icu63

rwaldron:
@FrankYFTang thanks for the additional insight.

PR #1922, opened 2018-11-05
Add cases for valid names in private method accessors.

leobalter:
Supersedes #1680


leobalter:
cc @ktbee

src/accessor-names/private-escape-sequence-ZWNJ.case
```diff
@@ -0,0 +1,13 @@
...
+/*---
+desc: Private IdentifierName - ZWNJ
+esid: sec-private-names-static-semantics-stringvalue
+template: private
+---*/
+
+//- declareWith
+ZW_\u200C_NJ
+//- referenceWith
+ZW_‌_NJ
```
rwaldron: Nice

PR #1925, opened 2018-11-06
Add missing `./` to relative path in dynamic import

bakkot:


test/language/expressions/dynamic-import/update-to-dynamic-import.js
```diff
@@ -19,7 +19,7 @@ features: [dynamic-import]
 ---*/
 
 async function fn() {
-    const first = await import('update-to-dynamic-import_FIXTURE.js');
+    const first = await import('./update-to-dynamic-import_FIXTURE.js');
```
leobalter: good catch, thanks!

PR #1926, opened 2018-11-06
Use travis ci to run new or modified test files on engines on each PR

leobalter:
I'm experimenting with Travis CI configuration and this PR should help.

~~It's currently WIP, so please hang on, don't merge.~~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,55 @@
-language: python
-install:
-  - python -m pip install --requirement tools/generation/requirements.txt
-  - python -m pip install --requirement tools/lint/requirements.txt
-script:
-  - ./tools/scripts/ci_build.sh
-  - ./tools/generation/test/run.py
-  - ./tools/lint/test/run.py
-  - ./tools/scripts/ci_lint.sh
-after_success:
-  - ./tools/scripts/deploy.sh
+matrix:
+  include:
+  - language: python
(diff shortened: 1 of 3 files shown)


leobalter:
This is working as an initial thing.

- it should only run on PRs (not the master branch)
- for now it does not fail if new or modified tests fail (I want to improve this later). If it runs test262-harness, travis should get green, regardless the results.
- it works if there is no new or modified file in the test/ folder. It just skips the engine runs earlier.
- It has a lot of room for improvements, I'm not intending to ship this perfectly. It works, that's the most important fact.
- one of the desired improvements is comparing the results with the original files from the master branch.

Question I raised writing this:

- Is there a way to set these runs in an informative and fast way on the master branch? Just running all the tests is ridiculously slow and not appropriate for this repo.

ljharb:
Since the tests being slow on master won't actually block anything, is there a reason not to do it?

ljharb:
I suppose it could choke up the "5 simultaneous builds" queue, though.

leobalter:
> Since the tests being slow on master won't actually block anything, is there a reason not to do it?

they are slow enough to execute for Travis, actually, some might take more than 1.5 hour (on each engine using multiple child processes)

leobalter:
As we talked in private, we might have a nice alternative for the full runs announced soon.

PR #1934, opened 2018-11-07
Add cases for early error deleting private names

leobalter:
I want to add other cases with existing static names. It can be on another PR if this get reviewed and approved before.


jbhoosreddy:
@leobalter @rwaldron @littledan is it possible to update the test plan [here](https://github.com/tc39/test262/issues/1343#issuecomment-402505544) to reflect all the tests which have landed?

littledan:
@jbhoosreddy If you post an updated test plan, I will edit the original comment to use that text.

leobalter:
@jbhoosreddy I'm in the process of verifying the current coverage we have and identifying missing parts for the spec. I'll post something as soon as I have this verification done.

PR #1935, opened 2018-11-08
private elements: tests for private method semantics usage

jbhoosreddy:
Tracking issue: #1343

### Summary of changes:
1. Reading a private method separately from calling it
    + All instances of the class share the same function identity for the private method
2. The .name property of the private method is "#foo"
3. It's OK at parse-time to reference a static private method from an instance method, or vice versa, and you can make use of this at runtime with creative application of Function.prototype.call.

### DIFF
--- a/src/class-elements/private-method-multiple-instance-identity-same.case
+++ b/src/class-elements/private-method-multiple-instance-identity-same.case
@@ -1,0 +1,27 @@
+// Copyright (C) 2018 Jaideep Bhoosreddy (Bloomberg LP). All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: Multiple class instances of PrivateName Private method have same function identity 
+info: |
+  Static Semantics: PrivateStringValue
+    PrivateName::
+      #IdentifierName
+  Return the String value consisting of the sequence of code units corresponding to PrivateName
+
+template: productions
+features: [class-methods-private]
+---*/
+
+//- fields
+#m() {}
+//- privateinspectionfunctions
+method() {
+  return this.#m;
+}
+
+const c1 = new C();
+const c2 = new C();
+
+//- assertions
+assert.sameValue(c1.method(), c2.method());
--- a/src/class-elements/private-method-name-property.case
+++ b/src/class-elements/private-method-name-property.case
@@ -1,0 +1,23 @@
+// Copyright (C) 2018 Jaideep Bhoosreddy (Bloomberg LP). All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: PrivateName IdentifierName PropName StringValue
+info: |
+  Static Semantics: PrivateStringValue
+    PrivateName::
+      #IdentifierName
+  Return the String value consisting of the sequence of code units corresponding to PrivateName
+
+template: productions
+features: [class-methods-private]
+---*/
+
+//- fields
+#foo() {}
+//- privateinspectionfunctions
+method() {
+  return this.#foo;
+}
+//- assertions
+assert.sameValue(c.method().name, '#foo');
--- a/src/class-elements/private-static-method-function-call-access.case
+++ b/src/class-elements/private-static-method-function-call-access.case
@@ -1,0 +1,28 @@
+// Copyright (C) 2018 Jaideep Bhoosreddy (Bloomberg LP). All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: PrivateName IdentifierName PropName StringValue
+info: |
+  Static Semantics: PrivateStringValue
+    PrivateName::
+      #IdentifierName
+  Return the String value consisting of the sequence of code units corresponding to PrivateName
+
+template: productions
+features: [class-methods-private]
+---*/
+
+//- fields
+#foo() { return 'test'; }
+static #bar() { return '262'; }
+//- privateinspectionfunctions
+method1() {
+  return this.#foo.call(C);
+}
+method2() {
+  return C.#bar.call(this);
+}
+//- assertions
+assert.sameValue(c.method1(), 'test');
+assert.sameValue(c.method2(), '262');


jbhoosreddy:
@mkubilayk for initial review

src/class-elements/private-method-multiple-instance-identity-same.case
```diff
@@ -0,0 +1,27 @@
...
+
+//- fields
+#m() {}
+//- privateinspectionfunctions
+method() {
+  return this.#m;
+}
+
+const c1 = new C();
+const c2 = new C();
```
rwaldron: This renders _inside_ the class body: 

```js
var C = class {
  *m() { return 42; } #m() {};
  method() {
    return this.#m;
  }

  const c1 = new C();
  const c2 = new C();

}
```

src/class-elements/private-method-multiple-instance-identity-same.case
```diff
@@ -0,0 +1,27 @@
...
+
+/*---
+desc: Multiple class instances of PrivateName Private method have same function identity 
+info: |
+  Static Semantics: PrivateStringValue
+    PrivateName::
+      #IdentifierName
+  Return the String value consisting of the sequence of code units corresponding to PrivateName
+
+template: productions
```
rwaldron: Please use the "default" templates, as "productions" will render tests that will contain a bunch of unnecessary and irrelevant code.

src/class-elements/private-method-name-property.case
```diff
@@ -0,0 +1,23 @@
...
+
+/*---
+desc: PrivateName IdentifierName PropName StringValue
+info: |
+  Static Semantics: PrivateStringValue
+    PrivateName::
+      #IdentifierName
+  Return the String value consisting of the sequence of code units corresponding to PrivateName
+
+template: productions
```
rwaldron: As requested in the previous review comment, please use the "default" template. This will produce a test that looks like this: 

```js
var C = class {
  static m() { return 42; } #foo() {};
  method() {
    return this.#foo;
  }
}

var c = new C();

assert.sameValue(C.m(), 42);
assert.sameValue(Object.hasOwnProperty.call(c, "m"), false);
assert.sameValue(Object.hasOwnProperty.call(C.prototype, "m"), false);

verifyProperty(C, "m", {
  enumerable: false,
  configurable: true,
  writable: true,
});

assert.sameValue(c.method().name, '#foo');
```

When it really only needs to produce a test that looks like this: 

```js
var C = class {
  #foo() {};
  method() {
    return this.#foo;
  }
}

var c = new C();

assert.sameValue(c.method().name, '#foo');
```

src/class-elements/private-static-method-function-call-access.case
```diff
@@ -0,0 +1,28 @@
+// Copyright (C) 2018 Jaideep Bhoosreddy (Bloomberg LP). All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: PrivateName IdentifierName PropName StringValue
```
leobalter: The description should tell what's being verified in this test, I can't read this and tell anything.

src/class-elements/private-static-method-function-call-access.case
```diff
@@ -0,0 +1,28 @@
...
+info: |
+  Static Semantics: PrivateStringValue
+    PrivateName::
+      #IdentifierName
+  Return the String value consisting of the sequence of code units corresponding to PrivateName
```
leobalter: This text doesn't match the test contents

src/class-elements/private-static-method-function-call-access.case
```diff
@@ -0,0 +1,28 @@
...
+template: productions
+features: [class-methods-private]
+---*/
+
+//- fields
+#foo() { return 'test'; }
+static #bar() { return '262'; }
+//- privateinspectionfunctions
+method1() {
+  return this.#foo.call(C);
```
leobalter: This needs adding the matching reference from the spec that allows this in the desc/info above. Otherwise it's hard to tell the goal of this test. The top comment on this PR clarified it a bit for me.

This needs adding a group of tests that assert other behaviors:

- [ ] when the function is called with other values, like `this.#foo.call(this)`, `this.#foo.call({})`, `this.#foo.call(null)`, `C.#bar.call(C)`, including other values as the instance variable name (instead of `this`), `undefined`, etc etc
- [ ] other possible private methods.
- [ ] fields with function values, perhaps?

src/class-elements/private-method-multiple-instance-identity-same.case
```diff
@@ -0,0 +1,27 @@
...
+features: [class-methods-private]
+---*/
+
+//- fields
+#m() {}
+//- privateinspectionfunctions
+method() {
+  return this.#m;
+}
```
leobalter: this can (and should) also be verified for other methods types including assessor methods

src/class-elements/private-method-multiple-instance-identity-same.case
```diff
@@ -0,0 +1,27 @@
+// Copyright (C) 2018 Jaideep Bhoosreddy (Bloomberg LP). All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: Multiple class instances of PrivateName Private method have same function identity 
+info: |
+  Static Semantics: PrivateStringValue
+    PrivateName::
+      #IdentifierName
+  Return the String value consisting of the sequence of code units corresponding to PrivateName
```
leobalter: this is also not testing the name property

src/class-elements/private-method-multiple-instance-identity-same.case
```diff
@@ -0,0 +1,27 @@
...
+
+template: productions
+features: [class-methods-private]
+---*/
+
+//- fields
+#m() {}
+//- privateinspectionfunctions
+method() {
+  return this.#m;
```
leobalter: I believe we already have tests to assert the method relationship with the class prototype? In this case, it should be good to assert which object has their own property and the shared relationship.

jbhoosreddy:
I'm temporarily closing this.. need some time to look at it.

PR #1937, opened 2018-11-08
intl: increase coverage for DateTimeFormat

ryzokuken:
Increase coverage for the Intl.DateTimeFormat constructor by adding a
test that confirms that ToObject(...) is appropriately called inside the
implementation.

/cc @gsathya @rwaldron @littledan

### DIFF
--- a/test/intl402/DateTimeFormat/constructor-options-toobject.js
+++ b/test/intl402/DateTimeFormat/constructor-options-toobject.js
@@ -1,0 +1,42 @@
+// Copyright (C) 2018 Ujjwal Sharma. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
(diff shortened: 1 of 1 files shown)


test/intl402/DateTimeFormat/constructor-options-toobject.js
```diff
@@ -0,0 +1,42 @@
...
+const toObjectResults = [
+  [true, new Boolean(true)],
+  [42, new Number(42)],
+  ['foo', new String('foo')],
+  [{}, {}],
+  [Symbol(), Object(Symbol())]
+];
+
+// Test if ToObject is used to convert primitives to Objects.
+toObjectResults.forEach(pair => {
```
gsathya: nit: you could use destructuring in the args:
```suggestion
toObjectResults.forEach(([value, result]) => {
```

PR #1939, opened 2018-11-08
Add V8 run with all --harmony flags

rwaldron:


.travis.yml
```diff
@@ -32,6 +32,12 @@ matrix:
     node_js: "node"
     script: ./tools/scripts/ci_test.sh
     env: T262ENGINE=v8 hostType=d8 hostArgs='--harmony'
+  - language: node_js
+    name: "New or modified tests on V8 with all --harmony flags"
+    if: type = pull_request
+    node_js: "node"
+    script: ./tools/scripts/ci_test.sh
+    env: T262ENGINE=v8 hostType=d8 hostArgs=$(v8 --help | grep "\--harmony" | perl -p -e "s/ \(enable.*\n//")
```
rwaldron: @leobalter 

```
$ v8 --help | grep "\--harmony" | perl -p -e "s/ \(enable.*\n//"
  --harmony  --harmony-shipping  --harmony-do-expressions  --harmony-class-fields  
--harmony-await-optimization  --harmony-regexp-sequence  --harmony-weak-refs  
--harmony-locale  --harmony-private-fields  --harmony-numeric-separator  --harmony-string-matchall  
--harmony-intl-list-format  --harmony-intl-segmenter  --harmony-namespace-exports  
--harmony-sharedarraybuffer  --harmony-import-meta  --harmony-dynamic-import  
--harmony-array-prototype-values  --harmony-array-flat  --harmony-symbol-description  
--harmony-global  --harmony-json-stringify  --harmony-public-fields  --harmony-static-fields  
--harmony-intl-relative-time-format
```

mathiasbynens: Why not just pass `--harmony`, which enables all `--harmony-*` flags?

mathiasbynens: Ah, `--harmony` only enables staged features. Then again, you probably don’t want to run tests against unstaged WIP implementations.

littledan: WIP features might be very broken. I bet you'll get more meaningful feedback from `--harmony`.

leobalter: the ci jobs are generally not slow and we can afford to have extra runs. I'm not sure if we can see add advantage on adding every flag here but I'd be game to sport available flags from all the engines.

PR #1943, opened 2018-11-08
Rename targets in templates for class elements

leobalter:
Also remove the onlyStrict restriction as the class body is always on strict mode and
the failures are still valid


src/class-elements/delete-error/cls-decl-field-delete-covered.template
```diff
@@ -21,8 +21,6 @@ info: |
     CoverParenthesizedExpressionAndArrowParameterList ultimately derives a
     phrase that, if used in place of UnaryExpression, would produce a
     Syntax Error according to these rules. This rule is recursively applied.
-
-flags: [onlyStrict]
```
rwaldron: Should these tests include something in the `info` that explains why they _must not_  have an `onlyStrict` flag?

leobalter: ok, I'll do it

leobalter:
3b3be75 has picked some errors from the last updates to the Master branch

src/class-elements/delete-error/cls-decl-method-delete-covered.template
```diff
@@ -3,9 +3,12 @@
 
 /*---
 esid: sec-class-definitions-static-semantics-early-errors
-path: language/statements/class/err-method-delete-covered-
+path: language/statements/class/syntax/early-errors/delete/method-delete-covered-
 name: in method, covered
 info: |
+  This file must never use the onlyStrict flag as the strict mode should always
+  be observed inside class bodies.
```
rwaldron: Thanks!

PR #1948, opened 2018-11-13
Update generated tests

leobalter:
this is from #1801, there are some other tests to update for matching.

This should fix Travis CI

PR #1957, opened 2018-11-14
Use $DONOTEVALUATE for newer tests

mathiasbynens:


bakkot:
While you're at it, I just noticed that [CONTRIBUTING.md](https://github.com/tc39/test262/blob/master/CONTRIBUTING.md#handling-errors-and-negative-test-cases) still says "the exact `throw` statement specified in this example" despite the example not containing a `throw` statement.

mathiasbynens:
@bakkot Good catch! Fixed.

leobalter:
#1948

mathiasbynens:
@leobalter Done. Note that this patch now only updartes `CONTRIBUTING.md`.

PR #1959, opened 2018-11-16
Remove invalid invalid values in the tests

FrankYFTang:
I believe these value of "invalid" in the test to test invalid are not valid. 
There are no where in the ECMA 402 spec mentioned the string "invalid" have to be treated as an invalid value and cannot be a valid value. Therefore, we need to remove them from the test. (or we can add "invalid" as an invalid value to the spec.)

### DIFF
--- a/test/intl402/Collator/ignore-invalid-unicode-ext-values.js
+++ b/test/intl402/Collator/ignore-invalid-unicode-ext-values.js
@@ -24,15 +24,15 @@ var defaultLocale = defaultOptions.locale;
 var defaultSortedArray = testArray.slice(0).sort(defaultCollator.compare);
 
 var keyValues = {
-    "co": ["standard", "search", "invalid"],
-    "ka": ["noignore", "shifted", "invalid"],
-    "kb": ["true", "false", "invalid"],
-    "kc": ["true", "false", "invalid"],
-    "kh": ["true", "false", "invalid"],
-    "kk": ["true", "false", "invalid"],
-    "kr": ["latn-hira-hani", "hani-hira-latn", "invalid"],
-    "ks": ["level1", "level2", "level3", "level4", "identic", "invalid"],
-    "vt": ["1234-5678-9abc-edf0", "invalid"]
+    "co": ["standard", "search"],
+    "ka": ["noignore", "shifted"],
+    "kb": ["true", "false"],
+    "kc": ["true", "false"],
+    "kh": ["true", "false"],
+    "kk": ["true", "false"],
+    "kr": ["latn-hira-hani", "hani-hira-latn"],
+    "ks": ["level1", "level2", "level3", "level4", "identic"],
+    "vt": ["1234-5678-9abc-edf0"]
 };
 
 Object.getOwnPropertyNames(keyValues).forEach(function (key) {
--- a/test/intl402/DateTimeFormat/ignore-invalid-unicode-ext-values.js
+++ b/test/intl402/DateTimeFormat/ignore-invalid-unicode-ext-values.js
@@ -20,9 +20,9 @@ locales.forEach(function (locale) {
     var defaultFormatted = defaultDateTimeFormat.format(input);
 
     var keyValues = {
-        "cu": ["USD", "EUR", "JPY", "CNY", "TWD", "invalid"], // DateTimeFormat internally uses NumberFormat
-        "nu": ["native", "traditio", "finance", "invalid"],
-        "tz": ["usnavajo", "utcw01", "aumel", "uslax", "usnyc", "deber", "invalid"]
+        "cu": ["USD", "EUR", "JPY", "CNY", "TWD"], // DateTimeFormat internally uses NumberFormat
+        "nu": ["native", "traditio", "finance"],
+        "tz": ["usnavajo", "utcw01", "aumel", "uslax", "usnyc", "deber"]
     };
     
     Object.getOwnPropertyNames(keyValues).forEach(function (key) {
--- a/test/intl402/NumberFormat/ignore-invalid-unicode-ext-values.js
+++ b/test/intl402/NumberFormat/ignore-invalid-unicode-ext-values.js
@@ -20,8 +20,8 @@ locales.forEach(function (locale) {
     var defaultFormatted = defaultNumberFormat.format(input);
 
     var keyValues = {
-        "cu": ["USD", "EUR", "JPY", "CNY", "TWD", "invalid"],
-        "nu": ["native", "traditio", "finance", "invalid"]
+        "cu": ["USD", "EUR", "JPY", "CNY", "TWD"],
+        "nu": ["native", "traditio", "finance"]
     };
     
     Object.getOwnPropertyNames(keyValues).forEach(function (key) {


littledan:
Well, I wonder if we can treat these tests similarly to the `locale` metadata: even if the spec doesn't say it, we could make a test which is valid in practice given real-world data, with metadata to enable engines to skip when it is not appropriate. Would this be useful? Cc @ms2ger

FrankYFTang:
not sure what do you mean

FrankYFTang:
without this, I need to explicit add the string "invalid" to a blacklist in the v8 to make this test pass. I do not mind to do so if the spec include such value. But w/o such value mentioned in the spec, it is hard to justify for that.

littledan:
I misunderstood. I think this test is trying to check against any unexpected value. Why would this need a blacklist?

Ms2ger:
@NorbertLindenberg: looks like you wrote these tests; could you clarify what they're doing?

NorbertLindenberg:
I wrote these tests for the first edition of the standard, but it seems the following is still true for the fifth edition: The standard specifies, via the RelevantExtensionKeys and LocaleData internal slots of the constructors, which extension keys and values are supported within the Unicode extension of a locale string. Only a subset of the extension keys defined in UTS 35 is supported, and even for supported extension keys not all values defined in UTS 35 are supported. The standard specifies, via the ResolveLocale abstract operation, that all unsupported keys and values are ignored. (The word "invalid" may be a bit strong in this context, since unsupported keys and values don’t cause an exception or otherwise block the use of a locale string containing them).

The tests verify that both unsupported keys and unsupported values are indeed ignored. For example, for NumberFormat, the "cu" key is unsupported, so seemingly valid currency codes such as "USD" should be ignored just like the string "invalid". The "nu" key itself is supported, but the specification excludes the values "native", "traditio", and "finance", so these must be ignored just like the string "invalid".

Note that any string that’s not explicitly allowed by UTS 35 is unsupported, so "invalid" is just one sample out of a universe of unsupported strings. A blacklist of unsupported values is the wrong way to pass these tests.

I’m not up-to-date on current discussions, so it’s possible that changes to the specification after the fifth edition may invalidate some or all of what I said above.

FrankYFTang:
Thanks @NorbertLindenberg (long long time no see... 20 years?) It help me a lot. But I still have some puzzles-
The test require Intl.DateTimeFormat("en-u-nu-invalid").resolvedOptions().locale to be "en" instead of "en-u-nu-invalid". Some code have to change the string from "en-u-nu-invalid" to "en". From https://www.ecma-international.org/ecma-402/#sec-resolvelocale I have hard time to figure out where is such step. I understand the concept @NorbertLindenberg talk about here. But where is such requirement in the spec? Basically, if I follow @NorbertLindenberg  's logic, 
Intl.DateTimeFormat("en-u-nu-foobar").resolvedOptions().locale  should also be "en", right? 
I just have hard time to find the spot in the spec of such "The standard specifies, via the ResolveLocale abstract operation, that all unsupported keys and values are ignored." it will be nice if you can point it out.

NorbertLindenberg:
Step 8 of the ResolveLocale abstract operation iterates over the keys in relevantExtensionKeys to extract information from the Unicode extension. This means, any key not contained in relevantExtensionKeys is ignored. Step 8.h.ii. checks whether the value defined in the Unicode extension is either one of the values in keyLocaleData, or is empty where keyLocaleData contains "true". This means, any value not contained in keyLocaleData is ignored. keyLocaleData is obtained from the localeData parameter in steps 8.a-c.

rwaldron:
@FrankYFTang @littledan @Ms2ger ping

FrankYFTang:
> Note that any string that’s not explicitly allowed by UTS 35 is unsupported, so "invalid" is just one sample out of a universe of unsupported strings. A blacklist of unsupported values is the wrong way to pass these tests.

> Step 8 of the ResolveLocale abstract operation iterates over the keys in relevantExtensionKeys to extract information from the Unicode extension. This means, any key not contained in relevantExtensionKeys is ignored. Step 8.h.ii. checks whether the value defined in the Unicode extension is either one of the values in keyLocaleData, or is empty where keyLocaleData contains "true". This means, any value not contained in keyLocaleData is ignored. keyLocaleData is obtained from the localeData parameter in steps 8.a-c.

Try to understand your comments. Do you mean from the spec point of view, only those listed as type name in https://www.unicode.org/repos/cldr/tags/latest/common/bcp47/number.xml (but also exclude "finance" /  "native" / "traditio" ) should be allowed for "nu" and similar for value of other key?

rwaldron:
@NorbertLindenberg ^^

FrankYFTang:
ok, I understand now. Thanks for the education. Mark it close

sffc:
I don't disagree with the conclusion to this thread.  I just want to point out that I do not see where in the spec we are required to reject invalid numbering systems.

[Section 9.2.6 step 8](https://www.ecma-international.org/ecma-402/#sec-resolvelocale) says,

> Let foundLocaleData be localeData.`[[<foundLocale>]]`.
> Let keyLocaleData be foundLocaleData.`[[<key>]]`.
> Let value be keyLocaleData[0].
> Let supportedExtensionAddition be "".
> If r has an [[extension]] field, then
>   Let requestedValue be UnicodeExtensionValue(r.`[[extension]]`, key).
>   If requestedValue is not undefined, then
>     If requestedValue is not the empty String, then
>       If keyLocaleData contains requestedValue, then
>         // the value is valid and put it into the resolved locale

The variable `localeData` comes from the format object.  Here is how it is defined in NumberFormat, [11.3.3](https://www.ecma-international.org/ecma-402/#sec-intl.numberformat-internal-slots):

> The value of the `[[RelevantExtensionKeys]]` internal slot is « "nu" ».
> 
> NOTE 1: Unicode Technical Standard 35 describes two locale extension keys that are relevant to number formatting, "nu" for numbering system and "cu" for currency. Intl.NumberFormat, however, requires that the currency of a currency format is specified through the currency property in the options objects.
> 
> The value of the `[[LocaleData]]` internal slot is implementation defined within the constraints described in 9.1 and the following additional constraints:
> 
> The list that is the value of the "nu" field of any locale field of `[[LocaleData]]` must not include the values "native", "traditio", or "finance".

It says nowhere what values should be in that list.  What stops us from saying that any string in the universe, except for the three bad ones, are in the list?  Practically, we could say that you can give us whatever numbering system you want, and if we don't have data for it right now, we give you a latin fallback.

sffc:
[Section 9.1](https://www.ecma-international.org/ecma-402/#sec-internal-slots) says vaguely,

> `[[RelevantExtensionKeys]]` is a List of keys of the language tag extensions defined in Unicode Technical Standard 35 that are relevant for the functionality of the constructed objects.

It does not appear to put any constraints on the possible *values* for those keys, only for the keys themselves.  I guess maybe you can read into the spec that the intention is to validate values as well as keys.  I just don't see that explicitly required.

sffc:
Okay, and then [UTS 35 section 3.6.2](https://unicode.org/reports/tr35/tr35.html#Numbering%20System%20Data) says,

> LDML supports multiple numbering systems. The identifiers for those numbering systems are defined in the file bcp47/number.xml.

which points to https://unicode.org/repos/cldr/tags/latest/common/bcp47/number.xml.

I'm convinced.  Sorry for the noise.

Ms2ger:
@macchiati since it seems people agree that this test is correct according to the current specification, perhaps your comment would be more actionable as an issue on the specification?

sffc:
> @macchiati since it seems people agree that this test is correct according to the current specification, perhaps your comment would be more actionable as an issue on the specification?

Done: https://github.com/tc39/ecma402/issues/301

FrankYFTang:
> > Read the bullet point right after that:
> > > [[SortLocaleData]] and [[SearchLocaleData]] (for Intl.Collator) and [[LocaleData]] (for Intl.NumberFormat, Intl.DateTimeFormat, and Intl.PluralRules) are records that have fields for each locale contained in [[AvailableLocales]]. The value of each of these fields must be a record that has fields for each key contained in [[RelevantExtensionKeys]]. The value of each of these fields must be a non-empty list of those values defined in Unicode Technical Standard 35 for the given key that are supported by the implementation for the given locale, with the first element providing the default value.
> 
> I think this is dangerous. Validity checking should be an option — not a requirement within other services like matching.
> 
> Why? Because if you bake validity checking into code that could be running on out-of-date software (eg older version of ICU), then you will reject locales that are perfectly valid on an up-to-date system. For general-purpose software that is not absolutely guaranteed to be up-to-date, it is much better to reject ill-formed locales, but allow invalid ones — and ignore tags that you don't understand. And I don't think we can ever assume that JS implementations will always be up-to-date with the latest CLDR.
> 
> See also [Locale Class](https://docs.google.com/document/d/12ZpGW4qcNV5NLCPQHR1-viMpVs9Pod81ZkwKJvyRfqQ/edit?ts=5bebde67#heading=h.mszsvv7qat7p).
> 
> FYI There is a validity checker on the Java side, but it is not public API:
> 
> com.ibm.icu.impl.locale.LocaleValidityChecker.java
> See also com.ibm.icu.dev.test.util.TestLocaleValidity.java
> 
> _**However, you have to be very cautious in its use, as remarked above.**_

Mark. I think we have some misunderstanding here.
When I start this issue, the focus is on the test for to the constructor of each Intl formatter/collator objects EXCEPT the Intl.Locale itself. 

The issue related to this is in an abstract routine ResolveLocale() 

https://ecma-international.org/ecma-402/#sec-resolvelocale

This routine is called by the constructor of Intl.DateTimeFormat, Intl.NumberFormat, Intl.Segment, Intl.Collator, Intl.PluralRules, Intl.ListFormat, Intl.RelativeTimeFormat, and probably some other newer Intl.* class but it is intentionally, I assume, NOT by Intl.Locale, neither in the constructor, nor anywhere in the spec of other routine.

It is part of the process to decide which exact concrete locale to be used for those operational classes (format, segment, sort, etc) , since the constructor allow the caller to pass in a list of locales with several different options. And in this stage, the software need to figure out exactly which one it should use. Therefore, such validation in this stage should be reasonable, right?

Therefore, such validation exists ONLY in the time of construction of these intl objects, but in particular, will NOT impact the construction or other method of the Intl.Locale object itself.

PR #1963, opened 2018-11-16
Basic tests for private methods

leobalter:


test/language/expressions/class/private-methods/prod-private-async-generator.js
```diff
@@ -0,0 +1,142 @@
...
+ * MethodDefinition : ClassElementName ( UniqueFormalParameters ) { FunctionBody }
+ * 
+ * 1. Let methodDef be DefineMethod of MethodDefinition with argument homeObject.
+ * ...
+ */
+assert.sameValue(c.ref, other.ref, 'The method is defined once, and reused on every new instance');
+
+assert.sameValue(c.ref.name, '#m', 'function name is preserved external reference');
+ctorPromise.then(() => {
+    var iter = c.ref();
```
rwaldron: I think it would be helpful if there was a comment that accompanied this line, which noted that this is actually calling `this.#m`

test/language/expressions/class/private-methods/prod-private-async-method.js
```diff
@@ -0,0 +1,137 @@
...
+
+  get ref() { return this.#m; }
+
+  constructor() {
+    assert.sameValue(
+      hasOwnProperty(this, '#m'), false,
+      'private methods are defined in an special internal slot and cannot be found as own properties'
+    );
+    assert.sameValue(typeof this.#m, 'function');
+    assert.sameValue(this.ref(), this.#m, 'returns the same value');
```
rwaldron: Same as above, should be fixed in https://github.com/tc39/test262/pull/1963/files#diff-0d2c472101ba48bbc07d267a50c4f19bR91

src/class-elements/private-methods/cls-decl.template
```diff
@@ -0,0 +1,123 @@
...
+
+esid: prod-MethodDefinition
+---*/
+
+/*** template notes
+ * method should always be #m
+ * the template provides c.ref() for external reference
+ */
+
+function hasOwnProperty(obj, name) {
```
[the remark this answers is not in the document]
leobalter: Interesting... using Reflect.has I can also confirm the private method is not in the prototype chain. I wonder if I should use both.

PR #1965, opened 2018-11-22
Fix a couple of minor issues in dynamic-import tests

jonco3:
Hi,

This is to fix what I think are a couple of minor issues in the new dynamic import tests:

 - indirect-resolution-1_FIXTURE.js has a bare import specifier.  This should be a relative specifier to find the indirect-resolution-2_FIXTURE.js module.

 - default-property-not-set-own.js should be marked as a module, so that when it imports itself it doesn't get run again.  If this happens $DONE will complain about being called twice.

### DIFF
--- a/test/language/expressions/dynamic-import/namespace/default-property-not-set-own.js
+++ b/test/language/expressions/dynamic-import/namespace/default-property-not-set-own.js
@@ -5,7 +5,7 @@
 description: The default property is not set the if the module doesn't export any default
 esid: sec-finishdynamicimport
 features: [dynamic-import]
-flags: [async]
+flags: [async, module]
 info: |
     Runtime Semantics: FinishDynamicImport ( referencingScriptOrModule, specifier, promiseCapability, completion )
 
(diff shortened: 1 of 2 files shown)


leobalter:
Hi @jonco3, thanks for the PR!

The first fix is correct, the second one is indeed a problem but I’d rather keep the file as a non module code and refactor the code to avoid calling import twice.

We have some other test files doing this approach but I’m only back to my computer on Monday after the holidays. Let me know if you need any support.

jonco3:
Ah yes, that would be better.  I'll see what I can do.

test/language/expressions/dynamic-import/namespace/default-property-not-set-own.js
```diff
@@ -26,7 +26,7 @@ info: |
         5. Return namespace.
 ---*/
 
-import('./default-property-not-set-own.js').then(ns => {
+import('./empty_FIXTURE.js').then(ns => {
```
leobalter: +1

PR #1977, opened 2018-11-29
Add tests for Array.prototype.sort stability

mathiasbynens:
https://github.com/tc39/ecma262/pull/1340


test/built-ins/Array/prototype/sort/stability-11-elements.js
```diff
@@ -0,0 +1,37 @@
...
+];
+assert.sameValue(array.length, 11);
+
+// Sort the elements by `rating` in descending order.
+// (This updates `array` in place.)
+array.sort((a, b) => b.rating - a.rating);
+
+const reduced = array.reduce((acc, element) => acc + element.name, '');
+const isProbablyStable = reduced === 'DGBEFHACIJK';
+assert(isProbablyStable);
```
leobalter: can we use `assert.sameValue(reduced, 'DGBEFHACIJK')` instead? same for the other tests.

the current way is also correct but we could use sameValue here for convention, please.

mathiasbynens: I did it this way intentionally, as it’s important to clarify that this is only a best-effort test. An engine could implement a completely randomized sorting algorithm, and still sometimes get lucky and pass the test. There is no way to write a single test that proves stability. The `isProbablyStable` name is important.

mathiasbynens:
Feedback addressed.

rhuanjl:
Just thought I should note that the stable sort I wrote for Chakracore switches algorithm (from insertion sort to merge sort) at length = 2048 - so this wouldn't fully test the stability of CC's sort 

(Note the stable sort in question is still only in chakracore master, not yet in any release - I wrote it as an external contributor and do not know when it will be included in a release - it is enabled by default in master though)

mathiasbynens:
Thanks, @rhuanjl — I’ll add another test with exactly 2048 elements then.

PR #1981, opened 2018-11-30
Add tests for invalid private names 

leobalter:
I'll follow up with similar tests for valid non early error cases with private names.


littledan:
Cc @mkubilayk @jbhoosreddy

PR #1982, opened 2018-12-03
Expand tests for duplicate lexical names of vardeclarednames in the same statementlist

leobalter:
cc @rkirsling


leobalter:
another ping to @rkirsling who's been helping me verifying these tests. thanks!

rkirsling:
Should this expansion apply to the cases where the `var` precedes the `let` / `const` too?

anba:
> cc @gsathya @anba @allenwb, PTAL on [81c8259](https://github.com/tc39/test262/commit/81c825984dc2dcd6787253b641c6313f9a146406)?

Yes, the test is correct. 

Some of the new tests are actually already present in <https://github.com/tc39/test262/tree/master/test/language/block-scope/syntax/redeclaration>. For example "language/statements/const/syntax/declare-const-binding-with-concurrent-function-declaration.js" is identical to the existing test "language/block-scope/syntax/redeclaration/const-declaration-attempt-to-redeclare-with-function-declaration.js". So it probably makes sense to remove the duplicates and move the new tests into "language/block-scope/syntax/redeclaration" next to the existing tests.

leobalter:
@rkirsling I'm not sure it's fully necessary as we already verify hoisting and these tests capturing lexical names declared before the var declarations in the code. I could do it if you believe it's still necessary. 

@anba thanks! I'll move the files there and fix the duplicates.

rkirsling:
@leobalter I guess my concern is for e8d3aced6f646bceb485402d32e92458d8297fdb in particular.  The existing tests in [language/block-scope/syntax/redeclaration](https://github.com/tc39/test262/tree/master/test/language/block-scope/syntax/redeclaration) cover "`var` before" and "`var` after" alike, so it'd be nice to see the same mirroring for the nested block cases too.

leobalter:
Ok, the tests are now on the block-scope sub folders and using the generation tool previously used for similar cases.

On this rework I discarded verifications on other places we find `Block` being positioned, as `if`, `while`, `try`, `catch`, etc. It would be just too much and I believe the current set is enough to capture the necessary bugs.

leobalter:
@rkirsling I've got tests positioning the var declaration before and after the lexical declarations, same goes for the inner blocks.

@anba PTAL?

anba:
Two nits:
- Some of the template files use `/*{ body }*/;` which results in double semicolons, like for example `const f = 0;;`. I haven't checked if the semicolon in the template file can be omitted in all cases, though!
- The generated file names are a bit misleading, for example "var-declaration-attempt-2-to-redeclare-with-async-function-declaration.js" reads as if a var-declaration is redeclared with an async function declaration, but the actual test case covers the reverse case, i.e. an async function declaration is redeclared with a var-declaration.

Otherwise looks good to me!

leobalter:
Thanks, @anba! I tried a different approach to fix the names and having them simplified.

for the semicolon part, it was easier to change the templates to keep avoiding double semicolons.

PTAL

PR #1985, opened 2018-12-04
test: Ensure private methods are visible from all initializers

mkubilayk:
Adds a test calling private methods from field initializers – according to the test plan posted in #1343.

> Execution order -- private methods are added before any field initializer is run, even if they appear textually later

I will commit the generated tests once the PR is approved.

cc: @littledan @jbhoosreddy


mkubilayk:
I was using that to check the value after `constructor` returns.

mkubilayk:
Thanks for the review. Pushed the generated files.

PR #1988, opened 2018-12-05
Fix various bugs in test cases

anba:
- src/class-elements/prod-private-async-method.case
- test/language/expressions/class/private-methods/prod-private-async-method.js
- test/language/statements/class/private-methods/prod-private-async-method.js
  - Use the `value` argument instead of calling `#m()` again, because the latter returns a new Promise object.

- test/built-ins/Array/prototype/flatMap/non-callable-argument-throws.js
  - `assert(...)` requires a boolean argument, switch to `assert.sameValue(...)`

- test/built-ins/Reflect/ownKeys/return-on-corresponding-order-large-index.js
  - Update test case per https://github.com/tc39/ecma262/pull/1242
  - Also add `4294967294` (maximum array index) and `4294967295` (not an array index) while there.

- test/intl402/NumberFormat/currencyDisplay-unit.js
  - The option value is "narrow-symbol" instead of "narrowSymbol".

- test/intl402/NumberFormat/style-unit.js
  - "unit" style without a unit value leads to a TypeError, use the "generic" unit here.

- test/intl402/Segmenter/iterator/following.js
- test/intl402/Segmenter/iterator/preceding.js
  - Remove `| 0` to ensure we actually test the `ToIndex` operation in the `following` resp. `preceding` methods.
  - ToIndex(1.4) is 1, so don't expect a TypeError for `iter.preceding(1.4)`

- test/intl402/Segmenter/iterator/granularity.js
- test/intl402/Segmenter/prototype/segment/segment-line.js
- test/intl402/Segmenter/prototype/segment/segment-sentence.js
- test/intl402/Segmenter/prototype/segment/segment-word.js
  - The initial value for `breakType` is always `undefined`.

### DIFF
--- a/src/class-elements/prod-private-async-method.case
+++ b/src/class-elements/prod-private-async-method.case
@@ -17,7 +17,7 @@ async #m() { return 42; }
 //- constructor
 assert.sameValue(this.#m.name, '#m', 'function name inside constructor');
 ctorPromise = this.#m().then(value => {
-    assert.sameValue(this.#m(), 42, 'already defined in the ctor');
+    assert.sameValue(value, 42, 'already defined in the ctor');
 }, $DONE);
 
 //- assertions
--- a/test/built-ins/Array/prototype/flatMap/non-callable-argument-throws.js
+++ b/test/built-ins/Array/prototype/flatMap/non-callable-argument-throws.js
@@ -7,7 +7,7 @@ description: >
 features: [Array.prototype.flatMap]
 ---*/
 
-assert(Array.prototype.flatMap);
+assert.sameValue(typeof Array.prototype.flatMap, "function");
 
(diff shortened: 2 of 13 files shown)


src/class-elements/prod-private-async-method.case
```diff
@@ -17,7 +17,7 @@ async #m() { return 42; }
 //- constructor
 assert.sameValue(this.#m.name, '#m', 'function name inside constructor');
 ctorPromise = this.#m().then(value => {
-    assert.sameValue(this.#m(), 42, 'already defined in the ctor');
+    assert.sameValue(value, 42, 'already defined in the ctor');
```
leobalter: oops! Thanks again for catching this!

PR #1993, opened 2018-12-10
Get rid of the leading empty line in info block

caitp:
This makes v8's _monkeyYaml happy --- without it, the info block never ends, and we fail to parse the `features` key.

An explicit indentation marker isn't used because monkeyYaml doesn't supported it, and adding it might slightly slow things down.

Using PyYAML instead was considered, but after trying it, I don't think it looks viable. [simple test cases](https://gist.github.com/caitp/4a59a67f671907c11051a1b7bd5fdaae)

### DIFF
--- a/test/built-ins/Object/fromEntries/requires-argument.js
+++ b/test/built-ins/Object/fromEntries/requires-argument.js
@@ -5,7 +5,6 @@
 esid: sec-object.fromentries
 description: Throws when called without an argument.
 info: |
-
   Object.fromEntries ( iterable )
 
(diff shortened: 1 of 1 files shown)


gsathya:
cc @leobalter @rwaldron 

Could we add a lint check for this too to prevent future regressions?

leobalter:
> Could we add a lint check for this too to prevent future regressions?

Yes! As I've mentioned on #1996, we should restrict our lint check to require a highly interoperable yaml formatting. Fixing parsers is not so easy and this problem is not exclusive to a single test runner or consumer of Test262.

PR #1996, opened 2018-12-15
Remove extra spaces from features list

psmarshall:
This broke my (admittedly terrible) feature parser. Although I guess the requirement is that it is valid YAML, which it currently is. Still, the formatting is very consistent except for this case.


leobalter:
sure thing. I'm trying to collect all adversities to tighten our permitted YAML on Test262. I've seen many parsers having trouble with details here and there.

Thanks for getting this fixed!

PR #2001, opened 2018-12-17
Fix more false positives throwing TypeError in functions

leobalter:


FrankYFTang:
We got the following breakage from this PR w/ v8

/usr/local/google/home/ftang/v8/v8/test/test262/data/test/intl402/RelativeTimeFormat/constructor/constructor/locales-invalid.js:19: TypeError: Cannot convert a Symbol value to a string
    }, `using ${String(locales)} expects ${expectedError}`);
                ^
TypeError: Cannot convert a Symbol value to a string
    at Array.join (<anonymous>)
    at Array.toString (<anonymous>)
    at String (<anonymous>)
    at /usr/local/google/home/ftang/v8/v8/test/test262/data/test/intl402/RelativeTimeFormat/constructor/constructor/locales-invalid.js:19:17

The problem is  `... ${String(locales)} ... `

Looks like all other locales-invalid.js tests does not print ${String(locales)} .

PR #2008, opened 2018-12-18
refactor coverage for flatMap

leobalter:
I found some gaps and false positives to cover flatMap.

I should still take a look on `[].flat` but I'd need more time.


leobalter:
cc @michaelficarra

PR #2009, opened 2018-12-19
remove "TypeError: Cannot convert a Symbol value to a string"

FrankYFTang:
Fix "TypeError: Cannot convert a Symbol value to a string" problem caused by 
printing ${String(locales)}
in https://github.com/tc39/test262/pull/2001
@leobalter @rwaldron

### DIFF
--- a/test/intl402/RelativeTimeFormat/constructor/constructor/locales-invalid.js
+++ b/test/intl402/RelativeTimeFormat/constructor/constructor/locales-invalid.js
@@ -14,7 +14,5 @@ features: [Intl.RelativeTimeFormat]
 assert.sameValue(typeof Intl.RelativeTimeFormat, "function");
 
 for (const [locales, expectedError] of getInvalidLocaleArguments()) {
-    assert.throws(expectedError, function() {
-        new Intl.RelativeTimeFormat(locales)
-    }, `using ${String(locales)} expects ${expectedError}`);
+    assert.throws(expectedError, function() { new Intl.RelativeTimeFormat(locales) });
 }


test/intl402/RelativeTimeFormat/constructor/constructor/locales-invalid.js
```diff
@@ -14,7 +14,5 @@ features: [Intl.RelativeTimeFormat]
 assert.sameValue(typeof Intl.RelativeTimeFormat, "function");
 
 for (const [locales, expectedError] of getInvalidLocaleArguments()) {
-    assert.throws(expectedError, function() {
-        new Intl.RelativeTimeFormat(locales)
-    }, `using ${String(locales)} expects ${expectedError}`);
```
ljharb: I’m confused; this should not be throwing an error since String() works on Symbols.

FrankYFTang: > I’m confused; this should not be throwing an error since String() works on Symbols.

for all the returned of getInvalidLocaleArguments() in https://github.com/tc39/test262/blob/master/harness/testIntl.js ?

FrankYFTang: The problem is String([Symbol.toPrimitive])  
Notice, not String(Symbol.toPrimitive)

FrankYFTang: According to http://ecma-international.org/ecma-262/#sec-text-processing
String(value) will call ToString(value) first in step 2.b
And according to http://ecma-international.org/ecma-262/#sec-tostring
I see no Array list there. 

I got the following in v8 if I try
String([Symbol.toPrimitive])

Uncaught TypeError: Cannot convert a Symbol value to a string
    at Array.toString (native)
    at String (<anonymous>)
    at <anonymous>:1:5

Not sure this is correct behavior or not but surely not a issue of Intl.RelativeTimeFormat .

ljharb: ah, because you're stringifying an array, sure. You'd need `String(arr.map(String))` instead.

leobalter: Lovely. I just tried to provide some information on the current assertion but this seems to require more work to do. I’ll need a bit more time to think through something else. For now, let the test remain correct.

PR #2010, opened 2018-12-20
Move localeMatcher before type and style

FrankYFTang:
Sync with https://github.com/tc39/proposal-intl-list-format/pull/36


FrankYFTang:
> LGTM. Please let me know when it’s ready to land

the Spec changes is landed now.

PR #2023, opened 2019-01-05
Test update for proposed Annex B.3.5 simplification

rkirsling:
This PR updates tests for https://github.com/tc39/ecma262/pull/1393.


test/annexB/language/statements/try/catch-redeclared-for-of-var.js
```diff
@@ -0,0 +1,27 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
```
leobalter: Nit pick: you should modify this to your name and this year.

test/annexB/language/statements/try/catch-redeclared-for-of-var.js
```diff
@@ -0,0 +1,27 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-variablestatements-in-catch-blocks
+es6id: B.3.5
```
leobalter: nic pick #2: this should be removed as this behavior is not compatible with ES6 anymore.

test/annexB/language/statements/try/catch-redeclared-for-of-var.js
```diff
@@ -0,0 +1,27 @@
+// Copyright (C) 2019 Sony Interactive Entertainment Inc.
```
rkirsling: Hmm, so the validator is complaining because I dropped the `All rights reserved.` here, which we [usually](https://github.com/WebKit/webkit/blob/master/Source/WebInspectorUI/UserInterface/Views/Layers3DContentView.js#L2) don't include in our copyright line.

I guess #851 is still outstanding, otherwise I could just drop it altogether. 😅

ljharb:
https://github.com/tc39/ecma262/pull/1393 has achieved consensus; it'd be great to merge these tests so we can update the labels there :-)

test/annexB/language/statements/try/catch-redeclared-for-of-var.js
mathiasbynens: I guess this is blocking the merge? Can this be addressed somehow?

rkirsling:
Thanks for catching that!

test/annexB/language/statements/try/catch-redeclared-for-of-var.js
rkirsling: Hoping @leobalter or @littledan can advise. 🤞

rkirsling: I'll apply your suggestion for now and others can object / accept when they get to it. 🙂

Issue #1026, opened 2017-05-07
Specify rules for integration tests in test262


ajvincent:
I have been working on a [ECMAScript-based Membranes project](https://github.com/ajvincent/es7-membrane/) which combines several different ECMAScript features:
- Proxies
- WeakMaps
- Symbols
- (to-do) Modules
- (in the "laboratory" pages) Template strings

It is their integration into one project that we (Dr. Mark Miller and I) are interested in adding to test262:  specifically, as a set of integration tests.  Test262 appears to contain only unit tests right now, so it's not entirely clear where integration tests might go in the first place.  So we're definitely breaking new ground here.

Similarly, because we're talking about testing a subset of my membrane project's code, I would have to extract and reduce the library being tested into a form appropriate for inclusion as an example for the tests to exercise.  This, I can start as a branch in my own project repository.

This ticket is to establish the guidelines for integration tests _in general_, using my project as a model to begin from.  Hopefully, a positive resolution will include adding a set of integration tests to exercise a minimal membrane implementation.

rwaldron:
Since test262 is explicitly for specification conformance, perhaps creating something like the [test262-parser-tests](https://github.com/tc39/test262-parser-tests) project would be appropriate.

leobalter:
I saw the membranes project and I have some questions:

- How those tests are applied to other projects?
- The tests are using Node + jasmine. Test262 is neutral and not providing favor to any JS implementation, that can’t be achieved using Node. The same goes to choosing Jasmine or any specific test framework.
- Considering all the dependencies - including above - merging this with Test262 would also mean requiring an immense work changing all the runners and their API. The ideal is to adapt all those new tests to match Test262 style.
- Even if we adopt a new integration tests project/repo at TC39, we can’t have a Node + Jasmine (or other test framework) dependency and require other runtimes to include these dependencies to their CI. For that, we will need consensus.

Saying that, those tests look indeed specific to the project you mentioned. We need the feedback from other implementors saying those tests might be useful for other projects, related to the specs.

ajvincent:
[...]

PR #2024, opened 2019-01-05
Integration tests: Membrane (Proxy, WeakMap, Reflect, Map).

ajvincent:
See #1026 - this is a first set of integration tests.

A few trouble spots:

- I didn't pick out a specific esid for any of the tests.  It's a required field for the tests' frontmatter... but these are integration tests, so no single section of ECMAScript is being tested.  I suggest an "integration-tests(section, section, ...)" esid format, for clarity.
- In the [recommended test-262 web runner](https://bakkot.github.io/test262-web-runner/), I couldn't put the integration-Membrane.js file into a subdirectory of harness/ .  (This is also the only test runner I actually used to develop and run these tests.)
- This is, I believe, my first contribution to test-262, and of course it's in a new area entirely.  So it needs to be more carefully reviewed, for both design and correctness.  At TC39 in July 2018, it was suggested that test-262 might not be the right repository for integration tests, but this is just a starting point.

### DIFF
--- a/harness/features.yml
+++ b/harness/features.yml
@@ -4,3 +4,4 @@ testAtomics.js: [ArrayBuffer, Atomics, DataView, SharedArrayBuffer, Symbol, Type
 testBigIntTypedArray.js: [BigInt, TypedArray]
 testTypedArray.js: [TypedArray]
 isConstructor.js: [Reflect.construct]
+integration/Membrane.js: [Proxy, WeakMap, Reflect, Map]
--- a/harness/integration-Membrane.js
+++ b/harness/integration-Membrane.js
@@ -1,0 +1,273 @@
+/**
+ * @fileoverview This tests how multiple features of ECMAScript work together.
+ * Specifically, Proxy, WeakMap, Reflect, and Map.  Membranes, the primary
+ * motivation behind both Proxy and WeakMap, require plenty of overhead to work.
+ *
+ * For an introduction to Membranes, read https://tvcutsem.github.io/js-membranes .
+ */
+
+const ORIGIN_GRAPH = Symbol("Origin graph");
+
+/**
+ * @constructor
+ */
+function Membrane() {
+  /**
+   * @private
+   */
+  this.graphs = new Map(/*
+    name: new ObjectGraph(name, this)
+  */);
+
+  /**
+   * @private
+   */
+  this.mappings = new WeakMap(/*
+    object: new Map(
+      string or symbol: { proxy, revoke }
+    )
+  */);
+}
+
+/**
+ * Get a named object graph.  Create it if necessary.
+ *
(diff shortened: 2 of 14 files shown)


ajvincent:
That's really odd:  I had a linting failure for the features part of the frontmatter, but when I fixed that, another automated test failed which had passed only a few minutes ago.

leobalter:
I'd like to give these tests a proper space for many reasons.

For now I created https://github.com/tc39/test262-integration for them just like the test262-parser-tests project.

These tests have different requirements than the traditional unit tests from Test262 and we can have a better perspective of failure/passing reports if we can set them there. Also, the details for Test262 (this project) should not be a blocker for the new tests.

If they come as well as different projects, we could also improve timing when they are consumed. e.g. I could have a special CI job for these.

WDYT?

If you bring these tests there we can set goals to get these tests running on different targets and identify how to better get results and effects from this work.

rwaldron:
@leobalter I think this is a brilliant idea—looking forward to seeing how that project evolves!

littledan:
I think it's important that this project make it easy for JS implementations to integrate the tests. If it starts breaking into multiple repositories, with their own setups, that could be hard. Do we know how successful test262-parser-tests has been at getting integrated into JS engines?

leobalter:
> Do we know how successful test262-parser-tests has been at getting integrated into JS engines?

I don't know. I didn't work on the engagement for engines to integrate these tests neither I can afford with time to maintain that project.

In other scope, I can find some plans for the integration tests and I'd like to fit more relaxed goals for the new project that wouldn't just fit into Test262 that we could rely on. Saying that, we can work on some integration and even CI jobs, e.g. webdriver to experiment outreach ourselves.

Multiple repositories is not a bad thing and git has nice tools to integrate things.

> with their own setups

Do we have all the necessary setup identified for these integration tests? Do we want or need more? Can we fit all of them on Test262?

We have a very similar plan for stress tests and this integration project might be the answer or just go as parallel.

If we identify all of these on the new project we can have all the answers before just pushing more content that won't be used at its best.

ajvincent:
Honestly, I'm fine with whatever the consensus is... but I do need a framework to put the tests in.  I notice that test262-integration is a bare-bones repository having only two files in it, which is obviously insufficient.

I also don't mind being the guinea pig with these membrane-oriented tests as the starting point.  But my time is limited, particularly when my university classes are in session.  In short, I must fill a support role, not a lead one.

littledan:
Would it make sense to share tests between JavaScript implementations in this integration test repo, or should that be a fourth place?

littledan:
cc @gsathya

ajvincent:
I think I can close out this pull request now, as test262-integration has the tests.

PR #2025, opened 2019-01-08
Correct the expectation for format-fraction-digits-precision.

FrankYFTang:
The expectation is simply wrong.


test/intl402/NumberFormat/prototype/format/format-fraction-digits-precision.js
```diff
@@ -23,8 +23,8 @@ var numberingSystems = [
 ];
 var testData = {
     // Ref tc39/ecma402#128
-    "12344501000000000000000000000000000": "12344501000000000487815444678311936.0",
-    "-12344501000000000000000000000000000": "-12344501000000000487815444678311936.0"
+    "12344501000000000000000000000000000": "12344501000000000000000000000000000.0",
+    "-12344501000000000000000000000000000": "-12344501000000000000000000000000000.0"
```
leobalter: cc @littledan @Yaffle @sffc @caridy

leobalter: Anyway, this matches engines behavior with no known exception so far.

PR #2026, opened 2019-01-08
Update same-line-async-gen-private-field-usage.js

nicolo-ribaudo:
I caught this because it was failing in Babel's tests.


nicolo-ribaudo:
Oh, I just noticed that that file is auto generated.

nicolo-ribaudo:
I'm not sure why travis's status is still pending :thinking:

leobalter:
sounds like travis failed to answer... the change looks fine to me.

PR #2027, opened 2019-01-09
Fix the rounding issue

FrankYFTang:


FrankYFTang:
@anba @littledan @gsathya @zbraniecki @jungshik

sffc:
The spec says to use half-up rounding for both fraction and scientific.

https://www.ecma-international.org/ecma-402/5.0/index.html#sec-torawprecision

> Let e and n be integers such that 10p–1 ≤ n < 10p and for which the exact mathematical value of n × 10e–p+1 – x is as close to zero as possible. If there are two such sets of e and n, pick the e and n for which n × 10e–p+1 is larger.

and

https://www.ecma-international.org/ecma-402/5.0/index.html#sec-torawfixed

> Let n be an integer for which the exact mathematical value of n ÷ 10f – x is as close to zero as possible. If there are two such n, pick the larger n.

In both cases, if there is a tie, pick the "larger" one.  This part of the spec operates only on positive numbers, so this means to round away from zero.

test/intl402/NumberFormat/prototype/format/format-significant-digits-precision.js
```diff
@@ -23,8 +23,8 @@ var numberingSystems = [
 ];
 var testData = {
     // Ref tc39/ecma402#128
```
ljharb: this comment links to the still-open issue: https://github.com/tc39/ecma402/issues/128 - are we confident about the conclusion both there and here?

sffc:
I've commented before on tc39/ecma402#128, and the current web reality, which is consistent with my interpretation of the spec, is that rounding behavior in the spec should not be influenced by floating-point noise.

```javascript
(123.445).toLocaleString("en", {maximumFractionDigits:2})  // 123.45 in Chrome and Firefox
```

I also found the PR that added the current test262 behavior: #857

It does make sense to resolve tc39/ecma402#128 first, and then update test262.

test/intl402/NumberFormat/prototype/format/format-significant-digits-precision.js
leobalter: are you sure I added this ref? I can't find a git log doing it. I'm not challenging you, but actually I don't remember the reason it's added here or anything. In any case, I can review this for sure.

leobalter: https://github.com/tc39/test262/commit/27b156c31e644166e606df21476e39b9a75f4fbc

found it. That was a just a request.

leobalter:
I'm convinced by @sffc's interpretation and also the fact this reflects web-reality.

the proposed changes have a passing result in all 4 major engines.

PR #2028, opened 2019-01-09
Remove no-longer supported privae tag

FrankYFTang:
Per https://github.com/tc39/proposal-intl-locale/pull/66
private tags are no longer supported. Remove them from tests.


FrankYFTang:
@littledan @zbraniecki @anba @srl295 @Ms2ger

FrankYFTang:
based on discussion with @littledan I will change this PR to testing these same locale but with different expectation instead. Stay tune.

PR #2029, opened 2019-01-09
Remove unsupported irregular grandfathered tags.

FrankYFTang:
Per https://github.com/tc39/proposal-intl-locale/pull/66 irregular grandfathered tag is no longer supported. Remove them from the test.


littledan:
Rather than removing these tests, can you test for the behavior when they are used (which is still well-defined)?

FrankYFTang:
> Rather than removing these tests, can you test for the behavior when they are used (which is still well-defined)?

That seems reasonable. Will work on that.

FrankYFTang:
> Rather than removing these tests, can you test for the behavior when they are used (which is still well-defined)?

Actually, I figure this is a hard part for now.
https://github.com/tc39/ecma402/pull/289
is still open. In particular, the behavior of how to react to 1) irregular 2) private 3) 4 letter language code 4) 2*3ALPHA "-" extlang is allowed in the current shape of ECMA 402 IsStructurallyValidLanguageTag which the text said " a well-formed BCP 47 language tag as specified in RFC 5646 section 2.1" but the rest of the intl-locale spec is referring to Unicode Locale Identifier that disallow the above 4 cases. so if Language Tags /  IsStructurallyValidLanguageTag /  CanonicalizeLanguageTag is change to UTS35 Unicode Locale Identifier then I can easily change the test.

test/intl402/Locale/constructor-parse-twice.js
```diff
@@ -13,77 +13,35 @@ info: |
   30. Let r be ! ApplyUnicodeExtensionToTag(tag, opt, relevantExtensionKeys).
 
   ApplyOptionsToTag( tag, options )
-  9. If tag matches neither the privateuse nor the grandfathered production, then
```
leobalter:
```
ApplyOptionsToTag( tag, options )

1. Assert: Type(tag) is String.
2. If IsStructurallyValidLanguageTag(tag) is false, throw a RangeError exception.
```

is probably the most fundamental piece in these descriptions to make a reference for the RangeError assertions

test/intl402/Locale/constructor-parse-twice.js
```diff
@@ -94,3 +52,13 @@ for (const {tag, options, canonical} of testData) {
     `new Intl.Locale("${tag}", ${options}).toString() returns "${canonical}"`
   );
 }
+
+assert.throws(RangeError, () =>
+    new Intl.Locale("no-bok", {region: "NO", calendar: "gregory"}).toString());
```
leobalter: these `.toString()` should never happen, right? Please remove them in the other files as well.

test/intl402/Locale/likely-subtags-grandfathered.js
```diff
@@ -93,19 +110,7 @@ const irregularGrandfathered = [
...
-
-    assert.sameValue(loc.maximize().toString(), maximized);
-    assert.sameValue(loc.maximize().maximize().toString(), maximized);
-
-    assert.sameValue(loc.minimize().toString(), minimized);
-    assert.sameValue(loc.minimize().minimize().toString(), minimized);
-
-    assert.sameValue(loc.maximize().minimize().toString(), minimized);
-    assert.sameValue(loc.minimize().maximize().toString(), maximized);
+    assert.throws(RangeError, () => new Intl.Locale(tag));
```
leobalter: only tag is used, cleanup the bindings, please

test/intl402/Locale/likely-subtags-grandfathered.js
```diff
@@ -169,6 +155,32 @@ for (const {tag, canonical, maximized = canonical, minimized = canonical} of reg
...
+    },
+    {
+        tag: "zh-min-nan",
+        canonical: "nan",
+        maximized: "nan-Hans-CN",
+    },
+];
+
+for (const {tag, canonical, maximized = canonical, minimized = canonical} of regularGrandfatheredWithExtLang) {
+    assert.throws(RangeError, () => new Intl.Locale(tag));
```
leobalter: only tag is used, please cleanup the bindings for this for loop

FrankYFTang:
@leobalter thanks. Fix the last missing toString. PTAL

PR #2040, opened 2019-01-19
Add test for 2.1 CanonicalizeLocaleList ( locales )

FrankYFTang:
See https://tc39.github.io/proposal-intl-locale/#sec-canonicalizelocalelist

This test tests the change in https://github.com/tc39/proposal-intl-locale/pull/64

@littledan @anba @zbraniecki


FrankYFTang:
@leobalter

PR #2041, opened 2019-01-19
Remove "line" tests and ensure error under "line"

FrankYFTang:
Sync test262 expectations with https://github.com/tc39/proposal-intl-segmenter/pull/60
@littledan  @zbraniecki @gsathya @mathiasbynens 
Some test changed to test what should happen if caller still use "line"
Some tests removed due to redundancy after the spec change.


FrankYFTang:
@littledan @mathiasbynens @litherum @sffc @zbraniecki  @gsathya 
could some of you review these changes?

test/intl402/Segmenter/constructor/constructor/options-valid-combinations.js
```diff
@@ -7,16 +7,12 @@ description: Checks handling of valid values for the granularity option to the S
 info: |
     Intl.Segmenter ([ locales [ , options ]])
 
-    9. Let lineBreakStyle be ? GetOption(options, "lineBreakStyle", "string", « "strict", "normal", "loose" », "normal").
-    13. Let granularity be ? GetOption(options, "granularity", "string", « "grapheme", "word", "sentence", "line" », "grapheme").
+    13. Let granularity be ? GetOption(options, "granularity", "string", « "grapheme", "word", "sentence" », "grapheme").
```
sffc: Are the line numbers still the same, or do they need to be updated?

gsathya:
cc original author @Ms2ger

test/intl402/Segmenter/constructor/constructor/options-valid-combinations.js
```diff
@@ -25,35 +21,16 @@ combinations.push([
...
-    undefined,
-  ]);
-}
-
 for (const granularity of granularityOptions) {
   combinations.push([
     { granularity },
     granularity,
-    granularity === "line" ? "normal" : undefined,
+    undefined,
```
sffc: You can delete the third element of the array and then remove it in the for loop on line 32, where it is mapped to the local variable "lineBreakStyle".

FrankYFTang: since lineBreakStyle was in the spec, I follow @littledan suggestion in other tests PR to keep it with a different expectation so see that property / value will actually have the right output instead of ignore and not checking it.

test/intl402/Segmenter/constructor/constructor/options-valid-combinations.js
FrankYFTang: DONE

FrankYFTang:
@leobalter

FrankYFTang:
Ping!

test/intl402/Segmenter/constructor/constructor/options-valid-combinations.js
sffc: I see.  Makes sense.

littledan:
Where's the test that using "line" is an error?

FrankYFTang:
@littledan ah... I added one in test/intl402/Segmenter/constructor/constructor/options-granularity-valid.js

FrankYFTang:
@leobalter could you merge this? Thanks

PR #2043, opened 2019-01-20
Improve coverage for UniqueFormalParameters

jugglinmike:
Previously, the early error prohibiting duplicate entries in
UniqueFormalParameters was only tested in terms of async functions. In
one case, this was misattributed to UniqeFormalParameters and only
enforced for strict mode code.

Extend coverage to the other function-creating productions which include
UniqueFormalParameters (i.e. method definitions and non-async arrow
functions), and update the existing tests to more accurately describe
the source of the error.

---

Found some missing coverage while implementing some of this for JSHint. @bterlson might want to take a look.


bterlson:
Nice find!

PR #2005, opened 2018-12-18
UpdateExpressions: add necessary changes to tests to reflect latest specification.

rwaldron:

PR #2045, opened 2019-01-24
UpdateExpressions: No early error for arguments and eval in non-strict mode

adrianheine:
According to sec-identifiers-static-semantics-assignmenttargettype,
`AssignmentTargetType` of `arguments` and `eval` in non-strict mode code is
`simple`.
sec-update-expressions-static-semantics-early-errors mandates early errors for
`UpdateExpression` if `AssignmentTargetType` is `invalid` or `strict`.

These were introduced in #2005.

### DIFF
--- a/test/language/expressions/postfix-decrement/arguments-nostrict.js
+++ b/test/language/expressions/postfix-decrement/arguments-nostrict.js
@@ -7,24 +7,19 @@ description: >
   It is an early Reference Error if AssignmentTargetType of LeftHandSideExpression is invalid. (arguments)
 info: |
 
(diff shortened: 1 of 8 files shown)


leobalter:
cc @rwaldron

leobalter:
Just informing I'll leave this to Rick for review as he worked on the previous tests.

rwaldron:
Thanks for surfacing this, I'm realizing that I constructed the evaluation in the wrong order. I'm going to merge this, but will follow up with some frontmatter cleanup. 

Thanks!!

PR #2047, opened 2019-01-25
move more case from valid to invalid

FrankYFTang:
This is based on UTS35 in the new spec.


FrankYFTang:
@littledan @gsathya @zbraniecki @anba @leobalter

PR #2048, opened 2019-01-25
change irregular grandfather tag to it's canonical

FrankYFTang:
To sync with the latest spec which referring to UTS35


FrankYFTang:
@littledan @gsathya @zbraniecki @anba @leobalter

PR #2052, opened 2019-01-29
Change expectation to keep variant

FrankYFTang:


FrankYFTang:
@littledan @leobalter

FrankYFTang:
yield to https://github.com/tc39/test262/pull/2054

PR #2053, opened 2019-01-29
Fix invalidLanguageTags array

FrankYFTang:
@littledan @leobalter


FrankYFTang:
> Weird, it was passing in an array instead? Good fix.

It was my mistake. I changed it recently. I copied the code of passing object for the valid cases to this checking of invalid cases but forgot to change the { } to [].

PR #2054, opened 2019-01-29
Remove 'cel-gaulish' from test

FrankYFTang:
'cel-gaulish', according to https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry

is 
"
Type: grandfathered
Tag: cel-gaulish
Description: Gaulish
Added: 2001-05-25
Deprecated: 2015-03-29
Comments: see xcg, xga, xtg
"
I think there are no point to test a deprecated + grandfathered tag regardless what behavior it should be. It is pointless to assure the quality of ECMA toward a deprecated grandfathered tag.

### DIFF
--- a/test/intl402/Locale/extensions-grandfathered.js
+++ b/test/intl402/Locale/extensions-grandfathered.js
@@ -22,18 +22,6 @@ features: [Intl.Locale]
 ---*/
 
 const testData = [
-    // Regular grandfathered without modern replacement.
-    {
-        tag: "cel-gaulish",
-        options: {
-            language: "fr",
-            script: "Cyrl",
-            region: "FR",
-            numberingSystem: "latn",
-        },
-        canonical: "fr-Cyrl-FR-u-nu-latn",
-    },
-
     // Regular grandfathered with modern replacement.
     {
         tag: "art-lojban",
--- a/test/intl402/Locale/getters-grandfathered.js
+++ b/test/intl402/Locale/getters-grandfathered.js
@@ -26,13 +26,6 @@ features: [Intl.Locale]
 ---*/
 
 // Regular grandfathered language tag.
-var loc = new Intl.Locale("cel-gaulish");
-assert.sameValue(loc.baseName, "cel-gaulish"); // Step 5.
-assert.sameValue(loc.language, "cel-gaulish");
-assert.sameValue(loc.script, undefined);
-assert.sameValue(loc.region, undefined);
-
-// Regular grandfathered language tag.
 assert.throws(RangeError, () => new Intl.Locale("zh-min"));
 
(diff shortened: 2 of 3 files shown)


anba:
I'd like to keep tests for grandfathered tags, because one reason to switch to UTS35 was to get *better* handling of grandfathered tags.

anba:
> there is no language in ECMA-402 that says anything specific about deprecated grandfathered tags.

This is not correct.

https://tc39.github.io/ecma402/#sec-canonicalizelanguagetag

>  A conforming implementation shall take the steps specified in the “BCP 47 Language Tag to Unicode BCP 47 Locale Identifier” algorithm, from Unicode Technical Standard #35 LDML § 3.3.1 BCP 47 Language Tag Conversion. 

And http://unicode.org/reports/tr35/#Language_Tag_to_Locale_Identifier
> 5. If the tag is one of the five deprecated grandfathered tags (cel-gaulish, i-default, i-enochian, i-mingo, zh-min) remaining after step #1, prefix by "und-x-".

So this case is definitely covered.

FrankYFTang:
Notice my KEY rational to remove "cel-gaulish" is NOT ONLY it is a grandfathered tag, but MORE IMPORTANTLY it is a DEPRECATED grandfathered tag. If we keep it inside the test, then test262 is not supporting such "deprecation", right?

anba:
The definition of "deprecated" from Webster isn't too relevant in this context, instead [RFC 5646, 3.1.6.  Deprecated Field](https://tools.ietf.org/html/rfc5646#section-3.1.6) needs to be taken into account. 

> but why should the implementer to spend enough energy to pass a test for something deprecated?

V8 is certainly free to not support this specific case and instead simply exclude this test. I'm open to split this specific case into a separate test file, so other tests in the same test file can still be tested in V8. But in the general case, V8 should definitely aim to support proper replacements of deprecated language subtags, because deprecation is a general mean of the IANA language tag registry to update language subtags.

> Notice my KEY rational to remove "cel-gaulish" is NOT ONLY it is a grandfathered tag, but MORE IMPORTANTLY it is a DEPRECATED grandfathered tag.

All grandfathered tags but "i-default" and "i-mingo" have a "Deprecated" field. So when following the argument that a deprecated tag shouldn't be tested, any tests except ones using "i-default" or "i-mingo" would need to be removed.

FrankYFTang:
Thanks for pointing out these. I really appreciate it. I  think it is a good idea to split the test as you suggest. In particular, I think the   five deprecated grandfathered tags (cel-gaulish, i-default, i-enochian, i-mingo, zh-min) as mentioned by you should be always separated from other test files. I will also file ICU ticket against these.

PR #2060, opened 2019-02-01
Add test for ({* identifier})

TimothyGu:
Counterpart to language/expressions/object/prop-def-invalid-async-prefix.js.


leobalter:
The test is correct but I wonder what motivated it beyond the other referenced test file. Just out of curiosity, for sure.

TimothyGu:
Prior to a fix I made (https://github.com/acornjs/acorn/commit/26b6b90abd4265c854254f984b5e46f8ff1efa06), Acorn incorrectly parsed this as valid.

Issue #2066, opened 2019-02-06
Missing test case for private fields storing functions


caiolima:
Some VM's, such as JSC, treats ```o.member()``` as a special AST node. In such case, it would be valuable having a test case for private fields covering ```o.#member()``` case, where ```#member = () => 'test252';``` for example.

littledan:
Sounds like a great idea to test the combination here, analogously to how many template tests get at other combinations.

leobalter:
#2067 should have fixed this.

PR #2067, opened 2019-02-06
Adding case where private field stores a function

caiolima:
This is handling missing tests from #2066.

### DIFF
--- a/src/class-elements/private-field-as-async-arrow-function.case
+++ b/src/class-elements/private-field-as-async-arrow-function.case
@@ -1,6 +1,4 @@
 // Copyright (C) 2019 Caio Lima. All rights reserved.
-// Copyright (C) 2018 Jaideep Bhoosreddy (Bloomberg LP). All rights reserved.
-// Copyright (C) 2018 Leo Balter. All rights reserved.
 // This code is governed by the BSD license found in the LICENSE file.
 
 /*---
@@ -18,6 +16,7 @@ info: |
     CallExpression[?Yield, ?Await].PrivateName
 template: default
 features: [class-fields-private, async-functions, arrow-function]
+flags: [async]
 ---*/
 
 //- elements
@@ -29,9 +28,6 @@ method() {
 //- assertions
 let c = new C();
 
-async function asyncRun() {
-    assert.sameValue(await c.method(), 'test262');
-}
-
-asyncRun();
+c.method().then((value) => assert.sameValue(value, 'test262'))
+  .then($DONE, $DONE);
 
--- a/src/class-elements/private-field-as-async-arrow-function.case
+++ b/src/class-elements/private-field-as-async-arrow-function.case
@@ -1,0 +1,37 @@
+// Copyright (C) 2019 Caio Lima. All rights reserved.
+// Copyright (C) 2018 Jaideep Bhoosreddy (Bloomberg LP). All rights reserved.
+// Copyright (C) 2018 Leo Balter. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: Calling async arrow function returned from private field access
+info: |
+  Updated Productions
+
+  CallExpression[Yield, Await]:
+    CoverCallExpressionAndAsyncArrowHead[?Yield, ?Await]
+    SuperCall[?Yield, ?Await]
+    CallExpression[?Yield, ?Await]Arguments[?Yield, ?Await]
+    CallExpression[?Yield, ?Await][Expression[+In, ?Yield, ?Await]]
+    CallExpression[?Yield, ?Await].IdentifierName
+    CallExpression[?Yield, ?Await]TemplateLiteral[?Yield, ?Await]
+    CallExpression[?Yield, ?Await].PrivateName
+template: default
+features: [class-fields-private, async-functions, arrow-function]
+---*/
+
+//- elements
+#m = async () => 'test262';
+
+method() {
+  return this.#m();
+}
+//- assertions
+let c = new C();
+
+async function asyncRun() {
+    assert.sameValue(await c.method(), 'test262');
+}
+
+asyncRun();
+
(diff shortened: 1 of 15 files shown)


src/class-elements/private-field-as-function.case
```diff
@@ -0,0 +1,31 @@
...
+
+  CallExpression[Yield, Await]:
+    CoverCallExpressionAndAsyncArrowHead[?Yield, ?Await]
+    SuperCall[?Yield, ?Await]
+    CallExpression[?Yield, ?Await]Arguments[?Yield, ?Await]
+    CallExpression[?Yield, ?Await][Expression[+In, ?Yield, ?Await]]
+    CallExpression[?Yield, ?Await].IdentifierName
+    CallExpression[?Yield, ?Await]TemplateLiteral[?Yield, ?Await]
+    CallExpression[?Yield, ?Await].PrivateName
+
```
leobalter: Remove this empty line, please.

The trailing empty line is breaking some yaml parsers for engines consuming Test262.

src/class-elements/private-field-as-function.case
```diff
@@ -0,0 +1,31 @@
...
+template: productions
```
leobalter: I believe the `default` templates should be better here.

In this case, productions are observing positioning of fields and this case is testing a specific behavior rather than positioning.

src/class-elements/private-field-as-function.case
```diff
@@ -0,0 +1,31 @@
...
+features: [class-fields-private]
+---*/
+
+//- elements
+#m = () => 'test262';
+//- privateinspectionfunctions
```
leobalter: if you change the template to `default` this line can be removed.

src/class-elements/private-field-as-function.case
```diff
@@ -0,0 +1,31 @@
...
+    CallExpression[?Yield, ?Await].IdentifierName
+    CallExpression[?Yield, ?Await]TemplateLiteral[?Yield, ?Await]
+    CallExpression[?Yield, ?Await].PrivateName
+
+template: productions
+features: [class-fields-private]
+---*/
+
+//- elements
+#m = () => 'test262';
```
leobalter: I wonder if it's interesting to also add other form of functions, e.g.:

- `#m = async () => 'test262';`
- `#m = function() { 'test262' };`
- etc

It seems it produces a similar effect to ASTs as reported in #2066.

WDYT?

src/class-elements/private-field-as-function.case
caiolima: Ok!

src/class-elements/private-field-as-function.case
caiolima: I think you are right.

src/class-elements/private-field-as-function.case
caiolima: I think it is a good idea.

caiolima:
@leobalter thx for the review. I added more cases now. One important is ```src/class-elements/private-field-super-access-throws.case``` that has different behaviour from private accessors and private methods.

src/class-elements/private-field-as-async-arrow-function.case
```diff
@@ -0,0 +1,37 @@
...
+  return this.#m();
+}
+//- assertions
+let c = new C();
+
+async function asyncRun() {
+    assert.sameValue(await c.method(), 'test262');
+}
+
+asyncRun();
```
leobalter: this won't really work without a few updates. Test262 tests need to indicate when tests are async so the engine will really wait for completion.

My suggestion is, add a flag for async in the frontmatter: `flags: [async]`

now you need to call `$DONE()` for completion, and give a "truthy" argument for failures, e.g.: `$DONE(1)`

For these calls, the Promise syntax should probably be easier to follow:

```js
c.method().then((value) => assert.sameValue(value, 'test262'))
  .then($DONE, $DONE);
```


Using the current async function:

```js
asyncRun().then($DONE, $DONE);
```

or maybe (definitely not the best candidate):

```js
async function asyncRun() {
  try {
    assert.sameValue(await c.method(), 'test262');
    $DONE();
  } catch(e) { $DONE(e); }
}

asyncRun();
```

src/class-elements/private-field-as-async-function.case
```diff
@@ -0,0 +1,37 @@
+// Copyright (C) 2019 Caio Lima. All rights reserved.
+// Copyright (C) 2018 Jaideep Bhoosreddy (Bloomberg LP). All rights reserved.
+// Copyright (C) 2018 Leo Balter. All rights reserved.
```
leobalter: feel free to remove these extra copyright lines if you prefer.

src/class-elements/private-field-as-async-function.case
```diff
@@ -0,0 +1,37 @@
...
+
+method() {
+  return this.#m();
+}
+//- assertions
+let c = new C();
+
+async function asyncRun() {
+    assert.sameValue(await c.method(), 'test262');
+}
```
leobalter: same thing for async tests.

src/class-elements/private-field-super-access-throws.case
```diff
@@ -0,0 +1,42 @@
+// Copyright (C) 2019 Caio Lima. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: Acessing private field from super shoudl throw an error
```
leobalter: typo: s/shoudl/should

src/class-elements/private-field-super-access-throws.case
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  method() {
+    return super.#m();
+  }
+}
+
+//- assertions
+assert.throws(TypeError, function() {
+  (new (new C()).Child).method();
```
leobalter: total nitpicking, and please disregard this but... `new new C().Child().method()` works too

leobalter: actually `(new (new C()).Child()).method();` is more consistent (adding parentheses to Child).

leobalter:
> One important change is `src/class-elements/private-field-super-access-throws.case` that has different behaviour from private accessors and private methods.

Thanks for adding these! I've given a new round of review and we are pretty close. Just a quick fix for the async tests and we should be good to land this PR.

src/class-elements/private-field-as-async-arrow-function.case
caiolima: Thx for let me know this!

PR #2069, opened 2019-02-07
Split Intl.getCanonicalLocales from Locale

FrankYFTang:
I found some of our tests for Locale failed on the result of Intl.getCanonicalLocales which should not be part of the verification for Locale but instead for getCanonicalLocales so I MOVE these test from Locale under Intl/getCanonicalLocales to make us have better picture which PART is failing.


FrankYFTang:
@littledan @leobalter @anba

FrankYFTang:
@rwaldron @Ms2ger

FrankYFTang:
@gsathya @zbraniecki

FrankYFTang:
ping

PR #2073, opened 2019-02-08
Removed invalid super access tests and created properly syntax error

### LINKED ISSUE #2071: Private access from `super` is invalid syntax
Test `src/class-elements/private-field-super-access-throws.case` is verifying private fields super access. However this is invalid syntax and should throw SyntaxError instead.

Looking private fields proposal, we have changes on [`MemberExpression`](https://tc39.github.io/proposal-class-fields/#prod-MemberExpression), however this rule doesn't include `super.PrivateIdentifier` as valid. Check [https://tc39.github.io/ecma262/#prod-MemberExpression](https://tc39.github.io/ecma262/#prod-MemberExpression) for clarification.

caiolima:
This closes #2071.


src/class-elements/grammar-private-field-super-access.case
```diff
@@ -0,0 +1,26 @@
...
+template: syntax/invalid
+features: [class-fields-private]
+---*/
+
+//- elements
+#m = function() { return 'test262'; };
+
+Child = class extends C {
+  access() {
+    return super.#m;
```
leobalter: seems correct to me, but I'd like to confirm with @littledan before landing these tests.

littledan: That's right, super.#x is not in the grammar.

caiolima: I think we are good to move forward, right?

rwaldron:
@caiolima can you add a second commit with the regenerated tests? Thanks!!

caiolima:
Thx a lot!

PR #2075, opened 2019-02-08
Add tests for calling BigInt functions with fewer arguments than required

cxielarko:
BigInt.asIntN and BigInt.asUintN should throw TypeErrors when the required bigint argument is not present

See https://bugzilla.mozilla.org/show_bug.cgi?id=1526279

### DIFF
--- a/test/built-ins/BigInt/asIntN/bigint-tobigint-errors.js
+++ b/test/built-ins/BigInt/asIntN/bigint-tobigint-errors.js
@@ -11,6 +11,14 @@ features: [BigInt, computed-property-names, Symbol, Symbol.toPrimitive]
 ---*/
 assert.sameValue(typeof BigInt, 'function');
 assert.sameValue(typeof BigInt.asIntN, 'function');
+
+assert.throws(TypeError, function () {
+  BigInt.asIntN();
+}, "ToBigInt: no argument => undefined => TypeError");
+assert.throws(TypeError, function () {
+  BigInt.asIntN(0);
+}, "ToBigInt: no argument => undefined => TypeError");
+
 assert.throws(TypeError, function() {
   BigInt.asIntN(0, undefined);
 }, "ToBigInt: undefined => TypeError");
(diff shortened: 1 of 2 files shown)


test/built-ins/BigInt/asIntN/bigint-tobigint-errors.js
```diff
@@ -11,6 +11,14 @@ features: [BigInt, computed-property-names, Symbol, Symbol.toPrimitive]
 ---*/
 assert.sameValue(typeof BigInt, 'function');
 assert.sameValue(typeof BigInt.asIntN, 'function');
+
+assert.throws(TypeError, function () {
+  BigInt.asIntN();
+}, "ToBigInt: no argument => undefined => TypeError");
+assert.throws(TypeError, function () {
+  BigInt.asIntN(0);
+}, "ToBigInt: no argument => undefined => TypeError");
```
rwaldron: Is this intentional?

test/built-ins/BigInt/asUintN/bigint-tobigint-errors.js
```diff
@@ -12,6 +12,13 @@ features: [BigInt, computed-property-names, Symbol, Symbol.toPrimitive]
 assert.sameValue(typeof BigInt, 'function');
 assert.sameValue(typeof BigInt.asUintN, 'function');
 
+assert.throws(TypeError, function () {
+  BigInt.asUintN();
+}, "ToBigInt: no argument => undefined => TypeError");
+assert.throws(TypeError, function () {
+  BigInt.asUintN(0);
+}, "ToBigInt: no argument => undefined => TypeError");
```
rwaldron: Same question

test/built-ins/BigInt/asIntN/bigint-tobigint-errors.js
cxielarko: yes, it's due to asIntN using the default behavior for missing arguments, combined with ToBigInt(undefined) throwing an exception (ToBigInt only does implicit conversions for booleans and strings)

test/built-ins/BigInt/asUintN/bigint-tobigint-errors.js
cxielarko: (same reasoning as for asIntN)

PR #2081, opened 2019-02-22
Fix parseTestRecord.py for optional top license header

### LINKED ISSUE #2080: Fix parseTestRecord.py
Ref #2065

leobalter:
This should work for the hashbang tests.

Note that this is legacy code and I'm first trying to implement a fix and we can later improve this script. The change is conservative to preserve legacy functionality as I'm avoiding breaking other things.

Fixes #2080 

cc @caitp @gsathya I appreciate if you could take a look on this one, thanks.

### DIFF
--- a/tools/packaging/test/test_common.py
+++ b/tools/packaging/test/test_common.py
@@ -37,7 +37,7 @@ declaration;""", record['commentary'])
                          record['description'])
         self.assertEqual("", record['onlyStrict'])
         self.assertEqual("SyntaxError", record['negative'])
-        self.assertEqual("http://wiki.ecmascript.org/doku.php?id=conventions:no_non_standard_strict_decls",
+        self.assertEqual('http://wiki.ecmascript.org/doku.php?id=conventions:no_non_standard_strict_decls',
                          record['bestPractice'])
 
 
@@ -55,9 +55,23 @@ class TestYAMLParsing(unittest.TestCase):
         self.assertEqual(['onlyStrict'], record['flags'])
         self.assertEqual("", record['onlyStrict'])
         self.assertEqual("SyntaxError", record['negative'])
-        self.assertEqual("http://wiki.ecmascript.org/doku.php?id=conventions:no_non_standard_strict_decls",
+        self.assertEqual('"http://wiki.ecmascript.org/doku.php?id=conventions:no_non_standard_strict_decls"',
                          record['bestPractice'])
 
+    def test_no_header(self):
+        name = 'fixtures/test262-yaml-headers-no-cr.js'
+        contents = slurpFile(name)
+        record = convertDocString(contents)
+
+        self.assertEqual("The production Block { } in strict code can't contain function declaration;\n", record['commentary'])
+
+        self.assertEqual("Trying to declare function at the Block statement",
+                         record['description'])
+        self.assertEqual(['onlyStrict'], record['flags'])
+        self.assertEqual("", record['onlyStrict'])
+        self.assertEqual("SyntaxError", record['negative'])
+        self.assertEqual('"http://wiki.ecmascript.org/doku.php?id=conventions:no_non_standard_strict_decls"',
+                         record['bestPractice'])
 
 if __name__ == '__main__':
     unittest.main()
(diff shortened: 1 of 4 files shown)


gsathya:
@caitp @mathiasbynens ptal

tools/packaging/test/test_common.py
```diff
@@ -55,9 +55,23 @@ def test_overview(self):
         self.assertEqual(['onlyStrict'], record['flags'])
         self.assertEqual("", record['onlyStrict'])
         self.assertEqual("SyntaxError", record['negative'])
-        self.assertEqual("http://wiki.ecmascript.org/doku.php?id=conventions:no_non_standard_strict_decls",
+        self.assertEqual('"http://wiki.ecmascript.org/doku.php?id=conventions:no_non_standard_strict_decls"',
```
mathiasbynens: Shouldn’t the expectation be that this is unquoted? YAML doesn’t distinguish between:

```yaml
foo: 'bar'
```

and 

```yaml
foo: bar
```

tools/packaging/test/test_common.py
```diff
@@ -55,9 +55,23 @@ def test_overview(self):
...
+        record = convertDocString(contents)
+
+        self.assertEqual("The production Block { } in strict code can't contain function declaration;\n", record['commentary'])
+
+        self.assertEqual("Trying to declare function at the Block statement",
+                         record['description'])
+        self.assertEqual(['onlyStrict'], record['flags'])
+        self.assertEqual("", record['onlyStrict'])
+        self.assertEqual("SyntaxError", record['negative'])
+        self.assertEqual('"http://wiki.ecmascript.org/doku.php?id=conventions:no_non_standard_strict_decls"',
```
mathiasbynens: Same question here

tools/packaging/test/test_parseTestRecord.py
```diff
@@ -161,7 +160,7 @@ def test_overview(self):
         self.assertEqual(['onlyStrict'], record['flags'])
         self.assertEqual("", record['onlyStrict'])
         self.assertEqual("SyntaxError", record['negative'])
-        self.assertEqual("http://wiki.ecmascript.org/doku.php?id=conventions:no_non_standard_strict_decls",
+        self.assertEqual('"http://wiki.ecmascript.org/doku.php?id=conventions:no_non_standard_strict_decls"',
```
mathiasbynens: and here

website/default.html
```diff
@@ -0,0 +1,203 @@
+﻿<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
```
mathiasbynens: What is this?

The commit adding this has “delete unused files” as its message, but it doesn’t actually delete anything. I’m confused!

mathiasbynens:
There seem to be a lot of unrelated changes in this patch, which makes it a little hard to review

tools/packaging/test/test_common.py
leobalter: honestly, I was just restoring the original test to a passing state. These are legacy tests and I believe the current tools are just quoting the string. Which is not wrong, as you already mentioned.

I'd rather spend time fixing this in another PR if you agree.

website/default.html
leobalter: I meant to remove these files... not sure what is happening here

leobalter:
I removed the bad commits and squashed the other fixups.

caitp:
It generally looks good to me.

The copyright header regex looks extra strict, is the expected license formatting documented somewhere?

tools/packaging/test/test_common.py
mathiasbynens: > Which is not wrong, as you already mentioned.

Quoting the string in the input is fine, but expecting the quotes to still be there after "parsing" the YAML _is_ wrong. (I'm fine with this being fixed separately though.)

leobalter:
> The copyright header regex looks extra strict, is the expected license formatting documented somewhere?

@caitp not much else than in the [CONTRIBUTING.md](https://github.com/tc39/test262/blob/master/CONTRIBUTING.md#test-case-style) and the current convention from new tests.

This regexp I used is a copy and paste from [the linting script](https://github.com/tc39/test262/blob/master/tools/lint/lib/checks/license.py), which I believe accepts the formats identified in the pre-existing test files.

I reused this regexp to guarantee this works with whatever the linter will accept.

We can surely improve this parseTestRecord.py but this PR is just trying to fix one blocking problem without messing around with other usages.

Please, LMK if this works for you, otherwise I'll be happy to find a better solution.

tools/packaging/test/test_common.py
leobalter: I agree with you.

PR #2085, opened 2019-02-28
Update parseTestRecord to support hashbang tests

anba:
41edfcebce1a850968c38f9897d5b47a78a738f4
- Clean-up imports.
- Remove support for the old test format, because it's no longer used.
- Simplify the regular expressions (e.g. `[cC]` is equal to `C` when `re.IGNORECASE` is set; `r"\/\/"` can be simplified to `r"//"`, etc.)
- Add custom error handling, defaulting to use the  `print` function. This will help downstream users to catch any format errors.
- Don't try two different ways to find and parse the YAML frontmatter, instead simply search for the `/*--- [...] ---*/` block. This matches how the linter works and allows to parse all existing hashbang tests correctly (without this patch, the frontmatter is not found in all hashbang tests!).
- Add some additional format checks to ensure the license and frontmatter blocks are handled correctly. Combined with the new error handler, this may give downstream users a better chance to identify misformatted tests.
- Clean-up the test case to make it more readable.

I've verified `parseTestRecord` can now parse all tests and doesn't report any errors.

334e905d2e592984ce231ed31598920d0f874127
- Add license comments now that they're properly supported by `parseTestRecord`.


anba:
This issue is blocking <https://bugzilla.mozilla.org/show_bug.cgi?id=1531202>.

test/language/comments/hashbang/escaped-bang-041.js
```diff
@@ -1,4 +1,8 @@
 #\041
+
+// Copyright (C) 2019 Leo Balter. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
```
leobalter: seems that we can't simply forget these headers...

PR #2086, opened 2019-02-28
Fix existing linting errors, update error phase, and add a new flags linter

anba:
- While working on the other PR, I saw that there are some existing linting errors in "test/language/directive-prologue".
- And the SpiderMonkey test262 was choking on "test/language/comments/hashbang/use-strict.js", because the flags combination "raw" and "noStrict" confused it.
- And finally there were also some tests using the old "early" phase instead of "parse".

### DIFF
--- a/test/language/directive-prologue/10.1.1-2gs.js
+++ b/test/language/directive-prologue/10.1.1-2gs.js
@@ -13,5 +13,5 @@ flags: [raw]
 ---*/
 
 "use strict"
-throw new Error("This code should not execute");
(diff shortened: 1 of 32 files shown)


test/language/directive-prologue/10.1.1-2gs.js
```diff
@@ -13,5 +13,5 @@ flags: [raw]
 ---*/
 
 "use strict"
-throw new Error("This code should not execute");
+throw "Test262: This statement should not be evaluated.";
```
leobalter: +1

PR #2096, opened 2019-03-11
Test cases for async function and async generator edge-cases

anba:
Test cases for <https://bugzilla.mozilla.org/show_bug.cgi?id=1530754> and other related issues I found (or found again), like tc39/ecma262#1474 and tc39/ecma262#1470, while working on that bug. 

SpiderMonkey passes these tests locally with <https://bugzilla.mozilla.org/show_bug.cgi?id=1530754> fixed. (Some of the tests fail on V8/JSC, so don't be surprised if you see any errors.)

### DIFF
--- a/test/language/statements/async-generator/return-undefined-implicit-and-explicit.js
+++ b/test/language/statements/async-generator/return-undefined-implicit-and-explicit.js
@@ -1,0 +1,80 @@
+// Copyright (C) 2019 André Bargull. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-return-statement-runtime-semantics-evaluation
+description: >
+  Return with an explicit return value awaits this value.
+info: |
+  13.10.1 Runtime Semantics: Evaluation
+
+    ReturnStatement : return;
+      1. Return Completion { [[Type]]: return, [[Value]]: undefined, [[Target]]: empty }.
+
+    ReturnStatement : return Expression ;
+      1. Let exprRef be the result of evaluating Expression.
+      2. Let exprValue be ? GetValue(exprRef).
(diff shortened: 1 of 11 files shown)


test/language/statements/async-generator/return-undefined-implicit-and-explicit.js
```diff
@@ -0,0 +1,80 @@
...
+  "g4 ret",
+];
+
+var actual = [];
+
+Promise.resolve(0)
+  .then(() => actual.push("tick 1"))
+  .then(() => actual.push("tick 2"))
+  .then(() => {
+    assert.compareArray(actual, expected, "Ticks for implicit and explicit return undefined");
```
adrianheine: This doesn't actually test the return values, does it?

anba: Correct, it doesn't test that each of the async generators actually returns `undefined`. But that was actually intentional, because the test should only ensure that the async generators finish at the correct turn. :-)

adrianheine: Ok, that sounds right :) Then the assertion text should be changed.

test/language/statements/async-generator/yield-star-return-then-getter-ticks.js
```diff
@@ -0,0 +1,145 @@
...
+  "get then",
+
+  // `Await(resumptionValue.[[Value]])` promise resolved.
+  "tick 3",
+
+  // Get iterator "return" method.
+  "get return",
+
+  // "then" of `received.[[Value]]` accessed.
+  "get then",
```
leobalter: wow! This is tricky

PR #2097, opened 2019-03-11
Fix various test issues (Was: Unicode BCP 47 Locale Identifier changes)

anba:
ECMA-402 was changed to use "Unicode BCP 47 Locale Identifiers" instead of "BCP 47 Language Tags" in <https://github.com/tc39/ecma402/pull/289>. Some tests need to be updated because of that PR.

d1d17999353269dd896e6b42f0b76fb86dcf5937
- Updates the expected canonicalisation per <http://unicode.org/reports/tr35/#Language_Tag_to_Locale_Identifier>

c7f53698882a7a7ce556a1266caed83dab508b5c
- Updates parsing per <http://unicode.org/reports/tr35/#Unicode_locale_identifier> and <http://unicode.org/reports/tr35/#BCP_47_Conformance>

AFAIK this is nowhere (*) implemented, so testing and verifying these changes will be... interesting, I guess. Maybe it makes sense to wait until some other implementation also passes all changed tests, so we can double check the changes?

(*) Except locally in es6draft, but that doesn't count! :-)


anba:
Ugh. The updates are based on UTS35, rev34, but I just saw that rev35 will contain many changes for  Unicode BCP 47 Locale Identifiers, so it probably now makes even more sense to hold off merging this PR. :-/

littledan:
cc @Ms2ger

anba:
I've updated the PR to cover language tag canonicalisation and also fix other test bugs accumulated over time.

leobalter:
Thanks for the update!

PR #2099, opened 2019-03-11
Adding tests to verify private fields on Proxy Objects

### LINKED ISSUE #2098: Missing coverage of private fields on Proxy Objects
We should have tests to verify that we are properly adding private fields and methods to Proxy objects.

The following is a valid program:

```
let arr = [];
	
class ProxyBase {
  constructor() {
    return new Proxy(this, {
      get: function (obj, prop) {
        arr.push(prop);
          return obj[prop];
      }
    });
  }
}

class Test extends ProxyBase {
  #f() {
    return 3;
  }
  method() {
    return this.#f;
  }
}

let t = new Test();
let r = t.method(); // r is 3 here
```

caiolima:
It closes #2098.

### DIFF
--- a/test/language/statements/class/elements/privatefield-on-proxy.js
+++ b/test/language/statements/class/elements/privatefield-on-proxy.js
@@ -2,7 +2,7 @@
 // This code is governed by the BSD license found in the LICENSE file.
 
 /*---
-description: Successfully access private field on Proxy objects
+description: Sucessyfully get private reference without using [[Get]]
 esid: sec-getvalue
 info: |
   GetValue(V)
@@ -14,6 +14,7 @@ info: |
         ii. Let field be ? ResolveBinding(GetReferencedName(V), env).
         iii. Assert: field is a Private Name.
         iv. Return ? PrivateFieldGet(field, base).
+      c. Return ? base.[[Get]](GetReferencedName(V), GetThisValue(V)).
   PrivateFieldGet(P, O)
     1. Assert: P is a Private Name value.
     2. If O is not an object, throw a TypeError exception.
@@ -24,12 +25,8 @@ includes: [compareArray.js]
 features: [class, class-fields-private, Proxy]
 ---*/
 
-function assertArray(l, r) {
-  assert(compareArray(l, r), r);
-}
-
 let arr = [];
-	
+
 class ProxyBase {
   constructor() {
     return new Proxy(this, {
@@ -52,5 +49,5 @@ let t = new Test();
 let r = t.method();
 assert.sameValue(r, 3);
 
-assertArray(arr, ['method']);
+assert.compareArray(arr, ['method']);
 
--- a/test/language/statements/class/elements/privatefield-on-proxy.js
+++ b/test/language/statements/class/elements/privatefield-on-proxy.js
@@ -1,0 +1,56 @@
+// Copyright (C) 2019 Caio Lima. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Successfully access private field on Proxy objects
+esid: sec-getvalue
+info: |
+  GetValue(V)
+    ...
+    5. If IsPropertyReference(V), then
+      ...
+      b. If IsPrivateReference(V), then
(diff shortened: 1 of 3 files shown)


caiolima:
@littledan Could you take a look on those tests we discussed last week?

test/language/statements/class/elements/privatefield-on-proxy.js
```diff
@@ -0,0 +1,56 @@
...
+  method() {
+    return this.#f;
+  }
+}
+
+let t = new Test();
+let r = t.method();
+assert.sameValue(r, 3);
+
+assertArray(arr, ['method']);
```
leobalter: You don't need assertArray, you can use `assert.compareArray(l, r)` directly and it will handle a nice error message if necessary. Just preserve the included harness file (`compareArray.js`) and you have it.

test/language/statements/class/elements/privatemethods-on-proxy.js
```diff
@@ -0,0 +1,49 @@
...
+includes: [compareArray.js]
+features: [class, class-fields-private, Proxy]
+---*/
+
+function assertArray(l, r) {
+  assert(compareArray(l, r), r);
+}
+
+let arr = [];
+
```
leobalter: there is a trailing tab character here, invisible on the PR diff

test/language/statements/class/elements/privatefield-on-proxy.js
```diff
@@ -0,0 +1,56 @@
...
+    4. If entry is empty, throw a TypeError exception.
+    5. Return entry.[[PrivateFieldValue]].
+includes: [compareArray.js]
+features: [class, class-fields-private, Proxy]
+---*/
+
+function assertArray(l, r) {
+  assert(compareArray(l, r), r);
+}
+
```
leobalter: here too, a trailing tab character

test/language/statements/class/elements/privatefield-on-proxy.js
```diff
@@ -0,0 +1,56 @@
+// Copyright (C) 2019 Caio Lima. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Successfully access private field on Proxy objects
```
leobalter: I believe these descriptions should extend to something like: _Sucessyfully get private reference without using [[Get]]`_

and the part using `[[Get]]` could be used in the info:

```
Return ? base.[[Get]](GetReferencedName(V), GetThisValue(V)).
```

test/language/statements/class/elements/privatemethods-on-proxy.js
```diff
@@ -0,0 +1,49 @@
...
+esid: sec-privatefieldget
+info: |
+  PrivateFieldGet(P, O)
+    ...
+    4. Perform ? PrivateBrandCheck(O, P).
+    5. If P.[[Kind]] is "method",
+      a. Return P.[[Value]].
+      ...
+includes: [compareArray.js]
+features: [class, class-fields-private, Proxy]
```
leobalter: `class-methods-private` instead of class-fields-private

caiolima:
Thank you very much for the reviews and I'm sorry for the late reply. I got pretty busy with other tasks last week.

> Just a few small details.
> 
> Do you think this should be extended to cases like `get #f() { return 3; }` as well?

Yes. I'll add them as well.

test/language/statements/class/elements/privatemethods-on-proxy.js
caiolima: Oops.

test/language/statements/class/elements/privatefield-on-proxy.js
caiolima: Oh nice.

test/language/statements/class/elements/privatemethods-on-proxy.js
caiolima: Fixed.

test/language/statements/class/elements/privatefield-on-proxy.js
caiolima: Fixed.

PR #2101, opened 2019-03-13
Use a line terminator for HTMLCloseComments

### LINKED ISSUE #2095: Incorrect tests: createdynfn-html-close-comment-(body|params).js
[test/annexB/built-ins/Function/createdynfn-html-close-comment-body.js](https://github.com/tc39/test262/blob/master/test/annexB/built-ins/Function/createdynfn-html-close-comment-body.js) and [test/annexB/built-ins/Function/createdynfn-html-close-comment-params.js](https://github.com/tc39/test262/blob/master/test/annexB/built-ins/Function/createdynfn-html-close-comment-params.js) i.e. `Function("-->");` and `Function("-->", "");` are supposed to test HTML-like comments and their parsing is expected to pass. Unfortunately, they are incorrect because `-->` is not a valid comment. Note that [HTMLCloseComment](https://tc39.github.io/ecma262/#prod-annexB-HTMLCloseComment) can be either part of [SingleLineHTMLCloseComment](https://tc39.github.io/ecma262/#prod-annexB-SingleLineHTMLCloseComment) or [MultiLineComment](https://tc39.github.io/ecma262/#prod-annexB-MultiLineComment). So it has to be preceeded by `LineTerminatorSequence` (to be `SingleLineHTMLCloseComment`) or by `*/` (to have a chance to be `MultiLineComment`). Hence, `-->` cannot be parsed as [Comment](https://tc39.github.io/ecma262/#prod-annexB-Comment) and the mentioned tests should result in `SyntaxError` instead.

Side note: It seems that major JavaScript engines share my opinion regarding `createdynfn-html-close-comment-params.js` test, i.e., all of them [fail it](https://test262.report/browse/annexB/built-ins/Function/createdynfn-html-close-comment-params.js). The only reason that some major JavaScript engines [pass](https://test262.report/browse/annexB/built-ins/Function/createdynfn-html-close-comment-body.js) `createdynfn-html-close-comment-body.js` test is IMHO the fact that they tend to parse `Function("params","body")` as `function anonymous(params\n) {\nbody\n}`, i.e., add an artificial line terminator before the body (which turns the body of the form `-->` into a valid comment effectively).

leobalter:
Closes #2095


rwaldron:
This bug was certainly educational.

PR #2102, opened 2019-03-13
Add new tests to observe required leading line terminator

leobalter:
Ref #2095

cc @anba @gnarf @iamstolis @gsathya @rwaldron


leobalter:
Talking to @gnarf we decided to open a patch on tc39/ecma262, changing the item 18 to wrap the bodyText between line feeds before parsing it.

This will match the current behavior or V8, SpiderMonkey, and JSC. At the same time, this change will become beneficial to avoid viciously successful parsings of body text contents that are actual hashbangs. `Function("!#")`

rwaldron:
Excellent outcome

leobalter:
I'm merging this and opening a new PR to observe proposed changes for #1479 on top.

PR #2103, opened 2019-03-15
Ensure Atomics functions validate the TypedArray type before coercing the other arguments

anba:
Note: SpiderMonkey currently fails these tests (except for `Atomics.wait` and `Atomics.notify`).

### DIFF
--- a/test/built-ins/Atomics/and/validate-arraytype-before-index-coercion.js
+++ b/test/built-ins/Atomics/and/validate-arraytype-before-index-coercion.js
@@ -1,0 +1,43 @@
+// Copyright (C) 2019 André Bargull. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-atomics.add
+description: >
+  TypedArray type is validated before `index` argument is coerced.
+info: |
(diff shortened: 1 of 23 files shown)


test/built-ins/Atomics/and/validate-arraytype-before-index-coercion.js
```diff
@@ -0,0 +1,43 @@
...
+
+var index = {
+  valueOf() {
+    throw new Test262Error("index coerced");
+  }
+};
+
+var badArrayTypes = [
+  Uint8ClampedArray, Float32Array, Float64Array
+];
```
rwaldron: This is 2/3 a duplicate of an existing test :|

rwaldron:
Each of these directories has a file called "shared-nonint-views.js" which does nearly the same thing as these tests you've added. Your new tests are more thorough, so I'd like to keep the new bits, but you'll need to either: 

1. Delete each of the `shared-nonint-views.js` files (likely the easiest)
2. Combine your new additions to each of the `shared-nonint-views.js` files (a lot of work)

My vote is for 1.

rwaldron:
@anba ping

PR #2104, opened 2019-03-16
Add tests for termination of "hashbang" comment

jugglinmike:


jugglinmike:
I almost forgot about ECMAScript parsers. How could I?

At the risk of making this test completely inscrutable, I've updated the "pass signal" to come from an early error. That should ensure that parsers also correctly detect the end of the comment. I realize that this extends the semantics under test, but I think a slight mixing of concerns is justified by the extended reach it provides. My bigger concern is if this pattern will make sense to others; do you folks think the explanation in the "info" metadata is sufficient?

adrianheine:
That's pretty ingenious :) How about test cases that only _pass_ if the hashbang line is correctly detected, like this:
```
#!xyz LINE_TERMINATOR {
/* frontmatter */
}

jugglinmike:
That's much more clear. It has some risk of spurious passes if the consumer incorrectly ignores too much, but that same could be said for every test. Great idea!

PR #2105, opened 2019-03-17
Refactor white space tests for parsers

jugglinmike:
The tests for the parsing of white space were expressed using eval. This made the tests more complex than necessary and also prevented the tests from providing value to ECMAScript parsers.

Remove the use of eval and instead express the expectations with literal source text.


jugglinmike:

test/language/white-space/between-vertical-tab.js
```diff
@@ -7,14 +7,6 @@ es5id: 7.2_A1.2_T2
 description: Insert real VERTICAL TAB between tokens of var x=1
 ---*/
 
-//CHECK#1
 varx=1;
```
leobalter: what a beautiful code! I could use this to always find whitespace in my code.

test/language/white-space/between-form-feed.js
```diff
@@ -7,14 +7,6 @@ es5id: 7.2_A1.3_T2
 description: Insert real FORM FEED between tokens of var x=1
 ---*/
 
-//CHECK#1
 varx=1;
```
leobalter: and this should always be used for code golfing.

PR #2106, opened 2019-03-20
Add CircleCI

leobalter:
This would help us having checks for text execution on each PR


leobalter:
This is fair enough for now and I need to have this tested in the master branch.

PR #2109, opened 2019-03-26
Fix tests for HTML comments in function body evaluation

leobalter:
Ref tc39/ecma262#1479


leobalter:
The respective spec has been merged in Nov 2019

PR #2110, opened 2019-03-26
Circle CI maintenance, fix engines path

leobalter:


leobalter:
2b7c14a should be removed from this PR before it's merged, I'm using it to get results from the CI

leobalter:
Docker used by Circle makes it not too friendly with the .jsvu path, so I'm copying the whole stuff to my work folder...

PR #2111, opened 2019-03-26
Update list pattern to sync with CLDR 35

FrankYFTang:
update list pattern to sync with CLDR 35

CLDR 35 change "and" to "&" for the list form pattern "short"
see https://unicode.org/cldr/trac/ticket/11520
Also compare
https://www.unicode.org/repos/cldr/tags/release-34/common/main/en.xml
against
https://www.unicode.org/repos/cldr/tags/release-35-beta/common/main/en.xml

Notice the addition of
```
<listPattern type="standard-short">
<listPatternPart type="start">{0}, {1}</listPatternPart>
<listPatternPart type="middle">{0}, {1}</listPatternPart>
<listPatternPart type="end">{0}, & {1}</listPatternPart>
<listPatternPart type="2">{0} & {1}</listPatternPart>
</listPattern>
```


FrankYFTang:
@zbraniecki @anba @littledan @leobalter

PR #2114, opened 2019-03-29
Add tests for string representations of Date objects with negative years

gibson042:
Ref https://github.com/tc39/ecma262/issues/1035


leobalter:
I'd recommend checking for open issues in the engines shown as failing here. This is the kind of change that can land without much noise otherwise and tests would fail for a long time.

gibson042:
Looks like that's V8, JSC, and XS. Let's take the opportunity to document in this repository instructions on how to report bugs and links to the relevant trackers, because I don't want to look up how to do so for XS more than once—can you take #2121?

PR #2118, opened 2019-04-01
 Refactor `with` statement tests for parsers 

jugglinmike:
The tests for the parsing of the `with` statement were expressed using
eval. This made the tests more complex than necessary and also prevented
the tests from providing value to ECMAScript parsers.

Remove the use of eval and instead express the expectations with literal
source text. Rename the files to make each test's purpose more clear.

(Also removed a few redundant tests.)


leobalter:
This is the kind of PR I triple check if you open on this day.

Not a joke! I’ll finish a review soon.

PR #2119, opened 2019-04-02
 Update test to allow 'narrow' style for all types 

FrankYFTang:
Based on the changes in tc39/proposal-intl-list-format#43
We now allow  'narrow' style for all types.
Change testing expectation
and remove test which is obsoleted and conditions that are covered in the above test.


FrankYFTang:
@littledan @anba @zbraniecki @leobalter

PR #2123, opened 2019-04-03
correct the test case

FrankYFTang:
I somehow missed the last line last time I proposed https://github.com/tc39/test262/pull/2111


FrankYFTang:
@littledan @anba @zbraniecki @leobalter

PR #2126, opened 2019-04-09
Initial tests for BigInt.prototype.toLocaleString().

Ms2ger:
Many of these tests are heavily based on the tests for
Number.prototype.toLocaleString().

### DIFF
--- a/test/intl402/BigInt/prototype/toLocaleString/builtin.js
+++ b/test/intl402/BigInt/prototype/toLocaleString/builtin.js
@@ -1,5 +1,5 @@
 // Copyright 2012-2019 Mozilla Corporation; Igalia, S.L. All rights reserved.
-// This code is governed by the license found in the LICENSE file.
+// This code is governed by the BSD license found in the LICENSE file.
 
 /*---
 esid: sec-bigint.prototype.tolocalestring
--- a/test/intl402/BigInt/prototype/toLocaleString/builtin.js
+++ b/test/intl402/BigInt/prototype/toLocaleString/builtin.js
@@ -1,0 +1,28 @@
+// Copyright 2012-2019 Mozilla Corporation; Igalia, S.L. All rights reserved.
+// This code is governed by the license found in the LICENSE file.
+
+/*---
+esid: sec-bigint.prototype.tolocalestring
+description: >
+    Tests that BigInt.prototype.toLocaleString meets the requirements
+    for built-in objects defined by the introduction of chapter 17 of
+    the ECMAScript Language Specification.
+author: Norbert Lindenberg
+includes: [isConstructor.js]
+features: [Reflect.construct, BigInt]
+---*/
+
+assert.sameValue(Object.prototype.toString.call(BigInt.prototype.toLocaleString), "[object Function]",
+                 "The [[Class]] internal property of a built-in function must be " +
+                 "\"Function\".");
+
+assert(Object.isExtensible(BigInt.prototype.toLocaleString),
+       "Built-in objects must be extensible.");
+
+assert.sameValue(Object.getPrototypeOf(BigInt.prototype.toLocaleString), Function.prototype);
+
+assert.sameValue(BigInt.prototype.toLocaleString.hasOwnProperty("prototype"), false,
+                 "Built-in functions that aren't constructors must not have a prototype property.");
+
+assert.sameValue(isConstructor(BigInt.prototype.toLocaleString), false,
+                 "Built-in functions don't implement [[Construct]] unless explicitly specified.");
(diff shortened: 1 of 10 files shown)


Ms2ger:
One bug in V8: it doesn't call `ThisBigIntValue()` in `toLocaleString()`.

CC @gsathya @FrankYFTang

test/intl402/BigInt/prototype/toLocaleString/builtin.js
```diff
@@ -0,0 +1,28 @@
+// Copyright 2012-2019 Mozilla Corporation; Igalia, S.L. All rights reserved.
+// This code is governed by the license found in the LICENSE file.
```
FrankYFTang: I got the following error
"No license found in: intl402/BigInt/prototype/toLocaleString/builtin"

test/intl402/BigInt/prototype/toLocaleString/taint-Intl-NumberFormat.js
```diff
@@ -0,0 +1,14 @@
+// Copyright 2013-2019 Mozilla Corporation; Igalia, S.L. All rights reserved.
+// This code is governed by the license found in the LICENSE file.
```
FrankYFTang: Got a "No license found in: intl402/BigInt/prototype/toLocaleString/taint-Intl-NumberFormat"
I think you need to change the line "// This code is governed by the license found in the LICENSE file." to
something like "// This code is governed by the BSD license found in the LICENSE file."

test/intl402/BigInt/prototype/toLocaleString/this-value-invalid.js
```diff
@@ -0,0 +1,28 @@
...
+  NaN,
+  -Infinity,
+  {valueOf: function () { return 5; }},
+  {valueOf: function () { return 5n; }},
+];
+
+for (const value of invalidValues) {
+  assert.throws(TypeError, function() {
+    BigInt.prototype.toLocaleString.call(value);
+  }, "BigInt.prototype.toLocaleString did not reject this = " + value + ".");
```
FrankYFTang: the ' + value + ' won't work for the test case of Symbol()

FrankYFTang:
> One bug in V8: it doesn't call `ThisBigIntValue()` in `toLocaleString()`.
> 
> CC @gsathya @FrankYFTang

Just FYI
https://bugs.chromium.org/p/v8/issues/detail?id=9109 for v8 bug and
https://chromium-review.googlesource.com/c/v8/v8/+/1559214 for fix in v8

test/intl402/BigInt/prototype/toLocaleString/this-value-invalid.js
```diff
@@ -0,0 +1,28 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-bigint.prototype.tolocalestring
+description: Tests that toLocaleString handles "thisBigIntValue" correctly.
+features: [BigInt]
+---*/
+
+var invalidValues = [
+//  undefined,
```
FrankYFTang: Please either remove the // or remove the whole line

littledan:
It would be nice to additionally include tests that:
- Some locale other than en-US works, e.g., using the appropriate thousands separator
- Some value outside of Number range works, without rounding

gsathya:
@Ms2ger I still get the missing license error that @FrankYFTang mentioned. These tests need to be updated:

```
No license found in: intl402/BigInt/prototype/toLocaleString/builtin
No license found in: intl402/BigInt/prototype/toLocaleString/default-options-object-prototype
No license found in: intl402/BigInt/prototype/toLocaleString/length
No license found in: intl402/BigInt/prototype/toLocaleString/returns-same-results-as-NumberFormat
No license found in: intl402/BigInt/prototype/toLocaleString/taint-Intl-NumberFormat
No license found in: intl402/BigInt/prototype/toLocaleString/this-value-invalid
No license found in: intl402/BigInt/prototype/toLocaleString/throws-same-exceptions-as-NumberFormat
```

Ms2ger:
@gsathya I can't reproduce that locally on master. Could you clarify where you see those errors?

Ms2ger:
More tests: https://github.com/tc39/test262/pull/2136

gsathya:
> @gsathya I can't reproduce that locally on master. Could you clarify where you see those errors?

@Ms2ger, these tests break the license regexp in the test262 harness: 
https://github.com/test262-utils/test262-harness-py/blob/master/src/parseTestRecord.py#L27

Ms2ger:
https://github.com/test262-utils/test262-harness-py/pull/6

PR #2128, opened 2019-04-12
Add a couple more tests for ASI with fields.

jorendorff:
I didn't see any tests for this case.

These tests are slightly silly, but I would check these into our repo without hesitation, so I figure there's no reason not to at least offer them.


Ms2ger:
ping @leobalter

littledan:
Thanks for the tests, @jorendorff , and for being flexible about landing them, @leobalter .

PR #2129, opened 2019-04-17
Refactor module _FIXTURE.js to have no harness dependencies

rwaldron:
```
rwaldron in ~/clonez/test262 on master
$ test262_runner test/language/module-code/instn-iee-bndng-{cls,const,fun,gen,let,var}.js
-----------------------------------------------------------------------------------------
V8 (jsvu)

test262-harness --hostArgs='' --hostType=d8 --hostPath=/Users/rwaldron/.jsvu/v8 test/language/module-code/instn-iee-bndng-cls.js test/language/module-code/instn-iee-bndng-const.js test/language/module-code/instn-iee-bndng-fun.js test/language/module-code/instn-iee-bndng-gen.js test/language/module-code/instn-iee-bndng-let.js test/language/module-code/instn-iee-bndng-var.js

Ran 12 tests
12 passed
0 failed

-----------------------------------------------------------------------------------------
ChakraCore (jsvu)

test262-harness --hostArgs='' --hostType=ch --hostPath=/Users/rwaldron/.jsvu/chakra test/language/module-code/instn-iee-bndng-cls.js test/language/module-code/instn-iee-bndng-const.js test/language/module-code/instn-iee-bndng-fun.js test/language/module-code/instn-iee-bndng-gen.js test/language/module-code/instn-iee-bndng-let.js test/language/module-code/instn-iee-bndng-var.js

Ran 12 tests
12 passed
0 failed

-----------------------------------------------------------------------------------------
JavaScriptCore (jsvu)

test262-harness --hostArgs='' --hostType=jsc --hostPath=/Users/rwaldron/.jsvu/javascriptcore test/language/module-code/instn-iee-bndng-cls.js test/language/module-code/instn-iee-bndng-const.js test/language/module-code/instn-iee-bndng-fun.js test/language/module-code/instn-iee-bndng-gen.js test/language/module-code/instn-iee-bndng-let.js test/language/module-code/instn-iee-bndng-var.js

Ran 12 tests
12 passed
0 failed

-----------------------------------------------------------------------------------------
SpiderMonkey (jsvu)

[...]

PR #2130, opened 2019-04-17
INTERPRETING.md: explicitly disallow use of Test262-Defined Bindings and Host-Defined Functions in _FIXTURE.js files

rwaldron:
@jugglinmike @leobalter This change simply codifies the predominant pattern in Test262. There were 6 total _FIXTURE.js files that used assertions, but those have been updated in #2129.

### DIFF
--- a/INTERPRETING.md
+++ b/INTERPRETING.md
@@ -137,9 +137,9 @@ This must precede any additional text modifications described by test metadata.
 Test262 includes tests for ECMAScript 2015 module code, denoted by the "module"
 metadata flag. Files bearing a name ending in `_FIXTURE.js`:
 
-- **MUST** not be interpreted as standalone tests; they are intended to be referenced by test files.
-- **MUST** not refer to, or make use of any [Test262-Defined Bindings](#test262-defined-bindings) in any way. 
-- **MUST** not refer to, or make use of any [Host-Defined Functions](#host-defined-functions) in any way. 
+- **MUST NOT** be interpreted as standalone tests; they are intended to be referenced by test files.
+- **MUST NOT** refer to, or make use of any [Test262-Defined Bindings](#test262-defined-bindings) in any way. 
+- **MUST NOT** refer to, or make use of any [Host-Defined Functions](#host-defined-functions) in any way. 
 
 All module specifiers used by Test262 begin with the character sequence `./`.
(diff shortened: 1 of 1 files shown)


INTERPRETING.md
```diff
@@ -135,9 +135,11 @@ This must precede any additional text modifications described by test metadata.
...
 
 Test262 includes tests for ECMAScript 2015 module code, denoted by the "module"
-metadata flag. Files bearing a name ending in `_FIXTURE.js` should not be
-interpreted as standalone tests; they are intended to be referenced by test
-files.
+metadata flag. Files bearing a name ending in `_FIXTURE.js`:
+
+- **MUST** not be interpreted as standalone tests; they are intended to be referenced by test files.
+- **MUST** not refer to, or make use of any [Test262-Defined Bindings](#test262-defined-bindings) in any way. 
+- **MUST** not refer to, or make use of any [Host-Defined Functions](#host-defined-functions) in any way.
```
leobalter: thanks for replacing should with must. That's accurate.

What do you think of highlighting `**MUST NOT**` instead? It sounds more clear.

rwaldron: Done!

INTERPRETING.md
```diff
@@ -135,9 +135,11 @@ This must precede any additional text modifications described by test metadata.
...
 
 Test262 includes tests for ECMAScript 2015 module code, denoted by the "module"
-metadata flag. Files bearing a name ending in `_FIXTURE.js` should not be
-interpreted as standalone tests; they are intended to be referenced by test
-files.
+metadata flag. Files bearing a name ending in `_FIXTURE.js`:
+
+- **MUST NOT** be interpreted as standalone tests; they are intended to be referenced by test files.
+- **MUST NOT** refer to, or make use of any [Test262-Defined Bindings](#test262-defined-bindings) in any way. 
+- **MUST NOT** refer to, or make use of any [Host-Defined Functions](#host-defined-functions) in any way.
```
jugglinmike: @rwaldron Come to think of it, the last two bullet points are actually contribution guidelines. For this document, we might want to instead extend the instruction about "must not be interpreted as standalone tests" to say how the Realm modifications described elsewhere in this document must not be applied.

Sorry to be getting to this so late

PR #2137, opened 2019-04-24
Set npm scripts to run tests

leobalter:
This is just helper scripts with parts I use everyday.

@mathiasbynens Ideally I'd love to set jsvu to use a custom folder inside my project's path. This would be very helpful for the CI tools as well as CircleCI's docker get's bonkers with the current path outside, regardless I add anything to the PATH for jsvu.

the `npm test` is literally just a shortcut for the local test262-harness.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,14 +18,13 @@
   },
   "scripts": {
     "ci": "./tools/scripts/ci_test.sh",
-    "jsvu": "jsvu",
     "test": "test262-harness",
     "diff": "git diff --diff-filter ACMR --name-only master -- test/",
     "test:diff": "npm run test:diff:v8 && npm run test:diff:spidermonkey && npm run test:diff:chakra && npm run test:diff:javascriptcore",
     "test:diff:v8": "test262-harness -t 8 --hostType=d8 --hostPath=v8 $(npm run diff)",
     "test:diff:spidermonkey": "test262-harness -t 8 --hostType=jsshell --hostPath=spidermonkey $(npm run diff)",
-    "test:diff:chakra": "test262-harness -t 8 --hostType=d8 --hostPath=chakra $(npm run diff)",
-    "test:diff:javascriptcore": "test262-harness -t 8 --hostType=d8 --hostPath=javascriptcore $(npm run diff)",
-    "test:diff:xs": "test262-harness -t 8 --hostType=d8 --hostPath=xs $(npm run diff)"
+    "test:diff:chakra": "test262-harness -t 8 --hostType=ch --hostPath=chakra $(npm run diff)",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -17,6 +17,15 @@
...
   "scripts": {
-    "test": "./tools/scripts/ci_test.sh"
+    "ci": "./tools/scripts/ci_test.sh",
+    "jsvu": "jsvu",
+    "test": "test262-harness",
+    "diff": "git diff --diff-filter ACMR --name-only master -- test/",
+    "test:diff": "npm run test:diff:v8 && npm run test:diff:spidermonkey && npm run test:diff:chakra && npm run test:diff:javascriptcore",
+    "test:diff:v8": "test262-harness -t 8 --hostType=d8 --hostPath=v8 $(npm run diff)",
+    "test:diff:spidermonkey": "test262-harness -t 8 --hostType=jsshell --hostPath=spidermonkey $(npm run diff)",
+    "test:diff:chakra": "test262-harness -t 8 --hostType=d8 --hostPath=chakra $(npm run diff)",
```
rwaldron: `--hostType=d8` => `--hostType=ch`

package.json
```diff
@@ -17,6 +17,15 @@
...
+    "test:diff:javascriptcore": "test262-harness -t 8 --hostType=d8 --hostPath=javascriptcore $(npm run diff)",
```
rwaldron: `--hostType=d8` => `--hostType=jsc`

package.json
```diff
@@ -17,6 +17,15 @@
...
+    "test:diff:xs": "test262-harness -t 8 --hostType=d8 --hostPath=xs $(npm run diff)"
```
rwaldron: `--hostType=d8` => `--hostType=xs`

package.json
```diff
@@ -17,6 +17,15 @@
     "test262-harness": "^6.2.0"
   },
   "scripts": {
-    "test": "./tools/scripts/ci_test.sh"
+    "ci": "./tools/scripts/ci_test.sh",
+    "jsvu": "jsvu",
```
rwaldron: This seems unnecessary

leobalter: I only had it in a hope to eventually being able to hook jsvu usage in a local folder, so I could call a npm script to install engines locally as dependencies before running the tests.

So far, jsvu serves the CircleCI project, otherwise it's hard to convince anyone for other usages.

mathiasbynens: I’m not familiar with your particular setup but it really sounds like it should be possible to access and run things in your home directory, even within [the CircleCI environment](https://circleci.com/docs/2.0/circleci-images/). I’m hoping we can figure this out, and avoid adding flags to `jsvu` as a workaround.

leobalter: This is such a low priority at this point - I'm unblocked to use jsvu where I need to - so I'm not sure we should worry about this for now. Maybe another time.

PR #2134, opened 2019-04-19
Add tests for Intl.DateTimeFormat formatRange and formatRangeToParts

FrankYFTang:
I did copy some code from pre-existing tests. I am not sure how to make the copyright / license text correctly for that. Please comment about that part.

@leobalter @anba @littledan @Ms2ger @gsathya

PR #2139, opened 2019-04-25
Add tests for Intl.DateTimeFormat formatRange

Ms2ger:


Ms2ger:
To add to https://github.com/tc39/test262/pull/2134

Ms2ger:
cc @FrankYFTang

test/intl402/DateTimeFormat/prototype/formatRange/argument-date-string.js
```diff
@@ -0,0 +1,38 @@
...
+
+const dtf = new Intl.DateTimeFormat();
+const dateTimeString = "2017-11-10T14:09:00.000Z";
+const date = new Date(dateTimeString);
+// |dateTimeString| is valid ISO-8601 style date/time string.
+assert.notSameValue(date, NaN);
+
+// Ensure string input values are not converted to time values by calling the
+// Date constructor.
+assert.throws(RangeError, function() {
```
leobalter: It would be helpful to tell that `"2017-11-10T14:09:00.000Z"` would be read as NaN.

test/intl402/DateTimeFormat/prototype/formatRange/en-US.js
```diff
@@ -0,0 +1,36 @@
...
+description: Basic tests for the en-US output of formatRange()
+info: >
+  Intl.DateTimeFormat.prototype.formatRange ( startDate , endDate )
+
+  8. Return ? FormatDateTimeRange(dtf, x, y).
+locale: [en-US]
+features: [Intl.DateTimeFormat-formatRange]
+---*/
+
+const date1 = new Date("2019-01-03");
```
FrankYFTang: I believe these need to be
```
const date1 = new Date("2019-01-03T00:00");
const date2 = new Date("2019-01-05T00:00");
const date3 = new Date("2019-03-04T00:00");
const date4 = new Date("2020-03-04T00:00");
```
to make it clear that date is in local time. Otherwise, if we use new Date("2019-01-03"); on a Machine in PST. The date object is in UTC time but the formatted result will be in local time zone. 

This is because in 
https://ecma-international.org/ecma-262/#sec-date-time-string-format
it stated:
```
... When the time zone offset is absent, date-only forms are interpreted as a UTC time and date-time 
forms are interpreted as a local time."
```

test/intl402/DateTimeFormat/prototype/formatRangeToParts/argument-to-integer.js
```diff
@@ -0,0 +1,53 @@
...
+}
+
+// Switch to a time format instead of using DateTimeFormat's default date-only format.
+const dtf = new Intl.DateTimeFormat(undefined, {
+    hour: "numeric", minute: "numeric", second: "numeric"
+});
+const date = Date.now();
+const expected = dtf.formatRangeToParts(0, date);
+
+compare(dtf.formatRangeToParts(-0.9, date), expected, "formatToParts(-0.9)");
```
FrankYFTang: the compare function has only 2 args but in here you pass in 3 args
Also the third arg should be "formatRangeToParts" not "formatToParts"

test/intl402/DateTimeFormat/prototype/formatRangeToParts/en-US.js
```diff
@@ -0,0 +1,199 @@
...
+    assert.sameValue(actualEntry.type, expectedEntry.type, `type for entry ${i}`);
+    assert.sameValue(actualEntry.value, expectedEntry.value, `value for entry ${i}`);
+    assert.sameValue(actualEntry.source, expectedEntry.source, `source for entry ${i}`);
+  }
+}
+
+const date1 = new Date("2019-01-03");
+const date2 = new Date("2019-01-05");
+const date3 = new Date("2019-03-04");
+const date4 = new Date("2020-03-04");
```
FrankYFTang: I believe these need to be
```
const date1 = new Date("2019-01-03T00:00");
const date2 = new Date("2019-01-05T00:00");
const date3 = new Date("2019-03-04T00:00");
const date4 = new Date("2020-03-04T00:00");
```
to make it clear that date is in local time. Otherwise, if we use new Date("2019-01-03"); on a Machine in PST. The date object is in UTC time but the formatted result will be in local time zone. 

This is because in 
https://ecma-international.org/ecma-262/#sec-date-time-string-format
it stated:
```
... When the time zone offset is absent, date-only forms are interpreted as a UTC time and date-time forms 
 are interpreted as a local time."
```

test/intl402/DateTimeFormat/prototype/formatRangeToParts/prop-desc.js
```diff
@@ -0,0 +1,17 @@
+// Copyright 2016-2019 Mozilla Corporation, Igalia S.L. All rights reserved.
+// This code is governed by the license found in the LICENSE file.
+
+/*---
+description: Property type and descriptor.
+includes: [propertyHelper.js]
+---*/
```
FrankYFTang: Please add line
features: [Intl.DateTimeFormat-formatRange]

test/intl402/DateTimeFormat/prototype/formatRangeToParts/en-US.js
```diff
@@ -27,10 +27,10 @@ function compare(actual, expected) {
   }
 }
 
-const date1 = new Date("2019-01-03");
-const date2 = new Date("2019-01-05");
-const date3 = new Date("2019-03-04");
-const date4 = new Date("2020-03-04");
+const date1 = new Date("2019-01-03T00:00:00Z");
```
FrankYFTang: If you use "2019-01-03T00:00:00Z" here, it mean the date value is in UTC. But when you format the date, the date will be output according to local timezone, therefore it won't be 2019-01-03 in some timezone the machine is running. I think you need to use "2019-01-03T00:00:00" or "2019-01-03T00:00" instead so the date value will be in locale timezone of 2019-01-03. For example, my machine is in PST. if you use  "2019-01-03T00:00:00Z" then date 1 is UTC of 2019-01-03 00:00 which is 2019-01-02 16:00:00 in PST. then when you format it, sine the dft is not created for a particular timezone, it will use local time zone (PST) to format, and therefore output as 2019/1/2 instead of 2019/1/3 as what you expected. In the other hand, if you use new Date("2019-01-03T00:00") , ECMA 262 said it will be interpret as using local time zone. Notice new Date("2019-01-03"), which you previously use, will be created as in UTC, not local timezone.

test/intl402/DateTimeFormat/prototype/formatRange/en-US.js
```diff
@@ -12,10 +12,10 @@ locale: [en-US]
 features: [Intl.DateTimeFormat-formatRange]
 ---*/
 
-const date1 = new Date("2019-01-03");
-const date2 = new Date("2019-01-05");
-const date3 = new Date("2019-03-04");
-const date4 = new Date("2020-03-04");
+const date1 = new Date("2019-01-03T00:00:00Z");
```
FrankYFTang: Since you are checking the output value of the formatRange, please either use const date1 = new Date("2019-01-03T00:00") so the date object is created in local timezone not UTC. OR you specify the timezone of dtf in UTC so it will match. Otherwise, it won't be 1/3 in some timezone. Create the date1 ... date4 and your format output in the same time zone- either both in UTC or both in local timezone. See my comments in formatRangeToParts/en-US.js for more.

test/intl402/DateTimeFormat/prototype/formatRange/argument-tonumber-throws.js
```diff
@@ -0,0 +1,54 @@
+// Copyright 2016-2019 Leonardo Balter, Igalia S.L. All rights reserved.
```
jorydotcom: @Ms2ger this reads as though @leobalter works at Igalia; can you move the second attribution to a different line?

leobalter: I confirm it's weird to see both names together.

I don't mind if you remove my name attribution and keep it to a single line.

Also we conventionally go with a single year for the copyright, where the range is usually for legacy tests. I believe the following should be ideal here:

```
// Copyright 2019 Igalia S.L. All rights reserved.
```

test/intl402/DateTimeFormat/prototype/formatRangeToParts/en-US.js
Ms2ger: Ugh.

leobalter:
I know @spectranaut is reviewing this PR today.

@Ms2ger Please, fix the Copyright issues in the other files as well.

PR #2140, opened 2019-04-25
Use .. for git diffs in the CI routines

leobalter:
the double dot gets only commits not in the left side of the double dot.

The right side assumes HEAD.

This will allow us to fetch the correct list of new or modified files in the current branch.

Ref #2137


leobalter:
Thanks a lot to @gnarf who helped me with this issue. This was becoming annoying where some PRs were checking for files recently introduced in the master branch but not yet in the PR.

PR #2145, opened 2019-05-02
Adding case to verify if PrivateEnvironment is properly set when evaluating 'ClassHeritage'

### LINKED ISSUE #2141: Private fields test coverage: Heritage does not have inner private environment
See the test case in https://bugs.chromium.org/p/v8/issues/detail?id=9177 . Would anyone be interested in making test262 tests for this case? cc @gsathya @spectranaut @caiolima

caiolima:
It closes #2141.

This case is relevant because implementors can misunderstand how to properly set `PrivateEnvironment` when evaluating classes, since it differs from `LexicalEnvironment`.

### DIFF
--- a/src/class-elements/grammar-private-environment-on-class-heritage-chained-usage.case
+++ b/src/class-elements/grammar-private-environment-on-class-heritage-chained-usage.case
@@ -27,7 +27,7 @@ features: [class-fields-private, class-fields-public]
 
 //- heritage
 
-extends class extends class extends class { x = this.#foo; } { #foo; x = this.#ba;r } { #bar; x = this.#fuz; }
+extends class extends class extends class { x = this.#foo; } { #foo; x = this.#bar; } { #bar; x = this.#fuz; }
 
 //- elements
 
--- a/src/class-elements/grammar-private-environment-on-class-heritage-chained-usage.case
+++ b/src/class-elements/grammar-private-environment-on-class-heritage-chained-usage.case
@@ -1,0 +1,34 @@
+// Copyright (C) 2019 Caio Lima (Igalia S.L). All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: It's a SyntaxError if a class expression evaluated on ClassHeritage uses an private name declared on subclass.
+info: |
+    Runtime Semantics: ClassDefinitionEvaluation
+
+    ClassTail : ClassHeritage { ClassBody }
+        ...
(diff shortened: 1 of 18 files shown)


leobalter:
cc @gsathya

src/class-elements/grammar-private-environment-on-class-heritage.case
```diff
@@ -0,0 +1,34 @@
...
+            a. Set the running execution context's LexicalEnvironment to classScope.
+            b. NOTE: The running execution context's PrivateEnvironment is outerPrivateEnvironment when evaluating ClassHeritage.
+        ...
+template: syntax/invalid
+features: [class-fields-private, class-fields-public]
+---*/
+
+//- heritage
+
+extends class { x = this.#foo; }
```
gsathya: Needs more test cases. ClassHeritage can be other productions like FunctionLiteral

leobalter: lol, this is why I love Test262... now you get me thinking about further chained heritage and there we go to more and more cases...

`class A extends class B extends class C extends class {} {} {} {}`

caiolima: Well...actually we would need more cases, since:

```
ClassHeritage[Yield, Await]:
    extendsLeftHandSideExpression[?Yield, ?Await]
```

It is gonna be fun. I'll work on that tomorrow morning.

src/class-elements/grammar-private-environment-on-class-heritage-chained-usage.case
```diff
@@ -0,0 +1,34 @@
...
+            a. Set the running execution context's LexicalEnvironment to classScope.
+            b. NOTE: The running execution context's PrivateEnvironment is outerPrivateEnvironment when evaluating ClassHeritage.
+        ...
+template: syntax/invalid
+features: [class-fields-private, class-fields-public]
+---*/
+
+//- heritage
+
+extends class extends class extends class { x = this.#foo; } { #foo; x = this.#ba;r } { #bar; x = this.#fuz; }
```
leobalter: `this.#ba;r` not wrong (r is a field) but I don't think this is intended here.

caiolima: It isn't. Sorry for this typo.

leobalter: don't be sorry.

Thanks for fixing it.

PR #2146, opened 2019-05-02
Use verifyProperty and remove dup check

FrankYFTang:


FrankYFTang:
@littledan @leobalter @Ms2ger

leobalter:
Thanks!

PR #2154, opened 2019-05-08
Tests to cover private methods comparison

### LINKED ISSUE #2153: Add tests to validate private methods comparison
According to spec, it is possible to get a private method function object. We need to have tests to verify such behavior.

caiolima:
It closes #2153.

### DIFF
--- a/src/class-elements/private-method-comparison.case
+++ b/src/class-elements/private-method-comparison.case
@@ -1,0 +1,43 @@
+// Copyright (C) 2019 Caio Lima (Igalia SL). All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: PrivateFieldGet of a private method returns the same function object to every instance of the same class
+esid: sec-privatefieldget
+info: |
+  PrivateFieldGet (P, O)
+    1. Assert: P is a Private Name.
+    2. If O is not an object, throw a TypeError exception.
+    3. If P.[[Kind]] is "field",
+      a. Let entry be PrivateFieldFind(P, O).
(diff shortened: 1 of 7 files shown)


src/class-elements/private-method-get-and-call.case
```diff
@@ -0,0 +1,45 @@
...
+    6. Else,
+      a. Assert: P.[[Kind]] is "accessor".
+      b. If P does not have a [[Get]] field, throw a TypeError exception.
+      c. Let getter be P.[[Get]].
+      d. Return ? Call(getter, O).
+
+  PrivateBrandCheck(O, P)
+    1. If O.[[PrivateBrands]] does not contain an entry e such that SameValue(e, P.[[Brand]]) is true,
+      a. Throw a TypeError exception.
+
```
leobalter: because of YAML compatibilities, please remove the trailing empty line.

src/class-elements/private-method-comparison.case
```diff
@@ -0,0 +1,43 @@
...
+    6. Else,
+      a. Assert: P.[[Kind]] is "accessor".
+      b. If P does not have a [[Get]] field, throw a TypeError exception.
+      c. Let getter be P.[[Get]].
+      d. Return ? Call(getter, O).
+
+  PrivateBrandCheck(O, P)
+    1. If O.[[PrivateBrands]] does not contain an entry e such that SameValue(e, P.[[Brand]]) is true,
+      a. Throw a TypeError exception.
+
```
leobalter: here as well

leobalter:
I'm applying these fixes here before landing.

PR #2155, opened 2019-05-08
Improve GetFunctionRealm coverage

shvaikalesh:


test/built-ins/Function/prototype/bind/get-fn-realm.js
```diff
@@ -7,17 +7,20 @@ description: >
...
 features: [cross-realm]
 ---*/
 
 var other = $262.createRealm().global;
 var C = new other.Function();
-var B = Function.prototype.bind.call(C);
+C.prototype = null;
+var B = C.bind();
 
-assert.sameValue(Object.getPrototypeOf(new B()), C.prototype);
```
shvaikalesh: Please note that this test does not invoke [`GetFunctionRealm`](https://tc39.github.io/ecma262/#sec-getfunctionrealm):

[`new`](https://tc39.github.io/ecma262/#sec-new-operator) calls bound function's [`[[Construct]]`](https://tc39.github.io/ecma262/#sec-bound-function-exotic-objects-construct-argumentslist-newtarget), which invokes regular [`[[Construct]]`](https://tc39.github.io/ecma262/#sec-ecmascript-function-objects-construct-argumentslist-newtarget) on target, which invokes [`GetPrototypeFromConstructor`](https://tc39.github.io/ecma262/#sec-getprototypefromconstructor) also on target.

The target, unlike the bound function, does have an object with non-enumerable `constructor` as its `prototype` property, so step 4 of [`GetPrototypeFromConstructor`](https://tc39.github.io/ecma262/#sec-getprototypefromconstructor)  won't be `true`.

test/built-ins/Proxy/get-fn-realm.js
```diff
@@ -6,24 +6,22 @@ es6id: 7.3.22
...
-       b. Let proxyTarget be the value of obj's [[ProxyTarget]] internal slot.
+       a. If obj.[[ProxyHandler]] is null, throw a TypeError exception.
+       b. Let proxyTarget be obj.[[ProxyTarget]].
        c. Return ? GetFunctionRealm(proxyTarget).
 features: [cross-realm, Proxy]
 ---*/
 
 var other = $262.createRealm().global;
 var C = new other.Function();
-// Ensure that the proxy does not report a `prototype` property
```
shvaikalesh: This test is correct, but it requires Proxy `get` trap to work, as well as [`[[Construct]]`](https://tc39.github.io/ecma262/#sec-proxy-object-internal-methods-and-internal-slots-construct-argumentslist-newtarget) method of Proxy to pass correct `NewTarget` to ordinary [`[[Construct]]`](https://tc39.github.io/ecma262/#sec-ecmascript-function-objects-construct-argumentslist-newtarget).

PR #2157, opened 2019-05-10
Fix invalid YAML in formatRange* tests.

rkirsling:
Correction for tests introduced in tc39#2134.

### DIFF
--- a/test/intl402/DateTimeFormat/prototype/formatRange/date-x-greater-than-y-throws.js
+++ b/test/intl402/DateTimeFormat/prototype/formatRange/date-x-greater-than-y-throws.js
@@ -2,9 +2,9 @@
 // This code is governed by the BSD license found in the LICENSE file.
 
 /*---
-description: |
+description: >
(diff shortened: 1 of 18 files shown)


test/intl402/DateTimeFormat/prototype/formatRange/date-x-greater-than-y-throws.js
```diff
@@ -2,9 +2,9 @@
 // This code is governed by the BSD license found in the LICENSE file.
 
 /*---
-description:
+description: |
   Throws a RangeError if date x is greater than y.
-info:
+info: >
```
leobalter: Thanks for the quick fix.

Just a nit: use > for description and | for info. The former will ignore line breaker while the latter will preserve them in the parsed values.

rkirsling: Whoops, can do! I actually just mimicked the the other files in these two directories (from #2139). I can update all of them accordingly though.

PR #2159, opened 2019-05-12
Improve Proxy/construct coverage

shvaikalesh:
Follow-up of #2156.


test/built-ins/Proxy/construct/trap-is-undefined-proto-from-ctor-realm.js
```diff
@@ -25,9 +25,10 @@ features: [cross-realm, Proxy, Reflect, Reflect.construct]
...
 var other = $262.createRealm().global;
 var C = new other.Function();
+C.prototype = null;
 
 var P = new Proxy(function() {}, {});
 
 var p = Reflect.construct(P, [], C);
 
-assert.sameValue(Object.getPrototypeOf(Object.getPrototypeOf(p)), other.Object.prototype);
+assert.sameValue(Object.getPrototypeOf(p), other.Object.prototype);
```
leobalter: the former formatting were passing on SpiderMonkey and the new version is not.

I'd suggest creating a new bug for FF to report it.

Meanwhile, can we have both versions in 2 separate files? I'd like to preserve the old test as well.

shvaikalesh: 021019b

test/built-ins/Proxy/construct/trap-is-undefined-proto-from-cross-realm-newtarget.js
```diff
@@ -0,0 +1,47 @@
...
+features: [cross-realm, Proxy, Reflect, Reflect.construct]
+---*/
+
+var other = $262.createRealm().global;
+var C = new other.Function();
+
+var P = new Proxy(function() {}, {});
+var p = Reflect.construct(P, [], C);
+
+assert.sameValue(Object.getPrototypeOf(p), C.prototype);
```
shvaikalesh: I've made the assertion more direct. Previously, it was:

```js
assert.sameValue(Object.getPrototypeOf(Object.getPrototypeOf(p)), other.Object.prototype);
```

We can be sure here that `Object.getPrototypeOf(C.prototype)` is indeed `other.Object.prototype`.

test/built-ins/Proxy/construct/return-not-object-throws-boolean-realm.js
```diff
@@ -0,0 +1,27 @@
+// Copyright (C) 2019 Aleksey Shvayka. All rights reserved.
```
shvaikalesh: We have 12 test files like this one: 6 for each primitive type + 6 cross-realm ones.
Do you mind if I make a follow up PR to merge them into 2 files (without loops)?
It feels like they take too much visual space in file system and test run logs.

test/built-ins/Proxy/construct/trap-is-undefined-proto-from-ctor-realm.js
shvaikalesh: Test results recap (this file, new version):

[SpiderMonkey v67.0b19](https://circleci.com/gh/tc39/test262/1260): FAIL
[SpiderMonkey v68.0b3](https://circleci.com/gh/tc39/test262/1705): PASS

[ChakraCore v1.11.8](https://circleci.com/gh/tc39/test262/1261): FAIL
[ChakraCore v1.11.9](https://circleci.com/gh/tc39/test262/1706): FAIL

PR #2161, opened 2019-05-13
Update RegExp property escape tests per Unicode v12.1.0

mathiasbynens:
Ref. https://github.com/tc39/ecma262/issues/1530.


mathiasbynens:
We have to skip the affected tests in the V8 test runner until this is merged: https://bugs.chromium.org/p/v8/issues/detail?id=9131#c3

leobalter:
Sorry it took so long!

> We have to skip the affected tests in the V8 test runner until this is merged

not anymore :)

PR #2162, opened 2019-05-14
[WIP] Default parameter value for catch binding

chicoxyzzy:
Tests for tc39/ecma262#1126

Based on @leobalter's commits from https://github.com/tc39/test262/pull/1483#issuecomment-372428065

### DIFF
--- a/src/dstr-binding/error/try-catch-init.template
+++ b/src/dstr-binding/error/try-catch-init.template
@@ -1,0 +1,22 @@
+// Copyright (C) 2018 Leo Balter. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+path: language/statements/try/dstr-catch-init-
+name: try catch with initializer statement
+esid: sec-runtime-semantics-catchclauseevaluation
+features: [destructuring-binding, catch-initializer]
+info: |
+  https://github.com/tc39/ecma262/pull/1126
+
+  Catch : catch ( CatchParameter ) Block
+
+  CatchParameter[Yield, Await] :
+    FormalParameter[?Yield, ?Await]
+---*/
+
+assert.throws(/*{ error }*/, function() {
+    try {
+      throw undefined;
+    } catch (/*{ elems }*/ = /*{ vals }*/) {}
+  });
+  
--- a/src/dstr-binding/error/try-catch-init.template
+++ b/src/dstr-binding/error/try-catch-init.template
@@ -1,5 +1,6 @@
 // Copyright (C) 2018 Leo Balter. All rights reserved.
 // This code is governed by the BSD license found in the LICENSE file.
+
 /*---
 path: language/statements/try/dstr-catch-init-
 name: try catch with initializer statement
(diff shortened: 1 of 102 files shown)


leobalter:
the current set is fine, this is still missing cases with a SingleNameBinding and a present Initializer. (like I mentioned before, `try{ throw undefined; } catch(e = 'oi!') { console.log(e == 'oi!'); }`

leobalter:
The files in `src/function-forms/*.case` might give direction for the missing cases here, but they are not directly applied.

The major difference there is that it aims for FormalParameters and not FormalParameter, so you can't rely on lists.

This way could just skip the test generation and write static test files.

Some of the cases interesting for coverage:

- _Abrupt completion returned by evaluation of initializer_:

```js
try {
 throw {}; // won't trigger the error
} catch (_ = (function() { throw new Test262Error(); }())) {
 // assert this block will be evaluated
}

// vs

try {
 throw undefined; // will trigger the error
} catch (_ = (function() { throw new Test262Error(); }())) {
 // assert this block will not be evaluated
} // also, split this test for cases with and without a finally block and verify its execution.
```

- Assert initializer won't be triggered if binding value is falsy but not undefined. (`false`, `''`, `0`, `null`, etc):

```js
try {
  throw undefined;
} catch (e = 42) {
  // assert e === 42
}

// and 

try {
  throw null; // and other cases where the value is 0, false, '', etc.
} catch (e = 42) {
  // assert the value is the thrown value, and not 42.
}
```

- The initializer is the binding name:

```
var e = 42;

try {
} catch(e = e) {} // no error, initializer is not used

try {
  throw 'foo';
} catch(e = e) {} // no error, initializer is not used

try {
  throw undefined;
} catch(e = e) {} // ReferenceError
// from specs, all the way to: Return ? base.GetBindingValue(GetReferencedName(V), IsStrictReference(V))
```

src/dstr-binding/error/try-catch-init.template
```diff
@@ -0,0 +1,22 @@
...
+  CatchParameter[Yield, Await] :
+    FormalParameter[?Yield, ?Await]
+---*/
+
+assert.throws(/*{ error }*/, function() {
+    try {
+      throw undefined;
+    } catch (/*{ elems }*/ = /*{ vals }*/) {}
+  });
+
```
leobalter: in the meanwhile, it would be nice if you could fix this EOF for consistency, please.

chicoxyzzy:
I think that I should use `feature` field somehow, but I don't understand how exactly. Should I just use `catch-initializer` in both templates and cases?

chicoxyzzy:
Or should I create a separate directory for my test templates and cases?

chicoxyzzy:
@leobalter I need your help :) Should I create new directory with my new templates and cases? Or should I rewrite my templates somehow?

leobalter:
> I need your help :) Should I create new directory with my new templates and cases? Or should I rewrite my templates somehow?

yes, please. If it ends up just like another folder, you can reuse it then. A new folder you give you some freedom to author tests and you can just copy other existing templates.

leobalter:
In a second thought, let me finish looking at the current tests.

leobalter:
Closing this PR as the ECMA-262 change has found no consensus at the TC39 meeting in 2019-10-01.

Thanks for the hard work, @chicoxyzzy!

PR #2163, opened 2019-05-15
Improve Function#apply coverage

### LINKED ISSUE #1727: Function#apply: second argument is not an array-like value.
This is feedback after reviewing this [contributed test](https://github.com/tc39/test262/blob/master/implementation-contributed/javascriptcore/stress/apply-second-argument-must-be-array-like.js)

- [ ] Cover cases where the second argument is not an array like value.

shvaikalesh:
Closes #1727.


test/built-ins/Function/prototype/apply/this-not-callable-realm.js
```diff
@@ -0,0 +1,32 @@
...
+
+  1. Let func be the this value.
+  2. If IsCallable(func) is false, throw a TypeError exception.
+features: [cross-realm]
+---*/
+
+var OFunction = $262.createRealm().global.Function;
+
+assert.throws(TypeError, function() {
+  OFunction.prototype.apply.call(undefined, {}, []);
```
leobalter: so here the error is from the internal steps after calling `OFunction#apply`, so I assume the TypeError should follow the realm.

See something analog to the opposite effect [https://github.com/tc39/test262/pull/688#issuecomment-229995414](https://github.com/tc39/test262/pull/688#issuecomment-229995414)

The active function here is apply and it should probably capture the TypeError from the same realm.

All engines are actually doing this already, and they are correct.

leobalter: Let me go a bit deep on the specs:

Runtime Semantics: EvaluateCall

...
8. Let result be Call(**func**, thisValue, argList).

---

Call ( **F**, V [ , argumentsList ] )

...
3. Return ? **F**.[[Call]](V, argumentsList).

---

[[[Call]]](https://tc39.github.io/ecma262/#sec-ecmascript-function-objects-call-thisargument-argumentslist)  ( thisArgument, argumentsList )

The [[Call]] internal method for an ECMAScript function **object F** is called with parameters thisArgument and argumentsList, a List of ECMAScript language values. The following steps are taken:

...
3. Let callerContext be the running execution context.
4. Let calleeContext be [PrepareForOrdinaryCall](https://tc39.github.io/ecma262/#sec-prepareforordinarycall)(**F**, undefined).

---

PrepareForOrdinaryCall ( **F**, newTarget )

...
2. Let callerContext be the running execution context.
3. Let calleeContext be a new ECMAScript code execution context.
4. Set the Function of calleeContext to F.
5. Let calleeRealm be F.[[Realm]].
6. Set the Realm of calleeContext to calleeRealm.
...
12. Push calleeContext onto the execution context stack; calleeContext is now the running execution context.
13. NOTE: Any exception objects produced after this point are associated with calleeRealm.
14. Return calleeContext.

---

At this point it's a bit clearer that the TypeError - the exception produced from inside .apply is associated with the calleeRealm.

The next steps from `[[Call]]` will make sure the correct Realm is set:

6. Perform OrdinaryCallBindThis(F, calleeContext, thisArgument).
7. Let result be OrdinaryCallEvaluateBody(F, argumentsList).

leobalter: This test file and `test/built-ins/Function/prototype/apply/argarray-not-object-realm.js` can be easily fixed by changing the references to TypeError in `assert.throws(TypeError, ...` to a TypeError from the given realm, e.g.:

```js
 // Following consistency with OFunction: (please don't use this comment line)
var OTypeError = $262.createRealm().global.TypeError;

assert.throws(OTypeError, ...
```

shvaikalesh: I deeply appreciate detailed write-up. Thank you, @leobalter.

5a725fe

test/built-ins/Function/prototype/apply/argarray-not-object-realm.js
```diff
@@ -0,0 +1,38 @@
...
+
+  [...]
+  2. If Type(obj) is not Object, throw a TypeError exception.
+features: [cross-realm]
+---*/
+
+var other = $262.createRealm().global;
+var fn = new other.Function();
+
+assert.throws(other.TypeError, function() {
```
leobalter: nice!

PR #2166, opened 2019-05-15
Add tests for [[Set]] performing proto lookup onto empty typedarrays

leobalter:
Ref https://twitter.com/bmeurer/status/1128674484217679873

### DIFF
--- a/test/built-ins/Reflect/set/set-to-array-with-empty-typedarray-proto.js
+++ b/test/built-ins/Reflect/set/set-to-array-with-empty-typedarray-proto.js
@@ -1,0 +1,85 @@
+// Copyright (C) 2019 Leo Balter. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-reflect.set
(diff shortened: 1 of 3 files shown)


ljharb:
Looks like this is also https://github.com/tc39/ecma262/issues/1541

leobalter:
It is directly related as you might also see the reference.

I plan to write tests for the proposed changes as well, I don't expect this one will really see the light of the day in Test262 but it may help to illustrate the problem and what's affected.

ljharb:
Given that eshost seems to report that all engines are already implementing the proposed bug fix (https://github.com/tc39/ecma262/issues/1541#issuecomment-492793709), why would you not expect it to land?

leobalter:
> why would you not expect it to land?

I don't expect to land the current tests as they are in this PR. They are following the current specs, without the proposed spec fix. They serve as an illustration.

As I mentioned, I'll have another PR with tests for the proposed changes.

We can discard this PR when TC39 achieve consensus in the next meeting.

PR #2167, opened 2019-05-16
change the removed unit 'generic' to 'gigabit'

FrankYFTang:
https://github.com/tc39/proposal-unified-intl-numberformat/pull/42 removed 'generic'. So we need to change the value to a still valid unit.
@sffc @littledan @leobalter


sffc:
LGTM

PR #2175, opened 2019-05-22
 Update tests description for super.x in eval

### LINKED ISSUE #2127: language/expressions/class/elements/derived-cls-direct-eval-err-contains-superproperty-1.js looks incorrect
https://github.com/tc39/test262/blob/master/test/language/expressions/class/elements/derived-cls-direct-eval-err-contains-superproperty-1.js

The comment correctly says the direct `eval` should hit an Early Error.

But the test code expects the eval to succeed.

leobalter:
Closes #2127 

cc @jorendorff

### DIFF
--- a/src/class-elements/eval-contains-superproperty-1.case
+++ b/src/class-elements/eval-contains-superproperty-1.case
@@ -2,7 +2,7 @@
 // This code is governed by the BSD license found in the LICENSE file.
 
 /*---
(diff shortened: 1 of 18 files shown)


jorendorff:
Oh, I see, this is supposed to work. Well, maybe the tests should be renamed—they still have `-err-` in the name.

PR #2176, opened 2019-05-23
Add test for multiple evaluations of classes with private

jbhoosreddy:
Related to: https://github.com/tc39/test262/issues/1343#issuecomment-446225159
### Summary of changes
1. This PR attempts to add case for multiple class evaluation in different realms would fail a private brand check.

> Class factories -- If you have a function which returns a class expression, then each time this function is run, it will create different internal private names, and the private methods from instances of one will not exist on the other
 _**Ditto, if the same script is evaluated in two different realms, the clasess can't access each other's private methods**_


jbhoosreddy:
@caiolima @mkubilayk @littledan @leobalter

test/language/statements/class/elements/private-getter-brand-check-multiple-evaluations-of-class-realm.js
```diff
@@ -0,0 +1,54 @@
...
+esid: sec-privatefieldget
+info: |
+  ClassTail : ClassHeritage { ClassBody }
+    ...
+    11. Let proto be ObjectCreate(protoParent).
+    ...
+    31. If PrivateBoundIdentifiers of ClassBody contains a Private Name P such that the P's [[Kind]] field is either "method" or "accessor",
+      a. Set F.[[PrivateBrand]] to proto.
+    ...
+
```
caiolima: I think we need to remove this blank line to avoid issues with YAML parser.

test/language/statements/class/elements/private-getter-brand-check-multiple-evaluations-of-class-realm.js
```diff
@@ -0,0 +1,54 @@
...
+    access(o) {
+      return o.#m;
+    }
+  }
+)`;
+
+let createAndInstantiateClass = function (eval) {
+  let C = eval(classStringExpression);
+  let c = new C();
+  return c;
```
caiolima: I think we could `return new C();` here.

jbhoosreddy:
thanks for pointing that out!

test/language/statements/class/elements/private-getter-brand-check-multiple-evaluations-of-class-realm.js
leobalter: Thanks for the reminder, but empty lines in between the info block is fine, just avoid leading or trailing lines when possible.

test/language/statements/class/elements/private-getter-brand-check-multiple-evaluations-of-class-realm.js
leobalter: what happens here if:

1. c1 and c2 are done both with `eval1`? (or just the same reference to eval)
2. you use a direct eval:  `let C = $262.createRealm().global.eval(classStringExpression);`?
3. use a local eval, direct and indirect?

Those might be different purposes but for n. 1, I'm not sure it will differentiate from the current test status. So you might only need one reference to eval instead of `eval1` and `eval2`. WDYT?

leobalter: These questions extends to the part you verify the access method as well.

test/language/statements/class/elements/private-getter-brand-check-multiple-evaluations-of-class-realm.js
```diff
@@ -0,0 +1,54 @@
...
+class {
+    get #m() { return 'test262'; }
+
+    access(o) {
+      return o.#m;
+    }
+  }
+)`;
+
+let createAndInstantiateClass = function (eval) {
```
leobalter: nit pick: to clarify it's a custom indirect eval, you might rename this to something else, e.g. `_eval`.

test/language/statements/class/elements/private-getter-brand-check-multiple-evaluations-of-class-realm.js
leobalter: for the code golf, you might also just `return new eval(classStringExpression);`

test/language/statements/class/elements/private-getter-brand-check-multiple-evaluations-of-class-realm.js
```diff
@@ -0,0 +1,54 @@
...
+      a. Throw a TypeError exception.
+features: [class, class-methods-private]
+flags: [noStrict]
+---*/
+
+let eval1 = $262.createRealm().global.eval;
+let eval2 = $262.createRealm().global.eval;
+
+let classStringExpression = `(
+class {
```
leobalter: also another nit pick: this is a class expression in the class statements folder.

test/language/statements/class/elements/private-getter-brand-check-multiple-evaluations-of-class-realm.js
jbhoosreddy: This is interesting, if we forsake the different realms aspect of this test, and use only one eval, it should still work. it would be a different form of the class factory stuff @caiolima wrote.

If I do that, I'll probably use eval from the same context instead of another realm.. thoughts @leobalter ?

jbhoosreddy: Based on your comments I can leave the realm stuff and address the other issues

jbhoosreddy:
@caiolima @leobalter ready for review again

test/language/expressions/class/private-getter-brand-check-multiple-evaluations-of-class-function-ctor.js
```diff
@@ -0,0 +1,48 @@
...
+  get m() { return 'test262'; }
+
+  access(o) {
+    return o.m;
+  }
+}
+`;
+
+let createAndInstantiateClass = function () {
+  return new (new (Function(classStringExpression)));
```
caiolima: This test is confusing me a little bit. We want to do this test using `Function` constructor, right? IIUC, `createAndInstantiateClass` will evaluate a new `ClassExpression` and then return an instantiation of this class. Is there any reason why you're using `new` to call `Function(classStringExpression)`? Also, I think it would be good if we construct the new function using `new operator`. What I have in mind is something like this:

```
let createAndInstantiateClass = function () {
    return new ((new Function(classStringExpression))());
}
```
or being more verbose, but making the code easier to follow, IMHO:
```
let createAndInstantiateClass = function () {
  let classFactoryFunction = new Function(classStringExpression);
  let unamedClass = classFactoryFunction();
	return new unamedClass();
}
```

test/language/expressions/class/private-getter-brand-check-multiple-evaluations-of-class-realm-realm-function-ctor.js
```diff
@@ -0,0 +1,48 @@
...
+  get #m() { return 'test262'; }
+
+  access(o) {
+    return o.#m;
+  }
+}
+`;
+
+let createAndInstantiateClass = function () {
+  return new (new ($262.createRealm().global.Function)(classStringExpression));
```
caiolima: I think this is going to return a class expression, instead of an instance of evaluated class. In this case, `c1.access(c1)` will fail to execute. Ditto for making it more verbose. IMHO, in the case of this test fails in an VM, less mental effort to implementors understand tests helps a lot.

test/language/expressions/class/private-getter-brand-check-multiple-evaluations-of-class-realm.js
```diff
@@ -0,0 +1,52 @@
...
+class {
+  get #m() { return 'test262'; }
+
+  access(o) {
+    return o.#m;
+  }
+}
+)`;
+
+let createAndInstantiateClass = function (eval) {
```
caiolima: nit: I think we could use `_eval` here too.

test/language/expressions/class/private-method-brand-check-multiple-evaluations-of-class-eval-indirect.js
```diff
@@ -0,0 +1,49 @@
...
+  }
+}
+)`;
+
+let createAndInstantiateClass = function (_eval) {
+  return new (_eval(classStringExpression));
+};
+
+let c1 = createAndInstantiateClass();
+let c2 = createAndInstantiateClass();
```
caiolima: `eval` arguments are missing here.

test/language/expressions/class/private-method-brand-check-multiple-evaluations-of-class-factory.js
```diff
@@ -0,0 +1,47 @@
...
+  const C = class {
+    #m() { return 'test262'; }
+
+    access(o) {
+      return o.#m();
+    }
+  }
+
+  let c = new C();
+  return c;
```
caiolima: nit: `return new C()`.

test/language/expressions/class/private-method-brand-check-multiple-evaluations-of-class-function-ctor.js
```diff
@@ -0,0 +1,48 @@
...
+  #m() { return 'test262'; }
+
+  access(o) {
+    return o.#m();
+  }
+}
+`;
+
+let createAndInstantiateClass = function () {
+  return new (new (Function(classStringExpression)));
```
caiolima: Ditto for the comments I've made into getter test.

test/language/expressions/class/private-method-brand-check-multiple-evaluations-of-class-realm-function-ctor.js
```diff
@@ -0,0 +1,48 @@
...
+  #m() { return 'test262'; }
+
+  access(o) {
+    return o.#m();
+  }
+}
+`;
+
+let createAndInstantiateClass = function () {
+  return new (new ($262.createRealm().global.Function)(classStringExpression));
```
caiolima: Ditto for this returning a class expression instead of an instance of the evaluated class.

test/language/expressions/class/private-method-brand-check-multiple-evaluations-of-class-realm.js
```diff
@@ -0,0 +1,52 @@
...
+class {
+  get #m() { return 'test262'; }
+
+  access(o) {
+    return o.#m;
+  }
+}
+)`;
+
+let createAndInstantiateClass = function (eval) {
```
caiolima: Ditto for `_eval`.

test/language/expressions/class/private-method-brand-check-multiple-evaluations-of-class-realm.js
```diff
@@ -0,0 +1,52 @@
...
+features: [class, class-methods-private]
+flags: [noStrict]
+---*/
+
+let eval1 = $262.createRealm().global.eval;
+let eval2 = $262.createRealm().global.eval;
+
+let classStringExpression = `(
+class {
+  get #m() { return 'test262'; }
```
caiolima: I think you want to test a method here.

test/language/expressions/class/private-setter-brand-check-multiple-evaluations-of-class-factory.js
```diff
@@ -0,0 +1,49 @@
...
+  const C = class {
+    set #m(v) { this._v = v; }
+
+    access(o, v) {
+      o.#m = v;
+    }
+  }
+
+  let c = new C();
+  return c;
```
caiolima: nit: `return new C()`.

test/language/expressions/class/private-setter-brand-check-multiple-evaluations-of-class-realm-function-ctor.js
```diff
@@ -0,0 +1,50 @@
...
+  set #m(v) { this._v = v; }
+
+  access(o, v) {
+    o.#m = v;
+  }
+}
+`;
+
+let createAndInstantiateClass = function () {
+  return new (new ($262.createRealm().global.Function)(classStringExpression));
```
caiolima: Ditto for this return a class expression.

test/language/expressions/class/private-setter-brand-check-multiple-evaluations-of-class-realm.js
```diff
@@ -0,0 +1,54 @@
...
+class {
+  set #m(v) { this._v = v; }
+
+  access(o, v) {
+    o.#m = v;
+  }
+}
+)`;
+
+let createAndInstantiateClass = function (eval) {
```
caiolima: Ditto for `_eval`

test/language/expressions/class/private-setter-brand-check-multiple-evaluations-of-class-function-ctor.js
```diff
@@ -0,0 +1,50 @@
...
+  set #m(v) { this._v = v; }
+
+  access(o, v) {
+    o.#m = v;
+  }
+}
+`;
+
+let createAndInstantiateClass = function () {
+  return new (new (Function(classStringExpression)));
```
caiolima: Ditto for making it verbose.

jbhoosreddy:
Thanks for the review @caiolima. Ready for your review again.

rwaldron:
@leobalter ping.

jbhoosreddy:
Thank you @leobalter !

PR #2178, opened 2019-05-24
Check private methods are not installed before super returns

mkubilayk:
Adds tests calling private methods before super is constructed - according to the test plan posted in #1343.

>  Private methods are installed "when super returns" and no earlier. A similar test is possible with field initializers instead of code called in the constructor.

I will commit the generated tests once the PR is approved.

cc: @littledan @jbhoosreddy @caiolima

### DIFF
--- a/src/class-elements/prod-private-getter-before-super-return-in-constructor.case
+++ b/src/class-elements/prod-private-getter-before-super-return-in-constructor.case
@@ -1,0 +1,42 @@
+// Copyright (C) 2019 Kubilay Kahveci (Bloomberg LP). All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: Private getters are installed "when super returns" and no earlier (call in constructor)
+info: |
+  SuperCall: super Arguments
+    1. Let newTarget be GetNewTarget().
+    2. If newTarget is undefined, throw a ReferenceError exception.
+    3. Let func be ? GetSuperConstructor().
+    4. Let argList be ArgumentListEvaluation of Arguments.
+    5. ReturnIfAbrupt(argList).
+    6. Let result be ? Construct(func, argList, newTarget).
+    7. Let thisER be GetThisEnvironment( ).
+    8. Let F be thisER.[[FunctionObject]].
+    9. Assert: F is an ECMAScript function object.
+    10. Perform ? InitializeInstanceElements(result, F).
+
+  EDITOR'S NOTE:
+    Private fields are added to the object one by one, interspersed with
+    evaluation of the initializers, following the construction of the
+    receiver. These semantics allow for a later initializer to refer to
+    a previous private field.
+template: default
+features: [class-methods-private]
+---*/
+
+//- elements
+constructor() {
+    this.f();
+}
+
+//- assertions
+class D extends C {
+    f() { this.#m; }
+    get #m() { return 42; }
+}
+
+assert(D.prototype.hasOwnProperty('f'));
+assert.throws(TypeError, function() {
+    var d = new D();
+}, 'private getters are not installed before super returns');
--- a/src/class-elements/prod-private-getter-before-super-return-in-field-initializer.case
+++ b/src/class-elements/prod-private-getter-before-super-return-in-field-initializer.case
@@ -1,0 +1,40 @@
+// Copyright (C) 2019 Kubilay Kahveci (Bloomberg LP). All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: Private getters are installed "when super returns" and no earlier (call in field initializer)
+info: |
+  SuperCall: super Arguments
+    1. Let newTarget be GetNewTarget().
+    2. If newTarget is undefined, throw a ReferenceError exception.
+    3. Let func be ? GetSuperConstructor().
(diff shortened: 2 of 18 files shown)


mkubilayk:
@caiolima updated with tests covering private accessors. I thought we generate tests once the PR is approved. I sent them as a separate commit just in case.

src/class-elements/prod-private-getter-before-super-return-in-constructor.case
```diff
@@ -0,0 +1,43 @@
...
+    8. Let F be thisER.[[FunctionObject]].
+    9. Assert: F is an ECMAScript function object.
+    10. Perform ? InitializeInstanceElements(result, F).
+
+  EDITOR'S NOTE:
+    Private fields are added to the object one by one, interspersed with
+    evaluation of the initializers, following the construction of the
+    receiver. These semantics allow for a later initializer to refer to
+    a previous private field.
+
```
caiolima: nit: This trailing line makes YAML break sometimes. Could you remove them?

src/class-elements/prod-private-getter-before-super-return-in-field-initializer.case
```diff
@@ -0,0 +1,41 @@
...
+    8. Let F be thisER.[[FunctionObject]].
+    9. Assert: F is an ECMAScript function object.
+    10. Perform ? InitializeInstanceElements(result, F).
+
+  EDITOR'S NOTE:
+    Private fields are added to the object one by one, interspersed with
+    evaluation of the initializers, following the construction of the
+    receiver. These semantics allow for a later initializer to refer to
+    a previous private field.
+
```
caiolima: ditto.

src/class-elements/prod-private-method-before-super-return-in-field-initializer.case
```diff
@@ -0,0 +1,41 @@
...
+    8. Let F be thisER.[[FunctionObject]].
+    9. Assert: F is an ECMAScript function object.
+    10. Perform ? InitializeInstanceElements(result, F).
+
+  EDITOR'S NOTE:
+    Private fields are added to the object one by one, interspersed with
+    evaluation of the initializers, following the construction of the
+    receiver. These semantics allow for a later initializer to refer to
+    a previous private field.
+
```
caiolima: Ditto.

src/class-elements/prod-private-setter-before-super-return-in-constructor.case
```diff
@@ -0,0 +1,43 @@
...
+    8. Let F be thisER.[[FunctionObject]].
+    9. Assert: F is an ECMAScript function object.
+    10. Perform ? InitializeInstanceElements(result, F).
+
+  EDITOR'S NOTE:
+    Private fields are added to the object one by one, interspersed with
+    evaluation of the initializers, following the construction of the
+    receiver. These semantics allow for a later initializer to refer to
+    a previous private field.
+
```
caiolima: ditto.

src/class-elements/prod-private-method-before-super-return-in-constructor.case
```diff
@@ -0,0 +1,43 @@
...
+    8. Let F be thisER.[[FunctionObject]].
+    9. Assert: F is an ECMAScript function object.
+    10. Perform ? InitializeInstanceElements(result, F).
+
+  EDITOR'S NOTE:
+    Private fields are added to the object one by one, interspersed with
+    evaluation of the initializers, following the construction of the
+    receiver. These semantics allow for a later initializer to refer to
+    a previous private field.
+
```
caiolima: ditto.

src/class-elements/prod-private-setter-before-super-return-in-field-initializer.case
```diff
@@ -0,0 +1,41 @@
...
+    8. Let F be thisER.[[FunctionObject]].
+    9. Assert: F is an ECMAScript function object.
+    10. Perform ? InitializeInstanceElements(result, F).
+
+  EDITOR'S NOTE:
+    Private fields are added to the object one by one, interspersed with
+    evaluation of the initializers, following the construction of the
+    receiver. These semantics allow for a later initializer to refer to
+    a previous private field.
+
```
caiolima: ditto.

caiolima:
> I thought we generate tests once the PR is approved. I sent them as a separate commit just in case.

Oh right. I don't think there is a strict rule here, but IMHO it is good to have them generated on first PR, because some templates are hard to review without generated tests. Also, the CI checks if they are missing.

mkubilayk:
@caiolima thanks for the review. Removed the trailing blank lines.

src/class-elements/prod-private-getter-before-super-return-in-constructor.case
```diff
@@ -0,0 +1,42 @@
...
+---*/
+
+//- elements
+constructor() {
+    this.f();
+}
+
+//- assertions
+class D extends C {
+    f() { this.#m; }
```
caiolima: nit: Oops, I missed this in the first review, but I think the code style uses 2 space bars.

leobalter: well, this doesn't change the test and there is more than just this to get fixed. 

I'd say we defer style for a follow up or a general fix in the test files.

mkubilayk:
Thanks!

PR #2179, opened 2019-05-27
Improve "name" tests of Promise built-in functions

### LINKED ISSUE #1752: Review name property for the executor functions in Promise.resolve calls
This is feedback after reviewing this [contributed test](https://github.com/tc39/test262/blob/master/implementation-contributed/javascriptcore/stress/private-name-as-anonymous-builtin.js)

- [ ] Review name property for the executor functions in Promise.resolve calls

shvaikalesh:
| Engine              | `name`  | `Promise.allSettled` | CI |
| -------------- | --------- | -------------------- | -- |
| ChakraCore     | 👌          | ❌                               | ❌|
| JSC                  | 👌          | ❌                               | ❌|
| SpiderMonkey | ❌          | ❌                              | ❌|
| V8                    | ❌          | 👌                               | ❌|

Closes #1752.

### DIFF
--- a/test/built-ins/Promise/all/resolve-element-function-name.js
+++ b/test/built-ins/Promise/all/resolve-element-function-name.js
@@ -2,6 +2,7 @@
 // This code is governed by the BSD license found in the LICENSE file.
 
 /*---
+esid: sec-promise.all-resolve-element-functions
 es6id: 25.4.4.1.2
 description: The `name` property of Promise.all Resolve Element functions
 info: |
@@ -29,3 +30,6 @@ NotPromise.resolve = function(v) {
 Promise.all.call(NotPromise, [thenable]);
 
 assert.sameValue(Object.prototype.hasOwnProperty.call(resolveElementFunction, "name"), false);
+assert.sameValue(Object.getOwnPropertyNames(resolveElementFunction).join(), "length");
+assert.sameValue(delete resolveElementFunction.name, true);
+assert.sameValue(resolveElementFunction.name, "");
--- a/test/built-ins/Promise/all/resolve-element-function-name.js
+++ b/test/built-ins/Promise/all/resolve-element-function-name.js
@@ -30,6 +30,4 @@ NotPromise.resolve = function(v) {
 Promise.all.call(NotPromise, [thenable]);
 
 assert.sameValue(Object.prototype.hasOwnProperty.call(resolveElementFunction, "name"), false);
-assert.sameValue(Object.getOwnPropertyNames(resolveElementFunction).join(), "length");
-assert.sameValue(delete resolveElementFunction.name, true);
 assert.sameValue(resolveElementFunction.name, "");
--- a/test/built-ins/Promise/all/resolve-element-function-name.js
+++ b/test/built-ins/Promise/all/resolve-element-function-name.js
@@ -3,7 +3,6 @@
 
 /*---
 esid: sec-promise.all-resolve-element-functions
-es6id: 25.4.4.1.2
 description: The `name` property of Promise.all Resolve Element functions
 info: |
   A promise resolve function is an anonymous built-in function.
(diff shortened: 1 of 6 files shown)


test/built-ins/Promise/executor-function-name.js
```diff
@@ -22,3 +23,6 @@ function NotPromise(executor) {
 Promise.resolve.call(NotPromise);
 
 assert.sameValue(Object.prototype.hasOwnProperty.call(executorFunction, "name"), false);
+assert.sameValue(Object.getOwnPropertyNames(executorFunction).join(), "length");
+assert.sameValue(delete executorFunction.name, true);
```
leobalter: why the delete here after `assert.sameValue(Object.prototype.hasOwnProperty.call(executorFunction, "name"), false);`?

test/built-ins/Promise/executor-function-name.js
```diff
@@ -22,3 +23,6 @@ function NotPromise(executor) {
 Promise.resolve.call(NotPromise);
 
 assert.sameValue(Object.prototype.hasOwnProperty.call(executorFunction, "name"), false);
+assert.sameValue(Object.getOwnPropertyNames(executorFunction).join(), "length");
```
leobalter: this seems a bit out of scope for this test. This is asserting the only own string-based property names in executorFunction is `"length"` right after you already check that `name` is not an own property of `executorFunction`.

test/built-ins/Promise/executor-function-name.js
```diff
@@ -22,3 +23,6 @@ function NotPromise(executor) {
...
+assert.sameValue(delete executorFunction.name, true);
+assert.sameValue(executorFunction.name, "");
```
leobalter: this assertion is ok, but the others just seem unnecessary.

test/built-ins/Promise/executor-function-name.js
shvaikalesh: I bet the original test was written at the time when `Function.name` was non-standard and non-configurable.
Totally useless, sorry.

71e1291

test/built-ins/Promise/executor-function-name.js
shvaikalesh: My understanding here is that `Object.getOwnPropertyNames` is used in original test file because it is a stress test and `Object#hasOwnProperty` in hot loop may give incorrect results.

Stress tests are surely out of scope.

71e1291

test/built-ins/Promise/all/resolve-element-function-name.js
```diff
@@ -2,6 +2,7 @@
 // This code is governed by the BSD license found in the LICENSE file.
 
 /*---
+esid: sec-promise.all-resolve-element-functions
 es6id: 25.4.4.1.2
```
rwaldron: When you add an `esid`, you should remove the `es6id`

shvaikalesh: Will do, thank you.
65c230a

test/built-ins/Promise/allSettled/reject-element-function-name.js
```diff
@@ -30,3 +30,4 @@ NotPromise.resolve = function(v) {
 Promise.allSettled.call(NotPromise, [thenable]);
 
 assert.sameValue(Object.prototype.hasOwnProperty.call(rejectElementFunction, 'name'), false);
+assert.sameValue(rejectElementFunction.name, '');
```
leobalter: I'm caught from another topic, but if I remember correctly we've had some discussion on anonymous function names in the last TC39 meeting. 

cc @gsathya, mind to remember me if something has changed there?

gsathya: The recent change I presented was to make the `name` property an `own` named property on the anonymous function. These tests just seem to be testing that the value of the property is an empty string, not the actual position of the property.

LGTM

PR #2180, opened 2019-05-27
Verify that symbols are not callable

### LINKED ISSUE #1747: Verify correct errors calling Symbol values
This is feedback after reviewing this [contributed test](https://github.com/tc39/test262/blob/master/implementation-contributed/javascriptcore/stress/error-description-on-symbols-should-not-crash.js)

- [ ] Symbols are not callable... `Symbol(1)()`

shvaikalesh:
Closes #1747.

### DIFF
--- a/test/built-ins/Symbol/not-callable.js
+++ b/test/built-ins/Symbol/not-callable.js
@@ -1,0 +1,25 @@
+// Copyright (C) 2019 Aleksey Shvayka. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-properties-of-symbol-instances
+description: >
+  Symbol primitives and objects are not callable.
+info: |
+  Properties of Symbol Instances
+
+  Symbol instances are ordinary objects that inherit properties from the
+  Symbol prototype object. Symbol instances have a [[SymbolData]] internal slot.
+  The [[SymbolData]] internal slot is the Symbol value represented by this
+  Symbol object.
+features: [Symbol]
+---*/
+
+assert.throws(TypeError, function() {
+  Symbol('desc')();
+});
+
+assert.throws(TypeError, function() {
+  let sym = Symbol();
+  let obj = Object(sym);
+  obj();
+});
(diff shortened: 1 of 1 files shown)


test/built-ins/Symbol/not-callable.js
```diff
@@ -0,0 +1,25 @@
...
+
+  Symbol instances are ordinary objects that inherit properties from the
+  Symbol prototype object. Symbol instances have a [[SymbolData]] internal slot.
+  The [[SymbolData]] internal slot is the Symbol value represented by this
+  Symbol object.
+features: [Symbol]
+---*/
+
+assert.throws(TypeError, function() {
+  Symbol('desc')();
```
leobalter: if you store the symbol before this function, you avoid a false positive if the impl does not support symbols yet.

So:

```js
var s = Symbol();
  s();
});
```

test/built-ins/Symbol/not-callable.js
```diff
@@ -0,0 +1,25 @@
...
+});
+
+assert.throws(TypeError, function() {
+  let sym = Symbol();
+  let obj = Object(sym);
+  obj();
+});
```
leobalter: Same here as you still don't have a wrong test result if in the impl does not support `Object(sym)` yet. The idea is always using a minimal code piece inside the error catching function...

```js
var obj = Object(Symbol());
  obj();
});
```

test/built-ins/Symbol/not-callable.js
```diff
@@ -0,0 +1,25 @@
+// Copyright (C) 2019 Aleksey Shvayka. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-properties-of-symbol-instances
+description: >
+  Symbol primitives and objects are not callable.
```
leobalter: to make sure they are really not callable, it might be worth adding a check for `new (Symbol());` (not `new Symbol()`)

test/built-ins/Symbol/not-callable.js
shvaikalesh: 134e812

test/built-ins/Symbol/not-callable.js
shvaikalesh: 134e812

test/built-ins/Symbol/not-callable.js
shvaikalesh: da71fce

PR #2182, opened 2019-05-28
Add for-of head Initializer tests

### LINKED ISSUE #2113: Initializer in for-of statement heads
In acornjs/acorn#821 it was brought up that there is no test for `for (var x = 42 of list) process(x);` (it should fail).

shvaikalesh:
Closes #2113.

### DIFF
--- a/test/language/statements/for-await-of/async-gen-dstr-const-ary-ptrn-init-err.js
+++ b/test/language/statements/for-await-of/async-gen-dstr-const-ary-ptrn-init-err.js
@@ -1,0 +1,20 @@
+// Copyright (C) 2019 Aleksey Shvayka. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-iteration-statements
+description: >
+  Initializer is not allowed in head's ForDeclaration position.
+info: |
+  IterationStatement:
+    for await (ForDeclaration of AssignmentExpression) Statement
+negative:
+  phase: parse
+  type: SyntaxError
+features: [async-iteration]
+---*/
+
+$DONOTEVALUATE();
+
+async function fn() {
+  for await (const [x] = 1 of []) {}
+}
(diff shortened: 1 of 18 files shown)


test/language/statements/for-await-of/async-gen-dstr-const-ary-ptrn-init-err.js
```diff
@@ -0,0 +1,20 @@
...
+negative:
+  phase: parse
+  type: SyntaxError
+features: [async-iteration]
+---*/
+
+$DONOTEVALUATE();
+
+async function fn() {
+  for await (const [x] = 1 of []) {}
```
leobalter: here we go again with tests for syntax that doesn't exist in the specs.

The problem is is that IterationStatement won't have any production for the `for-await-of` with an Initializer after the ForBinding`

```
[+Await]for await ( var ForBinding[?Yield, ?Await] of AssignmentExpression[+In, ?Yield, ?Await] ) Statement[?Yield, ?Await, ?Return]
[+Await]for await ( ForDeclaration[?Yield, ?Await] of AssignmentExpression[+In, ?Yield, ?Await] ) Statement[?Yield, ?Await, ?Return]
```

It's also hard to assume how the given code would work and all the actual implications if it did exist.

this would also need to ignore cases matching yield and await, etc.

The description says the Initializer is not allowed, but it's actually not specified in ECMA-262 at all. 

The file name also assumes it's an initializer error `-init-err.js` but there isn't any error for the initializer, it can't be something that exist.

---

Is this a behavior observed from any implementation? Otherwise I'm not compelled to add more tests for "this is not specified code". Otherwise, everything not in the spec would become a valid test.

leobalter:
well... it's weird code but guess what? ChakraCore and XS are failing...

```
$ eshost -x "for (var x = 1 of [undefined, ,2, false]) { console.log(x)}"
#### d8
SyntaxError: for-of loop variable declaration may not have an initializer.

#### ch
undefined
undefined
2
false

#### jsc
SyntaxError: Cannot assign to the loop variable inside a for-of loop header.

#### jsshell
SyntaxError: a declaration in the head of a for-of loop can't have an initializer:

#### node
SyntaxError: for-of loop variable declaration may not have an initializer.


$ xs -e "for (var x = 1 of [undefined, ,2, false]) { print(x)}"
1
1
2
false
```

PR #2184, opened 2019-05-28
Add tests for NumberFormat constructor edge cases.

Ms2ger:
The behavior here changed in <https://github.com/tc39/ecma402/pull/352>.


leobalter:
the referenced issue was rejected in the Intl meeting.

PR #2185, opened 2019-05-29
Use SameValue in verifyProperty helper

### LINKED ISSUE #1146: Improvements for the verifyProperty helper
2 suggestions:

On https://github.com/tc39/test262/blob/master/harness/propertyHelper.js#L42-L43

It should use the assert.sameValue to properly compare NaN as well.

Also, it still needs implementation to verify `get` and `set` values in accessor descriptors, those won't need to use assert.sameValue.

cc @rwaldron, you might be interested on this.

shvaikalesh:
Closes #1146.

### DIFF
--- a/harness/propertyHelper.js
+++ b/harness/propertyHelper.js
@@ -47,7 +47,7 @@ function verifyProperty(obj, name, desc, options) {
   var failures = [];
 
   if (Object.prototype.hasOwnProperty.call(desc, 'value')) {
-    if (desc.value !== originalDesc.value) {
+    if (!assert._isSameValue(desc.value, originalDesc.value)) {
       failures.push("descriptor value should be " + desc.value);
     }
   }
--- a/harness/propertyHelper.js
+++ b/harness/propertyHelper.js
@@ -79,12 +79,6 @@ function verifyProperty(obj, name, desc, options) {
     }
   }
 
-  if (Object.prototype.hasOwnProperty.call(desc, 'set')) {
-    if (desc.set !== originalDesc.set) {
-      failures.push('descriptor set should be ' + desc.set);
-    }
-  }
-
   assert(!failures.length, failures.join('; '));
 
   if (options && options.restore) {
--- a/harness/propertyHelper.js
+++ b/harness/propertyHelper.js
@@ -73,12 +73,6 @@ function verifyProperty(obj, name, desc, options) {
     }
   }
 
-  if (Object.prototype.hasOwnProperty.call(desc, 'get')) {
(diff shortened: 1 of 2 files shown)


harness/propertyHelper.js
```diff
@@ -73,6 +73,18 @@ function verifyProperty(obj, name, desc, options) {
     }
   }
 
+  if (Object.prototype.hasOwnProperty.call(desc, 'get')) {
```
shvaikalesh: This one is not quite useful.

There are ~30 accessors in spec.
All of them are non-enumerable, configurable, and lack `set` (well, except for `Object#__proto__`).
We can introduce a new helper to check their descriptors:

```js
verifyGetterProperty(Map.prototype, 'size');
```

harness/propertyHelper.js
```diff
@@ -47,7 +47,7 @@ function verifyProperty(obj, name, desc, options) {
   var failures = [];
 
   if (Object.prototype.hasOwnProperty.call(desc, 'value')) {
-    if (desc.value !== originalDesc.value) {
+    if (!assert._isSameValue(desc.value, originalDesc.value)) {
```
leobalter: this one depends on `assert` having a public property `_isSameValue` which not documented and might break if any runner uses a custom harness.

I'd be fine to actually expand this code here to verify for `NaN` cases, but still avoiding an abstraction that is not in this file.

harness/propertyHelper.js
leobalter: I'd prefer deferring this change to when we have usage and tests for it. It's pretty tricky to check/compare the value of the get functions and we might end up not having a proper expected value.

harness/propertyHelper.js
```diff
@@ -73,6 +73,18 @@ function verifyProperty(obj, name, desc, options) {
...
+    if (desc.get !== originalDesc.get) {
+      failures.push('descriptor get should be ' + desc.get);
+    }
+  }
+
+  if (Object.prototype.hasOwnProperty.call(desc, 'set')) {
+    if (desc.set !== originalDesc.set) {
+      failures.push('descriptor set should be ' + desc.set);
+    }
+  }
```
leobalter: same for set, better to defer this until we write tests using it.

shvaikalesh: 5ac819f

harness/propertyHelper.js
shvaikalesh: 4f4d2d6

harness/propertyHelper.js
shvaikalesh: 6935757

PR #2186, opened 2019-05-29
Adding new test cases to validate public class fields initialization.

caiolima:
This PR is adding following missing cases:

1. Following the Spec, `DefineField(receiver, fieldRecord)` calls `CreateDataPropertyOrThrow(receiver, fieldName, initValue)` and this operation can be user observable.
2. Frozen objects can't be changed.
3. `[[DefineOwnProperty]]` is used to define a public field, and it doesn't call setter of `Super`.

CC: @leobalter @caitp @xanlpz

### DIFF
--- a/test/language/statements/class/elements/class-field-is-observable-by-proxy.js
+++ b/test/language/statements/class/elements/class-field-is-observable-by-proxy.js
@@ -18,28 +18,29 @@ features: [class, class-fields-public, Proxy]
 ---*/
 
 let arr = [];
-
-class ProxyBase {
-  constructor() {
-    return new Proxy(this, {
-      defineProperty: function (target, key, descriptor) {
-        arr.push(key);
-        assert(descriptor.enumerable);
-        assert(descriptor.configurable);
-        assert(descriptor.writable);
-        return Reflect.defineProperty(target, key, descriptor);
-      }
-    });
-  }
+let expectedTarget = null;
+function ProxyBase() {
+  expectedTarget = this;
+  return new Proxy(this, {
+    defineProperty: function (target, key, descriptor) {
+      arr.push(key);
+      arr.push(descriptor.value);
+      arr.push(target);
+      assert.sameValue(descriptor.enumerable, true);
+      assert.sameValue(descriptor.configurable, true);
+      assert.sameValue(descriptor.writable, true);
+      return Reflect.defineProperty(target, key, descriptor);
+    }
+  });
 }
 
 class Test extends ProxyBase {
   f = 3;
-  g = "test";
+  g = "Test262";
 }
 
 let t = new Test();
 assert.sameValue(t.f, 3);
-assert.sameValue(t.g, "test");
+assert.sameValue(t.g, "Test262");
 
-assert.compareArray(arr, ["f", "g"]);
+assert.compareArray(arr, ["f", expectedTarget, 3, "g", expectedTarget, "Test262"]);
--- a/test/language/statements/class/elements/class-field-is-observable-by-proxy.js
+++ b/test/language/statements/class/elements/class-field-is-observable-by-proxy.js
@@ -1,0 +1,45 @@
+// Copyright (C) 2019 Caio Lima. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Public class fields initialization calls [[DefineOwnProperty]]
+esid: sec-define-field
+info: |
+  DefineField(receiver, fieldRecord)
+    ...
+    8. If fieldName is a Private Name,
+      a. Perform ? PrivateFieldAdd(fieldName, receiver, initValue).
+    9. Else,
+      a. Assert: IsPropertyKey(fieldName) is true.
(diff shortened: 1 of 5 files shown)


test/language/statements/class/elements/class-field-is-observable-by-proxy.js
```diff
@@ -0,0 +1,45 @@
...
+let arr = [];
+
+class ProxyBase {
+  constructor() {
+    return new Proxy(this, {
+      defineProperty: function (target, key, descriptor) {
+        arr.push(key);
+        assert(descriptor.enumerable);
+        assert(descriptor.configurable);
+        assert(descriptor.writable);
```
leobalter: nit: in the future it's better to be explicit about the expected value being `true`, otherwise `assert()` will only expect for any true'ish vale.

also, good catch here verifying the descriptor created for [the data property](https://tc39.github.io/ecma262/#sec-createdataproperty)

test/language/statements/class/elements/class-field-is-observable-by-proxy.js
```diff
@@ -0,0 +1,45 @@
...
+      b. Perform ? CreateDataPropertyOrThrow(receiver, fieldName, initValue).
+    10. Return.
+includes: [compareArray.js]
+features: [class, class-fields-public, Proxy]
+---*/
+
+let arr = [];
+
+class ProxyBase {
+  constructor() {
```
leobalter: total nit: it could be just a `function ProxyBase() {` here ...

test/language/statements/class/elements/class-field-is-observable-by-proxy.js
```diff
@@ -0,0 +1,45 @@
...
+    return new Proxy(this, {
+      defineProperty: function (target, key, descriptor) {
+        arr.push(key);
```
leobalter: add `arr.push(descriptor.value); arr.push(target);` here.

test/language/statements/class/elements/class-field-is-observable-by-proxy.js
```diff
@@ -0,0 +1,45 @@
...
+class Test extends ProxyBase {
+  f = 3;
+  g = "test";
+}
+
+let t = new Test();
+assert.sameValue(t.f, 3);
+assert.sameValue(t.g, "test");
+
+assert.compareArray(arr, ["f", "g"]);
```
leobalter: now you can assert the given property values and receiver.

test/language/statements/class/elements/private-class-field-on-frozen-objects.js
```diff
@@ -0,0 +1,31 @@
...
+  f = Object.freeze(this);
+  #g = "Test262";
+
+  get g() {
+    return this.#g;
+  }
+}
+
+let t = new Test();
+assert.sameValue(t.g, "Test262");
```
leobalter: assert that `t.f` is also set.

test/language/statements/class/elements/public-class-field-initialization-on-super-class-with-setter.js
```diff
@@ -0,0 +1,34 @@
...
+
+class Base extends Super {
+  f = "Test262";
+}
+
+let o = new Base();
+assert.sameValue(o.f, "Test262");
+verifyEnumerable(o, "f");
+verifyWritable(o, "f");
+verifyConfigurable(o, "f");
```
leobalter:
```js
verifyProperty(o, "f", {
  value: "Test262",
  enumerable: true,
  writable: true,
  configurable: true,
});
```

test/language/statements/class/elements/public-class-field-initialization-on-super-class-with-setter.js
```diff
@@ -0,0 +1,34 @@
...
+      a. Assert: IsPropertyKey(fieldName) is true.
+      b. Perform ? CreateDataPropertyOrThrow(receiver, fieldName, initValue).
+    10. Return.
+includes: [propertyHelper.js]
+features: [class, class-fields-public]
+---*/
+
+class Super {
+  set f(v) {
+    throw new Test262Error();
```
leobalter: it would be great to have something similar from this and the first file:

```js
  return new Proxy(this, {
      defineProperty: function (target, key, descriptor) {
        throw new Test262Error();
      });
  }

...

assert.throws(Test262Error, () => { new Base(); });
```

caiolima:
@leobalter Thank you very much for the review! PR is ready for another look.

test/language/statements/class/elements/private-class-field-on-frozen-objects.js
```diff
@@ -19,8 +19,8 @@ flags: [onlyStrict]
 ---*/
 
 class Test {
-  f = Object.freeze(this);
-  #g = "Test262";
+  f = this;
+  #g = (Object.freeze(this), "Test262");
```
leobalter: why the previous version is not correct?

caiolima: Because it fails to create `f` because the Object is already frozen.

PR #2188, opened 2019-05-30
Verify that private methods and accessors aren't visible to [[GetOwnProperty]] and [[HasProperty]].

### LINKED ISSUE #2187: [[HasOwnProperty]] / [[GetOwnProperty]] does not find private methods/accessors
Private methods and accessors are not stored directly into Objects (see [here](https://tc39.github.io/proposal-private-methods/#sec-private-names)). This way, they can't be retrieved by `[[HasOwnProperty]]` or `[[GetOwnProperty]]`.

caiolima:
It closes #2187.

I'm not sure if I should consider `__lookupGetter__` and `__lookupSetter__` on those tests, since it is deprecated.

CC: @leobalter @jbhoosreddy @mkubilayk

### DIFF
--- a/src/class-elements/private-getter-is-not-a-own-property.case
+++ b/src/class-elements/private-getter-is-not-a-own-property.case
@@ -1,0 +1,48 @@
+// Copyright (C) 2019 Caio Lima (Igalia SL). All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: Private getter is not stored as an own property of objects
+info: |
+  PrivateFieldGet (P, O)
+    1. Assert: P is a Private Name.
+    2. If O is not an object, throw a TypeError exception.
+    3. If P.[[Kind]] is "field",
+      a. Let entry be PrivateFieldFind(P, O).
+      b. If entry is empty, throw a TypeError exception.
+      c. Return entry.[[PrivateFieldValue]].
+    4. Perform ? PrivateBrandCheck(O, P).
+    5. If P.[[Kind]] is "method",
+      a. Return P.[[Value]].
+    6. Else,
+      a. Assert: P.[[Kind]] is "accessor".
+      b. If P does not have a [[Get]] field, throw a TypeError exception.
+      c. Let getter be P.[[Get]].
+      d. Return ? Call(getter, O).
+template: default
+features: [class-methods-private]
+---*/
+
+//- elements
+get #m() { return "Test262"; }
+
+checkPrivateGetter() {
+  assert.sameValue(Object.getOwnPropertyDescriptor(this, "#m"), undefined);
+  assert.sameValue(Reflect.getOwnPropertyDescriptor(this, "#m"), undefined);
+
+  assert.sameValue(this.hasOwnProperty("#m"), false);
+  assert.sameValue(Reflect.has(this, "#m"), false);
+
+  assert.compareArray(Object.getOwnPropertyNames(this), []);
+  assert.compareArray(Reflect.ownKeys(this), []);
+
+  assert.sameValue("#m" in this, false);
+
+  const descriptors = Object.getOwnPropertyDescriptors(this);
+  assert.sameValue("#m" in descriptors, false);
+
+  assert.sameValue(this.#m, "Test262");
+}
+//- assertions
+let c = new C();
+c.checkPrivateGetter();
--- a/src/class-elements/private-getter-is-not-a-own-property.case
+++ b/src/class-elements/private-getter-is-not-a-own-property.case
@@ -27,22 +27,15 @@ features: [class-methods-private]
 get #m() { return "Test262"; }
 
 checkPrivateGetter() {
-  assert.sameValue(Object.getOwnPropertyDescriptor(this, "#m"), undefined);
-  assert.sameValue(Reflect.getOwnPropertyDescriptor(this, "#m"), undefined);
-
   assert.sameValue(this.hasOwnProperty("#m"), false);
-  assert.sameValue(Reflect.has(this, "#m"), false);
-
-  assert.compareArray(Object.getOwnPropertyNames(this), []);
-  assert.compareArray(Reflect.ownKeys(this), []);
-
   assert.sameValue("#m" in this, false);
 
-  const descriptors = Object.getOwnPropertyDescriptors(this);
-  assert.sameValue("#m" in descriptors, false);
+  assert.sameValue(this.__lookupGetter__("#m"), undefined);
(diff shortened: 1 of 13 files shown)


rwaldron:
What do you mean by "deprecated"? Those properties are defined in Annex B, which is required when the host is a browser.

caiolima:
> What do you mean by "deprecated"? Those properties are defined in Annex B, which is required when the host is a browser.

@rwaldron: I meant this warning on [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/__lookupGetter__). It is usually to JS developers, but I was not totally sure if we intended to cover those cases. I'll update tests to include those.

src/class-elements/private-getter-is-not-a-own-property.case
```diff
@@ -0,0 +1,48 @@
...
+    5. If P.[[Kind]] is "method",
+      a. Return P.[[Value]].
+    6. Else,
+      a. Assert: P.[[Kind]] is "accessor".
+      b. If P does not have a [[Get]] field, throw a TypeError exception.
+      c. Let getter be P.[[Get]].
+      d. Return ? Call(getter, O).
+template: default
+features: [class-methods-private]
+---*/
```
leobalter: missing: `includes: [compareArray.js]`

src/class-elements/private-getter-is-not-a-own-property.case
```diff
@@ -0,0 +1,48 @@
...
+  assert.sameValue(this.hasOwnProperty("#m"), false);
+  assert.sameValue(Reflect.has(this, "#m"), false);
+
+  assert.compareArray(Object.getOwnPropertyNames(this), []);
+  assert.compareArray(Reflect.ownKeys(this), []);
+
+  assert.sameValue("#m" in this, false);
+
+  const descriptors = Object.getOwnPropertyDescriptors(this);
+  assert.sameValue("#m" in descriptors, false);
```
leobalter: for the `"#m" in this` I understand you're going through the inheritance chain, but here a hasOwnProperty should do the trick.

src/class-elements/private-getter-is-not-a-own-property.case
```diff
@@ -0,0 +1,48 @@
...
+
+  assert.sameValue(this.#m, "Test262");
+}
```
leobalter: nit: return something here (`0`, `true`, etc) and assert the return after calling `c.checkPrivateGetter();`.

src/class-elements/private-getter-is-not-a-own-property.case
leobalter: One way or another, I wonder if should just have a simplification of all these assertions with the following only:

```js
assert.sameValue(this.hasOwnProperty("#m"), false, '#m is not listed as an own property');
assert.sameValue("#m" in this, false, '#m is also not found in the prototype chain');
```

Other than that it seems we are verifying the built-in methods rather than anything else.

src/class-elements/private-method-is-not-a-own-property.case
```diff
@@ -0,0 +1,49 @@
...
+
+  assert.compareArray(Object.getOwnPropertyNames(this), []);
+  assert.compareArray(Reflect.ownKeys(this), []);
+
+  assert.sameValue("#m" in this, false);
+
+  const descriptors = Object.getOwnPropertyDescriptors(this);
+  assert.sameValue("#m" in descriptors, false);
+
+  assert.sameValue(this.#m(), "Test262");
```
leobalter: same observations from src/class-elements/private-getter-is-not-a-own-property.case can be applied here

src/class-elements/private-setter-is-not-a-own-property.case
```diff
@@ -0,0 +1,49 @@
...
+  assert.compareArray(Object.getOwnPropertyNames(this), []);
+  assert.compareArray(Reflect.ownKeys(this), []);
+
+  assert.sameValue("#m" in this, false);
+
+  const descriptors = Object.getOwnPropertyDescriptors(this);
+  assert.sameValue("#m" in descriptors, false);
+
+  this.#m = "Test262";
+  assert.sameValue(this._v, "Test262");
```
leobalter: same observations from src/class-elements/private-getter-is-not-a-own-property.case can be applied here

src/class-elements/private-getter-is-not-a-own-property.case
caiolima: Oops. Solved.

src/class-elements/private-getter-is-not-a-own-property.case
caiolima: The main idea here is to cover all users of `[[GetOwnProperty]]` and `[[HasOwnProperty]]`. But maybe only testing own properties and going through the inheritance chain can be enough. WDYT?

src/class-elements/private-getter-is-not-a-own-property.case
caiolima: Done.

test/language/statements/class/elements/private-field-is-not-clobbered-by-computed-property.js
```diff
@@ -0,0 +1,43 @@
+// Copyright (C) 2019 Caio Lima (Igalia SL). All rights reserved.
```
caiolima: This is not a generated file.

test/language/statements/class/elements/private-getter-is-not-clobbered-by-computed-property.js
```diff
@@ -0,0 +1,43 @@
+// Copyright (C) 2019 Caio Lima (Igalia SL). All rights reserved.
```
caiolima: This is not a generated test.

test/language/statements/class/elements/private-method-is-not-clobbered-by-computed-property.js
```diff
@@ -0,0 +1,43 @@
+// Copyright (C) 2019 Caio Lima (Igalia SL). All rights reserved.
```
caiolima: This is not a generated test.

test/language/statements/class/elements/private-setter-is-not-clobbered-by-computed-property.js
```diff
@@ -0,0 +1,44 @@
+// Copyright (C) 2019 Caio Lima (Igalia SL). All rights reserved.
```
caiolima: This is not a generated test.

PR #2189, opened 2019-05-31
remove (min|max)imumFractionDigits w/ significant

FrankYFTang:
Per change in https://tc39.github.io/proposal-unified-intl-numberformat/section11/numberformat_diff_out.html
we should not output minimumFractionDigits and maximumFractionDigits if minimumSignificantDigits or maximumSignificantDigits are set.

### DIFF
--- a/test/intl402/NumberFormat/prototype/resolvedOptions/order.js
+++ b/test/intl402/NumberFormat/prototype/resolvedOptions/order.js
@@ -22,8 +22,6 @@ const expected = [
   "currency",
   "currencyDisplay",
   "minimumIntegerDigits",
-  "minimumFractionDigits",
-  "maximumFractionDigits",
   "minimumSignificantDigits",
   "maximumSignificantDigits",
   "useGrouping",


FrankYFTang:
@sffc @littledan @Ms2ger @zbraniecki @anba @gsathya @leobalter

FrankYFTang:
Notice
```
Set intlObj.[[MinimumFractionDigits]] to mnfd.
Set intlObj.[[MaximumFractionDigits]] to mxfd.
```
are moved from setting unconditionally to the Else of
```
14. If mnsd is not undefined or mxsd is not undefined, then
...
15. Else If mnfd is not undefined or mxfd is not undefined, then
...
  e. Set intlObj.[[MinimumFractionDigits]] to mnfd.
  f. Set intlObj.[[MaximumFractionDigits]] to mxfd.
...
17. Else,
  ...
  b. Set intlObj.[[MinimumFractionDigits]] to mnfdDefault.
  c. Set intlObj.[[MaximumFractionDigits]] to mxfdDefault.
```

sffc:
Yes, this is an observable change but it is intentional.  Lgtm

FrankYFTang:
ping

PR #2197, opened 2019-06-12
Add more class declaration formations

jbhoosreddy:
Related Issue: https://github.com/tc39/test262/issues/1343#issuecomment-446225159

This PR attempts to adds additional class declaration formations as per @littledan's test plan

In particular:

   - [x] Types of declarations
       - [x] export default class
       - [x] export class X
       - [x] class declaration at the top level of a script/module
       - [x] class declaration in a block
       - [x] class declaration inside a function at the top level

### DIFF
--- a/src/class-elements/syntax/valid/cls-decl-elements-valid-syntax-export-default.template
+++ b/src/class-elements/syntax/valid/cls-decl-elements-valid-syntax-export-default.template
@@ -1,0 +1,14 @@
+// Copyright (C) 2019 Jaideep Bhoosreddy (Bloomberg LP). All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: prod-ClassElement
+name: class declaration module default export
+path: language/statements/class/elements/syntax/valid/export-default-
+flags: [module]
+features: [class]
+---*/
+
+export default class /*{ heritage }*/{
+  /*{ elements }*/
+}
--- a/test/language/statements/class/elements/syntax/valid/export-default-grammar-class-body-ctor-no-heritage.js
+++ b/test/language/statements/class/elements/syntax/valid/export-default-grammar-class-body-ctor-no-heritage.js
@@ -1,0 +1,22 @@
+// This file was procedurally generated from the following sources:
+// - src/class-elements/grammar-class-body-ctor-no-heritage.case
+// - src/class-elements/syntax/valid/cls-decl-elements-valid-syntax-export-default.template
+/*---
+description: A constructor is valid without a super call in the constructor and heritage (class declaration module default export)
+esid: prod-ClassElement
+features: [class]
+flags: [generated, module]
+info: |
+    ClassTail : ClassHeritageopt { ClassBody }
+
+    It is a Syntax Error if ClassHeritage is not present and the following algorithm evaluates to true:
+      1. Let constructor be ConstructorMethod of ClassBody.
+      2. If constructor is empty, return false.
+      3. Return HasDirectSuper of constructor.
+
+---*/
+
+
+export default class {
+  constructor() {}
+}
--- a/test/language/statements/class/elements/syntax/valid/export-default-grammar-field-classelementname-initializer-alt.js
+++ b/test/language/statements/class/elements/syntax/valid/export-default-grammar-field-classelementname-initializer-alt.js
@@ -1,0 +1,70 @@
+// This file was procedurally generated from the following sources:
+// - src/class-elements/grammar-field-classelementname-initializer-alt.case
+// - src/class-elements/syntax/valid/cls-decl-elements-valid-syntax-export-default.template
+/*---
+description: FieldDefinition, ClassElementName, PropertyName = Initializer Syntax (class declaration module default export)
+esid: prod-ClassElement
+features: [class-fields-public, class]
+flags: [generated, module]
+info: |
+    ClassElement :
+      ...
+      FieldDefinition ;
+      ;
+
+    FieldDefinition :
+      ClassElementName Initializer _opt
+
+    ClassElementName :
+      PropertyName
+      PrivateName
+
+    PropertyName :
+      LiteralPropertyName
+      ComputedPropertyName
+
+    LiteralPropertyName :
+      IdentifierName
+
+    Initializer :
+      = AssignmentExpression
+
+    IdentifierName ::
+      IdentifierStart
+      IdentifierName IdentifierPart
+
+    IdentifierStart ::
+      UnicodeIDStart
+      $
+      _
+      \ UnicodeEscapeSequence
+
+    IdentifierPart ::
+      UnicodeIDContinue
+      $
+      \ UnicodeEscapeSequence
+      <ZWNJ> <ZWJ>
+
+    UnicodeIDStart ::
+      any Unicode code point with the Unicode property "ID_Start"
+
+    UnicodeIDContinue ::
+      any Unicode code point with the Unicode property "ID_Continue"
+
+
+    NOTE 3
+    The sets of code points with Unicode properties "ID_Start" and
+    "ID_Continue" include, respectively, the code points with Unicode
+    properties "Other_ID_Start" and "Other_ID_Continue".
+
+---*/
+
+
+export default class {
+  $ = 1;
+  _ = 2;
+  \u{6F} = 3;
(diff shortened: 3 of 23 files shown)


leobalter:
Can you push the generated tests too? (in a separate commit)

caiolima:
Also, maybe in another PR, but we are missing covering of `__lookupGetter__` and `__lookupSetter__` on accessors. I think we should create new templates for `private-accessors` case. Right now we are using `private-methods` on those tests.

src/class-elements/private-methods/cls-decl-export-default.template
```diff
@@ -0,0 +1,115 @@
...
+
+esid: prod-MethodDefinition
+---*/
+
+/*** template notes
+ * method should always be #m
+ * the template provides c.ref() for external reference
+ */
+
+function hasProp(obj, name, expected, msg) {
```
rwaldron: I'm curious to know why this is here and you're not using `verifyProperty` via `includes: [propertyHelper.js]`

leobalter:
I'm afraid this adds too much without much useful coverage.

It's probably good for tests verifying only the syntactical production, but not for multiplying so many tests. The formats used here are well tested in other places, and maybe these templates should be limited to a minimal production helps leaning the new additions to something less than 7k LOC. WDYT?

cc @gsathya

src/class-elements/private-methods/cls-decl-export-default.template
leobalter: this is from the other templates we already have and could be improved. hasProp is always used for non existing own properties, and verifyProperty is meant for expected to exist properties.

this could be shortened to `hasNoProp(obj, name, msg)` and doing the following assertions:

```js
function hasNoProp(obj, name, msg) {
  var hasOwnProperty = Object.prototype.hasOwnProperty.call(obj, name);
  assert.sameValue(hasOwnProperty, false, msg);

  var hasProperty = Reflect.has(obj, name);
  assert.sameValue(hasProperty, false, msg);

// or `name in obj` instead of Reflect.has
}
```

leobalter: in any case, this should observe any conclusion for https://github.com/tc39/test262/pull/2197#issuecomment-503722953

gsathya:
From an engine perspective, there is difference in how classes are parsed in a module, top level script, inside a block, inside a function (both eager and lazy parsed) -- they trigger different scope analysis paths, so in general these seem ok to me. But I also see @leobalter's concerns here as the spec doesn't really differentiate all these cases as they're more of an "implementation detail" which isn't really under the purview of test262 and these are a lot of tests too. I haven't looked at this in detail but I wonder could just trim down these to test the different scoping rules.

littledan:
Pragmatically, what if we include a few tests for classes in these different environments, without generating all the combinations?

leobalter:
@littledan:

> Pragmatically, what if we include a few tests for classes in these different environments, without generating all the combinations?

This matches my comments:

> It's probably good for tests verifying only the syntactical production

We could extend the templates at the `src/class-elements/syntax/valid` folder for the productions tested here, instead of full cases from `src/class-elements/private-methods`.

src/class-elements/private-methods/cls-decl-export-default.template
rwaldron: Can we move `hasProp` to `propertyHelper.js` so there isn't many copies of it floating around?

leobalter: yes, but this needs to be done in another PR for the existing tests. It should be quick. I'm addressing it tomorrow.

littledan:
Sure, or we could even include a few individual tests (that could also test basic functionality) and not extend any templates.

leobalter:
> Sure, or we could even include a few individual tests (that could also test basic functionality) and not extend any templates.

yes, this would be fine too

jbhoosreddy:
Thanks @gsathya @littledan @leobalter and @rwaldron. So I'll follow @leobalter's suggestion and extend these templates `src/class-elements/syntax/valid`.

If you'd like me to go the other route and not extend templates, I can do that also.

jbhoosreddy:
@leobalter I added those forms in the templates here `/class-elements/syntax/valid` but it still produced 88 files. I'll try to see tomorrow if these tests are worth the maintenance. I may end up writing some non-templated tests instead.

jbhoosreddy:
In today's call @caiolima and I discussed that there might be some value to having these templated tests.
so I'll probably wait for @leobalter 's thoughts on the PR in it's current form.
Happy weekend!

jbhoosreddy:
ping @leobalter for thoughts on this

src/class-elements/syntax/valid/cls-decl-elements-valid-syntax-module.template
```diff
@@ -0,0 +1,14 @@
...
+esid: prod-ClassElement
+name: class declaration module
+path: language/statements/class/elements/syntax/valid/module-
+flags: [module]
+features: [class]
+---*/
+
+class C /*{ heritage }*/{
+  /*{ elements }*/
+}
```
leobalter: how is this related to a module? Just the fact it asserts a class inside a module?

It should better to remove this template. There is a template for Script code that can be reused by the runners as modules if necessary. Otherwise this opens a precedent for every syntax we test in this suite.

src/class-elements/syntax/valid/cls-decl-elements-valid-syntax-inside-function.template
```diff
@@ -0,0 +1,17 @@
...
+features: [class]
+---*/
+
+function getC() {
+  class C /*{ heritage }*/{
+    /*{ elements }*/
+  }
+  return C;
+}
+var C = getC();
```
leobalter: I believe this is really excessive. A single test to verify the production of a class inside a function would me more than enough.

Also, the `var C = getC();` is going beyond a test for valid syntax.

I recommend removing this.

src/class-elements/syntax/valid/cls-decl-elements-valid-syntax-block.template
```diff
@@ -0,0 +1,15 @@
...
+name: class declaration inside block
+path: language/statements/class/elements/syntax/valid/inside-function-
+features: [class]
+---*/
+
+{
+  class C /*{ heritage }*/{
+    /*{ elements }*/
+  }
+}
```
leobalter: a class declaration inside a block is a class declaration and we should limit the tests to itself alone, no expanding to class elements.

I recommend removing this like in src/class-elements/syntax/valid/cls-decl-elements-valid-syntax-inside-function.template

src/class-elements/syntax/valid/cls-decl-elements-valid-syntax-export.template
```diff
@@ -0,0 +1,14 @@
...
+esid: prod-ClassElement
+name: class declaration module export
+path: language/statements/class/elements/syntax/valid/export-
+flags: [module]
+features: [class]
+---*/
+
+export class C /*{ heritage }*/{
+  /*{ elements }*/
+}
```
leobalter: I still believe a single test containing a `export ClassDeclaration` is sufficient. This does not observe special syntax for elements in this position.

src/class-elements/syntax/valid/cls-decl-elements-valid-syntax-export-default.template
```diff
@@ -0,0 +1,14 @@
...
+esid: prod-ClassElement
+name: class declaration module default export
+path: language/statements/class/elements/syntax/valid/export-default-
+flags: [module]
+features: [class]
+---*/
+
+export default class C /*{ heritage }*/{
+  /*{ elements }*/
+}
```
leobalter: Maybe this is the only compelling template in this PR, IMO as the class declaration has an odd variation that is not found elsewhere.

In this case, the variation would be defining a declaration without a BindingIdentifier. This would be a production I believe that is good enough for checking syntax.

In this case, I would just change this template code to be like:

```
export default class /*{ heritage }*/{
  /*{ elements }*/
}
```

anba:
I've removed the tests again as part of #2097 (https://github.com/tc39/test262/pull/2097/commits/64b6b844ebfe1c82645d0ff7f325d4453f338cad), because they were violating <https://github.com/tc39/test262/blob/master/src/class-elements/conventions.md>, which rendered them useless.

PR #2199, opened 2019-06-15
Add unary operators with NewTarget test

shvaikalesh:
JSC bug: [Early error on ANY operator before new.target](https://bugs.webkit.org/show_bug.cgi?id=157970).

### DIFF
--- a/test/language/expressions/new.target/unary-expr.js
+++ b/test/language/expressions/new.target/unary-expr.js
@@ -1,0 +1,38 @@
+// Copyright (C) 2019 Alexey Shvayka. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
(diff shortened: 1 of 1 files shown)


test/language/expressions/new.target/unary-expr.js
```diff
@@ -0,0 +1,38 @@
...
+  ~(new.target);
+}
+
+function c() {
+  delete void typeof +-~! (new.target);
+}
+
+async function d() {
+  await new.target;
+}
```
leobalter: it would be nice to call these functions and assert the values... Test262 might be the only useful place to assert `delete void typeof +-~! (new.target)` is `true`.

shvaikalesh: 07bb71f

PR #2203, opened 2019-06-19
Add poisoned __proto__ tests

shvaikalesh:
JSC bug: [super should not depend on `__proto__`](https://bugs.webkit.org/show_bug.cgi?id=157972).


claudepache:
Another JSC bug caught by the first two tests (they fail as soon as the class construct with extends clause is evaluated): https://bugs.webkit.org/show_bug.cgi?id=205848

PR #2204, opened 2019-06-19
Remove duplicated `with` statement tests

shvaikalesh:
The same tests are in `statements/with` directory:
* [unscopables-get-err.js](https://github.com/tc39/test262/blob/master/test/language/statements/with/unscopables-get-err.js)
* [has-property-err.js](https://github.com/tc39/test262/blob/master/test/language/statements/with/has-property-err.js)

They have better naming and more precise `assert.throws`.


leobalter:
these are not duplicate tests, they are verifying similar abrupt completions using an expression rather than a statement. That's why you found them to be like the others. The production is slightly different.

cc @jugglinmike.

PR #2211, opened 2019-06-21
Add a test for FinalizationGroupCleanupIteratorPrototype

leobalter:
Also includes related tests for FinalizationGroup#cleanupSome

I plan to add more tests but these are ready to review. 

The %FinalizationGroupCleanupIteratorPrototype% Object
https://tc39.github.io/proposal-weakrefs/#sec-%finalizationgroupcleanupiterator%-object

- [x] is an ordinary object.
- [x] has a [[Prototype]] internal slot whose value is the intrinsic object %IteratorPrototype%.

%FinalizationGroupCleanupIteratorPrototype%.next ( )
https://tc39.github.io/proposal-weakrefs/#sec-%finalizationgroupcleanupiterator%.next

1. Let iterator be the this value.
2. [x] If Type(iterator) is not Object, throw a TypeError exception.
3. [x] If iterator does not have a [[FinalizationGroup]] internal slot, throw a TypeError exception.
4. Let finalizationGroup be iterator.[[FinalizationGroup]].
5. If finalizationGroup is undefined, then
  a. [x] Return CreateIterResultObject(undefined, true).
6. Assert: finalizationGroup has [[Cells]] and [[IsFinalizationGroupCleanupJobActive]] internal slot.
7. [x] If finalizationGroup.[[IsFinalizationGroupCleanupJobActive]] is false, throw a TypeError exception.
8. If finalizationGroup.[[Cells]] contains a Record cell such that cell.[[Target]] is empty,
  a. Choose any such cell.
  b. Remove cell from finalizationGroup.[[Cells]].
  c. [x] Return CreateIterResultObject(cell.[[Holdings]], false).
9. [x] Otherwise, return CreateIterResultObject(undefined, true).

%FinalizationGroupCleanupIteratorPrototype% [ @@toStringTag ]
https://tc39.github.io/proposal-weakrefs/#sec-%finalizationgroupcleanupiteratorprototype%-@@tostringtag

- [x] The initial value of the @@toStringTag property is the String value "FinalizationGroup Cleanup Iterator".
- [x] This property has the attributes { [[Writable]]: false, [[Enumerable]]: false, [[Configurable]]: true }.

cc @mhofman @gsathya @littledan

### DIFF
--- a/test/built-ins/FinalizationGroup/prototype/cleanupSome/return-undefined-with-gc.js
+++ b/test/built-ins/FinalizationGroup/prototype/cleanupSome/return-undefined-with-gc.js
@@ -1,0 +1,84 @@
+// Copyright (C) 2019 Leo Balter. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-finalization-group.prototype.cleanupSome
+description: Return undefined regardless the result of CleanupFinalizationGroup
+info: |
+  FinalizationGroup.prototype.cleanupSome ( [ callback ] )
+
+  1. Let finalizationGroup be the this value.
+  2. If Type(finalizationGroup) is not Object, throw a TypeError exception.
+  3. If finalizationGroup does not have a [[Cells]] internal slot, throw a TypeError exception.
+  4. If callback is not undefined and IsCallable(callback) is false, throw a TypeError exception.
+  5. Perform ! CleanupFinalizationGroup(finalizationGroup, callback).
+  6. Return undefined.
+features: [FinalizationGroup, arrow-function, async-functions, async-iteration, class, host-gc-required]
+---*/
+
+var called;
+var fn = function() {
+  called += 1;
+  return 39;
+};
+var cb = function() {
+  called += 1;
+  return 42;
+};
+var poisoned = function() { throw new Test262Error(); };
+var fg = new FinalizationGroup(fn);
+
+function emptyCells(custom) {
+  called = 0;
+  if (!custom) {
+    custom = fg;
+  }
+  (function() {
+    var o = {};
+    custom.register(o);
+  })();
+  $262.gc();
+}
+
+emptyCells();
+assert.sameValue(fg.cleanupSome(cb), undefined, 'regular callback');
+assert.sameValue(called, 1);
+
+emptyCells();
+assert.sameValue(fg.cleanupSome(fn), undefined, 'regular callback, same FG cleanup function');
+assert.sameValue(called, 1);
+
+emptyCells();
+assert.sameValue(fg.cleanupSome(() => 1), undefined, 'arrow function');
+
+emptyCells();
+assert.sameValue(fg.cleanupSome(fg.cleanupSome), undefined, 'cleanupSome itself');
+
+emptyCells();
+assert.sameValue(fg.cleanupSome(poisoned), undefined, 'poisoned');
+
+emptyCells();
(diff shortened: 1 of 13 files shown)


leobalter:
as the CI won't capture the results with flags:

```
test262-harness --hostType=d8 --hostPath=v8 --hostArgs='--harmony-weak-refs --expose-gc  --noincremental-marking' $(git diff --name-only master test/)
FAIL test/built-ins/FinalizationGroup/FinalizationGroupCleanupIteratorPrototype/Symbol.toStringTag.js (default)
  descriptor value should be FinalizationGroup Cleanup Iterator

FAIL test/built-ins/FinalizationGroup/FinalizationGroupCleanupIteratorPrototype/Symbol.toStringTag.js (strict mode)
  descriptor value should be FinalizationGroup Cleanup Iterator

FAIL test/built-ins/FinalizationGroup/FinalizationGroupCleanupIteratorPrototype/next-job-not-active-throws.js (default)
  Iter should fail if not called during the cleanupSome call Expected a TypeError to be thrown but no exception was thrown at all

FAIL test/built-ins/FinalizationGroup/FinalizationGroupCleanupIteratorPrototype/next-job-not-active-throws.js (strict mode)
  Iter should fail if not called during the cleanupSome call Expected a TypeError to be thrown but no exception was thrown at all

FAIL test/built-ins/FinalizationGroup/prototype/cleanupSome/iterator-holdings-multiple-values.js (default)
  Expected no error, got TypeError: Cannot convert a Symbol value to a string

FAIL test/built-ins/FinalizationGroup/prototype/cleanupSome/iterator-holdings-multiple-values.js (strict mode)
  Expected no error, got TypeError: Cannot convert a Symbol value to a string

Ran 26 tests
20 passed
6 failed
```

rwaldron:
Wow, excellent coverage!

test/built-ins/FinalizationGroup/prototype/cleanupSome/return-undefined-with-gc.js
```diff
@@ -0,0 +1,84 @@
...
+assert.sameValue(called, 1);
+
+emptyCells();
+assert.sameValue(fg.cleanupSome(() => 1), undefined, 'arrow function');
+
+emptyCells();
+assert.sameValue(fg.cleanupSome(fg.cleanupSome), undefined, 'cleanupSome itself');
+
+emptyCells();
+assert.sameValue(fg.cleanupSome(poisoned), undefined, 'poisoned');
```
mhofman: Um that doesn't seem right! I don't think the intent was for `cleanupSome` to swallow user errors here.
Will open a spec issue

leobalter: I'll review it this week

mhofman:
>     FAIL test/built-ins/FinalizationGroup/prototype/cleanupSome/iterator-holdings-multiple-values.js (default)
>       Expected no error, got TypeError: Cannot convert a Symbol value to a string

Interesting, my tests had missed that one!

test/built-ins/FinalizationGroup/prototype/cleanupSome/return-undefined-with-gc.js
mhofman: https://github.com/tc39/proposal-weakrefs/pull/145

leobalter: this test is incorrect for sure, my bad. I'll get it fixed this week when I'm back on test262 time.

PR #2212, opened 2019-06-24
feat: adding tests for optional chaining proposal

bcoe:
starting to work on a few tests for optional chaining as I read and review the proposal:

## Punctuators

- [x] ?.[lookahead ∉ DecimalDigit] [link](https://github.com/tc39/test262/pull/2212/files#diff-33a81e6f5b8c579cc77a2bd78d65b1b1).

## Left-Hand-Side Expressions

OptionalExpression:
   
- [x] MemberExpression OptionalChain
- [x] CallExpression OptionalChain
- [x] OptionalExpression OptionalChain

OptionalChain:

- [x] ?. [ expression ]
- [x] ?. IdentifierName
- [x] ?. Arguments
- [x] ?. TemplateLiteral  (this production is an early error). [link](https://github.com/tc39/test262/pull/2212/files#diff-13bec8e378e2faca88c8ed0f30faa43a)
- [x] OptionalChain [ expression ]
- [x] OptionalChain IdentifierName
- [x] OptionalChain Arguments
- [x] OptionalChain TemplateLiteral (this production is an early error).

## Static Semantics: IsValidSimpleAssignmentTarget

- [x] LeftHandSideExpression: OptionalExpression  [link](https://github.com/tc39/test262/pull/2212/files#diff-e40e9de1ab929495989843b75f82e653).


test/language/optional-chaining/destructuring.js
```diff
@@ -0,0 +1,20 @@
+/*---
+esid: pending
+info: |
+  destructuring is not supported for properties of optional chain
+description: >
+  Static Semantics: IsDestructuring
+    LeftHandSideExpression:
+      CallExpression
```
bcoe: @jridgewell @DanielRosenwasser @rkirsling, I read this rule as stating that destructuring is not supported in an optional chain, I also gathered that this was the outcome of:

https://github.com/tc39/proposal-optional-chaining/issues/74

This test is failing because the current babel implementation supports destructuring.

Am I correct in my reading of this part of the proposal, if so I think we'll want to update the babel implementation.

rkirsling: Hmm, it seems that IsDestructuring is only checked from [ForIn/OfBodyEvaluation](https://tc39.es/ecma262/#sec-runtime-semantics-forin-div-ofbodyevaluation-lhs-stmt-iterator-lhskind-labelset), and I think we're just saying that if you see an OptionalExpression as an LHS, then (just like a CallExpression) that's not an example of destructuring.

So I don't know that there's anything that needs to be tested here—note that this test doesn't actually involve an optional chain at present, and the linked idea about optional destructuring is something that would require additional syntax.

bcoe: @rkirsling I understand, it's the concept of "optional destructuring" that's out of scope, and the proposal was something along the lines of `foo?.bar?`; my bad misreading that section of the spec.

leobalter:
I recommend you create a checklist of all the new productions possible by the new syntax rather than having just generic tests. Perhaps, one test here and there might work nice for feature check but there is still a big chunk of missing coverage here.

bcoe:
@leobalter I fleshed out the issue body, based on your recommendations. Is that looking like what you were picturing, if so, any recommendations for naming conventions of tests?

test/language/expressions/optional-chaining/early-errors-tail-position-template-string.js
```diff
@@ -0,0 +1,21 @@
...
+      OptionalChain TemplateLiteral
+features: [optional-chaining]
+negative:
+  type: SyntaxError
+  phase: parse
+---*/
+
+$DONOTEVALUATE();
+
+const a = {fn: () => {}};
```
rwaldron: Non-blocking, entirely out of curiosity question: Why the arrow function here, and not just `{fn(){}}`?

test/language/expressions/optional-chaining/member-expression-optional-chain.js
```diff
@@ -0,0 +1,51 @@
...
+const fn = (arg1, arg2) => {
+  return arg1 + arg2;
+}
+const i = 0;
+const obj = {
+  a: 'hello',
+  b: {val: 13},
+  c: (arg1) => {
+    return arg1 * 2;
+  },
```
rwaldron: Same as above... non-blocking, entirely out of curiosity question: Why the arrow function here, and not just `{c(arg){ return arg * 2 }}`?

bcoe: no good reason, went ahead and used the more concise form.

bcoe:
@leobalter I offered to review the spec text of optional chaining, as part of the entrance criteria for stage 3.

Writing tests has been helping me better dive into edge-cases of the proposal, and get a better feel for the specification.

I'd love to help see these tests further along, as it's a great learning experience. Would you rather that I close this draft until we've reached stage 3 approval? or is it okay if I continue putting some work into it.

leobalter:
Closing is not necessary but we won't be able to merge the tests until it reaches stage 3.

I'm ok to leave this open, including this new "Draft" option in the PR as a good flag. As I pointed out there is a good chunk of coverage to be done yet. I might wait until Stage 3 for another review, but feel free to ping me for any questions.

bcoe:
@leobalter I made an effort to enumerate more test cases today, reading through the spec. Let me know if this is looking more like it.

https://gist.github.com/bcoe/b0e6ae3f93f8c8b8f8559f4428187f25

test/language/expressions/optional-chaining/call-expression-optional-chain.js
```diff
@@ -0,0 +1,17 @@
+/*---
```
leobalter: Can you add the copyright headers into each file? They are required to part tests correctly in different test runners used to consume Test262.

test/language/expressions/optional-chaining/call-expression-optional-chain.js
```diff
@@ -0,0 +1,17 @@
...
+esid: pending
```
leobalter: please use `esid: prod-OptionalExpression` instead of any pending in the test files.

test/language/expressions/optional-chaining/call-expression-optional-chain.js
```diff
@@ -0,0 +1,17 @@
...
+info: |
+  optional chain on call expression
+description: >
```
leobalter: invert usage of info and description. Description is the small line, info is the large chunk.

You may keep the block characters `info: |` (pipe) and `description: >` (greater than)

bcoe:
@leobalter I have a fairly light weekend, I'll make an effort to put in a few hours expanding the cases outlined in the gist.

leobalter:
> @leobalter I have a fairly light weekend, I'll make an effort to put in a few hours expanding the cases outlined in the gist.

Sure! If you can first address the small items in here, I can already land these tests. I've wrote some other tests (related) to #2262 and it's just a very small chunk from the complete coverage.

test/language/expressions/optional-chaining/early-errors-tail-position-template-string.js
```diff
@@ -0,0 +1,21 @@
...
+  type: SyntaxError
+  phase: parse
+---*/
+
+$DONOTEVALUATE();
+
+const a = {fn() {}};
+
+a?.fn
+  `hello`;
```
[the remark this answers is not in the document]
bcoe: I was specifically trying to address this section of the spec:

> NOTE This production exists in order to prevent automatic semicolon insertion rules (11.9) to be applied to the following code.

... however, by adding a semi-colon to the end of the expression I was failing to actually exercise this behavior, I've split it into another test.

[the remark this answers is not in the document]
leobalter: thanks for the clarification

test/language/expressions/optional-chaining/runtime-semantics-evaluation.js
```diff
@@ -0,0 +1,18 @@
...
+  accessing optional value on undefined or null returns undefined.
+info: |
+  If baseValue is undefined or null, then
+    Return undefined.
+features: [optional-chaining]
+---*/
+
+const nul = null;
+const undf = undefined;
+assert.sameValue(undefined, nul?.a);
```
leobalter: for a follow up:

```js
var x = 0;
var y = null?.[(function() { x += 1 }())];

// assert value of x to verify if the `[Expression]` has got evaluated
// assert y is undefined.
```

leobalter: oh! you did that in `test/language/expressions/optional-chaining/short-circuiting.js`, nice!

test/language/expressions/optional-chaining/static-semantics-simple-assignment.js
```diff
@@ -0,0 +1,23 @@
...
+negative:
+  type: SyntaxError
+  phase: parse
+---*/
+
+$DONOTEVALUATE();
+
+const obj = {};
+
+obj?.a = 33;
```
leobalter: nice!

bcoe:
@leobalter :tada: I have [this PR](https://github.com/bcoe/test262/pull/1) open on my fork of test262, which I've been adding a more thorough set of grammar productions to.

I will make this PR official now that this has landed, and address your comments in it.

PR #2214, opened 2019-06-26
Adding tests to validate HomeObject setup on private methods and accessors

caiolima:
Those tests are adding tests to verify if private methods and accessors have `[[HomeObject]]` configured correctly.

Cc. @jbhoosreddy @mkubilayk


caiolima:
Thx for the reviews @rwaldron and @leobalter.

PR #2222, opened 2019-07-03
Correct features to Intl.DateTimeFormat-dayPeriod

FrankYFTang:
Mistakes which I didn't catch while reviewing #2217


leobalter:
thanks!

FrankYFTang:
change all 6 files now. Please review and submit

PR #2223, opened 2019-07-03
Add tests for rest parameter with "use strict" (expecting syntax error)

gibson042:


leobalter:
nice one! Sounds like most of the tests for `ContainsUseStrict is true` + `IsSimpleParameterList of FormalParameters is false` are only matching the form of a default parameter `x = 0`.

This could cover more parts not limited to rest, but also using Object or Array patterns. e.g.: `function fn ([x]) { 'use strict'; }`, etc.

Would you mind adding the mentioned cases here as well? Also, please feel free to generate the tests and commit them (separately is even better). This way we get a quick feedback from the CI.

gibson042:
New test cases added and new tests generated.

src/function-forms/array-destructuring-param-strict-body.case
```diff
@@ -0,0 +1,103 @@
+// Copyright (C) 2018 Richard Gibson. All rights reserved.
```
rwaldron: Should this be 2019?

PR #2225, opened 2019-07-03
Fix test:diff npm scripts

gibson042:
* Include untracked files in `npm run diff`
* Suppress non-file output from nested `npm run diff`


leobalter:
there is another issue with this diff script where it's still capturing a diff from the files added in previous commits.

PR #2228, opened 2019-07-10
Added tests to cover access of private members on inner functions

caiolima:
We are adding test cases to verify that private members(fields, methods and accessors) are visible on scope of inner functions and arrow functions.


caiolima:
@leobalter Thx a lot for the review!

PR #2231, opened 2019-07-11
Add missing BigInt to Number conversion case

caiolima:
There is an infinite set of `BigInt` values that can be represented as Numbers. This is the case when a `BigInt` is out of safe integer range, for example. This PR is adding such cases to `Number(BigInt)` constructor case. Some [bugs](https://bugzilla.mozilla.org/show_bug.cgi?id=1558538) with those edge cases were reported on implementations already.

P.S.: This PR is inspired by the bug mentioned above.

### DIFF
--- a/test/built-ins/Number/bigint-conversion.js
+++ b/test/built-ins/Number/bigint-conversion.js
@@ -9,3 +9,15 @@ features: [BigInt]
 
 assert.sameValue(Number(0n), 0);
 assert.sameValue(+(new Number(0n)), +(new Number(0)));
+
+assert.sameValue(Number(2n**53n), 9007199254740992);
+assert.sameValue(Number(2n**53n + 1n), 9007199254740992);
+assert.sameValue(Number(2n**53n + 2n), 9007199254740994);
+assert.sameValue(Number(2n**53n + 3n), 9007199254740996);
+assert.sameValue(Number(2n**53n + 4n), 9007199254740996);
+
+assert.sameValue(Number(-(2n**53n)), -9007199254740992);
+assert.sameValue(Number(-(2n**53n + 1n)), -9007199254740992);
+assert.sameValue(Number(-(2n**53n + 2n)), -9007199254740994);
+assert.sameValue(Number(-(2n**53n + 3n)), -9007199254740996);
+assert.sameValue(Number(-(2n**53n + 4n)), -9007199254740996);


test/built-ins/Number/bigint-conversion.js
```diff
@@ -9,3 +9,15 @@ features: [BigInt]
 
 assert.sameValue(Number(0n), 0);
 assert.sameValue(+(new Number(0n)), +(new Number(0)));
+
+assert.sameValue(Number(2n**53n), 9007199254740992);
```
leobalter: These tests are wrong. The numbers here are represented above the max safe integer - observable at `Number.MAX_SAFE_INTEGER` or `9007199254740991` - and there isn't a guarantee you can see an implementation returning anything above it.

You could use assertions limited to `Number(2n**53n - 1n)` instead.

Checkout the specs for the representable Number values. https://tc39.es/ecma262/#sec-ecmascript-language-types-number-type

caiolima: Maybe I'm misunderstanding the spec, but there seems to be defined rule for BingInt -> Number conversion when BigInt is out of Safe Integer range. Looking here the definition of [the Number value for x](https://tc39.es/ecma262/#number-value), there is the following text:

> Choose the member of this set that is closest in value to x. If two values of the set are equally close, then the one with an even significand is chosen...

And looking into [Number constructor](https://tc39.es/proposal-bigint/#sec-number-constructor-number-value) change, we can see in 2.b: `If Type(prim) is BigInt, let n be the Number value for prim.`.

That's is basically the section I'm trying to cover with those changes.

leobalter: From the specs:

> Note that all the positive and negative mathematical integers whose magnitude is no greater than 253 are representable in the Number type (indeed, the mathematical integer 0 has two representations, +0 and -0).

There is no normative consistency in the specs for numbers >= `(2 ** 53) - 1` using the Number value. Anything above is implementation-defined, therefore we can't assert them here.

leobalter: Take the example of `2 ** 54` in V8:

```
x = 2 ** 54; // 18014398509481984
x === (x + 1); // true

y = 18014398509481983; // equivalent of 2 ** 54 - 1
x === y; // true
```
```

At some point, the implementations are not safe about number representation. This is one of the biggest reasons we have BigInt today.

Comment (caiolima) [reply test/built-ins/Number/bigint-conversion.js]: I think I follow what you mean. Up to know, I was sure that Numbers out of `2**53 - 1` where specified as [double-precision 64-bit binary format IEEE 754-2008 value](https://tc39.es/ecma262/#sec-terms-and-definitions-number-value), meaning that all implementations should have a defined behavior. I'm not an expert on double precision numbers, but my understanding form spec was that such conversion where well defined. If that's not the case, we should close this PR.

Comment (leobalter) [reply test/built-ins/Number/bigint-conversion.js]: I believe it's ok to convert these assertions in the boundaries of representable numbers, like `Number(2n ** 53n - 1n)` and it's negative form too.

Comment (jmdyck) [reply test/built-ins/Number/bigint-conversion.js]: @leobalter:
> 
> There is no normative consistency in the specs for numbers >= `(2 ** 53) - 1` using the Number value. Anything above is implementation-defined, therefore we can't assert them here.

Can you point to where the spec *says* it's implementation-defined? I don't think it does.

Comment (allenwb) [reply test/built-ins/Number/bigint-conversion.js]: > There is no normative consistency in the specs for numbers >= (2 ** 53) - 1 using the Number value. Anything above is implementation-defined, therefore we can't assert them here.

What exactly do you mean by the above?  The result for all operations upon IEEE-768 values are normatively specified by the I(EEE spec and hence normatively specified by ECMA=262. Because of the limited precision of the IEEE format, some results are not mathematically precise but the specific result is still well specified. 

The only allowable implementation variation in the  current specification involves   determining the "Number Value" of a literal numeric string:

> Otherwise, the rounded value must be the Number value for the MV (in the sense defined in 6.1.6), unless the literal includes a StrUnsignedDecimalLiteral and the literal has more than 20 significant digits, in which case the Number value may be either the Number value for the MV of a literal produced by replacing each significant digit after the 20th with a 0 digit or the Number value for the MV of a literal produced by replacing each significant digit after the 20th with a 0 digit and then incrementing the literal at the 20th digit position. A digit is significant if it is not part of an ExponentPart and
>
>it is not 0; or
>there is a nonzero digit to its left and there is a nonzero digit, not in the ExponentPart, to its right.

Comment (leobalter) [reply test/built-ins/Number/bigint-conversion.js]: > Because of the limited precision of the IEEE format, some results are not mathematically precise but the specific result is still well specified.

This is what I wanted to express when I mentioned no normative consistency.  Maybe I should have said "no observable consistency" instead.

The result might be well specified, but it's not well observed as the limited precision we have flakiness in the used values. e.g. in my examples where `2 ** 54 === (2 ** 54) - 1; // true`. The assertion might just never fail over the precision boundaries and I personally don't consider it relevant.

Comment (allenwb) [reply test/built-ins/Number/bigint-conversion.js]: > This is what I wanted to express when I mentioned no normative consistency. Maybe I should have said "no observable consistency" instead.

It is observable.  In you example, you are observing that those two values are both represented by the same IEEE binary 64-bit value.  It would be a spec. violation if the result returned was false.

You could also observe this by bit-encoding IEEE values using Typed Array overlays.

Comment (leobalter) [reply test/built-ins/Number/bigint-conversion.js]: it is too fragile to rely on tests. Because the limited precision makes a number return true to more than one comparison with other numbers...

```js
2 ** 53 === 9007199254740992; // true
2 ** 53 + 1 === 9007199254740992; // true
```

What I'm suggesting is to have assertions with values within the precision boundaries.

jmdyck: > the limited precision makes a number return true to more than one comparison with other numbers...

Yes! And that's the behavior that the spec *requires*. So it's worth testing.

PR #2232, opened 2019-07-11
Remove tests for "quarter" option of Intl.DateTimeFormat

FrankYFTang:
During July 2019 ECMA402 meeting, we believe the issue with 'quarter' is too complicated and need to bring to CLDR to address first. Therefore, the PR is now consider stage 1 and we should remove tests until the PR is in good shape.

@leobalter @sffc @srl295 
see https://github.com/tc39/ecma402/pull/345 for more detail.


leobalter:
Thanks for the cleanup!

PR #2235, opened 2019-07-12
Check static private methods on super class are called with correct receiver

mkubilayk:
Adds coverage for (static) private methods according to the test plan posted in #1343.

> Semantics > Error on missing method
> - Subclass receiver for static private

cc: @caiolima @jbhoosreddy

### DIFF
--- a/src/class-elements/static-private-method-subclass-receiver.case
+++ b/src/class-elements/static-private-method-subclass-receiver.case
@@ -1,0 +1,33 @@
+// Copyright (C) 2019 Kubilay Kahveci (Bloomberg LP). All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: Static private methods on the super-class cannot be called with sub-class as the receiver
+info: |
+  PrivateFieldGet (P, O)
+    1. Assert: P is a Private Name.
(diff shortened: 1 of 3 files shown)


src/class-elements/static-private-method-subclass-receiver.case
```diff
@@ -0,0 +1,33 @@
...
+    4. Perform ? PrivateBrandCheck(O, P).
+    5. If P.[[Kind]] is "method",
+      a. Return P.[[Value]].
+    ...
+
+  PrivateBrandCheck(O, P)
+    1. If O.[[PrivateBrands]] does not contain an entry e such that SameValue(e, P.[[Brand]]) is true,
+      a. Throw a TypeError exception.
+template: default
+features: [class-static-methods-private]
```
leobalter: please add public as well

mkubilayk: I couldn't find a feature tag for static public methods in the codebase and they are not part of https://github.com/tc39/proposal-static-class-features/. Should I just create one in `features.txt`?

leobalter: yes

PR #2238, opened 2019-07-15
Add coverage for non object values of the Properties arg of Object.create

### LINKED ISSUE #2237: Object.create properties parameter can be a non-object
JSC got a bug report on this. It appears all other engines pass the test but I didn't see a test262 checking correct behavior. I made this test for JSC, perhaps it could be ported to test262?

```js
let toObjectablePrimitives = [true, false, 1, 2, "hello", Symbol(), BigInt(1)];

for (let primitive of toObjectablePrimitives)
    Object.create({}, 2);

try {
    Object.create({}, null);
} catch (e) {
    if (!(e instanceof TypeError))
        throw e;
}
```

leobalter:
Closes #2237

cc @kmiller68

### DIFF
--- a/test/built-ins/Object/create/properties-arg-to-object-bigint.js
+++ b/test/built-ins/Object/create/properties-arg-to-object-bigint.js
@@ -1,0 +1,35 @@
+// Copyright (C) 2019 Leo Balter. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-object.create
+description: >
+  The Properties argument is cast to an object if it's a BigInt value
+info: |
+  Object.create ( O, Properties )
+
+  3. If Properties is not undefined, then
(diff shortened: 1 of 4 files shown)


test/built-ins/Object/create/properties-arg-to-object-bigint.js
```diff
@@ -0,0 +1,35 @@
...
+  3. If Properties is not undefined, then
+    a. Return ? ObjectDefineProperties(obj, Properties).
+
+  Runtime Semantics: ObjectDefineProperties ( O, Properties )
+
+  2. Let props be ? ToObject(Properties).
+  3. Let keys be ? props.[[OwnPropertyKeys]]().
+  ...
+  // All enumerable keys are added to the created object.
+features: [symbol, BigInt]
```
rwaldron: Remove `symbol`

PR #2241, opened 2019-07-15
Adding case where PrivateName is used inside computed property.

caiolima:
PrivateNames are visible on `ComputedProperty` evaluation of classes. This PR is adding tests to verify those cases for private fields, methods and accessors.

### DIFF
--- a/src/class-elements/grammar-privatename-in-computed-property-missing.case
+++ b/src/class-elements/grammar-privatename-in-computed-property-missing.case
@@ -1,0 +1,24 @@
+// Copyright (C) 2019 Caio Lima (Igalia SL). All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: Use of udeclared PrivateName in ComputedProperty is a syntax error
+info: |
+  ClassTail : ClassHeritage { ClassBody }
+    1. Let lex be the LexicalEnvironment of the running execution context.
+    2. Let classScope be NewDeclarativeEnvironment(lex).
+    3. Let classScopeEnvRec be classScope's EnvironmentRecord.
+    ...
+    15. Set the running execution context's LexicalEnvironment to classScope.
+    16. Set the running execution context's PrivateEnvironment to classPrivateEnvironment.
+    ...
+    27. For each ClassElement e in order from elements
+      a. If IsStatic of e is false, then
+        i. Let field be the result of ClassElementEvaluation for e with arguments proto and false.
+    ...
+features: [class-fields-private]
+template: syntax/invalid
+---*/
+
+//- elements
+[this.#f] = 'Test262'
--- a/src/class-elements/grammar-privatename-in-computed-property-missing.case
+++ b/src/class-elements/grammar-privatename-in-computed-property-missing.case
@@ -2,21 +2,40 @@
 // This code is governed by the BSD license found in the LICENSE file.
 
 /*---
-desc: Use of udeclared PrivateName in ComputedProperty is a syntax error
+desc: Use of undeclared PrivateName in ComputedProperty is a syntax error
 info: |
-  ClassTail : ClassHeritage { ClassBody }
-    1. Let lex be the LexicalEnvironment of the running execution context.
-    2. Let classScope be NewDeclarativeEnvironment(lex).
-    3. Let classScopeEnvRec be classScope's EnvironmentRecord.
-    ...
-    15. Set the running execution context's LexicalEnvironment to classScope.
-    16. Set the running execution context's PrivateEnvironment to classPrivateEnvironment.
-    ...
-    27. For each ClassElement e in order from elements
-      a. If IsStatic of e is false, then
-        i. Let field be the result of ClassElementEvaluation for e with arguments proto and false.
-    ...
-features: [class-fields-private]
+  ClassElementName:
+    PropertyName
+    PrivateIdentifier
+
+  PropertyName:
+    LiteralPropertyName
+    ComputedPropertyName
+
+  ComputedPropertyName:
+    [ AssignmentExpression ]
+
+  AssignmentExpression ... MemberExpression
+
+  MemberExpression:
+    MemberExpression . PrivateName
+
+  Static Semantics: AllPrivateIdentifiersValid
+    AllPrivateIdentifiersValid is an abstract operation which takes names as an argument.
+
+    MemberExpression : MemberExpression . PrivateIdentifier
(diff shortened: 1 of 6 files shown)


src/class-elements/grammar-privatename-in-computed-property-missing.case
```diff
@@ -0,0 +1,24 @@
+// Copyright (C) 2019 Caio Lima (Igalia SL). All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: Use of udeclared PrivateName in ComputedProperty is a syntax error
```
leobalter: typo

src/class-elements/grammar-privatename-in-computed-property-missing.case
```diff
@@ -0,0 +1,24 @@
...
+    3. Let classScopeEnvRec be classScope's EnvironmentRecord.
+    ...
+    15. Set the running execution context's LexicalEnvironment to classScope.
+    16. Set the running execution context's PrivateEnvironment to classPrivateEnvironment.
+    ...
+    27. For each ClassElement e in order from elements
+      a. If IsStatic of e is false, then
+        i. Let field be the result of ClassElementEvaluation for e with arguments proto and false.
+    ...
+features: [class-fields-private]
```
leobalter: public too

src/class-elements/grammar-privatename-in-computed-property-missing.case
```diff
@@ -0,0 +1,24 @@
+// Copyright (C) 2019 Caio Lima (Igalia SL). All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: Use of udeclared PrivateName in ComputedProperty is a syntax error
+info: |
+  ClassTail : ClassHeritage { ClassBody }
+    1. Let lex be the LexicalEnvironment of the running execution context.
```
leobalter: This info is using a Syntax template, therefore it would be nice to use the matching grammar:

```
ClassElementName:
  PropertyName
  PrivateIdentifier

PropertyName:
  LiteralPropertyName
  ComputedPropertyName

ComputedPropertyName:
  [ AssignmentExpression ]

AssignmentExpression ... MemberExpression

MemberExpression:
  MemberExpression . PrivateName
```

It also needs to specify where the invalid syntax is happening then, once the current text is showing part of the [Evaluation phase](https://tc39.es/proposal-class-fields/#runtime-semantics-class-definition-evaluation).

src/class-elements/grammar-privatename-in-computed-property-missing.case
caiolima: Oops. Fixed.

mathiasbynens:
Some of these file names have typos in them. `prviate` --> `private`

anba:
@caiolima Is there a missing `strict` directive in <https://github.com/tc39/test262/blob/master/test/language/statements/class/elements/private-field-is-visible-in-computed-properties.js>?

The test currently requires a `TypeError`, but that error is only thrown when the complete test file is run in strict mode. In non-strict mode, the test case should expect a `ReferenceError`, shouldn't it?

strict-mode case:
- `this.#f` throws a `TypeError`, because `this` is `undefined`. (RequireObjectCoercible is called before MakePrivateReference.)

non-strict mode case:
- `this` is the global object, so the RequireObjectCoercible check passes.
- MakePrivateReference is called, where ResolveBinding returns a reference to the uninitialised binding for `#f`, triggering a `ReferenceError` in step 3 when GetValue is called, cf. GetBindingValue step 3.

caiolima:
> @caiolima Is there a missing `strict` directive in https://github.com/tc39/test262/blob/master/test/language/statements/class/elements/private-field-is-visible-in-computed-properties.js?

I think this test wrong. There is a proposed PR to fix it in https://github.com/tc39/test262/pull/2284.

> The test currently requires a `TypeError`, but that error is only thrown when the complete test file is run in strict mode. In non-strict mode, the test case should expect a `ReferenceError`, shouldn't it?
> strict-mode case:
> 
> * `this.#f` throws a `TypeError`, because `this` is `undefined`. (RequireObjectCoercible is called before MakePrivateReference.)

I'm having a hard time to understand why it should be `TypeError` on strict mode. Do you mind placing in the spec where it says that `this` is going to be `undefined`? IIUC, it should be the global object also.

> non-strict mode case:
> 
> * `this` is the global object, so the RequireObjectCoercible check passes.
> * MakePrivateReference is called, where ResolveBinding returns a reference to the uninitialised binding for `#f`, triggering a `ReferenceError` in step 3 when GetValue is called, cf. GetBindingValue step 3.

You are right here.

anba:
> I'm having a hard time to understand why it should be `TypeError` on strict mode. Do you mind placing in the spec where it says that `this` is going to be `undefined`? IIUC, it should be the global object also.

[OrdinaryCallBindThis](https://tc39.es/ecma262/#sec-ordinarycallbindthis) only changes `null` and `undefined` to the global object in non-strict mode. And the this-value for a function call with a non-reference is `undefined` per [EvaluateCall](https://tc39.es/ecma262/#sec-evaluatecall), step 2.

caiolima:
> [OrdinaryCallBindThis](https://tc39.es/ecma262/#sec-ordinarycallbindthis) only changes `null` and `undefined` to the global object in non-strict mode. And the this-value for a function call with a non-reference is `undefined` per [EvaluateCall](https://tc39.es/ecma262/#sec-evaluatecall), step 2.

But I'm not sure how `OrdinaryCallBindThis` is related with this test. IIUC, the `PrimaryExpression : this` is going to call [ResolveThisBinding ](https://tc39.es/ecma262/#sec-resolvethisbinding) that is going to resolve to global object, because it is going to fallback to global Environment Record.

IIRC, `OrdinaryCallBindThis` is called whenever we do a function call without a reciever and that's when the result of strict and sloppy mode differs.

I also tested the following code into Safari and Chrome and the result is the same in both strict and sloppy mode:

```
this.x = 'foo';

class C {
  [this.x]() {
    return 'test';
  }
}

let c = new C();

c.foo(); // returns test
```

caiolima:
> [OrdinaryCallBindThis](https://tc39.es/ecma262/#sec-ordinarycallbindthis) only changes `null` and `undefined` to the global object in non-strict mode. And the this-value for a function call with a non-reference is `undefined` per [EvaluateCall](https://tc39.es/ecma262/#sec-evaluatecall), step 2.

Oh, I get it now. This test is inside a function! I think we should use another `base` to avoid this result, since the part of spec we want to test is not this one.

PR #2251, opened 2019-07-22
Fix structure for numeric literal tests

leobalter:


leobalter:
Merging as this is a minor patch.

PR #2253, opened 2019-07-22
Add tests for BigInt disallowed in LegacyOctalLike and NonOctal literals

leobalter:
Ref https://github.com/tc39/proposal-bigint/issues/208

cc @rkirsling

### DIFF
--- a/test/language/literals/bigint/legacy-octal-like-invalid-07n.js
+++ b/test/language/literals/bigint/legacy-octal-like-invalid-07n.js
@@ -1,4 +1,4 @@
-// Copyright (C) 2017 Robin Templeton. All rights reserved.
+// Copyright (C) 2019 Leo Balter. All rights reserved.
 // This code is governed by the BSD license found in the LICENSE file.
 
 /*---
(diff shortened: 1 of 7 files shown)


test/language/literals/bigint/non-octal-like-invalid-09n.js
```diff
@@ -0,0 +1,28 @@
...
+    NonZeroDigit DecimalDigits_opt BigIntLiteralSuffix
+features: [BigInt]
+negative:
+  phase: parse
+  type: SyntaxError
+---*/
+
+$DONOTEVALUATE();
+
+08n;
```
bakkot: Given the filename, this should perhaps be `09n`?

test/language/literals/bigint/legacy-octal-like-invalid-07n.js
```diff
@@ -0,0 +1,28 @@
+// Copyright (C) 2017 Robin Templeton. All rights reserved.
```
caiolima: Is this header correct?

leobalter: I copy and past the files structure usually from the same folder. I changed this file but somehow forgot to save it. Thanks!

leobalter:
Thanks @rkirsling, @bakkot, @devsnek, and @caiolima for the reviews!

leobalter:
I merged it too early. The non octal parts need consensus from TC39. I'll hold on reverting as the meeting starts tomorrow, let's see what should be done to fix this here.

bakkot:
For posterity: the change to the proposal [got consensus](https://github.com/tc39/proposal-bigint/issues/208#issuecomment-514385025), so these tests need not be reverted.

PR #2265, opened 2019-08-01
 Add cases for dstr assignment w/ inner edgy LHSExpressions

### LINKED ISSUE #2262: Add destructuring assignment tests for object literal accessor property targets
Creating this issue as a result of [this v8 bug](https://bugs.chromium.org/p/v8/issues/detail?id=9560).

Could we add tests for destructuring assignment to object literal accessor properties? Examples:

```js
[{ set prop(val) { console.log(val); } }.prop = 12 ] = [ 1 ];
({ obj: { set prop(val) { console.log(val); } }.prop = 12 } = { obj: 1 });
({ obj: { set prop(val) { console.log(val); } }.prop } = { obj: 1 });
[{ set prop(val) { console.log(val); } }.prop ] = [ 1 ];
```

leobalter:
cc @joeywatts

Closes #2262

### DIFF
--- a/test/language/expressions/assignment/dstr/array-elem-put-obj-literal-prop-ref-init.js
+++ b/test/language/expressions/assignment/dstr/array-elem-put-obj-literal-prop-ref-init.js
@@ -1,0 +1,69 @@
+// This file was procedurally generated from the following sources:
+// - src/dstr-assignment/array-elem-put-obj-literal-prop-ref-init.case
+// - src/dstr-assignment/default/assignment-expr.template
+/*---
+description: The DestructuringAssignmentTarget of an AssignmentElement can extend to LHSExpressions if it is neither an ObjectLiteral nor an ArrayLiteral and its AssignmentTargetTyp is simple. Using MemberExpression (ObjLiteral + identifier) with initializer. (AssignmentExpression)
+esid: sec-variable-statement-runtime-semantics-evaluation
+es6id: 13.3.2.4
+features: [destructuring-binding]
+flags: [generated]
+info: |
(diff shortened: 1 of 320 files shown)


test/language/expressions/assignment/dstr/array-elem-put-obj-literal-prop-ref-init.js
```diff
@@ -0,0 +1,69 @@
+// This file was procedurally generated from the following sources:
+// - src/dstr-assignment/array-elem-put-obj-literal-prop-ref-init.case
+// - src/dstr-assignment/default/assignment-expr.template
+/*---
+description: The DestructuringAssignmentTarget of an AssignmentElement can extend to LHSExpressions if it is neither an ObjectLiteral nor an ArrayLiteral and its AssignmentTargetTyp is simple. Using MemberExpression (ObjLiteral + identifier) with initializer. (AssignmentExpression)
+esid: sec-variable-statement-runtime-semantics-evaluation
+es6id: 13.3.2.4
```
rwaldron: Get rid of these please

leobalter: these are originally from the template files, which I prefer to do in another PR, replacing them with the proper esid's if necessary.

leobalter: in any case, these are valid ES6 codes.

leobalter: The new commits address only the templates used in here.

PR #2269, opened 2019-08-02
Remove duplicated es6id if esid exists

leobalter:
Includes generated test files and tests files from the built-ins folder. Ignore regular tests in the language folder as it contains tests with EOL special characters where the editor modifies it in the automatic search and replace.


leobalter:
To make sure this PR is safe for EOL special characters, the status saying it does have 0 additions is a goal to be kept here. The lint will verify if the tags were safely removed.

rwaldron:
Nice

PR #2270, opened 2019-08-02
 Add cases for Optional Expressions as DestructuringAssignmentTarget

leobalter:
cc @ljharb @bcoe

### DIFF
--- a/src/dstr-assignment/array-elem-nested-memberexpr-optchain-prop-ref-init.case
+++ b/src/dstr-assignment/array-elem-nested-memberexpr-optchain-prop-ref-init.case
@@ -1,0 +1,55 @@
+// Copyright (C) 2019 Leo Balter. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: prod-DestructuringAssignmentTarget
+desc: >
+  It is a Syntax Error if LeftHandSideExpression of an DestructuringAssignmentTarget is neither an
+  ObjectLiteral nor an ArrayLiteral and AssignmentTargetType(LeftHandSideExpression) is not simple
+  Using Object
+  (MemberExpression OptionalChain .IdentifierName Initializer)
+info: |
+  Syntax
+
+  AssignmentElement : DestructuringAssignmentTarget Initializer_opt
+  DestructuringAssignmentTarget : LeftHandSideExpression
+
+  Static Semantics: Early Errors
+
+  OptionalExpression:
+    MemberExpression OptionalChain
+    CallExpression OptionalChain
+    OptionalExpression OptionalChain
+
+  OptionalChain:
+    ?. [ Expression ]
+    ?. IdentifierName
+    ?. Arguments 
+    ?. TemplateLiteral
+    OptionalChain [ Expression ]
+    OptionalChain .IdentifierName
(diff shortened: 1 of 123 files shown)


src/dstr-assignment/array-elem-nested-memberexpr-optchain-prop-ref-init.case
```diff
@@ -0,0 +1,55 @@
...
+  type: SyntaxError
+---*/
+
+//- setup
+$DONOTEVALUATE();
+var x = {};
+//- elems
+[x?.y = 42]
+//- vals
+[23]
```
bcoe: what purpose does `[23]` serve here?

src/dstr-assignment/array-elem-nested-memberexpr-optchain-prop-ref.case
```diff
@@ -0,0 +1,55 @@
...
+negative:
+  phase: parse
+  type: SyntaxError
+---*/
+
+//- setup
+$DONOTEVALUATE();
+var x = {};
+//- elems
+[x?.y]
```
bcoe: :point_up: does `[x.y]` without a RHS evaluate as a destructuring assignment?

I think I would have assumed that this was an `[ Expression ]`. Would this production work:

```
const arr = [x?.y, x?.z];
```

src/dstr-assignment/array-elem-put-obj-literal-optchain-prop-ref.case
```diff
@@ -0,0 +1,57 @@
...
+---*/
+
+//- setup
+$DONOTEVALUATE();
+//- elems
+[{
+  set y(val) {
+    throw new Test262Error('The property should not be accessed.');
+  }
+}?.y]
```
bcoe: same question as above, probably me failing to read this production properly.

rkirsling:
I'm not sure if some of these tests were correct?

None of these are a SyntaxError with `.`, so I don't believe `?.` should change that:
https://github.com/tc39/test262/blob/master/test/language/statements/for-of/dstr/obj-prop-elem-target-obj-literal-optchain-prop-ref.js
https://github.com/tc39/test262/blob/master/test/language/statements/for-of/dstr/array-elem-nested-memberexpr-optchain-prop-ref.js
https://github.com/tc39/test262/blob/master/test/language/statements/for-of/dstr/obj-prop-elem-target-memberexpr-optchain-prop-ref.js
https://github.com/tc39/test262/blob/master/test/language/statements/for-of/dstr/array-elem-put-obj-literal-optchain-prop-ref.js
https://github.com/tc39/test262/blob/master/test/language/statements/for-in/dstr/obj-prop-elem-target-obj-literal-optchain-prop-ref.js
https://github.com/tc39/test262/blob/master/test/language/statements/for-in/dstr/obj-prop-elem-target-memberexpr-optchain-prop-ref.js
https://github.com/tc39/test262/blob/master/test/language/statements/for-in/dstr/array-elem-put-obj-literal-optchain-prop-ref.js
https://github.com/tc39/test262/blob/master/test/language/statements/for-in/dstr/array-elem-nested-memberexpr-optchain-prop-ref.js

leobalter:
@rkirsling I edited your comment to fix the links

PR #2273, opened 2019-08-07
tools: update lint.py to work with pip 10+ some more

Ms2ger:
main() and InstallRequirement were also removed from the public API.

### DIFF
--- a/tools/lint/lint.py
+++ b/tools/lint/lint.py
@@ -5,11 +5,12 @@
 import argparse
 import inflect
 import os
-import pip
 try:
-    from pip._internal.req import parse_requirements
+    from pip._internal import main as pip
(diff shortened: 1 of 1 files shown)


leobalter:
While this looks good to me, I'd like to get feedback from @jugglinmike.

jugglinmike:
I'm not able to test this locally, but I trust @Ms2ger knows what's up.

Still, we ought to be testing the environments that we intend to support. That means extending the CI configuration to run these tests with pip 10 (likely as a follow-up, unless anyone involved can make quick work of it here).

Ms2ger:
I'm not planning to try to set up CI for this at this time, so I'd prefer deferring that to a followup.

leobalter:
Let's land this for now and we can quickly react to any further report if it causes an unexpected problems.

Issue #2218, opened 2019-06-28
Top Level Await


rwaldron:
ref: https://tc39.es/proposal-top-level-await/

@leobalter please fill in the following work lists and commence work on new test material: 

- [ ]  [AsyncFunctionStart ( promiseCapability, asyncFunctionBody )](https://tc39.es/proposal-top-level-await/#sec-async-functions-abstract-operations-async-function-start)
  - [ ] ...
- [ ]  [AsyncBlockStart ( promiseCapability, asyncBody, asyncContext )](https://tc39.es/proposal-top-level-await/#sec-asyncblockstart)
  - [ ] ...
- [ ]  [Abstract Module Records](https://tc39.es/proposal-top-level-await/#sec-abstract-module-records)
  - [ ] ...
- [ ]  [Cyclic Module Records](https://tc39.es/proposal-top-level-await/#sec-cyclic-module-records)
  - [ ] ...

- [ ] Table 3: Additional Fields of Cyclic Module Records

  - [ ] [[Async]] 
    - [ ] ...
  - [ ] [[AsyncEvaluating]] true or false Whether this module is currently awaiting async fulfillment.
    - [ ] ...
  - [ ] [[TopLevelCapability]]
    - [ ] ...
  - [ ] [[AsyncParentModules]]
    - [ ] ...
  - [ ] [[PendingAsyncDependencies]]
    - [ ] ...

- [ ] Table 4: Additional Abstract Methods of Cyclic Module Records
  - [ ] ExecuteModule( [ promiseCapability ] )
    - [ ] ... 
- [ ] [Link ( ) Concrete Method](https://tc39.es/proposal-top-level-await/#sec-moduledeclarationlinking)
  - [ ] ...
- [ ] InnerModuleLinking ( module, stack, index )
  - [ ] ...
- [ ] Evaluate ( ) Concrete Method
  - [ ] ...
- [ ] InnerModuleEvaluation( module, stack, index )
  - [ ] ...
- [ ] ExecuteAsyncModule ( module )
  - [ ] ...
- [ ] GetCycleRoot ( module )
  - [ ] ...
- [ ] AsyncModuleExecutionFulfilled ( module )
  - [ ] ...
- [ ] AsyncModuleExecutionRejected ( module, error )
  - [ ] ...
- [ ] Source Text Module Records
  - [ ] ...
- [ ] ParseModule ( sourceText, realm, hostDefined )
  - [ ] ...
- [ ] ExecuteModule ( [ capability ] )
[...]

PR #2274, opened 2019-08-07
More tests for Top Level Await

leobalter:
Ref #2218

### DIFF
--- a/test/language/module-code/top-level-await/if-await-expr.js
+++ b/test/language/module-code/top-level-await/if-await-expr.js
@@ -1,0 +1,32 @@
+// Copyright (C) 2019 Leo Balter. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: >
+  Evaluate Await expression for IfStatement
+info: |
+  ModuleItem:
+    StatementListItem[~Yield, +Await, ~Return]
+
+  ...
+
+  UnaryExpression[Yield, Await]
+    [+Await]AwaitExpression[?Yield]
+
+  AwaitExpression[Yield]:
+    await UnaryExpression[?Yield, +Await]
+esid: prod-AwaitExpression
+flags: [module, async]
+features: [top-level-await]
+---*/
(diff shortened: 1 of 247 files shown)


leobalter:
I'm gonna ready this for review.

There is a final part pending which is evaluating importing and exporting modules using top level await. I want to work in a separate PR for that.

I'm intentionally covering a subset of allowed syntaxes for TLA here, there are much more I could do. If you have any suggestions in my, please let me know and I'd be happy to add.

leobalter:
I just added more tests I forgot to commit.

test/language/module-code/top-level-await/if-await-expr.js
```diff
@@ -0,0 +1,32 @@
...
+---*/
+
+var completed = 0;
+var p = Promise.resolve(false);
+
+if (await p) {
+  completed += 1;
+}
+
+assert.sameValue(completed, 1);
```
leobalter: seems like this might be wrong, verify (reminder for myself)

test/language/module-code/top-level-await/syntax/catch-parameter.js
```diff
@@ -0,0 +1,47 @@
...
+
+  CatchParameter[Yield, Await]:
+    BindingIdentifier[?Yield, ?Await]
+    BindingPattern[?Yield, ?Await]
+esid: prod-AwaitExpression
+flags: [module]
+features: [top-level-await, dynamic-import]
+---*/
+
+try {} catch (await e) {}
```
leobalter: this test is wrong, remove (reminder to myself in the future)

leobalter:
This is ready for review. I was able to check the tests with V8 using the TLA flag and the only failures were related to the **top level `for await`** syntax. I believe the tests remain correct.

leobalter:
The only missing part for TLA is checking importing modules with the TLA peaces. Otherwise, everything seems fine. I plan to add that in a follow up PR.

PR #2275, opened 2019-08-09
Unified Intl.NumberFormat: Various tests

### LINKED ISSUE #2233: Add additional tests for Unified NumberFormat
https://github.com/tc39/proposal-unified-intl-numberformat

Some things that should be tested:

1. Compact rounding (also see https://github.com/tc39/proposal-unified-intl-numberformat/issues/47):

```
{ style: "compact" }
```

| Input | Expected Compact Output, en-US |
|---|---|
| 0.0123 | 0.012 |
| 0.123 | 0.12 |
| 1.23 | 1.2 |
|12.3 | 12 |
| 123 | 123 |
| 1230 | 1.2K |
| 12300 | 12K |
| 123000 | 123K |
| 1230000 | 1.2M |

2. Measurement unit: perhaps say, the string should not be the same when a unit is applied.

```js
for (let simpleUnit of [ /* all supported units */ ]) {
  check(simpleUnit);
  for (let simpleUnit2 of [ /* all supported units */ ]) {
    check(simpleUnit + "-per-" + simpleUnit2);
    check(simpleUnit2 + "-per-" + simpleUnit);
  }
}

function check(unit) {
  const s1 = (123).toLocaleString(undefined, { style: "unit", unit: unit });
  const s2 = (123).toLocaleString();
  assert(s1 != s2);
  // consider a more sophisticated check, like, the difference between s1 and s2 should contain non-empty or letter-like characters
}
```

3. If necessary, more tests about sign display.  Test the two tables in the proposal readme:

```
{ signDisplay: x }
```

signDisplay | -1  | -0  | 0   | 1
----------- | --- | --- | --- | ---
auto        | -1  | -0  | 0   | 1
always      | -1  | -0  | +0  | +1
never       | 1   | 0   | 0   | 1
except-zero | -1  | -0  | 0   | +1

```
{ style: "currency", currency: "USD", currencyDisplay: "accounting", signDisplay: x }
```

signDisplay | -1     | -0     | 0      | 1
----------- | ------ | ------ | ------ | ------
auto        | ($1.00) | ($0.00) | $0.00   | $1.00
always      | ($1.00) | ($0.00) | +$0.00  | +$1.00
never       | $1.00   | $0.00   | $0.00   | $1.00
except-zero | ($1.00) | ($0.00) | $0.00   | +$1.00

4. Something about scientific notation, such as the rounding strategy for scientific notation, which is specified and consistent across locales.

```
{ notation: x }
```

| Input | Expected Output, scientific | Expected Output, engineering |
|---|---|---|
| 543 | 5.43E2 | 543 |
| 5430 | 5.43E3 | 5.43E3 |
| 543000 | 5.43E5 | 543E3 |
| 543211.1 | 5.432E5 | 543.211E5 |

The last case tests that numbers should be rounded to the default of 3 fraction digits.

Ms2ger:
Fixes #2233 .

### DIFF
--- a/test/intl402/NumberFormat/prototype/format/unit-de-DE.js
+++ b/test/intl402/NumberFormat/prototype/format/unit-de-DE.js
@@ -1,0 +1,69 @@
+// Copyright 2019 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-intl.numberformat.prototype.format
+description: Checks handling of the unit style.
+locale: [de-DE]
+features: [Intl.NumberFormat-unified]
+---*/
+
+
+const tests = [
+  [
+    -987,
+    {
+      "short": "-987 m",
+      "narrow": "-987 m",
+      "long": "-987 Meter",
+    }
+  ],
+  [
+    -0.001,
+    {
+      "short": "-0,001 m",
+      "narrow": "-0,001 m",
+      "long": "-0,001 Meter",
+    }
+  ],
+  [
+    -0,
+    {
+      "short": "-0 m",
+      "narrow": "-0 m",
+      "long": "-0 Meter",
+    }
+  ],
+  [
+    0,
+    {
+      "short": "0 m",
+      "narrow": "0 m",
+      "long": "0 Meter",
+    }
+  ],
+  [
+    0.001,
+    {
+      "short": "0,001 m",
+      "narrow": "0,001 m",
+      "long": "0,001 Meter",
+    }
+  ],
+  [
+    987,
+    {
+      "short": "987 m",
+      "narrow": "987 m",
+      "long": "987 Meter",
+    }
+  ],
+];
+
+for (const [number, expectedData] of tests) {
+  for (const [unitDisplay, expected] of Object.entries(expectedData)) {
+    const nf = new Intl.NumberFormat("de-DE", { style: "unit", unit: "meter", unitDisplay });
+    assert.sameValue(nf.format(number), expected);
+  }
+}
+
--- a/test/intl402/NumberFormat/prototype/formatToParts/engineering-scientific-zh-TW.js
+++ b/test/intl402/NumberFormat/prototype/formatToParts/engineering-scientific-zh-TW.js
@@ -21,6 +21,11 @@ function verifyFormatParts(actual, expected, message) {
 
 const tests = [
   [
+    "0.000345",
+    [{"type":"integer","value":"345"},{"type":"exponentSeparator","value":"E"},{"type":"exponentMinusSign","value":"-"},{"type":"exponentInteger","value":"6"}],
+    [{"type":"integer","value":"3"},{"type":"decimal","value":"."},{"type":"fraction","value":"45"},{"type":"exponentSeparator","value":"E"},{"type":"exponentMinusSign","value":"-"},{"type":"exponentInteger","value":"4"}],
+  ],
+  [
     "0.345",
     [{"type":"integer","value":"345"},{"type":"exponentSeparator","value":"E"},{"type":"exponentMinusSign","value":"-"},{"type":"exponentInteger","value":"3"}],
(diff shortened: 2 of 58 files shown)


Ms2ger:
CC @sffc

test/intl402/NumberFormat/prototype/formatToParts/engineering-scientific-zh-TW.js
```diff
@@ -0,0 +1,66 @@
...
+    [{"type":"integer","value":"543"},{"type":"exponentSeparator","value":"E"},{"type":"exponentInteger","value":"3"}],
+    [{"type":"integer","value":"5"},{"type":"decimal","value":"."},{"type":"fraction","value":"43"},{"type":"exponentSeparator","value":"E"},{"type":"exponentInteger","value":"5"}],
+  ],
+  [
+    "543211.1",
+    [{"type":"integer","value":"543"},{"type":"decimal","value":"."},{"type":"fraction","value":"211"},{"type":"exponentSeparator","value":"E"},{"type":"exponentInteger","value":"3"}],
+    [{"type":"integer","value":"5"},{"type":"decimal","value":"."},{"type":"fraction","value":"432"},{"type":"exponentSeparator","value":"E"},{"type":"exponentInteger","value":"5"}],
+  ],
+];
+
```
sffc: You should add a formatToParts scientific/exponent test that has a negative exponent, such that we get the exponentMinusSign field.

test/intl402/NumberFormat/prototype/formatToParts/signDisplay-currency-en-US.js
```diff
@@ -0,0 +1,60 @@
...
+  for (let i = 0; i < actual.length; ++i) {
+    assert.sameValue(actual[i].type, expected[i].type, `${message}: parts[${i}].type`);
+    assert.sameValue(actual[i].value, expected[i].value, `${message}: parts[${i}].value`);
+  }
+}
+
+const tests = [
+  [
+    "auto",
+    [{"type":"literal","value":"("},{"type":"currency","value":"$"},{"type":"integer","value":"987"},{"type":"decimal","value":"."},{"type":"fraction","value":"00"},{"type":"literal","value":")"}],
```
sffc: I made a bug to verify that this is correct:

https://github.com/tc39/proposal-unified-intl-numberformat/issues/68

Ms2ger:
I added a few tests around units; not sure if it covers "unit fields". Let me know if you have thoughts about other tests.

test/intl402/NumberFormat/constructor/unit.js
```diff
@@ -0,0 +1,40 @@
...
+    assert.sameValue(nf.resolvedOptions().unit, undefined);
+    assert.sameValue(called, 1);
+  }
+}
+
+const nf = new Intl.NumberFormat([], {
+  style: "percent",
+});
+assert.sameValue(nf.resolvedOptions().style, "percent");
+assert.sameValue(nf.resolvedOptions().unit, "percent");
```
sffc: This should be undefined as per the spec.  This is a relatively recent change.

https://github.com/tc39/proposal-unified-intl-numberformat/pull/57

test/intl402/NumberFormat/prototype/format/unit-de-DE.js
```diff
@@ -0,0 +1,69 @@
...
+---*/
+
+
+const tests = [
+  [
+    -987,
+    {
+      "short": "-987 m",
+      "narrow": "-987 m",
+      "long": "-987 Meter",
```
sffc: This test might be too sensitive to locale data.  @spectranaut @leobalter, thoughts?  If this is OK, then we don't need further changes.  If we need to change this to be less sensitive to locale data, we could do something along the lines of:

1. Verify that `formatToParts` output, when concatenated, equals `format` output (actually, this is probably something all the `formatToParts` tests should do anyhow)
2. Verify that there is a nonempty entry in the `formatToParts` output with `type: "unit"`

leobalter: I believe the data is just fine, considering this is a specific test for a specific locale. At one point there is some mitigation over having tests for the same units using different locales, therefore it's not required - although not forbidden - to have multiple.

At the same time, the idea for an almost generic test using formatToParts would be a nice addition, like in using a default locale. I'm +1 to these two items without necessarily removing the current items.

PR #2277, opened 2019-08-09
Lint: Support Python 3

Ms2ger:
In light of the planned end-of-life of Python 2 at the end of the year, this ensures that the lint tool supports Python 3 as well. The CI continues running the lint using Python 2, while the unit tests for the lint tool itself are run using both Python 2 and 3, to avoid regressions.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,6 +62,18 @@ jobs:
       # - run:
       #     name: "Run deploy"
       #     command: ./tools/scripts/deploy.sh
+  "Project lint unit tests on Python 3":
+    docker:
+      - image: circleci/python:3.7.4
+    working_directory: ~/test262
+    steps:
+      - checkout
+      - run:
+          name: "Install requirements for linter tool"
+          command: python -m pip install --user --requirement tools/lint/requirements.txt
+      - run:
+          name: "Test the lint tool"
+          command: ./tools/lint/test/run.py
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -62,6 +62,18 @@ jobs:
...
+      - image: circleci/python:3.7.4
+    working_directory: ~/test262
+    steps:
+      - checkout
+      - run:
+          name: "Install requirements for linter tool"
+          command: python -m pip install --user --requirement tools/lint/requirements.txt
+      - run:
+          name: "Test the lint tool"
+          command: ./tools/lint/test/run.py
```
leobalter: thanks for adding this!

leobalter: Btw, can we add the `"Project lint unit tests on Python 3"` to the workflows, after line 139 in this file?

I was going to do it in a separate PR but I think it's good to have it going green in this PR as well.

leobalter:
The CI should go green for the Python 3 run, right? Do you know what might be blocking it?

Ms2ger:
It looks like it's trying to build PyYAML from source, and failing. My suspicion is that the current pinned version (3.11) is too old; current is 5.1.2. I've dropped the commit to run in CI for now; I'll try to take another look next week.

leobalter:
Please, rebuild this with the most recent changes.

PR #2279, opened 2019-08-12
Tools: Update PyYAML version.

### LINKED ISSUE #2278: Consider updating PyYAML
We're on 3.11; current is 5.1.2.

Ms2ger:
Fixes #2278.


tools/generation/lib/template.py
```diff
@@ -154,20 +154,20 @@ def _frontmatter(self, case_filename, case_values):
         features += self.attribs['meta'].get('features', [])
         features = list(OrderedDict.fromkeys(features))
         if len(features):
-            lines += ['features: ' + re.sub('\n\s*', ' ', yaml.dump(features).strip())]
+            lines += ['features: ' + re.sub('\n\s*', ' ', yaml.dump(features, default_flow_style=True).strip())]
```
Ms2ger: Default changed in https://github.com/yaml/pyyaml/pull/256.

leobalter: It is weird how the library picks an option where a _default_ option needs to explicitly set as `True`. The actual default action is without the _default_ option is the equivalent of `False`. Sounds very counterintuitive.

Anyway, it's not a problem of Test262 and thanks for fixing this up.

PR #2280, opened 2019-08-12
CI: Don't mark CI as broken if a new test fails.

Ms2ger:
In general, the CI job should only be marked as broken (with a red cross) if
the result should block the PR from landing (or if the result should at least
be carefully reviewed before landing).

Having failing tests in some implementations, on the other hand, does not block
the PR. On the contrary, it means the test is valuable, because it proves the
presence of a bug in the implementation.

The current setup is unnecessarily confusing, particularly for newcomers to the
project.

### DIFF
--- a/tools/scripts/ci_test.sh
+++ b/tools/scripts/ci_test.sh
@@ -13,12 +13,5 @@ if [ "$CIRCLE_PULL_REQUEST" != "" ]; then
   echo ""
 
   echo "Running the tests with test262-harness"
-  test262-harness -t 1 --hostType=$hostType --hostPath=$hostPath --hostArgs="$hostArgs" -- $paths | tee exec.out
-
-  if grep -q '^[1-9][0-9]* failed$' exec.out; then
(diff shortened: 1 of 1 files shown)


leobalter:
The original request from implementors using Test262 was to show information if we got tests failing. 

That's why we use an explicit Required flag for a few of checks. Other checks do not block the PR from being merged.

This change doesn't provide anything good to the information used in the PRs. We are not testing the test runner, and this would be the only thing to not bring a green check if we apply this change. If the feature is not useful it should then be totally removed, and not providing false positives.

Maintaining this project regularly, it's pretty clear to me if I have test executions going red but at the same time I get a green check on the lint + build processes.

leobalter:
cc @mathiasbynens @gsathya

leobalter:
also, I would be happy to shut this feature down if someone creates a bot that execute new and modified tests for each new PR (and updates within) reporting any tests failing. The CI is the best immediate solution for now.

PR #2282, opened 2019-08-12
Tests to cover simple declaration of static fields.

caiolima:
Adding tests to validate public static fields declaration.

Cc @mkubilayk, @jbhoosreddy.


rwaldron:
@leobalter @spectranaut

caiolima:
Thx a lot for the review!

PR #2283, opened 2019-08-12
Added cases to cover public static field initializer operations with 'this' biding.

caiolima:
Following the spec, `this` is referring to class constructor when used into static field initializer. We are adding tests with this usage to validate such behavior. It also includes tests of:

- `this` referenve inside direct eval;
- Redeclaration;
- Initialization that throws exception;
-  `this` reference inside arrow function;


caiolima:
Thx a lot for the review.

PR #2284, opened 2019-08-12
accessing private fields in TDZ should throw a ReferenceError

joyeecheung:
When accessing a private field that is declared afterwards in a computed property name, it is similar to accessing a variable in TDZ.

cc @caiolima

### DIFF
--- a/test/language/statements/class/elements/private-field-is-visible-in-computed-properties.js
+++ b/test/language/statements/class/elements/private-field-is-visible-in-computed-properties.js
@@ -46,9 +46,10 @@ info: |
 features: [class-fields-private, class]
 ---*/
 
+const self = this;
 assert.throws(ReferenceError, function() {
   class C {
-    [this.#f] = 'Test262';
+    [self.#f] = 'Test262';
     #f = 'foo';
   }
 }, 'access to a not defined private field in object should throw a ReferenceError');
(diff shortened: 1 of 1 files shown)


caiolima:
Nice catch! LGTM.

test/language/statements/class/elements/private-field-is-visible-in-computed-properties.js
```diff
@@ -10,20 +10,46 @@ info: |
...
+    3. If the binding for N in envRec is an uninitialized binding, throw a ReferenceError exception.
+    ...
+
 features: [class-fields-private, class]
 ---*/
 
-assert.throws(TypeError, function() {
+assert.throws(ReferenceError, function() {
   class C {
     [this.#f] = 'Test262';
```
caiolima: The usage of `this` here makes this test wrong into strict mode (see discussion in https://github.com/tc39/test262/pull/2241#issuecomment-520872478). We should use any other base (for example using `self` to properly catch global object) or flag this test as `noStrict`. I prefer the former approach.

joyeecheung: Thanks for catching that, fixed

caiolima:
Nice! LGTM.

Cc @leobalter @spectranaut

PR #2288, opened 2019-08-13
Generation: Support Python 3

Ms2ger:
Like #2277.


tools/generation/lib/test.py
```diff
@@ -19,7 +19,7 @@ def __init__(self, file_name, source=None):
 
     def load(self, prefix = None):
         location = os.path.join(prefix or '', self.file_name)
-        with open(location) as handle:
+        with open(location, 'rb') as handle:
```
adrianheine, line 22: I actually reverted this in #2618, because otherwise `find_comments` wouldn't work.

PR #2290, opened 2019-08-13
Remove double error tests

Snapstromegon:
This PR adapts tests for tc39/ecma262#1645.
Due to changing idempotency for dynamic import on failures these tests are no longer needed.
After tc39/ecma262#1645 idempotency is only required after completing normally.


leobalter:
It's unfortunate we can't really test anything else as counter tests. If I read the proposed text correctly, this behavior remains correct but not required. I'm not sure what should be done in a second call.

PR #2291, opened 2019-08-13
Private static members with multiple evaluation of class

caiolima:
We are adding test to verify that every evaluation of a class generates a new Private Name or Private Brand.

Cc. @spectranaut and @leobalter


test/language/expressions/class/private-static-field-multiple-evaluations-of-class-eval-indirect.js
```diff
@@ -0,0 +1,68 @@
...
+    34. For each item fieldRecord in order from staticFields,
+      a. Perform ? DefineField(F, field).
+    ...
+
+  DefineField(receiver, fieldRecord)
+    ...
+    8. If fieldName is a Private Name,
+      a. Perform ? PrivateFieldAdd(fieldName, receiver, initValue).
+features: [class, class-static-fields-private]
+flags: [noStrict]
```
leobalter: +1

caiolima:
Thank you for the review!

PR #2293, opened 2019-08-13
Adding tests to verify if '.name' of static members are configured properly

caiolima:
Cc @leobalter @spectranaut


caiolima:
Thank you for the review!

Issue #2285, opened 2019-08-12
Add test for `const a = {def\u0061ult: 12}`


weswigham:
Given
```js
const a = {def\u0061ult: 12}
```

In chrome and safari this issues an error, while in firefox and chakra it does not. Some discussion in IRC seemed to lead to believing this is valid and should not error (as a property name is an [`IdentifierName`](https://tc39.es/ecma262/#prod-IdentifierName) which lacks the "no reserved words" restriction of an identifier and therefore should not be bound by the "no escapes in keywords" rule, as it is not a keyword position). There should be a test covering it - [chrome](https://bugs.chromium.org/p/chromium/issues/detail?id=993000) and [safari](https://bugs.webkit.org/show_bug.cgi?id=200638) have bugs open to rectify their behavior.

The extended escape form:
```js
const a = {def\u{0061}ult: 12}
```
is probably also worth a test.

bakkot:
Copying from #2286:

```js
({ def\u0061ult: 0 });
```
and
```js
({ def\u{61}ult: 0 });
```
are both legal, since `def\u0061ult` and `def\u{61}ult` are both [`IdentifierName`](https://tc39.es/ecma262/#prod-IdentifierName)s and therefore [`LiteralPropertyName`](https://tc39.es/ecma262/#prod-LiteralPropertyName)s. The restriction on keywords not containing escape sequences - which prevents `export def\u0061ult function (){}` from being legal - is not relevant in this context.

This case appears not to be covered by current tests - at least, I couldn't find one at a quick glance.

Currently, out of Chakra, JavascriptCore, SpiderMonkey, V8, and XS, only Chakra and SpiderMonkey allow the first form, and only SpiderMonkey allows the second.

bakkot:
Though note that `({ def\u0061ult })` is forbidden, and that should probably be tested too, if it isn't.

rkirsling:
Also note that of the five, it's only V8 that doesn't allow `foo = {}; foo.def\u{61}ult = 3;`.

leobalter:
@weswigham @bakkot @rkirsling I believe I got this covered at #2295 and #2294. Would you mind giving a look there? Thanks!

leobalter:
I also found this is not only with `default` but almost all the reserved words...

```js
var arr = [
    'aw\\u0061it',
    'bre\\u0061k',
    'c\\u0061se',
    'c\\u0061tch',
    'cl\\u0061ss',
    '\\u0063onst',
[...]

PR #2294, opened 2019-08-14
Add cases for `default` identifier names in accessor-names

leobalter:
This is a pre-work for #2285


bakkot:
These looks correct, but I'm not sure if you intended for the property names to have quotes around them:

```js
var C = class {
  get 'def\u0061ult'() { return 'get string'; }
  set 'def\u0061ult'(param) { stringSet = param; }
};
```
is legal (and is what this PR currently adds), but is a less interesting case than

```js
var C = class {
  get def\u0061ult() { return 'get string'; }
  set def\u0061ult(param) { stringSet = param; }
};
```

which is also legal and which I don't think is covered by theses tests.

leobalter:
@bakkot this is partially covered at #2295. I can extend the templates there to get/set accessors there but I wanted to add them here as well (re-using string literals). As I also mentioned, this PR is only a pre-work, while I was trying to locate possible tests using the targeted units.

PR #2295, opened 2019-08-14
Add tests for valid identifier names also listed as reserved words 

### LINKED ISSUE #2285: Add test for `const a = {def\u0061ult: 12}`

leobalter:
Closes #2285


bakkot:
Looks good to me!

One note: because of the difficulty involved in parsing `CoverParenthesizedExpressionAndArrowParameterList` and, separately, object assignment targets, it might be worth adding distinct tests for

```js
 // the parentheses matter because without them it is obvious before reaching the `:` that this cannot be the parameters of an arrow expression
var obj = ({ def\u{61}ult: 42 });
assert(obj.default === 42);
```
and
```js
// the '0,' matters because the code following it is in a context which allows `{` to start an AssignmentExpression without requiring a leading parenthesis, which might be interpreted by a parser as beginning the parameters of an arrow expression
0, { def\u{61}ult: x } = { default: 42 };
assert(x === 42);
```
(which are legal) and
```js
({ def\u{61}ult }) => 42;
```
and
```js
0, { def\u{61}ult } = { default: 42 };
```
(which are not).

But those would just be nice additions. This PR looks correct and covers #2285 without them. Thanks Leo!

leobalter:
@bakkot thanks for the review, the new commits should be addressing your feedback. I'll try to mention you in the parts because this is getting huge.

leobalter:
There are some bugs to catch, I'm fixing them.

leobalter:
Ok, the new patch should be fine!

test/language/expressions/arrow-function/dstr/syntax-error-ident-ref-break-escaped.js
```diff
@@ -0,0 +1,40 @@
...
+
+    Identifier:
+      IdentifierName but not ReservedWord
+
+---*/
+
+
+$DONOTEVALUATE();
+
+var x = ({ bre\u0061k }) => {};
```
leobalter: @bakkot here is a test covering the parameter list of an arrow function.

test/language/expressions/assignment/dstr/ident-name-prop-name-literal-case-escaped.js
```diff
@@ -0,0 +1,42 @@
...
+      ...
+
+    Reserved Words
+
+    A reserved word is an IdentifierName that cannot be used as an Identifier.
+
+---*/
+
+
+var y = { c\u0061se: x } = { case: 42 };
```
leobalter: @bakkot here is an assignment pattern with a property name. I did as a replacement for `0, {} = {}`. I hope this works for you. WDYT?

test/language/expressions/assignment/dstr/syntax-error-ident-ref-default-escaped-ext.js
```diff
@@ -0,0 +1,40 @@
...
+
+    Identifier:
+      IdentifierName but not ReservedWord
+
+---*/
+
+
+$DONOTEVALUATE();
+
+var x = { def\u{61}ult } = { default: 42 };
```
leobalter: and we also capture the error in case it's an identifier reference and not a property name.

test/language/expressions/object/covered-ident-name-prop-name-literal-default-escaped-ext.js
```diff
@@ -0,0 +1,39 @@
...
+    LiteralPropertyName:
+      IdentifierName
+      ...
+
+    Reserved Words
+
+    A reserved word is an IdentifierName that cannot be used as an Identifier.
+---*/
+
+var obj = ({ def\u{61}ult: 42 });
```
leobalter: @bakkot here goes the `({...})`

PR #2299, opened 2019-08-15
Assign "name" property to all anonymous functions

anba:
Implements the changes from <https://github.com/tc39/ecma262/pull/1490>.


leobalter:
That's awesome, thank you!

cc @caiolima if you're working on the tests for class elements this ref PR in ECMA-262 might affect them as well.

PR #2312, opened 2019-08-24
Add Promise#finally invoked on thenable test

### LINKED ISSUE #2302: Add tests for Promise#finally with thenable objects
cc @rkirsling @ljharb

shvaikalesh:
Closes #2302.


test/built-ins/Promise/prototype/finally/this-value-thenable.js
```diff
@@ -0,0 +1,17 @@
...
+features: [Promise.prototype.finally]
+---*/
+
+var thenResult = {};
+var Thenable = function() {};
+Thenable.prototype.then = function() { return thenResult; };
+
+assert.sameValue(
+  Promise.prototype.finally.call(new Thenable()),
+  thenResult
```
ljharb: shouldn’t the result of the then call be wrapped in Promise.resolve, not returned directly?

shvaikalesh: Please see step 7 of [Promise.prototype.finally](https://tc39.es/ecma262/#sec-promise.prototype.finally): it forwards to `then` and returns result right away, just like `Promise.prototype.catch` does.

It is `Promise.prototype.then` that wraps in a Promise, but in this test case we have custom `then`:

```js
Thenable.prototype.then = function() { return thenResult; };
```

ljharb: ah right, in order to support subclasses :-) thanks, my mistake

PR #2316, opened 2019-08-28
Adding tests to verify early error when mixing private static and instance accessors

caiolima:
This is not defined into the spec yet, but it seems to be a bug, according to [https://github.com/tc39/proposal-static-class-features/issues/48](https://github.com/tc39/proposal-static-class-features/issues/48). Maybe this PR needs to wait until this issue is fixed.

Cc @leobalter @littledan


caiolima:
The spec changes blocking this PR was merged. @leobalter, I think we are good to go to merge it.

PR #2317, opened 2019-08-29
Add tests for nested private fields

syg:
This came up with a V8 bug where private fields weren't resolved properly from nested classes where both the inner and the outer class had private fields.

### DIFF
--- a/src/class-elements/private-field-on-nested-class.case
+++ b/src/class-elements/private-field-on-nested-class.case
@@ -1,0 +1,42 @@
+// Copyright (C) 2019 Shu-yu Guo. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: PrivateName CallExpression usage (private field)
+info: |
+  Updated Productions
+
+  CallExpression[Yield, Await]:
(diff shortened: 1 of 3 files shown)


syg:
What am I doing wrong here? I have `features: [class-fields-private]`

leobalter:
There isn't anything wrong, the features are set correctly. Thanks for the tests!

syg:
> There isn't anything wrong, the features are set correctly. Thanks for the tests!

Ah great. I was asking about the bunch of the CI tasks that failed.

caiolima:
@syg Tests are going to fail if VMs don't implement the feature. Tests still run in this case and that's why you have failures into SpiderMonkey, JSC and ChakraCore.

leobalter:
like @caiolima said, we have split CI runs for project lint and build and other runs for executing new tests files on the go.

After the PR is merged, we can't see it anymore, but while it's open, it shows the lint + build runs as the only runs required to pass. There is a "Required" label for these. The other are informative runs are unfortunately don't have the privilege to distinguish failures with other symbols rather than the red cross used for failures.

ljharb:
You can click "view details" to see it after it's merged.

leobalter:
@ljharb I can't see the required label after it's merged.

PR #2320, opened 2019-08-29
Avoid usage of Object.keys when asserting order of returned values

### LINKED ISSUE #2226: do not use compareArray(Object.keys(...
It doesn't guarantee an order of keys.
see `test262/test$ grep -r 'compareArray(Object.keys(' .`

leobalter:
Fixes #2226


leobalter:
Merging based on https://github.com/tc39/test262/issues/2226#issuecomment-526565330

PR #2322, opened 2019-08-30
Added private method length cases

caiolima:
Cc @leobalter @spectranaut

### DIFF
--- a/src/class-elements/fields-anonymous-function-length.case
+++ b/src/class-elements/fields-anonymous-function-length.case
@@ -1,0 +1,42 @@
+// Copyright (C) 2019 Caio Lima (Igalia SL). All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: Anonymous function in field initilizer have length properly set
+info: |
+  Updated Productions
+
+  FieldDefinition :
+    ClassElementName Initializer_opt
+
+  InitializeInstanceFields ( O, constructor )
+    1. Assert: Type ( O ) is Object.
+    2. Assert: constructor is an ECMAScript function object.
+    3. Let fields be the value of constructor.[[Fields]].
+    4. For each item fieldRecord in order from fields,
+      a. Perform ? DefineField(O, fieldRecord).
+    5. Return.
+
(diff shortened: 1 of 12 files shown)


src/class-elements/fields-anonymous-function-length.case
```diff
@@ -0,0 +1,42 @@
...
+  DefineField(receiver, fieldRecord)
+    1. Assert: Type(receiver) is Object.
+    2. Assert: fieldRecord is a Record as created by ClassFieldDefinitionEvaluation.
+    3. Let name be fieldRecord.[[Name]].
+    4. Let initializer be fieldRecord.[[Initializer]].
+    5. If initializer is not empty, then
+      a. Let initValue be ? Call(initializer, receiver).
+    ...
+template: default
+features: [class-static-fields-private, class-static-fields-public]
```
adrianheine, line 29: This should be `class-fields-private, class-fields-public`, right?

test/language/expressions/class/elements/fields-anonymous-function-length.js
```diff
@@ -0,0 +1,46 @@
+// This file was procedurally generated from the following sources:
+// - src/class-elements/fields-anonymous-function-length.case
+// - src/class-elements/default/cls-expr.template
+/*---
+description: Anonymous function in field initilizer have length properly set (field definitions in a class expression)
+esid: prod-FieldDefinition
+features: [class-static-fields-private, class-static-fields-public, class]
```
adrianheine, line 7: See above

test/language/statements/class/elements/fields-anonymous-function-length.js
```diff
@@ -0,0 +1,46 @@
+// This file was procedurally generated from the following sources:
+// - src/class-elements/fields-anonymous-function-length.case
+// - src/class-elements/default/cls-decl.template
+/*---
+description: Anonymous function in field initilizer have length properly set (field definitions in a class declaration)
+esid: prod-FieldDefinition
+features: [class-static-fields-private, class-static-fields-public, class]
```
adrianheine, line 7: Same

Issue #2239, opened 2019-07-15
gc-cleanup-not-prevented-with-wr-deref.js is invalid


kmiller68:
I was looking through the new WeakRef tests and I noticed gc-cleanup-not-prevented-with-wr-deref.js, which I think has two bugs.

1) I think the WeakRef constructor prevents collection until the next event loop turn: https://tc39.es/proposal-weakrefs/#sec-weak-ref-target step 4.

2) There is no guarantee that as GC collect the target, which means it's still valid to return target at the assertion point.

Maybe I'm wrong though?

leobalter:
> 1. I think the WeakRef constructor prevents collection until the next event loop turn: https://tc39.es/proposal-weakrefs/#sec-weak-ref-target step 4.

This is correct. I'd need to make the test async to fix this.

>  2. There is no guarantee that as GC collect the target, which means it's still valid to return target at the assertion point.

This part is super tricky.

If we see Execution:

> At any time, if an object obj is not live, an ECMAScript implementation may perform the following steps atomically:

It is not explicit if the steps must be part of a Garbage Collection or if they are optional during the GC.

Taking the facts we have, Execution is not guaranteed even with a GC hook and this invalidates almost all the tests we have relying on it.

cc @mhofman @gsathya @mathiasbynens.

I'd like to confirm the intention of Execution here, and possibly set the text for it to describe execution must happen in a GC.

leobalter:
one thing that could be done is specifying the `$262.gc` hook should always trigger Execution.

kmiller68:
Even still, the object might not be dead to the VM. Some implementations use a conservative GC, which means if any number on the C stack looks like a pointer to a JS object that object will be kept alive. So even if you require `$262.gc()` trigger Execution, that doesn't mean any WeakRef or FinalizationGroup will be cleared/run, respectively.

kmiller68:
Also, it needs to be more than just async because microtasks don't clear the keepAlive list. You'll probably need a hook into the VM to turn the event loop.

leobalter:
@kmiller68 my example could be like this: 

```js
var deref = false;
var derefAsync = false;

function emptyCells() {
  var wr;
  (function() {
    var a = {};
    wr = new WeakRef(a);
  })();
  $262.gc();
  deref = wr.deref();
  return Promise.resolve().then(() => {
    $262.gc();
    derefAsync = wr.deref();
  });
}

emptyCells().then(() => {
  assert.sameValue(derefAsync, undefined);
[...]

PR #2329, opened 2019-09-04
Add first tentative for an Async GC

leobalter:
The commit 6bc1b35 is only addressing a single test file to illustrate the usage of the new async GC call. I don't have an immediate to the eshost hooks so this is the closest I can get.

Ref #2239

cc @kmiller68 @mhofman @gsathya @ajklein @mathiasbynens, WDYT?

### DIFF
--- a/harness/asyncGC.js
+++ b/harness/asyncGC.js
@@ -6,22 +6,31 @@ description: >
 ---*/
 
 function asyncGc(target) {
-  var wr = new WeakRef(target);
-  return asyncGcDeref(wr).then(() => {
-    if (wr.deref()) {
+  var fg = new FinalizationGroup(() => {});
+  fg.register(target, 'target');
+
+  target = null;
+
+  return Promise.resolve('tick').then(() => asyncGcDeref()).then(() => {
+    var names;
+
+    // consume iterator to capture names
+    fg.cleanupSome(iter => { names = [...iter]; });
+
+    if (!names) {
       throw new Test262Error('Object was not collected');
     }
   });
 }
 
-async function asyncGcDeref(wr) {
+async function asyncGcDeref() {
   var trigger;
-
+  
   // TODO: Remove this when $262.clearKeptObject becomes documented and required
   if ($262.clearKeptObjects) {
     trigger = $262.clearKeptObjects();
   }
-
+  
   await $262.gc();
 
   return Promise.resolve(trigger);
--- a/harness/asyncGC.js
+++ b/harness/asyncGC.js
@@ -7,11 +7,14 @@ description: >
 
 function asyncGc(target) {
   var wr = new WeakRef(target);
-  return asyncGcDeref(wr);
-  
+  return asyncGcDeref(wr).then(() => {
+    if (wr.deref()) {
+      throw new Test262Error('Object was not collected');
+    }
+  });
 }
 
-function asyncGcDeref(wr) {
+async function asyncGcDeref(wr) {
   var trigger;
 
   // TODO: Remove this when $262.clearKeptObject becomes documented and required
@@ -19,13 +22,9 @@ function asyncGcDeref(wr) {
     trigger = $262.clearKeptObjects();
   }
 
-  return Promise.resolve(trigger).then(() => {
-    // TODO: Change this accordingly to the documented sync $262.gc
-    $262.gc();
-    if (wr.deref()) {
-      throw new Test262Error('Object was not collected');
-    }
-  });
+  await $262.gc();
+
+  return Promise.resolve(trigger);
 }
 
 // function emptyCells() {
--- a/harness/asyncGC.js
+++ b/harness/asyncGC.js
@@ -1,0 +1,44 @@
+// Copyright (C) 2019  Ecma International.  All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+description: >
+    Collection of functions used to capture references cleanup from garbage collectors
+---*/
+
+function asyncGc(target) {
+  var wr = new WeakRef(target);
+  return asyncGcDeref(wr);
+  
+}
+
+function asyncGcDeref(wr) {
+  var trigger;
+
+  // TODO: Remove this when $262.clearKeptObject becomes documented and required
+  if ($262.clearKeptObjects) {
+    trigger = $262.clearKeptObjects();
+  }
+
+  return Promise.resolve(trigger).then(() => {
+    // TODO: Change this accordingly to the documented sync $262.gc
+    $262.gc();
+    if (wr.deref()) {
+      throw new Test262Error('Object was not collected');
+    }
+  });
+}
+
+// function emptyCells() {
+//   var target = {};
+//   var wr = new WeakRef(target);
+//   var collected = asyncGc(target, clearKeptObjects());
+//   target = null;
+//   return collected.then(() => {
+//     return wr.deref();
+//   });
+// }
+
+// emptyCells().then((derefAsync) => {
+//   assert.sameValue(derefAsync, undefined);
+// });
+
--- a/test/built-ins/FinalizationGroup/prototype/cleanupSome/cleanup-prevented-with-unregister.js
+++ b/test/built-ins/FinalizationGroup/prototype/cleanupSome/cleanup-prevented-with-unregister.js
@@ -47,8 +47,8 @@ function emptyCells() {
     fg.register(e, 'e', unregDE);
   })();
 
-  var res = fg.unregister(unregC); // unregister 'c' before GC
-  assert.sameValue(res, true, 'unregister c before GC');
+  var res = fg.unregister(unregC);
+  assert.sameValue(res, true, 'unregister c before iterating over it in cleanup');
 
   $262.gc();
 }
@@ -59,7 +59,7 @@ var res = fg.unregister(unregDE);
 assert.sameValue(res, true, 'unregister d and e after GC');
 
 fg.cleanupSome(function cb(iterator) {
-  var res = fb.unregister(unregA);
+  var res = fg.unregister(unregA);
   assert.sameValue(res, true, 'unregister a before the iterator is consumed.');
 
   holdingsList = [...iterator];
@@ -72,13 +72,11 @@ assert.sameValue(holdingsList[0], 'b');
 assert.sameValue(holdingsList.length, 1);
 
 // Second run
-res = fg.unregister(unregB); // let's empty the cells
-assert.sameValue(res, true, 'unregister B for cleanup');
 holdingsList = undefined;
 emptyCells();
 
 fg.cleanupSome(function cb(iterator) {
-  var res = fb.unregister(unregDE);
+  var res = fg.unregister(unregDE);
   assert.sameValue(res, true, 'unregister d and e before the iterator is consumed.');
   holdingsList = [...iterator];
 });
--- a/test/built-ins/FinalizationGroup/prototype/cleanupSome/return-undefined-with-gc.js
+++ b/test/built-ins/FinalizationGroup/prototype/cleanupSome/return-undefined-with-gc.js
@@ -48,9 +48,6 @@ emptyCells();
 assert.sameValue(fg.cleanupSome(() => 1), undefined, 'arrow function');
 
 emptyCells();
-assert.sameValue(fg.cleanupSome(fg.cleanupSome), undefined, 'cleanupSome itself');
-
-emptyCells();
 assert.sameValue(fg.cleanupSome(class {}), undefined, 'class expression');
 
 emptyCells();
--- a/test/built-ins/FinalizationGroup/prototype/cleanupSome/return-undefined-with-gc.js
+++ b/test/built-ins/FinalizationGroup/prototype/cleanupSome/return-undefined-with-gc.js
@@ -14,6 +14,8 @@ info: |
   5. Perform ? CleanupFinalizationGroup(finalizationGroup, callback).
   6. Return undefined.
 features: [FinalizationGroup, arrow-function, async-functions, async-iteration, class, host-gc-required]
+includes: [asyncGC.js]
+flags: [async]
 ---*/
 
 var called;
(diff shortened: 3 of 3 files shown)


mhofman:
I think you can implement `asyncGc` based on `FinalizationGroup` instead of `WeakRef`. It wouldn't depend on `clearKeptObjects`, with all the timing implications.

Sorry I meant to share this code earlier:

```typescript
export interface AsyncGc {
    (target?: object, trigger?: Promise<any>): Promise<boolean>;
}

export function makeAsyncGc(
    gc: () => Promise<any> | void,
    FinalizationGroup: FinalizationGroup.Constructor
): AsyncGc | undefined {
    type Holdings = { collected: boolean };

    function makeFinalizationGroup() {
        try {
            return new FinalizationGroup<Holdings>(holdings => {
                for (const holding of holdings) {
                    holding.collected = true;
                }
            });
        } catch (error) {
            return undefined;
        }
    }

    const finalizationGroup = makeFinalizationGroup();

    return !finalizationGroup
        ? undefined
        : function asyncGc(
              target?: object,
              trigger?: Promise<any>
          ): Promise<boolean> {
              let checker = {};

              const checkerHolding = {
                  collected: false,
              };

              const holding =
                  target != null ? { collected: false } : checkerHolding;
              finalizationGroup.register(target || checker, holding, holding);
              finalizationGroup.register(
                  checker,
                  checkerHolding,
                  checkerHolding
              );

              target = checker = undefined!;

              return Promise.resolve(trigger)
                  .then(() => {
                      // gc() may return a promise that signals asynchronous completion
                      return gc();
                  })
                  .then(() => {
                      finalizationGroup.cleanupSome();

                      finalizationGroup.unregister(checkerHolding);
                      finalizationGroup.unregister(holding);

                      if (!checkerHolding.collected) {
                          return Promise.reject(
                              new Error("Couldn't collect checker")
                          );
                      }

                      return holding.collected;
                  });
          };
}
```

leobalter:
@mhofman the idea is to rely on WeakRef to mitigate requirements and use the most simple piece. There is something bugging me which is the order for calling `$262.gc()` and `$262.clearKeptObjects()`.

I reordered the calls but neither way I can't get a passing state using V8 w/ flags. See f6dfd71

leobalter:
> I think you can implement `asyncGc` based on `FinalizationGroup` instead of `WeakRef`. It wouldn't depend on `clearKeptObjects`, with all the timing implications.
> 
> Sorry I meant to share this code earlier ...

@mhofman I'm confused with this code. It does have a big contrast with the current asyncGC implementation that is more simple and easy to get it. I honestly hope we can stick with the simplicity if possible.

leobalter:
well, I used your suggestion on using FG and got something that "works in V8".

@mhofman 7bc9abb9df WDYT?

mhofman:
Yeah I'm a little uneasy as well with the dependency in the test hooks on features we're testing. But unless the engine exposes an `asyncGc` itself, I don't see a way around it.

My issue with using a regular `WeakRef` to implement `asyncGc` is that it requires to wait until the kept objects have been cleared to trigger the gc, making it impossible to test that WeakRef does keep objects in the constructor and deref.

A `FinalizationGroup` based implementation doesn't have that timing constraint.

mhofman:
A couple thing:
- Why isn't the "trigger" passed as an argument to `asyncGc`? The idea was that gc should be triggered on the next microtask unless the test requires a later provided trigger, e.g. when testing `WeakRef` keeping objects for the current turn.
- The object not being collected is a valid outcome, e.g. when making sure WeakRef holds onto the object. That's why I had `asyncGc` return a boolean. In my implementation, to differentiate between the target not being collected (e.g. observable or kept), and the engine not being capable of collecting, I used a checker object which should get collected. The idea is that `asyncGc` can return an error if it can't collect the target due to implementation limitations, not because the target is still alive.

harness/asyncGC.js
```diff
@@ -0,0 +1,52 @@
+// Copyright (C) 2019  Ecma International.  All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+description: >
+    Collection of functions used to capture references cleanup from garbage collectors
+---*/
+
+function asyncGc(target) {
```
mathiasbynens: the file name is `asyncGC` but the function name is `asyncGc`. Let’s use `asyncGc` for both?

harness/asyncGC.js
```diff
@@ -0,0 +1,52 @@
...
+    if (!names) {
+      throw new Test262Error('Object was not collected');
+    }
+  });
+}
+
+async function asyncGcDeref() {
+  var trigger;
+  
+  // TODO: Remove this when $262.clearKeptObject becomes documented and required
```
mathiasbynens: `clearKeptObject` or `clearKeptObjects`?

leobalter:
answering https://github.com/tc39/test262/pull/2329#issuecomment-528071528:

> Why isn't the "trigger" passed as an argument to asyncGc? The idea was that gc should be triggered on the next microtask unless the test requires a later provided trigger, e.g. when testing WeakRef keeping objects for the current turn.

the only reason so far is that I need to find a use case for it. I'm not fully discarding this option but I'm ok to use it if a test requires. The code is yet in Test262 and the single placement just made it cleaner for now.

> The object not being collected is a valid outcome, e.g. when making sure WeakRef holds onto the object. That's why I had asyncGc return a boolean. In my implementation, to differentiate between the target not being collected (e.g. observable or kept), and the engine not being capable of collecting, I used a checker object which should get collected. The idea is that asyncGc can return an error if it can't collect the target due to implementation limitations, not because the target is still alive.

This is a major problem then. We need to keep Test262 having only deterministic tests. Having two possible exits would create flaky tests, where one could pass and the other might fail in different runs because we can't tell if the GC and refs are cleaned or not.

I had some estimate that within the example given, which I'm basing this current PR, had an assertive failure if the refs couldn't be collected.

If this is not possible I believe we exhausted all the options for Test262. I don't consider Flaky tests to be an option for Test262, and I already discussed this with some people who worked in this project. We have an agreement to continue striving for deterministic tests asserting conformance to the specs, whenever the specs prose can be assertive too.

I'm open to sport more solutions, if they fit, but for now, my plan is to remove the existing tests relying in the `$262.gc()` hook, as they are not compatible to some engines and not a requirement from the specs.

harness/asyncGC.js
leobalter: thanks for the feedback, I agree with it. The only reason I'm not applying is that I'm currently suspending the work on this PR. See https://github.com/tc39/test262/pull/2329#issuecomment-528424274

mhofman:
> This is a major problem then. We need to keep Test262 having only deterministic tests. Having two possible exits would create flaky tests, where one could pass and the other might fail in different runs because we can't tell if the GC and refs are cleaned or not.

Actually that's exactly why I suggested having `asyncGc` distinguish between the 2 cases:
- If the engine cannot determine collection eligibility of the target, `asyncGc` throws, and the test is "skipped". This is not a failure of the test as you point out, collection is not a requirement.
- If the engine can perform collection, but the target is not eligible for collection (e.g. in the kept objects, or simply observable), `asyncGc` returns false. Tests for `new WeakRef()` and `wr.deref()` would probably be testing for that outcome.

I believe that should help keeping test deterministic, as long as the following hooks invariants are respected:
- `ClearKeptObjects` is not performed automatically by the engine between a call to `asyncGc` and the actual collection occurs.
- `asyncGc` returns false only if the engine is capable of collecting the target, but it's not eligible for collection. If the engine cannot determine collection eligibility of the target, it should throw.

Does that make sense?

leobalter:
> asyncGc returns false only if the engine is capable of collecting the target, but it's not eligible for collection. If the engine cannot determine collection eligibility of the target, it should throw.

This is describing a third state for the tests, which we don't have any precedent in Test262.

I consider it be flaky tests. The real use case is that some engines have a non deterministic GC usage. So it means we can't force collection or even estimate how the GC will operate.

A success in tests like these would mean two complete different things.

mhofman:
> A success in tests like these would mean two complete different things.

If there is no third "skipped" state, and marking a non conclusive test as a success is not an option, then yes, there doesn't seem to be a way out.

My feeling is that it should be ok to mark non conclusive collection as a success since collection is not required. That simply makes the collection a best effort, and if it occurs, the behaviors tied to it can be asserted.

ajklein:
Deleting these tests is a step backward for WeakRefs & test262, this doesn't seem like the proper resolution of this discussion.

CC @tschneidereit (co-champion) and @syg (who's also been following along with this stuff)

mhofman:
I'm really bummed tests requiring gc are getting the boot. Most of this spec is only testable after collection has occurred.

Would it make sense to go back to a sync gc approach, and engines that cannot guarantee a complete synchronous collection simply have the tests depending on `$262.gc()` disabled?

leobalter:
Only the current tests requiring a sync GC hook are removed. I did this to stop them showing as "failures" in engines that can't support them at all. 

> engines that cannot guarantee a complete synchronous collection simply have the tests depending on $262.gc() disabled?

That was part of the initial idea, but we got pushback for this approach too.

kmiller68:
I don't think it's acceptable to have tests that explicitly exclude valid (and expected) implementation  strategies. I expect developers use test262 when trying understand the details of how a feature works. Adding tests that rely on a `$262.gc()`only some engines are able support would be misleading and counter productive, especially given how much discussion went into ensuring different GC implementations would be compatible when designing WeakRefs.

While I support the desire to have deterministic tests, I'm not sure how that's possible given this feature was designed with the idea that it would be non-deterministic. It seems like it's more valuable to have tests that cover interesting edge cases in spec. As a developer, I would rather not have to deal with incompatibility because there's a test than have a small number of possibly flakey test262 tests. It's also a little unclear to me how the non-determinism of `$262.asyncGC()` is different than something like a GC bug that collects live objects. Such a bug would cause tests to be flaky as well.

syg:
Removing the tests here is the wrong decision. There is plenty of test surface for WeakRefs that are of the form "if the engine collected this object, then the following must hold". Removing those tests because the **antecedent** sometimes doesn't hold is wrong: we're not testing the antecedent! In other words, we're not testing that the engine collected an object; that property is not simply not deterministically testable anyway.

@leobalter and @rwaldron, there is a difference between "if the precondition for a test isn't always satisfied and the part we want to test might never execute" and a test being flaky. The implementers are saying that these tests aren't flaky: the fact that the test might not execute because the preconditions aren't met is different from a flaky test that, when executed, sometimes spuriously fail.

The WeakRef tests don't have a three state outcome in the same way implications don't have three state outcomes. The tests are of the form "if object is collected, then some predicate must hold". There are two outcomes: fail when the object is collected but the predicate does not hold; pass when the object is not collected *or* the object is collected and the predicate holds.

mhofman:
Well in my opinion there are 3 outcomes:
- fail when the object is collected but the predicate does not hold
- pass when the object is collected and the predicate holds
- did not test because the object is not collected (precondition not met)

Right now it seems "did not test" can only occur statically when a feature is not implemented / the test is disabled. This is why I suggested as a stop-gap to disable the gc dependent tests for engines that cannot guarantee full collection.

The best approach would be to report a "did not test" outcome based on a dynamic precondition of the test (object was not collected). I understand this is not something the test262 framework supports.

The middle ground would be reporting a test pass when the predicate was in fact not tested.
While it feels dishonest, it's not completely wrong in a binary choice world as the test did not actually fail, and the precondition is not a requirement.

Where we could see flaky results though is if an engine sometimes cannot collect an object, but when it can the predicate does not hold. You would only have a flaky result because the test should be failing in the first place. IMO, it's better to report a failure when we can, than to never run the test.

leobalter:
Let me try with some example:

I have a test that asserts a call to FinalizationGroup#cleanupSome calls the given callback function or defer to the instance callback when there isn’t a callback given to the method call.

If an engine cannot collect the references, it’s indeed not an incorrect implementation, what fails is the expected result of my test. If you re read the paragraph above, that’s gonna be unreachable and impossible to tell.

Ideally, a test like this should be skipped. Test262 doesn’t describe artifacts for tests being skipped, this is also optional per implementations of test runners. 

From Test262, the perspective of an execution that can’t satisfy what I’m asserting is a failure. The problem here is the failure comes from my test plan, I can’t test what I can’t determine a path that should always exit that same way.

If I flag my execution as success, for one implementation perspective I’m fine, but the shared results are not honest to the contract established by my plan. In my example, if I call a test successful because it cannot determine when a ref is collected, I’m lying about saying cleanupSome calls the callback function once, because I simply can’t know if that actually happens, not if I force another path. The day it collect, it might continue successful, it might not.

For a test file, all I want to know is if the callback function is called, and in the same example is if I defer to the instance callback function only if a new callback function is not given.

syg:
> If an engine cannot collect the references, it’s indeed not an incorrect implementation, what fails is the expected result of my test. If you re read the paragraph above, that’s gonna be unreachable and impossible to tell.

No, it is not an incorrect implementation. An implementation that never collects anything is a correct implementation.

> If I flag my execution as success, for one implementation perspective I’m fine, but the shared results are not honest to the contract established by my plan. 

I'm puzzled, bare with me. It seems like you're saying "each test262 test must have every line executed by every engine, otherwise it's a useless test and users of test262 wouldn't be able to trust test262 anymore". The reality of WeakRefs is that a WeakRef test won't have every line executed by every engine, but every line executed by one set of engines, and only some lines executed by a different set of engines. The users of test262 (engines!) are explicitly saying that reality is not only useful but desirable.

>  In my example, if I call a test successful because it cannot determine when a ref is collected, I’m lying about saying cleanupSome calls the callback function once, because I simply can’t know if that actually happens, not if I force another path. The day it collect, it might continue successful, it might not.

That's an incorrect understanding of `cleanupSome`. The spec for it does not say `cleanupSome` calls the callback function once. It says it calls the callback function once **in the case of [CheckForEmptyCells](https://tc39.es/proposal-weakrefs/#sec-check-for-empty-cells) returning true**. Whether or not it returns true is **specced** to be nondeterministic. Since both true and false return values for CheckForEmptyCells are compliant, we should test both: some tests only apply to executions that return true, and some tests only apply to executions that return false. If this doesn't fit into the existing infrastructure for test262, test262 needs to be changed so that it does. The conclusion should not be that we simply omit those tests.

leobalter:
I've come with a test plan: https://github.com/tc39/test262/issues/2339

I truly believe this will work for everyone.

PR #2330, opened 2019-09-05
Remove tests based on the $262.gc hook

### LINKED ISSUE #2239: gc-cleanup-not-prevented-with-wr-deref.js is invalid

leobalter:
So far, the tests based on the `$262.gc` hook are proving to be not executable by many engines, as they can't provide a deterministic host defined hook for a GC that actually clear dead references. See #2239.

At the same time, we can't find a solution for deterministic tests using the hooks or not.

The goal of this project is to provide clear tests for what is described in the ECMAScript Suite specs, and the spec text is usually the limit for it. If a spec text says something is implementation defined or dependent, we cannot limit how an implementation - or many implementations - should behave beyond that point.

On a personal note, it's unfortunate we can't have these tests yet, but I hope Test262 continues to a shared base of reliable tests for every JS engine.

Fixes #2239
Closes #2329


rwaldron:
What a shame.

Issue #2263, opened 2019-07-30
Add tests for optional chaining


DanielRosenwasser:
Filing an issue to add tests for https://github.com/tc39/proposal-optional-chaining/

## Statement List

* ### UpdateExpression[Yield, Await]
  * - [ ] OptionalExpression[Yield, Await]
    * - [ ] MemberExpression[Yield, Await] **(member-expression.js)**
      * - [ ] PrimaryExpression[?Yield, ?Await]
        * - [ ] this
        * - [ ] IdentifierReference[?Yield, ?Await]
          * - [ ] Identifier
        * - [ ] Literal
        * - [ ] ArrayLiteral[?Yield, ?Await]
        * - [ ] ObjectLiteral[?Yield, ?Await]
        * - [ ] FunctionExpression
        * - [ ] ClassExpression[?Yield, ?Await]
        * - [ ] GeneratorExpression
        * - [ ] AsyncFunctionExpression
        * - [ ] AsyncGeneratorExpression
        * - [ ] RegularExpressionLiteral
        * - [ ] TemplateLiteral[?Yield, ?Await, ~Tagged]
        * - [ ] CoverParenthesizedExpressionAndArrowParameterList[?Yield, ?Await]
      * - [ ] MemberExpression[?Yield, ?Await][Expression[+In, ?Yield, ?Await]]
      * - [ ] MemberExpression[?Yield, ?Await].IdentifierName
      * - [ ] MemberExpression[?Yield, ?Await]TemplateLiteral[?Yield, ?Await, +Tagged]
      * - [ ] SuperProperty[?Yield, ?Await]
      * - [ ] MetaProperty
      * - [ ] newMemberExpression[?Yield, ?Await]Arguments[?Yield, ?Await]
    * - [ ] CallExpression[Yield, Await] **(call-expression.js)**
      * - [ ] CoverCallExpressionAndAsyncArrowHead[?Yield, ?Await]
      * - [ ] SuperCall[?Yield, ?Await]
      * - [ ] CallExpression[?Yield, ?Await]Arguments[?Yield, ?Await]
      * - [ ] CallExpression[?Yield, ?Await][Expression[+In, ?Yield, ?Await]]
      * - [ ] CallExpression[?Yield, ?Await].IdentifierName
      * - [ ] CallExpression[?Yield, ?Await]TemplateLiteral[?Yield, ?Await, +Tagged]
    * - [ ] OptionalExpression[Yield, Await]
[...]

PR #2337, opened 2019-09-08
test: adding async optional chaining tests

bcoe:
CC: @leobalter, @rkirsling 

see: #2263

### DIFF
--- a/test/language/expressions/optional-chaining/member-expression-async-identifier.js
+++ b/test/language/expressions/optional-chaining/member-expression-async-identifier.js
@@ -18,7 +18,7 @@ function returnSoon(val) {
   });
 }
 const a = undefined;
-const c = {d: returnSoon(11)};
+const c = {d: Promise.resolve(11)};
 async function checkAssertions() {
   assert.sameValue(undefined, await a?.b);
   assert.sameValue(await a?.b, undefined);
--- a/test/language/expressions/optional-chaining/member-expression-async-identifier.js
+++ b/test/language/expressions/optional-chaining/member-expression-async-identifier.js
@@ -1,0 +1,26 @@
+// Copyright 2019 Google, Inc.  All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: prod-OptionalExpression
+description: >
+  optional chain on member expression in async context
+info: |
+  Left-Hand-Side Expressions
+    OptionalExpression
+      MemberExpression [PrimaryExpression identifier] OptionalChain
+features: [optional-chaining]
+flags: [async]
+---*/
+
+function returnSoon(val) {
+  return new Promise(resolve => {
+    setTimeout(() => resolve(val), 5);
+  });
+}
+const a = undefined;
+const c = {d: returnSoon(11)};
+async function checkAssertions() {
+  assert.sameValue(undefined, await a?.b);
+  assert.sameValue(11, await c?.d);
+}
+checkAssertions().then($DONE, $DONE);
--- a/test/language/expressions/optional-chaining/member-expression-async-identifier.js
+++ b/test/language/expressions/optional-chaining/member-expression-async-identifier.js
@@ -21,6 +21,18 @@ const a = undefined;
 const c = {d: returnSoon(11)};
 async function checkAssertions() {
   assert.sameValue(undefined, await a?.b);
-  assert.sameValue(11, await c?.d);
+  assert.sameValue(await a?.b, undefined);
+  assert.sameValue(await c?.d, 11);
+  
+  Promise.prototype.x = 42;
+  var res = await Promise.resolve(undefined)?.x;
+  assert.sameValue(res, 42, 'await unwraps the evaluation of the whole optional chaining expression #1');
+
+  Promise.prototype.y = 43;
+  var res = await Promise.reject(undefined)?.y;
+  assert.sameValue(res, 43, 'await unwraps the evaluation of the whole optional chaining expression #2');
+  
+  c.e = Promise.resolve(39);
+  assert.sameValue(await c?.e, 39, 'await unwraps the promise given after the evaluation of the OCE');
 }
 checkAssertions().then($DONE, $DONE);
--- a/test/language/expressions/optional-chaining/member-expression-async-identifier.js
+++ b/test/language/expressions/optional-chaining/member-expression-async-identifier.js
@@ -12,11 +12,6 @@ features: [optional-chaining]
 flags: [async]
 ---*/
 
-function returnSoon(val) {
-  return new Promise(resolve => {
-    setTimeout(() => resolve(val), 5);
-  });
-}
 const a = undefined;
 const c = {d: Promise.resolve(11)};
 async function checkAssertions() {
--- a/test/language/expressions/optional-chaining/member-expression-async-identifier.js
+++ b/test/language/expressions/optional-chaining/member-expression-async-identifier.js
@@ -20,7 +20,6 @@ function returnSoon(val) {
 const a = undefined;
 const c = {d: Promise.resolve(11)};
 async function checkAssertions() {
-  assert.sameValue(undefined, await a?.b);
   assert.sameValue(await a?.b, undefined);
   assert.sameValue(await c?.d, 11);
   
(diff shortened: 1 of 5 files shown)


test/language/expressions/optional-chaining/member-expression-async-identifier.js
```diff
@@ -0,0 +1,26 @@
...
+  Left-Hand-Side Expressions
+    OptionalExpression
+      MemberExpression [PrimaryExpression identifier] OptionalChain
+features: [optional-chaining]
+flags: [async]
+---*/
+
+function returnSoon(val) {
+  return new Promise(resolve => {
+    setTimeout(() => resolve(val), 5);
```
leobalter: We can't use `setTimeout`, it's not specified in ECMAScript. In any case, I believe you can simplify this code.

test/language/expressions/optional-chaining/member-expression-async-identifier.js
```diff
@@ -0,0 +1,26 @@
...
+  });
+}
+const a = undefined;
+const c = {d: returnSoon(11)};
```
leobalter:
```suggestion
const c = {d: Promise.resolve(11)};
```

leobalter: In this case, you can remove the returnSoon function.

test/language/expressions/optional-chaining/member-expression-async-identifier.js
```diff
@@ -0,0 +1,26 @@
...
+async function checkAssertions() {
+  assert.sameValue(undefined, await a?.b);
+  assert.sameValue(11, await c?.d);
```
leobalter: We usually set the actual value and the expected values in the order opposed to this one presented here. I'd like to suggest two other assertions and so I'll propose it considering the conventional order:

```suggestion
  assert.sameValue(await a?.b, undefined);
  assert.sameValue(await c?.d, 11);
  
  Promise.prototype.x = 42;
  var res = await Promise.resolve(undefined)?.x;
  assert.sameValue(res, 42, 'await unwraps the evaluation of the whole optional chaining expression #1');

  Promise.prototype.y = 43;
  var res = await Promise.reject(undefined)?.y;
  assert.sameValue(res, 43, 'await unwraps the evaluation of the whole optional chaining expression #2');
  
  c.e = Promise.resolve(39);
  assert.sameValue(await c?.e, 39, 'await unwraps the promise given after the evaluation of the OCE');
```

test/language/expressions/optional-chaining/member-expression-async-this.js
```diff
@@ -0,0 +1,25 @@
...
+
+function returnSoon(val) {
+  return new Promise(resolve => {
+    setTimeout(() => resolve(val), 5);
+  });
+}
+async function thisFn() {
+  return await this?.a
+}
+thisFn.call({a: returnSoon(33)}).then(function checkAssertions(arg) {
```
leobalter: same problem with the setTimeout, a Promise.resolve should be enough here. You also don't need the function name.

```suggestion
thisFn.call({a: Promise.resolve(33)}).then(function(arg) {
```

test/language/expressions/optional-chaining/optional-chain-async-optional-chain-square-brackets.js
```diff
@@ -0,0 +1,38 @@
...
+}
+function rejectSoon(error) {
+  return new Promise((resolve, reject) => {
+    setTimeout(() => reject(error), 5);
+  });
+}
+async function checkAssertions() {
+  assert.sameValue(11, await {a: [11]}?.a[0]);
+  const b = {c: [22, 33]};
+  assert.sameValue(33, b?.c[await returnSoon(1)]);
```
leobalter:
```suggestion
  assert.sameValue(33, b?.c[await Promise.resolve(1)]);
```

test/language/expressions/optional-chaining/optional-chain-async-optional-chain-square-brackets.js
```diff
@@ -0,0 +1,38 @@
...
+  function e(val) {
+    return val;
+  }
+  assert.sameValue(55, {d: e}?.d(await returnSoon([44, 55]))[1]);
```
leobalter:
```suggestion
  assert.sameValue(55, {d: e}?.d(await Promise.resolve([44, 55]))[1]);
```

test/language/expressions/optional-chaining/optional-chain-async-optional-chain-square-brackets.js
```diff
@@ -0,0 +1,38 @@
...
+  assert.sameValue(undefined, undefined?.arr[
+    await rejectSoon(Error('unreachable'))
```
leobalter:
```suggestion
    await Promise.reject(new Error('unreachable'))
```

Also adding `new` for the explicit constructor usage

test/language/expressions/optional-chaining/optional-chain-async-square-brackets.js
```diff
@@ -0,0 +1,34 @@
...
+  });
+}
+function rejectSoon(error) {
+  return new Promise((resolve, reject) => {
+    setTimeout(() => reject(error), 5);
+  });
+}
+async function checkAssertions() {
+  assert.sameValue(11, await [11]?.[0]);
+  assert.sameValue(33, [22, 33]?.[await returnSoon(1)]);
```
leobalter:
```suggestion
  assert.sameValue(33, [22, 33]?.[await Promise.resolve(1)]);
```

test/language/expressions/optional-chaining/optional-chain-async-square-brackets.js
```diff
@@ -0,0 +1,34 @@
...
+  assert.sameValue(55, [44, await returnSoon(55)]?.[1]);
```
leobalter:
```suggestion
  assert.sameValue(55, [44, await Promise.resolve(55)]?.[1]);
```

test/language/expressions/optional-chaining/optional-chain-async-square-brackets.js
```diff
@@ -0,0 +1,34 @@
...
+  assert.sameValue(undefined, undefined?.[
+    await rejectSoon(Error('unreachable'))
```
leobalter:
```suggestion
    await Promise.reject(new Error('unreachable'))
```

bcoe:
@leobalter I've updated the tests, I'm not following this one production however:

```js
  Promise.prototype.y = 43;
  var res = await Promise.reject(undefined)?.y;
  assert.sameValue(res, 43, 'await unwraps the evaluation of the whole optional chaining expression #2');
```

I believe this would result in an exception being thrown, not 43 being resolved; can we just drop it?

leobalter:
> @leobalter I've updated the tests, I'm not following this one production however:
> 
> ```js
>   Promise.prototype.y = 43;
>   var res = await Promise.reject(undefined)?.y;
>   assert.sameValue(res, 43, 'await unwraps the evaluation of the whole optional chaining expression #2');
> ```
> 
> I believe this would result in an exception being thrown, not 43 being resolved; can we just drop it?

await does not unwrap the rejected promise, that's the tricky part. The optional chaining will evaluate the `Promise.reject(undefined)` to an instance of Promise, than it gets the `y` property. the `await` expression will never see the rejected promise there.

This might illustrate better the same resolution:

```js
var prom = Promise.reject(undefined);
var res = await (prom?.y);

// or

var evaluation = prom?.y;
var res = await evaluation;
```

The test case is interesting as it verifies the order for runtime evaluation. We can add this in a follow up too.

bcoe:
@leobalter my concern was, at least in Node.js:

```
var res = await Promise.reject(undefined)?.y;
```

bubbles to an unhandled rejection and fails the test.

leobalter:
that doesn't make the test incorrect, right? We can't block tests because a specific host has a non specific behavior, and I believe Node might have some flag to ignore non handled rejections. In this case, Node is even doing the correct evaluation but the warning is triggering a test failure in some test262 runner.

It's important to have these tests, and I believe this is not gonna be the first test here with this rejection swallowing.

bcoe:
@leobalter I trust you 😄, was mainly concerned I was unable to validate in the test runner I had pulled together.

PR #2342, opened 2019-09-11
Add "proto-from-ctor-realm" tests for async functions

shvaikalesh:
Similar tests:

* [Function/proto-from-ctor-realm.js](https://github.com/tc39/test262/blob/master/test/built-ins/Function/proto-from-ctor-realm.js)
* [GeneratorFunction/proto-from-ctor-realm.js](https://github.com/tc39/test262/blob/master/test/built-ins/GeneratorFunction/proto-from-ctor-realm.js)

Related: #2334, #2335, #2336, #2340, #2346.

### DIFF
--- a/test/built-ins/AsyncFunction/proto-from-ctor-realm.js
+++ b/test/built-ins/AsyncFunction/proto-from-ctor-realm.js
@@ -1,0 +1,62 @@
+// Copyright (C) 2019 Alexey Shvayka. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-async-function-constructor-arguments
+description: Default [[Prototype]] value derived from realm of the NewTarget.
+info: |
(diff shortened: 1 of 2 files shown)


shvaikalesh:
Tests are failed with `TypeError`s  because [`assert.sameValue` calls `String`](https://github.com/tc39/test262/blob/master/harness/assert.js#L45) on [`AsyncFunction.prototype`](https://tc39.es/ecma262/#sec-async-function-prototype-properties)s from different realms.

They call inherited [`Function.prototype.toString`](https://tc39.es/ecma262/#sec-function.prototype.tostring) and it fails, because [`AsyncFunction.prototype`](https://tc39.es/ecma262/#sec-async-function-prototype-properties) is not a function (unlike [`Function.prototype`](https://tc39.es/ecma262/#sec-properties-of-the-function-prototype-object)).

Fixed by #2343.

PR #2343, opened 2019-09-11
Gracefully handle errors from ToString in harness/assert

shvaikalesh:
[Motivation](https://github.com/tc39/test262/pull/2342#issuecomment-530604306).

Follow-up of #2342.


leobalter:
I have to say it's a bit overkill but it works. Thanks for adding the tests!

PR #2348, opened 2019-09-15
Remove deprecated $ERROR function

chicoxyzzy:
I've replaced all `$ERROR` calls by `new Test262Error` calls, removed `$ERROR` function and its mention in `CONTRIBUTING.md` and `.jshint`


chicoxyzzy:
I forgot some more files. Will fix later tonight.

chicoxyzzy:
Turns out `$ERROR` is actually could be useful in some cases like `executor(resolve, $ERROR);` or `}).catch($ERROR);`

chicoxyzzy:
Maybe I should just rename `$ERROR` and don't expose for those cases?

leobalter:
> Turns out $ERROR is actually could be useful in some cases like executor(resolve, $ERROR); or }).catch($ERROR);

That seems like cases where we would have async tests and therefore we can simply use `$DONE`,  which throws a Test262Error if called with at least one argument or called more than once. And it's successful if called once with zero arguments.

I'm looking forward for you to apply the other missing cases as I expect at least a couple extra jobs in the CI to pass here.

leobalter:
We've got more updates in the master branch adding more `$ERROR`, please rebase. Thanks!

rwaldron:
My concern is that there may be engine specific test suites that are overriding this function with some custom implementation.

chicoxyzzy:
> My concern is that there may be engine specific test suites that are overriding this function with some custom implementation.

There was 2 tests (with V8 team copyright) in `test/harness/`:
- one testing that `$ERROR` throws `Test262Error`
- and second testing overriding `$ERROR`

I've removed both files.

chicoxyzzy:
Would it be better to split this PR to a number of small PRs replacing `$ERROR` step by step?

chicoxyzzy:
I will split this PR into a bunch of smaller PRs

PR #2349, opened 2019-09-16
Remove expectation of UM 49 to ISO 639 conversion that not exist in the spec

FrankYFTang:
Nowhere in the spec transform 554 to NZ so the locale should stay as en-554 instead of en-NZ.

### DIFF
--- a/test/intl402/Locale/constructor-options-region-valid.js
+++ b/test/intl402/Locale/constructor-options-region-valid.js
@@ -31,8 +31,8 @@ features: [Intl.Locale]
 const validRegionOptions = [
   [undefined, undefined],
   ['FR', 'en-FR'],
-  ['554', 'en-NZ'],
-  [554, 'en-NZ'],
+  ['554', 'en-554'],
+  [554, 'en-554'],
 ];
 for (const [region, expected] of validRegionOptions) {
   let options = { region };


FrankYFTang:
@anba  @peterwmwong @leobalter

FrankYFTang:
@Ms2ger

Ms2ger:
This expectation was introduced in #2097.

leobalter:
In this case, can we wait a bit for feedback from @anba?

anba:
[CanonicalizeLanguageTag](https://tc39.es/ecma402/#sec-canonicalizelanguagetag) refers to [BCP 47 Language Tag Conversion](https://unicode.org/reports/tr35/#BCP_47_Language_Tag_Conversion) which contains:
```
3. If the BCP 47 region subtag matches the type attribute of a territoryAlias element in
   Supplemental Data, replace the language subtag with the replacement value, as follows: 
  a. If there is a single territory in the replacement, use it.
```
And [common/supplemental/supplementalMetadata.xml](https://github.com/unicode-org/cldr/blob/9ab29f8e12433b6b6ffcc4f0cb51b3026344967d/common/supplemental/supplementalMetadata.xml#L988) in turn contains:
```xml
<territoryAlias type="554" replacement="NZ" reason="overlong"/> <!-- New Zealand -->
```

leobalter:
@FrankYFTang PTAL

FrankYFTang:
hum... that make sense. Sorry about this.

PR #2353, opened 2019-09-18
The async GC, with a resolving method to avoid failures

leobalter:


leobalter:
this is ready for review, it only misses adding documentation on the new flag.

rwaldron:
Does this pass in V8 the same way that it did prior to these changes? (I won't have time to do a full local run today)

leobalter:
Interesting question. I don't have a clear answer because I also fixed things in the tests, some of them reported in other issues here.

I've also ran the tests with the current graceful end of resolveAsyncGC and checked the same tests without the graceful end. There are two tests failing with this change ("Test did not run to completion"). Let's see the report:

```
test262-harness -t 32 --hostType=d8 --hostPath=v8 --hostArgs="--harmony-weak-refs"  $(git diff --name-only --diff-filter=d master -- test)
FAIL test/built-ins/FinalizationGroup/FinalizationGroupCleanupIteratorPrototype/next-job-not-active-throws.js (default)
  Test262Error: Iter should fail if not called during the cleanupSome call Expected a TypeError to be thrown but no exception was thrown at all

FAIL test/built-ins/FinalizationGroup/FinalizationGroupCleanupIteratorPrototype/next-job-not-active-throws.js (strict mode)
  Test262Error: Iter should fail if not called during the cleanupSome call Expected a TypeError to be thrown but no exception was thrown at all

FAIL test/built-ins/FinalizationGroup/prototype/cleanupSome/cleanup-prevented-with-reference.js (strict mode)
  Test did not run to completion

FAIL test/built-ins/FinalizationGroup/prototype/cleanupSome/cleanup-prevented-with-reference.js (default)
  Test did not run to completion

FAIL test/built-ins/FinalizationGroup/prototype/cleanupSome/poisoned-callback-throws.js (strict mode)
  Expected no error, got ReferenceError: iteraror is not defined

FAIL test/built-ins/FinalizationGroup/prototype/cleanupSome/poisoned-callback-throws.js (default)
  Expected no error, got ReferenceError: iteraror is not defined

FAIL test/built-ins/WeakRef/prototype/deref/gc-cleanup-not-prevented-with-wr-deref.js (default)
  Test did not run to completion

FAIL test/built-ins/WeakRef/prototype/deref/gc-cleanup-not-prevented-with-wr-deref.js (strict mode)
  Test did not run to completion

Ran 36 tests
28 passed
8 failed
```

So I reviewed and found to minor typos in these files, already fixed in 93b3963, but the error persists for `test/built-ins/WeakRef/prototype/deref/gc-cleanup-not-prevented-with-wr-deref.js`, as the WeakRef instance is the referenced part and is probably preventing the collection of the target object.

harness/async-gc.js
```diff
@@ -0,0 +1,56 @@
...
+  return Promise.resolve(trigger);
+}
+
+function resolveAsyncGC(err) {
+  // if (err === asyncGC.notCollected) {
+  //   // Do not fail as GC can't provide necessary resources.
+  //   $DONE();
+  // }
+
+  $DONE(err);
```
syg: Am I correct in understanding that `resolveAsyncGC` passes along `asyncGC.notCollected` and report executions that didn't collect all the targets as FAILED, and it'll be up to implementations to interpret that as okay because of the `non-deterministic` flag?

leobalter: Sorry, I didn’t mean to commit The commented lines, the intention is to call $DONE without args, which will end with a passing state. Thanks for catching this.

harness/async-gc.js
```diff
@@ -0,0 +1,56 @@
...
+  var trigger;
+
+  // TODO: Remove this when $262.clearKeptObject becomes documented and required
+  if ($262.clearKeptObjects) {
+    trigger = $262.clearKeptObjects();
+  }
+
+  await $262.gc();
+
+  return Promise.resolve(trigger);
```
mhofman: Shouldn't `gc()` be performed after trigger?
```javascript
return Promise.resolve(trigger).then(() => $262.gc());
```

PR #2354, opened 2019-09-18
Add missing feature to prod-private-method-initialize-order.case

nicolo-ribaudo:
Bug found while parsing Test262 tests with Babel


leobalter:
Thank you! I appreciate this PR a lot, regardless it seems such a small fix.

PR #2355, opened 2019-09-20
Test specific error in super-private-access-invalid.case

nicolo-ribaudo:
Without declaring `#x` in the class body, the thrown error could have been about the undeclared private name rather than about the SuperProperty.


leobalter:
good caught, thanks!

PR #2360, opened 2019-09-23
Uncomment lines

leobalter:
Ref #2353

This uncomment lines meant to be in the code for the graceful end of the async GC executions.


leobalter:
cc @syg

PR #2367, opened 2019-09-25
lint: add rule to verify use of harness files

jugglinmike:
Verify that every test file which references a harness file using the
"includes" directive also contains at least one reference to a value
defined in the harness file.

To support this check, extend each harness file with a list of values
which it defines.

I used this check to identify the changes in gh-2366. If you folks want to
include it upstream, then that patch should be merged first.

### DIFF
--- a/tools/lint/lib/checks/includes.py
+++ b/tools/lint/lib/checks/includes.py
@@ -1,0 +1,65 @@
+import os
+import re
+
+from ..check import Check
+from ..frontmatter import parse
+
+class CheckIncludes(Check):
+    '''Ensure tests make use of the harness files that they require via the
+    `includes` directive.'''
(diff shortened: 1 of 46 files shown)


tools/lint/lib/checks/includes.py
```diff
@@ -0,0 +1,65 @@
+import os
```
leobalter: I really like this check. Thanks!

jugglinmike:
> I really like this check. Thanks!

Sure thing! It's a little more complicated than I first expected because we currently have inter-dependencies between harness files. That's new since the last time I contributed to the project. Making each harness file truly independent could improve the contribution experience. If we ever do that, we should simplify this check.

PR #2369, opened 2019-09-25
Changing test to follow spec and throw TypeError instead of ReferenceError

caiolima:
This should be landed only after we resolve https://github.com/tc39/proposal-class-fields/pull/269.

Cc. @leobalter, @spectranaut, @littledan


caiolima:
It is a dup of #2390.

PR #2383, opened 2019-09-30
Add tests for 'numberingSystem' and 'calendar' options

FrankYFTang:


FrankYFTang:
@Ms2ger @rwaldron @leobalter @littledan @zbraniecki @anba

PR #2380, opened 2019-09-27
Add tests for ListFormat StringListFromIterable ( iterable )

FrankYFTang:
Add tests to test the iterable

PR #2387, opened 2019-10-03
Add more tests for ListFormat

### LINKED ISSUE #2386: Need more tests on StringListFromIterable for ListFormat
During the review of https://github.com/tc39/test262/pull/2380 

@Ms2ger  mentioned  
```
LGTM.

I don't think there's any tests yet for

iterable = undefined
GetIterator throws
IteratorStep throws
IteratorValue throws
IteratorClose does anything interesting
Maybe file an issue for those?
```

File this issue to track those.

FrankYFTang:
Fix  #2386
In addition to what I already put into https://github.com/tc39/test262/pull/2380
@Ms2ger @rwaldron @leobalter @littledan @zbraniecki

### DIFF
--- a/test/intl402/ListFormat/prototype/format/iterable-getiterator-throw.js
+++ b/test/intl402/ListFormat/prototype/format/iterable-getiterator-throw.js
@@ -1,0 +1,27 @@
+// Copyright 2019 Google Inc.  All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-Intl.ListFormat.prototype.format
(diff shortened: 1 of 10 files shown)


rkirsling:
These tests have `features: [Intl.ListFormat]` but it doesn't work somehow (at least with WK's test runner)? I wonder if it has to do with the `info` block?

leobalter:
@rkirsling maybe the description with the [double lines](https://github.com/tc39/test262/pull/2387/files#diff-3f5c1e8a2e65ec22bebb8243bc8403f6R6-R7) I sent a patch fixing it recently for other files, I'll repeat the same with these.

PR #2389, opened 2019-10-03
tests for String.prototype.matchAll changes

ljharb:
Per https://github.com/tc39/ecma262/pull/1716

cc @mathiasbynens


ljharb:
Not that I could see.

PR #2390, opened 2019-10-04
Undefined private name access in computed property keys should throw TypeError

joyeecheung:
Refs: https://github.com/tc39/proposal-class-fields/pull/269

cc @caiolima @littledan @syg


leobalter:
> And Inc ethos merge, we can close the other =)

What this means? I'm fine merging one or another, but I need a clear direction which one should prevail here

caiolima:
> > And Inc ethos merge, we can close the other =)
> 
> What this means? I'm fine merging one or another, but I need a clear direction which one should prevail here

I have no idea =O.

I think we should merge this one, because it has a better `info` section.

PR #2391, opened 2019-10-04
Test coverage for `proposal-optional-chaining`

ryaninvents:
Started to write tests for `proposal-optional-chaining`. I've added the required metadata into the file headers, but I'm not sure if I've done it properly so I'd be happy to rework it if someone gives me pointers.

I've tested this by passing the following script as `--transformer`.

<details>
<summary>transform.js</summary>
<pre>
    const babel = require("@babel/core");
    const optionalChaining = require('@babel/plugin-proposal-optional-chaining');
    module.exports = (source) => {
      try {
        const result = babel.transform(source, {
          babelrc: false,
          plugins: [optionalChaining],
          parserOpts: {
            plugins: ['optionalChaining']
          }
        });
        return result.code;
      } catch (error) {
        return `throw new SyntaxError(${JSON.stringify(error.message)});`;
      }
    }
</pre>
</details>

### DIFF
--- a/test/language/punctuators/optional-chaining/computed-property-access.js
+++ b/test/language/punctuators/optional-chaining/computed-property-access.js
@@ -1,0 +1,11 @@
+/*---
+esid: pending
+features: [optional-chaining]
+info: |
+  Access computed properties using optional chaining.
+---*/
+let x = 'my-key';
+let keyAbsent = undefined;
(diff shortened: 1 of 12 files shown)


ryaninvents:
Oops, after looking up the related issue, it looks like someone's already written these tests 😅

Closing my PR.

PR #2395, opened 2019-10-08
Add missing `regexp-match-indices` feature tags

mathiasbynens:
https://chromium-review.googlesource.com/c/v8/v8/+/1845656/7/test/test262/testcfg.py#60

cc @rbuckton


mathiasbynens:
Not sure why the diff is so large. Line endings?

rbuckton:
Definitely line endings. The diff is a lot smaller if you hide whitespace changes:
image

PR #2401, opened 2019-10-09
Use canonical format for the description in the metatags

leobalter:
Ref #2387

cc @rkirsling


rkirsling:
Looks like there's also test/intl402/DateTimeFormat/prototype/formatToParts/related-year.js from #2379.

leobalter:
Thanks for catching it. Seems like my regexp skills are rusting a little bit.

I added this file now. PTAL

PR #2402, opened 2019-10-09
Add tests for Nullish Coalesce Expression

leobalter:
FYI, the only error using V8 with the matching flag is due to the lack of support for TCO in a test file that requires it.

```
test262-harness -t 32 --hostType=d8 --hostPath=v8 --hostArgs='--harmony-nullish' $(git diff --name-only master test/)
FAIL test/language/expressions/coalesce/tco-pos-null.js (strict mode)
  Expected no error, got RangeError: Maximum call stack size exceeded

FAIL test/language/expressions/coalesce/tco-pos-undefined.js (strict mode)
  Expected no error, got RangeError: Maximum call stack size exceeded

Ran 44 tests
42 passed
2 failed
```


rwaldron:
This feature is absurd.

PR #2405, opened 2019-10-17
Add tests for Intl.DisplayNames#resolvedOptions

leobalter:
cc @FrankYFTang

### DIFF
--- a/test/intl402/DisplayNames/prototype/resolvedOptions/return-object.js
+++ b/test/intl402/DisplayNames/prototype/resolvedOptions/return-object.js
@@ -1,0 +1,95 @@
+// Copyright (C) 2019 Leo Balter. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-Intl.DisplayNames.prototype.resolvedOptions
+description: >
+  Returns a new ordinary object on each call, with data properties containing values from internals
+info: |
+  Intl.DisplayNames.prototype.resolvedOptions ()
+
+  1. Let pr be the this value.
+  2. If Type(pr) is not Object or pr does not have an [[InitializedDisplayNames]] internal slot,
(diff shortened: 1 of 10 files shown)


test/intl402/DisplayNames/prototype/resolvedOptions/return-object.js
```diff
@@ -0,0 +1,95 @@
...
+
+assert.sameValue(
+  explicit.hasOwnProperty('localeMatcher'),
+  false,
+  'the localeMatcher option is not set, option was explicitly set'
+);
+
+var extra = new Intl.DisplayNames('en', { chaos: 'yes', random: 'sure', '0': 42 }).resolvedOptions();
+
+assert.compareyArray(
```
FrankYFTang: Extra 'y' before "Array"
It should be
assert.compareArray(
not
assert.compareyArray(

test/intl402/DisplayNames/prototype/resolvedOptions/return-object.js
```diff
@@ -0,0 +1,95 @@
...
+
+verifyProperty(options, 'locale', {
+  value: 'en-US',
+  writable: true,
+  enumerable: true,
+  configurable: true
+});
+
+assert.compareArray(
+  Reflect.ownKeys(options),
```
FrankYFTang: verifyProperty has side effect to change options so you cannot pass options to Refelct.ownKeys after calling verifyProperty. Change to use "Reflect.ownKeys(other)" instead of pass in "options" here will avoid the side effect.

leobalter: oh good catch, thanks!

leobalter:
cc @FrankYFTang sorry it took so long. It should be good to go

FrankYFTang:
LGTM. Merge it !

PR #2408, opened 2019-10-20
Adding case to test invalid syntax when using PrivateName after '?.'

### LINKED ISSUE #2407: Missing test including optional chaining and private identifier
The spec of both proposals don't allow the syntax for `obj?.#privateID`. We need to add test case to cover this invalid syntax.

caiolima:
It closes #2407.


leobalter:
good catch! cc @bcoe

PR #2419, opened 2019-11-11
Fix tests for AggregateError

leobalter:
Ref tc39/proposal-promise-any#49

cc @anba


leobalter:
It's easier to revert this if necessary.

anba:
LGTM! Thanks for the fix.

PR #2425, opened 2019-11-18
feat(optional-chaining): add tests for IterationStatement

bcoe:
Adds tests for IterationStatements, which we were lacking.

Brings up the somewhat interesting case of the production:

```js
const obj = undefined;
for (const key of obj?.a) {
  str += key;
}
```

Which will throw a `TypeError` if `obj` happens to be `undefined` or `null` ... I think this falls into a similar category to:

```
class Foo extends obj?.Base {
// method bodies.
}
```

i.e., ☝️ don't do this.

see:  #2263

### DIFF
--- a/test/language/expressions/optional-chaining/iteration-statement-for-of-type-error.js
+++ b/test/language/expressions/optional-chaining/iteration-statement-for-of-type-error.js
@@ -1,0 +1,18 @@
+// Copyright 2019 Google, LLC.  All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: prod-OptionalExpression
+description: >
+  optional chain returning undefined in RHS of for of statement
+info: |
+  IterationStatement
+    for (LeftHandSideExpression of Expression) Statement
+features: [optional-chaining]
+---*/
+
+assert.throws(TypeError, function() {
+  const obj = undefined;
+  for (const key of obj?.a) {
+    str += key;
+  }
+});
--- a/test/language/expressions/optional-chaining/iteration-statement-for-of-type-error.js
+++ b/test/language/expressions/optional-chaining/iteration-statement-for-of-type-error.js
@@ -11,8 +11,18 @@ features: [optional-chaining]
 ---*/
 
 assert.throws(TypeError, function() {
-  const obj = undefined;
-  for (const key of obj?.a) {
-    str += key;
-  }
+  for (const key of {}?.a) ;
+});
+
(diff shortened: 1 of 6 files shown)


test/language/expressions/optional-chaining/iteration-statement-for-of-type-error.js
```diff
@@ -0,0 +1,18 @@
...
+info: |
+  IterationStatement
+    for (LeftHandSideExpression of Expression) Statement
+features: [optional-chaining]
+---*/
+
+assert.throws(TypeError, function() {
+  const obj = undefined;
+  for (const key of obj?.a) {
+    str += key;
```
leobalter: this line is not necessary.

To avoid any sort of false positives, I'd suggest splitting in 4 simple statements:

```js
  for (const key of {}?.a) ;
});

  for (const key of {}?.a) {}
});

const obj = undefined; // outside of the error catching.
  for (const key of obj?.a) {}
});

  for (const key of obj?.a) ;
});
```

test/language/expressions/optional-chaining/iteration-statement-for.js
```diff
@@ -0,0 +1,27 @@
...
+const obj = {a: true};
+for (count = 0; obj?.a; count++) {
+  if (count > 0) break;
+}
+assert.sameValue(1, count);
+
+// OptionalExpression in init/test/update.
+let count2 = 0;
+const obj2 = undefined;
+for (obj?.a; obj2?.a; obj?.a) {
```
leobalter: nit pick, but might be interesting adding another sample with `for (obj?.a; undefined?.a; obj?.a) { count2++ }`, also, change the third part:

```js
for (obj?.a; obj2?.a; count2++) { count2++; }
assert.sameValue(count2, 0); // personal nit: I usually have "actual, expected"

for (obj?.a; undefined?.a; count2++) { count2++; }
assert.sameValue(count2, 0);
```

this leads to hunting another way to also test optional chaining there:

```js
let touched = 0;
const obj3 = {
  get a() {
    count++;
    return undefined; // explicit for clarity
  }
};
for (count = 0; true; obj3?.a?.[touched++]) {
  if (count > 0) { break; }
}
assert.sameValue(count, 1);
assert.sameValue(touched, 0);
```

bcoe:
@leobalter all you recommendations seem reasonable to me 👌 I noticed I hadn't set an `async` flag, and also fixed this.

Issue #2414, opened 2019-11-07
Optional chain call with super and meta properties


jridgewell:
Babel and TypeScript both ran into issues with `super.method?.()` by trying to use `super` as a regular variable. Can we add tests around this, and `new.target`?

```js
class Base {
    method() {
        return this;
    }
}

class Foo extends Base {
    method() {
        super.method?.();
    }
}

function Bar() {
    return new.target?.();
}
Reflect.construct(Bar, [], function () { return this; })
```

[Babel repl](https://babeljs.io/repl/#?babili=false&browsers=ie%2011&build=&builtIns=usage&spec=false&loose=false&code_lz=MYGwhgzhAEBCkFNoG8BQ0PQLYIC4AsB7AEwAoBKFdTGgJzwFdaA7aAgSwgG5qMBfVANShIMAGKFC0BAA9cCZsRjwISNDRwESFKjRoQGABwS0AdJqLEA_KYo8aAoQDMGzYLnaFW8WjvWZ6XCZWZgQAd1NcMFoAczwbO0FUACUEJxAEd1NgLwhcWgZ3Uh8AGmgAbQBdMpc3Dy9oP2hA4LZ8Ti5oPnIgA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env&prettier=false&targets=&version=7.7.2&externalPlugins=%40babel%2Fplugin-proposal-optional-chaining%407.6.0), [Typescript repl](https://www.typescriptlang.org/play/#code/MYGwhgzhAEBCkFNoG8BQ0PQLYIC4AsB7AEwAoBKFdTGgJzwFdaA7aAgSwgG5qMBfVANShIMAGKFC0BAA9cCZsRjwISNDRwESFKjRoQGABwS0AdJqLEA-KYo8aAoQDMGzYLnaFW8WjvWZ6XCZWZgQAd1NcMFoAczwbO0FUACUEJxAEd1NgLwhcWgZ3Uh8AGmgAbQBdMpc3Dy9oP2hA4LZ8Ti5oPnIgA)

bcoe:
@jridgewell with regards to `super.method?.()`, the expected response would be `this` in your example?

Should ` new.target?.()` throw a `TypeError`?

bcoe:
@jridgewell with regards to `super.method?.()`, the expected response would be `this` in your example?

[...]

PR #2429, opened 2019-11-19
feat: regression tests for optional call

bcoe:
Adds regression tests for #2414, CC: @jridgewell.

fixes: #2414

### DIFF
--- a/test/language/expressions/optional-chaining/new-target-optional-call.js
+++ b/test/language/expressions/optional-chaining/new-target-optional-call.js
@@ -9,14 +9,18 @@ info: |
     MemberExpression OptionalChain
       NewTarget OptionalChain
 features: [optional-chaining]
+includes: [fnGlobalObject.js]
 ---*/
+
+var global = fnGlobalObject();
 
 let called = false;
 function Base() {
   called = true;
+  return this;
 }
 function Foo(blerg) {
-  new.target?.();
+  return new.target?.();
 }
-Reflect.construct(Foo, [], Base);
+assert(Reflect.construct(Foo, [], Base) === global);
 assert.sameValue(called, true);
--- a/test/language/expressions/optional-chaining/new-target-optional-call.js
+++ b/test/language/expressions/optional-chaining/new-target-optional-call.js
@@ -15,12 +15,17 @@ includes: [fnGlobalObject.js]
 const newTargetContext = (function() { return this; })();
 
 let called = false;
+// should be set to 'undefined' or global context, depending on whether
+// mode is strict or sloppy.
+let context = null;
 function Base() {
   called = true;
-  return this;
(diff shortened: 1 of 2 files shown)


test/language/expressions/optional-chaining/new-target-optional-call.js
```diff
@@ -0,0 +1,22 @@
...
+
+let called = false;
+function Base() {
+  called = true;
+}
+function Foo(blerg) {
+  new.target?.();
+}
+Reflect.construct(Foo, [], Base);
+assert.sameValue(called, true);
```
jridgewell: We should also check that the `this` context is correct?

bcoe: @jridgewell which `this` context:

```js
function Base() {
  return this;
}
function Foo(blerg) {
  return new.target();
}
const result = Reflect.construct(Foo, [], Base);
```

☝️ the `this` context of `Base` in `v8` appears to be the `global` object in this example, i.e.,

```
result istanceof Base // is false.
```

```js
function Foo(blerg) {
  return new.target?.();
}
```

behaves similarly.

jridgewell: It should be the global in sloppy mode and `undefined` in strict.

bcoe:
```js
let called = false;
function Base() {
  called = true;
  return this;
}
function Foo(blerg) {
  return new.target?.();
}
assert(Reflect.construct(Foo, [], Base) === this);
assert.sameValue(called, true);
```

jridgewell: Yup. Then for `super.method?.()`, too (which has a much easier `this` context to understand).

leobalter: just adding a quick tip: `var t = (function() { return this; })();` so you reuse the default this value for sloppy and strict mode.

bcoe: @leobalter oh, If I understand, we already run tests both in both `sloppy` and `strict` mode, so I would be able to simply have one test:

```js
const t = (function() { return this; })();
assert(Reflect.construct(Foo, [], Base) === t);
```

leobalter: assuming you want the global object, yes.

For consistency, you can also load a harness file, adding to the yaml meta:

```yaml
includes: [fnGlobalObject.js]
---*/
```

so in your js you can use: `const global = fnGlobalObject();`

bcoe:
@jridgewell @leobalter take a look, sorry it took me a few days to get back to this.

PR #2430, opened 2019-11-20
Decouple checking module tests from having cyclic capabilities

rricard:
Hello, I'm opening this draft PR because I'm currently testing a module loader against test262. This module loader at the moment does not support cyclic dependencies. Unfortunately, many test262 module tests export and import everything in the same file and self-reference themselves. This is practical because it does not require to add fixture files but it does put a hard dependency of having a module loader able to handle cycles even with the intent of testing a different feature of the module loader. I just started working on this, please let me know if the approach is not sound. Thank you!


rricard:
wasn't open as draft PR

PR #2431, opened 2019-11-20
Decouple module tests from having cyclic capabilities

rricard:
Hello, I'm opening this draft PR because I'm currently testing a module loader against test262. This module loader at the moment does not support cyclic dependencies. Unfortunately, many test262 module tests export and import everything in the same file and self-reference themselves. This is practical because it does not require to add fixture files but it does put a hard dependency of having a module loader able to handle cycles even with the intent of testing a different feature of the module loader. I just started working on this, please let me know if the approach is not sound. Thank you!

### DIFF
--- a/test/language/module-code/eval-export-dflt-cls-anon.js
+++ b/test/language/module-code/eval-export-dflt-cls-anon.js
@@ -26,8 +26,7 @@ info: |
 flags: [module]
 ---*/
 
-export default class { valueOf() { return 45; } }
-import C from './eval-export-dflt-cls-anon.js';
+import C from './eval-export-dflt-cls-anon_FIXTURE.js';
 
 assert.sameValue(new C().valueOf(), 45, 'binding initialized');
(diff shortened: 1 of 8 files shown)


leobalter:
For much of these tests, it's intentional to have cyclic dependencies to observe conformance.

The tests are currently correct and the coverage should not be limited to support specific implementations. I suggest filtering these tests away.

rricard:
I understand why you would close that PR, some tests indeed require self-reference and I made the mistake to include them in that rewrite, that being said, there are tests that do not require self-referencing and that if I unroll them in two files, those tests will fail for the reason they're supposed to fail (and not due to the fact they are introducing a cyclic dep, ex: `eval-export-dflt-cls-anon` will rightfully complain about a wrongly named class in our implementation). At the moment I've been ignoring those tests but I wanted to see if I could include them because it gives us information for partial conformance.

rricard:
Note that the PR is in a draft state as I do not think it should be merged now.

PR #2432, opened 2019-11-23
Add test for for-in order

bakkot:
This adds some simple tests for the stage 3 [for-in ordering proposal](https://github.com/tc39/proposal-for-in-order/).

I am happy to make changes as necessary, or you are welcome to push them to this branch yourselves.

### DIFF
--- a/features.txt
+++ b/features.txt
@@ -162,6 +162,10 @@ AggregateError
 # https://github.com/tc39/proposal-string-replaceall
 String.prototype.replaceAll
 
+# Enumeration order for for-in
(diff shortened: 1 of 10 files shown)


bakkot:
The ChakraCore failure is also observable with `Reflect.ownKeys`; see [bug](https://github.com/Microsoft/ChakraCore/issues/4486). This proposal just adds the ordering requirement from `ownKeys` to more cases.

The JavaScriptCore failure corresponds to [this bug](https://bugs.webkit.org/show_bug.cgi?id=38970).

PR #2436, opened 2019-11-27
Add test verifying that optional call preserves `this`.

### LINKED ISSUE #2292: Add tests for Opt. Chain. inside Opt.Chain. computed property names
Based on https://bugs.webkit.org/show_bug.cgi?id=200199#c17

```js
x?.[y?.z];
```

Also, "what happens when `y?.z` fails".

rkirsling:
Resolves #2292.


test/language/expressions/optional-chaining/optional-call-preserves-this.js
```diff
@@ -0,0 +1,27 @@
+// Copyright (C) 2019 Sony Interactive Entertainment Inc.
```
leobalter:
```suggestion
// Copyright (C) 2019 Sony Interactive Entertainment Inc. All rights reserved.
```

this would be a technical fix for the linter

PR #2437, opened 2019-11-27
Add Promise.any tests

### LINKED ISSUE #2410: Add coverage for Promise.any proposal
This issue is to track the progress of the test262 coverage of the stage 3 Promise.any proposal. https://github.com/tc39/proposal-promise-any. ~Tests also should cover `AggregateError`.~ (covered by https://github.com/tc39/test262/pull/2406)

I'm starting to work on this.

chicoxyzzy:
Resolves #2410

Still in-progress

### DIFF
--- a/test/built-ins/Promise/any/reject-immed.js
+++ b/test/built-ins/Promise/any/reject-immed.js
@@ -1,0 +1,15 @@
+// Copyright (C) 2019 Sergey Rubanov. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Promise.any([]) rejects immediately
+esid: sec-promise.any
+flags: [async]
+includes: [promiseHelper.js]
+features: [Promise.any]
+---*/
+
+Promise.any([])
+  .then(function() {
+    $DONE('The promise should not be fulfilled.');
+  }, $DONE);
--- a/features.txt
+++ b/features.txt
@@ -7,6 +7,10 @@
 #
 # https://github.com/tc39/process-document
 
+# Promise.any
+# https://github.com/tc39/proposal-promise-any
+Promise.any
+
 # Missing checks in Proxy internal methods
 # https://github.com/tc39/ecma262/pull/666
 proxy-missing-checks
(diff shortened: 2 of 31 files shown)


chicoxyzzy:
Is there any way to skip CI tests?

test/built-ins/Promise/any/reject-immed.js
```diff
@@ -0,0 +1,15 @@
...
+esid: sec-promise.any
+flags: [async]
+includes: [promiseHelper.js]
+features: [Promise.any]
+---*/
+
+Promise.any([])
+  .then(function() {
+    $DONE('The promise should not be fulfilled.');
+  }, $DONE);
```
spectranaut: If Promise.any([]) should reject immediately, this test will fail, because $DONE will be called with an error instead of `undefined`. You might want to write: `() => $DONE()`, but furthermore, you might want to include an assertion about the error before `$DONE` is called. It would also be nice to include the spec text where this error is being thrown from.

chicoxyzzy:
Thank you for your review @spectranaut! It seems that I won't have time for further changes until at least Tuesday. Feel free to push any changes here :) Thank you again!

spectranaut:
I didn't get to it yesterday @chicoxyzzy, so feel free to add more tests whenever and I'll review them!

Looks like there are more tests that are mostly straightforward copies from Promise.all.

rwaldron:
Just a quick check-in: what's the current status of these tests?

chicoxyzzy:
@rwaldron no progress since my last comment here. I need to address @spectranaut's comments and add a lot more test cases like for Promise.all. Though, I don't know when I'll have time to get back to this :(

rwaldron:
@chicoxyzzy ok, thank you for the update!

rwaldron:
@chicoxyzzy would you be open to letting me pick up where you left off?

chicoxyzzy:
@rwaldron sure! I'll be grateful if you will pick it up!

rwaldron:
New work has moved into https://github.com/tc39/test262/pull/2533

PR #2443, opened 2019-12-03
Adding tests for Array.prototype.copyWithin

natashenka:
A security test for copyWithin. This checks that Array.prototype.copyWithin behaves correctly in the case that converting the start parameter to an integer has side effects. This caused a security bug in Moddable https://github.com/Moddable-OpenSource/moddable/issues/123

### DIFF
--- a/test/built-ins/Array/prototype/copyWithin/coerced-values-start-change-start.js
+++ b/test/built-ins/Array/prototype/copyWithin/coerced-values-start-change-start.js
@@ -1,0 +1,62 @@
+// Copyright (C) 2019 Google. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-array.prototype.copywithin
+description: >
+  SECURITY: start argument is coerced to an integer value
+  and side effects change the length of the array so that
+  the start is out of bounds
+info: |
+  22.1.3.3 Array.prototype.copyWithin (target, start [ , end ] )
+
+  ...
+  8. Let relativeStart be ToInteger(start).
+  ...
+includes: [compareArray.js]
+---*/
+
+
+// make a long integer Array
+function longDenseArray(){
+	var a = [0];
+	for(var i = 0; i < 1024; i++){
+		a[i] = i;
+	}
+	return a;
+}
+
+function shorten(){
+	currArray.length = 20;
+	return 1000;
+}
+
+var array = longDenseArray();
+array.length = 20;
+for(var i = 0; i < 20; i++){
+	array[i] = array[i+1000];
+}
+
+var currArray = longDenseArray();
+
+assert(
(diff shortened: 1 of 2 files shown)


ljharb:
I tried adding these tests to the es6-shim; the "coerced-values-start-change-start" tests appear to fail in latest node (where the implementation is native), as well as with the es6-shim applied in both latest node and 0.10.

Specifically, it's getting an empty array where it expects a dense array of 24 undefineds - perhaps my test framework's assertion is incorrect (maybe it's different than `compareArray`), but can you confirm what the expected values should be for these tests?

natashenka:
Yeah, I think this is a difference in how compare array works. The correct array to return is an empty array of length 20, but when it gets compared, the value is converted to undefined anyhow, but I'm guessing that doesn't happen in your harness. Anyhow, I updated it to compare the empty array, this makes it simpler, and hopefully will work with your harness as well.

test/built-ins/Array/prototype/copyWithin/coerced-values-start-change-start.js
```diff
@@ -0,0 +1,59 @@
...
+array2.length = 20;
+for(var i = 0; i < 24; i++){
+	array2[i] = Object.getPrototypeOf(currArray)[i+1000];
+}
+
+assert(
+  compareArray(
+    currArray.copyWithin(0, {valueOf: shorten}), array2
+  ),
+  'coercion side-effect makes start out of bounds with prototype'
```
ljharb: With the latest changes, i get 1 failure instead of 2 (here).

What's the state of the current implementation in v8 in latest node; is this expected to pass or fail?

natashenka: Can you link to what you're using to run the tests? I'm running these with: https://github.com/bterlson/test262-harness and getting passes on V8, JSC and Spidermonkey downloaded with JSVU today

ljharb: Sure! I'm running `npm run tests-only` and `npm run test:native` inside https://github.com/ljharb/es6-shim/commit/14866a78ee5b1af3f0ef1aac441c6562311be15e (clone/checkout and `npm install` first)

natashenka: I think this is because deep.equal fails if the prototype of the arrays are different, meanwhile compareArray only compares the elements

ljharb: aha, thanks, i can correct the test then. thanks for looking into it!!!

natashenka:
Just following up here. Is there anything I need to do to get this merged?

PR #2444, opened 2019-12-05
TypedArray copyWithin security tests

natashenka:
Adding tests for TypedArray.prototype.copyWithin, as it has had security issues in the past, for example https://bugs.chromium.org/p/project-zero/issues/detail?id=862

All three tests behave inconsistently on the browser JS engines.

test262/test/built-ins/TypedArray/prototype/copyWithin/coerced-values-end-detached.js

Passes on v8, fails on SpiderMonkey and JSC

test262/test/built-ins/TypedArray/prototype/copyWithin/coerced-values-end-detached-prototype.js

Passes on SpiderMonkey and JSC, fails on v8

test262/test/built-ins/TypedArray/prototype/copyWithin/coerced-values-start-detached.js

Passes on v8, fails on SpiderMonkey and JSC

I believe the behavior in the tests is correct according to the specification, but I would appreciate if someone else could check, as the method is fairly complicated.

The root problem here is that when copyWithin is applied to TypeArrays, detachment checks are needed, but the specification does not say where to place them.

### DIFF
--- a/test/built-ins/Array/prototype/copyWithin/coerced-values-start-change-target.js
+++ b/test/built-ins/Array/prototype/copyWithin/coerced-values-start-change-target.js
@@ -1,0 +1,46 @@
+// Copyright (C) 2019 Google. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-array.prototype.copywithin
+description: >
+  SECURITY: start argument is coerced to an integer value
+  and side effects change the length of the array so that
+  the target is out of bounds
+info: |
+  22.1.3.3 Array.prototype.copyWithin (target, start [ , end ] )
+
+  ...
+  8. Let relativeStart be ToInteger(start).
+  ...
+includes: [compareArray.js]
+---*/
+
+
+// make a long integer Array
+function longDenseArray(){
+	var a = [0];
+	for(var i = 0; i < 1024; i++){
+		a[i] = i;
+	}
+	return a;
+}
+
+function shorten(){
+	currArray.length = 20;
+	return 1;
+}
+
+var array = longDenseArray();
+array.length = 20;
+for(var i = 0; i < 19; i++){
+	array[i+1000] = array[i+1];
+}
+
+var currArray = longDenseArray();
+
+assert(
+  compareArray(
+    currArray.copyWithin(1000, {valueOf: shorten}), array
+  ),
+  'coercion side-effect makes target out of bounds'
+);
--- a/test/built-ins/TypedArray/prototype/copyWithin/coerced-values-end-detached-prototype.js
+++ b/test/built-ins/TypedArray/prototype/copyWithin/coerced-values-end-detached-prototype.js
@@ -24,7 +24,7 @@ info: |
   7. If end is undefined, let relativeEnd be len; else let relativeEnd be ?
   ToInteger(end).
   ...
-includes: [compareArray.js, testTypedArray.js, detachArrayBuffer.js]
+includes: [testTypedArray.js, detachArrayBuffer.js]
 features: [TypedArray]
 ---*/
 
--- a/test/built-ins/TypedArray/prototype/copyWithin/coerced-values-end-detached-prototype.js
+++ b/test/built-ins/TypedArray/prototype/copyWithin/coerced-values-end-detached-prototype.js
@@ -7,22 +7,15 @@ description: >
   causing array detachment, but the value is still defined
   by a prototype
 info: |
-  22.2.3.5 %TypedArray%.prototype.copyWithin (target, start [ , end ] )
-
-  %TypedArray%.prototype.copyWithin is a distinct function that implements the
-  same algorithm as Array.prototype.copyWithin as defined in 22.1.3.3 except
-  that the this object's [[ArrayLength]] internal slot is accessed in place of
-  performing a [[Get]] of "length" and the actual copying of values in step 12
-  must be performed in a manner that preserves the bit-level encoding of the
-  source data.
-
+  22.2.3.5%TypedArray%.prototype.copyWithin ( target, start [ , end ] )
   ...
-
-  22.1.3.3 Array.prototype.copyWithin (target, start [ , end ] )
-
-  ...
-  7. If end is undefined, let relativeEnd be len; else let relativeEnd be ?
-  ToInteger(end).
+  8. If end is undefined, let relativeEnd be len; else let relativeEnd be ? ToInteger(end).
+  ... 
+  10. Let count be min(final - from, len - to).
+  11. If count > 0, then
+    a. NOTE: The copying must be performed in a manner that preserves the bit-level encoding of the source data.
+    b. Let buffer be O.[[ViewedArrayBuffer]].
+    c. If IsDetachedBuffer(buffer) is true, throw a TypeError exception. 
   ...
 includes: [testTypedArray.js, detachArrayBuffer.js]
 features: [TypedArray]
@@ -36,7 +29,7 @@ testWithTypedArrayConstructors(function(TA) {
   function detachAndReturnIndex(){
       $DETACHBUFFER(ta.buffer);
       Object.setPrototypeOf(ta, array);
-      return 900;
+      return 101;
   }
 
   array.length = 10000; // big arrays are more likely to cause a crash if they are accessed after they are freed
@@ -44,6 +37,6 @@ testWithTypedArrayConstructors(function(TA) {
   ta = new TA(array);
   assert.throws(TypeError, function(){ 
     ta.copyWithin(0, 100, {valueOf : detachAndReturnIndex}); },
-  "should thow TypeError as array is detached");
+  "should throw TypeError as array is detached");
   
 });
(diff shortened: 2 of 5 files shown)


test/built-ins/TypedArray/prototype/copyWithin/coerced-values-end-detached-prototype.js
```diff
@@ -0,0 +1,49 @@
...
+  must be performed in a manner that preserves the bit-level encoding of the
+  source data.
+
+  ...
+
+  22.1.3.3 Array.prototype.copyWithin (target, start [ , end ] )
+
+  ...
+  7. If end is undefined, let relativeEnd be len; else let relativeEnd be ?
+  ToInteger(end).
```
leobalter: `Array#copyWithin` is never used or reached. in this test

test/built-ins/TypedArray/prototype/copyWithin/coerced-values-end-detached-prototype.js
```diff
@@ -0,0 +1,49 @@
+// Copyright (C) 2019 Google. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-%typedarray%.prototype.copywithin
+description: >
+  SECURITY: end argument is coerced to an integer values
+  causing array detachment, but the value is still defined
+  by a prototype
```
leobalter: this is wrong.

```
%TypedArray%.prototype.copyWithin ( target, start [ , end ] )

...
8. If end is undefined, let relativeEnd be len; else let relativeEnd be ? ToInteger(end).
9. If relativeEnd < 0, let final be max((len + relativeEnd), 0); else let final be min(relativeEnd, len).
10. Let count be min(final - from, len - to).
11. If count > 0, then
    a. NOTE: The copying must be performed in a manner that preserves the bit-level encoding of the source data.
    b. Let buffer be O.[[ViewedArrayBuffer]].
    c. If IsDetachedBuffer(buffer) is true, throw a TypeError exception.
```

Spec wise, there is always a check to validate the `this` value. 

In step 8, the TypedArray instance gets a detached buffer, and its prototype is changed, but the `TypedArray#copyWithin` function continues running. In step 11.c the method will throw the TypeError exception.

test/built-ins/TypedArray/prototype/copyWithin/coerced-values-start-detached.js
```diff
@@ -0,0 +1,43 @@
...
+  function detachAndReturnIndex(){
+      $DETACHBUFFER(ta.buffer);
+      return 100;
+  }
+
+  var array = [];
+  array.length = 10000; // big arrays are more likely to cause a crash if they are accessed after they are freed
+  array.fill(7, 0);
+  ta = new TA(array);
+  ta.copyWithin(0, {valueOf : detachAndReturnIndex}, 1000);
```
leobalter: This line should throw a TypeError

```
%TypedArray%.prototype.copyWithin ( target, start [ , end ] )

...
6. Let relativeStart be ? ToInteger(start).
...
10. Let count be min(final - from, len - to).
11. If count > 0, then
    a. NOTE: The copying must be performed in a manner that preserves the bit-level encoding of the source data.
    b. Let buffer be O.[[ViewedArrayBuffer]].
    c. If IsDetachedBuffer(buffer) is true, throw a TypeError exception.
```

test/built-ins/Array/prototype/copyWithin/coerced-values-start-change-target.js
```diff
@@ -0,0 +1,46 @@
...
+var array = longDenseArray();
+array.length = 20;
+for(var i = 0; i < 19; i++){
+	array[i+1000] = array[i+1];
+}
+
+var currArray = longDenseArray();
+
+assert(
+  compareArray(
```
leobalter: please use `assert.compareArray`, we've been trying to replace the direct usages of the compareArray method.

test/built-ins/TypedArray/prototype/copyWithin/coerced-values-end-detached.js
```diff
@@ -0,0 +1,44 @@
...
+  function detachAndReturnIndex(){
+      $DETACHBUFFER(ta.buffer);
+      return 900;
+  }
+
+  var array = [];
+  array.length = 10000; // big arrays are more likely to cause a crash if they are accessed after they are freed
+  array.fill(7, 0);
+  ta = new TA(array);
+  ta.copyWithin(0, 100, {valueOf : detachAndReturnIndex});
```
leobalter: line should throw a TypeError.

```
%TypedArray%.prototype.copyWithin ( target, start [ , end ] )

...
8. If end is undefined, let relativeEnd be len; else let relativeEnd be ? ToInteger(end).
9. If relativeEnd < 0, let final be max((len + relativeEnd), 0); else let final be min(relativeEnd, len).
10. Let count be min(final - from, len - to).
11. If count > 0, then
    a. NOTE: The copying must be performed in a manner that preserves the bit-level encoding of the source data.
    b. Let buffer be O.[[ViewedArrayBuffer]].
    c. If IsDetachedBuffer(buffer) is true, throw a TypeError exception.
```

leobalter:
> The root problem here is that when copyWithin is applied to TypeArrays, detachment checks are needed, but the specification does not say where to place them.

The spec seems pretty clear on where the checks are.

```
%TypedArray%.prototype.copyWithin ( target, start [ , end ] )

1. Let O be the this value.
2. Perform ? ValidateTypedArray(O).
3. Let len be O.[[ArrayLength]].
4. Let relativeTarget be ? ToInteger(target).
5. If relativeTarget < 0, let to be max((len + relativeTarget), 0); else let to be min(relativeTarget, len).
6. Let relativeStart be ? ToInteger(start).
7. If relativeStart < 0, let from be max((len + relativeStart), 0); else let from be min(relativeStart, len).
8. If end is undefined, let relativeEnd be len; else let relativeEnd be ? ToInteger(end).
9. If relativeEnd < 0, let final be max((len + relativeEnd), 0); else let final be min(relativeEnd, len).
10. Let count be min(final - from, len - to).
11. If count > 0, then
    a. NOTE: The copying must be performed in a manner that preserves the bit-level encoding of the source data.
    b. Let buffer be O.[[ViewedArrayBuffer]].
    c. If IsDetachedBuffer(buffer) is true, throw a TypeError exception.
    ...
```

Step 2: first check for a detached check.
Step 4, 6, and 8: observable and ordered `ToInteger` calls to each of the given arguments that can have a detached buffer, no detached checks between them.
Step 11.c: a hard check for detached buffer for each loop iteration, without leaving the copyWithin method, not avoidable if the buffer was detached in the steps 4, 6, or 8.

The only way to now trigger a typeerror after detachment from those steps is to have `count` that does not meet step 11 condition.

natashenka:
Eek, sorry, I was accidentally working off an older version of the specification: https://www.ecma-international.org/ecma-262/6.0/#sec-%typedarray%.prototype.copywithin . I'll submit an update soon.

test/built-ins/TypedArray/prototype/copyWithin/coerced-values-end-detached-prototype.js
```diff
@@ -0,0 +1,49 @@
...
+---*/
+
+testWithTypedArrayConstructors(function(TA) {
+  
+  var ta;
+  var array = [];
+
+  function detachAndReturnIndex(){
+      $DETACHBUFFER(ta.buffer);
+      Object.setPrototypeOf(ta, array);
```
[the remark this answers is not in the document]
natashenka: It behaves differently if it's implemented wrong, and they don't do the detachment check. In that case, some implementations call Get(), which will also throw a TypeError if the prototype doesn't define the property and the buffer is detached, but will not throw if the prototype does define the property.

test/built-ins/TypedArray/prototype/copyWithin/coerced-values-end-detached-prototype.js
natashenka: Replaced this with a correct description

test/built-ins/TypedArray/prototype/copyWithin/coerced-values-end-detached-prototype.js
natashenka: I've added a better description, though this is still what the test does. It shouldn't be necessary, but this does cause some engines to misbehave despite the specification.

test/built-ins/Array/prototype/copyWithin/coerced-values-start-change-target.js
natashenka: Removed this portion

test/built-ins/TypedArray/prototype/copyWithin/coerced-values-end-detached.js
```diff
@@ -0,0 +1,39 @@
...
+      return 900;
+  }
+
+  var array = [];
+  array.length = 10000; // big arrays are more likely to cause a crash if they are accessed after they are freed
+  array.fill(7, 0);
+  ta = new TA(array);
+  assert.throws(TypeError, function(){ 
+    ta.copyWithin(0, 100, {valueOf : detachAndReturnIndex});
+  "should throw TypeError as array is detached");
```
leobalter: needs to close the function from assert.throws

test/built-ins/TypedArray/prototype/copyWithin/coerced-values-end-detached-prototype.js
```diff
@@ -0,0 +1,42 @@
...
+      return 101;
```
leobalter: Interesting that setting the end being 1 number after start (`100` at line 39) makes it detect an implementation bug on V8.

PR #2445, opened 2019-12-13
Update tests for removal of per-parameter position var scope

syg:
See https://github.com/tc39/ecma262/pull/1046


test/language/function-code/each-param-has-own-scope.js
```diff
@@ -1,26 +0,0 @@
...
-  6. Let paramVarEnv be NewDeclarativeEnvironment(originalEnv).
-  ...
-
-features: [arrow-function, default-parameters]
----*/
-
-var y;
-function f(a = eval("var x = 1; y = 42; x"), b = x) {}
-
-assert.throws(ReferenceError, () => {
```
spectranaut: Do the changes in this PR introduce anything that will now cause a ReferenceError that hasn't before, and should we add tests for those new scenarios? For example, what about:

```
 function f(a = eval("var x = 1"), b =  eval("var x = 2")) {}
```

syg: That example shouldn't throw because it's because `var`s can redeclare other `var`s without error.

There *are* programs that didn't used to throw and now will throw, but I didn't see anything in test262 itself.

Programs that will now throw are ones where a sloppy direct eval introduce a `var` with the same name as one of the parameters. This is because the var scope for per-parameter sloppy eval used to be inside of the parameter name scope, and now it is outside of it. So something like this will now throw:

```javascript
(function f(a = eval("var a = 42")) {})();
```

It's a good thing you brought it up. These should be added to test262, perhaps in a later PR? Is there a way to generate that test for all function forms?

spectranaut: Oh right, of course, it's been so long since I've used vars.

And sure, I can open a different PR generating those tests.

PR #2447, opened 2019-12-17
Add test case for new ReferenceError after removal of per-parameter position var scope

spectranaut:
This test case is in response to the changes in https://github.com/tc39/ecma262/pull/1046 and discussion https://github.com/tc39/test262/pull/2445#discussion_r358789570

### DIFF
--- a/src/function-forms/eval-var-scope-syntax-err.case
+++ b/src/function-forms/eval-var-scope-syntax-err.case
@@ -1,0 +1,20 @@
+// Copyright (C) 2016 Valerie Young. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+desc: sloppy direct eval in params introduces var
+template: error
+info: |
+
+  Runtime Semantics: IteratorBindingInitialization
+  FormalParameter : BindingElement
+
+  1. Return the result of performing IteratorBindingInitialization for BindingElement with arguments iteratorRecord and environment.
+
+
+features: [default-parameters]
+---*/
+
+//- params
+a = eval("var a = 42")
+//- error
+SyntaxError
--- a/test/language/expressions/arrow-function/eval-var-scope-syntax-err.js
+++ b/test/language/expressions/arrow-function/eval-var-scope-syntax-err.js
@@ -1,0 +1,57 @@
+// This file was procedurally generated from the following sources:
+// - src/function-forms/eval-var-scope-syntax-err.case
+// - src/function-forms/error/arrow-function.template
+/*---
+description: sloppy direct eval in params introduces var (arrow function expression)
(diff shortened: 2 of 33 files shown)


spectranaut:
@syg this is how we generate tests for all function forms! Can you take a look at the case file and make sure the info quote and desc are reasonable?

src/function-forms/eval-var-scope-ref-err.case
```diff
@@ -0,0 +1,20 @@
...
+  1. Return the result of performing IteratorBindingInitialization for BindingElement with arguments iteratorRecord and environment.
+
+
+features: [default-parameters]
+---*/
+
+//- params
+a = eval("var a = 42")
+//- error
+ReferenceError
```
syg: Redeclarations are `SyntaxError`s not `ReferenceError`s.

patrick-soquet:
My understanding was that the specification change introduced a new ReferenceError in sloppy mode. But the committed test cases check for SyntaxError in both sloppy and strict modes. Did I miss something? Thank you

ljharb:
See #2460, i think

syg:
> My understanding was that the specification change introduced a new ReferenceError in sloppy mode. But the committed test cases check for SyntaxError in both sloppy and strict modes. Did I miss something? Thank you

They are indeed sloppy only. I wasn't aware of the noStrict flag and didn't know all tests ran both strict/sloppy.

patrick-soquet:
Since class code is strict mode code, it looks like the following test cases are incorrect:
```
language/expressions/class/async-gen-method/eval-var-scope-syntax-err.js
language/expressions/class/async-gen-method-static/eval-var-scope-syntax-err.js
language/expressions/class/gen-method/eval-var-scope-syntax-err.js
language/expressions/class/gen-method-static/eval-var-scope-syntax-err.js
language/expressions/class/method/eval-var-scope-syntax-err.js
language/expressions/class/method-static/eval-var-scope-syntax-err.js
language/statements/class/async-gen-method/eval-var-scope-syntax-err.js
language/statements/class/async-gen-method-static/eval-var-scope-syntax-err.js
language/statements/class/gen-method/eval-var-scope-syntax-err.js
language/statements/class/gen-method-static/eval-var-scope-syntax-err.js
language/statements/class/method/eval-var-scope-syntax-err.js
language/statements/class/method-static/eval-var-scope-syntax-err.js
```
I modified XS to conform to the specification change and to throw a SyntaxError. The new version will be published soon. Thank you.

syg:
@patrick-soquet good catch! @spectranaut is there a way to have the `noStrict` flag passed only to a subset of generated tests? I don't know how to work the python test generator.

PR #2448, opened 2019-12-19
Remove date related types for Intl.DisplayNames

FrankYFTang:
to sync with https://github.com/tc39/proposal-intl-displaynames/pull/63


FrankYFTang:
@rwaldron @leobalter

PR #2457, opened 2019-12-27
Add test case for BigInt in LiteralPropertyName

mysticatea:
Hello.

This PR adds a test case for BigInt literals in `LiteralPropertyName` production.

**Productions:**

[LiteralPropertyName] > [NumericLiteral] > [DecimalBigIntegerLiteral]

**Runtime Semantics:**

> ```
>    LiteralPropertyName: NumericLiteral
>        1. Let _nbr_ be the NumericValue of |NumericLiteral|.
>        1. Return ! ToString(_nbr_).
>```
> https://tc39.es/ecma262/#sec-object-initializer-runtime-semantics-evaluation

[LiteralPropertyName]: https://tc39.es/ecma262/#prod-LiteralPropertyName
[NumericLiteral]: https://tc39.es/ecma262/#prod-NumericLiteral
[DecimalBigIntegerLiteral]: https://tc39.es/ecma262/#prod-DecimalBigIntegerLiteral


mysticatea:
Are there any issues I should address?

test/language/expressions/object/literal-property-name-bigint.js
```diff
@@ -0,0 +1,18 @@
...
+    LiteralPropertyName: NumericLiteral
+        1. Let _nbr_ be the NumericValue of |NumericLiteral|.
+        1. Return ! ToString(_nbr_).
+features: [BigInt]
+---*/
+
+var obj = { 999999999999999999n: true };
+
+assert.sameValue(obj["999999999999999999"], true,
+    "the property name must be the string representation of the numeric value.");
```
caiolima: This test LGTM. I added some other cases we should be covering with BigInt literal as PropertyName in https://github.com/tc39/test262/pull/2486. Could you add those here? This way, I can mark my PR as duplicate.

mysticatea:
I'm sorry for my late response.

@rwaldron Sorry, how can I sign the CLA? Two links in that page are 404.

test/language/expressions/object/literal-property-name-bigint.js
mysticatea: Thank you. I will do.

test/language/expressions/object/literal-property-name-bigint.js
```diff
@@ -0,0 +1,60 @@
+// Copyright (C) 2019 Igalia S.L, Toru Nagashima. All rights reserved.
```
caiolima: nit: s/2019/2020/

test/language/expressions/object/literal-property-name-bigint.js
```diff
@@ -0,0 +1,60 @@
...
+
+let o = { 999999999999999999n: true }; // greater than max safe integer
+
+assert.sameValue(o["999999999999999999"], true,
+    "the property name must be the string representation of the numeric value.");
+
+o = { 1n: "foo" };
+assert.sameValue(o[1n], "foo");
+assert.sameValue(o[1], "foo");
+assert.sameValue(o["1"], "foo");
```
caiolima: I think you can remove this case here, since it's what you already is asserting on lines 31-34.

test/language/expressions/object/literal-property-name-bigint.js
```diff
@@ -0,0 +1,60 @@
...
+
+// MethodDeclaration
+
+o = { 1n() { return "bar"; } };
+assert.sameValue(o[1n](), "bar");
```
caiolima: Also, following your pattern from previous code, I think it is enough to have only 1 assert with string key. I would also recommend add messages on `assert.sameValue`, as you've done. The message here could be `Method name must be the string representation of the numeric value`.

rwaldron:
@mysticatea I linked directly to the CLA in my comment, here it is again: https://tc39.es/test262-cla/. I just did a test signing and the form worked as expected.

mysticatea:
Thank you for your review. I updated this PR.

@rwaldron Thank you. I have sent the form. That was my mistake -- One of my user scripts had affected the page badly.

PR #2459, opened 2020-01-03
Add tests for the numberingSystem option to RelativeTimeFormat.

Ms2ger:


Ms2ger:
https://github.com/tc39/proposal-intl-relative-time/pull/108

PR #2463, opened 2020-01-06
Add tests for Array methods throwing with string receiver.

### LINKED ISSUE #2359: Coverage: Array methods should throw TypeError upon attempting to modify a string
(Discovered in https://bugs.webkit.org/show_bug.cgi?id=201910.)

JSC and XS are both doing the wrong thing at present, but we lack the tests to show it:
```
#### jsc, xs
❌ did not throw: Array.prototype.push.call('')
❌ did not throw: Array.prototype.push.call('', 1)
❌ did not throw: Array.prototype.push.call('abc')
❌ did not throw: Array.prototype.push.call('abc', 1)
❌ did not throw: Array.prototype.pop.call('')
✅         threw: Array.prototype.pop.call('abc')
❌ did not throw: Array.prototype.shift.call('')
❌ did not throw: Array.prototype.shift.call('', 1)
✅         threw: Array.prototype.shift.call('abc')
✅         threw: Array.prototype.shift.call('abc', 1)
❌ did not throw: Array.prototype.unshift.call('')
❌ did not throw: Array.prototype.unshift.call('abc')
```

rkirsling:
Resolves #2359.

### DIFF
--- a/test/built-ins/Array/prototype/pop/throws-with-string-receiver.js
+++ b/test/built-ins/Array/prototype/pop/throws-with-string-receiver.js
@@ -1,0 +1,29 @@
+// Copyright (C) 2020 Sony Interactive Entertainment Inc. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-array.prototype.pop
+description: >
+  Array#pop throws TypeError upon attempting to modify a string
+info: |
+  Array.prototype.pop ( )
+  ...
+  3. If len is zero, then
+    a. Perform ? Set(O, "length", 0, true).
+    ...
+  4. Else,
+    ...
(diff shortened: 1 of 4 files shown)


test/built-ins/Array/prototype/pop/throws-with-string-receiver.js
```diff
@@ -0,0 +1,29 @@
...
+    a. Perform ? Set(O, "length", 0, true).
+    ...
+  4. Else,
+    ...
+    f. Perform ? Set(O, "length", newLen, true).
+
+  Set ( O, P, V, Throw )
+  ...
+  4. Let success be ? O.[[Set]](P, V, O).
+  5. If success is false and Throw is true, throw a TypeError exception.
```
leobalter: nice catch!

PR #2464, opened 2020-01-06
super evaluation order tests

ljharb:
For https://github.com/tc39/ecma262/pull/1775

### DIFF
--- a/test/language/expressions/class/super-evaluation-order.js
+++ b/test/language/expressions/class/super-evaluation-order.js
@@ -1,0 +1,29 @@
+// Copyright (C) 2020 Jordan Harband. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: prod-SuperCall
+description: SuperCall should evaluate Arguments prior to checking IsConstructable
+info: |
+  GetSuperConstructor ( )
+
+  [...]
+  1. Let _func_ be ! GetSuperConstructor().
+  2. Let _argList_ be ? ArgumentListEvaluation of |Arguments|.
+  3. If IsConstructor(_func_) is *false*, throw a *TypeError* exception.
+  [...]
+features: [default-parameters]
+---*/
+
+var x = 0;
+class C extends Object {
+  constructor() {
+    super(x = 123);
+  }
+}
+
+Object.setPrototypeOf(C, parseInt);
+
+assert.throws(TypeError, () => {
+  new C();
+});
+assert.sameValue(x, 123, 'via ArgumentListEvaluation');


test/language/expressions/class/super-evaluation-order.js
```diff
@@ -0,0 +1,26 @@
+// Copyright (C) 2019 Jordan Harband. All rights reserved.
```
leobalter: happy new year!

test/language/expressions/class/super-evaluation-order.js
```diff
@@ -0,0 +1,26 @@
...
+  2. Let _argList_ be ? ArgumentListEvaluation of |Arguments|.
+  3. If IsConstructor(_func_) is *false*, throw a *TypeError* exception.
+  [...]
+features: [default-parameters]
+---*/
+
+var i;
+class A extends {} {
+  constructor() {
+    super(i = 123);
```
leobalter: nice catch!

leobalter: I'm confused ... let me check the specs I'll brb...

bmeck: @leobalter in current spec this test fails, it is for a normative PR from top comment

leobalter:
```
Runtime Semantics: ClassDefinitionEvaluation

ClassTail :ClassHeritage { ClassBody }

5. If ClassHeritageopt is not present, then
  ...
6. Else,
  e. If superclass is null, then
    i. Let protoParent be null.
    ii. Let constructorParent be %Function.prototype%.
  f. Else if IsConstructor(superclass) is false, throw a TypeError exception.
```

the current syntax has a dead end.

I believe you need something like a `class A extends null {` here...

leobalter: > @leobalter in current spec this test fails, it is for a normative PR from top comment

is there any other spec PR allowing a class to extend from a non constructor object like `{}`?

I might really be not up to date and missing it. For the current linked change, the coverage here aims for a post evaluated class definition.

bmeck: looking at your snippet you might be right. changing to `null` should work

ljharb: I've updated to extend from a throwing function, so that if `extends null` is ever fixed, this test won't break.

test/language/expressions/class/super-evaluation-order.js
```diff
@@ -0,0 +1,26 @@
...
+  [...]
+  1. Let _func_ be ! GetSuperConstructor().
+  2. Let _argList_ be ? ArgumentListEvaluation of |Arguments|.
+  3. If IsConstructor(_func_) is *false*, throw a *TypeError* exception.
+  [...]
+features: [default-parameters]
+---*/
+
+var i;
+class A extends function () { throw new TypeError(); } {
```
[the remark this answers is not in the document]
leobalter: @ljharb 

```js
var x = 0;
class C extends Object {
  constructor() {
    super(x = 1);
  }
}

Object.setPrototypeOf(C, parseInt);
```

this will get you good for the assertions.

leobalter:
thanks @ljharb!

PR #2470, opened 2020-01-14
Adding security tests for TypedArray.prototype.fill

natashenka:
Adding security tests for TypedArray.prototype.fill based on CVE-2016-4734. These fail on V8 right now.


natashenka:
Any idea why all the tests are failing? This should pass on Chakra, JSC and SpiderMonkey, but it looks like the test script is hitting other errors

rwaldron:
> Any idea why all the tests are failing? 

The CI tasks for executing tests on the engines are not expected to pass. The only CI tasks that need to pass all the time are "Project lint and generation unit tests on Python 3" and "Project lint, generation tests and build".

PR #2472, opened 2020-01-16
Add test for NumericLiteralSeparator after decimal point followed by DecimalDigit

arai-a:
from https://bugzilla.mozilla.org/show_bug.cgi?id=1609547


rwaldron:
Thanks!

PR #2473, opened 2020-01-16
String.prototype.replace: add `replaceValue` evaluation order test

shvaikalesh:
JSC bug: [Fix broken String.prototype.replace() and replaceAll()](https://bugs.webkit.org/show_bug.cgi?id=204479).

//cc @rkirsling


rwaldron:
This makes me happy. Thanks!

PR #2474, opened 2020-01-16
Security test for TypedArray.prototype.sort based on CVE-2016-7288

natashenka:
Security test for TypedArray.prototype.sort based on CVE-2016-7288. This passes on all the major engines except for Chakra.


natashenka:
Checks for this are also failing, and based on the output, it seems to be due to a different commit. Have the checks broken somehow?

natashenka:
Would it be possible to get this reviewed soon?

rwaldron:
I believe this is a duplicate of: https://github.com/tc39/test262/blob/master/test/built-ins/TypedArray/prototype/sort/sort-tonumber.js

natashenka:
I think this tests a slightly different case. sort-tonumber.js does the detachment before the coercion function meanwhile this test does the detachment in the coercion function. So this test makes sure the detachment check is done after coercion (as opposed to just delaying the exception).

PR #2168, opened 2019-05-17
Improve JSON.stringify coverage

shvaikalesh:
Closes #758.

PR #2475, opened 2020-01-17
Test that JSON.stringify calls replacer on deleted properties

shvaikalesh:
Follow-up of #2168.


test/built-ins/JSON/stringify/replacer-function-object-deleted-property.js
```diff
@@ -0,0 +1,41 @@
...
+
+  [...]
+  12. Return ? SerializeJSONProperty(the empty String, wrapper).
+
+  SerializeJSONProperty ( key, holder )
+
+  1. Let value be ? Get(holder, key).
+  [...]
+  3. If ReplacerFunction is not undefined, then
+    a. Set value to ? Call(ReplacerFunction, holder, « key, value »).
```
rwaldron: The real thing be tested here is whether step 5 & 6 occur prior to the operations nested in step 8.

PR #2479, opened 2020-01-21
Add coverage for astral ID_Start and ID_Continue in identifiers

shvaikalesh:
While the test is correct as-is, it requires good unicode support in [`IdentifierPart`](https://tc39.es/ecma262/#prod-IdentifierPart) to work.

With this change, it would be possible to add surrogate pair support to [`RegExpIdentifierPart`](https://tc39.es/ecma262/#prod-RegExpIdentifierPart) and pass the test before [updating Unicode version in JSC](https://bugs.webkit.org/show_bug.cgi?id=79353).

//cc @mathiasbynens

### DIFF
--- a/test/language/identifiers/other_id_continue-astral.js
+++ b/test/language/identifiers/other_id_continue-astral.js
@@ -3,9 +3,9 @@
 
 /*---
 esid: sec-names-and-keywords
-description: Test grandfathered astral characters of ID_Continue.
+description: Test astral characters of ID_Continue.
 info: |
-  Grandfathered astral characters (Other_ID_Continue)
+  Astral characters (Other_ID_Continue)
 ---*/
 
 // Other_ID_Continue (Unicode 4.1)
--- a/test/language/identifiers/other_id_continue-astral.js
+++ b/test/language/identifiers/other_id_continue-astral.js
@@ -1,0 +1,45 @@
+// Copyright (C) 2020 Alexey Shvayka. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-names-and-keywords
+description: Test grandfathered astral characters of ID_Continue.
+info: |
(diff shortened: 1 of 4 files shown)


mathiasbynens:
1. I don't think this is a "simplification". If anything, the test as currently written is arguably simpler.
2. Instead of exchanging one test with another, we should ideally test both variations.

shvaikalesh:
Seems like [Unicode version in JSC](https://bugs.webkit.org/show_bug.cgi?id=79353) isn't the issue, but rather a [parser bug](https://bugs.webkit.org/show_bug.cgi?id=167940).

I've added astral identifiers tests, yet I believe we should use the simplest syntax possible to avoid unnecessary dependencies for the tests.

test/language/identifiers/other_id_continue-astral.js
```diff
@@ -0,0 +1,45 @@
+// Copyright (C) 2020 Alexey Shvayka. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-names-and-keywords
+description: Test grandfathered astral characters of ID_Continue.
```
mathiasbynens: “grandfathered” is the opposite of what happened here: astral characters were not valid in identifiers prior to ES2015. (Same in the other file.)

rwaldron:
@mathiasbynens can you take one last look?

mathiasbynens:
The changes in `test/language/identifiers/*` LGTM. For the changes in `test/built-ins/RegExp/named-groups/*`, [my earlier comment](https://github.com/tc39/test262/pull/2479#issuecomment-577209848) still stands. Adding tests for `.groups['STRING']` is great, but it should not come at the cost of deleting the tests for `.groups.IDENTIFIER`.

shvaikalesh:
This was superseded by amazing #2532.

Issue #2331, opened 2019-09-05
ECMA-402 PR coverage: Add calendar and numberingSystem options


leobalter:
Ref: https://github.com/tc39/ecma402/pull/175

I've might already have tests for this, see https://github.com/tc39/ecma402/pull/175#issuecomment-494533901

Close it if coverage is confirmed.

caiolima:
We are missing coverage on invalid options throwing `RangeError`.

caiolima:
BTW, I'm going to work on that now.

leobalter:
Closing after #2480

PR #2480, opened 2020-01-23
Added tests with invalid values for calendar and numberingSystem

caiolima:
It is related with #2331.

Cc @leobalter @littledan @FrankYFTang

### DIFF
--- a/test/intl402/DateTimeFormat/invalid-numbering-system-calendar-options.js
+++ b/test/intl402/DateTimeFormat/invalid-numbering-system-calendar-options.js
@@ -1,0 +1,30 @@
+// Copyright 2020 Google Inc, Igalia S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-initializedatetimeformat
+description: >
+    Tests that invalid numberingSystem and calendar option throws RangeError.
+author: Caio Lima
+---*/
+
+let invalidValues = ["ab", "aaaaaaaabbbbbababa."];
+
+var defaultLocale = new Intl.NumberFormat().resolvedOptions().locale;
+
+invalidValues.forEach(function (value) {
+    assert.throws(RangeError, function () {
+            return new Intl.DateTimeFormat([defaultLocale], {numberingSystem: value});
+    }, "Invalid numberingSystem value " + value + " was not rejected.");
+    assert.throws(RangeError, function () {
+            return new Intl.DateTimeFormat([defaultLocale + "-u-nu-" + value]);
+    }, "Invalid numberingSystem value " + value + " was not rejected.");
+
+    assert.throws(RangeError, function () {
+            return new Intl.DateTimeFormat([defaultLocale], {calendar: value});
+    }, "Invalid calendar value " + value + " was not rejected.");
+    assert.throws(RangeError, function () {
+            return new Intl.DateTimeFormat([defaultLocale + "-u-ca-" + value]);
+    }, "Invalid calendar value " + value + " was not rejected.");
+});
+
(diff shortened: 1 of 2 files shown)


iamstolis:
These tests expect `RangeError` from both an invalid `calendar`/`numberingSystem` option and the corresponding Unicode extension. While I see where the `RangeError` comes in the former case (If `_calendar_`/`_numberingSystem_` does not match the Unicode Locale Identifier `type` nonterminal, throw a *RangeError* exception.) I don't see where the specification requires `RangeError` in the latter case (when the invalid value is passed through a Unicode extension). Am I missing something?

caiolima:
> These tests expect `RangeError` from both an invalid `calendar`/`numberingSystem` option and the corresponding Unicode extension. While I see where the `RangeError` comes in the former case (If `_calendar_`/`_numberingSystem_` does not match the Unicode Locale Identifier `type` nonterminal, throw a _RangeError_ exception.) I don't see where the specification requires `RangeError` in the latter case (when the invalid value is passed through a Unicode extension). Am I missing something?

Sorry, I'm not sure if I understand the question. In those tests, both examples throws an `RangeError` because they don't match Unicode Locale Identifier `type` nonterminal. `"ab"` don't match the rule because it is only 2 alphanumeric characters and  `"aaaaaaaabbbbbababa."` contains a `"."`.

iamstolis:
OK, I will be more specific. I don't understand why `new Intl.DateTimeFormat(["en-u-nu-ab"])` should throw `RangeError`. Namely, the `RangeError` cannot come from
> If _numberingSystem_ does not match the Unicode Locale Identifier `type` nonterminal, throw a *RangeError* exception.

step because this step is executed only when the _numberingSystem_ is specified using an option (not using a Unicode extension).

`new Intl.DateTimeFormat(["en-u-nu-aaaaaaaabbbbbababa."])` throws `RangeError` because of ` IsStructurallyValidLanguageTag()` step of `CanonicalizeLocaleList()` i.e. `en-u-nu-aaaaaaaabbbbbababa.` does not represent a well-formed Unicode BCP 47 Locale Identifier as specified in Unicode Technical Standard 35 section 3.2.

Unfortunately, this also cannot be applied to `en-u-nu-ab` as it is well-formed AFAIK. Note that the corresponding grammar allows a key without value/type (`keyword = key (sep type)?`). Hence, `en-u-nu-ab` has a Unicode extension with two keys `nu` and `ab` without values/types IMHO.

FrankYFTang:
> > These tests expect `RangeError` from both an invalid `calendar`/`numberingSystem` option and the corresponding Unicode extension. While I see where the `RangeError` comes in the former case (If `_calendar_`/`_numberingSystem_` does not match the Unicode Locale Identifier `type` nonterminal, throw a _RangeError_ exception.) I don't see where the specification requires `RangeError` in the latter case (when the invalid value is passed through a Unicode extension). Am I missing something?
> 
> Sorry, I'm not sure if I understand the question. In those tests, both examples throws an `RangeError` because they don't match Unicode Locale Identifier `type` nonterminal. `"ab"` don't match the rule because it is only 2 alphanumeric characters and `"aaaaaaaabbbbbababa."` contains a `"."`.

Agree with caiolima!
@iamstolis here is the part of the spec throw RangeError

https://ecma-international.org/ecma-402/#sec-canonicalizelocalelist
9.2.1 CanonicalizeLocaleList ( locales )
step 7.c.iv If IsStructurallyValidLanguageTag(tag) is false, throw a RangeError exception.

Notice 
https://ecma-international.org/ecma-402/#sec-initializedatetimeformat
12.1.1 InitializeDateTimeFormat ( dateTimeFormat, locales, options )
step 1 is Let requestedLocales be ? CanonicalizeLocaleList(locales).

FrankYFTang:
ok, @iamstolis  does has a good point about the en-u-nu-ab has a Unicode extension with two keys nu and ab without values/types.

PR #2486, opened 2020-01-28
Adding test to cover BigInt literal as PropertyName.

### LINKED ISSUE #2485: Missing tests for BigInt literals as ProeprtyName
See related bug on [SpiderMonkey](https://bugzilla.mozilla.org/show_bug.cgi?id=1605835) and [JSC](https://bugs.webkit.org/show_bug.cgi?id=206888).

BigInt literals can be used as property name. In this case `{1n: "foo"}` should evaluate to an object with `{"1": "foo"}`.

caiolima:
It closes #2485.

Cc @littledan @spectranaut @leobalter

### DIFF
--- a/test/language/literals/bigint/use-as-property-name.js
+++ b/test/language/literals/bigint/use-as-property-name.js
@@ -1,0 +1,50 @@
+// Copyright (C) 2020 Igalia S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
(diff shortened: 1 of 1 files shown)


rwaldron:
This duplicates https://github.com/tc39/test262/pull/2457, however it covers more surface than the predecessor. Any thoughts on how I might resolve this?

caiolima:
Closing because it is being handled on #2457

PR #2498, opened 2020-02-11
Test \b escape inside CharacterClass in Unicode RegExp

shvaikalesh:


rwaldron:
@mathiasbynens can you take a quick look at this?

test/built-ins/RegExp/unicode_character_class_backspace_escape.js
```diff
@@ -0,0 +1,20 @@
...
+    b
+
+  Static Semantics: CharacterValue
+
+  ClassEscape :: b
+
+  1. Return the code point value of U+0008 (BACKSPACE).
+---*/
+
+assert(/[\b]/u.test('\b'));
```
mathiasbynens: You could imagine a broken engine that treats `\b` as `b` both in string literals and in `u` RegExps. Such an engine would still pass this test. How about:

```js
const backspace = '\b';
assert(backspace.charCodeAt() === 0x08);
assert(/[\b]/u.test(backspace));
```

shvaikalesh:
> LGTM with a minor suggestion

@rwaldron the suggestion was accounted for.

PR #2505, opened 2020-02-18
Fix invalid YAML from #2470 and #2474.

rkirsling:
The tests added in PRs #2470 and #2474 have invalid YAML (according to perl5/YAML/Loader.pm), which prevents JSC from importing the latest changes.


kmiller68:
LGTM

PR #2509, opened 2020-02-24
Test Array methods with "length" of MAX_SAFE_INTEGER

shvaikalesh:
JSC implements `Array.prototype.{copyWithin,fill,reduceRight}` methods in JavaScript.


rwaldron:
@phoddie @kmiller68 calling you both over to take a look at the failures here, which are coming from XS and JSC—I figured you'd want to be made aware of these.

shvaikalesh:
Regarding JSC, a fix would be to change a few types from `unsinged`. I will submit a patch as soon as this PR is merged and synced to WebKit's trunk.

Issue #2515, opened 2020-02-27
Resolve test validity claim for built-ins/AggregateError/newtarget-proto-fallback.js


rwaldron:
See: https://github.com/Moddable-OpenSource/moddable/blob/public/documentation/xs/XS%20Conformance.md#built-ins-1

Re: https://github.com/tc39/test262/blob/7040938bd06ded171250d1e945a969f90fb37165/test/built-ins/AggregateError/newtarget-proto-fallback.js

anba:
That claim is correct, we're also skipping that test in Firefox, because it expects the wrong prototype.

rwaldron:
Thanks @anba

PR #2518, opened 2020-03-05
Fix: AggregateError/newtarget-proto-fallback. Closes gh-2515

rwaldron:
@anba ptal?


anba:
LGTM! Thanks for fixing this bug.

PR #2461, opened 2020-01-04
Add test verifying offsetless date-time string is local time.

rkirsling:
Resolves #2450.

PR #2523, opened 2020-03-08
Fix DST sensitivity in Date/parse/without-utc-offset.js.

rkirsling:
Addresses https://github.com/tc39/test262/pull/2461#issuecomment-596199997.

(CC @anba)


anba:
Thanks! :smile:

PR #2525, opened 2020-03-10
Improve an error message in assert.throws().

Ms2ger:
A message like "Expected a TypeError but got a RangeError" is not
necessarily helpful to find the actual bug. Adding the message will
usually help more.

### DIFF
--- a/harness/assert.js
+++ b/harness/assert.js
@@ -84,7 +84,7 @@ assert.throws = function (expectedErrorConstructor, func, message) {
       message += 'Thrown value was not an object!';
       $ERROR(message);
     } else if (thrown.constructor !== expectedErrorConstructor) {
-      message += 'Expected a ' + expectedErrorConstructor.name + ' but got a ' + thrown.constructor.name;
+      message += 'Expected a ' + expectedErrorConstructor.name + ' but got a ' + thrown.constructor.name + ' (' + thrown.message + ')';
       $ERROR(message);
     }
     return;


harness/assert.js
```diff
@@ -84,7 +84,7 @@ assert.throws = function (expectedErrorConstructor, func, message) {
       message += 'Thrown value was not an object!';
       $ERROR(message);
     } else if (thrown.constructor !== expectedErrorConstructor) {
-      message += 'Expected a ' + expectedErrorConstructor.name + ' but got a ' + thrown.constructor.name;
+      message += 'Expected a ' + expectedErrorConstructor.name + ' but got a ' + thrown.constructor.name + ' (' + thrown.message + ')';
```
rwaldron: This whole consequent body only concerns the output when there is a constructor mismatch. If the error constructor is not what you wanted, the message is irrelevant.

leobalter: In fact, this has been discussed in the past and the current message is intentionally ignoring the error message.

assert.throws prioritizes a given custom message to the assertion. This is preferred usage and a good practice we should enforce.

Ideally, each test should distinguish assertions, even if it's minimal, e.g.:

```js
assert.throws(Test262Error, fn, "fn throws Test262Error the first time it's called");
assert.throws(TypeError, fn, "fn throws TypeError the second time it's called");
```

If Test262 relies on error messages, it would be expecting implementations to provide something more than the specification does.

leobalter:
While the changes are technically correct, I prefer we should instead be enforcing a good practice of assertion messages.

Thanks for understanding!

Ms2ger:
I'm not sure I follow the argument. I was using the third argument, which is how I knew which `assert.throws` call failed. What I had trouble figuring out while debugging that was the provenance of the incorrect exception. The type of the exception alone wasn't enough to pinpoint it, but seeing the message immediately pointed me to the bug I needed to find, so it was very much not irrelevant to me. Before this change, I got to see the message by moving the code out of the `assert.throws` call, instead invoking the throwing code directly; in that case, the failure message was something like

> Expected no error, got [error constructor]: [error message]

Should we remove the error message there as well?

leobalter:
By your question on removing an error message I’m generalizing to the matters if we should improve and clean up Test262 code.

My answer is positive. We need to spend some time improving the code quality for a better consistency. Unfortunately, this requires some more dedicated and sponsored time we could use to prioritize this work.

Ms2ger:
I don't understand how any of the comments in this PR apply to the proposed change. Could you clarify the harm you see?

Ms2ger:
@rwaldron @leobalter ping

shvaikalesh:
Proposed change will definitely improve debugging experience.

Test case:

```js
assert.throws(Test262Error, fn,
  "fn throws Test262Error the first time it's called.");
```

Output:

```diff
   fn throws Test262Error the first time it's called. Expected a Test262Error
-  but got a TypeError
+  but got a TypeError (x is not a function)
```

`assert` messages provided by test authors are not overridden.

Ms2ger:
@rwaldron @leobalter ping

rwaldron:
The error message is meaningless noise if the error isn't even the correct kind of error—that's what this path checks.

rwaldron:
Another issue to consider is there are unknown consumers of test262 output that rely on the output not changing, for the purpose of validation.

Ms2ger:
It's not meaningless, it's very helpful in figuring out where the incorrect exception comes from. I also don't buy your "unknown consumers" argument.

PR #2526, opened 2020-03-11
Update RegExp property escape tests per Unicode v13.0.0

mathiasbynens:
Upstream change in test generator:
https://github.com/mathiasbynens/unicode-property-escapes-tests/commit/11fa92ca644f1b6c0f3be40f32533ff0ec64e86c

Unicode v13.0.0:
https://unicode.org/versions/Unicode13.0.0/

Ref. https://github.com/tc39/ecma262/issues/1897 & https://github.com/tc39/ecma262/pull/1896.


leobalter:
BTW, good to keep track of Unicode 13 adoption here: https://test262.report/features/regexp-unicode-property-escapes/built-ins/RegExp/property-escapes

jirkamarsik:
I see that this new update includes aliases for the emoji properties (`EComp` for `Emoji_Component`, `EMod` for `Emoji_Modifier`, `EBase` for `Emoji_Modifier_Base`, `EPres` for `Emoji_Presentation` and `ExtPict` for `Extended_Pictographic`). The tests now check that implementations support these short-hand aliases. However, the spec currently forbids such aliases (see [Table 56](https://tc39.es/ecma262/#table-binary-unicode-properties)).

Is it the plan to update the spec to include these aliases or were they added here accidentally?

mathiasbynens:
@jirkamarsik Good catch! These aliases haven't yet been added to the spec. Patch: https://github.com/tc39/ecma262/pull/1939

PR #2534, opened 2020-03-16
Adding test coverage for 'OptionalChain'.PrivateIdentifier case

caiolima:
Test to add coverage to changes proposed at https://github.com/tc39/proposal-class-fields/pull/301. This should only merge after we get consensus on merging changes on that PR.


caiolima:
@leobalter Do you think it is ok to merge it before consensus from committee?

PR #2535, opened 2020-03-17
Add coverage for EvaluateNew

### LINKED ISSUE #2527: Add test for EvaluateNew execution order
I couldn't find a test that ensures that [EvaluateNew](https://tc39.es/ecma262/#sec-evaluatenew) steps 3 and 4 come before step 6: the reference to the constructor is evaluated before the arguments.

specifically, a test case like this:

```
(function (){var a=function(){};new a((a=1))})()
```

It should (I believe) construct the function with `1` as an arg, not try to construct `1` and `TypeError`.

I think this is relevant because this currently fails in JSC (macOS 10.15.3):

```
% cat test.js
(function (){var a=function(){};new a((a=1))})()
% /System/Library/Frameworks/JavaScriptCore.framework/Versions/A/Resources/jsc test.js
Exception: TypeError: 1 is not a constructor (evaluating 'new a((a=1))')
test.js:1:38
global code@test.js:1:47
```

leobalter:
Fix #2527

### DIFF
--- a/test/language/expressions/new/ctorExpr-isCtor-after-args-eval-fn-wrapup.js
+++ b/test/language/expressions/new/ctorExpr-isCtor-after-args-eval-fn-wrapup.js
@@ -1,0 +1,36 @@
+// Copyright (C) 2020 Leo Balter. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-new-operator
+description: >
+  The IsConstructor(ctor) happens after evaluating the arguments, use the correct ctor.
+  Function wrap-up to use the same function level binding ref
+info: |
+  NewExpression : new NewExpression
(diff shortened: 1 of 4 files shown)


leobalter:
cc @mroch @rkirsling

test/language/expressions/new/ctorExpr-isCtor-after-args-eval-fn-wrapup.js
```diff
@@ -0,0 +1,36 @@
...
+  8. Return ? Construct(constructor, argList). 
+---*/
+
+var ref;
+var argz;
+
+assert.throws(TypeError, function() {
+  var x = 42;
+  ref = x;
+  new x(x = function() {}, argz = 39);
```
leobalter: I'm trying to test the opposite direction of the one reported in #2527 here. And JSC is failing to throw the TypeError as well.

PR #2548, opened 2020-03-28
Test ProxyCreate with revoked target and handler

shvaikalesh:
Ref. https://github.com/tc39/ecma262/pull/1814


rwaldron:
@bathos can you take a look?

bathos:
@rwaldron I’m out of the loop re: any further discussion which might have taken place regarding this change, but this does LGTM.

I could see value in also testing for `typeof revoked === 'function'`, where the revoked proxy’s target was a function.

PR #2550, opened 2020-03-28
Adding case where 'static' is used as a field name

### LINKED ISSUE #2549: Missing tests with "static" valid public fields
There is a [bug](https://bugs.webkit.org/show_bug.cgi?id=209703) in JSC public class fields implementation that don't allow us create a field named as `static`, but JSC is passing all tests with `class-fields-public`. We should add cases to verify it.

caiolima:
It closes #2549.

### DIFF
--- a/src/class-elements/static-as-valid-instance-field.case
+++ b/src/class-elements/static-as-valid-instance-field.case
@@ -1,0 +1,32 @@
+// Copyright (C) 2020 Igalia S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: static is a valid name of an instance field
+info: |
+  ClassElement:
+    ...
+    FieldDefinition ;
+
+  FieldDefinition:
+    ClassElementName Initializer_opt
+
+  ClassElementName:
+    PropertyName
+template: default
+includes: [propertyHelper.js]
+features: [class-fields-public]
+---*/
+
+//- elements
+static = "foo";
+//- assertions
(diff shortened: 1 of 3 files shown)


src/class-elements/static-as-valid-instance-field.case
```diff
@@ -0,0 +1,32 @@
+// Copyright (C) 2020 Igalia S.L. All rights reserved.
```
ljharb: what is a .case file for?

test/language/expressions/class/elements/static-as-valid-instance-field.js
```diff
@@ -0,0 +1,36 @@
...
+      ClassElementName Initializer_opt
+
+    ClassElementName:
+      PropertyName
+
+---*/
+
+
+var C = class {
+  static = "foo";
```
ljharb: should this also verify a `static static = ` field?

src/class-elements/static-as-valid-instance-field.case
caiolima: It will create a version of this case for `ClassExpression` and another for `ClassDeclaration`.

test/language/expressions/class/elements/static-as-valid-instance-field.js
caiolima: Nice catch. I think we should do this in a separate file/PR, since static fields uses a different flag than `class-fields-public`

PR #2551, opened 2020-03-29
Test that async iterator methods don't pass absent values

shvaikalesh:
Ref. https://github.com/tc39/ecma262/pull/1776.
//cc @ljharb @spectranaut

### DIFF
--- a/test/built-ins/AsyncFromSyncIteratorPrototype/return/absent-value-not-passed.js
+++ b/test/built-ins/AsyncFromSyncIteratorPrototype/return/absent-value-not-passed.js
@@ -1,0 +1,43 @@
+// Copyright (C) 2020 Alexey Shvayka. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-%asyncfromsynciteratorprototype%.return
+description: >
+  `return` method does not pass absent `value`.
+info: |
+  %AsyncFromSyncIteratorPrototype%.return ( value )
+
+  [...]
+  8. If value is present, then
+    [...]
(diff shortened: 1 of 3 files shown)


rwaldron:
@leobalter would you be available for a review?

test/built-ins/AsyncFromSyncIteratorPrototype/return/absent-value-not-passed.js
```diff
@@ -0,0 +1,45 @@
...
+};
+
+var asyncIterator = (async function* () {
+  yield* syncIterator;
+})();
+
+asyncIterator.next().then(function() {
+  return asyncIterator.throw();
+}).then(function(result) {
+  throw new Test262Error("Promise should be rejected, got: " + result.value);
```
rwaldron: Use `$DONE("...")` here

shvaikalesh:
Directories and front matter for `throw` and `return` were mixed up, now fixed.

iamstolis:
These tests assume that `syncIterator.next/return/throw()` is called without argument when the corresponding method of `asyncIterator` is called without argument. I believe that this is a wrong assumption. The tests are using `YieldExpression : yield * AssignmentExpression`. [The runtime semantics](https://tc39.es/ecma262/#sec-generator-function-definitions-runtime-semantics-evaluation) of this construction says that `received.[[Value]]` is passed to `next/return/throw` unconditionally there. This is not affected by [the normative change](https://github.com/tc39/ecma262/pull/1776) these tests attempt to verify. In other words, it seems that these tests are incorrect. Am I missing anything?

PR #2552, opened 2020-03-30
Test unicode escapes without Hex4Digits in strings

shvaikalesh:
JSC bug: ['\u' should throw an early SyntaxError exception, but instead evaluates to 'u'](https://bugs.webkit.org/show_bug.cgi?id=198790).
//cc @mathiasbynens


mathiasbynens:
Thanks!

PR #2554, opened 2020-03-30
Upstream language tag canonicalisation and Intl.Locale tests from SpiderMonkey

anba:
5663696:
- Updates the list of Intl service constructors to ensure existing tests are also run with RelativeTimeFormat, ListFormat, and DisplayNames.

f2fc3af:
- Updates the list of numbering systems to match current CLDR.

15db7f1:
- Python script used for the next two commits.

79524bf and 32fd063:
- Update the language tag canonicalisation to match UTS 35.

2a7c9f8:
- Variant subtags need to be canonicalised (again).

2e32d50:
- Using UTS 35 means "cel-gaulish" can now be canonicalised consistently across implementations, so we can revert #2054.

cce81ab:
- Additional canonicalisation and Intl.Locale tests from SpiderMonkey.

### DIFF
--- a/harness/testIntl.js
+++ b/harness/testIntl.js
@@ -39,7 +39,8 @@ function testWithIntlConstructors(f) {
   var constructors = ["Collator", "NumberFormat", "DateTimeFormat"];
 
   // Optionally supported Intl constructors.
-  ["PluralRules"].forEach(function(constructor) {
+  // NB: Intl.Locale isn't an Intl service constructor!
+  ["PluralRules", "RelativeTimeFormat", "ListFormat", "DisplayNames"].forEach(function(constructor) {
     if (typeof Intl[constructor] === "function") {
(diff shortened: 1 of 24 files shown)


rwaldron:
@anba do I understand correctly: once each engine updates to http://unicode.org/reports/tr35/, the tests here will pass?

anba:
Yes, when all parts of UTS 35 are implemented, the tests should pass. (Unless I'm misunderstanding some of the more loosely specified parts of UTS 35 :smile:.) 

The tests are currently already passing in SpiderMonkey, because we aren't using ICU for language tag canonicalisation. Frank improved the UTS 35 canonicalisation for ICU in unicode-org/icu#951, so when ICU 67 gets released later this month (and V8 updates to ICU 67), the number of passing tests will also increase for V8.

PR #2555, opened 2020-03-30
Adding test to cover case where there is a static field named as 'static'

caiolima:
As we added a case for instance field in https://github.com/tc39/test262/pull/2550, we also need a test case for static field named as `static` as well.


rwaldron:
Can you add a case that covers assigning a value—thanks!

caiolima:
> Can you add a case that covers assigning a value—thanks!

Sure!

caiolima:
@ljharb and @rwaldron: I also added a version fo instance field without assignment.

PR #2558, opened 2020-03-30
Improve RegExp LegacyOctalEscapeSequence coverage

shvaikalesh:
JSC bug: [Octal escapes in strings and regexes should be max 3 digits, but leading zeros are allowed](https://bugs.webkit.org/show_bug.cgi?id=167067).


leobalter:
cc @mathiasbynens

mathiasbynens:
You're on a roll @shvaikalesh! Nice work.

PR #2562, opened 2020-03-31
Promise.any: remove erroneous SpeciesConstructor tests until a better solution can be found. 

rwaldron:


leobalter:
this seems very reasonable

PR #2566, opened 2020-04-01
Updating test with 'o?.#field' to be valid program

### LINKED ISSUE #2563: Fix tests of `?.#field`
After consensus reached towards tc39/proposal-class-fields#302 , src/class-elements/grammar-private-field-optional-chaining.case is not a valid test anymore. We should fix it to be a valid syntax.

caiolima:
It closes #2563.


rwaldron:
I pushed a commit with the removal of those two files.

caiolima:
> The test files that were generated from the .case file that you altered need to be removed:
> 
> * [test/language/expressions/class/elements/syntax/early-errors/grammar-private-field-optional-chaining.js](https://github.com/tc39/test262/blob/8c9b4767c5ee48ba6bea49fc38270f08d8f3ef0c/test/language/expressions/class/elements/syntax/early-errors/grammar-private-field-optional-chaining.js)
> * [test/language/statements/class/elements/syntax/early-errors/grammar-private-field-optional-chaining.js](https://github.com/tc39/test262/blob/8c9b4767c5ee48ba6bea49fc38270f08d8f3ef0c/test/language/statements/class/elements/syntax/early-errors/grammar-private-field-optional-chaining.js)
> 
> I recommend running `./make.py clean` before generating tests.

Oops. I totally missed this, but thank you for letting me know about `./make.py clean`.

PR #2481, opened 2020-01-26
Improve RegExp.prototype[@@replace] coverage

shvaikalesh:
JSC bugs:
* [RegExpPrototype @@replace builtin needs to perform toLength](https://bugs.webkit.org/show_bug.cgi?id=173867)
* [RegExp.prototype[@@replace] doesn't coerce result index to integer](https://bugs.webkit.org/show_bug.cgi?id=209323)

See also #2426.

PR #2567, opened 2020-04-02
Improve coverage of RegExp.prototype[@@replace] with named groups

shvaikalesh:
While we have a bit more extensive coverage for named capture groups with `String.prototype.replace` in [`/test/built-ins/RegExp/named-groups`](https://github.com/tc39/test262/tree/master/test/built-ins/RegExp/named-groups), JSC has `RegExp.prototype[@@replace]` as slow path for RegExp subclasses and it lacked named groups support until [recently](https://trac.webkit.org/changeset/254195). Please also see [engine262 coverage report](https://coveralls.io/builds/29870435/source?filename=src/runtime-semantics/GetSubstitution.mjs).

JSC bug: [RegExp.prototype[Symbol.replace] does not support named capture groups](https://bugs.webkit.org/show_bug.cgi?id=205783).
Follow-up of #2481.
//cc @mathiasbynens

### DIFF
--- a/test/built-ins/RegExp/prototype/Symbol.replace/named-groups-fn.js
+++ b/test/built-ins/RegExp/prototype/Symbol.replace/named-groups-fn.js
@@ -1,0 +1,59 @@
+// Copyright (C) 2020 Alexey Shvayka. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-regexp.prototype-@@replace
+description: >
+  "groups" value is passed as last argument of replacer unless it is undefined.
+info: |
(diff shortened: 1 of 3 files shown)


leobalter:
cc @anba PTAL? SpiderMonkey is [currently failing](https://circleci.com/gh/tc39/test262/10320?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) and I wonder if you have some feedback on these?

shvaikalesh:
Seems like RegExp named capture groups is not yet implemented in SpiderMonkey: https://bugzilla.mozilla.org/show_bug.cgi?id=1362154.

PR #2568, opened 2020-04-02
Improve malformed RegExp named groups coverage

shvaikalesh:
This change:
a) Moves named references with look-behind tests to separate file, allowing engines without look-behind assertions support (JSC) to run them.
b) Adds a few tests with both malformed named and **numeric** references, as they both require reparsing and may interfere with each other.

JSC bug: [Invalid numeric and named references should be early syntax errors](https://bugs.webkit.org/show_bug.cgi?id=178175).
//cc @mathiasbynens

### DIFF
--- a/test/annexB/built-ins/RegExp/named-groups/non-unicode-malformed.js
+++ b/test/annexB/built-ins/RegExp/named-groups/non-unicode-malformed.js
@@ -6,7 +6,7 @@ description: >
   Named groups in Unicode RegExps have some syntax errors and some
   compatibility escape fallback behavior.
 esid: prod-GroupSpecifier
-features: [regexp-named-groups, regexp-lookbehind]
+features: [regexp-named-groups]
 includes: [compareArray.js]
 ---*/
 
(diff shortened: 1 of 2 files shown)


rwaldron:
@mathiasbynens _gentle nudge_

test/annexB/built-ins/RegExp/named-groups/non-unicode-malformed.js
```diff
@@ -17,11 +17,9 @@ assert(/\k/.test("k"));
...
+assert.compareArray(["k<a>"], "xxxk<a>xxx".match(/\k<a>/));
+assert.compareArray(["k<a"], "xxxk<a>xxx".match(/\k<a/));
 
-// A couple of corner cases around '\k' as named back-references vs. identity
-// escapes.
-assert(/\k<a>(?<=>)a/.test("k<a>a"));
-assert(/\k<a>(?<!a)a/.test("k<a>a"));
 assert(/\k<a>(<a>x)/.test("k<a><a>x"));
+assert(/\k<a>\1/.test("k<a>\x01"));
+assert(/\1(b)\k<a>/.test("bk<a>"));
```
leobalter, line 25: while the assertions are fine, it seems it would be slightly better to capture the results if they were positioned in a new test file, for similar reasons some assertions here are being set in a new file.

PR #2569, opened 2020-04-05
Add test for \0 in Unicode patterns

shvaikalesh:
JSC bug: [Octal escapes should be max 3 digits and syntax errors in Unicode patterns](https://bugs.webkit.org/show_bug.cgi?id=167067).

When fixing [test/built-ins/RegExp/unicode_restricted_octal_escape.js](https://github.com/tc39/test262/blob/master/test/built-ins/RegExp/unicode_restricted_octal_escape.js) it was quite easy to miss `\0` case and no tests were failing.

//cc @mathiasbynens

### DIFF
--- a/test/language/literals/regexp/u-null-character-escape.js
+++ b/test/language/literals/regexp/u-null-character-escape.js
@@ -1,0 +1,25 @@
+// Copyright (C) 2020 Alexey Shvayka. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: prod-CharacterEscape
+description: >
+  Null character escape is permitted in Unicode patterns.
+info: |
+  CharacterEscape[U] ::
+    ControlEscape
+    c ControlLetter
+    0 [lookahead ∉ DecimalDigit]
+    HexEscapeSequence
+    RegExpUnicodeEscapeSequence[?U]
+    IdentityEscape[?U]
+
+  DecimalDigit :: one of
+    0 1 2 3 4 5 6 7 8 9
+---*/
+
+var nullChar = String.fromCharCode(0);
+assert.sameValue(/\0/u.exec(nullChar)[0], nullChar);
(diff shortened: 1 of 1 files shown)


test/language/literals/regexp/u-null-character-escape.js
```diff
@@ -0,0 +1,24 @@
...
+    IdentityEscape[?U]
+
+  DecimalDigit :: one of
+    0 1 2 3 4 5 6 7 8 9
+---*/
+
+assert(/\0/u.test(String.fromCharCode(0)));
+assert(/\0a/u.test('\0a'));
+assert(/\0②/u.test('\x00②'));
+assert(/\0፬/u.test('\u0000፬'));
```
mathiasbynens, line 24: Since the intention seems to be to match the entire string, maybe add `^` and `$`?

shvaikalesh: GraalJS fails the test without `^` and `$`, but passes with them.
Weird, I reverted adding the assertions.

mathiasbynens: Maybe test both then? Or use `match` and check what is being matched?

shvaikalesh: Resolved in 7da7f38, thank you.

It seems like GraalJS parses `\0` differently in beginning of the pattern vs as second char.
//cc @iamstolis

mathiasbynens: Woah, good catch! Interesting bug.

iamstolis: > It seems like GraalJS parses \0 differently at the end of the pattern vs not.

Thanks for the report. Yes, this is a bug in GraalJS. We attempt to read a character behind `\0` which is not a good idea when `\0` is the last character of the pattern.

PR #2572, opened 2020-04-07
verifyProperty: gracefully handle `originalDesc` being `undefined`

shvaikalesh:
Output without this change:

```yaml
test/built-ins/Proxy/getOwnPropertyDescriptor/trap-is-missing-target-is-proxy.js:
  default: "TypeError: undefined is not an object (evaluating 'originalDesc.enumerable')"
  strict mode: "TypeError: undefined is not an object (evaluating 'originalDesc.enumerable')"
test/built-ins/Proxy/getOwnPropertyDescriptor/trap-is-null-target-is-proxy.js:
  default: "TypeError: undefined is not an object (evaluating 'originalDesc.value')"
  strict mode: "TypeError: undefined is not an object (evaluating 'originalDesc.value')"
test/built-ins/Proxy/getOwnPropertyDescriptor/trap-is-undefined-target-is-proxy.js:
  default: "TypeError: undefined is not an object (evaluating 'originalDesc.value')"
  strict mode: "TypeError: undefined is not an object (evaluating 'originalDesc.value')"
```

### DIFF
--- a/harness/propertyHelper.js
+++ b/harness/propertyHelper.js
@@ -43,6 +43,12 @@ function verifyProperty(obj, name, desc, options) {
 
     // desc and originalDesc are both undefined, problem solved;
     return true;
+  } else {
+    assert.notSameValue(
+      originalDesc,
+      undefined,
+      "obj['" + nameStr + "'] descriptor should not be undefined"
+    );
   }
 
   assert(
--- a/test/harness/verifyProperty-undefined-desc.js
+++ b/test/harness/verifyProperty-undefined-desc.js
@@ -18,6 +18,10 @@ assert.sameValue(
 );
 
 assert.throws(Test262Error, () => {
+  verifyProperty(sample, "foo2", {value: undefined});
+}, "originalDesc is undefined, desc is not undefined");
+
+assert.throws(Test262Error, () => {
   verifyProperty(sample, 'bar', undefined);
 }, "dataDescriptor value is undefined");
 


shvaikalesh:
`hasOwnProperty(x)` should do `getOwnPropertyDescriptor(this, x) !== undefined`, yet there is a bug in JSC which causes `getOwnPropertyDescriptor(obj, x)` return different result than `hasOwnProperty(x)`.

I don't think it worth changing `propertyHelper.js` because of it.

ljharb:
@shvaikalesh could you elaborate, or link to, the relevant jsc bug?

shvaikalesh:
I am pretty sure it is [ Fix DebuggerScope and/or ProxyObject to remove a hack in JSObject::getOwnPropertyDescriptor](https://bugs.webkit.org/show_bug.cgi?id=200560).

Some relevant code:

```cpp
bool JSObject::getOwnPropertyDescriptor(JSGlobalObject* globalObject, PropertyName propertyName, PropertyDescriptor& descriptor)
{
    VM& vm = globalObject->vm();
    auto scope = DECLARE_THROW_SCOPE(vm);
    JSC::PropertySlot slot(this, PropertySlot::InternalMethodType::GetOwnProperty);

    bool result = methodTable(vm)->getOwnPropertySlot(this, globalObject, propertyName, slot);
    EXCEPTION_ASSERT(!scope.exception() || !result);
    if (!result)
        return false;


    // FIXME: https://bugs.webkit.org/show_bug.cgi?id=200560
    // This breaks the assumption that getOwnPropertySlot should return "own" property.
    // We should fix DebuggerScope, ProxyObject etc. to remove this.
    //
    // DebuggerScope::getOwnPropertySlot() (and possibly others) may return attributes from the prototype chain
    // but getOwnPropertyDescriptor() should only work for 'own' properties so we exit early if we detect that
    // the property is not an own property.
    if (slot.slotBase() != this && slot.slotBase()) {
        JSProxy* jsProxy = jsDynamicCast<JSProxy*>(vm, this);
        if (!jsProxy || jsProxy->target() != slot.slotBase()) {
            // Try ProxyObject.
            ProxyObject* proxyObject = jsDynamicCast<ProxyObject*>(vm, this);
            if (!proxyObject || proxyObject->target() != slot.slotBase())
                return false;
        }
    }
```

That last `return false` causes the issue if the method is called on a Proxy w/o `getOwnPropertyDescriptor` trap and with another Proxy as it's `[[ProxyTarget]]`. `Object.prototype.hasOwnProperty` uses different code path.

PR #2573, opened 2020-04-07
Define Test262Error.prototype.name

shvaikalesh:
Please consider current output of a failing `async` test:

```yaml
test/built-ins/AsyncFromSyncIteratorPrototype/next/absent-value-not-passed.js:
  default: 'Test262:AsyncTestFailure:Test262Error: Test262Error: Expected SameValue(«1», «0») to be true'
  strict mode: 'Test262:AsyncTestFailure:Test262Error: Test262Error: Expected SameValue(«1», «0») to be true'
```

<br>`Test262Error` is duplicated because `Test262Error` does not implement `Error` interface `$DONE` expects it to:

```js
function $DONE(error) {
  if (error) {
    if(typeof error === 'object' && error !== null && 'name' in error) { // Test262Error has no "name"
      __consolePrintHandle__('Test262:AsyncTestFailure:' + error.name + ': ' + error.message);
    } else {
      __consolePrintHandle__('Test262:AsyncTestFailure:Test262Error: ' + error);
    }
  } else {
    __consolePrintHandle__('Test262:AsyncTestComplete');
  }
}
```

<br> and gets converted to string.

### DIFF
--- a/harness/sta.js
+++ b/harness/sta.js
@@ -14,6 +14,7 @@ function Test262Error(message) {
   this.message = message || "";
 }
 
+Test262Error.prototype.name = "Test262Error";
 Test262Error.prototype.toString = function () {
   return "Test262Error: " + this.message;
 };
--- a/test/harness/sta.js
+++ b/test/harness/sta.js
@@ -9,6 +9,7 @@ description: >
 ---*/
 
 assert(typeof Test262Error === "function");
+assert.sameValue(Test262Error.prototype.name, "Test262Error");
 assert(typeof Test262Error.prototype.toString === "function");
 assert(typeof $ERROR === "function");
 assert(typeof $DONOTEVALUATE === "function");


rwaldron:
This can't be merged until it's been tested with all tools that consume these output messages.

leobalter:
This reminds me how some changes in the structure of this project might cause such burden we would not usually expect for.

These details might break test262 runners and I'm inclined to wait merging this until we get some stats on different runners. The CI here only show results caught from test262-harness, but most browsers use their own runner, and there are even some out there I've never touched.

While this seems too much precaution, I have to remember (myself) the days I almost broke results for Chrome by removing the **inline comments** from the head of each test file. Or when the YAML syntax needs to be heavily restricted and enforced for the metadata to not break many test runners. That's why we need a linter.

The benefit of this change is not having the Test262Error name being duplicated. It's unfortunate the cost is too high.

leobalter:

PR #2577, opened 2020-04-14
Add Proxy.revocable built-in test

shvaikalesh:


test/built-ins/Proxy/revocable/builtin.js
```diff
@@ -0,0 +1,18 @@
...
+features: [Proxy, Reflect.construct]
+---*/
+
+assert(Object.isExtensible(Proxy.revocable));
+assert.sameValue(typeof Proxy.revocable, 'function');
+assert.sameValue(Object.prototype.toString.call(Proxy.revocable), '[object Function]');
+assert.sameValue(Object.getPrototypeOf(Proxy.revocable), Function.prototype);
+
+assert.sameValue(Proxy.revocable.hasOwnProperty('prototype'), false);
+assert.sameValue(isConstructor(Proxy.revocable), false);
```
shvaikalesh, line 18: JSC defines throwing `[[Construct]]` method for a nicer error message instead of no method at all.
Using `isConstructor` is the simplest way to catch this kind of incompatibilities.

PR #2578, opened 2020-04-14
Assert that Symbol is a constructor

shvaikalesh:


test/built-ins/Symbol/is-constructor.js
```diff
@@ -0,0 +1,19 @@
...
+  [...]
+  * may be used as the value of an extends clause of a class definition
+  but a super call to it will cause an exception.
+  [...]
+  * has a [[Prototype]] internal slot whose value is %Function.prototype%.
+includes: [isConstructor.js]
+features: [Symbol, Reflect.construct]
+---*/
+
+assert(isConstructor(Symbol));
```
shvaikalesh, line 18: JSC doesn't define `[[Construct]]` method for `Symbol`, yet `class extends Symbol` works because [it doesn't perform `IsConstructor`](https://bugs.webkit.org/show_bug.cgi?id=193057).

PR #2580, opened 2020-04-15
Atomics wait async

rwaldron:
@leobalter can you give this a quick look? There is no way for you to run these test currently. I have a polyfill that I'm using to write tests against, but it currently requires a hacked up version of eshost.


test/built-ins/Atomics/wait/validate-arraytype-before-index-coercion.js
```diff
@@ -2,7 +2,7 @@
 // This code is governed by the BSD license found in the LICENSE file.
 
 /*---
-esid: sec-atomics.notify
+esid: sec-atomics.wait
 description: >
```
rwaldron, line 6: Note to myself: undo this

PR #2581, opened 2020-04-15
Simplify alphanum regex in testIntl.js

rkirsling:
JSC is currently [mishandling](https://bugs.webkit.org/show_bug.cgi?id=210576) the duplicate variant regex in testIntl.js.

That isn't test262's problem, but it's still incredibly silly that we're treating every `alphanum` subexpression as a capture group, so let's fix this.

### DIFF
--- a/harness/testIntl.js
+++ b/harness/testIntl.js
@@ -272,10 +272,10 @@ function isCanonicalizedStructurallyValidLanguageTag(locale) {
    *
    * Spec: https://unicode.org/reports/tr35/#Unicode_locale_identifier
    */
-  var alpha = "[a-z]",
+  var alpha = "[A-Za-z]",
     digit = "[0-9]",
-    alphanum = "(" + alpha + "|" + digit + ")",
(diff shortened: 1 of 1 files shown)


rwaldron:
How do these changes impact all of the test files that use this helper file?

rkirsling:
It's intended to be fully equivalent—the unnecessary capture group is rather confusing, but it only affects the backreference in the duplicate variant regex.

(Removing the `i` flag is just extra cleanup; I think it's a bit simpler but feel free to disagree.)

anba:
@rwaldron Please revert this change.

@rkirsling The `i` flag is required for these regular expressions to work correctly.

For example the following calls should both return `true`, but are currently returning `false`.

```js
js> isStructurallyValidLanguageTag("und-X-b")
false
js> isStructurallyValidLanguageTag("und-U-kf")      
false
```

And this call should return `false`, but is currently returning `true`.

```js
js> isStructurallyValidLanguageTag("und-variant-VARIANT")
true
```

PR #2583, opened 2020-04-18
Make GetFunctionRealm tests useful

shvaikalesh:
JSC bug: [InternalFunction::createSubclassStructure should use newTarget's globalObject](https://bugs.webkit.org/show_bug.cgi?id=202599).
Follow-up of #2155.

### DIFF
--- a/test/built-ins/Function/prototype/bind/get-fn-realm-recursive.js
+++ b/test/built-ins/Function/prototype/bind/get-fn-realm-recursive.js
@@ -3,23 +3,51 @@
 /*---
 esid: sec-getfunctionrealm
 description: >
-    The realm of a bound function exotic object is the realm of its target
-    function. GetFunctionRealm is called recursively.
+  The realm of a bound function exotic object is the realm of its target function.
+  GetFunctionRealm is called recursively.
 info: |
-    7.3.22 GetFunctionRealm ( obj )
+  Object ( [ value ] )
 
-    [...]
-    2. If obj has a [[Realm]] internal slot, then
-       a. Return obj.[[Realm]].
-    3. If obj is a Bound Function exotic object, then
-       a. Let target be obj.[[BoundTargetFunction]].
-       b. Return ? GetFunctionRealm(target).
-features: [cross-realm]
+  1. If NewTarget is neither undefined nor the active function, then
+    a. Return ? OrdinaryCreateFromConstructor(NewTarget, "%Object.prototype%").
+
+  OrdinaryCreateFromConstructor ( constructor, intrinsicDefaultProto [ , internalSlotsList ] )
+
+  [...]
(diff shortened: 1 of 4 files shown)


rwaldron:
There's no reason to change the existing tests, as they function correctly. If you want to add more _and different_ test coverage, then please do so.

shvaikalesh:
JSC have no implementation of `GetFunctionRealm`, yet all 4 of those tests pass.

test/built-ins/Function/prototype/bind/get-fn-realm-recursive.js
```diff
@@ -3,23 +3,51 @@
...
+  3. If obj is a bound function exotic object, then
+    a. Let target be obj.[[BoundTargetFunction]].
+    b. Return ? GetFunctionRealm(target).
+features: [cross-realm, Reflect]
 ---*/
 
-var other = $262.createRealm().global;
-var C = new other.Function();
-C.prototype = null;
-var B = C.bind().bind();
```
shvaikalesh, line 23: `Function.prototype.bind` is from the same Realm as `C`.

test/built-ins/Function/prototype/bind/get-fn-realm-recursive.js
```diff
@@ -3,23 +3,51 @@
...
+var realm1 = $262.createRealm().global;
```
shvaikalesh, line 40: I am using current Realm as less as possible to avoid false-positive test results.

rwaldron:
I wonder then, if the issue is a failure of JSC's $262 implementation

shvaikalesh:
I can confirm that JSC with [patch that implements `GetFunctionRealm`](https://bugs.webkit.org/show_bug.cgi?id=202599) passes these 4 tests.

JSC's `$262.createRealm` is a very thin wrapper around `GlobalObject::create`, I would not expect it to cause any issues.

The problem is that current tests (without proposed change) do not reach steps 3 and 4 of [`GetFunctionRealm`](https://tc39.es/ecma262/#sec-getfunctionrealm) they are supposed to (judging from `info` sections), as both bound function and `Proxy` exotics forward `[[Construct]]` calls to their targets, and `GetFunctionRealm` is end up being called on regular function.

To test steps 3 and 4, we need `Reflect.construct` and `NewTarget` this change is adding.

shvaikalesh:
Also, paths that existing tests take are covered by:
* [test/built-ins/Function/prototype/bind/proto-from-ctor-realm.js](https://github.com/tc39/test262/tree/master/test/built-ins/Function/prototype/bind/proto-from-ctor-realm.js)
* [test/built-ins/Proxy/construct/trap-is-undefined-proto-from-newtarget-realm.js](https://github.com/tc39/test262/tree/master/test/built-ins/Proxy/construct/trap-is-undefined-proto-from-newtarget-realm.js)

Issue #2585, opened 2020-04-19
Incorrect test-cases for Intl.NumberFormat with signDisplay: "exceptZero"


iamstolis:
When `Intl.NumberFormat` is used with `signDisplay` set to `exceptZero` then the resulting pattern (`zeroPattern`/`positivePattern`/`negativePattern`) is determined after rounding of the argument, see [step 15 of GetNumberFormatPattern](https://tc39.es/ecma402/#sec-getnumberformatpattern) and [steps 4 and 5 of PartitionNumberPattern](https://tc39.es/ecma402/#sec-partitionnumberpattern). Hence, a non-zero number that is rounded to zero should be formatted using `zeroPattern`, i.e., without sign. This is reflected by [signDisplay-rounding.js](https://github.com/tc39/test262/blob/master/test/intl402/NumberFormat/prototype/format/signDisplay-rounding.js) correctly. Unfortunately, there are other tests that fail to handle these corner cases correctly:

[format/signDisplay-de-DE.js](https://github.com/tc39/test262/blob/master/test/intl402/NumberFormat/prototype/format/signDisplay-de-DE.js#L50) expects `-0.0001` (and even `-0`) to be formatted as `-0` (but the correct result is `0`). Surprisingly, the same test expects no sign while formatting `0.0001` correctly.

[format/signDisplay-currency-de-DE.js](https://github.com/tc39/test262/blob/master/test/intl402/NumberFormat/prototype/format/signDisplay-currency-de-DE.js#L41) expects `-0.0001`, `-0`, `0.0001` to be formatted using signed pattern.

Other language variants of these tests as well as their `formatToParts` counterpars have the same issues.

FrankYFTang:
I agree with @iamstolis @tc39/ecma-fellows  I am crafting a PR to correct this. @Ms2ger

sffc:
We discussed this at length in the proposal and decided for a number of reasons that small, negative numbers that round to zero should display without the sign in exceptZero.

- Chart: https://github.com/tc39/proposal-unified-intl-numberformat#iii-sign-display
- Issue: https://github.com/tc39/proposal-unified-intl-numberformat/issues/66
- Spec: https://tc39.es/ecma402/#sec-getnumberformatpattern

So, I agree that the Test262 is incorrect and needs to be fixed.

FrankYFTang:
[...]

PR #2454, opened 2019-12-23
More tests for Unified NumberFormat.

Ms2ger:
Fixes #2449.

PR #2586, opened 2020-04-20
Correct the exception of signDisplay: "exceptZero" for Intl.NumberFormat

FrankYFTang:
Close https://github.com/tc39/test262/issues/2585 
See https://github.com/tc39/test262/issues/2585 for detail of why the current expectation is wrong.
-0.001, -0 and +0.001 should all be formatted the same as 0 when signDisplay: "exceptZero" since the sign is decided after the rounding. @iamstolis @tc39/ecma-fellows 

Address changes in https://github.com/tc39/proposal-unified-intl-numberformat/pull/72


FrankYFTang:
@leobalter @Ms2ger

Ms2ger:
Looks like I missed these in #2454; thanks for following up.

iamstolis:
LGTM. Thanks for a quick correction of these tests.

Issue #2589, opened 2020-04-22
expectation in intl402/RelativeTimeFormat/prototype/format*/pl-pl*js is incorrect


FrankYFTang:
As issues raised by @rkirsling  in https://github.com/tc39/test262/pull/1669 the expectation for pl is incorrect. 
```
rkirsling commented 5 days ago
Are these tests valid with the thousands separator?

I see that V8 and SM pass them, but this seems wrong, as CLDR has minimumGroupingDigits{"2"} for Polish and ureldatefmt_format[Numeric] respects that in ICU 64.

V8 and SM implementations of RelativeTimeFormat seem to ignore minimumGroupingDigits for other locales like es too, but I'm not seeing a reason for this in ECMA-402 or ICU's commit history.
```

Bug is filed against ICU in https://unicode-org.atlassian.net/browse/ICU-21086

This is indeed incorrect expectation in these test. there should be no grouping separator for pl while the value is < 10000 since pl has minimumGroupingDigits{"2"}

Discuss this in ICU-TC 4/22/2020 with Peter Edberg from Apple and all agree since relative time format is format with grouping, minimumGroupingDigits{"2"} should be respect.

FrankYFTang:
v8 issue filed in https://bugs.chromium.org/p/v8/issues/detail?id=10443

FrankYFTang:
@Ms2ger @leobalter @tc39/ecma-fellows @anba @zbraniecki

PR #2590, opened 2020-04-22
Correct pull/1669 for pl

FrankYFTang:
Correct tests landed in https://github.com/tc39/test262/pull/1669
to close https://github.com/tc39/test262/issues/2589

pl locale has minimumGroupingDigits{"2"}  and should not place \u00a0 if there are only 4 digits (since  minimumGroupingDigits{"2"}   means it require at least 2 digits on the left of the grouping separator)
@Ms2ger @leobalter @rkirsling


FrankYFTang:
The upstream problem why v8 and spidermonkey is broken is in ICU . https://unicode-org.atlassian.net/browse/ICU-21086 is filed.

FrankYFTang:
@anba

FrankYFTang:
@leobalter @rwaldron

PR #2592, opened 2020-04-22
Allow new property such as fractionalSecondDigits present in resolvedOptions result

### LINKED ISSUE #2591: missing fractionalSecondDigits in resolvedOptions/{order,order-dayPeriod}.js
Tests in 
intl402/DateTimeFormat/prototype/resolvedOptions/order* is written to
"Verifies the property order for the object returned by resolvedOptions()."
notice it is currently using assert.compareArray to ensure the exact order and element of the return object of resolvedOptions.
However, in PR https://github.com/tc39/ecma402/pull/347/files
The loop of "For each row of Table 6, except the header row, in table order, do" will always once have _prop_ as "fractionalSecondDigits" and the calling to GetNumberOption(options, "fractionalSecondDigits", 0, 3, 0). will get 0 if there are no fractionalSecondDigits or fractionalSecondDigits is undefined. Therefore, fractionalSecondDigits in the resolvedOptions should be 0 if it is undefined in the input options when there are no dateStyle nor timeStyle present.

Notice this is not the case for order-style since for the cases of dateStyle and timeStyle set, the loop in the Else won't be hit so that test is correctly specified without fractionalSecondDigits. 

Problem test
https://github.com/tc39/test262/blob/master/test/intl402/DateTimeFormat/prototype/resolvedOptions/order-dayPeriod.js
https://github.com/tc39/test262/blob/master/test/intl402/DateTimeFormat/prototype/resolvedOptions/order.js

FrankYFTang:
This fix the conflict with the https://github.com/tc39/ecma402/pull/347 
so the resolvedOptions contains fractionalSecondDigits even it is undefined in the input, except the cases which dateStyle or timeStyle is specified.
close #2591

### DIFF
--- a/test/intl402/DateTimeFormat/prototype/resolvedOptions/order-dayPeriod.js
+++ b/test/intl402/DateTimeFormat/prototype/resolvedOptions/order-dayPeriod.js
@@ -28,8 +28,8 @@ const expected = [
 
 let actual = Object.getOwnPropertyNames(options);
 
-// Ensure all expected items are in actual and also allow other property
-// implemented in other new proposal.
+// Ensure all expected items are in actual and also allow other properties
+// implemented in new proposals.
 assert(arrayContains(actual, expected));
 for (var i = 1; i < expected.length; i++) {
   // Ensure the order as expected but allow additional new property in between
--- a/test/intl402/DateTimeFormat/prototype/resolvedOptions/order-dayPeriod.js
+++ b/test/intl402/DateTimeFormat/prototype/resolvedOptions/order-dayPeriod.js
@@ -5,7 +5,7 @@
 esid: sec-intl.datetimeformat.prototype.resolvedoptions
 description: Verifies the property order for the object returned by resolvedOptions().
 includes: [compareArray.js]
-features: [Intl.DateTimeFormat-dayPeriod]
+features: [Intl.DateTimeFormat-dayPeriod,Intl.DateTimeFormat-fractionalSecondDigits]
 ---*/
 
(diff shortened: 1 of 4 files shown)


FrankYFTang:
@Ms2ger @zbraniecki @tc39/ecma-fellows @leobalter @anba

FrankYFTang:
> Could you do this in a new copy of the test, so implementations that don't implement [tc39/ecma402#347](https://github.com/tc39/ecma402/pull/347) yet don't lose coverage?

The point is the implementation which implement tc39/ecma402#347](https://github.com/tc39/ecma402/pull/347 won't be able to pass this test because the test is currently assert by comparing the array and the fractionalSecondDigits will always present in the resolvedOptions. I think I will change the test in a different way to cover both. Stay tune

FrankYFTang:
OK, I change the subject of this PR and the test to make it check
1. the actual result contains the expected
2. the item in the actual result follow the order of the expected
AND
3. allow other property from new proposal, such as fractionalSecondDigits, present in the actual result.
In this way engine which implement one new proposal won't fail test designed for the browser which has not support that prorposal.

FrankYFTang:
@rwaldron  I answered the issue- could you remove the awaiting-author l label now? Thanks

Ms2ger:
Do we not want a test that ensures "fractionalSecondDigits" is actually included?

FrankYFTang:
> Do we not want a test that ensures "fractionalSecondDigits" is actually included?

It is already there in order-fractionalSecondDigits.js and I make that test also not failing if dayPeriod is also supported.

PR #2597, opened 2020-04-24
Correctly tag AggregateError proto-from-ctor-realm test

syg:
It'd be nice to have a CI hook that enforces presence of feature tags for all tests in a built-in directory.


rwaldron:
> It'd be nice to have a CI hook that enforces presence of feature tags for all tests in a built-in directory

I agree! I've wanted that for years, but certain folks thought that hard enforcement of feature tags was too much of an imposition. Glad I have your support on this! I'll start working on ways to enforce feature tags.

syg:
> I agree! I've wanted that for years, but certain folks thought that hard enforcement of feature tags was too much of an imposition. 

Interesting point re: the imposition. My feeling here is that for the implementations that flag their features, the main feature flag needs to be there for stage 3 features or the test doesn't run correctly. Additional flags for stage 4 features that have been shipping for some time are probably needless imposition though.

PR #2598, opened 2020-04-25
Correct the expectation of zh-Hant for Locale.minimize

FrankYFTang:
It is incorrect to expect the minimize result of "zh-Hant" to be "zh-TW". It should be "zh-Hant". Why?

first, what we have in input for zh-Hant
lang = zh
region = [none]
script = Hant 

Now, look at the AddLikelySubtags algorithm in http://www.unicode.org/reports/tr35/#Likely_Subtags

```
Remove Likely Subtags: Given a locale, remove any fields that Add Likely Subtags would add.
The reverse operation removes fields that would be added by the first operation.
1. First get max = AddLikelySubtags(inputLocale). If an error is signaled, return it.
2. Remove the variants from max.
3. Then for trial in {language, language _ region, language _ script}
3-* If AddLikelySubtags(trial) = max, then return trial + variants.
4. If you do not get a match, return max + variants.
```

after 1 and 2, max = AddLikelySubtags(zh-Hant) = “zh-Hant-TW”
in 3, we loop through the following
A trial = language = zh. AddLikelySubtags(zh) = ”zh-Hans-CN” != max
[and there are “no region code”, so no language _ region for trial]
B. trial = language _ script = zh_Hant AddLikelySubtags(zh-Hant) = ”zh-Hant-TW” == max so we return trial + v which is “zh_Hant”

### DIFF
--- a/test/intl402/Locale/prototype/minimize/removing-likely-subtags-first-adds-likely-subtags.js
+++ b/test/intl402/Locale/prototype/minimize/removing-likely-subtags-first-adds-likely-subtags.js
@@ -34,7 +34,7 @@ var testDataMinimal = {
     // https://unicode-org.atlassian.net/browse/ICU-12345
     "und-CW": "pap-CW",
     "und-US": "en",
-    "zh-Hant": "zh-TW",
+    "zh-Hant": "zh-Hant",
     "zh-Hani": "zh-Hani",
 };
 


FrankYFTang:
@anba @Ms2ger @rwaldron @leobalter @littledan

anba:
@rwaldron Please revert this change.

@FrankYFTang The "zh-Hant" minimises to "zh-TW" example was taken straight from the specification, directly below the `Remove Likely Subtags` algorithm:

> Example:
>    - Input is zh_Hant. Maximize to get zh_Hant_TW.
>    - zh => zh_Hans_CN. No match, so continue.
>    - zh_TW => zh_Hant_TW. Matches, so return zh_TW.
>
> A variant of this favors the script over the region, thus using {language, language_script, language_region} in the above. If that variant is used, then the result in this example would be zh_Hant instead of zh_TW.

FrankYFTang:
hum.... @anba you are right. So... the language, language_region and language_script are referring to the one inside the max, not the one inside the input?

anba:
Yes, that seems to be what the spec implies when following the example given there. Maybe the spec would be clearer if it was changed to use subscripts, similar to the "Add Likely Subtags" operation.

PR #2600, opened 2020-04-27
Partly revert "Simplify alphanum regex in testIntl.js"

shvaikalesh:
This PR reverts `i` flag change per https://github.com/tc39/test262/pull/2581#issuecomment-619856574, but keeps unnecessary capture group removal since JSC currently [fails to match it correctly](https://bugs.webkit.org/show_bug.cgi?id=210576).

_cc_ @anba @rkirsling

### DIFF
--- a/harness/testIntl.js
+++ b/harness/testIntl.js
@@ -272,10 +272,10 @@ function isCanonicalizedStructurallyValidLanguageTag(locale) {
    *
    * Spec: https://unicode.org/reports/tr35/#Unicode_locale_identifier
    */
-  var alpha = "[A-Za-z]",
+  var alpha = "[a-z]",
     digit = "[0-9]",
-    alphanum = "[A-Za-z0-9]",
-    variant = "(" + alphanum + "{5,8}|(?:" + digit + alphanum + "{3}))",
+    alphanum = "(" + alpha + "|" + digit + ")",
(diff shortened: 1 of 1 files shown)


anba:
`alphanum` needs to be a non-capturing group or alternatively you also need to revert the group reference change in `duplicateVariant` from #2581.

rkirsling:
What? Why on earth would we need to write `(?:[a-z]|[0-9])` instead of `[a-z0-9]`? This was the whole point of the change (as JSC's current issue is not limited to capturing groups).

rkirsling:
Oh oops, I misunderstood the course of the commits in this PR. Yes, changing `([a-z]|[0-9])` to `[a-z0-9]` and decrementing the backreference is the only essential part of this change.

harness/testIntl.js
```diff
@@ -272,9 +272,9 @@ function isCanonicalizedStructurallyValidLanguageTag(locale) {
    *
    * Spec: https://unicode.org/reports/tr35/#Unicode_locale_identifier
    */
-  var alpha = "[A-Za-z]",
+  var alpha = "[a-z]",
     digit = "[0-9]",
-    alphanum = "[A-Za-z0-9]",
+    alphanum = "(?:" + alpha + "|" + digit + ")",
```
rkirsling, line 277:
```suggestion
    alphanum = "[a-z0-9]",
```

PR #2601, opened 2020-04-28
Revert "Correct the expectation of zh-Hant for Locale.minimize"

FrankYFTang:
Reverts tc39/test262#2598


FrankYFTang:
@anba @rwaldron

PR #2608, opened 2020-05-02
Correct value to canonicalized locale

FrankYFTang:
(new Intl.Locale("uz-UZ-cyrillic")).toString*) should return "uz-Latn-UZ-cyrillic" not "uz-UZ-cyrillic" because according to
https://tc39.es/ecma402/#sec-canonicalizeunicodelocaleid
"2.  Let localeId be the string localeId after performing the algorithm to transform it to canonical form. (The result is a Unicode BCP 47 locale identifier, in both canonical syntax and canonical form.)"

And https://unicode.org/reports/tr35/#Canonical_Unicode_Locale_Identifiers stated
* Replace aliases in the unicode_language_id and tlang (if any) using the following process:
** If the language subtag matches the type attribute of a languageAlias element in Supplemental Data, replace the language subtag with the replacement value.
*** If there are additional subtags in the replacement value, add them to the result, but only if there is no corresponding subtag already in the tag.

and in https://github.com/unicode-org/cldr/blob/master/common/supplemental/supplementalMetadata.xml
    		<languageAlias type="uz_UZ" replacement="uz_Latn_UZ" reason="legacy"/>
so uz_UZ will be replaced as uz_Latn_UZ during the 

"6.2.3 CanonicalizeUnicodeLocaleId ( locale )"

### DIFF
--- a/test/intl402/Intl/getCanonicalLocales/non-iana-canon.js
+++ b/test/intl402/Intl/getCanonicalLocales/non-iana-canon.js
@@ -23,6 +23,7 @@ var testData = [
     },
     {
         tag: "uz-UZ-cyrillic",
+        canonical: "uz-Latn-UZ-cyrillic",
     },
     {
         tag: "posix",
--- a/test/intl402/Locale/constructor-non-iana-canon.js
+++ b/test/intl402/Locale/constructor-non-iana-canon.js
@@ -32,6 +32,7 @@ var testData = [
     },
     {
         tag: "uz-UZ-cyrillic",
+        canonical: "uz-Latn-UZ-cyrillic",
         maximized: "uz-Latn-UZ-cyrillic",
         minimized: "uz-cyrillic",
     },
--- a/test/intl402/Segmenter/constructor/supportedLocalesOf/locales-specific.js
+++ b/test/intl402/Segmenter/constructor/supportedLocalesOf/locales-specific.js
@@ -19,4 +19,6 @@ assert.sameValue(typeof Intl.Segmenter.supportedLocalesOf, "function",
 assert.compareArray(Intl.Segmenter.supportedLocalesOf("sr"), ["sr"]);
 
 const multiLocale = ["sr-Thai-RS", "de", "zh-CN"];
-assert.compareArray(Intl.Segmenter.supportedLocalesOf(multiLocale), multiLocale);
+const multiCanonLocale = ["sr-Thai-RS", "de", "zh-Hans-CN"];
+
+assert.compareArray(Intl.Segmenter.supportedLocalesOf(multiLocale), multiCanonLocale);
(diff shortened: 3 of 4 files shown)


FrankYFTang:
@anba @Ms2ger @rwaldron

anba:
> If the **language subtag** matches the type attribute of a languageAlias element in Supplemental Data, replace the language subtag with the replacement value.

(Emphasis added)

Per UTS 35, the canonicalisation only checks for the "language subtag" when looking for replacements, so this entry shouldn't be taken into account:
```xml
<languageAlias type="uz_UZ" replacement="uz_Latn_UZ" reason="legacy"/>
```

Unless this is a spec bug in UTS 35 and the matching should include additional subtags besides the language subtag! 

If that's indeed a spec bug and the canonical form of "uz-UZ" is "uz-Latn-UZ", <https://github.com/tc39/test262/blob/master/tools/misc/make_intl_data.py> and <https://github.com/tc39/test262/blob/master/harness/testIntl.js> also need to be updated to properly handle this case.

---

There are other inconsistencies between UTS 35 and the actual data. For example [supplementalMetadata.xml](https://github.com/unicode-org/cldr/blob/da75520653344830bb194192ce4b19187e6c8bc6/common/supplemental/supplementalMetadata.xml#L465) includes
```xml
<scriptAlias type="Qaai" replacement="Zinh" reason="deprecated"/>
```
and [UTS 35, §3.4 Language Identifier Field Definitions](https://unicode.org/reports/tr35/tr35.html#unicode_script_subtag_validity) mentions that the canonicalised form for "Qaai" is "Zinh", but in [UTS 35, §3.2.1 Canonical Unicode Locale Identifiers](https://unicode.org/reports/tr35/tr35.html#Canonical_Unicode_Locale_Identifiers) script subtag aliases aren't handled.

FrankYFTang:
> > If the **language subtag** matches the type attribute of a languageAlias element in Supplemental Data, replace the language subtag with the replacement value.
> 
> (Emphasis added)
> 
> Per UTS 35, the canonicalisation only checks for the "language subtag" when looking for replacements, so this entry shouldn't be taken into account:
> 
> ```
> <languageAlias type="uz_UZ" replacement="uz_Latn_UZ" reason="legacy"/>
> ```
> 
> Unless this is a spec bug in UTS 35 and the matching should include additional subtags besides the language subtag!
> 
> If that's indeed a spec bug and the canonical form of "uz-UZ" is "uz-Latn-UZ", https://github.com/tc39/test262/blob/master/tools/misc/make_intl_data.py and https://github.com/tc39/test262/blob/master/harness/testIntl.js also need to be updated to properly handle this case.
> 
> There are other inconsistencies between UTS 35 and the actual data. For example [supplementalMetadata.xml](https://github.com/unicode-org/cldr/blob/da75520653344830bb194192ce4b19187e6c8bc6/common/supplemental/supplementalMetadata.xml#L465) includes
> 
> ```
> <scriptAlias type="Qaai" replacement="Zinh" reason="deprecated"/>
> ```
> 
> and [UTS 35, §3.4 Language Identifier Field Definitions](https://unicode.org/reports/tr35/tr35.html#unicode_script_subtag_validity) mentions that the canonicalised form for "Qaai" is "Zinh", but in [UTS 35, §3.2.1 Canonical Unicode Locale Identifiers](https://unicode.org/reports/tr35/tr35.html#Canonical_Unicode_Locale_Identifiers) script subtag aliases aren't handled.

your reply make sense. I filed https://unicode-org.atlassian.net/browse/CLDR-13758 and we need to discuss with Mark Davis about this.

FrankYFTang:
I added more changes to this PR and include other cases which will be impacted by the interpretation of https://tc39.es/ecma402/#sec-canonicalizeunicodelocaleid

rwaldron:
@anba can you take another look? Thanks!

FrankYFTang:
Please review but not merge. We are still nailing down the details in CLDR and UTS35

leobalter:

FrankYFTang:
The PR is not aligned with CLDR 38. Close

PR #2612, opened 2020-05-05
Fix testing for throw in an iterator via yield* throws

syg:
IsHTMLDDA is specified in INTERPRETING.md to return null on [[Call]]
when called with no arguments or with single argument "". Returning null
causes the iterator protocol to throw.

Also see star-iterable-return-emulates-undefined-throws-when-called.js


syg:
@shvaikalesh Does this seem like a bug or did I miss something for why this test doesn't currently test for throwing?

rwaldron:
Oof, I misinterpreted the CI results as well. Sorry about that folks.

PR #2615, opened 2020-05-08
Improve [[IsHTMLDDA]] coverage for built-ins

shvaikalesh:
JSC bug: [Use @isUndefinedOrNull instead of abstract equality with null](https://bugs.webkit.org/show_bug.cgi?id=210954).
Follow-up of #2579.


shvaikalesh:
Tests fail as `$262.IsHTMLDDA` is not properly implemented for most engines. 
https://github.com/bterlson/eshost/pull/105 attempts to solve it for JSC; I need to investigate CI failure though.

PR #2616, opened 2020-05-08
Fix exported bindings and ns

### LINKED ISSUE #2614: export-star-as-dflt seems wrong
The test seems to be testing that:

[L21](https://github.com/tc39/test262/blob/master/test/language/module-code/export-star-as-dflt.js#L21) Re-export all the imports as a named export named `default`
[L22](https://github.com/tc39/test262/blob/master/test/language/module-code/export-star-as-dflt.js#L22) Import the default export as a namespace object `ns` from itself.
[L23](https://github.com/tc39/test262/blob/master/test/language/module-code/export-star-as-dflt.js#L23) Test that the namespace object has a named export named `default`.

L21 doesn't actually make a default export (as opposed to a named export named `default`), so how does L22 import a namespace object?

@adrianheine What is this trying to test?

leobalter:
Closes #2614

cc @syg

### DIFF
--- a/test/language/module-code/export-star-as-dflt.js
+++ b/test/language/module-code/export-star-as-dflt.js
@@ -19,5 +19,10 @@ features: [export-star-as-namespace-from-module]
 ---*/
 
 export * as default from './export-star-as-dflt_FIXTURE.js';
-import ns from './export-star-as-dflt.js';
+import Self from './export-star-as-dflt.js';
+import { default as named } from './export-star-as-dflt.js';
+import * as ns from './export-star-as-dflt.js';
+
+assert.sameValue(Self.x, 1, 'Module was re-exported under the name `default`');
+assert.sameValue(named.x, 1, 'named binding was re-exported under the name `default`');
 assert.sameValue(ns.default.x, 1, 'namespace was re-exported under the name `default`');


test/language/module-code/export-star-as-dflt.js
```diff
@@ -19,5 +19,10 @@ features: [export-star-as-namespace-from-module]
...
 
 export * as default from './export-star-as-dflt_FIXTURE.js';
-import ns from './export-star-as-dflt.js';
+import Self from './export-star-as-dflt.js';
+import { default as named } from './export-star-as-dflt.js';
+import * as ns from './export-star-as-dflt.js';
+
+assert.sameValue(Self.x, 1, 'Module was re-exported under the name `default`');
+assert.sameValue(named.x, 1, 'named binding was re-exported under the name `default`');
 assert.sameValue(ns.default.x, 1, 'namespace was re-exported under the name `default`');
```
leobalter, line 28: @syg I added these other cases as valid ways to verify the same export-star-from.

At the same time, I just noticed we don't have a simple coverage for `import { default as ...` and this is the first actual test verifying a positive importing of a default named binding.

leobalter:
It's interesting to see SpiderMonkey failing here. @anba PTAL?

anba:
SpiderMonkey doesn't yet implement this syntax: https://bugzilla.mozilla.org/show_bug.cgi?id=1496852

leobalter:
> SpiderMonkey doesn't yet implement this syntax: https://bugzilla.mozilla.org/show_bug.cgi?id=1496852

Thanks! Somehow I assumed it had landed already.

PR #2620, opened 2020-05-11
Cover simple declaration of instance fields

adrianheine:
Copied from #2282. Numeric literal as field name wasn't covered so far.


src/class-elements/field-declaration.case
```diff
@@ -0,0 +1,95 @@
+// Copyright (C) 2019 Caio Lima (Igalia SL). All rights reserved.
```
rwaldron, line 1: This is not your name ;P

PR #2625, opened 2020-05-18
Add tests for direct eval calls in strict context

### LINKED ISSUE #2623: Coverage: eval(x) should test for "use strict" depart from the parameter "x".
We found a bug of the Hermes engine. When "use strict" departs from the execution statement "x" and reserved word is used as a variable name, all other engines throw a SyntaxError, but Hermes passed. I didn't see a test262 checking "use strict" outside of "x". Maybe test262 should test for this.

Bug issue from Hermes: https://github.com/facebook/hermes/issues/102

**Testcase**:
```js
function NISLFuzzingFunc(){
	"use strict";
	eval("var public = 1;");
	// when "use strict" is in eval, hermes reports SyntaxError. 
}
NISLFuzzingFunc();

```

leobalter:
Fixes #2623

cc @NWU-NISL

### DIFF
--- a/test/language/eval-code/direct/strict-caller-function-context.js
+++ b/test/language/eval-code/direct/strict-caller-function-context.js
@@ -1,0 +1,17 @@
+// Copyright (C) 2020 Leo Balter. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-performeval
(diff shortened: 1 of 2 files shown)


rwaldron:
I looked on circle ci and this operation finished. Not sure why it's not updated here.

PR #2626, opened 2020-05-18
Fix [[IsHTMLDDA]] Symbol.replaceAll test

shvaikalesh:
Follow-up of #2615.

### DIFF
--- a/test/annexB/built-ins/String/prototype/replaceAll/custom-replacer-emulates-undefined.js
+++ b/test/annexB/built-ins/String/prototype/replaceAll/custom-replacer-emulates-undefined.js
@@ -18,7 +18,7 @@ features: [Symbol.replace, String.prototype.replaceAll, IsHTMLDDA]
 
 var searchValue = $262.IsHTMLDDA;
 var replacerGets = 0;
-Object.defineProperty(searchValue, Symbol.replaceAll, {
+Object.defineProperty(searchValue, Symbol.replace, {
   get: function() {
     replacerGets += 1;
     return searchValue;


test/annexB/built-ins/String/prototype/replaceAll/custom-replacer-emulates-undefined.js
```diff
@@ -18,7 +18,7 @@ features: [Symbol.replace, String.prototype.replaceAll, IsHTMLDDA]
 
 var searchValue = $262.IsHTMLDDA;
 var replacerGets = 0;
-Object.defineProperty(searchValue, Symbol.replaceAll, {
+Object.defineProperty(searchValue, Symbol.replace, {
   get: function() {
     replacerGets += 1;
     return searchValue;
```
leobalter, line 24: this test still lacks the part where the returned value here gets called. 

`2.d.i. Return ? Call(replacer, searchValue, « O, replaceValue »).`

It's good to check for the GetMethod on `@@replace`, but it would be interesting to confirm it calls it with the given args and reuses the given return.

```suggestion
    return function(zero, value) {
      // assert zero
      // assert value
      // assert this
      replacerCalls += 1; // this should be asserted along with replacerGets
    };
```

shvaikalesh: We need to return `[[IsHTMLDDA]]` object here to make sure it is not treated as `null` by `GetMethod`.

JSC code (GetMethod is inlined):

```js
if (!@isUndefinedOrNull(search)) {
    if (@isRegExp(search) && !@stringIncludesInternal.@call(@toString(search.flags), "g"))
        @throwTypeError("String.prototype.replaceAll argument must not be a non-global regular expression");

    var replacer = search.@@replace;
    if (!@isUndefinedOrNull(replacer)) {
        return replacer.@call(search, this, replace);
    }
}
```

We do compare return value to `null`, and I am sure there is coverage for call arguments.

shvaikalesh:
> It might be just a thing with the IsHTMLDDA object.

Seems like `eshost` doesn't properly implement `$262.IsHTMLDDA`: https://github.com/bterlson/eshost/blob/master/runtimes/d8.js#L41.

test/annexB/built-ins/String/prototype/replaceAll/custom-replacer-emulates-undefined.js
rwaldron: > It's good to check for the GetMethod on @@replace, but it would be interesting to confirm it calls it with the given args and reuses the given return.

I suggest letting @shvaikalesh address changes that are not IsHTMLDDA specific in a follow up PR

rwaldron:
The engine failures represent no change from the master branch

rwaldron:
@shvaikalesh 

> Seems like eshost doesn't properly implement $262.IsHTMLDDA for V8.

This is intentional! There is no such thing as `IsHTMLDDA` in V8. IsHTMLDDA should only be implemented by hosts that need to support `document.all` behavior.

shvaikalesh:
@rwaldron

Since V8 is shipping in Chrome, it has to support `document.all`. The question is whether `d8` shell that `eshost` uses has it.

Judging by my experience with JSC, `[[IsHTMLDDA]]` logic is spread out across multiple ops/tiers: disabling it just for `jsc` shell would be very complicated. There is good chance that `d8` may easily expose it (or already does), like [we did for JSC](https://trac.webkit.org/changeset/259587).

It would be nice to display correct results for `[[IsHTMLDDA]]` in https://test262.report.

rwaldron:
> Since V8 is shipping in Chrome, it has to support document.all

That's incorrect. IsHTMLDDA would be tested by a host, eg. Chrome, that supports document.all. When V8 ships a `document.all`, or similar mechanism, then we can provide a non-no-op in eshost. The definition of IsHTMLDDA even states: 

> **`IsHTMLDDA`** - (present only in implementations that can provide it)

No engine is obligated to support IsHTMLDDA, or to even run those tests. 



> It would be nice to display correct results for [[IsHTMLDDA]] in https://test262.report.

As the owner of that project, I'd be more inclined to just omit IsHTMLDDA tests for those engines, since none of them are actually obligated to run them, let alone pass them.

Issue #2629, opened 2020-05-19
Missing tests for Promise combinators


marjakh:
I started looking into which tests exist for Promise.all/allSettled/any/race, and discovered a bunch of missing tests (e.g., a test for Promise.allSettled being there and the corresponding test for Promise.any not being there).

See the (incomplete) table here:
https://docs.google.com/document/d/1lrDTH6UtVj4a_SZcXKBSgknL-wOtDXC5mLQFVcjmgSE/edit?usp=sharing

rwaldron:
Thanks for filing—I'm going to prioritize this for tomorrow.

leobalter:
@marjakh thanks for sharing this.

Coverage for Test262 has always been a challenge as it sources from each step of the specs.

The Built-in APIs are relatively easier to verify the coverage if compared with the syntax grammar. We mostly have all steps from start to end. Promises being settled add a bit of complexity but it's still ok to be listed from a usage lifetime, from call to return + effects.

All the steps might be expanded in several ways. One case is within abstractions from ECMAScript, e.g. `ToLength(x) => ToInteger(x) => ToNumber(x)`. Another example is expanding input values to different types and values (see the different handlings from [ToNumber](https://tc39.es/ecma262/#sec-tonumber) itself. There are other more subtle expansions as well.

All of this are observable steps to the user of the language and we try to cover tests in the most reasonable ways.

What is complex for me here is the fact this document shows coverage not based on these steps in any order, but actually after each file name. I'm pretty sure naming the files after each step is not helpful as this project historically tried this and failed. We constantly change the steps and that naming convention does not support expansions to abstractions very well. We now have a convention, more implicit than anything, to name files after what is being covered. **There is no specific rule that maps these file names the steps being covered**.

Add this to the fact the project is maintained by different people with different perspectives and experiences. We can all try to maintain a consistent naming but it's impossible to achieve perfection. Add to the fact some times we need to apply subjective judgement on how much we expand each coverage. It's impossible to explore every expansion and if we ended up comparing methods tested by different people at different times, we will definitely find the subjectivity on each expansion here.

I'm not complaining, it's valid to flag missing coverage for specific parts. What feels odd to me is the document saying a specific test file name is missing in a different folder.

[...]

PR #2630, opened 2020-05-20
Promise.any: coverage updates

rwaldron:
This is related to #2629. More tests will be added over time.


test/built-ins/Promise/any/ctx-non-ctor.js
```diff
@@ -21,27 +21,3 @@ assert.sameValue(typeof Promise.any, 'function');
 assert.throws(TypeError, function() {
   Promise.any.call(eval);
 });
-
-assert.throws(TypeError, function() {
```
leobalter, line 25: While this change matches files with Promise.all, by convention we've been recently having a single file to non valid constructor types. While this remains correct I don't see any pros with this change here. I would not block it thou. We should avoid expanding every detail to a single file, right?

PR #2640, opened 2020-05-27
Test that yield* return() awaits received value

shvaikalesh:


test/language/statements/async-generator/yield-star-return-missing-value-is-awaited.js
```diff
@@ -0,0 +1,44 @@
...
+}
+
+var asyncIterator = asyncGenerator();
+asyncIterator.next().then(function() {
+  var promise = Promise.resolve(2).then(() => 3);
+  asyncIterator.return(promise).then(function(result) {
+    assert.sameValue(result.value, 3);
+    assert.sameValue(result.done, true);
+  }).then($DONE, $DONE);
+}).catch($DONE);
```
leobalter, line 44:
```suggestion
asyncIterator.next().then(function() {
  var promise = Promise.resolve(2).then(() => 3);
  return asyncIterator.return(promise).then(function(result) {
    assert.sameValue(result.value, 3);
    assert.sameValue(result.done, true);
  });
}).then($DONE, $DONE);
```

I'm not sure that we really need to have a chained `promise`. It feels like a non necessary step, but it doesn't hurt anyway.

The suggestion simplifies the `$DONE` execution to avoid extra ends or multiple calls on failure.

shvaikalesh: Thank you, Leo. This is a nice improvement and the way Promises are authored in the wild.
I guess making tests depend on promise chaining is fine.

PR #2642, opened 2020-05-27
Atomics.waitAsync: BigInt

rwaldron:


test/built-ins/Atomics/waitAsync/bigint/not-a-typedarray-throws.js
```diff
@@ -0,0 +1,39 @@
...
+features: [Atomics.waitAsync, arrow-function, Atomics]
+---*/
+const poisoned = {
+  valueOf() {
+    throw new Test262Error('should not evaluate this code');
+  }
+};
+
+assert.throws(TypeError, () => {
+  Atomics.waitAsync({}, 0, 0n, 0);
```
leobalter, line 34: this test should assert the existence of waitAsync before, otherwise it will pass if it's not implemented.

test/built-ins/Atomics/waitAsync/bigint/not-an-object-throws.js
```diff
@@ -0,0 +1,60 @@
...
+features: [Atomics.waitAsync, Symbol, arrow-function, Atomics]
+---*/
+const poisoned = {
+  valueOf() {
+    throw new Test262Error('should not evaluate this code');
+  }
+};
+
+assert.throws(TypeError, () => {
+  Atomics.waitAsync(null, poisoned, poisoned, poisoned);
```
leobalter, line 34: same here

PR #2643, opened 2020-05-28
Object.getOwnPropertyDescriptor: Improve coverage for primitives

### LINKED ISSUE #2636: Coverage: Function Object.getOwnPropertyDescriptor should convert a non-object parameter to object.
We found a bug of the QuickJS engine. When the first parameter of Object.getOwnPropertyDescriptor is not an object as expected, the JS engine should automatically covert a non-object to an object. However, QuickJS didn’t do that. After careful inspection, I found that test262 has many testcases whose first parameter is non-object such as Null, Undefined, Number, and Boolean, but except a String value. The following is the testcase that I think should be added to the test262 suite.

Testcase:
```
var NISLFuzzingFunc = function(){
	var desc = Object.getOwnPropertyDescriptor("prop", "0");
	print(typeof(desc));
	print(desc.value);
};
NISLFuzzingFunc();
```

shvaikalesh:
Resolves #2636.


test/built-ins/Object/getOwnPropertyDescriptor/length.js
```diff
@@ -0,0 +1,27 @@
+// Copyright (C) 2012 Ecma International. All rights reserved.
```
rwaldron, line 1: Since this is an entirely new file, containing an entirely different mechanism for testing, I'd say it cannot possibly be "Copyright (C) 2012 Ecma International."

test/built-ins/Object/getOwnPropertyDescriptor/primitive-string.js
```diff
@@ -0,0 +1,37 @@
...
+---*/
+
+assert.sameValue(Object.getOwnPropertyDescriptor('', '0'), undefined);
+
+var indexDesc = Object.getOwnPropertyDescriptor('foo', '0');
+
+assert.sameValue(indexDesc.value, 'f', '[[Value]]');
+assert.sameValue(indexDesc.writable, false, '[[Writable]]');
+assert.sameValue(indexDesc.enumerable, true, '[[Enumerable]]');
+assert.sameValue(indexDesc.configurable, false, '[[Configurable]]');
```
rwaldron, line 30: Why not `verifyProperty`? (same for all following tests)

shvaikalesh:
@rwaldron We can't use `verifyProperty` here since it performs `delete primitive[key]` to test configurability and `primitive[key] = 42` to test writability, which are both errors in strict mode.

Apart from that, `verifyProperty` is not documented to work with primitives and has no relevant test coverage, so the way it currently works with primitives is merely an implementation detail.

shvaikalesh:
I didn't rebase the PR properly, so V8 failures are unrelated.

PR #2645, opened 2020-06-02
Update Intl.Segmenter to latest spec

FrankYFTang:
Remove methods removed in the latest reversion.

Still need to add tests for:
1.5.2.1 %SegmentsPrototype%.containing ( index )
1.6.2.1 %SegmentIteratorPrototype%.next ()
1.6.2.2 %SegmentIteratorPrototype% [ @@toStringTag ]

@gibson042  @Ms2ger


FrankYFTang:
@leobalter

test/intl402/Segmenter/prototype/segment/segment-word-iterable.js
```diff
@@ -29,22 +29,12 @@ for (const text of [
...
     ]) {
   let segments = [];
-  // Create another %SegmentIterator% to compare with result from the one that
-  // created in the for of loop.
-  let iter = seg.segment(text);
-  let prev = 0;
   for (const v of seg.segment(text)) {
-    assert(["word", "none"].includes(v.breakType), v.breakType);
+    assert.sameValue("boolean", typeof v.isWordLike);
+    assert([true, false].includes(v.isWordLike), v.isWordLike);
```
caiolima, line 34: If this assert ever executes, `[true, false].includes(v.isWordLike)` will always be `true`, given we asserted above that `typeof v.isWordLike` is `boolean`. I think it is redundant if we assert it here as well. Also, I'm a bit confused why second argument is `v.isWordLike`. IIRC, the signature for this function is `function assert(mustBeTrue, message)` and passing a boolean to `message` seems a bit off.

FrankYFTang: > If this assert ever executes, `[true, false].includes(v.isWordLike)` will always be `true`, given we asserted above that `typeof v.isWordLike` is `boolean`. I think it is redundant if we assert it here as well. Also, I'm a bit confused why second argument is `v.isWordLike`. IIRC, the signature for this function is `function assert(mustBeTrue, message)` and passing a boolean to `message` seems a bit off.

Yes, it is redundant so I will remove it. 
Yes, the second argument is the message, I intend to print out the value in the error but since as you point out it is redundant and I will just remove it now it is not not important.

PR #2654, opened 2020-06-08
Add tests for \8 in template/strict strings.

### LINKED ISSUE #2653: Need tests for \8 in templates and strict strings
Looks like tests for `\8` were missed from https://github.com/tc39/test262/pull/1063.

Everyone agrees that `\08` is bad outside of sloppy non-template strings, but SM, V8, XS, and ChakraCore all accept <code>\`\8\`</code> and `'\8'` regardless of mode.

Templates should hit the NotEscapeSequence early errors in [12.2.9.1](https://tc39.es/ecma262/#sec-static-semantics-template-early-errors) while strict non-templates should fail to lex altogether per the definition of CharacterEscapeSequence in [11.8.4](https://tc39.es/ecma262/#prod-CharacterEscapeSequence).

rkirsling:
Resolves #2653.


leobalter:

leobalter:
This achieved consensus today.

PR #2667, opened 2020-06-22
add a test for `eval?.(code)`

mysticatea:
This PR adds a test for `eval?.(code)`. It's indirect eval.

> ```
>   12.3.9.2 Runtime Semantics: ChainEvaluation
>     OptionalChain: ?. Arguments
>       1. Let thisChain be this OptionalChain.
>       2. Let tailCall be IsInTailPosition(thisChain).
>       3. Return ? EvaluateCall(baseValue, baseReference, Arguments, tailCall).
> ```
> https://tc39.es/ecma262/#sec-optional-chaining-chain-evaluation

### DIFF
--- a/test/language/expressions/optional-chaining/eval-optional-call.js
+++ b/test/language/expressions/optional-chaining/eval-optional-call.js
@@ -1,0 +1,19 @@
+// Copyright 2020 Toru Nagashima.  All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-optional-chaining-chain-evaluation
+description: optional call invoked on eval function should be indirect eval.
+info: |
+  12.3.9.2 Runtime Semantics: ChainEvaluation
+    OptionalChain: ?. Arguments
+      1. Let thisChain be this OptionalChain.
+      2. Let tailCall be IsInTailPosition(thisChain).
+      3. Return ? EvaluateCall(baseValue, baseReference, Arguments, tailCall).
+
+features: [optional-chaining]
+---*/
+
+const a = "global";
+const b = (a => eval?.("a"))("local")
+
+assert.sameValue(b, a);


leobalter:
Calling implementors to check this out: @syg @yulia @anba @kmiller68 @rkirsling.

I hope we can avoid a new de-facto direct eval.

anba:
@mysticatea (thank you!) filed <https://bugzilla.mozilla.org/show_bug.cgi?id=1647169> against SpiderMonkey, which we already fixed today.

syg:
I actually feel like this would be the least confusing if it _were_ a direct eval.

leobalter:
I'm currently reviewing the current tests for Optional Chaining. I can update the tests from indirect to direct eval if we have a matching PR in ECMA-262. Just let me know!

rkirsling:
To echo @syg, I find it surprising (both as a reviewer and as an implementer) that this would be an exception to the "code with `?.` behaves the same as code without it when the LHS is non-nullish" rule of thumb.

https://github.com/tc39/proposal-optional-chaining/issues/21 suggests that this was not an accident, but given that browser-hosted engines all implemented what we expected the spec to say instead of what it actually says, I don't think it's appropriate to resolve this divergence (in either direction) without a proper discussion.

I'm not sure how this discussion should be conducted (or if there's a good precedent for such a situation)—does a GitHub thread suffice or should we have a meeting topic?

leobalter:
> I'm not sure how this discussion should be conducted (or if there's a good precedent for such a situation)—does a spec issue thread suffice or should we add a topic to the next meeting agenda?

I believe so. We will need someone to champion this work at ECMA-262. For this specific change I believe it needs a Normative PR making sure that `eval?.(x)` will trigger `PerformEval` with the direct flag set as `true`.

ljharb:
We can have the discussion without the normative PR, but filing it does accelerate things in the event the change gets consensus.

PR #2668, opened 2020-06-22
use $ERROR instead of throwing in 15.2.3.7-2-18.js

Meir017:


rwaldron:
Can you make sure that you've signed the CLA? Thanks!

Meir017:
@rwaldron I just signed the CLA

test/built-ins/Object/defineProperties/15.2.3.7-2-18.js
```diff
@@ -24,7 +24,9 @@ try {
 
   Object.defineProperties(obj, this);
 } catch (e) {
-  if (!(e instanceof TypeError)) throw e;
+  if (!(e instanceof TypeError)) {
+    $ERROR("Expected TypeError, got " + e);
+  }
   result = true;
 } finally {
   delete this.prop;
```
leobalter, line 32: just re-checking this code, it's fun how this part here is not necessary for the test itself and it seems to be guarding against leaking if tests would be running together.

PR #2671, opened 2020-06-24
replace Test262Error with $ERROR when possible

Meir017:


leobalter:
This is going the opposite path we want for Test262. Ideally the recommended form is throwing a new Test262Error rather than using $ERROR directly. 

The current `$ERROR` usage is within legacy code we'd like to change in the future.

leobalter:
I'd be fine to reopen this if we want to try the opposite approach. For now, I hope we don't go through this current proposed change.

rwaldron:
I was also going to add that this is clearly marked as **DEPRECATED** in https://github.com/tc39/test262/blob/master/CONTRIBUTING.md

leobalter:
To make things clear: #2668 was accepted due to the target test being legacy. The whole test needs improvement but the change just seemed better than the original usage (`$ERROR` is still better than `throw e` in a legacy test).

PR #2666, opened 2020-06-18
Promise.race: coverage updates (ref gh-2629)

rwaldron:

PR #2672, opened 2020-06-24
Promise: whenever evaluating checkSequence(sequence), also assert length of sequence

rwaldron:
This is a follow up from https://github.com/tc39/test262/pull/2666#pullrequestreview-434719626

I branched from the branch in that PR, to avoid making the same changes twice. This will be rebased after that is landed.

### DIFF
--- a/test/built-ins/Promise/allSettled/resolved-sequence-mixed.js
+++ b/test/built-ins/Promise/allSettled/resolved-sequence-mixed.js
@@ -37,22 +37,26 @@ sequence.push(1);
 
 p1.catch(function() {
   sequence.push(3);
+  assert.sameValue(sequence.length, 3);
   checkSequence(sequence, 'Expected to be called first.');
-}).catch($DONE);
+});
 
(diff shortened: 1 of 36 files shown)


test/built-ins/Promise/allSettled/resolved-sequence-mixed.js
```diff
@@ -37,22 +37,26 @@ sequence.push(1);
...
+  assert.sameValue(sequence.length, 3);
   checkSequence(sequence, 'Expected to be called first.');
-}).catch($DONE);
+});
 
 Promise.allSettled([p1, p2, p3]).then(function() {
   sequence.push(6);
+  assert.sameValue(sequence.length, 6);
   checkSequence(sequence, 'Expected to be called fourth.');
 }).then($DONE, $DONE);
```
leobalter, line 48: It is interesting how this last promise catches all the errors in this test. sequence will make sure of the order. Those .catch($DONE) are really ok to be removed.

PR #2676, opened 2020-06-24
Atomics.waitAsync: await $262.agent.getReportAsync() in a loop, instead of calling thrice asynchronously. Fixes #2664

### LINKED ISSUE #2664: built-ins/Atomics/waitAsync/bigint/good-views.js & built-ins/Atomics/waitAsync/good-views.js are incorrect (timing dependent)
In this test, all waitAsyncs resolve immediately; that doesn't create any timing issues. The worker does essentially this:

```
postMessage("message A");
postMessage("message B");
postMessage("message C");
```

The main thread does:

`Promise.all([getReportAsync(), getReportAsync(), getReportAsync()])`. When Promise.all resolves, it asserts that the resulting array is `["message A", "message B", "message C"]`.

Each `getReportAsync` creates a looper which checks if a message has been received, and if not, schedules itself to be re-executed with `setTimeout`: https://github.com/tc39/test262/blob/master/harness/atomicsHelper.js#L57

```
  $262.agent.getReportAsync = function() {
    return new Promise(function(resolve) {
      (function loop() {
        let result = getReport();
        if (!result) {
          setTimeout(loop, 1000);
        } else {
          resolve(result);
        }
      })();
    });
  };
```

Since the main thread does `getReportAsync` 3 times, the task queue will always have 3 tasks, related to 3 different `getReportAsync` calls. At some point, the foremost task will expire and post itself at the back of the queue etc.

Then a message is received. At that point, it's arbitrary which of the 3 loopers happens to own the topmost task of the task queue. That looper will get the message and stop looping.

As a result, the array returned by Promise.all will be an arbitrarily ordered collection of "message A", "message B" and "message C".

This is especially reproducible with d8, since d8 ignores the time parameter of `setTimeout`, spinning the loopers as fast as it can. With this setting, it's especially random which of the loopers happens to own the foremost task when a message arrives. With other js shells, it might not repro so easily; I'd assume the first 3 loops tasks would not find the message, but the next one (belonging to the first `getReportAsync`), executed 1 second after the time start, would.

Suggestions:
-> The test has to be fixed somehow. Probably would be best to separate it into 3 parts which are executed one at a time.
-> If this is how `getReportAsync` is implemented, test262 should allow only one `getReportAsync` to be active at any given time. This would prevent this error from happening again.

rwaldron:


test/built-ins/Atomics/waitAsync/bigint/good-views.js
```diff
@@ -45,11 +44,12 @@ $262.agent.start(`
...
-Promise.all([
-  $262.agent.getReportAsync(),
-  $262.agent.getReportAsync(),
-  $262.agent.getReportAsync(),
-]).then(outcomes => {
+(async () => {
+  const outcomes = [];
+
+  for (let i = 0; i < 3; i++) {
+    outcomes.push(await $262.agent.getReportAsync());
```
leobalter, line 51: _missed an opportunity to use async generators..._

To be honest this looks better anyway.

PR #2685, opened 2020-07-01
Intl.DateTimeFormat.{formatRange,formatRangeToParts} should throw TypeError instead of RangeError if either of parameter is undefined

Constellation:
http://tc39.es/proposal-intl-DateTimeFormat-formatRange/
The spec draft throws TypeError instead of RangeError.

	1.4.5 Intl.DateTimeFormat.prototype.formatRange ( startDate , endDate )
	...
	4. If startDate is undefined or endDate is undefined, throw a TypeError exception.

	1.4.6 Intl.DateTimeFormat.prototype.formatRangeToParts ( startDate , endDate )
	...
	4. If startDate is undefined or endDate is undefined, throw a TypeError exception.

### DIFF
--- a/test/intl402/DateTimeFormat/prototype/formatRange/date-undefined-throws.js
+++ b/test/intl402/DateTimeFormat/prototype/formatRange/date-undefined-throws.js
@@ -3,14 +3,14 @@
 
 /*---
 description: >
-  Throws a RangeError if startDate or endDate is undefined.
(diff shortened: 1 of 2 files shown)


sffc:
CC @fabalbon @littledan @leobalter

Constellation:
Thanks!!!!!

PR #2691, opened 2020-07-07
Promise.prototype.finally: Fix unhandled rejections

shvaikalesh:
Without this change, the tests pass on JSC's `test262/Runner.pm`, while failing on [`test262.report`](https://test262.report/browse/built-ins/Promise/prototype/finally?engines=javascriptcore).


test/built-ins/Promise/prototype/finally/species-symbol.js
```diff
@@ -1,22 +0,0 @@
-// Copyright (C) 2017 V8. All rights reserved.
-// This code is governed by the BSD license found in the LICENSE file.
```
ljharb, line 2: why is this deleted?

leobalter: This test seems like a false positive. It doesn't seem like the assertions are relying on the the Promise subclass at all, but just returning what finally returns.

I wonder if the goal was checking something like `.finally.call(MyPromise, () => {})` or the **resolved** value of the finally call.

For the current text, I don't see why `p instanceof MyPromise` would ever be true and that's why the info metatag is ever so important in tests like this.

test/built-ins/Promise/prototype/finally/subclass-reject-count.js
```diff
@@ -16,7 +16,8 @@ class FooPromise extends Promise {
...
 
-FooPromise.reject().finally(() => {}).then($ERROR).catch(() => {
-  assert.sameValue(7, count);
-  $DONE();
-});
+FooPromise.reject().finally(() => {}).then(value => {
+  throw new Test262Error("Expected Promise to be rejected, got: resolved with " + value);
+}, () => {
+  assert.sameValue(count, 7);
+}).then($DONE, $DONE);
```
leobalter, line 23: This is interesting, it seems like in the unchanged version we would not capture the `$DONE` for the async execution of the test properly. If the assertion failed at line 20 (`count === 7`) we would never reach $DONE and therefore we'd need to wait for a timeout. This format captures the end correctly if the test fails or pass.

leobalter:
cc @rkirsling @kmiller68 these remain failing on JSC.

test/built-ins/Promise/prototype/finally/species-symbol.js
shvaikalesh: @ljharb This test is great, but it is a 1:1 duplicate of [`subclass-species-constructor-resolve-count.js`](https://github.com/tc39/test262/blob/master/test/built-ins/Promise/prototype/finally/subclass-species-constructor-resolve-count.js).

leobalter: I opened #2693 so we can move this one forward.

Issue #2504, opened 2020-02-18
Computed Property Name and property ordering in tests


scole66:
Computed Property Name tests accidentally test for Class Definition evaluation order of operations, and as such many JS implementations fail to pass these three tests.

Not only that, but the order tested for doesn't match the order specified in the ecma-262 specification anyway.

Relevant tests are:

* test/language/computed-property-names/class/static/method-number.js
* test/language/computed-property-names/class/static/method-string.js
* test/language/computed-property-names/class/static/method-symbol.js

A sample of the problem is the final assertion of each test. The one from `method-number` looks like:

```javascript
assert(
  compareArray(Object.getOwnPropertyNames(C), ['1', '2', 'length', 'name', 'prototype', 'a', 'c']),
  "`compareArray(Object.getOwnPropertyNames(C), ['1', '2', 'length', 'name', 'prototype', 'a', 'c'])` returns `true`"
);
```

The problem is that `getOwnPropertyNames` returns its not-numeric, not-symbol property names in "order of creation". If we look at section 14.6.13 (ClassDefinitionEvaluation), we see `length` getting added in step 12; `prototype` getting added in step 16, and `name` getting created in step 18.a.
So an ecmascript implementation based solely off the spec cannot pass this test. (The order there would be `length`, `prototype`, `name`.)

But the different JS engines I checked all do it differently anyway. The pass rate for these tests on test-262-reports is a measly 50%. I really think these tests need to step away from creation order entirely (browsers are failing for reasons that have nothing to do with "can a number be a property name").

There are a lot of ways to make this better. We could treat the result as a Set, rather than an array, and then compare the Sets. We could sort the result of `getOwnPropertyNames` lexically, and get rid of order-of-creation as an artifact entirely. We could probe the properties by hand and merely check for existence.

ljharb:
I'm not sure it's accidental; computed property names seem like the only way to test ordering. (In 2015, the committee agreed on this ordering independent of Class Fields, *because* of class fields and decorators - to coordinate things between them)

scole66:
Hm. Take out `length`, `name`, and `prototype` then? Changing the comparison to match the order in the specification seems wrong; all the engines do things in a different order

ljharb:
The spec requires they do them in the same order, however, so I think it's important that the test fails for them.

scole66:
[...]

PR #2696, opened 2020-07-09
Refactor tests for static methods creation and fetching

leobalter:
Ref #2504
Ref https://github.com/tc39/ecma262/pull/1490

@scole66 The tests are now correct with the current specs since https://github.com/tc39/ecma262/pull/1490 has been merged.

There wasn't much to change I added some info content to help clarifying how the properties are set during the class evaluation and fetched from `Object.getOwnPropertyNames` and `Object.getOwnPropertySymbols`.

I also removed the assert messages as `compareArray` does a good job showing the actual and expected values if anything fails.

For the current tests failures, let's ping @codehag @anba @syg.

### DIFF
--- a/test/language/computed-property-names/class/static/method-number.js
+++ b/test/language/computed-property-names/class/static/method-number.js
@@ -1,11 +1,61 @@
 // Copyright (C) 2014 the V8 project authors. All rights reserved.
 // This code is governed by the BSD license found in the LICENSE file.
 /*---
-esid: sec-object-initializer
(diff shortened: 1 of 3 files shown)


anba:
SpiderMonkey test failures for these files are tracked at <https://bugzilla.mozilla.org/show_bug.cgi?id=1629803>.

leobalter:
Thank you, @anba, I value this information because sometime it's a bit hard to map failing tests to existing filed bugs, if any.

rwaldron:
@syg heads for failures in V8

PR #2697, opened 2020-07-10
Remove 0 from the range of fractionalSecondDigits 

FrankYFTang:
Update the test to sync with the least version of https://github.com/tc39/ecma402/pull/347
and remove 0 from the range of fractionalSecondDigits and just use undefined for that.

### DIFF
--- a/test/intl402/DateTimeFormat/constructor-options-fractionalSecondDigits-valid.js
+++ b/test/intl402/DateTimeFormat/constructor-options-fractionalSecondDigits-valid.js
@@ -14,9 +14,6 @@ features: [Intl.DateTimeFormat-fractionalSecondDigits]
 
 const validOptions = [
   [undefined, undefined],
-  [-0, undefined],
-  [0, undefined],
-  ["0", undefined],
   [1, 1],
   ["1", 1],
   [2, 2],
@@ -25,9 +22,8 @@ const validOptions = [
   ["3", 3],
   [2.9, 2],
   ["2.9", 2],
-  [0.00001, undefined],
+  [1.00001, 1],
   [{ toString() { return "3"; } }, 3],
-  [{ valueOf() { return -0; }, toString: undefined }, undefined],
 ];
 for (const [fractionalSecondDigits, expected] of validOptions) {
   const dtf = new Intl.DateTimeFormat("en", { fractionalSecondDigits });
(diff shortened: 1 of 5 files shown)


FrankYFTang:
@anba @zbraniecki @leobalter @rwaldron @gibson042

FrankYFTang:
@syg v8 fix is in https://chromium-review.googlesource.com/c/v8/v8/+/2290627

yamass:
@FrankYFTang 
I am super curious why this change was necessary. I don't see the reason why specifying `fractionalSecondDigits: 0` would do any harm. 
For example, Chrome will throw and exception when doing so. As a user of the API, I am very surprised about this behavior.

Try this in Chrome:
```
new Intl.DateTimeFormat("en", {fractionalSecondDigits: 0}) // Will throw Uncaught RangeError: fractionalSecondDigits value is out of range.
```

I was unable to find any good explanation for this change. The only things I found were your comments
https://github.com/tc39/ecma402/pull/347#issuecomment-636236094
https://github.com/tc39/ecma402/pull/347#issuecomment-618065161
but these are only regarding the behavior of resolvedOptions()

Please could you or anybody else give me a hint?

FrankYFTang:
here is the reason, it is related to resolvedOptions(). For argument sake, let' say we allow 
df1 = new Intl.DateTimeFormat("en", {fractionalSecondDigits: 0})
df2 = new Intl.DateTimeFormat("en")

both not throwing exception
now, both df1 and df2 will produce the same result while calling .format()
but what would df1.resolvedOption().fractionalSecondDigits and df2.resolvedOption().fractionalSecondDigits return?

df2.resolvedOption().fractionalSecondDigits should return undefined, right?
what should 
df1.resolvedOption().fractionalSecondDigits return? According to the rest of the spec, it need to return 0. 

However, in order to make it return undefined and 0 depending on the two different input, the implementation need to track the value and cannot depend on the required runtime information to do so. Which make the implement use more memory without providing any more functionality than what could already be provided with 
new Intl.DateTimeFormat("en") 
or
new Intl.DateTimeFormat("en", {fractionalSecondDigits: undefined}) 

Therefore, it is not needed to make the functionality better, but it is needed to make the implementation to delivery the feature w/ efficient usage of memory w/o reducing any exposed features.

PR #2698, opened 2020-07-11
Intl.DisplayNames make type required

FrankYFTang:
Update the tests to the latest version after we make
the type option now required

Spec:
https://tc39.es/proposal-intl-displaynames/

Changes of spec:
https://github.com/tc39/proposal-intl-displaynames/pull/76
https://github.com/tc39/proposal-intl-displaynames/pull/83

@anba @Ms2ger @rwaldron @zbraniecki @littledan @rkirsling @syg @leobalter

### DIFF
--- a/harness/testIntl.js
+++ b/harness/testIntl.js
@@ -40,7 +40,8 @@ function testWithIntlConstructors(f) {
 
   // Optionally supported Intl constructors.
   // NB: Intl.Locale isn't an Intl service constructor!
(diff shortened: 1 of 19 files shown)


leobalter:
@FrankYFTang we should have coverage verifying the behavior if the type options is not provided or invalid.

FrankYFTang:
> @FrankYFTang we should have coverage verifying the behavior if the type options is not provided or invalid.

Done

PR #2701, opened 2020-07-13
Remove unnecessary files containing whitelist and blacklist

rwaldron:


leobalter:
can `implementation-contributed/v8/mjsunit/es6/unscopables.js` also be removed?

PR #2702, opened 2020-07-15
Test [[IsHTMLDDA]] object as "prototype" of superclass

shvaikalesh:
Follow-up of #2579.
See also #2652.

### DIFF
--- a/test/annexB/language/statements/class/subclass/superclass-prototype-emulates-undefined.js
+++ b/test/annexB/language/statements/class/subclass/superclass-prototype-emulates-undefined.js
@@ -1,0 +1,29 @@
+// Copyright (C) 2020 Alexey Shvayka. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-runtime-semantics-classdefinitionevaluation
+description: >
+  [[IsHTMLDDA]] object as "prototype" of superclass: `null` check uses strict equality.
+info: |
+  ClassDefinitionEvaluation
+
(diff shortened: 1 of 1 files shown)


leobalter:

leobalter:
Interesting, somehow merging this PR did not trigger the GitHub workflow. I wonder if this because the pull request was originally open against another branch, but still sounds weird.

shvaikalesh:
@leobalter Thank you for asking!

For `Type(protoParent) is neither Object nor Null` prose, JSC [emits](https://github.com/WebKit/webkit/blob/79e9a9ebd57865e1ed894be711961aa338ec4054/Source/JavaScriptCore/bytecompiler/NodesCodegen.cpp#L4909-L4911) the same bytecode as for `typeof protoParent === "object" || typeof protoParent === "function"`.

Introducing new bytecode requires implementing it for all 4 tiers (+ slow path + 32-bit support). We try to avoid introducing new bytecode ops unless absolutely necessary, keeping their count < 256.

But why does the test pass if `typeof document.all === "undefined"`? There is a bug affecting interpreter and baseline JIT (first 2 tiers). When 3rd tier kicks in, there is an error (reproducible in Safari 13.1):

```js
function Superclass() {}
Superclass.prototype = document.all;

for (let i = 0; i < 1e6; ++i) {
  class C extends Superclass {}
  // throws TypeError: The value of the superclass's prototype property is not an object.
}
```

leobalter:
Thank you! I really appreciate the extra info!

PR #2706, opened 2020-07-17
Test [[IsHTMLDDA]] object with `typeof` operator

shvaikalesh:
Follow-up of #2579.

### DIFF
--- a/test/annexB/language/expressions/typeof/emulates-undefined.js
+++ b/test/annexB/language/expressions/typeof/emulates-undefined.js
@@ -1,0 +1,27 @@
+// Copyright (C) 2020 Alexey Shvayka. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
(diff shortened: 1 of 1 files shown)


leobalter:
I'm curious why most engines are failing. This may be related to copying the value from `$262.IsHTMLDDA` ?

shvaikalesh:
@leobalter CI results for `[[IsHTMLDDA]]` objects are unreliable because [`bterlson/eshost`](https://github.com/bterlson/eshost) defines regular function stub for most runtimes ([e.g. `d8`](https://github.com/bterlson/eshost/blob/efd1599c87ccd1e319a4559d1ce0068584856afa/runtimes/d8.js#L41)).

For JSC, test result is correct (thanks to https://github.com/bterlson/eshost/commit/ff43f5d7ea1db5072a818c9b67fb5d634df13bec): `typeof document.all === 'function'` and `typeof document.all === 'undefined'` are both `true` because of a bug in peephole optimization.

PR #2707, opened 2020-07-18
Update tests for tc39/ecma402#471

ryzokuken:
The other test that was affected by the change was one of V8's local tests.

Refs: https://github.com/tc39/ecma402/pull/471

/cc @littledan @sffc


sffc:
lgtm

PR #2711, opened 2020-07-20
Intl.DisplayNames.length should be 2

Constellation:
`locales` and `options` are not optional parameters. So the length should be 2.


Constellation:
CC @anba @Ms2ger @rwaldron @zbraniecki @littledan @rkirsling @syg @leobalter @FrankYFTang

FrankYFTang:
FYI- v8 track in https://bugs.chromium.org/p/v8/issues/detail?id=10733

PR #2713, opened 2020-07-21
Add feature tag for cleanupSome

leobalter:
The WeakRefs proposal advanced to Stage 4 but cleanupSome moved back to a Stage 2 status.

The tests should remain in Test262 for now but this feature flag should help users to filter tests out.

### DIFF
--- a/features.txt
+++ b/features.txt
@@ -133,6 +133,10 @@ import.meta
 WeakRef
 FinalizationRegistry
 
+# FinalizationRegistry#cleanupSome
+# link pending
+cleanupSome
+
 # Optional Chaining
 # https://github.com/tc39/proposal-optional-chaining
 optional-chaining
--- a/harness/async-gc.js
+++ b/harness/async-gc.js
@@ -3,9 +3,8 @@
 /*---
 description: >
     Collection of functions used to capture references cleanup from garbage collectors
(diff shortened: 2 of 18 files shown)


littledan:
+1 on this patch. I want to suggest that test262 maintain a mapping from feature flags to stages. That way, engines can automatically filter out things that are pre-Stage 3 (without needing manual review). It will be easy to filter out cleanupSome, but if we get more Stage 2 features with test262 tests, it could create a bunch of busy-work for engines.

leobalter:
> I want to suggest that test262 maintain a mapping from feature flags to stages. 

I'm not a big fan into adding extra info in the tests for this. We do have the features.txt file that does some conventional mapping to _recent_ proposals. I wish we could work in something broader that helps building both the proposals list (in the proposals repo) that could connect with the features.txt here.

ljharb:
Presumably we could come up with a way to programmatically derive stage and a feature flag name from a proposal in https://github.com/tc39/proposals, and then it wouldn't require manual maintenance.

littledan:
Don't take my suggestion as a precondition for landing this PR. I don't want to add more maintenance burden for people.

leobalter:
sure thing, but I like a further discussion with this new topic. There might be a good solution in the end.

PR #2714, opened 2020-07-23
Fix Segmenter tests: Check "input" in segment-*-iterable.js

FrankYFTang:
Sync with the Jul 21 2020 spec which promoted to Stage 3
https://tc39.es/proposal-intl-segmenter/

Remove dup test in segment-*.js
fix segment-tostring.js
@gibson042  @leobalter @rwaldron @anba @zbraniecki @littledan @Ms2ger 

We should add test to test containing(index)


ryzokuken:
@leobalter can we merge this in please?

leobalter:
> @leobalter can we merge this in please?

sure! I was just waiting for another reviewer, and this is not resolved! Thanks!

PR #2717, opened 2020-07-26
Intl.Segmenter iterator result should have isWordLike only if type is "word"

Constellation:
Update according to https://github.com/tc39/proposal-intl-segmenter/pull/128.


Constellation:
/CC @gibson042 @leobalter @rwaldron @anba @zbraniecki @littledan @Ms2ger @FrankYFTang

ryzokuken:
@rwaldron @leobalter can we merge this in?

Constellation:
Thanks so much!!

PR #2718, opened 2020-07-29
related-year-zh.js may not contain years

Constellation:
Apple ICU team is shipping macOS system ICU with newer CLDR, but it has many overrides on the top of it to make the formatted output suitable for the rest of the system (entire macOS components are using this system ICU).
So this is not using raw CLDR data.

In related-year-zh.js tests, Apple ICU intentionally overrides the CLDR data with the different format.
This change modifies the test to accept that alternative output.

I think this is OK since test262 is not a test for CLDR/ICU and the different output is OK so long as it is intentional :)
And formatToParts is showing right results for the generated output.

### DIFF
--- a/test/intl402/DateTimeFormat/prototype/format/related-year-zh.js
+++ b/test/intl402/DateTimeFormat/prototype/format/related-year-zh.js
@@ -1,4 +1,5 @@
 // Copyright 2019 Google Inc, Igalia S.L. All rights reserved.
+// Copyright 2020 Apple Inc. All rights reserved.
 // This code is governed by the BSD license found in the LICENSE file.
 
 /*---
@@ -11,4 +12,6 @@ locale: [zh-u-ca-chinese]
 
 const df = new Intl.DateTimeFormat("zh-u-ca-chinese", {year: "numeric"});
 const date = new Date(2019, 5, 1);
-assert.sameValue(df.format(date), "2019己亥年");
+const formatted = df.format(date);
+const expected = ["2019己亥年", "己亥年"];
+assert(expected.includes(formatted));
--- a/test/intl402/DateTimeFormat/prototype/formatToParts/related-year-zh.js
+++ b/test/intl402/DateTimeFormat/prototype/formatToParts/related-year-zh.js
@@ -1,4 +1,5 @@
 // Copyright 2019 Google Inc, Igalia S.L. All rights reserved.
+// Copyright 2020 Apple Inc. All rights reserved.
 // This code is governed by the BSD license found in the LICENSE file.
 
 /*---
@@ -9,21 +10,24 @@ description: >
 locale: [zh-u-ca-chinese]
 ---*/
 
-function verifyFormatParts(actual, expected, message) {
-  assert.sameValue(Array.isArray(expected), true, `${message}: expected is Array`);
-  assert.sameValue(Array.isArray(actual), true, `${message}: actual is Array`);
-  assert.sameValue(actual.length, expected.length, `${message}: length`);
-
-  for (let i = 0; i < actual.length; ++i) {
-    assert.sameValue(actual[i].type, expected[i].type, `${message}: parts[${i}].type`);
-    assert.sameValue(actual[i].value, expected[i].value, `${message}: parts[${i}].value`);
-  }
-}
-
 const df = new Intl.DateTimeFormat("zh-u-ca-chinese", {year: "numeric"});
 const date = new Date(2019, 5, 1);
-verifyFormatParts(df.formatToParts(date), [
+const actual = df.formatToParts(date);
+
+const expected = [
   {type: "relatedYear", value: "2019"},
   {type: "yearName", value: "己亥"},
   {type: "literal", value: "年"},
-]);
+];
+
+assert.sameValue(Array.isArray(actual), true, 'actual is Array');
+
+if (actual.length <= 2) {
+  expected.shift(); // removes the relatedYear
+}
+
+actual.forEach(({ type, value }, i) => {
+  const { type: eType, value: eValue } = expected[i];
+  assert.sameValue(type, eType, `actual[${i}].type should be ${eType}`);
+  assert.sameValue(value, eValue, `actual[${i}].value should be ${eValue}`);
+});


Constellation:
Radar for Apple readers <rdar://problem/61842636> (where JavaScriptCore team and Apple ICU team discussed about it).
/CC @rkirsling who found this in WebKit.

Constellation:
And CC Apple ICU folk :) @pedberg-icu

test/intl402/DateTimeFormat/prototype/formatToParts/related-year-zh.js
```diff
@@ -12,18 +13,25 @@ locale: [zh-u-ca-chinese]
...
   {type: "relatedYear", value: "2019"},
   {type: "yearName", value: "己亥"},
   {type: "literal", value: "年"},
-]);
+];
+const expected2 = [
+  {type: "yearName", value: "己亥"},
+  {type: "literal", value: "年"},
+];
+assert.sameValue(verifyFormatParts(df.formatToParts(date), expected1) || verifyFormatParts(df.formatToParts(date), expected2), true);
```
leobalter, line 37: This || breaks a bit of the test determinism and we can try to move this branching to a lower layer here. The test seems correct, but let's see if we can reduce this.

test/intl402/DateTimeFormat/prototype/format/related-year-zh.js
```diff
@@ -11,4 +12,5 @@ locale: [zh-u-ca-chinese]
 
 const df = new Intl.DateTimeFormat("zh-u-ca-chinese", {year: "numeric"});
 const date = new Date(2019, 5, 1);
-assert.sameValue(df.format(date), "2019己亥年");
+const formatted = df.format(date);
+assert.sameValue(formatted === "2019己亥年" || formatted === "己亥年", true);
```
leobalter, line 16:
```suggestion
const expected = ["2019己亥年", "己亥年"];
assert(expected.includes(formatted));
```

test/intl402/DateTimeFormat/prototype/formatToParts/related-year-zh.js
leobalter, line 37: We can remove the whole `verifyFormatParts` function, and then:

```suggestion
const actual = df.formatToParts(date);

assert.sameValue(Array.isArray(actual), true, 'actual is Array');

if (actual.length > 2) {
  expected.shift(); // removes the relatedYear
}

actual.forEach(({ type, value }, i) => {
  const { type: eType, value: eValue } = expected[i];
  assert.sameValue(type, eType, `actual[${i}].type should be ${eType}`);
  assert.sameValue(value, eValue, `actual[${i}].value should be ${eValue}`);
}); 
```

test/intl402/DateTimeFormat/prototype/formatToParts/related-year-zh.js
```diff
@@ -12,18 +13,25 @@ locale: [zh-u-ca-chinese]
...
+  return true;
 }
 
 const df = new Intl.DateTimeFormat("zh-u-ca-chinese", {year: "numeric"});
 const date = new Date(2019, 5, 1);
-verifyFormatParts(df.formatToParts(date), [
+const expected1 = [
   {type: "relatedYear", value: "2019"},
   {type: "yearName", value: "己亥"},
   {type: "literal", value: "年"},
```
leobalter, line 31:
```suggestion
const expected = [
  {type: "relatedYear", value: "2019"},
  {type: "yearName", value: "己亥"},
  {type: "literal", value: "年"},
```

test/intl402/DateTimeFormat/prototype/formatToParts/related-year-zh.js
```diff
@@ -12,18 +13,25 @@ locale: [zh-u-ca-chinese]
 function verifyFormatParts(actual, expected, message) {
```
leobalter, line 13: it's fun how `message` was never used here.

leobalter:
Tests are correct, I just suggested some _ready to pick_ nits.

If you prefer I can do these changes in a follow up but I believe the changes improve the test readability and the array "shift" is set only for the non-deterministic point.

leobalter:
cc @gibson042 @FrankYFTang @sffc @jswalden

test/intl402/DateTimeFormat/prototype/format/related-year-zh.js
Constellation: Nice! I'll fix it.

test/intl402/DateTimeFormat/prototype/formatToParts/related-year-zh.js
Constellation: Sounds pretty nice. Fixing.

leobalter: if you prefer (but not limited to) you can apply these changes through the GitHub interface. I should probably squash them at the end.

test/intl402/DateTimeFormat/prototype/formatToParts/related-year-zh.js
Constellation: Lol, right, we should just remove it.

Constellation:
@leobalter OK! I've revised and updated the commit :) Thanks!

test/intl402/DateTimeFormat/prototype/formatToParts/related-year-zh.js
Constellation: I've updated the PR by squashing changes :)

Constellation: Fixed! Thanks

test/intl402/DateTimeFormat/prototype/formatToParts/related-year-zh.js
Constellation: Nice, changed!

FrankYFTang:
The change make senes to me.

leobalter:
Thanks!

Constellation:
Thanks all!!!

Issue #2478, opened 2020-01-21
Missing coverage: Redeclaration of "arguments" in direct eval in parameter expressions


anba:
There doesn't seem to be coverage for redeclaring "arguments" in direct eval in function parameter expressions. For example:
```js
function f(p = eval("var arguments")) {}
f();
```

should throw a SyntaxError when evaluating [EvalDeclarationInstantiation](https://tc39.es/ecma262/#sec-evaldeclarationinstantiation) in step 5.d.ii.2.a.i.

I've written some basic tests at <https://phabricator.services.mozilla.com/D60508>, but those could probably be expanded through the test generation python script.

PR #2720, opened 2020-07-30
Redeclaration of arguments in direct eval in parameter expressions, ref gh-2478

rwaldron:
@anba I used your report and spidermonkey tests (from here https://github.com/tc39/test262/issues/2478) as the basis for the test cases & templates. Can you take a look?

### DIFF
--- a/src/direct-eval-code/default/func-decl/a-following-parameter-is-named-arguments.template
+++ b/src/direct-eval-code/default/func-decl/a-following-parameter-is-named-arguments.template
@@ -1,0 +1,19 @@
+// Copyright (C) 2020 Rick Waldron, André Bargull. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+path: language/eval-code/direct/func-decl-a-following-parameter-is-named-arguments-
+name: Declare |arguments| when a following parameter is named |arguments|.
+esid: sec-evaldeclarationinstantiation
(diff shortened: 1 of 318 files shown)


src/direct-eval-code/default/func-decl/a-following-parameter-is-named-arguments.template
```diff
@@ -0,0 +1,19 @@
...
+
+/*---
+path: language/eval-code/direct/func-decl-a-following-parameter-is-named-arguments-
+name: Declare |arguments| when a following parameter is named |arguments|.
+esid: sec-evaldeclarationinstantiation
+flags: [noStrict]
+---*/
+
+
+assert.sameValue("arguments" in this, false, "No global 'arguments' binding");
```
leobalter, line 12:
```suggestion
assert.sameValue("arguments" in this, false, "No global 'arguments' binding before eval");
```

leobalter: this applies to the other matching templates (40 ish of them)

rwaldron: Those assertions are gone because it erroneously penalizes engines for having a global `arguments` object, when there is no rule against that.

rwaldron:
@syg heads up for new failures in V8

PR #2722, opened 2020-08-01
Add tests for \9 in template/strict strings.

rkirsling:
Supplement to #2654, per @syg's request.

I just included tests for \8 in that PR but we really should check \9 as well for completeness.


gibson042:
Should this also add tests for noctals in tagged templates? SpiderMonkey and V8 currently evaluate ```((strings => strings[0])`\9!`)``` as `"9!"` rather than the [specified](https://tc39.es/ecma262/#sec-static-semantics-tv-and-trv) `undefined`.

rkirsling:
Oh, you're right, we should add that! I hadn't done so because the early errors (pertinent to the associated normative PR) only apply to untagged templates, but TV is an existing matter that isn't being tested/respected.

rwaldron:

PR #2723, opened 2020-08-05
Add tests for ZWJ and ZWNJ in identifiers

akluball:
It's possible I missed it, but I couldn't find tests for zwj/zwnj in identifiers. This adds negative tests for zwj/zwnj as identifier start and positive tests for zwj/zwnj as an identifier part (https://tc39.es/ecma262/#sec-names-and-keywords).


akluball:
I think https://github.com/engine262/engine262/pull/119 should make engine262 tests pass

leobalter:
Thanks, @akluball! Can you sign the [Test262 CLA](http://tc39.github.io/test262-cla), please?

test/language/identifiers/part-zwj-escaped.js
```diff
@@ -0,0 +1,10 @@
+// Copyright (C) 2020 Adam Kluball. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+author: Adam Kluball
+esid: sec-names-and-keywords
+description: zero width joiner is a valid identifier part
+---*/
+
+var $\u200D;
```
leobalter, line 10: I believe for this specific case we need to improve this test a little bit to avoid false positives, such as an implementation taking `\u200D` and `\u200C` as white space, supposedly. 

Maybe this can be done w/ a new file leaving the current one as is, or just blending this file with the `test/language/identifiers/part-zwnj-escaped.js`. I prefer the former option, creating a new test file to distinguish those identifiers.

```suggestion
var $ = 1;
var $\u200D = 2;
var $\u200C = 3;

assert.sameValue($, 1);
assert.sameValue($\u200D, 2);
assert.sameValue($\u200C, 3);
```

akluball: Added an assingment test. It seems like it makes the other tests redundant. I could remove them and rename the assignment test to part-zwj-zwnj-unescaped.js.

leobalter: Thanks! I believe it's fine either way. Just let me know when it's ready.

akluball: It's ready

PR #2725, opened 2020-08-10
Test [[IsHTMLDDA]] object with equality operations

### LINKED ISSUE #1372: Coverage area: document.all
V8's test262 test harness doesn't yet expose anything for document.all-like objects. The only test262 tests I see failing due to that omission are related to iteration. I would've expected there to be some additional tests failing, such as ensuring that the document.all-like object will be `===` to neither `null` nor `undefined` (which would fail for us, as it is just `undefined`). There's a bunch of places in the spec where SameValue(undefined) comes up; a few more of these could be tested.

shvaikalesh:
Closes #1372.
Related: #2726, #2706.

Tested manually in FF 81 and Chrome 86: all tests pass.

### DIFF
--- a/test/annexB/built-ins/Object/is/emulates-undefined.js
+++ b/test/annexB/built-ins/Object/is/emulates-undefined.js
@@ -1,0 +1,26 @@
+// Copyright (C) 2020 Alexey Shvayka. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-object.is
+description: >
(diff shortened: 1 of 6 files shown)


rwaldron:
@syg heads up for failures in V8 (althought, that might not be an issue for your local test runner since you have https://github.com/v8/v8/blob/master/test/test262/harness-ishtmldda.js ... can we expose this in a way that allows `eshost` to include it in V8 runtime?)

PR #2734, opened 2020-08-20
Test the sequence of potentially abrupt steps in String.prototype.split

gibson042:


test/built-ins/String/prototype/split/limit-touint32-error.js
```diff
@@ -0,0 +1,22 @@
...
+info: |
+  1. If _limit_ is *undefined*, let _lim_ be 2<sup>32</sup> - 1; else let _lim_ be ? ToUint32(_limit_).
+  1. Let _R_ be ? ToString(_separator_).
+  1. If _lim_ = 0, return _A_.
+---*/
+
+var nonStringableSeparator = {};
+nonStringableSeparator[Symbol.toPrimitive] = $DONOTEVALUATE;
+nonStringableSeparator.toString = $DONOTEVALUATE;
+nonStringableSeparator.valueOf = $DONOTEVALUATE;
```
rwaldron, line 15: `$DONOTEVALUATE` must only be used in tests that are explicitly a negative, parse phase Syntax Error test. The definition of `$DONOTEVALUATE` is allowed to be replaced by an engine's own runner, and that definition may not be what you want for these tests. 

For tests like this, just create your own function that throws whatever you want it to throw. If you don't have time or bandwidth, I will happily make the change for you

rwaldron: @gibson042 I also want to acknowledge that this was never explicitly stated in the documentation for test contributing or interpreting and I apologize for that. I've updated the docs and I'm awaiting @leobalter's review.

gibson042: Thanks for the explanation. I've refactored to avoid violating such expectations. Should I also go a step further and generalize ExpectedError into the test harness and CONTRIBUTING.md?

harness/assert.js
```diff
@@ -84,7 +84,7 @@ assert.throws = function (expectedErrorConstructor, func, message) {
       message += 'Thrown value was not an object!';
       $ERROR(message);
     } else if (thrown.constructor !== expectedErrorConstructor) {
-      message += 'Expected a ' + expectedErrorConstructor.name + ' but got a ' + thrown.constructor.name;
+      message += 'Expected a ' + expectedErrorConstructor.name + ' but got a ' + thrown.constructor.name + ' ' + thrown.message;
```
leobalter, line 87: I might be wrong but IIRC we've had a reason to not include message here.

Anyway, the fact test runners might try to parse this line for the whatever reason, I'm not sure if this would require a new version. One quick workaround would be putting this in a new line and handle if message is not undefined.

This is probably too much of a nit pick.

test/built-ins/String/prototype/split/this-value-tostring-error.js
```diff
@@ -0,0 +1,45 @@
...
+  this.message = message || "";
+}
+ExpectedError.prototype.toString = function () {
+  return "ExpectedError: " + this.message;
+};
+
+var split = String.prototype.split;
+
+var nonStringableReceiver = {};
+nonStringableReceiver.toString = function() { throw new ExpectedError("receiver.toString"); };
```
leobalter, line 25:
```suggestion
var receiverToString = 0;
nonStringableReceiver.toString = function() {
  receiverToString += 1;
  throw new ExpectedError("receiver.toString");
};
```

and after the `split.call(nonStringableReceiver, splitter` line:

```js
assert.sameValue(receiverToString, 0, 'ToString should not be called on the receiver when the separator has a @@split method.');
```

and at the end, after the last assert.throws, you can check `receiverToString` is 1, without errors being swallowed or re-checked.

harness/assert.js
gibson042: Hmm, so should I revert this change? It made confirmation better for me, but I'm not married to it.

test/built-ins/String/prototype/split/this-value-tostring-error.js
```diff
@@ -0,0 +1,45 @@
...
+
+var splitter = {};
+splitter[Symbol.split] = function() {};
+
+try {
+  split.call(nonStringableReceiver, splitter, Symbol());
+} catch (e) {
+  assert.sameValue(e, undefined,
+      'ToString should not be called on the receiver when the separator has a @@split method.');
+}
```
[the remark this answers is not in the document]
gibson042: The actual invocation of a separator with Symbol.split is tested elsewhere and I think this output looks better with try-catch and minimal assertions; do you still want this change?

test/built-ins/String/prototype/split/this-value-tostring-error.js
gibson042: I'm not sure what this suggestion is getting at. The way I have things now, the only source of an ExpectedError is from ToString on the receiver, which _must not_ take place as part of `split.call(nonStringableReceiver, splitter, Symbol())` (because the Symbol.split path is special) and _must_ take place as part of `split.call(nonStringableReceiver, nonStringableSeparator, Symbol())` (which is inside the `assert.throws` call that looks for ExpectedError). So the presence of a thrown ExpectedError is already a reliable means of determining that `toString` was called and absence of such an error is already a reliable means of determining that it was not, leaving no apparent reason for an invocation counter.

test/built-ins/String/prototype/split/this-value-tostring-error.js
```diff
@@ -0,0 +1,45 @@
...
+esid: sec-string.prototype.split
+description: Abrupt completion from ToString on the "this" value
+info: |
+  1. Let _O_ be ? RequireObjectCoercible(*this* value).
+  1. If _separator_ is neither *undefined* nor *null*, then
+    1. Let _splitter_ be ? GetMethod(_separator_, @@split).
+    1. If _splitter_ is not *undefined*, then
+      1. Return ? Call(_splitter_, _separator_, &laquo; _O_, _limit_ &raquo;).
+  1. Let _S_ be ? ToString(_O_).
+---*/
```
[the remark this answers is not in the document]
gibson042: What's the guidance here? I don't think I'd want this file to be run with a `-f Symbol` filter, so if it's just from use of `Symbol()` then I can replace that with a `nonNumberableLimit`.

PR #2744, opened 2020-08-21
Coverage: `new await ...` and `new (await ...)`. Fixes gh-1898

rwaldron:


rwaldron:
@syg heads up for new failures in V8

PR #2745, opened 2020-08-23
Property clearing code of RegExp constructor is wrong

Constellation:
This code is setting values to $1 etc., which can be non-writable or accessors.
For example, https://github.com/tc39/proposal-regexp-legacy-features/ is specifying them as accessors, and they should throw an error when
it is done in strict code. We should execute `(/(?:)/).test("");` to clear them.


Constellation:
CC @anba @Ms2ger @rwaldron @zbraniecki @littledan @rkirsling @syg @leobalter @FrankYFTang

rwaldron (approved):

Constellation:
Thank you for merging!

PR #2746, opened 2020-08-23
Introduce a new pattern for testing the sequence of potentially abrupt steps

gibson042:
I like the coverage improvements from #2734, but not the file sprawl or difficulty in establishing confidence that there are no further gaps. With correcting that in mind, I wanted to get some feedback on this alternative approach in which a single file is used to test every abrupt completion by iteratively replacing input parameters capable of raising unique errors with values that should get further through an algorithm.

Sample failure output, courtesy Chakra:
```
> test262-harness -t 8 --hostType=ch --hostPath=chakra $(npm run --silent diff)

FAIL test/built-ins/String/prototype/split/errors.js (default)
  Abrupt completions from GetMethod(_separator_, @@split) are propagated. Expected a SeparatorGetSplitError but got a ReceiverToStringError

FAIL test/built-ins/String/prototype/split/errors.js (strict mode)
  Abrupt completions from GetMethod(_separator_, @@split) are propagated. Expected a SeparatorGetSplitError but got a ReceiverToStringError
```


gibson042:
Nice; it correctly found bugs!

## Chakra
```
FAIL test/built-ins/String/prototype/split/errors.js (default)
  Abrupt completions from GetMethod(_separator_, @@split) are propagated. Expected a SeparatorGetSplitError but got a ReceiverToStringError

FAIL test/built-ins/String/prototype/split/errors.js (strict mode)
  Abrupt completions from GetMethod(_separator_, @@split) are propagated. Expected a SeparatorGetSplitError but got a ReceiverToStringError
```
(exposed by source text like `String.prototype.split.call({toString(){throw new UnexpectedError()}}, {[Symbol.split]: 0})`)

## XS
```
FAIL test/built-ins/String/prototype/split/errors.js (default)
  Abrupt completions from ToString(_separator_) are propagated. Expected a SeparatorToStringError to be thrown but no exception was thrown at all

FAIL test/built-ins/String/prototype/split/errors.js (strict mode)
  Abrupt completions from ToString(_separator_) are propagated. Expected a SeparatorToStringError to be thrown but no exception was thrown at all
```
(exposed by source text like `"".split({toString(){throw new ExpectedError()}}, 0)`)

rwaldron:
Both @leobalter and I are on vacation this week. I will look at this first thing next week.

test/built-ins/String/prototype/split/errors.js
```diff
@@ -0,0 +1,124 @@
...
+    { toString() { return undefined; } },
+    Symbol.split,
+    { value: undefined })
+};
+
+// End with a verification of non-abrupt completion when invoking
+// the final (all "good") invoker.
+lastCallCount = callCount;
+invoker(input)();
+assert.sameValue(callCount, lastCallCount + 1);
```
rwaldron, line 124: One bit before I have to put this away until next week: we're trying to break tests up into smaller chunks, not consolidate. I'd be more interested in seeing a new harness helper file, if you want to propose something like that? Either way, these all need to return to their original files.

gibson042: My position motivating this PR is that having such granularity and spreading tests over so many files invites coverage gaps—I'm explicitly advocating _for_ consolidation, specifically of all abrupt completions for an operation into one file.

leobalter: _It's not a big deal for a simple case like this PR, but I would slightly prefer detailed results_.

@gibson042 I understand this motivation and sympathize with it but there are some catches.

The flow for tests execution and error reporting depends on the way we handle assertions. It is ridiculously unfortunate that our assertions are designed to throw on failures rather than reporting or storing the failures and preventing code to continue executing. All of us here knows the advantages of an improved flow like we've had from QUnit. I miss it a lot in Test262.

That leads us for the case where an initial failure does prevent a detailed report, hence that's one big reason why we have multiple files.
 
My biggest take here is that I'd love to enhance the current system available for Test262 API, with something that allows a better grouping of assertions to preserve a consistent error reporting. It's important that we provide better tools to permit consolidation of tests capture the same groups of abrupt completions, etc. I believe we lack those in our API and there are different ways to tackle this. I believe we are in a good opportunity here as our bias somehow connect to QUnit and it might be a good path to discuss for what we might want to design.

gibson042: So is there a path forward here? Ordering is an observable and therefore testable part of the spec, but having O(N) files dedicated to it for _each_ N-parameter function seems completely unmaintainable.

Issue #1183, opened 2017-08-16
`SuperCall` tests missing parens


mroch:
https://github.com/tc39/test262/blob/eaffe8671cade47b1990a1b703672b3098112a1f/test/language/expressions/object/method-definition/name-super-call-body.js#L17
https://github.com/tc39/test262/blob/eaffe8671cade47b1990a1b703672b3098112a1f/test/language/expressions/object/method-definition/name-super-call-param.js#L16

`HasDirectSuper` is true when there is a `SuperCall`. `super;` is not a `SuperCall` without parens.

https://tc39.github.io/ecma262/#sec-method-definitions-static-semantics-hasdirectsuper

This makes these tests test whether `super;` is a syntax error, not whether `super()` is allowed outside of constructors.

littledan:
Sounds like it's valuable to have both tests.

PR #2754, opened 2020-09-04
SuperCall includes parenthesis. Fixes gh-1183

rwaldron:


test/language/expressions/object/method-definition/name-super-call-body.js
```diff
@@ -14,6 +14,6 @@ $DONOTEVALUATE();
 
 ({
   method() {
-    super;
+    super();
```
shvaikalesh, line 17: I second [Daniel's comment](https://github.com/tc39/test262/issues/1183#issuecomment-323088157) on having two tests.

leobalter: The problem here is finding what part of ES matches for `super;`.

Quick answer: none. So this is quite similar to have other known keywords like in `method() { if; }`. The addition of wrong syntax is subjective and could be expanded to infinity.

The current tests are aiming to [a specific point of the spec](https://tc39.es/ecma262/#sec-object-initializer-static-semantics-early-errors) that invalidates code that could be seen as valid for parsers otherwise.

Due to popularity of `super` I can even understand adding one single test but I still want a better reason, like maybe a mainstream implementation currently accepting it as valid.

leobalter (approved):

test/language/expressions/object/method-definition/name-super-call-body.js
rwaldron: @shvaikalesh another way to view this is that ES doesn't have any explicit rule that says an engine cannot intentionally extend its syntax to include some meaning for `super;`. That's very much allowed, as is any extension, as long as it doesn't violate the [Forbidden Extensions](https://www.ecma-international.org/ecma-262/#sec-forbidden-extensions)

PR #2757, opened 2020-09-06
Test "length" [[Value]] coercion & overflow check order in ArraySetLength

shvaikalesh:
JSC bug: [Assertion failure when redefining array length](https://bugs.webkit.org/show_bug.cgi?id=158791).


test/built-ins/Array/length/define-own-prop-length-value-coercion-order.js
```diff
@@ -0,0 +1,69 @@
+// Copyright (C) 2020 Alexey Shvayka. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+author: André Bargull
```
shvaikalesh, line 5: All credits go to @anba.

rwaldron (approved):

PR #2758, opened 2020-09-09
Improve RegExp.prototype[@@replace] poisoned stdlib test

shvaikalesh:
JSC bug: [Use `@putByValDirect` instead `Array.prototype.@push` in builtins](https://bugs.webkit.org/show_bug.cgi?id=175432).

Follow-up of #2481.

### DIFF
--- a/test/built-ins/RegExp/prototype/Symbol.replace/poisoned-stdlib.js
+++ b/test/built-ins/RegExp/prototype/Symbol.replace/poisoned-stdlib.js
@@ -17,6 +17,17 @@ assert(delete String.prototype.indexOf);
 assert(delete String.prototype.slice);
 assert(delete String.prototype.substring);
 
+for (let i = 0; i < 5; ++i) {
+    Object.defineProperty(Array.prototype, i, {
+        get: function() {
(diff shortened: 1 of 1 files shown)


test/built-ins/RegExp/prototype/Symbol.replace/poisoned-stdlib.js
```diff
@@ -17,6 +17,17 @@ assert(delete String.prototype.indexOf);
 assert(delete String.prototype.slice);
 assert(delete String.prototype.substring);
 
+for (let i = 0; i < 5; ++i) {
+    Object.defineProperty(Array.prototype, i, {
+        get: function() {
+            throw new Test262Error(i + " getter should be unreachable.");
+        },
+        set: function(_value) {
+            throw new Test262Error(i + " setter should be unreachable.");
```
shvaikalesh, line 26: While JSC uses private copy of `Array.prototype.push` that is unaffected by `delete` at L11, it still performs the [`push()`](https://tc39.es/ecma262/#sec-array.prototype.push) which does `[[Set]]` at step **5.a**.

rwaldron (approved):

PR #2771, opened 2020-09-11
Remove invalid tests. Fixes gh-427

rwaldron:


leobalter:
This would still need new tests to cover respective items from 427 properly.

Issue #2121, opened 2019-04-03
Document how to report bugs against each ES implementation


gibson042:
We should make it easy for test authors or other consumers/contributors of test262 to inform engines of bugs when they are discovered.

PR #2775, opened 2020-09-14
Contributing: add list of links to implementation bug trackers. Fixes gh-2121

rwaldron:


anba:
Maybe `https://bugzilla.mozilla.org/enter_bug.cgi?product=Core&component=JavaScript%20Engine` for SpiderMonkey, so it's not necessary to find the correct component? We can move to the more narrow "JavaScript: Standard Library" or "JavaScript: Internationalization API" components when triaging.

Issue #2773, opened 2020-09-12
Missing tests for Array.prototype.shift method


NWU-NISL:
**Description:**

ES specification states that if the length property of the argument passed to `Array.prototype.shift` method cannot be changed , JS engines should throw a `TypeError`. So if the inputting argument is a function object whose length property is unconfigurable as the following test case, JS engines should throw a `TypeError`. We didn't find a similar test case from Test262 test suite.

```js
var NISLFuzzingFunc =function(array){
    return Array.prototype.shift.call(array);
};
var NISLParameter0=function(config){};
NISLFuzzingFunc(NISLParameter0);
```

The above test case triggers a bug of JavaScriptCore and Rhino(https://github.com/mozilla/rhino/issues/612).

**Engine Version:**
Rhino-1.7.11
JavaScriptCore-d940b47

**Output:**

```
 
```

**Expected Output:**

```
TypeError:"length"is read-only
```

**ECMA262 Link:**

1. https://tc39.es/ecma262/#sec-array.prototype.shift
   **22.1.3.24 Array.prototype.shift ( )**
   When the `reduceRight` method is called with one or two arguments, the following steps are taken: 
   …… 
   6. ......
      e. Else, 
        i. Assert: fromPresent is false.
       ii. Perform? [DeletePropertyOrThrow](https://tc39.es/ecma262/#sec-deletepropertyorthrow)(O, to)
      ……
2. https://tc39.es/ecma262/#sec-deletepropertyorthrow
   **7.3.9 DeletePropertyOrThrow ( O, P )**
      ......

      3. Let success be ? O.[[Delete]] (p)
      4. If success is false, throw a TypeError exception.
         ......

rwaldron:
This _is_ tested, but not with a function object. I'll update the test to include more objects where `length` would produce this outcome.

PR #2776, opened 2020-09-14
Array#shift throws TypeError if this value's "length" property was defined with [[Writable]]: false. Fixes gh-2773

rwaldron:


test/built-ins/Array/prototype/shift/throws-when-this-value-length-is-writable-false.js
```diff
@@ -25,3 +25,11 @@ assert.throws(TypeError, () => {
 assert.throws(TypeError, () => {
   Array.prototype.shift.call('abc');
 }, "Array.prototype.shift.call('abc')");
+
+assert.throws(TypeError, () => {
+  Array.prototype.shift.call(function() {});
```
leobalter, line 30: neat

leobalter (approved):

PR #2779, opened 2020-09-15
Intl.NumberFormat + currency will throw an error if maximumFractionDigits is less than currencyDigits

Constellation:
In 12.1.1 SetNumberFormatDigitOptions step 12.d[1], mnfd (minimum fraction digits) becomes the same to currencyDigits (mxfdDefault in this case).
It is 2 for USD, 4 for CLF. So, if maximumFractionDigits is less than that, we should throw RangeError.

[1]: https://tc39.es/ecma402/#sec-setnfdigitoptions


rwaldron (approved):

Constellation:
Thanks!

Issue #1268, opened 2017-10-12
Missing coverage: Early error for assignment to template literal


littledan:
Reported as a Node.js bug, now being [fixed in V8](https://chromium-review.googlesource.com/c/v8/v8/+/712794), the patch doesn't turn on or off any test262 tests, so I suspect it is missing coverage.

PR #2781, opened 2020-09-15
AssignmentTargetType cases. Fixes gh-1268

rwaldron:


rwaldron:
The actual generated test files start here: https://github.com/tc39/test262/pull/2781/files#diff-4de39601c68307203517b0771dc12073 if anyone is interested.

rwaldron:

src/assignment-target-type/invalid/direct.template
```diff
@@ -0,0 +1,15 @@
...
+path: language/expressions/assignmenttargettype/direct-
+name: Direct assignment
+info: |
+  Direct assignment
+---*/
+$DONOTEVALUATE();
+
+function _() {
+  /*{ assignmenttarget }*/ /*{ operator }*/ /*{ value }*/;
+}
```
mathiasbynens, line 14: Since this function is never called, it defeats the purpose of the `$DONOTEVALUATE()` mechanism: in engines that implement the error as a runtime error (instead of an early/parse error) no error is thrown because the code isn’t executed.

Could we either

a) rewrite this test without the function, or
b) keep it as-is but call the function at the end?

cc @leszekswirski

src/assignment-target-type/invalid/parenthesized.template
```diff
@@ -0,0 +1,18 @@
...
+info: |
+  ParenthesizedExpression: (Expression)
+
+  Return AssignmentTargetType of Expression.
+---*/
+$DONOTEVALUATE();
+
+function _() {
+  (/*{ assignmenttarget }*/) /*{ operator }*/ /*{ value }*/;
+}
```
mathiasbynens, line 17: Same comment here.

src/assignment-target-type/invalid/direct.template
mathiasbynens: Filed https://github.com/tc39/test262/issues/3160 for this

Issue #1179, opened 2017-08-15
Add coverage for reserved words in shorthand properties


mroch:
[test/language/expressions/object/let-non-strict-syntax.js](https://github.com/tc39/test262/blob/master/test/language/expressions/object/let-non-strict-syntax.js)
[test/language/expressions/object/yield-non-strict-syntax.js](https://github.com/tc39/test262/blob/master/test/language/expressions/object/yield-non-strict-syntax.js)

these test that `var object = {let};` is valid in non-strict mode. there is no corresponding test asserting that it's a SyntaxError in strict mode.

rwaldron:
Thanks for the report.

> there is no corresponding test asserting that it's a SyntaxError in strict mode.

That's correct, because it's not a SyntaxError to use `let` (or any reserved word) as a property name: https://tc39.github.io/ecma262/#prod-PropertyName

```js
(function() { 
  "use strict"; var o = {let: 1}; 
  console.log(o.let); 
})();
// 1
```

mroch:
```
(function() {
  var yield = 1;
  (function() {
    var o = {yield}; // ok
    console.log(o.yield);
  })();
})();
```

```
(function() {
  var yield = 1;
  (function() {
    "use strict"; 
    var o = {yield}; // not ok
    console.log(o.yield);
  })();
})();
```

rwaldron:
Ok, I see now, sorry for the misunderstanding.

rwaldron:
Do you want to add the test for `void let;`?

rwaldron:
Covered in test/language/expressions/assignment/dstr/obj-id-identifier-yield-ident-invalid.js

PR #2782, opened 2020-09-15
Add coverage for reserved words in shorthand properties. Fixes gh-1179

rwaldron:


test/language/expressions/object/identifier-shorthand-implements-invalid-strict-mode.js
```diff
@@ -0,0 +1,29 @@
+// Copyright (C) 2020 Rick Waldron. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: >
+  Object literal shorthands are only valid with identifier references,
+  not property names. (implements)
```
leobalter, line 7: this phrasing is weird.

Perhaps _Object literal shorthands are limited to valid identifier references. Future Reserved Words are disallowed in Strict Mode._

test/language/expressions/object/identifier-shorthand-implements-invalid-strict-mode.js
```diff
@@ -0,0 +1,29 @@
...
+esid: sec-object-initializer
+flags: [noStrict]
+info: |
+  PropertyDefinition:
+    IdentifierReference
+    CoverInitializedName
+    PropertyName : AssignmentExpression
+    MethodDefinition
```
leobalter, line 15: Adding this would be pretty helpful:

```
  Identifier : IdentifierName but not ReservedWord
    It is a Syntax Error if this phrase is contained in strict mode code and
    the StringValue of IdentifierName is: "implements", "interface", "let",
    "package", "private", "protected", "public", "static", or "yield". 
```

leobalter (commented):
I believe we do, but I wonder if we have a test for `({ await })` as valid in Strict Mode.

Issue #1742, opened 2018-09-12
Add tests for completion value resolution in try/catch/finally blocks


leobalter:
This is feedback after reviewing this [contributed test](https://github.com/tc39/test262/blob/master/implementation-contributed/javascriptcore/stress/completion-value.js)

- [ ] v8: add tests for completion value resolution in try/catch/finally blocks (anything returning -2)

```js
eval('99; do { -99; try { 42 } catch (e) { -1 } finally { -2; break; -3 }; } while (false);')
eval('99; do { -99; try { [].x.x } catch (e) { 42; } finally { -2; break; -3 }; } while (false);')
eval('99; do { -99; try { 42 } catch (e) { -1 } finally { -2; break; -3 }; -77 } while (false);')
eval('99; do { -99; try { [].x.x } catch (e) { 42; } finally { -2; break; -3 }; -77 } while (false);')
eval('99; do { -99; try { 42 } catch (e) { -1 } finally { -2; continue; -3 }; } while (false);')
eval('99; do { -99; try { [].x.x } catch (e) { 42; } finally { -2; continue; -3 }; } while (false);')
eval('99; do { -99; try { 42 } catch (e) { -1 } finally { -2; continue; -3 }; -77 } while (false);')
eval('99; do { -99; try { [].x.x } catch (e) { 42; } finally { -2; continue; -3 }; -77 } while (false);')
```

expected: 42
actual: -2

PR #2792, opened 2020-09-17
Coverage: try/catch/finally completion values. Fixes gh-1742

rwaldron:


rwaldron:
@syg heads up for failures in V8—these might be unexpected.

leobalter:
@rwaldron After battle testing this PR, I took the liberty to fix the eval tests and add some others involving return completions.

There is still a chance to add tests for try-catch-finally messing up returns using break or continue, lol.

```js
function fn() {
  do {
    try {
      return 'NOT ME';
    } catch {}
    finally {
      break;
    }
  } while (false);
  return 42;
}

fn(); // 42;
```

test/language/statements/try/completion-values-fn-finally-abrupt.js
```diff
@@ -0,0 +1,79 @@
...
+  } finally {
+    count.finally += 1;
+    throw new Test262Error('finally'); // If F.[[Type]] is normal, set F to C.
+  }
+  return 'wat';
+};
+
+assert.throws(Test262Error, fn, '1: try Abrupt, catch Abrupt, finally Abrupt; Completion: finally');
+assert.sameValue(count.catch, 1, '1: catch count');
+assert.sameValue(count.finally, 1, '1: fiinally count');
```
bakkot, line 39:
```suggestion
assert.sameValue(count.finally, 1, '1: finally count');
```

rwaldron:
@bakkot nice catch!

PR #2798, opened 2020-09-19
Connector of date and time style can be " at " or ", "

Constellation:
macOS system ICU is shipping new CLDR, but it has many overrides on the top of it to make the formatted output suitable for the system.
And in timedatestyle-en.js tests, we intentionally override the CLDR data with the different format.
This change modifies the test to accept that alternative output.


rwaldron (approved):

Constellation:
Thanks!

Issue #2089, opened 2019-03-03
Missing coverage for interaction of rest properties with number literal keys and numeric computed keys


adrianheine:
Examples:
```js
const source = { 1: 'one', 2: 'two' };
const { 1.: _, ...rest } = source;
assert(!(1 in rest));
```
```js
const source = { 1: 'one', 2: 'two' };
const { 1e0: _, ...rest } = source;
assert(!(1 in rest));
```
```js
const source = { 1: 'one', 2: 'two' };
const { [1+0]: _, ...rest } = source;
assert(!(1 in rest));
```

leobalter:
Thanks for reporting this. It's an interested part to be covered.

cc @caiolima

caiolima:
Sorry for delay here! I'll update this later this week.

rwaldron:
@leobalter @caiolima did this ever get resolved?

caiolima:
I don't think it was solved. I don't remember adding those cases and last update on obj rest tests are from https://github.com/tc39/test262/commit/29627e892be455f2685057a68d62c64e0d7aa30c

shvaikalesh:
JSC bug: [Add the property to be excluded from the rest operation of the object as string](https://bugs.webkit.org/show_bug.cgi?id=216437).

A coverage PR is on the way.

rwaldron:
Resolved in https://github.com/tc39/test262/commit/be99a3e519841d2f495cc64a272a1e3eb4440b5b

PR #2799, opened 2020-09-19
Add coverage for rest properties with computed key

HyeockJinKim:
gh-2089

### DIFF
--- a/src/dstr-assignment/obj-rest-non-string-computed-property.case
+++ b/src/dstr-assignment/obj-rest-non-string-computed-property.case
@@ -1,0 +1,32 @@
+// Copyright (C) 2020 HyeockJin Kim. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: >
+    Destructuring field can be a non-string computed property, i.e it
+    can be defined only at runtime. Rest operation needs to skip
+    these properties as well.
+template: default
+includes: [propertyHelper.js]
+features: [object-rest]
+---*/
+
+//- setup
+var a = 1.;
+var b, rest;
+//- elems
+{[a]:b, ...rest}
+//- vals
+{[a]: 1, bar: 2 }
+//- body
+assert.sameValue(b, 1);
+
+assert.sameValue(Object.getOwnPropertyDescriptor(rest, "1"), undefined);
+
+verifyProperty(rest, "bar", {
+  value: 2,
+  enumerable: true,
+  writable: true,
+  configurable: true
+});
+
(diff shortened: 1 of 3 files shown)


rwaldron:
Looks like the generated tests in this commit don't match whatever this case is generating. Can you make one commit with the case and another with the generated tests? Thanks!

HyeockJinKim:
I generated the test code for the test case again.
But now 5 CIs fail. What should I do?

shvaikalesh:
@HyeockJinKim CI detects changed tests via `git`, so please rebase on top of `main` to fix the failures.

src/dstr-assignment/obj-rest-non-string-computed-property.case
```diff
@@ -0,0 +1,33 @@
+// Copyright (C) 2020 HyeockJin Kim. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: >
+    Destructuring field can be a non-string computed property, i.e it
+    can be defined only at runtime. Rest operantion needs to skip
```
shvaikalesh, line 7: nit: typo (operantion)

src/dstr-assignment/obj-rest-non-string-computed-property.case
```diff
@@ -0,0 +1,33 @@
...
+    these properties as well.
+template: default
+esid: pending
```
shvaikalesh, line 10: We can remove this since `esid` is also generated.

src/dstr-assignment/obj-rest-non-string-computed-property.case
```diff
@@ -0,0 +1,33 @@
...
+includes: [propertyHelper.js]
+features: [object-rest]
+---*/
+
+//- setup
+var a = 1.;
+var b, rest;
+//- elems
+{[1.]:b, ...rest}
```
shvaikalesh, line 19: I would prefer `{[a]: b` here, as it is a bit more readable.
It would be nice to have more combinations in future:

| `elems` | `vals` |
|-------|------|
| `1.0`   | `1`    |
| `1.`    | `[1]`  |
| `"1"`   | `1.0`  |
| `[1]`   | `"1"`  |
| `1e0`   | `[1]`  |
| `[1e0]`   | `1.0`  |

src/dstr-assignment/obj-rest-non-string-computed-property.case
```diff
@@ -0,0 +1,33 @@
...
+//- vals
+{[a]: 1, bar: 2 }
+//- body
+assert.sameValue(b, 1);
+assert.sameValue(rest.bar, 2);
```
shvaikalesh, line 24: Alternatively, you can test `value: 2` in `verifyProperty`.

shvaikalesh (commented):
Thank you for contributing the tests!

src/dstr-assignment/obj-rest-non-string-computed-property.case
HyeockJinKim: Then, should I create and upload another `.case` file for the combination you presented?

rwaldron (approved):

rwaldron:
@HyeockJinKim can you please sign the [CLA](https://tc39.es/test262-cla/) if you have not yet done so? Thanks!

HyeockJinKim:
I signed it now. Thanks. :)

PR #2801, opened 2020-09-20
Use \t instead of TAB in test/built-ins/parseInt/S15.1.2.2_A2_T1.js

Constellation:
This is not testing the TAB parsing in this string. So let's use \t instead, which makes integration in WebKit tree easy.


rwaldron:
This test needs to be rebased, but more importantly: please note that the entire format of these tests has been updated.

jugglinmike:
I'm going to close this pull request because it's been over a year since changes were requested. If anyone is still interested in this (and particularly, if they can give some detail about how this makes integration in the WebKit tree easy), then I'll be happy to help update the patch and merge it!

Issue #2803, opened 2020-09-22
Missing tests for Array.prototype.pop method


NWU-NISL:
**Description:**
We found that Test262 missing the test case where the argument of the "Array.prototype.pop.call" method is non-object type, such as the following example:

```js
var foo = function(){
   Array.prototype.pop.call(false);
};
foo();
```
This test case caused a [bug] (https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-8240298) of Nashorn which threw a`TypeError` exception because it didn't perform `ToObject` operation when the argument is false. 

 **Nashorn Version:**

nashorn-13.0.1

 
**Nashorn Output:**

```js
TypeError: false is not an Object.
```

**Expected Output:**

```js

```

 

**ECMA262 Link:**

1. https://tc39.es/ecma262/#sec-array.prototype.pop

    **22.1.3.19 Array.prototype.pop ( )**

   When the `pop` method is called, the following steps are taken:

   1. Let O be ? [ToObject](https://tc39.es/ecma262/#sec-toobject)(this value).

      ......

2. [https://tc39.es/ecma262/#sec-toobject](#sec-toobject)

PR #2807, opened 2020-09-22
Coverage: calling array methods with boolean as thisValue. Fixes gh-2803

rwaldron:


test/built-ins/Array/prototype/sort/call-with-boolean.js
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) 2020 Rick Waldron.  All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-array.prototype.sort
+description: Array.prototype.sort applied to boolean primitive
+---*/
+
+assert.sameValue(Array.prototype.sort.call(true), true);
```
shvaikalesh, line 9: Per https://github.com/tc39/ecma262/pull/1585, `? ToObject(this value)` is returned.

test/built-ins/Array/prototype/reverse/call-with-boolean.js
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) 2020 Rick Waldron.  All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-array.prototype.reverse
+description: Array.prototype.reverse applied to boolean primitive
+---*/
+
+assert.sameValue(Array.prototype.reverse.call(true) instanceof Boolean, true);
```
shvaikalesh, line 9:
```suggestion
assert(Array.prototype.reverse.call(true) instanceof Boolean);
```

test/built-ins/Array/prototype/sort/call-with-boolean.js
rwaldron: I'll back this one out

mathiasbynens: Filed https://bugs.chromium.org/p/v8/issues/detail?id=11362

PR #2812, opened 2020-09-22
Coverage: Array.prototype.item, String.prototype.item, TypedArray.prototype.item. Closes gh-2809, gh-2810, gh-2811

rwaldron:
https://tc39.es/proposal-item-method/


rwaldron:
cc @syg

Issue #2813, opened 2020-09-23
Incomplete test coverage of String.prototype.split method


NWU-NISL:
**Description:**

We found that Test262 may lack the following two test cases: 

* Testcase1:

```js
var NISLFuzzingFunc = function(){
  var a = "test".split(/^/);
  print(a);
  print(a.length);
};
NISLFuzzingFunc();
```

**Engine Version:**

[jerry-996bf76](https://github.com/jerryscript-project/jerryscript/commit/996bf76f5956c65342c1726d3f629fbea41320ab)

**Output 1:**

```js
t,e,s,t
4
```

**Expected Output 1:**

```js
test
1
```

------

* Testcase2:

 ```js
var NISLFuzzingFunc =function(){
    var b = 'test'.split(/$/); 
    print(b); 
    print(b.length);
};
NISLFuzzingFunc();
 ```

**Engine Version:**

[hermes-b6530ae](https://github.com/facebook/hermes/tree/b6530ae)

**Output 2:**

```js
test,
2
```

**Expected Output 2:**

```js
test
1
```

The above two test cases respectively triggered a bug of JerryScript (https://github.com/jerryscript-project/jerryscript/issues/3914) and Hermes (https://github.com/facebook/hermes/issues/277). We found no similar test cases from the test suite of Test262.

**ECMA262 Link:**

1. https://tc39.es/ecma262/#sec-string.prototype.split

PR #2815, opened 2020-09-23
Coverage: String.prototype.split method, regexp separator. Fixes gh-2813

rwaldron:


rkirsling:
These filenames are *extremely* long, to the point that it can easily cause breakage on Windows. 😓  Is there something else we can do here, so as to avoid having an entire hyphenated sentence like `arguments-are-objects-and-instance-is-string-first-object-have-overrided-to-string-function-and-value-of-function-that-throw-exception-second-object-have-overrided-value-of-function-that-throw-exception`?

PR #2817, opened 2020-09-23
Revert "Intl.NumberFormat + currency will throw an error if maximumFractionDigits is less than currencyDigits"

Constellation:
This reverts commit a7b9067fe129b8ded01a4b86f8295ed40fe2d9e5.
I fixed the test with a7b9067fe129b8ded01a4b86f8295ed40fe2d9e5 based on the current spec, but this test was based on WIP spec change.


Constellation:
And JSC change https://trac.webkit.org/changeset/267500/webkit

PR #2836, opened 2020-09-29
tests for https://github.com/tc39/ecma262/pull/2154

bmeck:
see https://github.com/tc39/ecma262/pull/2154

### DIFF
--- a/test/language/module-code/export-expname-binding-string.js
+++ b/test/language/module-code/export-expname-binding-string.js
@@ -1,0 +1,17 @@
+// Copyright (C) 2020 Bradley Farias. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+description: Module Namespace Objects can have non-Identifier properties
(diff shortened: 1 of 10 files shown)


test/language/module-code/export-expname-binding-string.js
```diff
@@ -0,0 +1,17 @@
+// Copyright (C) 2020 Bradley Farias. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+description: Module Namespace Objects can have non-Identifier properties
+esid: prod-ExportSpecifier
+info:
```
ExE-Boss, line 6: The lint task is failing because without this, `info` is being interpreted as an object, rather than as a literal multiline string:

```suggestion
info: |
```

rwaldron:
I'm going to merge this and fix the CI issue directly.

ExE-Boss:
This should’ve been squash merged.

Issue #1739, opened 2018-09-12
Built-in function objects that are not identified as constructors do not implement the [[Construct]]


leobalter:
This is feedback after reviewing this [contributed test](https://github.com/WebKit/webkit/blob/master/JSTests/stress/builtin-function-is-construct-type-none.js)

- [ ] Add a forEach test with a new.target call `new Array.prototype.forEach(function () { });`
- [ ] Extend it to other methods

```
Built-in Function Objects

...
Built-in function objects that are not identified as constructors do not implement the [[Construct]] internal method unless otherwise specified in the description of a particular function. When a built-in constructor is called as part of a new expression the argumentsList parameter of the invoked [[Construct]] internal method provides the values for the built-in constructor's named parameters.

Built-in functions that are not constructors do not have a prototype property unless otherwise specified in the description of a particular function.
```

PR #2842, opened 2020-10-01
Built-in function objects that are not identified as constructors must throw a TypeError exception when new'ed. Fixes gh-1739

rwaldron:


test/built-ins/Function/built-in-function-objects-that-are-not-identified-as-constructors.js
```diff
@@ -0,0 +1,1611 @@
...
+
+info: |
+  ECMAScript Function Objects
+
+  All ECMAScript function objects have the [[Call]] internal method defined here. ECMAScript functions that are also constructors in addition have the [[Construct]] internal method.
+features: [arrow-function, Array.prototype.flatMap, Array.prototype.values, Symbol, Symbol.iterator, ArrayBuffer, Atomics, TypedArray, SharedArrayBuffer, BigInt, DataView, FinalizationRegistry, generators, Map, Object.fromEntries, Object.is, Promise.allSettled, Promise.any, Reflect, Reflect.construct, Reflect.set, Reflect.setPrototypeOf, Symbol.match, Symbol.matchAll, Symbol.replace, Symbol.search, Symbol.split, Set, String.prototype.endsWith, String.prototype.matchAll, string-trimming, String.prototype.trimEnd, String.prototype.trimStart, WeakRef, WeakMap, WeakSet]
+---*/
+assert.throws(TypeError, () => {
+  new decodeURI('');
+}, '`new decodeURI("")` throws TypeError');
```
ljharb, line 16: this doesn't verify it lacks [[Construct]] - you'd need Proxy to do that https://esdiscuss.org/topic/isconstructor#content-11

shvaikalesh: Agreed, asserting that `new` throws `TypeError` gives false-positives.
[`isConstructor`](https://github.com/tc39/test262/blob/main/harness/isConstructor.js) harness we already have is way more useful.
It helped cover two bugs related to missing/extra `[[Construct]]` internal method in JSC: [`Proxy.revocable`](https://bugs.webkit.org/show_bug.cgi?id=210959) and [`Symbol` constructor](https://bugs.webkit.org/show_bug.cgi?id=211050).

rwaldron: You're right, and I think that helped me recenter myself on these tests. I'm going to do both.

rwaldron: > Agreed, asserting that new throws TypeError gives false-positives.

Yes, and I took steps to prevent false positives, on a case by case basis. 

Either way, I've re-worked the entire approach to include both the isConstructor checks and asserting "`new ...` + preventions" throws.

test/annexB/built-ins/Date/prototype/getYear/not-a-constructor.js
```diff
@@ -0,0 +1,27 @@
...
+
+assert.sameValue(
+  isConstructor(Date.prototype.getYear),
+  false,
+  'isConstructor(Date.prototype.getYear) must return false'
+);
+
+assert.throws(TypeError, () => {
+  let date = new Date(Date.now()); new date.getYear();
+}, '`let date = new Date(Date.now()); new date.getYear()` throws TypeError');
```
ljharb, line 26: these assertions are "fine" but i don't see what value they add, and this seems like the wrong place to put them - they don't test what the file is claiming to test.

rwaldron: I think maybe you're missing the context for why these tests are even coming into existence? When we were actively capturing JSC tests, for the purpose of curating novel material, or finding gaps in existing coverage, https://github.com/tc39/test262/issues/1739 was filed to call out this test: https://github.com/WebKit/webkit/blob/master/JSTests/stress/builtin-function-is-construct-type-none.js. 

I've added further specification citations to the info for these tests, citing the step in `EvaluateNew ( constructExpr, arguments )`.

ljharb: I was missing the context; however, the test files say they're testing the lack of a [[Construct]] slot, which is achieved by the first test.

rwaldron: Got it, and yes—that's what I thought you were referring to and what I hoped to address with: 

> I've added further specification citations to the info for these tests, citing the step in EvaluateNew ( constructExpr, arguments ).

rwaldron: I thought I had added that language to all of them, looks like I missed some :| Fixing that now.

test/built-ins/BigInt/is-a-constructor.js
```diff
@@ -0,0 +1,33 @@
...
+  - is not intended to be used with the new operator or to be subclassed. It may be used as the value of an extends clause of a class definition but a super call to the BigInt constructor will cause an exception.
+
+  sec-bigint-constructor-number-value
+
+  1. If NewTarget is not undefined, throw a TypeError exception.
+includes: [isConstructor.js]
+features: [BigInt, Reflect.construct, arrow-function]
+---*/
+
+assert.sameValue(
```
shvaikalesh, line 20: Nice, this caught [`BigInt` constructor should be constructible while it always throws an error](https://bugs.webkit.org/show_bug.cgi?id=217575).

Issue #2845, opened 2020-10-03
Coverage: missing tests for Object.seal method


NWU-NISL:
**Description:**

We found that Test262 may miss the test case where the argument of the `Object.seal` method is a string instance created by the `new` operator, such as the following example:

```js
function main(){
    var a = new String(1);
    Object.seal(a);
};
main();
```

This test case caused a [bug](https://github.com/mozilla/rhino/issues/770) of rhino. 

**Rhino Version:**

rhino-1.7.12

**Rhino Output:**

```js
Exception in thread "main" java.lang.NullPointerException
```

**Expected Output:**

```js

```

 

**ECMA262 Link:**

1. https://tc39.es/ecma262/#sec-object.seal

PR #2848, opened 2020-10-05
Coverage: Object.seal instances of all built-in constructors. Fixes gh-2845

rwaldron:


rkirsling:
@rwaldron Seems like https://github.com/tc39/test262/pull/2815#issuecomment-701021649 might have gotten overlooked, but this new naming practice might not be ideal for cross-platform usability (`object-seal-the-configurable-attribute-of-own-accessor-property-of-o-is-set-from-true-to-false-and-other-attributes-of-the-property-are-unaltered` isn't quite as long as the previous, but it's certainly up there).

PR #2858, opened 2020-10-09
Add a test verifying that duplicate variants in the tlang component of a language tag make it structurally invalid

jswalden:
(This is currently un-exercised so may contain tpyos or similar -- posting so I can test that it works correctly against SM by importing it.)


FrankYFTang:
could we separate the dup variant and dup tl into two different test files?

FrankYFTang:
Just split them into two files like
test/intl402/Locale/reject-duplicate-variants.js
test/intl402/Locale/reject-duplicate-in-tlang.js

jswalden:
I would be surprised if we don't have tests for rejecting duplicate variants already -- I included them mostly for reader understandability -- but sure, split into separate tests, with corresponding wording changes and so on.

rwaldron (approved):

rwaldron:

PR #2868, opened 2020-10-17
Add SyntaxError tests for string literal export name

syg:
cc @bfarias-godaddy @rwaldron


test/language/module-code/export-expname-string-binding.js
```diff
@@ -0,0 +1,24 @@
...
+negative:
+  phase: parse
+  type: SyntaxError
+---*/
+
+export { "foo" as "bar" }
+
+function foo() {}
+
+$DONOTEVALUATE();
```
rwaldron, line 23: I'm going to fix these after this lands

Issue #2879, opened 2020-10-27
Align Detached Buffer Semantics: test update needs, from QuickJS


rwaldron:
From QuickJS: 

- [x] test/built-ins/TypedArray/prototype/every/BigInt/callbackfn-detachbuffer.js:35: Test262Error: Expected a TypeError to be thrown but no exception was thrown at all (Testing with BigInt64Array.)
    - updated
- [x] test/built-ins/TypedArray/prototype/every/callbackfn-detachbuffer.js:28: Test262Error: Expected SameValue(«1», «2») to be true (Testing with Float64Array.)
    - this one is correct
- [x] test/built-ins/TypedArray/prototype/filter/BigInt/callbackfn-detachbuffer.js:27: Test262Error: Expected a TypeError to be thrown but no exception was thrown at all (Testing with BigInt64Array.)
    - updated
- [x] test/built-ins/TypedArray/prototype/filter/callbackfn-detachbuffer.js:20: Test262Error: Expected SameValue(«1», «2») to be true (Testing with Float64Array.)
    - this one is correct
- [x] test/built-ins/TypedArray/prototype/find/BigInt/predicate-may-detach-buffer.js:43: Test262Error: throws a TypeError getting a value from the detached buffer Expected a TypeError to be thrown but no exception was thrown at all (Testing with BigInt64Array.)
    - updated
- [x] test/built-ins/TypedArray/prototype/findIndex/BigInt/predicate-may-detach-buffer.js:36: Test262Error: throws a TypeError getting a value from the detached buffer Expected a TypeError to be thrown but no exception was thrown at all (Testing with BigInt64Array.)
    - updated
- [x] test/built-ins/TypedArray/prototype/forEach/BigInt/callbackfn-detachbuffer.js:35: Test262Error: Expected a TypeError to be thrown but no exception was thrown at all (Testing with BigInt64Array.)
    - updated
- [x] test/built-ins/TypedArray/prototype/forEach/callbackfn-detachbuffer.js:28: Test262Error: Expected SameValue(«1», «2») to be true (Testing with Float64Array.)
    - this one is correct
- [x] test/built-ins/TypedArray/prototype/map/BigInt/callbackfn-detachbuffer.js:27: Test262Error: Expected a TypeError to be thrown but no exception was thrown at all (Testing with BigInt64Array.)
    - updated
[...]

PR #2880, opened 2020-10-28
Align detached buffer semantics with web reality, R4. Fixes gh-2879

rwaldron:
@rkirsling ptal, thanks!


rkirsling (approved):
Seems good! Sorry that I've not been examining the BigInt tests as closely—JSC still doesn't have BigInt typed arrays. 😅

PR #2887, opened 2020-10-31
Add tests for [[Description]] of Intl [[FallbackSymbol]]

Constellation:
If normative optional is implemented and [[FallbackSymbol]] is used, its description should be "IntlLegacyConstructedSymbol"

https://tc39.es/ecma402/#intl-object
> The Intl object has an internal slot, [[FallbackSymbol]], which is a new %Symbol% in the current realm with the [[Description]] "IntlLegacyConstructedSymbol".

Tests for https://trac.webkit.org/changeset/266655/webkit change.

### DIFF
--- a/test/intl402/DateTimeFormat/intl-legacy-constructed-symbol-on-unwrap.js
+++ b/test/intl402/DateTimeFormat/intl-legacy-constructed-symbol-on-unwrap.js
@@ -1,0 +1,31 @@
+// Copyright 2020 Apple Inc. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-unwrapdatetimeformat
+description: >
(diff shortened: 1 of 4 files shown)


Constellation:
PTAL @rwaldron  :)

Constellation:
Oops, I should add unwrap side check too.

rwaldron (approved):
Since "normative optional" is only required by web browsers, any tests related to it should be distinguishable. I'm going to accept these as-is, but will make changes in a follow up.

My plan is: 

- add `features: [intl-normative-optional]` (and put it in features.txt) 
- split these into two tests each, one for testing implementations that support "intl-normative-optional", and one and one for testing implementations that do not support "intl-normative-optional".

rwaldron:
Actually, I'm not going to split them up, I'm just going to add the feature so engines can easily filter on them.

Constellation:
Thanks!

Issue #2850, opened 2020-10-06
Permit Symbol as WeakMap key and WeakSet entry


rwaldron:
https://github.com/tc39/ecma262/pull/2038

ptomato:
@rwaldron Mind if I take on updating this to the version of the proposal that eventually reached stage 3?

PR #2890, opened 2020-11-02
Permit Symbol as WeakMap key and WeakSet entry. Closes gh-2850

rwaldron:
For changes that are explicitly relevant to "Permit Symbol as WeakMap key and WeakSet entry", see commit 1.


ExE-Boss:
This needs to be updated so that it also checks that symbols are valid `WeakRef` targets.

Ms2ger (commented):
I reviewed this PR. Nearly everything landed in #3678, except for a few minor changes that I rewrote in #3715. I haven't checked the semi-related changes to feature lists; those are covered by #3665.

PR #2891, opened 2020-11-02
Re-resolve unresolvable references on the global in PutValue. Supports tc39/ecma262#2205

rwaldron:


test/language/identifier-resolution/assign-to-globalThis-undefined.js
```diff
@@ -0,0 +1,20 @@
...
+    If IsStrictReference(V) is true, then
+      Let realm be the current Realm Record.
+      Set V to ? ResolveBinding(realm.[[GlobalEnv]], name).
+      Assert: IsStrictReference(V) is true.
+      If IsUnresolvableReference(V) is true, throw a ReferenceError exception.
+
+flags: [onlyStrict]
+---*/
+
+undeclared = (globalThis.undeclared = 5);
```
jugglinmike, line 20: I don't understand how the `globalThis` binding is related to the normative change. By my reading, this test only differs from the existing test in that it also includes semantics that are verified by [`test/built-ins/global/global-object.js`](https://github.com/tc39/test262/blob/26f1f4567ee7e33163d961c867d689173cbb9065/test/built-ins/global/global-object.js).

test/language/identifier-resolution/assign-to-global-undefined.js
```diff
@@ -3,18 +3,18 @@
...
+      Assert: IsStrictReference(V) is true.
+      If IsUnresolvableReference(V) is true, throw a ReferenceError exception.
 
 flags: [onlyStrict]
-negative:
-  phase: runtime
-  type: ReferenceError
 ---*/
 
 undeclared = (this.undeclared = 5);
```
jugglinmike, line 20: Now that we're expecting this test to execute, we should include an assertion for the result of the expression, like  `assert.sameValue(undeclared, 5);`

jugglinmike (changes requested):

PR #2907, opened 2020-11-18
Fix missing renames for item->at

leobalter:
includes nit pick using verifyProperty

Ref #2905

### DIFF
--- a/test/built-ins/TypedArray/prototype/at/prop-desc.js
+++ b/test/built-ins/TypedArray/prototype/at/prop-desc.js
@@ -17,6 +17,8 @@ assert.sameValue(
   'The value of `typeof TypedArray.prototype.at` is "function"'
 );
 
-verifyNotEnumerable(TypedArray.prototype, 'item');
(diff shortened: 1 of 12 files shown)


rwaldron (dismissed):

rkirsling:
Thanks @leobalter! :bow:

test/built-ins/TypedArray/prototype/at/prop-desc.js
```diff
@@ -17,6 +17,8 @@ assert.sameValue(
   'The value of `typeof TypedArray.prototype.at` is "function"'
 );
 
-verifyNotEnumerable(TypedArray.prototype, 'item');
-verifyWritable(TypedArray.prototype, 'item');
-verifyConfigurable(TypedArray.prototype, 'item');
+verifyProperty(TypedArray.prototype.at, 'name', {
```
rkirsling, line 20: Oops, this and the String one are mistakenly checking `name` here. Will submit another PR to fix.

PR #2910, opened 2020-11-24
Promise.prototype.finally creates thenFinally and catchFinally in onFinally's realm

arai-a:
Test for https://github.com/tc39/ecma262/pull/2233


rwaldron (approved):

test/built-ins/Promise/prototype/finally/then-catch-finally-realm.js
```diff
@@ -0,0 +1,29 @@
...
+    a. Let realm be ? GetFunctionRealm(_onFinally_).
+    c. Let _thenFinally_ be ! CreateBuiltinFunction(_stepsThenFinally_, &laquo; [[Constructor]], [[OnFinally]] &raquo;, realm).
+    g. Let _catchFinally_ be ! CreateBuiltinFunction(_stepsCatchFinally_, &laquo; [[Constructor]], [[OnFinally]] &raquo;, realm).
+  7. Return ? Invoke(_promise_, *"then"*, &laquo; _thenFinally_, _catchFinally_ &raquo;).
+features: [Promise.prototype.finally]
+---*/
+
+var other = $262.createRealm().global;
+
+var onFinally = () => {};
```
jugglinmike, line 20: This test would be more precise if it used the "other" realm to create the `onFinally` function rather than using it to call `Promise.prototype.finally.` That kind of precision would reduce the risk of false positives (e.g. if the runtime incorrectly used the realm of the `then` function) and also give readers a stronger indication of the expected semantics (almost to say, "it should use the realm from this value, right here!").

test/built-ins/Promise/prototype/finally/then-catch-finally-realm.js
```diff
@@ -0,0 +1,29 @@
...
+
+var [thenFinally, catchFinally] = other.Promise.prototype.finally.call({
```
jugglinmike, line 22: Asserting the values returned by `Promise.prototype.finally` expands the scope of this test to include semantics which are unrelated to the normative change. That creates an opportunity (however unlikely) for implementations to spuriously pass if they modify the value returned by the `then` method. More importantly, it makes the test more challenging for human readers to understand.

jugglinmike (commented):
This test appears to correctly assert the proposed semantics, so it should be okay to merge when the change to ECMA262 is accepted.

I've left some thoughts about how the precision and readability could be improved.

Issue #2915, opened 2020-12-08
Detached Array Buffer semantics, follow-ups. 


rwaldron:
From @rkirsling 

> @syg brought up array.length = 0; as an analogous operation to buffer detachment, and I'm thinking that's probably the best way to view the intent here. Thus in the above examples, includes should return true and join should return ",", but we should restore the HasProperty checks for indexOf / lastIndexOf such that they return -1.

I'll make these updates during my 12/9 work day.

rwaldron:
@rkirsling I have a few questions, each of which assumes the following: 

```js
function detach(ab) {
  if (ArrayBuffer.transfer) {
    ArrayBuffer.transfer(ab);
  } else if (ArrayBuffer.detach) {
    ArrayBuffer.detach(ab);
  } else if (typeof detachArrayBuffer === "function") {
    detachArrayBuffer(ab);
  } else if (typeof transferArrayBuffer === "function") {
    transferArrayBuffer(ab)
  } else if (typeof Worker === "function") {
    try { 
      eval("%ArrayBufferDetach(ab)") 
    } catch (e) {
      try {
        var w = new Worker("", {type: "string"});
        w.postMessage(ab, [ab]);
        w.terminate();
      } catch (e) {
        var w = new Worker("", {type: "classic"});
        w.postMessage(ab, [ab]);
        w.terminate();
      }
    }
  } else {
    throw new TypeError("cannot detach array buffer");
  }
}

let u8 = new Uint8Array([0xFF, 0xFF, 0xFF, 0xFF]);
detach(u8.buffer);

let a = [0xFF, 0xFF, 0xFF, 0xFF];
a.length = 0;
```

> includes should return true 

`a.includes(/* anything */)` would return `false`, should `u8.includes(/* anything */)` be consistent?

> join should return ","

Similarly, `a.join() === ""`—shouldn't `u8.join() === ""`?

rwaldron:
I think I answered my own question—those expectations are when detachment occurs after `ValidateTypedArray`

PR #2918, opened 2020-12-10
Align detached buffer semantics with web reality, R6. Fixes gh-2915

rwaldron:
@rkirsling PTAL


test/built-ins/TypedArray/prototype/indexOf/detached-buffer-during-fromIndex-returns-minus-one.js
```diff
@@ -0,0 +1,49 @@
...
+testWithTypedArrayConstructors(function(TA) {
+  const sample = new TA(1);
+  const fromIndex = {
+    valueOf() {
+      $DETACHBUFFER(sample.buffer);
+      return 0;
+    }
+  };
+
+  assert.sameValue(sample.indexOf(0, fromIndex), -1);
```
rkirsling, line 48: It's okay that we test for `indexOf(0, ...)` (and likewise for `lastIndexOf`), but the `indexOf(undefined, ...)` case is the more important one, since it's the one that SM currently fails.

rkirsling (dismissed):
LGTM other than one comment. Thanks again!

test/built-ins/TypedArray/prototype/indexOf/detached-buffer-during-fromIndex-returns-minus-one.js
rwaldron: Got it

rwaldron:
@rkirsling one last look?

rkirsling (approved):

PR #2921, opened 2020-12-14
Set function `length` and `name` in `CreateBuiltinFunction`

ExE-Boss:
**Spec:** <https://github.com/tc39/ecma262/pull/2116>

---

I’ve also added tests for all built‑in functions that had their tests updated in <https://github.com/tc39/test262/pull/2299>.

---

/cc @bakkot @ljharb @michaelficarra @syg

### DIFF
--- a/test/built-ins/Function/property-order.js
+++ b/test/built-ins/Function/property-order.js
@@ -1,0 +1,15 @@
+// Copyright (C) 2020 ExE Boss. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-createbuiltinfunction
+description: Function constructor property order
+info: |
+  Set order: "length", "name", ...
+---*/
+
+var propNames = Object.getOwnPropertyNames(Function);
+var lengthIndex = propNames.indexOf("length");
+var nameIndex = propNames.indexOf("name");
+
+assert(lengthIndex >= 0 && nameIndex === lengthIndex + 1,
+  "The `length` property comes before the `name` property on built-in functions");
--- a/test/built-ins/Promise/all/resolve-element-function-property-order.js
+++ b/test/built-ins/Promise/all/resolve-element-function-property-order.js
@@ -1,0 +1,30 @@
+// Copyright (C) 2020 ExE Boss. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-createbuiltinfunction
+description: Promise.all resolve element function property order
+info: |
+  Set order: "length", "name"
+---*/
+
+var resolveElementFunction;
+var thenable = {
+  then: function(fulfill) {
+    resolveElementFunction = fulfill;
+  }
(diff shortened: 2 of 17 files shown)


rwaldron:
@ExE-Boss I'd like to wait until at least one from your requested "cc" list responds with an approval, otherwise this looks good to me.

test/built-ins/Function/property-order.js
```diff
@@ -0,0 +1,14 @@
...
+esid: sec-createbuiltinfunction
+description: Function constructor property order
+info: |
+  Set order: "length", "name", "prototype", ...
+includes: [compareArray.js]
+---*/
+
+assert.compareArray(
+  Object.getOwnPropertyNames(Function).slice(0, 3),
+  ['length', 'name', 'prototype']
```
bakkot, line 13: The order in which the `prototype` property is added is not actually specified. All we can say is that `length` must be before `name`.

(Also, strictly speaking, engines are permitted to add other properties if they want, so I'm not certain we should be asserting that these are the only properties present.)

ExE-Boss: See the `slice(0, 3)` call above.

ljharb: I'm not sure that would do it - I think the check would need to compare `.filter(x => x === 'name' || x === 'prototype')` to `['name', 'prototype']`.

ExE-Boss: <https://github.com/tc39/ecma262/pull/2116> makes it so that the `length` and `name` properties are the first properties that are added.

bakkot: The problem is that the spec also says that (some of) these things have a "prototype" property without mandating that the property be added at any particular time, and some implementations (spidermonkey, e.g.) create the property at the same time as the function, before adding "length" and "name". I'm not convinced that's forbidden per the spec. (It would be nice to nail this down further, but I'm not convinced it's worth the time it would take to make that happen.)

I'd prefer to just stick with asserting that "length" comes before "name" for now.

bakkot: @ExE-Boss This is still asserting that "length" and "name" come before all other properties. Were you up for changing that so it just asserted the order of those properties, without other changes?

test/built-ins/Promise/all/resolve-element-function-property-order.js
```diff
@@ -0,0 +1,30 @@
+// Copyright (C) 2020 ExE Boss. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-createbuiltinfunction
+description: Promise.all resolve element function property order
+info: |
+  Set order: "length", "name"
+includes: [compareArray.js]
+
```
rwaldron, line 9: Missing the closing `---*/`

rwaldron (changes requested):

rwaldron (approved):

PR #2923, opened 2020-12-28
Getting the test262.py works with python3 for es5-tests.

lygstate:
As for IOT javascript such as jerryscript, they may want only support for ES5 subset feature, and not the up to date
feature, and now python2 are getting deprecating, so getting  test262.py  can be used with python3

### DIFF
--- a/tools/packaging/test262.py
+++ b/tools/packaging/test262.py
@@ -41,8 +41,8 @@ def ReportError(s):
 
 
 if not os.path.exists(EXCLUDED_FILENAME):
-    print "Cannot generate (JSON) test262 tests without a file," + \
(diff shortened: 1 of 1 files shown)


leobalter (changes requested):
I'm not compelled to accept the changes for a branch that should remain stationery. The old testing structure is intended to remain for saving historical purposes, and perhaps being even a TAG, not a branch.

This `es5-tests` branch is no maintained by people who actively working on this project. Even thou, the main challenge to accept new changes here would be setting this branch compatible with the current CI structure.

The CI was left with failure since day 1 of this PR. For that, I'm closing it.

I'd be happy to reopen for review if the CI part can be addressed.

PR #2925, opened 2020-12-28
Add tests for intl402 formatRange(ToParts) to check the behavior when startDate is the same as endDate.

fabalbon:
Add tests for Intl.DateTimeFormat formatRange(ToParts) to check the behavior when startDate is the same as endDate.
This is based on the discussion and conclusion on tc39/proposal-intl-DateTimeFormat-formatRange#19.

### DIFF
--- a/test/intl402/DateTimeFormat/prototype/formatRange/date-same-returns-single-date.js
+++ b/test/intl402/DateTimeFormat/prototype/formatRange/date-same-returns-single-date.js
@@ -1,4 +1,4 @@
-// Copyright 2019 Google Inc. All rights reserved.
+// Copyright 2021 Google Inc. All rights reserved.
 // This code is governed by the BSD license found in the LICENSE file.
 
 /*---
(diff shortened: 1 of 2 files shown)


test/intl402/DateTimeFormat/prototype/formatRange/date-same-returns-single-date.js
```diff
@@ -0,0 +1,42 @@
+// Copyright 2019 Google Inc. All rights reserved.
```
rwaldron, line 1: Surely this is just copy pasta?

test/intl402/DateTimeFormat/prototype/formatRangeToParts/date-same-returns-single-date.js
```diff
@@ -0,0 +1,57 @@
+// Copyright 2019 Google Inc. All rights reserved.
```
rwaldron, line 1: Same here

test/intl402/DateTimeFormat/prototype/formatRange/date-same-returns-single-date.js
fabalbon: Yes, I fixed the date. Thanks

test/intl402/DateTimeFormat/prototype/formatRangeToParts/date-same-returns-single-date.js
fabalbon: Yes, I fixed the date. Thanks

PR #2927, opened 2020-12-30
Line 17 are ending with \r\n, does the String.raw need respect that?

lygstate:
Signed-off-by: Yonggang Luo <luoyonggang@gmail.com>

### DIFF
--- a/test/built-ins/String/raw/special-characters.js
+++ b/test/built-ins/String/raw/special-characters.js
@@ -16,6 +16,6 @@ assert.sameValue(
   String.raw`\\
 \
 `,
-  '\\\n\\\n\\\n',
+  '\\\r\\\r\n\\\n',
   'Literal characters'
 );


jugglinmike:
Thanks for your interest in Test262, @lygstate! This is a tricky test, but it's correct. Here's why:

`String.raw` returns the "raw" representation of the string template. In the ECMAScript spec, there's a whole algorithm dedicated to translating from literal code points (the bytes that show up in the source code) to the raw representation. It's called [the Template Raw Value, usually abbreviated to "TRV"](https://tc39.es/ecma262/#sec-static-semantics-tv-and-trv).

Here are the relevant parts for this particular test:

> The TRV of [*LineTerminatorSequence*](https://tc39.es/ecma262/#prod-LineTerminatorSequence) :: <LF> is the String value consisting of the code unit 0x000A (LINE FEED).
> The TRV of [*LineTerminatorSequence*](https://tc39.es/ecma262/#prod-LineTerminatorSequence) :: <CR> is the String value consisting of the code unit 0x000A (LINE FEED).
>
> [...]
>
> The TRV of [*LineTerminatorSequence*](https://tc39.es/ecma262/#prod-LineTerminatorSequence) :: <CR> <LF> is the String value consisting of the code unit 0x000A (LINE FEED).

The specification also has this helpful note for exactly this case:

> TV excludes the code units of [*LineContinuation*](https://tc39.es/ecma262/#prod-LineContinuation) while TRV includes them. <CR><LF> and <CR> [*LineTerminatorSequence*](https://tc39.es/ecma262/#prod-LineTerminatorSequence)s are normalized to <LF> for both TV and TRV. An explicit [*EscapeSequence*](https://tc39.es/ecma262/#prod-EscapeSequence) is needed to include a <CR> or <CR><LF> sequence.

That's why the test we're discussing is already verifying the correct behavior. I'm going to close this pull request, but I'm happy to elaborate if my explanation isn't clear!

PR #2929, opened 2021-01-07
Fix formatRangeToParts/date-same-returns-single-date.js and add tests using practically-equal two dates

Constellation:
intl402/DateTimeFormat/prototype/formatRangeToParts/date-same-returns-single-date.js is using `formatRange` and `format`.
Fix this test to use `formatRangeToParts` and `formatToParts` since it is the intention of this test.

This patch adds additional tests to intl402/DateTimeFormat/prototype/formatRangeToParts/date-same-returns-single-date.js and
intl402/DateTimeFormat/prototype/formatRange/date-same-returns-single-date.js. The new test uses two dates that are practially-equal,
and ensures the implementation uses `format` or `formatToParts` by detecting they are practically-equal.


Constellation:
@rwaldron PTAL :)

Constellation:
BTW WebKit change: https://github.com/WebKit/WebKit/commit/0abeff78063c5fdfe4f9cb7684ec171ede6b8a90

rwaldron (approved):

PR #2932, opened 2021-01-13
Adding test case for super access from class field defined as arrow function

### LINKED ISSUE #2931: Missing tests for super biding on class fields initializer
We have tests verifying super biding on initializer and even super called from eval, but we are missing tests where super is used within arrow function defined as class field. The motivation behind this test case is a [bug](https://bugs.webkit.org/show_bug.cgi?id=220558) found on JSC implementation.

caiolima:
It closes #2931.

### DIFF
--- a/src/class-elements/super-access-from-arrow-func-on-field.case
+++ b/src/class-elements/super-access-from-arrow-func-on-field.case
@@ -1,0 +1,35 @@
+// Copyright (C) 2021 Caio Lima (Igalia SL). All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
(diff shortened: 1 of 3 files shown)


src/class-elements/super-access-from-arrow-func-on-field.case
```diff
@@ -0,0 +1,35 @@
+// Copyright (C) 2020 Caio Lima (Igalia SL). All rights reserved.
```
rwaldron, line 1: 2021 ;)

caiolima: Oops, time flies...

PR #2939, opened 2021-01-22
Add tests for more precise Array#sort

mathiasbynens:
https://github.com/tc39/ecma262/pull/1585


shvaikalesh:
Thank you for adding these, Mathias!
I can confirm the proposal was implemented in [`r267514`](https://trac.webkit.org/changeset/267514), passing all JSC's benchmarks.

A few possible improvements on the coverage:
* an engine may have different code paths/sorting algorithms for `arr.sort()` and `arr.sort(comparator)` (JSC does): it would be nice to use a dummy comparator in some tests;
* an engine may use different sorting algorithms for small and large arrays (V8 used to do this a while ago);
* an engine may attempt to skip creation of intermediate array for non-sparse and otherwise normal receiver: it would be nice to add a variation of [`precise-comparefn-throws.js`](https://github.com/tc39/test262/pull/2939/files#diff-23a36b56654761afcf9ae41eede5606a7781c6c6cc1572525d831773cc8f7446) test;
* there is [a report](https://bugs.webkit.org/show_bug.cgi?id=202582#c15) on cross-engine inconsistency on using [`HasProperty`](https://tc39.es/ecma262/#sec-hasproperty) vs [`HasOwnProperty`](https://tc39.es/ecma262/#sec-hasownproperty), though I've failed to come up with a test case;
* IIRC `Array.prototype.sort()` in V8 skipped `ToObject` when called on a primitive (at least boolean): please see https://github.com/tc39/test262/pull/2807#discussion_r492844425.

mathiasbynens:
Thanks for those excellent suggestions @shvaikalesh! I'm hoping these can be tackled as a separate follow-up patch. Would you be interested in writing some of these tests?

rwaldron (approved):
These look great—nice work!

PR #2943, opened 2021-02-03
Fix assign-to-object-literal tests

strager:
src/assignment-target-type/primaryexpression-objectliteral.case and
test/language/expressions/assignmenttargettype/direct-primaryexpression-objectliteral.js
try to assign to an object literal. These tests in fact do something
different: they create a block statement, followed by an equal sign with
no left-hand expression:

    {}   // OK: block statement
    = 1  // ERROR: assignment with no left-hand expression

Further, the intended test should *not* produce a parse error, because
assigning to an object literal is a legal destructuring assignment:

    ({} = 1)  // OK: destructuring assignment

Delete these bogus tests. Add a new test in place of the
previously-generated legitimate
test/language/expressions/assignmenttargettype/parenthesized-primaryexpression-objectliteral.js.


jmdyck:
Note that src/assignment-target-type/primaryexpression-objectliteral.case isn't itself a test. Rather, it's one of the ingredients from which tests are generated.

As for test/language/expressions/assignmenttargettype/direct-primaryexpression-objectliteral.js (which *is* a test), the metadata says that it should throw a SyntaxError, and you agree that it's an error, so I'm not sure what the problem is. It doesn't really matter whether you think of it as an attempt to assign to an object literal or as a block statement followed by a malformed assignment; the text doesn't conform to the grammar, so there isn't a "correct" way to describe its (lack of) structure.

(And if `({} = 1)` doesn't produce a SyntaxError, then that's clearly not the "intended" test.)

rwaldron:
What you've described are runtime semantics, but these tests should never even reach runtime execution, as they are intended to throw SyntaxError exceptions during parse phase.

strager:
> Note that src/assignment-target-type/primaryexpression-objectliteral.case isn't itself a test. Rather, it's one of the ingredients from which tests are generated.

Ah. All the .case files I looked at seem to be valid test cases, so I assumed they were supposed to be tested directly too.

> the metadata says that it should throw a SyntaxError, and you agree that it's an error, so I'm not sure what the problem is.

The problem is that an implementation could pass the test but still have a bug. (This is the case for my implementation; my implementation reports a SyntaxError before this PR, but does not report a SyntaxError with this PR.) The test isn't testing what it claims to test.

> the text doesn't conform to the grammar, so there isn't a "correct" way to describe its (lack of) structure.

I agree that my analysis is misleading. I should rephrase it.

> What you've described are runtime semantics

Where did I talk about runtime semantics? The new test I added references only "Static Semantics" parts of the standard, and my commit message talks about how the code is parsed.

rwaldron:
I think I misread/misunderstood something in you're original post. I don't remember why I (mis) interpreted what you wrote as describing runtime behavior. 

I need to go digging for the original issue that motivated me to write those cases. I have no intention of accepting this PR until I have time to re-review.

jmdyck:
> The problem is that an implementation could pass the test but still have a bug.

Yes, but that's always true. 

> my implementation reports a SyntaxError before this PR,

(As expected.)

> but does not report a SyntaxError with this PR.

I.e., it doesn't report a SyntaxError on `({}) = {};`, thus revealing a bug in your implementation? Okay, if there's no other test in test262 that reveals this bug, then it should maybe be added, but that doesn't mean there's a problem with direct-primaryexpression-objectliteral.js.

> The test isn't testing what it claims to test.

It isn't crystal clear, but I believe the test's metadata is claiming that it's invalid (a parse error) to have an ObjectLiteral as the direct target of an assignment expression. From your initial post, I think you agree that that's what it's claiming.

And I would say that the test's code attempts to have an ObjectLiteral as the target of an assignment, which matches the metadata. But you're saying that the test's code *doesn't* attempt to have an ObjectLiteral as the target of an assignment?

strager:
> And I would say that the test's code attempts to have an ObjectLiteral as the target of an assignment, which matches the metadata. But you're saying that the test's code _doesn't_ attempt to have an ObjectLiteral as the target of an assignment?

Agreed.

To me, the following code (from the test) *does not look* like an ObjectLiteral as the target of an assignment:

https://github.com/tc39/test262/pull/2943/commits/c808b0b05cd790cd07ffe0f6d5bf8f6de3cbbe0c#diff-69059bda39f4e43cd1a711c3ec4f3e4634216c5277fafbed41740a01bdc9401d

```javascript
function _() {
  {} = 1;
}
```

but the following code *does look* like an ObjectLiteral as the target of an assignment:

```javascript
function _() {
  ({} = 1);
}
```

(I have a separate commit (no PR yet) which changes the test to include the outer parentheses. I didn't realize that the `({} = 1);` test case is missing from this PR. Maybe this is the cause for the confusion?)

jmdyck:
Perhaps a .template should be added that puts parentheses around the whole assignment? (parenthesized.template only puts parentheses around the left-hand-side.)

strager:
> Perhaps a .template should be added that puts parentheses around the whole assignment? (parenthesized.template only puts parentheses around the left-hand-side.)

I agree! I'm implementing that in another PR (not published [1]). But if we make that change, then test/language/expressions/assignmenttargettype/direct-primaryexpression-objectliteral.js starts failing. Adding `( )` around the assignment is how I noticed the problem with this test in the first place. I'm making this PR (#2943) to fix the test before introducing the `( )`.

[1] Here's a draft: https://github.com/strager/test262/commit/2846d9d79a1c203eb0ca41a41b41ba821a9c27df

jmdyck:
(Please note that I suggested *adding* a .template, not changing an existing one.)

strager:
> (Please note that I suggested _adding_ a .template, not changing an existing one.)

Adding a new template will still lead to a test failure. Do we want broken tests in test262?

jmdyck:
I don't think you've established that direct-primaryexpression-objectliteral.js is a broken test. After all, the test asserts that
```
function _() {
  {} = 1;
}
```
is a parse-time SyntaxError, and you agree that it is, right? So the test is basically valid, in that the stated expectation is correct for the given code.

Instead, it seemed to me that your point was that the test's *description* didn't match what it was actually doing. So perhaps rather than trying to change the test to match the description, do the reverse: suggest a change to the description that you think better describes what the test is doing. (Bearing in mind that this is a generated test, so a change to a .case or .template file will have to make sense for all tests generated from it.)

strager:
> Instead, it seemed to me that your point was that the test's *description* didn't match what it was actually doing.

Correct.

> suggest a change to the description that you think better describes what the test is doing

For direct.template, I don't know what the description could be. For some cases, it tests an assignment with a negative Static Semantic check; for other cases, it tests a context-free syntax error.

For primaryexpression-objectliteral.case, I don't know what the description could be. That .case file has two templates: one which would treat the `{}` as who-knows-what (src/assignment-target-type/invalid/direct.template), and one which would treat `{}` as an expression (src/assignment-target-type/invalid/parenthesized.template).

> So perhaps rather than trying to change the test to match the description

In my view, this PR is doing the following:

* Delete a test (test/language/expressions/assignmenttargettype/direct-primaryexpression-objectliteral.js)
* Rename a test (test/language/expressions/assignmenttargettype/parenthesized-primaryexpression-objectliteral.js -> test/language/expressions/assignment/target-parenthesized-object.js)

I could undelete the deleted test, making it static (not generated). Would that make more sense?

test/language/expressions/assignment/target-parenthesized-object.js
```diff
@@ -0,0 +1,35 @@
...
+
+  1. Return invalid.
+negative:
+  phase: parse
+  type: SyntaxError
+---*/
+
+$DONOTEVALUATE();
+
+({}) = {};
```
leobalter, line 35: there is no reason to replace `1` with `{}`. They end with the same result already pointed out in the metadata for this file.

test/language/expressions/assignmenttargettype/direct-primaryexpression-objectliteral.js
```diff
@@ -1,19 +0,0 @@
...
-  phase: parse
-  type: SyntaxError
-info: |
-    Direct assignment
----*/
-
-$DONOTEVALUATE();
-
-function _() {
-  {} = 1;
```
leobalter, line 17: differently from what this PR points out, there is no specified qualification if this is interpreted as a Block and an invalid Syntax `= 1`. Actually, the spec qualifies some initial grammar for `{} = 1;` that is then prohibited only by the static error pointed out in this test. This AssignmentExpression is invalid because the LHS Expression is an invalid AssignmentTargetType. Look at what folder this test is?

test/language/expressions/assignmenttargettype/parenthesized-primaryexpression-objectliteral.js
```diff
@@ -1,22 +0,0 @@
...
-info: |
-    ParenthesizedExpression: (Expression)
-
-    Return AssignmentTargetType of Expression.
----*/
-
-$DONOTEVALUATE();
-
-function _() {
-  ({}) = 1;
```
leobalter, line 20: this test is valid for the same reason of the test above, also deleted in this PR, going through the same route described in the newly introduced test.

leobalter (commented):
Despite what the PR says, the tests this PR intends to delete are valid.

> they create a block statement, followed by an equal sign with no left-hand expression

Nope, the reasons are described in the other review comments.

leobalter:
I believe this PR is wrong.

There is no qualifi

PR #2944, opened 2021-02-04
Remove mention of "phase: early", since it's never used

jmdyck:
(It was folded into "phase: parse" in PR #1366.)


rwaldron:
Woo-hoo! I forgot that this was still in "INTERPRETING.md". This also means we need to bump the version number.

rwaldron:
Actually, the version doesn't need to be bumped yet, because `test262-stream` will still correctly parse `phase: early` and `phase: parse`.

rwaldron (approved):

PR #2954, opened 2021-02-24
Promise.anySettled

chicoxyzzy:
Tests for https://github.com/tc39/ecma262/pull/2226

There is only one new file `race.js` which checks that both functions are the same function object, everywhere else it's renaming `race` to `anySettled`


rwaldron:
Bold move!

chicoxyzzy:
I was not sure what's the best way to add tests for this change. Adding a separate `anySettled` directory would make it harder to update tests for both combinators. Another option is to keep `Promise.race` tests as before (only change `Promise.race.name` test if that will be web compatible), but add `anySettled.js` with the same source code as currently in `race.js`

rwaldron:
> I was not sure what's the best way to add tests for this change. Adding a separate `anySettled` directory would make it harder to update tests for both combinators. Another option is to keep `Promise.race` tests as before (only change `Promise.race.name` test if that will be web compatible), but add `anySettled.js` with the same source code as currently in `race.js`

I didn't mean to imply that this wasn't the best way to make this change—I think it is!

leobalter:
> I was not sure what's the best way to add tests for this change. Adding a separate `anySettled` directory would make it harder to update tests for both combinators. Another option is to keep `Promise.race` tests as before (only change `Promise.race.name` test if that will be web compatible), but add `anySettled.js` with the same source code as currently in `race.js`

We still need tests for `Promise.race` (not anySettled) asserting a few things:

- `assert.sameValue(Promise.race, Promise.anySettled);`
- prop-desc.js
- length.js // maybe?
- name.js // which is the most important to track name would need to change to `anySettled`.

These would validate `Promise.race` as a copy of anySettled and I couldn't find them quick browsing this PR.

Assuming everything else is a copy and paste with the proper renamings from `race` to `anySettled`, it LGTM. Even the `name` test - test/built-ins/Promise/anySettled/name.js - is good!

Nice job!

chicoxyzzy:
> assert.sameValue(Promise.race, Promise.anySettled);

This one [is in commit already](https://github.com/tc39/test262/pull/2954/files#diff-18643571bf12fc0d2b0e80f35d0a8879a20915e8d6528cc00b9d9bd7931efe36), but maybe it's better to move it to `/race` directory and I need to find a better filename.

> - prop-desc.js
> - length.js // maybe?
> - name.js // which is the most important to track name would need to change to anySettled.

Yes, we definitely need these tests too! I'll add them on Saturday afternoon in my TZ. Thank you Leo!

leobalter:
> > assert.sameValue(Promise.race, Promise.anySettled);
> 
> This one [is in commit already](https://github.com/tc39/test262/pull/2954/files#diff-18643571bf12fc0d2b0e80f35d0a8879a20915e8d6528cc00b9d9bd7931efe36), but maybe it's better to move it to `/race` directory and I need to find a better filename.

it could be just a `.../Promise/race/anySettled.js` file.

> > * prop-desc.js
> > * length.js // maybe?
> > * name.js // which is the most important to track name would need to change to anySettled.
> 
> Yes, we definitely need these tests too! I'll add them on Saturday afternoon in my TZ. Thank you Leo!

I'm looking forward to it, but please take your time!

chicoxyzzy:
Now these files are preserved in `/Promise/race` directory
- prop-desc.js
- length.js
- name.js
- not-a-constructor.js

 and  `anySettled.js/race` is now to `race/anySettled.js`

test/built-ins/Promise/race/name.js
```diff
@@ -0,0 +1,26 @@
...
+
+    [...]
+
+    Unless otherwise specified, the name property of a built-in Function
+    object, if it exists, has the attributes { [[Writable]]: false,
+    [[Enumerable]]: false, [[Configurable]]: true }.
+includes: [propertyHelper.js]
+---*/
+
+assert.sameValue(Promise.race.name, 'race');
```
leobalter: the value should be `allSettled`, right? Ref https://github.com/tc39/ecma262/pull/2226#discussion_r520910084

test/built-ins/Promise/race/length.js
```diff
@@ -0,0 +1,25 @@
+// Copyright (C) 2021 Sergey Rubanov. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 25.4.4.3
```
leobalter: use `esid: sec-promise.race` instead. Same for the other files in this commit with es6id. Most of these are no longer a true reference to es6.

test/built-ins/Promise/race/name.js
```diff
@@ -0,0 +1,26 @@
...
+    identified as an anonymous function has a name property whose value is a
+    String. Unless otherwise specified, this value is the name that is given to
+    the function in this specification.
+
+    [...]
+
+    Unless otherwise specified, the name property of a built-in Function
+    object, if it exists, has the attributes { [[Writable]]: false,
+    [[Enumerable]]: false, [[Configurable]]: true }.
+includes: [propertyHelper.js]
```
leobalter: add `features: [Promise.anySettled]`

test/built-ins/Promise/race/not-a-constructor.js
```diff
@@ -0,0 +1,29 @@
+// Copyright (C) 2021 Sergey Rubanov. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-ecmascript-standard-built-in-objects
```
leobalter: `esid: sec-promise.race`

test/built-ins/Promise/race/not-a-constructor.js
```diff
@@ -0,0 +1,29 @@
...
+includes: [isConstructor.js]
+features: [Reflect.construct, arrow-function]
+---*/
+
+assert.sameValue(isConstructor(Promise.race), false, 'isConstructor(Promise.race) must return false');
+
+assert.throws(TypeError, () => {
+  new Promise.race([]);
+}, '`new Promise.race([])` throws TypeError');
+
```
leobalter: this test file seems redundant as you already have `test/built-ins/Promise/race/anySettled.js`. It doesn't hurt one way or another.

test/built-ins/Promise/race/anySettled.js
```diff
@@ -7,4 +7,4 @@ description: >
   The initial value of the *"race"* property is the same function object as the initial value of the `Promise.anySettled` property.
 ---*/
```
leobalter, line 8: add features: [Promise.anySettled]

leobalter (changes requested):
Just a few minor changes requested, otherwise LGTM

test/built-ins/Promise/anySettled/S25.4.4.3_A1.1_T1.js
```diff
@@ -0,0 +1,13 @@
+// Copyright 2014 Cubane Canada, Inc.  All rights reserved.
+// See LICENSE for details.
+
+/*---
+info: Promise.anySettled is callable
+es6id: S25.4.4.3_A1.1_T1
+author: Sam Mikes
+description: Promise.anySettled is callable
+---*/
```
leobalter, line 9: some meta here: for all the files in this `Promise.anySettled` folder please replace any es6id with `esid: sec-promise-anysettled` and make sure each one of them has the `features: [Promise.anySettled]`, adding the feature tag if they already have a `features` entry.

leobalter (changes requested):

akamsmrhill:
> Tests for https://github.com/tc39/ecma262/pull/2226
> 
> There is only one new file `race.js` which checks that both functions are the same function object, everywhere else it's renaming `race` to `anySettled`

test/built-ins/Promise/anySettled/ctx-ctor.js
```diff
@@ -3,18 +3,18 @@
...
     [...]
     6. Let promiseCapability be NewPromiseCapability(C).
     [...]
     10. Let iteratorRecord be Record {[[iterator]]: iterator, [[done]]: false}.
-    11. Let result be PerformPromiseRace(iteratorRecord, promiseCapability, C).
+    11. Let result be PerformPromiseAnySettled(iteratorRecord, promiseCapability, C).
     [...]
     13. Return Completion(result).
-features: [class]
+features: [classfeatures, Promise.anySettled]
```
leobalter, line 17:
```suggestion
features: [class, Promise.anySettled]
```

test/built-ins/Promise/anySettled/species-get-error.js
```diff
@@ -3,14 +3,14 @@
...
-es6id: 25.4.4.3
+    Promise.anySettled() does not retrieve `Symbol.species` property of the `this` value
+esid: sec-promise-anysettled
 info: |
     1. Let C be the this value.
     2. If Type(C) is not Object, throw a TypeError exception.
     3. Let promiseCapability be ? NewPromiseCapability(C).
     ...
-features: [Symbol.species]
+features: [Symbol.speciesPromise.anySettled]
```
leobalter, line 13:
```suggestion
features: [Symbol.species, Promise.anySettled]
```

leobalter:
Two minor typos, after that it LGTM.

Thanks for working through this, it's a nice job and I hope we ship Promise.anySettled in ECMA-262 soon!

chicoxyzzy:
Thank you for your help!

leobalter (approved):
cc @ljharb @mathiasbynens FYI

test/built-ins/Promise/anySettled/S25.4.4.3_A2.2_T3.js
```diff
@@ -3,14 +3,15 @@
...
     4. Let iterator be GetIterator(iterable).
     5. IfAbruptRejectPromise(iterator, promiseCapability)
-es6id: S25.4.4.3_A2.2_T3
+esid: sec-promise-anysettled
 author: Sam Mikes
-description: Promise.race rejects if GetIterator throws
+description: Promise.anySettled rejects if GetIterator throws
 features: [Symbol.iterator]
 flags: [async]
+features: [Promise.anySettled]
```
leobalter, line 14: another minor typo: duplicate features.

```suggestion
features: [Symbol.iterator, Promise.anySettled]
flags: [async]
```

test/built-ins/Promise/anySettled/S25.4.4.3_A4.1_T2.js
```diff
@@ -2,11 +2,12 @@
...
 
 /*---
-es6id: S25.4.4.3_A4.1_T2
+esid: sec-promise-anysettled
 author: Sam Mikes
-description: Promise.race rejects if IteratorStep throws
+description: Promise.anySettled rejects if IteratorStep throws
 features: [Symbol.iterator]
 flags: [async]
+features: [Promise.anySettled]
```
leobalter, line 10:
```suggestion
features: [Symbol.iterator, Promise.anySettled]
flags: [async]
```

leobalter (commented):
... We should be capturing these duplicate tags in the linting tool.

leobalter:
@chicoxyzzy apparently my eyes couldn't catch all the duplicate `features:` in the files and that causes a false positive in CircleCI (to be resolved separately).

Thanks for fixing that!

Just to register a quick helper: `git grep -o 'features: ' -- test/built-ins/Promise/anySettled | uniq -d` would print any files with duplicates.

test/built-ins/Promise/race/name.js
mathiasbynens: The comment you point to suggests using `anySettled`, not `allSettled`.

But this is something I will bring up in committee. I think the safest/most conservative option is to preserve the pre-existing function object, i.e. have the name be `race`. This matches what happened with `trimLeft`/`trimRight`.

It would be nice (IMHO, nicer) to have the name be `anySettled`, but there is a minor web compat risk, since it would change `Promise.race.name` to `'anySettled'`.

mathiasbynens (approved):

test/built-ins/Promise/race/name.js
chicoxyzzy: I 💯 agree

leobalter: > The comment you point to suggests using `anySettled`, not `allSettled`.

It was a typo. Thanks!

leobalter: I'm not taking any strong sides but I have a preference for the name being `anySettled`. I understand it's a breaking change but it feels like one we can go through without problems.

chicoxyzzy:
I think we can close this PR as normative change had no agreement on March meeting.

PR #2962, opened 2021-03-05
Add coverage for `Yield` parameter in RHS of `in`

jugglinmike:


rwaldron (commented):
@jugglinmike, these are certainly correct but I wonder if there is any back story here? Was there a report of missing coverage?

jugglinmike:
No report, no. I wrote them because I initially misinterpreted the spec (not appreciating the need for the grouping operator), and by the time I realized mistake, the tests were already written.

Test262 doesn't have coverage for `yield` in this position, but all engines I've tested pass these, so you can take them or leave them!

rwaldron:
Cool! Thanks for the additional context

PR #2963, opened 2021-03-06
Add tests for stage 3 proposal "private fields in"

jugglinmike:
I have two questions about the "private class fields" proposal which have bearing on the tests for this proposal:

- [the fallibility of ResolveBinding when applied to a PrivateEnvironment](https://github.com/tc39/ecma262/pull/1668/files#r584007453)
- [the expected behavior of GetValue when applied to a private name binding](https://github.com/tc39/ecma262/pull/1668/files#r570729429)

### DIFF
--- a/test/language/expressions/in/private-field-invalid-identifier-simple.js
+++ b/test/language/expressions/in/private-field-invalid-identifier-simple.js
@@ -1,0 +1,25 @@
+// Copyright 2021 the V8 project authors.  All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Invalid private identifier - simple case
+info: |
+  12.10.5 Static Semantics: AllPrivateIdentifiersValid
+
+  AllPrivateIdentifiersValid is an abstract operation which takes names as an argument.
+
+  RelationalExpression:PrivateIdentifierinShiftExpression
+
+  1. If StringValue of PrivateIdentifier is in names, return true.
+  2. Return false.
+
+esid: sec-relational-operators-runtime-semantics-evaluation
+negative:
+  phase: parse
+  type: SyntaxError
+features: [class-fields-private, class-fields-private-in]
+---*/
+
+$DONOTEVALUATE();
+
+#name in {};
--- a/test/language/expressions/in/private-field-invalid-rhs.js
+++ b/test/language/expressions/in/private-field-invalid-rhs.js
@@ -1,0 +1,26 @@
+// Copyright 2021 the V8 project authors.  All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Syntactic grammar restricts right-hand side
+info: |
+  Syntax
+    RelationalExpression[In, Yield, Await]:
+    [...]
+    [+In]PrivateIdentifier in ShiftExpression[?Yield, ?Await]
+esid: sec-relational-operators-runtime-semantics-evaluation
+negative:
+  phase: parse
+  type: SyntaxError
+features: [class-fields-private, class-fields-private-in]
+---*/
+
+$DONOTEVALUATE();
+
+class C {
+  #field;
+
+  constructor() {
+    #field in () => {};
+  }
+}
--- a/test/language/expressions/in/private-field-presence-accessor.js
+++ b/test/language/expressions/in/private-field-presence-accessor.js
@@ -1,0 +1,33 @@
+// Copyright 2021 the V8 project authors.  All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Value when private name describes an accessor method
+info: |
+  7. Let privateName be ? GetValue(privateNameBinding).
+  8. Assert: privateName is a Private Name.
+  [...]
+  10. Else,
+      a. Assert: privateName.[[Kind]] is "method" or "accessor".
+      b. If PrivateBrandCheck(rval, privateName) is not an abrupt completion,
+         then return true.
+  11. Return false.
+esid: sec-relational-operators-runtime-semantics-evaluation
+features: [class-static-methods-private, class-fields-private-in]
+---*/
+
+let count = 0;
+
+class Class {
+  get #accessor() {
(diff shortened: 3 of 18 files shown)


test/language/expressions/in/private-field-invalid-identifier-simple.js
```diff
@@ -0,0 +1,25 @@
...
+esid: sec-relational-operators-runtime-semantics-evaluation
+negative:
+  phase: parse
+  type: SyntaxError
+features: [class-fields-private-in]
+---*/
+
+$DONOTEVALUATE();
+
+#name in {};
```
ljharb, line 25: confirming i understand negative parse tests: this is ensuring that this is always a syntax error?

test/language/expressions/in/private-field-invalid-rhs.js
```diff
@@ -0,0 +1,26 @@
...
+features: [class-fields-private-in]
+---*/
+
+$DONOTEVALUATE();
+
+class C {
+  #field;
+
+  constructor() {
+    #field in () => {};
```
ljharb, line 24: this seems like it should parse just fine, but the test seems to be a negative one?

test/language/expressions/in/private-field-presence-accessor.js
```diff
@@ -0,0 +1,46 @@
...
+  7. Let privateName be ? GetValue(privateNameBinding).
+  8. Assert: privateName is a Private Name.
+  [...]
+  10. Else,
+      a. Assert: privateName.[[Kind]] is "method" or "accessor".
+      b. If PrivateBrandCheck(rval, privateName) is not an abrupt completion,
+         then return true.
+  11. Return false.
+esid: sec-relational-operators-runtime-semantics-evaluation
+features: [class-fields-private-in]
```
ljharb, line 16: does this also need private class fields themselves?

test/language/expressions/in/private-field-rhs-yield-absent.js
```diff
@@ -0,0 +1,33 @@
...
+features: [class-fields-private-in]
+---*/
+
+$DONOTEVALUATE();
+
+class C {
+  #field;
+
+  static method() {
+    #field in yield;
```
ljharb, line 31: #foundpoetry

test/language/expressions/in/private-field-presence-method.js
```diff
@@ -0,0 +1,46 @@
...
+let parentCount = 0;
+let childCount = 0;
+
+class Parent {
+  #method() {
+    parentCount += 1;
+  }
+
+  static init() {
+    Child = class {
```
ljharb, line 29: might be nice to also have a few of these tests where the child doesn’t shadow the field

test/language/expressions/in/private-field-invalid-identifier-simple.js
rwaldron: Correct!

test/language/expressions/in/private-field-rhs-yield-absent.js
jugglinmike: I'd be happy to add a `shield:` label if we can think of any way that might possibly improve coverage.

test/language/expressions/in/private-field-presence-method.js
jugglinmike: Will do

test/language/expressions/in/private-field-invalid-rhs.js
jugglinmike: I assumed that would be redundant, but it couldn't hurt!

test/language/expressions/in/private-field-presence-accessor.js
jugglinmike: I assumed that would be redundant, but it couldn't hurt!

test/language/expressions/in/private-field-invalid-rhs.js
jugglinmike: This doesn't parse because the nonterminal following the `in` keyword is ShiftExpression, but [ArrowFunction is an AssignmentExpression](https://tc39.es/ecma262/#sec-expressions). That's the same reason `yield` has to be parenthesized.

test/language/expressions/in/private-field-rhs-yield-absent.js
jugglinmike: `keeled: healed = #field in yield shield(peeled)`

test/language/expressions/in/private-field-invalid-rhs.js
jugglinmike: We could add one, but between the test for YieldExpression (which uses ParenthesizedExpression) and the test for ShiftExpression (which does not), that seems more like a test for the wider grammar than for this proposal specifically. It makes me wonder if anyone has looked into precisely verifying expression precedence...

ljharb (approved):
Thank you, this seems great to me!

jugglinmike:
@ljharb Sorry to bug you, but would you mind approving this again? I had to fix a typo in the features flags, and that cleared your review.

ljharb (dismissed):

rwaldron:
@jugglinmike please rebase and then I will merge this! Thanks!

jugglinmike:
Right on, @rwaldron. I've rebased, squashed, and force-pushed the result. [The prior version of this branch is available here](https://github.com/bocoup/test262/pull/new/private-fields-in-orig) in case that's of interest to anyone.

jugglinmike:
Strange. I force-pushed to this Pull Request's branch, and although the new commit is visible from the source repository's page, it's not reflected here.

jugglinmike:
I re-created the commit and force-pushed a second time, and the result is (finally) visible here. Ready when you are, @rwaldron

leobalter (approved):

leobalter:
I'm merging this - after a proper review - to avoid extending more time with this PR open. The tests look great and thanks for holding and cleaning up the rebased commit, @jugglinmike!

PR #2965, opened 2021-03-18
Add tests for stage 3 proposal error cause

legendecas:
Implementation Status:
- ChakraCore: https://github.com/chakra-core/ChakraCore/commit/b992a5d1092cb978a70e873e0ae661c835f05c8c
- JSC: https://github.com/WebKit/WebKit/commit/b03c4f4dada2c477376b9275332e08a3d08eba5f

### DIFF
--- a/test/built-ins/Error/cause_property.js
+++ b/test/built-ins/Error/cause_property.js
@@ -1,0 +1,36 @@
+// Copyright (C) 2021 Chengzhong Wu. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Error constructor creates own cause property
+info: |
+  Error ( message [ , options ] )
+
+  ...
+  4. Perform ? InstallErrorCause(O, options).
+  ...
+
+  20.5.8.1 InstallErrorCause ( O, options )
+
+  1. If Type(options) is Object and ? HasProperty(options, "cause") is true, then
+    a. Let cause be ? Get(options, "cause").
+    b. Perform ! CreateNonEnumerableDataPropertyOrThrow(O, "cause", cause).
+  ...
+
+esid: sec-error-message
+includes: [propertyHelper.js]
+---*/
+
+var message = "my-message";
+var cause = { message: "my-cause" };
+var error = new Error(message, { cause });
+
+verifyProperty(error, "cause", {
+  configurable: true,
+  enumerable: false,
+  writeable: true,
+  value: cause,
+});
+
+verifyProperty(new Error(message), "cause", undefined);
+verifyProperty(new Error(message, { cause: undefined }), "cause", { value: undefined });
--- a/test/built-ins/Error/cause_abrupt.js
+++ b/test/built-ins/Error/cause_abrupt.js
@@ -1,0 +1,49 @@
+// Copyright (C) 2021 Chengzhong Wu. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: InstallErrorCause on abrupt completions
+info: |
+  Error ( message [ , options ] )
+
+  ...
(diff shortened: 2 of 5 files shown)


test/built-ins/Error/cause_property.js
```diff
@@ -0,0 +1,34 @@
...
+    a. Let cause be ? Get(options, "cause").
+    b. Perform ! CreateNonEnumerableDataPropertyOrThrow(O, "cause", cause).
+  ...
+
+esid: sec-error-message
+includes: [propertyHelper.js]
+---*/
+
+var message = "my-message";
+var cause = new Error("my-cause");
```
jugglinmike, line 25: Although we expect developers to set the `cause` to Error instances, there's nothing in the proposal which requires that. Would you mind using a plain Object instead? That will help verify that the feature is value-agnostic without reducing the test's ability to assert identity.

jugglinmike (changes requested):
Thanks for the patch, @legendecas!

Test262 has some automated checks that run against all pull requests, and the results are available at the bottom of each GitHub Pull Request page. The checks are labeled as "build & lint tests." For this Pull Request, the check is reporting a problem:

    Linting complete. 3 errors found.
    /home/circleci/test262/test/built-ins/Error/cause_property.js: HARNESS - verifyConfigurable & verifyProperty may not be used in the same file
    /home/circleci/test262/test/built-ins/NativeErrors/AggregateError/cause-property.js: HARNESS - verifyConfigurable & verifyProperty may not be used in the same file
    /home/circleci/test262/test/built-ins/NativeErrors/cause_property_native_error.js: HARNESS - verifyConfigurable & verifyProperty may not be used in the same file

The `verifyProperty` is capable of verifying the entire property descriptor. Would you mind updating your patch to use that instead of `verifyNotEnumerable`, `verifyWritable`, and `verifyConfigurable`?

(We recommend using `verifyProperty` instead of a combination of the more targeted functions because it's so tricky to get the sequence of those functions right. As a bonus, `verifyProperty` fails with a message describing *every* violated expectation--implementers will only see the *first* violation from the more targeted functions.)

Here are some ideas about improving coverage:

- InstallErrorCause has two observable behaviors that we're not testing yet. They're strange edge cases, but those make them the most likely for implementers to miss!

  - > 1. If Type(options) is Object and ? HasProperty(options, "cause") is true, then

    Could you add a test for when HasProperty returns an abrupt completion?

  - > Let cause be ? Get(options, "cause").

    Could you add a test for when Get returns an abrupt completion?

- In all cases, the error "message" is installed before the error "cause." This might seem inconsequential, but if implementers get it backwards, then that could lead to interoperability bugs. For example:

      var sequence = [];
      new Error(
        {toString() { sequence.push('toString'); }},
        {get cause() { sequence.push('cause'); }}
      );

jugglinmike:
Hi again, @legendecas. Would you like a hand making those changes?

legendecas:
@jugglinmike sorry for the delay, just updated the tests to include your great suggestions. thanks.

jugglinmike (approved):
Hi there, @legendecas! I noticed that you pushed up a new version yesterday, so I've taken another look. (In the future, if you could avoid rewriting previously-submitted commits, that will make it easier for folks to review your changes.)

This new version is great! There are a few minor improvements that we can still make, but they don't concern correctness, so I'll make a follow-up patch to discuss further.

However, there was one issue that was related to correctness: the Proxy constructor was being invoked without the `new` operator. That prevented the test from passing in conforming runtimes. Since the correction was straightforward and since I'd like to make these tests available to consumers as soon as possible, I've gone ahead and pushed a fix to this branch. Hope you don't mind!

legendecas:
@jugglinmike thanks a lot for the correctness update! It is not quite easy to check the test correctness with tools that I'm aware of. I'm so sorry for the poor updates 😵 .

PR #2968, opened 2021-04-03
Add tests for "Class Static Init. Blocks" proposal

jugglinmike:
This proposal is currently at "stage 3" in TC39's standardization
process.


test/language/expressions/function/static-init-await-binding.js
```diff
@@ -0,0 +1,21 @@
...
+  - It is a Syntax Error if the code matched by this production is nested,
+    directly or indirectly (but not crossing function or static initialization
+    block boundaries), within a ClassStaticBlock and the StringValue of
+    Identifier is "await".
+features: [class-static-block]
+---*/
+
+class C {
+  static {
+    (function await(await) {});
```
JLHwung, line 19: Can you add a new test case on function declaration?
```js
class C {
  static {
    function await(await) {};
  }
}
```
Since how the binding identifier inherits `[Await]` is different from function expressions. We can also draft some test cases based on https://github.com/tc39/proposal-class-static-block/issues/43#issuecomment-812224819

jugglinmike: Sounds good. Added a few more

rwaldron:
@rbuckton I need you to review these. Thanks!

Vindiesel25 (dismissed):

Vindiesel25:
pvg

rwaldron:
@rbuckton ping

rwaldron:
These look great! I just tested in v8/d8 with `--harmony-class-static-blocks` turned on and wanted to give @syg the heads up on the following failures: 


- test/language/expressions/arrow-function/static-init-await-binding.js
    - Expected no error, got SyntaxError: Unexpected reserved word

- test/language/expressions/arrow-function/static-init-await-reference.js
    - Expected no error, got SyntaxError: Unexpected reserved word

- test/language/expressions/class/static-init-await-binding.js
    - Expected no error, got SyntaxError: Unexpected reserved word

- test/language/expressions/object/identifier-shorthand-static-init-await-invalid.js
    - Expected test to throw error of type SyntaxError, got Test262: This statement should not be evaluated.

- test/language/statements/class/static-init-await-binding-invalid.js
    - Expected test to throw error of type SyntaxError, got Test262: This statement should not be evaluated.

- test/language/statements/class/static-init-invalid-return.js
    - Expected test to throw error of type SyntaxError, got Test262: This statement should not be evaluated.

- test/language/statements/function/static-init-await-binding-invalid.js
    - Expected test to throw error of type SyntaxError, got Test262: This statement should not be evaluated.

- test/language/statements/try/static-init-await-binding-invalid.js
    - Expected test to throw error of type SyntaxError, got Test262: This statement should not be evaluated.

- test/language/statements/variable/dstr/obj-ptrn-elem-id-static-init-await-invalid.js
    - Expected test to throw error of type SyntaxError, got Test262: This statement should not be evaluated.

rwaldron (dismissed):

rwaldron:
@jugglinmike could you rebase your fixups and push back to the branch? Thanks!!

jugglinmike:
> These look great! I just tested in v8/d8 with `--harmony-class-static-blocks` turned on and wanted to give @syg the heads up on the following failures:

What service! Good news, V8 has tickets for these:

- [Issue 11718: Class static initialization blocks: restrictions on `await` identifier are incorrectly applied](https://bugs.chromium.org/p/v8/issues/detail?id=11718)
- [Issue 11719: Class static initialization blocks: ReturnStatement incorrectly allowed](https://bugs.chromium.org/p/v8/issues/detail?id=11719)

> @jugglinmike could you rebase your fixups and push back to the branch? Thanks!!

Certainly. That's done, and [the original version of this branch is available here](https://github.com/tc39/test262/compare/main...bocoup:class-static-initialization-blocks-prebase) in case that's interesting to anybody.

JLHwung:
@rbuckton mentioned in https://github.com/tc39/proposal-class-static-block/issues/27#issuecomment-812227741 that another way to handle `await` is to parse the static block in `+Await` and disallows if it Contains `AwaitExpression` or for-await-of.

If so, the following examples would throw, they are marked as valid as of ff9be07.

`await` is not allowed in arrow parameters in `+Await`
- `test/language/expressions/arrow-function/static-init-await-binding.js`
- `test/language/expressions/arrow-function/static-init-await-reference.js`

`await` not allowed in class binding in `+Await`
- `test/language/expressions/class/static-init-await-binding.js`

I agree with https://github.com/tc39/proposal-class-static-block/issues/43#issuecomment-829595666, either we come up a formal definition of function boundaries or we do in another way.

test/language/statements/class/static-init-scope-lex-close.js
```diff
@@ -0,0 +1,27 @@
...
+
+var test262 = 'outer scope';
+var probe;
+
+class C {
+  static {
+    let test262 = 'first block';
+  }
+  static {
+    probe = test262;
```
alflennik, line 23: I couldn't figure out how we're using probe - should there be an assertion for ti?

test/language/statements/class/static-init-scope-var-close.js
```diff
@@ -0,0 +1,27 @@
...
+
+var test262 = 'outer scope';
+var probe;
+
+class C {
+  static {
+    var test262 = 'first block';
+  }
+  static {
+    probe = test262;
```
alflennik, line 23: Similar comment to my other comment above - should we be using `probe` in an assertion?

jugglinmike:
Great catch, @alflennik! Fixed in another fixup commit

syg:
I'm confused about some of the `await` as identifier tests.

For class expression, [ClassExpression](https://tc39.es/ecma262/#prod-ClassExpression)'s BindingIdentifier doesn't have [~Await], why should await be allowed? This is plausibly a spec bug.

For arrow parameters, [ArrowFunction](https://tc39.es/ecma262/#prod-ArrowFunction)'s ArrowParameters also doesn't have [~Await]. By comparison, [FunctionExpression](https://tc39.es/ecma262/#prod-FunctionExpression)'s FormalParameters has [~Await].

test/language/expressions/class/static-init-await-binding.js
```diff
@@ -0,0 +1,21 @@
...
+  - It is a Syntax Error if the code matched by this production is nested,
+    directly or indirectly (but not crossing function or static initialization
+    block boundaries), within a ClassStaticBlock and the StringValue of
+    Identifier is "await".
+features: [class-static-block]
+---*/
+
+class C {
+  static {
+    (class await {});
```
syg, line 19: This shouldn't be allowed if the intention is that class static blocks are like async function bodies for the purposes of [Await].

`async function f() { (class await {}) }` doesn't parse.

test/language/expressions/arrow-function/static-init-await-binding.js
```diff
@@ -0,0 +1,21 @@
...
+  - It is a Syntax Error if the code matched by this production is nested,
+    directly or indirectly (but not crossing function or static initialization
+    block boundaries), within a ClassStaticBlock and the StringValue of
+    Identifier is "await".
+features: [class-static-block]
+---*/
+
+class C {
+  static {
+    (await => 0);
```
syg, line 19: Similarly here and in other arrow function tests, `await` is never allowed as an arrow parameter in async function bodies, so shouldn't here either.

syg:
> Currently, the proposal's restrictions on await are not expressed in terms of the Await production parameter. Although the text doesn't appear to align with the intent expressed in this GitHub comment, I've been trying to resolve the incongruities in the proposal itself. In the mean time, this patch is my attempt to be as faithful as possible to what's formally proposed.

Right enough. The editors would really like the draft spec to be reworked so that the await-as-identifier restrictions are expressed here exactly as they would be for async function bodies so the two align. We can wait for @rbuckton to do that.

rwaldron:
@alflennik thanks for taking time to review this! Syntax testing is not for the faint of heart, so I commend your willingness to dive right in to a review. That aside, we (which includes you) would benefit from the champion of this proposal being actively engaged. My understanding is that the proposal itself is now stale (that is, it does not reflect the intended semantics of the feature), which is not good for anyone. I also commend @jugglinmike for sticking to a disciplined approach and only testing the feature as it's currently written, not as its "intended". I'm going to attempt to escalate attention to this PR.

mgaudet:
Just want to leave a comment here: SpiderMonkey's prototype implmentation, landed under [Bug 1712138](https://bugzilla.mozilla.org/show_bug.cgi?id=1712138) passes all these tests _except_ some of the the await static semantic tests, so we're definitely interested in seeing more clarification around `await` semantics as well. 

We currently fail: 

```
expressions/object/identifier-shorthand-static-init-await-invalid.js
language/identifier-resolution/static-init-invalid-await.js
language/statements/class/static-init-await-binding-invalid.js
```

rwaldron:
@mgaudet Thanks for reporting in, that's actually really helpful. 

I've made several attempts to reach @rbuckton both here and via IRC (before and during the recent TC39 meeting). I'm not sure what to do when a champion refuses to engage with a required step in the process. Can someone else take over the proposal?

rbuckton:
I'd like to apologize for the delay on this, I've been out of the office this last week. I'm hoping to get caught up early next week.

rbuckton:
I put up a PR for `proposal-class-static-block` that should clarify some of the parsing behavior based on conversation's I've had with @bakkot and others: https://github.com/tc39/proposal-class-static-block/pull/46

I plan to fully review the test-262 tests and will start working on a PR for ECMA-262 this upcoming week.

test/language/expressions/arrow-function/static-init-await-binding.js
rbuckton: As mentioned by @syg, `await` should be disallowed here, per https://github.com/tc39/proposal-class-static-block/pull/46

test/language/expressions/arrow-function/static-init-await-binding.js
```diff
@@ -0,0 +1,21 @@
+// Copyright (C) 2021 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-identifiers
+description: The restriction on `await` as a BindingIdentifier does not cross function boundaries
+info: |
+  BindingIdentifier : Identifier
+
+  [...]
+  - It is a Syntax Error if the code matched by this production is nested,
```
rbuckton, line 10: As of https://github.com/tc39/proposal-class-static-block/pull/46, this is changing such that the block is parsed in `+Await` and we issue an error if `ContainsAwait` returns **true**, so this section has been removed.

test/language/expressions/arrow-function/static-init-await-reference.js
```diff
@@ -0,0 +1,28 @@
+// Copyright (C) 2021 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-identifiers
```
rbuckton, line 4: As of https://github.com/tc39/proposal-class-static-block/pull/46, there are no longer any changes to `sec-identifiers`, these tests will likely need to be moved to `sec-class-definitions` (due to the change to use `+Await`), or to `sec-class-definitions-static-semantics-early-errors` (due to the new early error when `ContainsAwait` is **true**).

test/language/statements/class/static-init-invalid-await.js
```diff
@@ -0,0 +1,27 @@
...
+/*---
+esid: sec-class-definitions
+description: The "Await" parsing context does not apply to the block's statement list
+info: |
+  Syntax
+
+  [...]
+
+  ClassStaticBlockStatementList :
+     StatementList[~Yield, ~Await, ~Return]opt
```
rbuckton, line 12: Per https://github.com/tc39/proposal-class-static-block/pull/46, this changes to:
```suggestion
     StatementList[~Yield, +Await, ~Return]opt
```
However _AwaitExpression_ and `for await` are both early errors in static semantics.

test/language/statements/class/static-init-invalid-yield.js
```diff
@@ -0,0 +1,27 @@
...
+/*---
+esid: sec-class-definitions
+description: The "Yield" parsing context does not apply to the block's statement list
+info: |
+  Syntax
+
+  [...]
+
+  ClassStaticBlockStatementList :
+     StatementList[~Yield, ~Await, ~Return]opt
```
rbuckton, line 12:
```suggestion
     StatementList[~Yield, +Await, ~Return]opt
```

test/language/statements/class/static-init-scope-var-close.js
```diff
@@ -0,0 +1,27 @@
...
+  }
+}
+
+assert.sameValue(probe262, 'outer scope');
```
rbuckton, line 27:
```suggestion
assert.sameValue(probe, 'outer scope');
```

rbuckton (changes requested):
This looks good, though all of the `esid: sec-identifier` cases will need to be updated once https://github.com/tc39/proposal-class-static-block/pull/46 has merged. That should help to clarify the intended `await` behavior.

Also, while I saw tests for _SuperProperty_ and _ThisExpression_ in a static block, I did not see any for accessing the bound name of the class, i.e.:

```js
var value;
class C { static { value = C; } }
// or
export default class C { static { value = C; } }
// or
(class C { static { value = C; } });
```

This is permitted in static fields today (i.e., `class C { static x = C; }`)

rwaldron:
@rbuckton thanks for the review!

rbuckton:
A few updates regarding the proposal:
- https://github.com/tc39/proposal-class-static-block/pull/46 (switching from `~Await` to `+Await`) has been merged.
- https://github.com/tc39/ecma262/pull/2440 has been created.
- https://github.com/tc39/proposal-class-static-block/pull/47 has been created to track https://github.com/tc39/ecma262/pull/2440 due to a delay in the PR diff being available on ecma262-compare.

https://github.com/tc39/ecma262/pull/2440 will be the definitive source for the specification going forward, though I'll try to keep https://github.com/tc39/proposal-class-static-block/pull/47 up to date with any changes for now.

jugglinmike:
@rbuckton regarding access to the bound names of the class, I decided to express that as a new kind of scope-probing test--one focused on the environment record's derivation (that also highlighted the need for a corresponding test for the variable lexical environment).

This patch is ready for another review.

rbuckton (approved):
This looks good to me, thanks for the hard work!

rwaldron (approved):

Issue #2967, opened 2021-04-02
test Symbol.for('bar').toString()


dckc:
the spec is pretty clear that `Symbol.for('bar').toString()` should evaluate to `"Symbol(bar)"`. Please add a test for this.

https://tc39.es/ecma262/#sec-symbol.prototype.tostring

MDN gives this example:
```
// The key is also used as the description
var sym = Symbol.for('mario');
sym.toString(); // "Symbol(mario)"
```
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/for

context :https://github.com/Moddable-OpenSource/moddable/issues/621

cc @erights @phoddie

jugglinmike:
Thanks for the report, @dckc! I opened gh-2970 to address this.

(I decided to use the `description` property rather than `toString` since that puts slightly less unrelated functionality under test. It still captures XS's bug, though.)

PR #2970, opened 2021-04-09
Add test for [[Description]] from `Symbol.for`

jugglinmike:
Resolves gh-2967. I reviewed the project's coverage for the rest of this algorithm, and it looks okay to me.

### DIFF
--- a/test/built-ins/Symbol/for/description.js
+++ b/test/built-ins/Symbol/for/description.js
@@ -10,7 +10,7 @@ info: |
        is stringKey.
     [...]
     6. Return newSymbol.
-features: [Symbol]
+features: [Symbol, Symbol.prototype.description]
 ---*/
 
 var symbol = Symbol.for({toString: function() { return 'test262'; }});
--- a/test/built-ins/Symbol/for/description.js
+++ b/test/built-ins/Symbol/for/description.js
@@ -1,0 +1,18 @@
+// Copyright (C) 2021 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-symbol.for
+description: Assigns the string value to the [[Description]] slot
+info: |
+    1. Let stringKey be ? ToString(key).
+    [...]
+    4. Let newSymbol be a new unique Symbol value whose [[Description]] value
+       is stringKey.
+    [...]
+    6. Return newSymbol.
+features: [Symbol]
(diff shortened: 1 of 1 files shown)


test/built-ins/Symbol/for/description.js
```diff
@@ -0,0 +1,18 @@
...
+esid: sec-symbol.for
+description: Assigns the string value to the [[Description]] slot
+info: |
+    1. Let stringKey be ? ToString(key).
+    [...]
+    4. Let newSymbol be a new unique Symbol value whose [[Description]] value
+       is stringKey.
+    [...]
+    6. Return newSymbol.
+features: [Symbol]
```
shvaikalesh, line 13:
```suggestion
features: [Symbol, Symbol.prototype.description]
```

shvaikalesh (approved):

leobalter (changes requested):
@jugglinmike The test looks good and it is just missing the feature tag as pointed out by @shvaikalesh.

A small nit pick, thou. I believe this test file should be at [test/built-ins/Symbol/prototype/description](https://github.com/tc39/test262/tree/main/test/built-ins/Symbol/prototype/description) and that folder might be missing other tests to capture description from ToString coerced values.

The decision on the file location might be subjective and it's ok to not go through that path.

jugglinmike:
Thanks for reviewing this!

> just missing the feature tag as pointed out by @shvaikalesh.

All set

> I believe this test file should be at test/built-ins/Symbol/prototype/description

Yeah, the test necessarily uses two different APIs, making the placement a little ambiguous. I chose the `for/` directory because we're specifically interested in how the `[[Description]]` internal slot is defined, not how the `[[Description]]` internal slot is referenced (the existing tests verify the latter).

> that folder might be missing other tests to capture description from ToString coerced values.

I can help with that. I don't see any string coercion in that algorithm, though--could you say more about what you have in mind?

rwaldron:
@leobalter are you good with @jugglinmike's response?

leobalter:
Yes, I just don't have the bandwidth to expand my thoughts on the ToString coercion, but hopefully I can explore it in a near future.

jugglinmike:
Thanks, @leobalter! Would you mind formally registering your approval using the GitHub.com UI? At some point, this became mandatory for Test262 pull requests.

leobalter (approved):

PR #2971, opened 2021-04-09
Fix typo in private getter/setter tests

strager:
Several tests for getters and setters claim to check for an early
SyntaxError regarding mixing static and non-static propeties with the
same name. However, the tests trigger another issue: the getters have no
method body; they're missing curlies.

Fix the tests to test only the intended SyntaxError, not unrelated
SyntaxError-s.


jugglinmike (approved):
Nice find! This exposes an error in QuickJS: version 2021-03-27 doesn't report the early errors. Thanks!

PR #2973, opened 2021-04-10
Test Array's toString() with non-callable "join"

shvaikalesh:
JSC bug: [Array's `toString()` is incorrect if `join()` is non-callable](https://bugs.webkit.org/show_bug.cgi?id=224215).

### DIFF
--- a/test/built-ins/Array/prototype/toString/non-callable-join-string-tag.js
+++ b/test/built-ins/Array/prototype/toString/non-callable-join-string-tag.js
@@ -12,16 +12,22 @@ info: |
     2. Let func be ? Get(array, "join").
     3. If IsCallable(func) is false, set func to the intrinsic function %Object.prototype.toString%.
     4. Return ? Call(func, array).
-features: [Symbol.toStringTag, Proxy, Reflect]
+features: [Symbol.toStringTag, Proxy, Reflect, BigInt]
 ---*/
 
 assert(delete Object.prototype.toString);
 
-Array.prototype.join = {};
-assert.sameValue(Array.prototype.toString.call({}), "[object Object]");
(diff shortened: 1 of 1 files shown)


jugglinmike:
Thanks for the patch! Glad to be improving coverage for such a well-established built-in.

The modifications to the Array prototype do not influence the test. The `toString` algorithm references the property named `join` on the running function's `this` value. The assertions pass as written because that property's value is `undefined` on all of the provided objects.

For this test, it seems less relevant to use built-in constructors since that amounts to verifying that `join` does not exist on the built-in prototypes. It would be more clear/direct to use purpose-built plain objects as the subjects, e.g.

```js
Array.prototype.toString.call({ join: 1n })
```

The behavior of `Object.prototype.toString` doesn't seem to be in question, so we don't have to go too deep into its semantics. That said, it does seem wise to use a custom `Symbol.toStringTag` method (as some of your assertions already do) in order to verify that implementations are passing along the `this` value. Using different constructors introduces an axis which appears to be relevant but is not (in other words: we're not interested in testing every combination of constructor and `join` value type).

shvaikalesh:
Thank you for review & detailed feedback!

> The modifications to the Array prototype do not influence the test.

Oops, that's embarrassing. I've removed the modifications and added tests for different `"join"` values separately from built-in constructors.

> in other words: we're not interested in testing every combination of constructor and join value type

That is well put! 👍

I think it would be helpful to keep all the pre-ES6 built-ins with special slots to ensure that runtimes perform full version of `Object.prototype.toString`, and not some part of it like JSC did.

jugglinmike (approved):
Thanks, @shvaikalesh -- this version is much more clear!

PR #2976, opened 2021-04-23
Increase precision of assertions for Error Cause

jugglinmike:
Minimize the code provided to the `assert.throws` utility in order to
reduce the possibility of false positives and to improve failure
messages in non-conforming runtimes.

---

@legendecas here's the improvement that [I mentioned earlier](https://github.com/tc39/test262/pull/2965#pullrequestreview-643575028).


legendecas (approved):
LGTM, thanks a lot for the follow-up improvements!

rwaldron (approved):

PR #2982, opened 2021-05-04
Add a test for `for await (async of`

### LINKED ISSUE #2979: Missing tests for `for await (async of`
It was recently noticed (https://github.com/babel/babel/issues/13235) that Babel failed to parse this code:

```js
async function f() {
    for await (async of []);
}
```

Unlike a regular for-of loop there is no ambiguity with `for (async of => {};;)`, and an async function expression is not legal here, so `async` can be used as the loop variable.

But it's easy for the implementation to mess up by parsing `async of` as an async function anyway, or by incorrectly sharing code that prohibits `for (async of` in a regular for-of.

Zalathar:
This adds a test for `for await (async of`, since the identifier `async` is legal here.

Fixes #2979; motivated by https://github.com/babel/babel/issues/13235.

### DIFF
--- a/test/language/statements/for-await-of/head-lhs-async.js
+++ b/test/language/statements/for-await-of/head-lhs-async.js
@@ -1,0 +1,23 @@
+// Copyright (C) 2021 Stuart Cook. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-for-in-and-for-of-statements
+description: >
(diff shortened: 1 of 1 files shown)


rwaldron (commented):
LGTM, however @jugglinmike may want to see this as a set of templates/cases.

jugglinmike (approved):
A single hand-written test seems appropriate here. Nice work, @Zalathar, and thank you!

PR #2983, opened 2021-05-05
Add more tests for `for (async of` edge-cases

Zalathar:
This adds tests for two more edge-cases related to the `for (async of` restriction (https://github.com/tc39/ecma262/pull/2256):

```js
for (\u0061sync of [7]);
```

In this case, the identifier `async` is written with an escape sequence. This avoids the `async of` lookahead, which only considers non-escaped words.

This detects implementations that check for an `async` token without considering whether it was escaped.

Recently fixed in V8: https://bugs.chromium.org/p/v8/issues/detail?id=11722

```js
for ((async) of [7]);
```

In this case, the identifier token `async` is preceded by a `(` token, which avoids the `async of` lookahead because `async` is no longer the first token in the head.

This detects implementations that check whether the LHS is an identifier at the AST level, without considering parentheses.

Motivated by https://github.com/babel/babel/pull/13244#pullrequestreview-649786288.

### DIFF
--- a/test/language/statements/for-of/head-lhs-async-escaped.js
+++ b/test/language/statements/for-of/head-lhs-async-escaped.js
@@ -1,0 +1,18 @@
+// Copyright (C) 2021 Stuart Cook. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-for-in-and-for-of-statements
+description: >
+  The left-hand-side of a for-of loop may be the identifier `async` written
+  with an escape sequence.
+info: |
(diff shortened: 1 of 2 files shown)


jugglinmike (approved):
Very nice. Thanks for revisiting this!

Vindiesel25:
- [ ]

Zalathar:
A CLA should be fine, but where can I find a privacy policy for the sensitive personal information required by the form?

jugglinmike:
@Zalathar I can't answer your questions myself, but I'm working to find someone who can (hence the delay). I'll keep you posted.

Issue #2984, opened 2021-05-06
Need tests for "Locale Info" API


FrankYFTang:
https://github.com/tc39/proposal-intl-locale-info advanced to Stage 3 in TC39 2021-Apr meeting. We need to add tests for this in test 262

@ryzokuken  @sffc

FrankYFTang:
v8 implementation is behind --harmony_intl_locale_info once we complete the landing of https://chromium-review.googlesource.com/c/v8/v8/+/2570218

jugglinmike:
@ryzokuken is in the process of contributing these tests. GitHub will not allow me to assign this issue to them, so I'm assigning it to myself to communicate that it's being worked on (and avoid duplicating efforts).

FrankYFTang:
I think this should be reopen. The test @ryzokuken  put in on #2987 is barely minimum and not enough to test the full functionality of the intended proposal. It is just the first step.

FrankYFTang:
somehow I cannot reopen this issue.

ryzokuken:
@FrankYFTang @jugglinmike said in #2987 

> Thanks for the patch! When we land it, I'll create a tracking issue for the follow-up you had in mind so folks know that it's a work-in-progress.

Would that work fine for you?

FrankYFTang:
> @FrankYFTang @jugglinmike said in #2987
> 
> > Thanks for the patch! When we land it, I'll create a tracking issue for the follow-up you had in mind so folks know that it's a work-in-progress.
> 
> Would that work fine for you?

Isn't https://github.com/tc39/test262/pull/2987 landed already?

sffc:
@ryzokuken Can you please link to the follow up issue?

ryzokuken:
I had landed, yes. I'm not sure if @jugglinmike has made the issue yet, but I'm confident that they will soon...

jugglinmike:
This issue ought to do just fine--reopening!

(I didn't recognize that GitHub would automatically close this issue when we merged gh-2987. That happened because the pull request's description included the text "Fixes: #2984" which is not entirely accurate).

FrankYFTang:
Here is what we have so far https://github.com/tc39/test262/search?q=Intl.Locale-info

FrankYFTang:
Could we identifiy what area require improvement before we close this issue?

ryzokuken:
[...]

PR #2987, opened 2021-05-11
Add basic tests for proposal-intl-locale-info

ryzokuken:
Add the basic tests (`name.js`, `branding.js` and `prop-desc.js`) for the Intl LocaleInfo proposal.

The existing getters on the `Intl.Locale` prototype only have these basic tests as well, so I think this should be sufficient for unblocking implementations. That said, I will try to add more comprehensive tests for these getters.

/cc @FrankYFTang

Fixes: #2984

### DIFF
--- a/features.txt
+++ b/features.txt
@@ -84,6 +84,10 @@ Intl.ListFormat
 # https://github.com/tc39/proposal-intl-locale
 Intl.Locale
 
+# Intl.Locale Info
+# https://github.com/tc39/proposal-intl-locale-info
+Intl.Locale-info
+
 # Intl.RelativeTimeFormat
 # https://github.com/tc39/proposal-intl-relative-time
 Intl.RelativeTimeFormat
--- a/test/intl402/Locale/prototype/calendars/branding.js
+++ b/test/intl402/Locale/prototype/calendars/branding.js
@@ -10,7 +10,7 @@ info: |
 
     2. If Type(loc) is not Object or loc does not have an [[InitializedLocale]] internal slot, then
         a. Throw a TypeError exception.
-features: [Intl.Locale]
+features: [Intl.Locale,Intl.Locale-info]
 ---*/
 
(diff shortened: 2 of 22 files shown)


Ms2ger (dismissed):

FrankYFTang:
Since Intl Locale info is a newer proposal than 
```
# Intl.Locale
# https://github.com/tc39/proposal-intl-locale
Intl.Locale
```
I suggest you add to 
https://github.com/tc39/test262/blob/main/features.txt
```
# Intl.Locale-info
# https://github.com/tc39/proposal-intl-locale-info
Intl.Locale-info
```

and change
```
features: [Intl.Locale]
```
to
```
features: [Intl.Locale,Intl.Locale-info]
```
so each browser can decide which flag to turn in their test framework based on that.

ryzokuken:
@FrankYFTang thanks, done. PTAL.

FrankYFTang:
LGTM. I think we need more test to ensure the return data type fit the spec but that could come from a later PR.

ryzokuken:
> I think we need more test to ensure the return data type fit the spec but that could come from a later PR.

That's precisely what I'm planning :smile:

jugglinmike (commented):
Thanks for the patch! When we land it, I'll create a tracking issue for the follow-up you had in mind so folks know that it's a work-in-progress.

Runtimes that don't implement this API will spuriously pass the "branding" tests--they will report a `TypeError` due to `propdesc.call` being undefined. Would you mind adding an assertion to guard against that?

```
assert.sameValue(typeof propdesc.get, "function");
```

ryzokuken:
@jugglinmike thanks, done, check it out. Should I make another PR to add those for the original `Intl.Locale` getters too?

FrankYFTang:
LGTM

FrankYFTang (approved):

jugglinmike:
> @jugglinmike thanks, done, check it out. Should I make another PR to add those for the original `Intl.Locale` getters too?

That would be wonderful!

jugglinmike (approved):

PR #2980, opened 2021-05-03
Add tests for import assertions

jugglinmike:
"Import Assertions" is a proposal that is currently in stage 3 of TC39's standards process: https://github.com/tc39/proposal-import-assertions/

PR #2990, opened 2021-05-14
Remove invalid test

jugglinmike:
@rwaldron I can't figure out how this test made it to `main`. We definitely removed it from gh-2980 via 92cf109, but it's somehow present in the final commit, 5ae04619a9348f9156382eb412e7b8458b8d3d30. Hinky

### DIFF
--- a/test/rejected.js
+++ b/test/rejected.js
@@ -1,1 +1,0 @@
-Promise.reject("anything");


leobalter (approved):
GitHub is not being helpful here, the diff shows like the line has been removed, but UI is unclear if the file remains.

I checked this branch locally and the file is definitely not there anymore.

PR #2995, opened 2021-05-25
Add tests for Object.hasOwn

jamiebuilds:
Proposal: https://github.com/tc39/proposal-accessible-object-hasownproperty

This proposal just reached stage 3, its not currently published in the draft but I've used `esid: sec-object.hasown` since that's what it's very likely going to end up being.

This PR is largely a copy-and-paste job from the old `Object.prototype.hasOwnProperty` tests, but I've tried to modernize anything I can.

### DIFF
--- a/test/built-ins/Object/hasOwn/hasown.js
+++ b/test/built-ins/Object/hasOwn/hasown.js
@@ -15,13 +15,5 @@ author: Jamie Kyle
 features: [Object.hasOwn]
 ---*/
 
-//CHECK#1
-if (typeof Object.hasOwn !== "function") {
-  $ERROR('#1: hasOwn method is defined');
-}
-
-//CHECK#2
-if (!(Object.hasOwn(Object, "hasOwn"))) {
-  $ERROR('#2: hasOwn method works properly');
-}
-//
+assert.sameValue(typeof Object.hasOwn, 'function');
+assert.sameValue(Object.hasOwn(Object, 'hasOwn'), false);
--- a/test/built-ins/Object/hasOwn/hasown.js
+++ b/test/built-ins/Object/hasOwn/hasown.js
@@ -1,0 +1,27 @@
+// Copyright 2021 Jamie Kyle.  All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-object.hasown
+info: |
+    Object.hasOwn ( _O_, _P_ )
+
+    1. Let _obj_ be ? ToObject(_O_).
+    2. Let _key_ be ? ToPropertyKey(_P_).
+    3. Return ? HasOwnProperty(_obj_, _key_).
+description: >
+    Checking type of the Object.hasOwn and the returned result
+author: Jamie Kyle
+features: [Object.hasOwn]
+---*/
+
+//CHECK#1
+if (typeof Object.hasOwn !== "function") {
+  $ERROR('#1: hasOwn method is defined');
+}
+
+//CHECK#2
+if (!(Object.hasOwn(Object, "hasOwn"))) {
+  $ERROR('#2: hasOwn method works properly');
+}
+//
--- a/test/built-ins/Object/hasOwn/hasown.js
+++ b/test/built-ins/Object/hasOwn/hasown.js
@@ -16,4 +16,4 @@ features: [Object.hasOwn]
 ---*/
 
 assert.sameValue(typeof Object.hasOwn, 'function');
-assert.sameValue(Object.hasOwn(Object, 'hasOwn'), false);
+assert(Object.hasOwn(Object, 'hasOwn'));
--- a/test/built-ins/Object/hasOwn/hasown_inherited_getter_and_setter.js
+++ b/test/built-ins/Object/hasOwn/hasown_inherited_getter_and_setter.js
@@ -14,7 +14,7 @@ var base = {
   get foo() {
     return 42;
   },
-  set foo(x) {;
+  set foo(x) {
   }
 };
 var o = Object.create(base);
--- a/test/built-ins/Object/hasOwn/hasown_inherited_getter_and_setter.js
+++ b/test/built-ins/Object/hasOwn/hasown_inherited_getter_and_setter.js
@@ -1,0 +1,22 @@
+// Copyright (c) 2021 Jamie Kyle.  All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-object.hasown
+description: >
+    Properties - [[HasOwnProperty]] (literal inherited getter/setter
+    property)
+author: Jamie Kyle
+features: [Object.hasOwn]
+---*/
+
+var base = {
+  get foo() {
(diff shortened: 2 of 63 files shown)


test/built-ins/Object/hasOwn/length.js
```diff
@@ -0,0 +1,31 @@
...
+  Unless otherwise specified, the "length" property of a built-in Function
+  object has the attributes { [[Writable]]: false, [[Enumerable]]: false,
+  [[Configurable]]: true }.
+includes: [propertyHelper.js]
+author: Jamie Kyle
+features: [Object.hasOwn]
+---*/
+
+verifyProperty(Object.hasOwn, "length", {
+  value: 1,
```
ljharb, line 27: the length of this function must be 2, since it takes two arguments
```suggestion
  value: 2,
```

test/built-ins/Object/hasOwn/name.js
```diff
@@ -0,0 +1,28 @@
...
+
+    Unless otherwise specified, the name property of a built-in Function
+    object, if it exists, has the attributes { [[Writable]]: false,
+    [[Enumerable]]: false, [[Configurable]]: true }.
+includes: [propertyHelper.js]
+author: Jamie Kyle
+features: [Object.hasOwn]
+---*/
+
+assert.sameValue(Object.hasOwn.name, "hasOwnProperty");
```
ljharb, line 24:
```suggestion
assert.sameValue(Object.hasOwn.name, "hasOwn");
```

ljharb (changes requested):

nicolo-ribaudo (commented):
I might have missed it while reading the tests, but I think we are not testing the property descriptor of `Object.hasOwn` (something like https://github.com/tc39/test262/blob/main/test/built-ins/Object/assign/assign-descriptor.js).

jamiebuilds:
@nicolo-ribaudo nope didn't have that, added

test/built-ins/Object/hasOwn/hasown_own_getter.js
```diff
@@ -0,0 +1,17 @@
...
+features: [Object.hasOwn]
+---*/
+
+var o = {
+  get foo() {
+    return 42;
+  }
+};
+
+assert(Object.hasOwn(o, "foo"), 'Object.hasOwn(o, "foo") !== true');
```
nicolo-ribaudo, line 17: It might be better to always use `assert.sameValue(Object.hasOwn(o, "foo"), true` in all these tests, to verify that the return value is actually `true` and not just truthy. You are already doing it for `assert.sameValue(Object.hasOwn(o, "foo"), false`.

nicolo-ribaudo: Nevermind, `assert` already does that.

jamiebuilds: oh lol, well I updated it to use `assert.sameValue` with `true`

test/built-ins/Object/hasOwn/toobject_before_topropertykey.js
```diff
@@ -0,0 +1,48 @@
...
+    callCount1++;
+    throw new Test262Error();
+  },
+  get valueOf() {
+    callCount1++;
+    throw new Test262Error();
+  },
+};
+
+assert.throws(Test262Error, function() {
```
jugglinmike, line 32: Here and below, don't we expect the TypeError from ToObject?

jugglinmike (changes requested):
Thanks for contributing these tests, Jamie!

test/built-ins/Object/hasOwn/hasown_inherited_getter_and_setter.js
```diff
@@ -0,0 +1,22 @@
...
+    property)
+author: Jamie Kyle
+features: [Object.hasOwn]
+---*/
+
+var base = {
+  get foo() {
+    return 42;
+  },
+  set foo(x) {;
```
rwaldron, line 17: lol, what is that semi-colon doing here? 🤣

rwaldron (commented):
There's a lot of test material here, which I generally really appreciate, but there are things being tested that aren't relevant to property "own"-ness.

jamiebuilds:
> There's a lot of test material here, which I generally really appreciate, but there are things being tested that aren't relevant to property "own"-ness.

I copied I think all of these tests from [test/built-ins/Object/prototype/hasOwnProperty](https://github.com/tc39/test262/tree/main/test/built-ins/Object/prototype/hasOwnProperty), I'm not entirely sure which ones should be removed.

rwaldron:
Well, then it seems that directory is filled with tests that are irrelevant to "own"-ness. We can deal with that later.

rwaldron:
Everyone that's reviewed should come back and resolve or re-review for @jamiebuilds

rwaldron (approved):

codehag:
Pinging this, in the hopes that it gets the reviews updated. @jugglinmike i think yours is the only request changes rn?

test/built-ins/Object/hasOwn/hasown.js
```diff
@@ -15,13 +15,5 @@ author: Jamie Kyle
...
-  $ERROR('#1: hasOwn method is defined');
-}
-
-//CHECK#2
-if (!(Object.hasOwn(Object, "hasOwn"))) {
-  $ERROR('#2: hasOwn method works properly');
-}
-//
+assert.sameValue(typeof Object.hasOwn, 'function');
+assert.sameValue(Object.hasOwn(Object, 'hasOwn'), false);
```
jugglinmike, line 19: I think we want

```diff
-assert.sameValue(Object.hasOwn(Object, 'hasOwn'), false);
+assert(Object.hasOwn(Object, 'hasOwn'));
```

test/built-ins/Object/hasOwn/prototype.js
```diff
@@ -10,8 +10,4 @@ author: Jamie Kyle
 features: [Object.hasOwn]
 ---*/
 
-//CHECK#1
-if (Object.hasOwn.prototype !== undefined) {
-  $ERROR('#1: Object.hasOwn has not prototype property' + Object.hasOwn.prototype);
-}
-//
+assert.sameValue(Object.hasOwn.prototype !== undefined, true);
```
jugglinmike, line 13: I think we want

```diff
-assert.sameValue(Object.hasOwn.prototype !== undefined, true);
+assert.sameValue(Object.hasOwn.prototype, undefined);
```

jugglinmike (changes requested):
We're close! It just looks like a couple small bugs slipped in through the latest refactoring.

jamiebuilds:
Okay @jugglinmike I've updated those test cases, this should now be good to go

jugglinmike (approved):
Looks good to me! Nice work :)

By the way, the CI failures are about failing tests. [We recently configured the system to tolerate test failures](https://github.com/tc39/test262/pull/2998), so my guess is that this was caused by the base branch being behind the upstream `main` branch.

PR #2999, opened 2021-05-28
Add tests for JSON modules

jugglinmike:
Designate the files which are intended to be interpreted as JSON modules
via a distinct file name suffix. Differentiating JSON files in this way
will assist consumers which require special handling of such files (e.g.
web browsers).

Change the pattern used to designate "fixture" files so that it may be
applied to files used for JSON modules.

Increment the project version number to alert consumers of this change
in interpreting instructions.


rwaldron (changes requested):
Please check linting errors

jugglinmike:
Got it Rick. I've split the infrastructural changes into a dedicated commit which precedes the commit introducing the "JSON modules" tests.

rwaldron (approved):

wangxiang365:
****

PR #3000, opened 2021-05-28
Don't use function argument destructuring in regExpUtils.js

linusg:
This increases compatibility with less advanced engines that only have partial support for recent language features like assignment patterns.

### DIFF
--- a/harness/regExpUtils.js
+++ b/harness/regExpUtils.js
@@ -6,7 +6,12 @@ description: |
 defines: [buildString, testPropertyEscapes, matchValidator]
 ---*/
 
-function buildString({ loneCodePoints, ranges }) {
+function buildString(args) {
+  // Use member expressions rather than destructuring `args` for improved
+  // compatibility with engines that only implement assignment patterns
+  // partially or not at all.
+  const loneCodePoints = args.loneCodePoints;
+  const ranges = args.ranges;
   const CHUNK_SIZE = 10000;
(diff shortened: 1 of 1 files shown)


jugglinmike (approved):
I was recently surprised to find how many ES6-isms have crept into the harness files. Out of curiosity: which engine is it that implements `let` and `const` but not function parameter destructuring?

linusg:
In this case that's [SerenityOS](https://github.com/SerenityOS/serenity)'s [LibJS](https://github.com/SerenityOS/serenity/tree/master/Userland/Libraries/LibJS).

Since (from-scratch) development started just about a year ago, there's not the usual progression of catching up with spec changes over time but instead we're implementing both old (ES5 support is pretty good by now) and new features (there's bleeding edge stuff like `Object.hasOwn()`, nullish coalescing, logical nullish assignment) at the same time, hence why it's a bit of a mixed bag right now :)

p-bakker:
Sry for hijacking this issue, but have a related question: as @jugglinmike mentioned above, a lot of ES6-isms have crept into the harness files, like newer regex flags, default param values, Template Literals, rest parameters etc etc.

This means lesser compatible engines cannot run (the latest) tests. In our case: Rhino

While we're trying to get closer to spec compliance, we're now looking a a ton of tests that fail, but some just fail because the harness code fails.

Shouldn't the harness code stick to either ES3 or ES5?

PR #3006, opened 2021-06-07
minor fixes to frontmatter

jmdyck:


jugglinmike (approved):
Thanks! We ought to be able to automate some of this--I'll follow up with a couple extensions to the linter

PR #3008, opened 2021-06-10
Fix tests on {localeMatcher: "lookup"}

FrankYFTang:
This PR fix the harness by passing the option for getLocaleSupportInfo instead of an internal hard coded {localeMatcher:  "lookup"} . It also fix 3 tests which have strong assumption based on only "lookup" localeMatcher
1. intl402/Segmenter/constructor/supportedLocalesOf/locales-specific.js

The expectation that "sr-Thai-RS" would be returned is only true with the 
9.2.2 BestAvailableLocale ( availableLocales, locale )
https://tc39.es/ecma402/#sec-bestavailablelocale
algorithm used by 9.2.3 LookupMatcher ( availableLocales, requestedLocales )
https://tc39.es/ecma402/#sec-lookupmatcher

The default for localeMatcher is "best fit" but not "lookup" for all Intl objects.

And for 9.2.4 BestFitMatcher ( availableLocales, requestedLocales )
https://tc39.es/ecma402/#sec-bestfitmatcher
It may not match "sr-Thai-RS" for "sr"  and return ["de", "zh-CN"] instead. Therefore, we need to change this test to only test on {localeMatcher: "lookup"}

2. The intl402/fallback-locales-are-supported.js is only valid for "lookup" localeMatcher since the "best fit" algorithm does not guarantee of the fallback assumption
3. The intl402/supportedLocalesOf-consistent-with-resolvedOptions.js and  supportedLocalesOf-unicode-extensions-ignored.js
Passing the localeMatcher to get the correct expectation.

### DIFF
--- a/test/intl402/supportedLocalesOf-unicode-extensions-ignored.js
+++ b/test/intl402/supportedLocalesOf-unicode-extensions-ignored.js
@@ -11,20 +11,18 @@ includes: [testIntl.js]
 ---*/
 
 testWithIntlConstructors(function (Constructor) {
-    var info = getLocaleSupportInfo(Constructor);
 
     // this test should work equally for both matching algorithms
     ["lookup", "best fit"].forEach(function (matcher) {
+        var opt = {localeMatcher: matcher};
+        var info = getLocaleSupportInfo(Constructor, opt);
         var allLocales = info.supported.concat(info.byFallback, info.unsupported);
         allLocales.forEach(function (locale) {
             var validExtension = "-u-co-phonebk-nu-latn";
             var invalidExtension = "-u-nu-invalid";
-            var supported1 = Constructor.supportedLocalesOf([locale],
-                {localeMatcher: matcher});
-            var supported2 = Constructor.supportedLocalesOf([locale + validExtension],
-                {localeMatcher: matcher});
-            var supported3 = Constructor.supportedLocalesOf([locale + invalidExtension],
-                {localeMatcher: matcher});
+            var supported1 = Constructor.supportedLocalesOf([locale], opt);
+            var supported2 = Constructor.supportedLocalesOf([locale + validExtension], opt));
+            var supported3 = Constructor.supportedLocalesOf([locale + invalidExtension], opt));
             if (supported1.length === 1) {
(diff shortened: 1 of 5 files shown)


FrankYFTang:
@anba  @Ms2ger @ryzokuken  @sffc @gibson042 @leobalter @ljharb

FrankYFTang:
v8 tracking https://bugs.chromium.org/p/v8/issues/detail?id=11868 for launching https://docs.google.com/document/d/1cPGfiihn76yj2iAomKcspPFyLLcnk3WkCiqceBQPQyk/edit#

FrankYFTang:
Agree

FrankYFTang:
PTAL

FrankYFTang:
@rwaldron

ryzokuken (approved):
LGTM, thanks.

FrankYFTang:
I don't have commit right, please land it for me if you can review and commit it. Thanks

rwaldron (approved):

test/intl402/supportedLocalesOf-unicode-extensions-ignored.js
```diff
@@ -11,20 +11,18 @@ includes: [testIntl.js]
...
             var invalidExtension = "-u-nu-invalid";
-            var supported1 = Constructor.supportedLocalesOf([locale],
-                {localeMatcher: matcher});
-            var supported2 = Constructor.supportedLocalesOf([locale + validExtension],
-                {localeMatcher: matcher});
-            var supported3 = Constructor.supportedLocalesOf([locale + invalidExtension],
-                {localeMatcher: matcher});
+            var supported1 = Constructor.supportedLocalesOf([locale], opt);
+            var supported2 = Constructor.supportedLocalesOf([locale + validExtension], opt));
+            var supported3 = Constructor.supportedLocalesOf([locale + invalidExtension], opt));
```
iamstolis, line 25: This test is not syntactically correct (there is a redundant closing parenthesis on this and the previous line).

FrankYFTang: sorry, I think it just got fixed in https://github.com/tc39/test262/commit/356afd79a81e232a3d1fe452dbe1e79ecdc17d38#diff-29fe459e743829b8576655c217425f9c83220c67d90264cfc55a959b8c2c6350

Issue #3002, opened 2021-06-04
Testing the Temporal proposal


jugglinmike:
[The Temporal proposal](https://github.com/tc39/proposal-temporal) reached Stage 3 on [2021-03-10](https://github.com/tc39/notes/blob/master/meetings/2021-03/mar-10.md), and like all ECMA262 proposals, it needs coverage in Test262 to advance to Stage 4. In terms of normative text, it is easily the largest proposal we've seen. Testing it will definitely require significant time, effort, and coordination (even with [the proposal's existing tests](https://github.com/tc39/proposal-temporal/tree/main/polyfill/test) as a starting point). We'd like to document progress publicly for transparency and to encourage participation.

To get the ball rolling, I'm suggesting that we track progress on an per-API basis using this GitHub.com issue. Compared to alternatives like [collaborative document editing services](https://sheets.google.com/) and an in-repository text file, this seems like it balances ease of hosting, maintainability, and proximity to the tests. [This approach worked well for ES6's well-known Symbols](https://github.com/bocoup/test262/issues/31)--the most similar (though much smaller) effort that comes to mind.

**Working mode**

Anyone can comment here when they'd like to begin writing tests for a specific API. A maintainer will update the list to reflect the ongoing and completed work. When you submit a pull request, please mention this issue (just write "gh-3002") so GitHub displays the status of your work in this thread.

Tests for multiple APIs can be submitted in a single pull request, but patches should be small enough for another contributor to review them in a single session. For instance, it probably makes sense to submit tests for all of the `@@toStringTag` methods in one patch. There's some subjectivity here, but when in doubt, we should err on the side of smaller patches (and remain open to reviewers' requests to break up pull requests).

**Status**

- [ ] Temporal.now.timeZone
- [ ] Temporal.now.instant
- [ ] Temporal.now.plainDateTime
- [ ] Temporal.now.plainDateTimeISO
- [ ] Temporal.now.zonedDateTime
- [ ] Temporal.now.zonedDateTimeISO
- [ ] Temporal.now.plainDate
- [ ] Temporal.now.plainDateISO
- [ ] Temporal.now.plainTimeISO
- [ ] Temporal.PlainDate
- [ ] Temporal.PlainDate.from
- [ ] Temporal.PlainDate.compare
- [ ] Temporal.PlainDate.prototype.constructor
- [ ] Temporal.PlainDate.prototype[ @@toStringTag ]
- [ ] Temporal.PlainDate.prototype.toPlainYearMonth
[...]

PR #3011, opened 2021-06-10
Add tests for Temporal.now.timezone

jugglinmike:
In service of gh-3002

/cc @ptomato @Ms2ger

### DIFF
--- a/test/built-ins/Temporal/now/timeZone/extensible.js
+++ b/test/built-ins/Temporal/now/timeZone/extensible.js
@@ -1,0 +1,15 @@
+// Copyright (C) 2021 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.now.timezone
+description: Temporal.now.timeZone is extensible.
+info: |
+  ## 17 ECMAScript Standard Built-in Objects
+
+  Unless specified otherwise, the [[Extensible]] internal slot
+  of a built-in object initially has the value true.
+features: [Temporal]
+---*/
+
+assert(Object.isExtensible(Temporal.now.timeZone));
--- a/test/built-ins/Temporal/now/timeZone/not-a-constructor.js
+++ b/test/built-ins/Temporal/now/timeZone/not-a-constructor.js
@@ -1,0 +1,21 @@
+// Copyright (C) 2021 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.now.timezone
+description: Temporal.now.timeZone does not implement [[Construct]]
+info: |
+  ECMAScript Function Objects
+
+  Built-in function objects that are not identified as constructors do not
+  implement the [[Construct]] internal method unless otherwise specified in
+  the description of a particular function.
+includes: [isConstructor.js]
+features: [Reflect.construct, Temporal]
+---*/
+
+assert.sameValue(isConstructor(Temporal.now.timeZone), false, 'isConstructor(Math.min) must return false');
+
+assert.throws(TypeError, () => {
+  new Temporal.now.timeZone();
+}, '`new Temporal.now.timeZone()` throws TypeError');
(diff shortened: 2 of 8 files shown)


test/built-ins/Temporal/now/timeZone/not-a-constructor.js
```diff
@@ -0,0 +1,21 @@
...
+  ECMAScript Function Objects
+
+  Built-in function objects that are not identified as constructors do not
+  implement the [[Construct]] internal method unless otherwise specified in
+  the description of a particular function.
+includes: [isConstructor.js]
+features: [Reflect.construct, Temporal]
+---*/
+
+assert.sameValue(isConstructor(Temporal.now.timeZone), false, 'isConstructor(Math.min) must return false');
```
Ms2ger, line 17:
```suggestion
assert.sameValue(isConstructor(Temporal.now.timeZone), false, 'isConstructor(Temporal.now.timeZone) must return false');
```

Ms2ger (approved):

test/built-ins/Temporal/now/timeZone/extensible.js
```diff
@@ -0,0 +1,15 @@
...
+description: Temporal.now.timeZone is extensible.
+info: |
+  ## 17 ECMAScript Standard Built-in Objects
+
+  Unless specified otherwise, the [[Extensible]] internal slot
+  of a built-in object initially has the value true.
+features: [Temporal]
+---*/
+
+assert(Object.isExtensible(Temporal.now.timeZone));
```
Ms2ger, line 15:
```suggestion
assert.sameValue(Object.isExtensible(Temporal.now.timeZone), true);
```

test/built-ins/Temporal/now/timeZone/not-a-constructor.js
jugglinmike: Whoops! Thanks

test/built-ins/Temporal/now/timeZone/extensible.js
jugglinmike: This looks equivalent to me. Could you say a little about why we ought to use `assert.sameValue` instead of `assert`?

Ms2ger: I understood the former to be preferred in test262, but I may be misremembering. Happy to go either way.

rwaldron: @jugglinmike you're right, this is fine as is. 

@Ms2ger `assert()` is discouraged in cases such as `assert(foo === "something interesting")`, which is better written as `assert.sameValue(foo, "something interesting")` because you'll get a more meaningful error message.

test/built-ins/Temporal/now/timeZone/return-value.js
```diff
@@ -0,0 +1,15 @@
...
+description: Temporal.now.timeZone returns an instance of the TimeZone constructor
+info: |
+  1. Return ? SystemTimeZone().
+features: [Temporal]
+---*/
+
+assert.sameValue(
+  Object.getPrototypeOf(Temporal.now.timeZone()),
+  Temporal.TimeZone.prototype
+);
```
rwaldron, line 15: This test just tell me that those two things are the same thing, but explicitly that either of them is an instance of `Temporal.TimeZone`. 

`assert(Temporal.now.timeZone() instanceof Temporal.TimeZone)`?

rwaldron (changes requested):

test/built-ins/Temporal/now/timeZone/return-value.js
jugglinmike: The instanceof operator has a lot of additional semantics which would make this test less precise. For example, it would be satisfied if the return value was an instance of a subclass of `Temporal.TimeZone`. I believe a strict equality check is the most precise way to verify this behavior.

Ms2ger: We could have both?

test/built-ins/Temporal/now/timeZone/extensible.js
Ms2ger: Sounds like I misremembered, thanks for the clarification. I couldn't find this discouragement in the documentation at https://github.com/tc39/test262/blob/main/CONTRIBUTING.md#test-environment - could you add it?

rwaldron (approved):

rwaldron:
Rebased locally and merged

PR #3013, opened 2021-06-20
Add more static class property ordering tests

TimothyGu:
Since https://github.com/tc39/ecma262/pull/1490, the "length" and "name" properties of a class are defined before any static methods. This is tested by #2057, in test/language/computed-property-names of all places.

At the same time, static methods with "name" as the name would overwrite the original property, but retain the original property enumeration order. This was not previously tested. In fact, the overwriting behavior was not tested at all for the "length" property.

This commit mends both holes in test coverage.

### DIFF
--- a/test/language/statements/class/definition/fn-length-static-precedence.js
+++ b/test/language/statements/class/definition/fn-length-static-precedence.js
@@ -4,19 +4,22 @@
 /*---
 esid: sec-runtime-semantics-classdefinitionevaluation
 description: >
-    Function `name` attribute not inferred in presence of static `name` method
+    Function `length` attribute not inferred in presence of static `length` method
 info: |
     ClassTail : ClassHeritage_opt { ClassBody_opt }
 
     14. If constructor is empty, then [...]
       b. Let F be ! CreateBuiltinFunction(steps, 0, className, « [[ConstructorKind]], [[SourceText]] », empty, constructorParent).
-    15. Else, [...]
-      d. Perform ! SetFunctionName(F, className).
+    15. Else,
+      a. Let constructorInfo be ! DefineMethod of constructor with arguments proto and constructorParent.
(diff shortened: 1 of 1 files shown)


TimothyGu:
XS and engine262 pass both tests. ChakraCore, GraalJS, JSC, SpiderMonkey, and V8 fail.

test/language/statements/class/definition/fn-length-static-precedence.js
```diff
@@ -0,0 +1,82 @@
...
+includes: [compareArray.js]
+features: [generators]
+---*/
+
+class A {
+  static method() {
+    $ERROR('Static method should not be executed during definition');
+  }
+  static length() {
+    $ERROR('Static method should not be executed during definition');
```
rwaldron, line 32: Please don't use `$ERROR` like this. `throw new Test262Error(...)` (This is clearly stated in the contributing docs https://github.com/tc39/test262/blob/main/CONTRIBUTING.md#test-environment)

rwaldron (changes requested):

rwaldron:
FWIW, I've updated the linter to check for $ERROR uses.

test/language/statements/class/definition/fn-length-static-precedence.js
TimothyGu: This is primarily copied from the existing fn-name-static-precedence.js test, which explains the existence of $ERROR. I can fix it up.

rwaldron: I appreciate the follow up. I just finished eliminating `$ERROR` from every test file in the project 🤣

rwaldron:
@TimothyGu please @ mention me when you're ready for another review.

TimothyGu:
@rwaldron This is ready for another look.

rwaldron (approved):

PR #3016, opened 2021-06-23
Add 'import-assertions' feature tag to json modules tests

dandclark:
The language/import/json-* tests depend on syntax from the Stage 3 import assertions proposal, so these tests should have the import-assertions feature tag.

Add it to the tests where it was missing.

### DIFF
--- a/test/language/import/json-extensibility-array.js
+++ b/test/language/import/json-extensibility-array.js
@@ -5,7 +5,7 @@ esid: sec-parse-json-module
 description: Creates extensible arrays
 flags: [module]
 includes: [propertyHelper.js]
-features: [json-modules]
+features: [import-assertions, json-modules]
 ---*/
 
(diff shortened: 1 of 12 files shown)


ljharb:
Isn’t that host-dependent? node has always planned to allow json importing without an assertion.

dandclark:
> Isn’t that host-dependent? node has always planned to allow json importing without an assertion.

That's true, but all of these tests are using the import assertions syntax, e.g.
```mjs
import value from './json-value-array_FIXTURE.json' assert { type: 'json' };
```

dandclark:
I updated the PR message to specify that the tests in question are using the import assertions syntax.

ljharb:
ah, k - it’d be great to add tests that cover json modules when not using the syntax.

rwaldron (approved):

PR #3017, opened 2021-06-24
tools: remove deprecated $ERROR api

rwaldron:


jugglinmike:
This check will alert contributors about the deprecation, but since it doesn't pass for the entire test suite, it will also interrupt some other workflows.

One is local usage by contributors who aren't seasoned git users. These folks will likely pass directories to the linting tool as a way to efficiently validate their work (rather than lint all of Test262). A filter like that may include tests beyond the ones they modified, but it is likely "good enough" to reduce execution time for most cases. With this check in place, a "good enough" filter like that may fail spuriously.

We could get around this by making the linter aware of git so that it's easier for all contributors to use the more advanced filter. I personally value the simplicity of the tooling, though.

More fundamentally, I appreciate the ability to validate all of Test262 at any time, especially as someone who isn't a full-time maintainer. It's a good way to prove that mistakes haven't been accidentally merged over a long period of time.

One addition that would address both of these concerns would be formally allowing all the current infractions (looks like there's around 4,700 of them) through additions to the linter's `lint.exceptions` file. This isn't a perfect solution, though: contributors who modify an existing file will not be notified of the use of `$ERROR` within that file. (Also, we've never made such extensive use of the "exceptions" file before, so I'd be curious to see how well it performs in that case.)

What do you think?

rwaldron:
I think I should commit to fixing them all, once and for all.

jugglinmike:
If we remove all the invocations, then is the implementation still necessary? Could we also remove the implementation, the harness tests, the documentation, and the linting check?

Also, I can't recall why $ERROR is deprecated in the first place. Since that isn't documented, it might be good to explain in the commit message.

rwaldron:
@jugglinmike even though we didn't get a chance to resolve the "Why?" question this weekend, I've taken your feedback and removed all trace of `$ERROR`.

leobalter:
Guessing here, but I believe the reason to deprecate $ERROR is in order to promote support for a single consistent way to handle errors. The current recommendation does not require usage of $ERROR and as pointed out, the usage of such API might leads to erratic or confusing practices.

With some thoughts, I would even say $ERROR is not that bad, but it sounds like a good fit to be used if the test runner could eventually not bail out on the first error, but it's not the case for Test262.

The pet peeve is more related to bad practices I observed around the api and not exactly within it. This refers to tests using $ERROR as part of a syntactically composed assertion, rather than using the assertion api itself. E.g.: `if (!x) $ERROR();` instead of `assert(x)`.

Using `throw new Test262Error();` creates a better communication for the intention the test might have and sounds like a better fit in conjunction with `assert.throws`, which `$ERROR` does not translate well to the custom error constructor. @rwaldron pointed this out.

The deprecated api is not buggy or problematic, but removing it we can achieve some better consistency and cleaner code, I believe. This will also help avoid new contributions using the undesired api, as people usually repeat test content for the observed style.

jugglinmike:
This all sounds good to me--thank you @rwaldron and thank you @leobalter.

(By the way, one subtle difference between `throw` and `$ERROR` is that since the latter is an expression, it's a more versatile mechanism for producing exceptions. As it happens, `$ERROR` is exclusively used as an ExpressionStatement today. That's all the evidence we need that when it comes to throwing from expression positions, YAGNI.)

jugglinmike (changes requested):
@rwaldron I'm not going to use in-line comments for this review because the size of the patch makes loss more likely.

If we move forward with this, we should remove references to `$ERROR` from the metadata in `harness/sta.js` and `test/harness/sta.js`.

Before that, though, `test/language/line-terminators/comment-multi-cr.js` has been invalidated:

```diff
diff --git a/test/language/line-terminators/comment-multi-cr.js b/test/language/line-terminators/comment-multi-cr.js
index 98d42edca0..7e2976abea 100644
--- a/test/language/line-terminators/comment-multi-cr.js
+++ b/test/language/line-terminators/comment-multi-cr.js
@@ -10,8 +10,11 @@ description: Insert real CARRIAGE RETURN into multi line comment
 
 var x = 0;
 /*
-multi^Mline^Mcomment^Mx = 1;
+multi
+line
+comment
+x throw new Test262Error
 */
 if (x !== 0) {
-  $ERROR('#1: var x = 0; /*\\rmulti\\rline\\rcomment\\rx = 1;\\r*/ x === 0. Actual: ' + (x));
+  throw new Test262Error('#1: var x = 0; /*\\rmulti\\rline\\rcomment\\rx = 1;\\r*/ x === 0. Actual: ' + (x));
 }
```

The fix is straightforward, but honestly, the mistake makes me suspicious of the rest of the patch. There's just too much change to review manually, and while my tool-assisted approach uncovered this regression, I'm not confident that it's exhaustive.

It may be that we can't remove `$ERROR` "once and for all." We may have to commit to a longer incremental process.

rwaldron:
> It may be that we can't remove $ERROR "once and for all." We may have to commit to a longer incremental process.

I've been sitting on this for a bit, and I agree with you. I'm going to do "sections at a time".

Issue #3003, opened 2021-06-05
Testing the "Resizable ArrayBuffer" proposal


jugglinmike:
https://github.com/tc39/proposal-resizablearraybuffer

This proposal advanced to stage 3 on 2021-05-25

jugglinmike:
I've begun writing these tests

jugglinmike:
I've submitted these tests for review, split up across 11 pull requests in order to facilitate incremental review (the pull requests are listed above).

There's one design decision which effected multiple patches, so to keep things organized, I'll describe it here:

Hosts are permitted to fail any resize operation at their own discretion. This is fairly straightforward to accommodate in direct tests (e.g. those for `ArrayBuffer.prototype.resize`), but it's more difficult to address for the features which integrate.

The approach I've settled on involves dynamically assigning assertion expectation values within `try`/`catch` statements. This is most jarring in those tests which concern abrupt completions: they use `Test262Error` and `assert.throws` to verify that a given invocation does *not* throw. In all cases, the design is intended to maximize coverage (by validating *something* for the "host-defined failure" case) while limiting the amount of branching in the test body (which is generally an invitation for test bugs and reader confusion).

This is all a little unorthodox, but I feel it's warranted by the proposal's unique behavior.

/cc @syg

PR #3027, opened 2021-06-25
Resizable ArrayBuffer: TypedArray methods

jugglinmike:
This is in service of gh-3003.

The files in the second commit are highly similar. Only the test for
`TypedArray.prototype.copyWithin` was written manually. The others were
generated from that test via the following script:

    #!/bin/bash

    set -eu

    names_cb='
    every
    filter
    find
    findIndex
    forEach
    map
    reduce
    reduceRight
    some
    '
    names_num='
    at
    fill
    includes
    indexOf
    join
    lastIndexOf
    slice
    '
    names_none='
    entries
    values
    keys
    reverse
    sort
    toLocaleString
    values
    '

    for name in $(printf "${names_cb} ${names_num} ${names_none}"); do
      lower=$(echo ${name} | tr '[:upper:]' '[:lower:]')
      if echo "$names_cb" | grep -xq $name; then
        value='() => {}'
      elif echo "$names_num" | grep -xq $name; then
        value='0'
      else
        value=''
      fi

      sed \
        -e "s/copywithin/${lower}/g" \
        -e "s/copyWithin/${name}/g" \
        -e "s/${name}(.*);/${name}(${value});/g" \
        ./test/built-ins/TypedArray/prototype/copyWithin/return-abrupt-from-this-oob.js \
        > ./test/built-ins/TypedArray/prototype/${name}/return-abrupt-from-this-oob.js
    done

### DIFF
--- a/test/built-ins/TypedArray/prototype/copyWithin/return-abrupt-from-this-out-of-bounds.js
+++ b/test/built-ins/TypedArray/prototype/copyWithin/return-abrupt-from-this-out-of-bounds.js
@@ -1,0 +1,56 @@
+// Copyright (C) 2021 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-%typedarray%.prototype.copywithin
+description: Return abrupt when "this" value fails buffer boundary checks
+includes: [testTypedArray.js]
+features: [TypedArray, resizable-arraybuffer]
+---*/
+
+assert.sameValue(
+  typeof TypedArray.prototype.copyWithin,
+  'function',
+  'implements TypedArray.prototype.copyWithin'
+);
+
+assert.sameValue(
+  typeof ArrayBuffer.prototype.resize,
+  'function',
+  'implements ArrayBuffer.prototype.resize'
+);
+
+testWithTypedArrayConstructors(TA => {
+  var BPE = TA.BYTES_PER_ELEMENT;
+  var ab = new ArrayBuffer(BPE * 4, {maxByteLength: BPE * 5});
+  var array = new TA(ab, BPE, 2);
+
+  try {
+    ab.resize(BPE * 5);
+  } catch (_) {}
+
+  // no error following grow:
+  array.copyWithin(new TA(), 0);
+
+  try {
+    ab.resize(BPE * 3);
+  } catch (_) {}
+
+  // no error following shrink (within bounds):
+  array.copyWithin(new TA(), 0);
+
+  var expectedError;
+  try {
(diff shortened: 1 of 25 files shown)


rwaldron:
Love it.

test/built-ins/TypedArray/prototype/entries/return-abrupt-from-this-out-of-bounds.js
```diff
@@ -0,0 +1,56 @@
...
+
+  var expectedError;
+  try {
+    ab.resize(BPE * 3);
+    expectedError = TypeError;
+  } catch (_) {
+    // The host is permitted to fail any "resize" operation at its own
+    // discretion. If that occurs, the entries operation should complete
+    // successfully.
+    expectedError = Test262Error;
```
rwaldron, line 49: Clever

test/built-ins/TypedArray/prototype/copyWithin/return-abrupt-from-this-out-of-bounds.js
```diff
@@ -0,0 +1,56 @@
+// Copyright (C) 2021 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-%typedarray%.prototype.copywithin
+description: Return abrupt when "this" value fails buffer boundary checks
+includes: [testTypedArray.js]
+features: [TypedArray, TypedArray.prototype.at, resizable-arraybuffer]
```
rwaldron, line 7: `TypedArray.prototype.at` isn't used in this test. You can remove this (and all of the others) or I can after this is merged. Let me know what you'd like to do.

rwaldron (changes requested):

test/built-ins/TypedArray/prototype/entries/return-abrupt-from-this-out-of-bounds.js
jugglinmike: :)

jugglinmike:
> `TypedArray.prototype.at` isn't used in this test. You can remove this (and all of the others) or I can after this is merged. Let me know what you'd like to do.

Good call. I took a few minutes to extend the Bash script so that this patch can still be explained as a mechanical transformation of a single file. Rebasing over the changes to the `main` branch was a good opportunity to update the commit message with the new version of the script.

Here's the original version of the branch in case that's interesting to anyone:

https://github.com/tc39/test262/compare/main...bocoup:resizable-arraybuffer-ta-methods-orig

rwaldron (approved):

PR #3029, opened 2021-06-25
Add missing features: [resizable-arraybuffer]

rwaldron:


jugglinmike (approved):
Good catch. Looks like I missed a few others. I added a commit to fix those ones, so I'll leave merging to you

PR #3030, opened 2021-06-25
Append missing pipe character on "info:" lines

jmdyck:


jugglinmike (approved):
Sorry about that; thanks for fixing it!

PR #3033, opened 2021-06-30
Add tests for Intl.DateTimeFormat conflicting options

katemihalikova:
Tests for step 36 of [ECMA 402 v8.0 / `Intl.DateTimeFormat` / _InitializeDateTimeFormat_](https://402.ecma-international.org/8.0/#sec-initializedatetimeformat) specs.


rwaldron:
Hi @katemihalikova! Thanks for the contribution :) As a first time contributor, I'm obligated to ask if you are employed by a member org, and if not would you kindly sign the CLA: http://tc39.github.io/test262-cla

katemihalikova:
Thanks, signed 👍

rwaldron (approved):

rwaldron:
Nice work! I cross checked with the tests in https://github.com/tc39/test262/tree/main/test/intl402/DateTimeFormat and found no overlap, so great catch :)

iamstolis:
I run the latest `test262` tests against our JavaScript engine (`graal-js`) and the test added by this PR appeared among the new failures. The reason of the failure was not the fact that we do not implement the step checked by this test but that this test is checking conflicts even with new properties (like `dayPeriod` and `fractionalSecondDigits`) that we do not support yet. So, I wonder if this test is missing something like
`features: [Intl.DateTimeFormat-dayPeriod, Intl.DateTimeFormat-fractionalSecondDigits, Intl.DateTimeFormat-datetimestyle]`?

PR #3036, opened 2021-07-12
Add test for 12.4.21 of Temporal

FrankYFTang:
12.4.21 Temporal.Calendar.prototype.fields ( fields )
https://tc39.es/proposal-temporal/#sec-temporal.calendar.prototype.fields

@sffc @Ms2ger @ryzokuken @ljharb


test/built-ins/Temporal/Calendar/fields.js
```diff
@@ -0,0 +1,27 @@
...
+let cal = new Temporal.Calendar("iso8601")
+const fields = {
+  *[Symbol.iterator]() {
+      let i = 0;
+      while (i++ < 1000001) {
+        yield "garbage " + i;
+      }
+  }
+}
+assert.deepEqual(cal.fields(fields), Array.from(fields));
```
Ms2ger, line 27: It doesn't seem like `deepEqual` actually checks the result is an Array. Not sure why we'd use it over `compareArray`

test/built-ins/Temporal/Calendar/fields.js
```diff
@@ -0,0 +1,27 @@
+// Copyright (C) 2021 the V8 project authors. All rights reserved.
```
Ms2ger, line 1: This should go in `test/built-ins/Temporal/Calendar/prototype/fields/`, but compare with the existing test https://github.com/tc39/proposal-temporal/blob/e057057bf18f81b226cd0d6b51faff3ac7da6ce7/polyfill/test/Calendar/prototype/fields/argument-iterable-not-array.js

test/built-ins/Temporal/Calendar/fields.js
FrankYFTang: DONE

test/built-ins/Temporal/Calendar/fields.js
FrankYFTang: DONE

rwaldron:
@FrankYFTang you should make sure you coordinate with @jugglinmike, as I know that he's working on migrating and expanding all of the Temporal tests

FrankYFTang:
> @FrankYFTang you should make sure you coordinate with @jugglinmike, as I know that he's working on migrating and expanding all of the Temporal tests

sure @jugglinmike do you think this test faithfully test against the current Temporal spec text?

jugglinmike (approved):
Yup, it looks accurate to me! Thanks for writing it!

(My guess is that @rwaldron was referring to [this issue I created to coordinate test-writing for the Temporal Proposal](https://github.com/tc39/test262/issues/3002). If you'd like to write more tests, please do let us know there. We still have more tests to write for this method, so I'll leave it as "unassigned" in that issue.)

PR #3037, opened 2021-07-14
Add tests for Temporal.now.plainDateTime

jugglinmike:
In service of gh-3002

/cc @ptomato @Ms2ger


test/built-ins/Temporal/now/plainDateTime/calendar-function.js
```diff
@@ -0,0 +1,46 @@
+// Copyright (C) 2020 Igalia, S.L. All rights reserved.
```
rwaldron, line 1: Is this correct?

Ms2ger: Yes, originates here: https://github.com/tc39/proposal-temporal/commit/60211b2c99b6169665c4f5e0bec8a5a34a77940b

test/built-ins/Temporal/now/plainDateTime/calendar-from-undefined.js
```diff
@@ -1,39 +0,0 @@
-// Copyright (C) 2020 Igalia, S.L. All rights reserved.
```
Ms2ger: Note to self: this was [initially](https://github.com/tc39/proposal-temporal/commit/b5700a83a67ea57c999afaf360342bbb4b2b0336) testing the case where `Calendar.from` was undefined:
```js

Object.defineProperty(Temporal.Calendar, "from", {
  get() {
    actual.push("get Temporal.Calendar.from");
    return undefined;
  },
});
```
Need to check what happened here.

rwaldron (approved):

Ms2ger:
I would appreciate it if you didn't merge Termporal tests over my outstanding comments.

rwaldron:
Sorry, I don't remember seeing that comment. I reviewed all of the files, spent about an hour cross checking them with the original sources in https://github.com/tc39/proposal-temporal/tree/main/polyfill/test/Now/plainDateTime and when I was satisfied, I approved and merged. I'm sure it's possible that in the time between my review and then my review of the proposal's tests, that your comment came in, but my view didn't update for any number of unknowable reasons. Anyway, please consider it nothing more than an accident.

rwaldron:
I am curious though... how would you have liked me to handle a comment that starts with "Note to self:"?

Issue #3038, opened 2021-07-14
Errors in new ArrayBuffer / SharedArrayBuffer tests (related to resizable / growable arraybuffers)


marjakh:
Problem 1: using .resizable instead of .growable:

built-ins/SharedArrayBuffer/options-non-object.js
built-ins/SharedArrayBuffer/options-maxbytelength-undefined.js

Problem 2: errors in variable names

built-ins/SharedArrayBuffer/prototype/maxByteLength/this-is-arraybuffer.js uses an undefined variable "sab" (should be "ab")

Problem 3: test is the wrong way around

test/built-ins/SharedArrayBuffer/prototype/growable/this-is-arraybuffer.js

Spec: if it's *not* a SAB, then throw.
Test: if it *is* a SAB, except and error thrown.

Curiously, the test file name is correct, it probably means to test the situation where `this` is an ArrayBuffer, not a SharedArrayBuffer. But the error message in the test is wrong, saying `this` cannot be a SharedArrayBuffer.

jugglinmike:
Thanks for the report! I'll have corrections ready for review later today

jugglinmike:
Here's the patch: gh-3040

PR #3040, opened 2021-07-14
Correct errors in SharedArrayBuffer tests

jugglinmike:
This resolves gh-3038.

### DIFF
--- a/test/built-ins/SharedArrayBuffer/prototype/growable/this-is-arraybuffer.js
+++ b/test/built-ins/SharedArrayBuffer/prototype/growable/this-is-arraybuffer.js
@@ -2,7 +2,7 @@
 // This code is governed by the BSD license found in the LICENSE file.
 /*---
 esid: sec-get-sharedarraybuffer.prototype.growable
-description: Throws a TypeError exception when `this` is an SharedArrayBuffer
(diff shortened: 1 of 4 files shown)


test/built-ins/SharedArrayBuffer/prototype/growable/this-is-arraybuffer.js
```diff
@@ -18,7 +18,7 @@ var growable = Object.getOwnPropertyDescriptor(
 );
 
 var getter = growable.get;
-var ab = new SharedArrayBuffer(4);
+var ab = new ArrayBuffer(4);
```
marjakh, line 21: The error messages below are still wrong.

marjakh (dismissed):
Thanks for the quick fix! Aside the erroneous error message, these changes look correct and they should fix the bug I reported. (Let's see if github allows me to approve them...)

test/built-ins/SharedArrayBuffer/prototype/growable/this-is-arraybuffer.js
jugglinmike: I'm moving a bit too fast for my own good. That's fixed now, too.

rwaldron (approved):

PR #3041, opened 2021-07-14
Cleanup generation code

gibson042:


rwaldron (approved):
I tested this by running `./make.py clean` and then `./make.py build` locally. I was surprised that these changes have no observable difference. Nice refactor!

PR #3043, opened 2021-07-15
Add whitespace-after-regular-expression-literal tests

### LINKED ISSUE #3039: Add tests for all insignificant characters immediately after a regular expression literal
Turns out that JSC tokenizes non-ASCII non-Latin1 WhiteSpace characters after a regular expression literal as part of its flags, and XS doesn't recognize them as WhiteSpace at all:
* https://bugs.webkit.org/show_bug.cgi?id=227944
* https://github.com/Moddable-OpenSource/moddable/issues/665

This repository should include a test verifying that all valid WhiteSpace and LineTerminator characters are insignificant after a regular expression literal, probably similar to the script in the moddable issue.

gibson042:
Fixes #3039

### DIFF
--- a/src/insignificant-input-elements/README.md
+++ b/src/insignificant-input-elements/README.md
@@ -1,0 +1,9 @@
+This directory should contain a `.case` test case file
+for each character that is either [_WhiteSpace_](https://tc39.es/ecma262/#sec-white-space)
+or [_LineTerminator_](https://tc39.es/ecma262/#sec-line-terminators).
+Because the former is defined in terms of the Unicode "Space_Separator" category,
+it can change over time even if the ECMAScript specification does not.
+
+A current list can be fetched as `[:Space_Separator:]` using a tool
+such as the Unicode UnicodeSet utility:
+https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3ASpace_Separator%3A%5D&g=&i=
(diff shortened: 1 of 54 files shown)


src/insignificant-input-elements/README.md
```diff
@@ -0,0 +1,9 @@
+This directory should contain a `.case` test case file
+for each character that is either [_WhiteSpace_](https://tc39.es/ecma262/#sec-white-space)
+or [_LineTerminator_](https://tc39.es/ecma262/#sec-line-terminators).
+Because the former is defined in terms of the Unicode "Space_Separator" category,
+it can change over time even if the ECMAScript specification does not.
+
+A current list can be fetched as `[:Space_Separator:]` using a tool
+such as the Unicode UnicodeSet utility:
+https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3ASpace_Separator%3A%5D&g=&i=
```
rwaldron, line 9: ❤️

rwaldron (approved):

jugglinmike:
The test generation tool adds the most value in instances with large or non-trivial combinations of testing conditions. With just one template file, the benefit may not be enough to warrant the indirection. I'm thinking of converting these to literal tests (essentially, removing the "case" files and updating the tests' license), and I'm curious about the author's intent. What do you think, @gibson042?

gibson042:
I used the template-and-cases pattern in order to make the whitespace more visible (expressed as a hexadecimal code point rather than an invisible character), and also to more conveniently accommodate future changes in the definition of whitespace. But if those goals can be satisfied equally well in another way, then I doubt I'll object.

PR #3046, opened 2021-07-16
Callable Boundary Realms tests

rwaldron:
### This set of tests is not intended to be exhaustive, but should cover enough ground for an implementation effort to work against.

Either @leobalter or I will be continually adding test material here, until coverage meets our expectations. 

In order to run these: 

1. Make sure you have the latest `test262-harness`
2. `git clone https://github.com/leobalter/realms-polyfill`
3. Run these with a prelude: 
```
test262-harness --prelude ../realms-polyfill/src/test262.js --hostType=jsc --hostPath=`which jsc` 'test/built-ins/Realm/**/*.js'
```


test/built-ins/Realm/prototype/evaluate/wrapped-function-from-return-values-share-no-identity.js
```diff
@@ -0,0 +1,67 @@
...
+arrow.x = 'secrets';
+pFn.x = 'secrets';
+aFn.x = 'secrets';
+genFn.x = 'secrets';
+`)
+
+const wrappedOrdinary = r.evaluate('() => fn')();
+assert.sameValue(typeof wrappedOrdinary, 'function', 'ordinary function wrapped');
+assert.sameValue(wrappedOrdinary(), 42, 'ordinary, return');
+assert.sameValue(wrappedOrdinary.x, undefined, 'ordinary, no property shared');
```
leobalter, line 45: please add assertions verifying `x in wrapped*` and/or `Object.prototype.hasOwnProperty.call(wrappedOrdinary, 'x')`.

test/built-ins/Realm/prototype/evaluate/wrapped-functions-new-wrapping-on-each-evaluation.js
```diff
@@ -0,0 +1,29 @@
...
+  'function',
+  'This test must fail if Realm.prototype.evaluate is not a function'
+);
+
+const r = new Realm();
+
+r.evaluate(`
+function fn() {
+    return 42;
+}
```
leobalter, line 21: oops, evaluate will return undefined here

test/built-ins/Realm/prototype/evaluate/wrapped-function-arguments-are-wrapped-into-the-inner-realm.js
```diff
@@ -0,0 +1,24 @@
...
+  'This test must fail if Realm.prototype.evaluate is not a function'
+);
+
+const r = new Realm();
+const blueFn = (x, y) => x + y;
+
+const redWrappedFn = r.evaluate(`
+0, function(blueWrappedFn, a, b, c) {
+    return blueWrappedFn(a, b) * c;
+}
```
leobalter, line 22: you might wanna use an arrow function here as it's already used on line 17.

leobalter (dismissed):
This is a partial review (incomplete). Overall this is great, I just need more time to finish.

jugglinmike:
> Either @leobalter or I will be continually adding test material here, until coverage meets our expectations.

By "here" do you mean "the `test/built-ins/Realm` directory (on the `main` branch)" or "this pull request"?

rwaldron:
@jugglinmike 

> By "here" do you mean "the test/built-ins/Realm directory (on the main branch)" or "this pull request"?

That's a great question. I think I originally meant "here in this pull request", but I think I mean "here in test/built-ins/Realm/*" now. @leobalter and I will figure out what we want to do and communicate the plan a little more clearly than I have :)

test/built-ins/Realm/prototype/evaluate/wrapped-functions-new-wrapping-on-each-evaluation.js
rwaldron: It doesn't need to be, since the test is evaluating `fn` later and getting the wrapped function there.

test/built-ins/Realm/prototype/evaluate/wrapped-function-from-return-values-share-no-identity.js
rwaldron: I've added tests that check `Object.prototype.hasOwnProperty.call(wrappedOrdinary, 'x') === false`

leobalter:
LGTM after the tweaks proposed in #3073.

The coverage for importValue needs to be extended but everything else looks about right.

PR #3054, opened 2021-07-19
Add presence check for property cause on Error.prototype

legendecas:
Property `cause` should not present on `Error.prototype`.

### DIFF
--- a/test/built-ins/Error/prototype/properties.js
+++ b/test/built-ins/Error/prototype/properties.js
@@ -1,0 +1,19 @@
+// Copyright (C) 2021 Chengzhong Wu. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Error prototype properties
+esid: sec-properties-of-the-error-prototype-object
+includes: [propertyHelper.js]
+---*/
+
+const desc = {
+  writable: true,
+  enumerable: false,
+  configurable: true
+};
+verifyProperty(Error.prototype, 'cause', undefined);
+verifyProperty(Error.prototype, 'constructor', desc);
+verifyProperty(Error.prototype, 'name', desc);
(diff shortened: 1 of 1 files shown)


rwaldron (changes requested):
- Some of this duplicates existing tests. 
- Please put these in separate files, in the appropriate directories (where you will find that some of this is already covered)

jugglinmike:
Thanks for the patch! [Hosts are permitted to define an `Error.prototype.cause` property](https://tc39.es/ecma262/#sec-host-built-in-objects-and-methods), so we should remove the assertion for its absence.

legendecas:
Thanks for the suggestions! Well, I didn't notice that there are statements about host extensions on built-in objects. I was noticing v8 is putting a `cause` with `undefined` value on `Error.prototype` which is not expected. If spec allows such extension, I'd close this PR and just file a bug to v8.

ljharb:
I added https://github.com/tc39/agendas/commit/964d2d36f80994b1b77972ef4867c8569b531874 to the next meeting to discuss tightening restrictions on hosts, so that helpful tests like this can be permitted.

leobalter:
@ljharb can you share what do you suggest/recommend as the next steps for the restrictions?

The issue on this one is that ECMAScript allows extensions. This is imperative as globals add names for the platform it is contained, and other names are not exactly defined by ECMAScript. Otherwise, it would be easy to enumerate the names expected in `Error.prototype`, but if the host adds anything (not limited to `cause`), how is that an error?

ljharb:
Allowing new globals, as well as property names on any object that are *not* specified in 262 or 402, is fine.

My hope is that we can forbid extensions on any specified objects whose names are specified elsewhere in 262 or 402, which would include “cause” on any language builtins other than the specified ones.

PR #3056, opened 2021-07-19
Add tests for Temporal.Calendar.p*.inLeapYear

FrankYFTang:


test/built-ins/Temporal/Calendar/prototype/inLeapYear/string.js
```diff
@@ -0,0 +1,45 @@
...
+assert.sameValue(false, cal.inLeapYear("2019-03-18T13:00:00Z"));
+assert.sameValue(true, cal.inLeapYear("2020-12-31T23:59:59Z"));
+assert.sameValue(false, cal.inLeapYear("2021-03-18T13:00:00Z"));
+assert.sameValue(false, cal.inLeapYear("2022-03-18T13:00:00Z"));
+assert.sameValue(false, cal.inLeapYear("2023-03-18T13:00:00Z"));
+assert.sameValue(true, cal.inLeapYear("2024-03-18T13:00:00Z"));
+assert.sameValue(false, cal.inLeapYear("2025-01-01T00:00:00Z"));
+assert.sameValue(false, cal.inLeapYear("2026-03-18T13:00:00Z"));
+
+assert.sameValue(false, cal.inLeapYear("2019-03"));
```
ptomato, line 38: This block isn't correct - these strings will fail ToTemporalDate.

FrankYFTang: hum... interesting... so cal.inLeapYear will take PlainYearMonth but not year-month string....

FrankYFTang: PTAL

test/built-ins/Temporal/Calendar/prototype/inLeapYear/string-throw-range-error.js
```diff
@@ -0,0 +1,29 @@
+// Copyright (C) 2021 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.calendar.prototype.inleapyear
+description: Temporal.Calendar.prototype.inLeapYear throw RangeError while ISO8601 string is not a date.
+  and return true or false that year is leap year.
+  and return Array of the same content.
```
rwaldron, line 8: @FrankYFTang Looks like some copy pasta?

leobalter: I'd recommend using the appropriate yaml block format to avoid breaking compatibility of Test262 with many runners using it.

```suggestion
description: >
  Temporal.Calendar.prototype.inLeapYear throw RangeError while ISO8601 string is not a date.
  and return true or false that year is leap year.
  and return Array of the same content.
```

Even thou, this test is only testing one thing here so the text can be improved too.

test/built-ins/Temporal/Calendar/prototype/inLeapYear/string-throw-range-error.js
```diff
@@ -0,0 +1,29 @@
...
+info: |
+  1. Let calendar be the this value.
+  2. Perform ? RequireInternalSlot(calendar, [[InitializedTemporalCalendar]]).
+  3. Assert: calendar.[[Identifier]] is "iso8601".
+  4. If Type(temporalDateLike) is not Object or temporalDateLike does not have an [[InitializedTemporalDate]] or [[InitializedTemporalYearMonth]] internal slot, then
+  a. Set temporalDateLike to ? ToTemporalDate(temporalDateLike).
+  5. Return ! IsISOLeapYear(temporalDateLike.[[ISOYear]]).
```
leobalter, line 15: This test could add the text that throws the RangeError:

```
ParseTemporalDateString ( isoString )

1. Assert: Type(isoString) is String.
2. If isoString does not satisfy the syntax of a TemporalDateString (see 13.33), then
  a. Throw a RangeError exception.
3. Let result be ? ParseISODateTime(isoString).
4. Return the new Record { [[Year]]: result.[[Year]], [[Month]]: result.[[Month]], [[Day]]: result.[[Day]], [[Calendar]]: result.[[Calendar]] }.
```

And also the relations to the [iso8601](https://tc39.es/proposal-temporal/#sec-temporal-iso8601grammar) could be valuable here

leobalter: This can be shortened by saying there is no syntax produced in the [iso8601 grammar section](https://tc39.es/proposal-temporal/#sec-temporal-iso8601grammar) for these format variations that otherwise look like regular dates.

test/built-ins/Temporal/Calendar/prototype/inLeapYear/plain-date-time.js
```diff
@@ -0,0 +1,30 @@
...
+  2. Perform ? RequireInternalSlot(calendar, [[InitializedTemporalCalendar]]).
+  3. Assert: calendar.[[Identifier]] is "iso8601".
+  4. If Type(temporalDateLike) is not Object or temporalDateLike does not have an [[InitializedTemporalDate]] or [[InitializedTemporalYearMonth]] internal slot, then
+  a. Set temporalDateLike to ? ToTemporalDate(temporalDateLike).
+  5. Return ! IsISOLeapYear(temporalDateLike.[[ISOYear]]).
+features: [Temporal]
+---*/
+let cal = new Temporal.Calendar("iso8601");
+
+assert.sameValue(false, cal.inLeapYear(new Temporal.PlainDateTime(1995, 8, 23, 5, 30, 13)));
```
leobalter, line 20: these tests are only touching the surface of inLeapYear and there is so much that could be done with the date validations. One thing that is interesting is verifying some edgy variations of the year, like years that are not the common 4 digit (0, 1, or "0000", "0001" for parsed strings), etc.

There is also so many different types accepted as argument. This should include things such as a `new Date`, a regular object, with valid and invalid (true and false returns), and abrupt completions (toPrimitive operations, symbols), and just weird args such as an integer, undefined, null, etc.

This would help guarantee the cross compatibility and avoid surprises.

leobalter (changes requested):
This is missing coverage beyond superficial API tests. The tests are generally ok other than formating. I don't have specific preference to address everything in a single PR, but early merging this should not consider the coverage as done for the method.

There are many of the spec parts missing here, like testing `2. Perform ? RequireInternalSlot(calendar, [[InitializedTemporalCalendar]]).` and all the ends from `ToTemporalDate`.

One suggestion, `ToTemporalDate` is reused in many parts of Temporal, so I'd suggest having a utility helper in the harness folder that can be reused, maybe as a factory for checking all the points the abstraction might be called by each api.

leobalter:
There is a lot of tests that should also be imported from Temporal. I'm not sure how much they are up to date, but the files give a good base for the necessary coverage too. 

https://github.com/tc39/proposal-temporal/tree/main/polyfill/test/Calendar/prototype/inLeapYear

FrankYFTang:
> There is a lot of tests that should also be imported from Temporal. I'm not sure how much they are up to date, but the files give a good base for the necessary coverage too.
> 
> https://github.com/tc39/proposal-temporal/tree/main/polyfill/test/Calendar/prototype/inLeapYear

It will be nice if whoever is working on that move import them to test262 ASAP.

ptomato:
I outlined the plan for moving them over in https://github.com/tc39/test262/issues/3002#issuecomment-887908934
It would be really inconvenient for ensuring that we correctly and completely converted the legacy test suite, if we were to move everything over now to this repo and let it start to diverge.

FrankYFTang:

PR #3058, opened 2021-07-19
Add tests for Temporal.Calendar.p*.monthDayFromFields

FrankYFTang:

### DIFF
--- a/test/built-ins/Temporal/Calendar/prototype/monthDayFromFields/overflow-undefined.js
+++ b/test/built-ins/Temporal/Calendar/prototype/monthDayFromFields/overflow-undefined.js
@@ -1,0 +1,45 @@
+// Copyright (C) 2021 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.calendar.prototype.monthdayfromfields
+description: Temporal.Calendar.prototype.monthDayFromFields will return correctly with default overflow which constrain date range.
+info: |
+  1. Let calendar be the this value.
+  2. Perform ? RequireInternalSlot(calendar, [[InitializedTemporalCalendar]]).
+  3. Assert: calendar.[[Identifier]] is "iso8601".
+  4. If Type(fields) is not Object, throw a TypeError exception.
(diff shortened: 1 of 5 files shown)


test/built-ins/Temporal/Calendar/prototype/monthDayFromFields/overflow-undefined.js
```diff
@@ -0,0 +1,45 @@
+// Copyright (C) 2021 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.calendar.prototype.monthdayfromfields
+description: Temporal.Calendar.prototype.monthDayFromFields will return correctly with default overflow which constrain date range.
```
ptomato, line 6: This one has a naming conflict with a test already in the proposal repo which still has to be copied over here: https://github.com/tc39/proposal-temporal/blob/main/polyfill/test/Calendar/prototype/monthDayFromFields/overflow-undefined.js
Consider combining them?

FrankYFTang: so... what make the tests in those places not yet move into test262 yet?

FrankYFTang: How about I change the file name and yield the file name for those in 
https://github.com/tc39/proposal-temporal/blob/main/polyfill/test/

ptomato: I outlined the plan for moving them over in https://github.com/tc39/test262/issues/3002#issuecomment-887908934
It would be really inconvenient for ensuring that we correctly and completely converted the legacy test suite, if we were to move everything over now to this repo and let it start to diverge.

FrankYFTang: ok, let me just abandon this PR and wait for @ptomato  and @Ms2ger  to move their test from Temporal repo instead.

FrankYFTang:

PR #3068, opened 2021-07-20
Add test for Temporal.Duration.from

FrankYFTang:

### DIFF
--- a/test/built-ins/Temporal/Duration/from/duration.js
+++ b/test/built-ins/Temporal/Duration/from/duration.js
@@ -3,7 +3,7 @@
 
 /*---
 esid: sec-temporal.duration.from
-description: Temporal.Duration.from will crate a Duration from a given Duration.
+description: Temporal.Duration.from will create a Duration from a given Duration.
 info: |
   1. If Type(item) is Object and item has an [[InitializedTemporalDuration]] internal slot, then
   a. Return ? CreateTemporalDuration(item.[[Years]], item.[[Months]], item.[[Weeks]], item.[[Days]], item.[[Hours]], item.[[Minutes]], item.[[Seconds]], item.[[Milliseconds]], item.[[Microseconds]], item.[[Nanoseconds]]).
--- a/test/built-ins/Temporal/Duration/from/duration.js
+++ b/test/built-ins/Temporal/Duration/from/duration.js
@@ -1,0 +1,32 @@
+// Copyright (C) 2021 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.duration.from
+description: Temporal.Duration.from will crate a Duration from a given Duration.
+info: |
+  1. If Type(item) is Object and item has an [[InitializedTemporalDuration]] internal slot, then
+  a. Return ? CreateTemporalDuration(item.[[Years]], item.[[Months]], item.[[Weeks]], item.[[Days]], item.[[Hours]], item.[[Minutes]], item.[[Seconds]], item.[[Milliseconds]], item.[[Microseconds]], item.[[Nanoseconds]]).
+  2. Return ? ToTemporalDuration(item).
+features: [Temporal]
+---*/
+function assertDuration(duration, years, months, weeks, days, hours,
+    minutes, seconds, milliseconds, microseconds, nanoseconds, sign, blank) {
+  assert.sameValue(years, duration.years, "years");
+  assert.sameValue(months, duration.months, "months");
+  assert.sameValue(weeks, duration.weeks, "weeks");
+  assert.sameValue(days, duration.days, "days");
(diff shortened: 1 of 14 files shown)


test/built-ins/Temporal/Duration/from/duration.js
```diff
@@ -0,0 +1,32 @@
+// Copyright (C) 2021 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.duration.from
+description: Temporal.Duration.from will crate a Duration from a given Duration.
```
rwaldron, line 6: s/crate/create

test/built-ins/Temporal/Duration/from/duration.js
```diff
@@ -0,0 +1,32 @@
...
+  assert.sameValue(days, duration.days, "days");
+  assert.sameValue(hours, duration.hours, "hours");
+  assert.sameValue(minutes, duration.minutes, "minutes");
+  assert.sameValue(seconds, duration.seconds, "seconds");
+  assert.sameValue(milliseconds, duration.milliseconds, "milliseconds");
+  assert.sameValue(microseconds, duration.microseconds, "microseconds");
+  assert.sameValue(nanoseconds, duration.nanoseconds, "nanoseconds");
+  assert.sameValue(sign, duration.sign, "sign");
+  assert.sameValue(blank, duration.blank, "blank");
+}
```
rwaldron, line 27: If this is going to be used many times, please add it to https://github.com/tc39/test262/blob/main/harness/temporalHelpers.js

test/built-ins/Temporal/Duration/from/string-throw-range-error.js
```diff
@@ -0,0 +1,56 @@
...
+assert.throws(RangeError, () => Temporal.Duration.from("PT,1H"),
+    "Invalid time value");
+assert.throws(RangeError, () => Temporal.Duration.from("PT.1M"),
+    "Invalid time value");
+assert.throws(RangeError, () => Temporal.Duration.from("PT,1M"),
+    "Invalid time value");
+assert.throws(RangeError, () => Temporal.Duration.from("PT.1S"),
+    "Invalid time value");
+assert.throws(RangeError, () => Temporal.Duration.from("PT,1S"),
+    "Invalid time value");
```
rwaldron, line 56: There are two options here: 

1. You can update these assert messages to be meaningful
2. I can run a script on them later and generate something more meaningful, but you'll have no say in them

test/built-ins/Temporal/Duration/from/string.js
```diff
@@ -0,0 +1,179 @@
...
+assertDuration(Temporal.Duration.from("PT3,025M"),
+    0, 0, 0, 0, 0, 3, 1, 500, 0, 0, 1, false);
+assertDuration(Temporal.Duration.from("PT3,01M"),
+    0, 0, 0, 0, 0, 3, 0, 600, 0, 0, 1, false);
+assertDuration(Temporal.Duration.from("PT3,005M"),
+    0, 0, 0, 0, 0, 3, 0, 300, 0, 0, 1, false);
+assertDuration(Temporal.Duration.from("PT3,001M"),
+    0, 0, 0, 0, 0, 3, 0, 60, 0, 0, 1, false);
+assertDuration(Temporal.Duration.from("PT3,006M"),
+    0, 0, 0, 0, 0, 3, 0, 360, 0, 0, 1, false);
```
rwaldron, line 179: What happens when the wrong number of arguments are passed?

test/built-ins/Temporal/Duration/from/string-throw-range-error.js
FrankYFTang: 2. is ok for me.

test/built-ins/Temporal/Duration/from/string.js
FrankYFTang: "PT3,001M" in ISO8601 is a different way to represent "PT3.001M" , whic means 3.001 minutes.
which is 3 minutes and 60 milliseconds

"PT3,006M" in ISO8601 is a different way to represent "PT3.006M" , whic means 3.006 minutes.
which is 3 minutes and 360 milliseconds

See https://tc39.es/proposal-temporal/#prod-DecimalSeparator both "." and "," are DecimalSeparator .

If your question is about the number of argument of the from method- there is only one 
https://tc39.es/proposal-temporal/#sec-temporal.duration.from

rwaldron:
@FrankYFTang the additions to `harness/temporalHelpers.js` are in `main` now, so you can rebase this and update it.

ptomato:
I'd recommend not merging this as it's obviously testing a mistake in the spec text, and we intend to ask for consensus at the next TC39 meeting to change it.

FrankYFTang:

PR #3069, opened 2021-07-20
Rename all uses of Temporal.now to Temporal.Now

linusg:
See: https://github.com/tc39/proposal-temporal/commit/d0acb66


jugglinmike (approved):
Thanks!

PR #3075, opened 2021-07-21
Add test for Temporal.Instant.p*.add

FrankYFTang:

### DIFF
--- a/test/built-ins/Temporal/Instant/prototype/add/throw-range-error-from-AddInstant.js
+++ b/test/built-ins/Temporal/Instant/prototype/add/throw-range-error-from-AddInstant.js
@@ -1,0 +1,21 @@
+// Copyright (C) 2021 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.instant.prototype.add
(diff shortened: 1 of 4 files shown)


test/built-ins/Temporal/Instant/prototype/add/throw-range-error-from-AddInstant.js
```diff
@@ -0,0 +1,21 @@
+// Copyright (C) 2021 the V8 project authors. All rights reserved.
```
ptomato, line 1: Consider combining with https://github.com/tc39/proposal-temporal/blob/main/polyfill/test/Instant/prototype/add/result-out-of-range.js

test/built-ins/Temporal/Instant/prototype/add/throw-type-error-from-RequireInternalSlot.js
```diff
@@ -0,0 +1,18 @@
+// Copyright (C) 2021 the V8 project authors. All rights reserved.
```
ptomato, line 1: Consider using https://github.com/tc39/proposal-temporal/blob/main/polyfill/test/Instant/prototype/add/branding.js

FrankYFTang:

PR #3114, opened 2021-07-27
Default to not regenerate test files that are already newer than their case and template

gibson042:
This reduces the time of `./make.py` by  an order of magnitude for me, from 110 seconds to 6.


rwaldron (approved):
I love it

PR #3127, opened 2021-08-03
Add various private field and private method tests

anba:
This adds tests for implementation bugs in SpiderMonkey [1], plus
additional tests for implementation bugs in V8 and JSC.

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=1723155

### DIFF
--- a/test/language/expressions/in/private-field-in-nested.js
+++ b/test/language/expressions/in/private-field-in-nested.js
@@ -1,0 +1,27 @@
+// Copyright (C) 2021 André Bargull. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: >
+  Can't nest `in` expressions when the left-hand side is PrivateIdentifier.
+info: |
+  Syntax
+    RelationalExpression[In, Yield, Await]:
+    [...]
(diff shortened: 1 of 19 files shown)


test/language/expressions/in/private-field-in-nested.js
```diff
@@ -0,0 +1,27 @@
...
+features: [class-fields-private, class-fields-private-in]
+---*/
+
+$DONOTEVALUATE();
+
+class C {
+  #field;
+
+  constructor() {
+    #field in #field in this;
```
ljharb, line 25: Why would this not evaluate? The RHS is an expression. Or is it that precedence would evaluate the first two pieces first, which would be a syntax error?

anba: The [grammar production](https://tc39.es/ecma262/#prod-RelationalExpression) is:
> RelationalExpression :  PrivateIdentifier in ShiftExpression

so it's not possible to nest private field checks.

`a in b in c` is allowed because the left-hand side in this RelationalExpression expression is another RelationalExpression:
> RelationalExpression :  RelationalExpression in ShiftExpression

rwaldron (approved):
As always, these are great.

PR #3128, opened 2021-08-03
Split files which test for the property order of functions

anba:
So SpiderMonkey can run the other parts of these tests.

The SpiderMonkey bug for this issue is <https://bugzilla.mozilla.org/show_bug.cgi?id=1629803>.

### DIFF
--- a/test/language/statements/class/definition/fn-name-static-precedence.js
+++ b/test/language/statements/class/definition/fn-name-static-precedence.js
@@ -17,7 +17,6 @@ info: |
       b. Else,
         i. Let field be ClassElementEvaluation of e with arguments F and false.
            [ This overwrites the name property on F. ]
-includes: [compareArray.js]
 features: [generators]
 ---*/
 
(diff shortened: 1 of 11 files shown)


test/language/statements/class/definition/fn-length-static-precedence.js
```diff
@@ -79,4 +75,3 @@ class E {
 }
 
 assert.sameValue(typeof E.length, 'function');
-assert(compareArray(Object.getOwnPropertyNames(E), ['length', 'name', 'prototype']))
```
rwaldron, line 82: Need to remove `includes: [compareArray.js]`

test/language/statements/class/definition/fn-name-static-precedence.js
```diff
@@ -76,4 +72,3 @@ class E {
 }
 
 assert.sameValue(typeof E.name, 'function');
-assert(compareArray(Object.getOwnPropertyNames(E), ['length', 'name', 'prototype']))
```
rwaldron, line 79: Need to remove `includes: [compareArray.js]`

rwaldron (changes requested):

test/language/statements/class/definition/fn-length-static-precedence.js
anba: Removed both includes.

rwaldron (approved):

rwaldron:
@anba thanks!

PR #3129, opened 2021-08-03
Derived class constructor checks are executed after leaving the function body

anba:
Upstreams the tests for <https://bugzilla.mozilla.org/show_bug.cgi?id=1722269>, because some of the tests also fail in JSC and V8.

### DIFF
--- a/test/language/statements/class/subclass/derived-class-return-override-for-of-arrow.js
+++ b/test/language/statements/class/subclass/derived-class-return-override-for-of-arrow.js
@@ -1,0 +1,38 @@
+// Copyright (C) 2021 André Bargull. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-ecmascript-function-objects-construct-argumentslist-newtarget
+description: >
+  ReferenceError when returning from a derived class constructor without calling
+  `super()` is thrown after the function body has been left, so an iterator
+  return handler can still call `super()`.
+---*/
+
+var iter = {
+  [Symbol.iterator]() {
+    return this;
+  },
+  next() {
+    return {done: false};
+  },
+  return() {
+    // Calls |super()|.
+    this.f();
+
+    return {done: true};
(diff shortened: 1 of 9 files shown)


test/language/statements/class/subclass/derived-class-return-override-for-of.js
```diff
@@ -0,0 +1,37 @@
+// Copyright (C) 2021 André Bargull. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-ecmascript-function-objects-construct-argumentslist-newtarget
+description: >
+  TypeError from `return 0` is thrown after the function body has been left, so
+  an error thrown from an iterator has precedence.
```
leobalter, line 8: this is a pretty nice test!

test/language/statements/class/subclass/derived-class-return-override-for-of-arrow.js
```diff
@@ -0,0 +1,38 @@
...
+    return {done: true};
+  },
+};
+
+class C extends class {} {
+  constructor() {
+    iter.f = () => super();
+
+    for (var k of iter) {
+      return;
```
leobalter, line 32: this test is great! 

I'm just not sure if this should leave any note (comment/metadata) saying this will provoke the iterator's return.

leobalter (approved):

test/language/statements/class/subclass/derived-class-return-override-for-of-arrow.js
anba: > this test is great!

Thanks! :-)

The description in the YAML block mentions that the "iterator return handler" is called. But I can also add a line comment here to clarify which code triggers the handler. WDYT?

leobalter: it should be ok without changes. I just had a random and noisy thought.

PR #3136, opened 2021-08-06
Automated assertion message update

rwaldron:
This also required manually fixing the "actual" and "expected" order of many assert.sameValue(...) calls.


leobalter (approved):

Ms2ger:
Hey @rwaldron - this made it significantly harder to verify changes between this repo and the proposal repo. If this is automated, can you share the automation so I can run it against the proposal repo, or make a PR there for review?

Issue #3137, opened 2021-08-06
Consider copying the `branding.js` test from https://github.com/tc39/proposal-temporal/blob/main/polyfill/test/Calendar/prototype/daysInMonth/branding.js which covers this case more comprehensively. (Same remark for the other PRs)


rwaldron:
Consider copying the `branding.js` test from https://github.com/tc39/proposal-temporal/blob/main/polyfill/test/Calendar/prototype/daysInMonth/branding.js which covers this case more comprehensively. (Same remark for the other PRs)

_Originally posted by @ptomato in https://github.com/tc39/test262/pull/3052#r683856883_

PR #3138, opened 2021-08-07
Add branding tests from Temporal

ptomato:
Closes: #3137

(Ms2ger originally authored all these tests so I'm making him the author of the commit)

### DIFF
--- a/test/built-ins/Temporal/Calendar/prototype/dateAdd/branding.js
+++ b/test/built-ins/Temporal/Calendar/prototype/dateAdd/branding.js
@@ -1,0 +1,22 @@
+// Copyright (C) 2021 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.calendar.prototype.dateadd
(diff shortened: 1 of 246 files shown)


leobalter (changes requested):
Please remember to add the necessary metadata so the CI can pass. Failing to do it might break existing test262 runners.

```
description: >
  Throw a TypeError if this is invalid
```

The example above should be good enough.

~~I also believe this is not exactly branding tests, the tests are verifying if `this` is invalid. Maybe calling the files `this-invalid-throws.js` would be more consistent with other Test262 tests.~~

leobalter:
I've got corrected these are branding tests, I still have the nit pick over the test files but don't worry about changing the test file names.

rwaldron:
> Please remember to add the necessary metadata so the CI can pass. Failing to do it might break existing test262 runners.

Yep! [`description` is required (as the contributing documentation makes clear)](https://github.com/tc39/test262/blob/main/CONTRIBUTING.md#frontmatter)

ptomato:
OK, thanks for the review! I've added descriptions to all the files.

leobalter (approved):

PR #3161, opened 2021-08-19
refactor: remove function syntax boundary from AssignmentTargetType templates. Closes #3160

### LINKED ISSUE #3160: Templates lacking proper `$DONOTEVALUATE()` support
The following two template lack proper `$DONOTEVALUATE()` support:

- https://github.com/tc39/test262/blob/main/src/assignment-target-type/invalid/direct.template
- https://github.com/tc39/test262/blob/main/src/assignment-target-type/invalid/parenthesized.template

They wrap the test code in a function that is never called, defeating the purpose of the `$DONOTEVALUATE()` mechanism. In engines that (incorrectly) implement the error as a runtime error (instead of an early/parse error) no error is thrown because the code isn’t executed.

Could we either

a) rewrite this test without the function, or
b) keep it as-is but call the function at the end?

I think option A makes the most sense.

Affected tests:
- https://github.com/tc39/test262/blob/main/test/language/expressions/assignmenttargettype/direct-callexpression-arguments.js
- https://github.com/tc39/test262/blob/main/test/language/expressions/assignmenttargettype/parenthesized-callexpression-arguments.js

rwaldron:


leobalter (approved):

test/language/expressions/assignmenttargettype/direct-callexpression-arguments.js
```diff
@@ -17,7 +17,4 @@ info: |
 
 $DONOTEVALUATE();
 
-function _() {
-  f() = 1;
-}
-
+f() = 1;
```
mathiasbynens, line 20: For this + other tests that do `f()`, I think we need to make sure `f` is a function — otherwise Web-compatible engines produce a `ReferenceError` which is not what the test is about.

Adding a line above that `function f() {}` might be sufficient. WDYT?

mathiasbynens:
#3160

PR #3162, opened 2021-08-24
memoize decimalToPercentHexString()

lightmare:
Some decodeURI/encodeURI tests call this function many times in nested loops, with the same argument.

### DIFF
--- a/harness/decimalToHexString.js
+++ b/harness/decimalToHexString.js
@@ -20,7 +20,15 @@ function decimalToHexString(n) {
   return s;
 }
 
+var decimalToPercentHexString$ = new Array(256);
+
 function decimalToPercentHexString(n) {
-  var hex = "0123456789ABCDEF";
-  return "%" + hex[(n >> 4) & 0xf] + hex[n & 0xf];
+  var s = decimalToPercentHexString$[n];
+  if (!s) {
+    var hex = "0123456789ABCDEF";
+    var cH = hex[(n >> 4) & 0xf];
+    var cL = hex[n & 0xf];
+    s = decimalToPercentHexString$[n] = `%${cH}${cL}`;
+  }
+  return s;
 }


jugglinmike:
Thanks for the patch! Many optimizations are lost in the enormous scale of Test262. In order to justify the complexity of an efficiency-motivated change, it helps to have some data about how much it meaningfully improves performance. Can you say a bit about how much this helps?

lightmare:
I'm testing engine262, where this kind of optimization is almost guaranteed to save some time. Not so sure about actual optimizing engine, though.

Without memoize:
```
> @engine262/engine262@0.0.1 test:test262 /home/foss❤/nodejs/engine262
> node test/test262/test262.js "--run-slow-tests" "test/built-ins/encodeURI/"


#######################
 engine262 Test Runner
 Detected 4 CPUs
 Not running on CI
#######################

[01:21|:   62|+   62|-    0|»    0] (0.00/s)

real	1m22.239s
user	1m33.420s
sys	0m1.785s
```

With memoize:
```
> @engine262/engine262@0.0.1 test:test262 /home/foss❤/nodejs/engine262
> node test/test262/test262.js "--run-slow-tests" "test/built-ins/encodeURI/"


#######################
 engine262 Test Runner
 Detected 4 CPUs
 Not running on CI
#######################

[01:16|:   62|+   62|-    0|»    0] (0.00/s)

real	1m16.893s
user	1m28.827s
sys	0m1.522s
```

I was actually thinking about precomputing the table and replacing the call with array lookup in individual tests. For example `test/built-ins/decodeURI/S15.1.3.1_A2.5_T1.js` has four nested loops, calling this function a million times. That approach would probably help more than memoization inside, but would also mean more changes sprinkled across multiple files.

rwaldron:
@jugglinmike https://jsbench.me/19ksrm2a0q/1

jugglinmike:
Thanks, you two. @rwaldron I have no doubt that this change makes things go faster in general; it's just that I'm interested in its effect on real-life workloads. That's why I'd like to ask @lightmare: is it common for you to run the `encodeURI` tests in isolation? When I validate changes to a parser I maintain, I find myself casting a wider net, so I'm curious if you're usually also watching results for, e.g., `encodeURIComponent`, `decodeURI`, and `decodeURIComponent`.

lightmare:
> is it common for you to run the `encodeURI` tests in isolation?

No, it is not common. I would only do that after making changes to that specific function.

The issue with engine262 is that `decodeURI` and `decodeURIComponent` take about a third of the whole test suite's runtime.

rwaldron:
Closing, per comment here https://github.com/tc39/test262/pull/3169#issuecomment-915045182

PR #3170, opened 2021-08-26
avoid unnecessary calls to decimalToHexString()

lightmare:
This removes calls to `decimalToHexString()` whose result was never used; and moves calls whose result was only used to construct error message, into the error-handling branch.


jugglinmike (approved):
Nice; thanks!

PR #3172, opened 2021-08-27
Normalize some YAML

### LINKED ISSUE #3171: valid but unusual YAML
In 47 recent Temporal tests, the `description` is a multi-line scalar in flow notation,  e.g.
```
description: Temporal.Calendar.prototype.dateFromFields should throw RangeError for
  input not in valid range.
```
which is valid YAML, but not something that the rest of test262 uses.
For multi-line scalars, test262 typically uses block notation, e.g.
```
description: >
  Temporal.Calendar.prototype.dateFromFields should throw RangeError for
  input not in valid range.
```
Would you like a PR to change them, or is that more bother than it's worth? (CONTRIBUTING.md isn't clear on this point.)

jmdyck:
In some recently added tests, the `description` is a multi-line scalar in flow notation. Change it to block notation.

Resolves #3171.


rwaldron (approved):
Thanks @jmdyck!

Issue #3139, opened 2021-08-10
Incorrect test expectation in type-calendar-valid.js


anba:
The underscore tests in <https://github.com/tc39/test262/blob/86e23998e37e2f1d8107afa51d7be355f3f65cca/test/intl402/DisplayNames/prototype/of/type-calendar-valid.js#L29-L32> should all be moved to <https://github.com/tc39/test262/blob/main/test/intl402/DisplayNames/prototype/of/type-calendar-invalid.js>.

The underscore character "_" is only accepted for the [sep](https://unicode.org/reports/tr35/#Unicode_language_identifier) production when the [backwards-compatibility extensions](https://unicode.org/reports/tr35/#BCP_47_Conformance) are enabled. But in ECMA-402 we [only accept](https://tc39.es/ecma402/#sec-language-tags) _Unicode BCP 47 locale identifiers_, so identifiers without the backwards-compatibility extensions.

jugglinmike:
Thanks for reporting this! And for effectively writing the commit message. Correction under review here: gh-3173

Just when you think you know a programming language...

PR #3173, opened 2021-09-01
Correct tests for Unicode Language Identifier

jugglinmike:
The underscore character "_" is only accepted for the sep production [1]
when the backwards-compatibility extensions [2] are enabled. However,
ECMA-402 only accepts Unicode BCP 47 locale identifiers [3], meaning
identifiers *without* the backwards-compatibility extensions.

[1] https://unicode.org/reports/tr35/#Unicode_language_identifier
[2] https://unicode.org/reports/tr35/#BCP_47_Conformance
[3] https://tc39.es/ecma402/#sec-language-tags

---

Resolves gh-3139. /cc @anba

### DIFF
--- a/test/intl402/DisplayNames/prototype/of/type-calendar-invalid.js
+++ b/test/intl402/DisplayNames/prototype/of/type-calendar-invalid.js
@@ -43,3 +43,19 @@ assert.throws(RangeError, function() {
 assert.throws(RangeError, function() {
   displayNames.of('abc def');
 }, 'interstitial space');
+
+assert.throws(RangeError, function() {
+  displayNames.of('123_abc');
+}, '2 segments, minimum length, underscore');
+
+assert.throws(RangeError, function() {
+  displayNames.of('12345678_abcdefgh');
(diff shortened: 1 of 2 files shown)


rwaldron (approved):

anba:
Thanks! :-)

FrankYFTang:
I disagree with this change

The spec stated
https://tc39.es/intl-displaynames-v2/#sec-canonicalcodefordisplaynames

a. If code does not match the Unicode Locale Identifier type nonterminal, throw a RangeError exception.

And 
sep | = [-_] ;
type(Also known as uvalue) | = alphanum{3,8}(sep alphanum{3,8})* ;

which means _  match "Unicode Locale Identifier type nonterminal" 

[3] https://tc39.es/ecma402/#sec-language-tags 

does not mention "type nonterminal" and the code pass into the "of()" is referring to language tag but only the type nonterminal.

FrankYFTang:
make it clear in https://github.com/tc39/ecma402/pull/647

PR #3177, opened 2021-09-03
Test Math.hypot argument coercion

gibson042:
https://bugs.webkit.org/show_bug.cgi?id=229843

### DIFF
--- a/test/built-ins/Math/hypot/Math.hypot_Zero_2.js
+++ b/test/built-ins/Math/hypot/Math.hypot_Zero_2.js
@@ -2,12 +2,16 @@
 // This code is governed by the BSD license found in the LICENSE file.
 
 /*---
+esid: sec-math.hypot
 es6id: 20.2.2.18
 author: Ryan Lewis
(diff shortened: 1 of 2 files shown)


test/built-ins/Math/hypot/Math.hypot_Zero_2.js
```diff
@@ -2,12 +2,16 @@
 // This code is governed by the BSD license found in the LICENSE file.
 
 /*---
+esid: sec-math.hypot
```
jugglinmike, line 5: Thanks for adding this!

jugglinmike (approved):
Nice catch! I'm curious: how did you discover the JSC bug?

gibson042:
> Nice catch! I'm curious: how did you discover the JSC bug?

I was reviewing the definition of Math.hypot as part of the https://github.com/js-choi/proposal-bigint-math discussion and was struck by the divergent handling of infinities vs. NaN, which I explored with eshost.

jugglinmike:
It's like spider-sense for test bugs :)

PR #3179, opened 2021-09-03
Simplify test

jugglinmike:
Remove semantics which are not necessary to exercise the behavior under
test.

### DIFF
--- a/test/built-ins/Math/hypot/Math.hypot_ToNumberErr.js
+++ b/test/built-ins/Math/hypot/Math.hypot_ToNumberErr.js
@@ -13,13 +13,10 @@ info: |
   3. For each element _number_ of _coerced_, do
 ---*/
 
-let uniqueErrorCount = 0;
-class UniqueError extends Error {
-  constructor() { super(); uniqueErrorCount++; }
-}
+var counter = 0;
 
 assert.throws(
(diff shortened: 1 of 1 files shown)


rwaldron (approved):

test/built-ins/Math/hypot/Math.hypot_ToNumberErr.js
```diff
@@ -13,26 +13,23 @@ info: |
...
     Math.hypot(
       Infinity,
       -Infinity,
       NaN,
       0,
       -0,
-      {valueOf(){ throw new UniqueError(); }},
-      {valueOf(){ throw new UniqueError(); }},
+      {valueOf: function(){ throw new Test262Error(); }},
+      {valueOf: function(){ counter++; }}
```
rwaldron, line 28: In the future: changing to _MethodDefinition_ syntax to _PropertyName_ `:` _AssignmentExpression_ syntax is unnecessary, please don't, unless you're specifically testing the latter syntax.

jugglinmike: We're not testing either syntax. The change is necessary for the purpose of simplification--the stated goal of this patch.

Issue #2978, opened 2021-05-01
Incorrect tests: tco-cross-realm-*


bakkot:
Before you read this issue: this is basically the lowest possible priority, since it concerns TCO and cross-realm calls interacting in a way which I suspect literally zero humans have encountered, and also I'm not at all convinced that the behavior being tested here is intentional. I'm writing this just so I don't forget about it. That said:

- [tco-cross-realm-class-construct.js](https://github.com/tc39/test262/blob/70bc32edab22b44db9d671ce505db8842ae200b6/test/language/expressions/call/tco-cross-realm-class-construct.js)
- [tco-cross-realm-class-derived-construct.js](https://github.com/tc39/test262/blob/70bc32edab22b44db9d671ce505db8842ae200b6/test/language/expressions/call/tco-cross-realm-class-derived-construct.js)
- [tco-cross-realm-fun-call.js](https://github.com/tc39/test262/blob/70bc32edab22b44db9d671ce505db8842ae200b6/test/language/expressions/call/tco-cross-realm-fun-call.js)
- [tco-cross-realm-fun-construct.js](https://github.com/tc39/test262/blob/70bc32edab22b44db9d671ce505db8842ae200b6/test/language/expressions/call/tco-cross-realm-fun-construct.js)

all became incorrect with the merge of https://github.com/tc39/ecma262/pull/2216, and didn't get updated in https://github.com/tc39/test262/pull/2878.

These are a little tricky to explain. For background, when the spec says to "Throw a TypeError exception", the realm of the constructor of that error is supposed to be derived from the execution context at the top of the execution context stack. [PrepareForTailCall](https://tc39.es/ecma262/multipage/#sec-preparefortailcall) is the abstract operation which is used to accomplish tail calls: its behavior is to pop the execution context stack, just before doing a call, with the idea that this represents popping the current function from the call stack so that it can be replaced by the new function. However, in a couple of places there is (or was) logic which could lead to an error being thrown between the point at which the execution context stack was popped and the point at which the new function's call pushed its own execution context. In that case, the error is (per spec) supposed to be constructed from the realm of the _caller_ of the thing which did the tail call, that is, of the grand-caller of the thing whose call ultimately was responsible for the error.

Calling a class as a function used to be such a place. (It's not anymore, which is why these tests are wrong; we'll get to that.)

[...]

PR #3181, opened 2021-09-03
Add test for normative change to TCO&revoked Proxy

jugglinmike:
Normative: phrase tail calls as discarding resources rather than popping
execution context stack
https://github.com/tc39/ecma262/pull/2495


bakkot:
While you're at it, you might want to delete the four `tco-cross-realm-*` tests in [this directory](https://github.com/tc39/test262/tree/70bc32edab22b44db9d671ce505db8842ae200b6/test/language/expressions/call). Those were already invalid since https://github.com/tc39/ecma262/pull/2216 (see https://github.com/tc39/test262/issues/2978), but are more obviously so as of tc39/ecma262#2495.

bakkot (approved):

jugglinmike:
Sounds good to me! gh-3182

rwaldron (approved):

PR #3189, opened 2021-09-09
RegExp.prototype.compile from other realm should throw TypeError from other realm

iamstolis:
This PR fixes the second test-case in `annexB/built-ins/RegExp/prototype/compile/this-cross-realm-instance.js` test. It is a symmetric test-case to the first test-case, i.e., the `TypeError` should come from the realm of the invoked built-in (not from the realm of the argument).


rwaldron:
Nice catch!

rwaldron (approved):

PR #3192, opened 2021-09-11
Test lastIndex overwrite with big lastIndex

gibson042:
Ref https://github.com/Moddable-OpenSource/moddable/issues/698

### DIFF
--- a/test/built-ins/RegExp/prototype/exec/failure-lastindex-set.js
+++ b/test/built-ins/RegExp/prototype/exec/failure-lastindex-set.js
@@ -1,0 +1,57 @@
+// Copyright (C) 2021 Richard Gibson. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: >
+    lastIndex is set to 0 after exhausting the string when global and/or sticky are set.
+esid: sec-regexpbuiltinexec
+info: |
+    RegExpBuiltinExec (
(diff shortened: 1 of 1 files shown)


test/built-ins/RegExp/prototype/exec/failure-lastindex-set.js
```diff
@@ -0,0 +1,71 @@
...
+
+  assert.sameValue(R_g.lastIndex, 0,
+      "global RegExp instance lastIndex must be reset after " + lastIndex +
+      " exceeds string length");
+  assert.sameValue(R_y.lastIndex, 0,
+      "sticky RegExp instance lastIndex must be reset after " + lastIndex +
+      " exceeds string length");
+  assert.sameValue(R_gy.lastIndex, 0,
+      "global sticky RegExp instance lastIndex must be reset after " + lastIndex +
+      " exceeds string length");
```
rwaldron, line 70: Not blocking, but in the future feel free to use template strings 👍

rwaldron (approved):

test/built-ins/RegExp/prototype/exec/failure-lastindex-set.js
gibson042: I'm glad you brought this up, because I was wondering as I wrote this (even though I ended up in an inconsistent state from use of `**`): #3196

PR #3195, opened 2021-09-13
SyntaxErrors in `ShadowRealm.evaluate` should propagate to parent

philomates:
If I'm following the proposed Shadow Realm spec, it says to throw a `SyntaxError` when the `sourceText` argument provided to `ShadowRealm.evaluate` is malformed syntax (https://tc39.es/proposal-shadowrealm/#sec-performrealmeval).

This PR updates the test262 test case to reflect that interpretation of the spec.

### DIFF
--- a/test/built-ins/ShadowRealm/prototype/evaluate/errors-from-the-other-realm-is-wrapped-into-a-typeerror.js
+++ b/test/built-ins/ShadowRealm/prototype/evaluate/errors-from-the-other-realm-is-wrapped-into-a-typeerror.js
@@ -15,7 +15,7 @@ assert.sameValue(
 
 const r = new ShadowRealm();
 
-assert.throws(TypeError, () => r.evaluate('...'), 'SyntaxError => TypeError');
+assert.throws(SyntaxError, () => r.evaluate('...'), 'SyntaxError exposed to Parent');
 assert.throws(TypeError, () => r.evaluate('throw 42'), 'throw primitive => TypeError');
 assert.throws(TypeError, () => r.evaluate('throw new ReferenceError("aaa")'), 'custom ctor => TypeError');
 assert.throws(TypeError, () => r.evaluate('throw new TypeError("aaa")'), 'Child TypeError => Parent TypeError');
--- a/test/built-ins/ShadowRealm/prototype/importValue/specifier-tostring.js
+++ b/test/built-ins/ShadowRealm/prototype/importValue/specifier-tostring.js
@@ -24,7 +24,7 @@ const specifier = {
 };
 
 assert.throws(Test262Error, () => {
-  r.importValue(specifier);
+  r.importValue(specifier, '');
 });
 
 assert.sameValue(count, 1);


test/built-ins/ShadowRealm/prototype/importValue/specifier-tostring.js
```diff
@@ -24,7 +24,7 @@ const specifier = {
 };
 
 assert.throws(Test262Error, () => {
-  r.importValue(specifier);
+  r.importValue(specifier, '');
```
philomates, line 27: `importValue` is a 2-arity function, so this change is needed (right?)

philomates:
@leobalter / @rwaldron: can you check this out when you have a moment, given you worked on these tests recently

test/built-ins/ShadowRealm/prototype/importValue/specifier-tostring.js
rwaldron: This change makes the test ambiguous. The test, as is, should only reach step 3: 

```
1. Let O be this value.
2. Perform ? ValidateRealmObject(O).
3. Let specifierString be ? ToString(specifier).
```

Adding a second empty string argument to this test would allow an implementation to validate the arguments out of order and still pass this test.

test/built-ins/ShadowRealm/prototype/evaluate/errors-from-the-other-realm-is-wrapped-into-a-typeerror.js
```diff
@@ -15,7 +15,7 @@ assert.sameValue(
 
 const r = new ShadowRealm();
 
-assert.throws(TypeError, () => r.evaluate('...'), 'SyntaxError => TypeError');
+assert.throws(SyntaxError, () => r.evaluate('...'), 'SyntaxError exposed to Parent');
```
rwaldron, line 18: Yeah, this looks right to me, given step 5 here https://tc39.es/proposal-shadowrealm/#sec-performrealmeval.

rwaldron (changes requested):
Remove the change in test/built-ins/ShadowRealm/prototype/importValue/specifier-tostring.js

rwaldron:
I've updated the polyfills to pass this test.

rwaldron:
@philomates I'm actually going to accept this by cherry picking the first commit! Thanks!!

rwaldron:
Cherrypicked: https://github.com/tc39/test262/commit/ba4bb68781821828985acfb7e51e272fa0d88a52

test/built-ins/ShadowRealm/prototype/importValue/specifier-tostring.js
philomates: what about making it explicit like this?
```javascript
const specifier = { ... };
const bindingName = {
  toString() {
    bindingNameCount += 1;
    throw new SomeOtherRandomError();
  }
};

 assert.throws(Test262Error, () => {  r.importValue(specifier, bindingName); });
```

PR #3198, opened 2021-09-14
Add tests for Unicode v14 Identifier{Start,Part}

mathiasbynens:
Issue: https://github.com/tc39/ecma262/issues/2514


jugglinmike:
@anba Would you mind also giving this a whirl in SpiderMonkey with Unicode 14?

anba:
The tests all pass when Unicode 14 is used.

PR #3199, opened 2021-09-14
Update RegExp property tests per Unicode v14

mathiasbynens:
Issue: https://github.com/tc39/ecma262/issues/2514


jugglinmike:
Thanks, @mathiasbynens! I'd like to help, but I frankly don't know how to constructively review tests like this. Do you have any thoughts about how a reviewer would validate your patch?

@anba: if this looks good to you, then I'd be willing to merge even before I'm ready to review personally.

mathiasbynens:
That's a good question, and to be honest the answer isn't crystal clear to me either.

The upstream change for the test generator is here: https://github.com/mathiasbynens/unicode-property-escapes-tests/commit/a38f6829de09bcf9cf928f764a9e6d03098ef25e#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519 which basically defers to other Unicode data-related packages. Tracing it all the way back to the source, we end up at https://github.com/node-unicode/node-unicode-data/commit/243f19e8f49e95bed3e92a1aeb7fce4c05b6f9f8#diff-f67dd27999ce20ea2e7a12a007c56ba704e590cb5bc0bef935928830e1df7a7b which affected the https://github.com/node-unicode/unicode-14.0.0 dependency. Perhaps you'd rather review those changes, and trust that the existing generator still does the right thing when given the right inputs based on the Unicode Standard?

As a double-check, we could also wait until a JS engine updates to Unicode v14 and then verify that they pass all the tests.

anba:
> @anba: if this looks good to you, then I'd be willing to merge even before I'm ready to review personally.

Yes, the changes look good to me.

> As a double-check, we could also wait until a JS engine updates to Unicode v14 and then verify that they pass all the tests.

I've built SpiderMonkey with Unicode 14 support and verified that all tests pass.

jugglinmike (approved):
That all sounds good--my thanks to both of you! I've left a note about this on [the original issue for Unicode property escapes](https://github.com/tc39/test262/issues/950) (which was still open, curiously).

PR #3144, opened 2021-08-11
Transform legacy format to harness assertions: test/built-ins/D*/**/*.js 

rwaldron:

PR #3201, opened 2021-09-15
Fix incorrect tests: new Date(-0).valueOf() and new Date(-1.23e-15).valueOf() should return +0.

woess:
Fixes test regressions introduced by https://github.com/tc39/test262/pull/3144 (1c24e4b915a090ae1878da606a9cd1e62ca815ca) which changed:
```diff
 var d5 = new Date(-0);
-if (d5.valueOf() !== -0) {
-  throw new Test262Error('#5: var d5 = new Date(-0); d5.valueOf() === -0;');
-}
+assert.sameValue(d5.valueOf(), -0, 'd5.valueOf() must return -0');
```
and
```diff
 var d12 = new Date(-1.23e-15);
-if (d12.valueOf() !== -0) {
-  throw new Test262Error('#12: var d12 = new Date(-1.23e-15); d12.valueOf() === -0;');
-}
+assert.sameValue(d12.valueOf(), -0, 'd12.valueOf() must return -0');
```

This is wrong. The correct result is positive 0. (See [TimeClip](https://tc39.es/ecma262/#sec-timeclip) and [ToIntegerOrInfinity](https://tc39.es/ecma262/#sec-tointegerorinfinity) which returns either an integer (cannot be -0), +∞, or -∞.)
Arguably, the tests were always wrong, suggesting that the correct result be `-0`, but technically they were valid since `0 === -0`.

### DIFF
--- a/test/built-ins/Date/prototype/valueOf/S9.4_A3_T1.js
+++ b/test/built-ins/Date/prototype/valueOf/S9.4_A3_T1.js
@@ -56,4 +56,4 @@ assert.sameValue(d11.valueOf(), 0, 'd11.valueOf() must return 0');
 
 // CHECK#12
 var d12 = new Date(-1.23e-15);
-assert.sameValue(d12.valueOf(), -0, 'd12.valueOf() must return -0');
+assert.sameValue(d12.valueOf(), 0, 'd12.valueOf() must return 0');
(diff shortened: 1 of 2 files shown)


jugglinmike (approved):
Good catch. Coincidentally, ChakraCore happens to pass the incorrect version of these tests.

It looks like the misleading pattern (strict equality with negative zero) occurs in other parts of the test suite, so I'll follow up with another patch to address that.

Thanks for the fix!

PR #3202, opened 2021-09-15
Improve coverage for ops involving negative zero

jugglinmike:
Prior to this commit, the modified tests used the strict equality
operator to compare computed values with negative zero. Due to the
semantics of that operator, these tests would spuriously pass if the
value under test was in fact positive zero.

Update the tests to be more precise by instead asserting equality with
the `assert.sameValue` utility method (since that method correctly
distinguishes between negative zero and positive zero).

Thanks to @woess for helping uncover this by [identifying a related problem](https://github.com/tc39/test262/pull/3201).


rwaldron (approved):
If only we had infinite time and resources to update all of those old Sputnik tests

PR #3205, opened 2021-09-17
Move IANA TimeZone tests to intl402

FrankYFTang:
IANA TimeZone name other than "UTC" is only required if the implementation " includes the ECMA-402 Internationalization API"
https://tc39.es/proposal-temporal/#sec-isvalidtimezonename
Therefore, we need to split the test and keep this one  without checking "America/Vancouver" and move those test on IATA timezone name other than UTC to under intl402
Add test cases to test withtout the ":" between TimeZoneUTCOffsetHour and TimeZoneUTCOffsetMinute
in the TimeZoneNumericUTCOffset string see https://tc39.es/proposal-temporal/#prod-TimeZoneNumericUTCOffset

### DIFF
--- a/test/built-ins/Temporal/Now/plainDateTimeISO/timezone-string-datetime.js
+++ b/test/built-ins/Temporal/Now/plainDateTimeISO/timezone-string-datetime.js
@@ -19,14 +19,6 @@ assert.throws(RangeError, () => Temporal.Now.plainDateTimeISO({ timeZone }), "ba
   "2021-08-19T1730-07:00",
   "2021-08-19T17:30-0700",
   "2021-08-19T1730-0700",
-  "2021-08-19T17:30[America/Vancouver]",
-  "2021-08-19T1730[America/Vancouver]",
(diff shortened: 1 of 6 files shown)


FrankYFTang:
@jugglinmike @Ms2ger

rwaldron:
@ptomato @Ms2ger PTAL

ptomato:
This is a good point, but I think it would be better to change these tests to use the `UTC` time zone instead of `America/Vancouver`. Otherwise the purpose of these tests is lost, because they are testing functionality (string parsing) that ought to be present in the engine regardless of whether it implements ECMA-402.

FrankYFTang:
> This is a good point, but I think it would be better to change these tests to use the `UTC` time zone instead of `America/Vancouver`. Otherwise the purpose of these tests is lost, because they are testing functionality (string parsing) that ought to be present in the engine regardless of whether it implements ECMA-402.

so you mean in the tests of the built-ins we keep have something like
"2021-08-19T17:30[UTC]", right?

rwaldron (approved):

PR #3212, opened 2021-09-23
Correct assertion messages and improve coverage

jugglinmike:
Following a recent normative change to the Resizable ArrayBuffer
proposal [1], the term "out of bounds" no longer applies to
"length-tracking" TypedArrays whose underlying ArrayBuffer has been
resized to match their byte offset.

Reflect this in the tests by renaming the condition from "out of bounds"
to "on boundary" and by adding new assertions for true "out of bounds"
conditions.

[1] https://github.com/tc39/proposal-resizablearraybuffer/pull/70


marjakh:
This PR is not modifying the byteOffset tests (which I originally reported), did you mean to do that in this PR or in a separate one?

jugglinmike:
gh-3211 is intended to resolve your bug report; this pull request addresses an omission from gh-3163 that I discovered while writing gh-3211.

rwaldron (approved):

PR #3213, opened 2021-09-23
Add test for freezing of module namespace object

mhofman:
Attempting to freeze the module namespace exotic object should not affect the `writable`-ity of the properties as that exercises the same `DefineOwnProperty` operation according to [`SetIntegrityLevel`](https://tc39.es/ecma262/#sec-setintegritylevel).

@erights discovered a [bug in v8](https://bugs.chromium.org/p/v8/issues/detail?id=12240) where, while the `Object.freeze` operation throws, it actually makes exported properties non-writable one by one.

At the request of @syg, I'm contributing a test against this behavior. The bug in v8 actually leads to a breakage of the objects invariants, however I'm not testing for that here as the root cause is the illegal freezing of the export.

### DIFF
--- a/test/language/module-code/namespace/internals/define-own-property.js
+++ b/test/language/module-code/namespace/internals/define-own-property.js
@@ -114,3 +114,24 @@ assert.throws(TypeError, function() {
       {value: "module", writable: false, enumerable: false,
        configurable: false});
 }, 'Object.defineProperty: Symbol.toStringTag');
+
+
+// Indirect change requested through Object.freeze
+
+// Try freezing more times than there are exported properties
+for (let i = 1; i < exported.length + 2; i++) {
+  assert.throws(
+    TypeError,
+    function () {
+      Object.freeze(ns);
+    },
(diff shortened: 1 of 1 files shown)


jugglinmike (commented):
[The `verifyProperty` function provided by the `propertyHelper.js` harness file](https://github.com/tc39/test262/blob/94500e1de995e34cc24e226d59d4c133b6bb4b4c/harness/propertyHelper.js#L27) checks the property descriptor *and* ensures that writing to the property actually works as expected. Would you mind adding

    includes: [propertyHelper.js]

to this test's metadata and using `verifyProperty` instead?

jugglinmike:
Oh, of course!

jugglinmike (approved):

mhofman:
@jugglinmike, could you help fix the erasure of my change that happened when #3219 got merged?

jugglinmike:
@mhofman Certainly. gh-3225 is a bit of a sledgehammer, which I'm reaching for only after a more nuanced response didn't pan out.

PR #3219, opened 2021-09-24
chore: convert compareArray() to assert.compareArray()

rwaldron:


harness/compareArray.js
```diff
@@ -27,7 +27,7 @@ compareArray.isSameValue = function(a, b) {
 };
 
 compareArray.format = function(spreadable) {
-  return `[${[...spreadable].map(String).join(', ')}]`;
+  return `[${[].map.call(spreadable, String).join(', ')}]`;
```
jugglinmike, line 30: Dunno if this will work--a "spreadable" is an iterable, right?

leobalter (dismissed):

harness/compareArray.js
leobalter: there is a whole problematic over this compareArray file and this format is being only used for the assertion message.

The assertion requires the values a and b (actual and expected) to be accessed in a for loop during the assertion, and the values would be accessed again for this formatting purposes. 

I believe this `[].map.call` is a conservative approach and I'd stick with it for discussing compareArray in a different PR.

harness/compareArray.js
```diff
@@ -26,8 +26,8 @@ compareArray.isSameValue = function(a, b) {
   return a === b;
 };
 
-compareArray.format = function(array) {
-  return `[${array.map(String).join(', ')}]`;
+compareArray.format = function(thingWithALengthPropertyAndNumericIndices) {
+  return `[${[].map.call(thingWithALengthPropertyAndNumericIndices, String).join(', ')}]`;
```
rwaldron, line 30: @jugglinmike @leobalter how about now?

jugglinmike: Anything that doesn't suggest we expect `map` to exhaust an iterable.

leobalter: The original code was using map directly on the given array-like object. Now it only uses map on top of the same object. I'd rather advance any further modifications/improvements to this file into another PR, WDYT?

jugglinmike: Sure (I'm not requesting any further modifications)

harness/compareArray.js
```diff
@@ -35,8 +35,11 @@ assert.compareArray = function(actual, expected, message) {
...
-    compareArray(actual, expected),
-    `Expected ${format(actual)} and ${format(expected)} to have the same contents. ${message}`
-  );
+  var result = compareArray(actual, expected);
+
+  // The following prevents actual and expected from being iterated and evaluated
+  // more than once unless absolutely necessary.
+  if (!result) {
+    assert(false, `Expected ${format(actual)} and ${format(expected)} to have the same contents. ${message}`);
+  }
```
leobalter, line 44: @jugglinmike Rick and I discussed about this assertion message that is now deferred to execute only if compareArray fails. That might _avoid_ conflicts for existing tests migrating from pure compareArray to assert.compareArray.

leobalter (approved):

mhofman:
This PR mangled my previous PR #3213 (and possibly other changes!)

test/language/module-code/namespace/internals/define-own-property.js
```diff
@@ -84,54 +84,33 @@ for (const key of ([...exported, Symbol.toStringTag])) {
...
-    "Object.freeze: " + String(i)
-  );
-}
-
-for (const key of exported) {
-  const desc = Object.getOwnPropertyDescriptor(ns, key);
-  assert.sameValue(desc.writable, true, String(key) + " writable");
-}
-
-assert(!Object.isFrozen(ns), "namespace object not frozen");
```
mhofman, line 137: These lines were removed

PR #3210, opened 2021-09-22
ShadowRealm: improved existing and additional coverage 

rwaldron:

PR #3220, opened 2021-09-27
Adapt ShadowRealm proxy coverage

philomates:
- I don't think `... instanceof Proxy` ([which is performed here](https://github.com/tc39/test262/pull/3210/files#diff-0dfc4b1a2e48f81f61b8ecf92f9aa16d306256682d107203c4372d4037b11beaR25)) is a valid check in JS; it throws an error on non-realm expressions as well
 - Add some additional coverage over proxy-ing wrapped functions to show that the underlying boundary checks are still made. Perhaps this is obvious and not worth having in the suite, so happy to remove it


philomates:
cc/ @rwaldron @leobalter

rwaldron (approved):

PR #3221, opened 2021-09-27
Split tests and fix false positive (ShadowRealms)

leobalter:
cc @philomates @rwaldron I did split this test file to separate the different goals of each test.

I did some other minor tweaks such as removing a false positive where the handler was calling `target(args)` and args is an array despite the `secretObj` sent and I avoided too much code inside the assert.throws function.


rwaldron:
image

rwaldron (approved):

philomates (commented):
thanks for the improvements, looks good!

PR #3222, opened 2021-09-29
Remove duplicate tests

ptomato:
While working on adding the Temporal tests from tc39/proposal-temporal (#3002) I noticed that these tests tested the same thing as each of the `branding.js` tests in the same directory. This removes the duplicate tests.


jugglinmike:
These tests definitely concern the same behavior; thanks for pointing out the redundancy! They aren't necessarily equivalent, though.

The tests named `branding.js` verify more receiver values, so at first glance, they seem like the ones we ought to keep. However, those tests also omit required parameters, and for some of the methods under test (e.g. [dateFromFields](https://tc39.es/proposal-temporal/#sec-temporal.calendar.prototype.datefromfields)), that omission would trigger an indistinguishable (to Test262) TypeError. The net effect is that the tests cannot identify common implementation bugs such as incorrect step sequences or omitted steps.

For other methods (e.g. [dateAdd](https://tc39.es/proposal-temporal/#sec-temporal.calendar.prototype.dateadd)), the omission would trigger a distinct error, so while it's a little misleading for readers, it's not as much of a concern.

We could address this by updating the `branding.js` tests to provide valid arguments. It's tempting to defer a change like that to a follow-up patch, but doing so would reduce coverage in the interim, so I'd rather address this issue atomically.

ptomato:
Thanks, that makes sense. I've updated the `branding.js` tests that correspond to the duplicate tests that I removed in this PR, for now, so that we don't lose any coverage. Maybe there should be a new issue for updating the rest of the branding tests to pass valid arguments?

jugglinmike (approved):

jugglinmike:
> Thanks, that makes sense. I've updated the `branding.js` tests that correspond to the duplicate tests that I removed in this PR, for now, so that we don't lose any coverage.

Thanks!

> Maybe there should be a new issue for updating the rest of the branding tests to pass valid arguments?

Sounds about right to me. See gh-3224

PR #3226, opened 2021-10-01
fix: corrections to assert.compareArray and assert.compareArray.format

rwaldron:
This is the just the updates assert.compareArray and assert.compareArray.format, extracted to a standalone commit

### DIFF
--- a/harness/compareArray.js
+++ b/harness/compareArray.js
@@ -26,8 +26,8 @@ compareArray.isSameValue = function(a, b) {
   return a === b;
 };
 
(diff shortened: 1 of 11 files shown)


jugglinmike (approved):
Thanks for splitting this out! The change highlights how our harness tests are currently a little inconsistent about when/how they use the harness to test the harness. You know I'd prefer to be more spartan about this kind of thing, and I know that you're rolling your eyes at me right now. Having said my piece, I'll save us all the heartache and let it go.

PR #3231, opened 2021-10-05
fix: ensure that symbol args as message don't break assert.compareArray

rwaldron:


jugglinmike (approved):
Elsewhere, I asked @rwaldron: do we *want* error messages to be Symbols? Would
it be better to encourage test authors to write words to describe the
assertion?

He pointed out that currently `assert.sameValue(x, y, Symbol())` is valid, so
it should be invalid for both, or valid for both.

PR #3236, opened 2021-10-05
Transform compareArray -> assert.compareArray: test/built-ins/Object/**/*.js

rwaldron:


jugglinmike (changes requested):
The assertion messages aren't accurate. We can probably remove them altogether in light of `assert.compareArray`'s built-in message generation

rwaldron:
@jugglinmike fixed!

rwaldron:
These CI failures are for generated tests, which there are none of in this PR

PR #3239, opened 2021-10-05
Transform compareArray -> assert.compareArray: test/language/**/*.js

rwaldron:


jugglinmike (changes requested):
Could you remove the references to `compareArray` from the assertion messages?

rwaldron:
@jugglinmike fixed!

PR #3248, opened 2021-10-07
mv tests depend on Intl to intl402.

FrankYFTang:
Tests which include Intl should not be under built-ins because some engine may not implement ECMA402. Move these test under intl402


FrankYFTang:
> I wonder if it would still be worth testing this in `built-ins/`?

Could we first move these tests to intl402 and if you can figure out some additional test without using Intl, we then add them into built-ins ?

Issue #3258, opened 2021-10-08
should when no complete records in built-ins/Temporal/PlainDate/prototype/toPlainDateTime/basic.js


FrankYFTang:
I saw a problem with built-ins/Temporal/PlainDate/prototype/toPlainDateTime/basic.js

The folllowing should thorw:
```
const date = new Temporal.PlainDate(2000, 5, 2);
const optionBag = date.toPlainDateTime({ hour: 11, minute: 30, second: 23 });
```
because 

https://tc39.es/proposal-temporal/#sec-temporal.plaindate.prototype.toplaindatetime
3.3.26 Temporal.PlainDate.prototype.toPlainDateTime ( [ temporalTime ] )
4. Set temporalTime to ? ToTemporalTime(temporalTime).

https://tc39.es/proposal-temporal/#sec-temporal-totemporaltime
4.5.2 ToTemporalTime ( item [ , overflow ] )
3-f. Let result be ? ToTemporalTimeRecord(item).

https://tc39.es/proposal-temporal/#sec-temporal-totemporaltimerecord
4.5.9 ToTemporalTimeRecord ( temporalTimeLike )
3. For each row of Table 3, except the header row, in table order, do
a. Let property be the Property value of the current row.
b. Let value be ? Get(temporalTimeLike, property).
c. If value is undefined, then
i. Throw a TypeError exception.

Internal Slot | Property
-- | --
[[Hour]] | "hour"
[[Microsecond]] | "microsecond"
[[Millisecond]] | "millisecond"
[[Minute]] | "minute"
[[Nanosecond]] | "nanosecond"
[[Second]] | "second"

So when 3-a get the second item as "microsecond" 
b. Let value be ? Get(temporalTimeLike, "microsecond"). will be undefined and throw.

FrankYFTang:
> This is a mistake in the spec text, I am planning to present this normative change at the next TC39 plenary: [tc39/proposal-temporal#1862](https://github.com/tc39/proposal-temporal/pull/1862) Assuming it is accepted, the test will subsequently be correct.

oh I see. I just cannot tell what is the intend. If that is the intend, then let me just close the test262 PR.

ptomato:
I think this can be closed, since the change was incorporated and the test is correct now.

ptomato:
Closing as complete, per above.

PR #3259, opened 2021-10-08
add  millisecond:0, microsecond:0, nanosecond: 0

FrankYFTang:
so it won't throw. address https://github.com/tc39/test262/issues/3258

Fix https://github.com/tc39/test262/issues/3258

@ptomato @Ms2ger @ryzokuken @jugglinmike


FrankYFTang:
So it is a spec bug not a test262 bug, let me close this PR for now.

Issue #3260, opened 2021-10-08
no spec text specify "local" must be an invalid calendar name


FrankYFTang:
In test/built-ins/Temporal/PlainYearMonth/calendar-invalid.js
const args = [
  TemporalHelpers.toPrimitiveObserver(actual, 1970, "year"),
  TemporalHelpers.toPrimitiveObserver(actual, 1, "month"),
  "local",
  TemporalHelpers.toPrimitiveObserver(actual, 1, "day")
];
assert.throws(RangeError, () => new Temporal.PlainYearMonth(...args));

This test is using the calendar name "locale" as an invalid calendar name.
However, in the spec text, the string "local" satisify the CalendarName and there are no spec text mandate "local" must be an invalid calendar name. Should we change it to "invalidCalendar" instead?

@ptomato @Ms2ger @jugglinmike

FrankYFTang:
ok, I see. One thing I would like to point out is in UTS35, it only define what is "Unicode Calendar Identifier", but does not define the term "Unicode BCP 47 calendar identifier"
https://unicode.org/reports/tr35/#UnicodeCalendarIdentifier
Should we change the spec text to remove "BCP 47" from that to make it clear?

@sffc

PR #3261, opened 2021-10-08
Use "invalidCalendar" which violate ISO8601 grammar

FrankYFTang:
no spec text mandate "local" MUST be an invalid calendar name therefore we should not used that in the text.

address https://github.com/tc39/test262/issues/3260
@Ms2ger @jugglinmike @ptomato

### DIFF
--- a/test/built-ins/Temporal/Calendar/from/calendar-object-invalid.js
+++ b/test/built-ins/Temporal/Calendar/from/calendar-object-invalid.js
@@ -7,5 +7,5 @@ description: Converting objects to Temporal.Calendar
 features: [Temporal]
 ---*/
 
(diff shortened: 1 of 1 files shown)


FrankYFTang:
There are no spec text mandate the implementation cannot use "iso-8601", "invalid-calendar" as part of the builtin calendar.
The spec text stated
https://tc39.es/proposal-temporal/#sup-temporal-isbuiltincalendar
"
15.6.1.1 IsBuiltinCalendar ( id )
This definition supersedes the definition provided in 12.1.2.

1. Let builtinCalendars be a List of Unicode BCP 47 calendar identifiers identifying the calendars for which the implementation provides the functionality of the constructed Temporal.Calendar objects. The list must include "iso8601". The ordering is irrelevant.
2. If builtinCalendars contains id, return true.
3. Return false.
"

PR #3273, opened 2021-10-19
Fix expectation in disambiguation-wrong-type.js

FrankYFTang:
@Ms2ger @ptomato @ryzokuken @jugglinmike


jugglinmike:
This test looks correct to me as it is written. Can you share your rationale for suggesting this change?

(It's generally a good practice to include that kind of information in the commit message so that, if the patch is accepted, its justification is accessible to others in the project's revision history.)

FrankYFTang:
epochNanoseconds of Temporal.ZonedDateTime should return the same value as the epochNanoseconds passed in in the constructor. The current expectation is simply different.

ptomato:
I agree with @jugglinmike here, I believe the test is correct as written. The epochNanoseconds property being tested here isn't set by the constructor, it's tested on the result of the call to `with()`.

FrankYFTang:
sorry, I got confused.

PR #3274, opened 2021-10-20
Revert #3250

### LINKED ISSUE #3272: Revert or remove #3250
The test changes in #3250 exposed a Temporal spec bug (https://github.com/tc39/proposal-temporal/issues/1878). The old tests before #3250 were validating the correct intended behavior, and they passed previously because the Temporal polyfill matches that behavior too. The spec will be updated to match that behavior.

One problem: we found this bug after the deadline for the Oct TC39 plenary meeting. That probably means that we probably can't merge the fix into the Temporal spec until after the December 2021 meeting.

How should we handle? Options include: 
a) Revert #2350, which would at least alert implementers that there's a problem, even if the spec doesn't yet match the tests
b) Remove these test cases until the Temporal spec catches up.
c) Something else?

We'd prefer (a), because we'd rather have implementers fail tests and reach out to us to find out why, vs. writing incorrect implementations and having to fix them later.  But we're flexible to whatever the Test262 maintainers think is best.

cc @Ms2ger @ptomato @12wrigja

justingrant:
Fixes #3272 by reverting a change to tests of`Temporal.PlainDateTime.prototype.since()`.

See tc39/proposal-temporal#1878 for details.

### DIFF
--- a/test/built-ins/Temporal/PlainDate/prototype/since/largestunit-higher-units.js
+++ b/test/built-ins/Temporal/PlainDate/prototype/since/largestunit-higher-units.js
@@ -21,6 +21,6 @@ TemporalHelpers.assertDuration(feb21.since(feb20, { largestUnit: "weeks" }), 0, 
 
 const lastFeb20 = Temporal.PlainDate.from("2020-02-29");
 const lastFeb21 = Temporal.PlainDate.from("2021-02-28");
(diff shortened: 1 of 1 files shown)


ptomato:
The corresponding change gained consensus in TC39, so this should be able to be merged now.

rwaldron (approved):

PR #3292, opened 2021-10-29
Fix ShadowRealm.prototype.evaluate wraps value in the builtin function realm

legendecas:
[[[Call]] of built-in function objects](https://tc39.es/ecma262/#sec-built-in-function-objects-call-thisargument-argumentslist) setup the execution context's realm to be the built-in function's creation realm. So the caller realm in https://tc39.es/proposal-shadowrealm/#sec-shadowrealm.prototype.evaluate should be this creation realm instead of the realm calling `ShadowRealm.prototype.evaluate`. Thus the return value should be wrapped in the same realm that creates the built-in function.

Refs: https://github.com/tc39/test262/pull/3230/files#r738857432
Refs: https://github.com/tc39/proposal-shadowrealm/issues/334#issuecomment-954314665


rwaldron (approved):

leobalter:
cc @philomates

PR #3294, opened 2021-10-29
Remove harness file, arrayContains.js

jugglinmike:
The `arrayContains` function has a number of deficiencies which make it
inappropriate for Test262:

- It apparently isn't very useful: despite being available for over 7
  years, fewer than ten tests use it
- It's misleading: its documentation reads, "Verify that a subArray is
  contained within an array." In reality, it only verifies that all the
  elements of one array are present in another--order does not matter.
- It's not ergonomic for test authors: it has been misused to create
  tests that were prone to false positives [1]
- It's not ergonomic for implementers: ostensibly designed for use with
  `assert`, the failure messages produced by tests that use it do not
  necessarily have very much context

All code in the "harness" directory adds to the total amount of
project-specific information which contributors are expected to to
learn. In light of the above deficiencies, the burden of this particular
harness file is unjustified.

Remove the harness file and its associated tests. Update the tests which
depend on it to express their expectations using alternate methods, and
strengthen the tests to assert element order wherever appropriate.

[1] https://github.com/tc39/test262/pull/3289

---

The new version of the tests for `for-in` is pretty verbose because the behavior under test is so low-level. I'd like to avoid using loops to test looping semantics. Also, [the order of properties on build-ins is undefined](https://github.com/tc39/ecma262/issues/2032), so the rewritten version of the two tests for built-ins preserves the surprising ambiguity of `arrayContains`.


rwaldron (approved):
Good riddance!

Issue #3245, opened 2021-10-06
Missing coverage: calling Array.prototype funcs with a TypedArray with a resizable / growable backing


marjakh:
Example:

```js
__v_1 = new Uint8Array();
Object.defineProperty(__v_1.__proto__, 'length', {value: 42});
Array.prototype.includes.call(new Uint8Array(), 2);
```

-> if we pass in a TA with resizable backing, weird things might happen. Especially, there are two different kinds of lengths going around, one from the length getter and one from the backing store, and the latter might change in surprising places.

Getting this right is also essential for security, so would be great if test262 had tests for this.

cc @jugglinmike

jugglinmike:
Thanks for the explanation!

> In this particular example, `__v_1.__proto__` is the same object as `Uint8Array.prototype`, so they're interchangable. Once we have a reference to the object and call Object.defineProperty on it, I don't think it's observable _how_ we got that reference (unless, getters or something, which this example doesn't have). Maybe I don't fully understand what you had in mind.

Got it--that all matches my understanding. (When I write tests, I try to avoid creating values that aren't strictly necessary, mostly to better highlight the behavior under test. By the same token, it's easy to overlook subtleties in the more particular tests, so I don't attempt simplification lightly.)

> Re: what should happen: In this example, the spec for `Array.prototype.includes` uses the length from the length getter to determine how many items to process, and then for each item, it calls Get. Get returns undefined if the index is out of bounds. So that's relatively straightforward.
>
> However, implementations might have all kinds of fast paths, and e.g., and an implementation might have the following logic: "If the receiver is a TypedArray, check its backing ArrayBuffer's length against the length from the getter and if the backing ArrayBuffer is big enough, loop over N elements and access raw memory." That would now be a security bug now that the backing ArrayBuffer's length can change (potentially during the iteration, for functions like `every`, `some`, `filter` etc).
>
> It's basically the same thing that happens for TypedArray.prototype functions. The purpose of this report was to point out we also need tests for Array.prototype functions where we pass a resizable TypedArray as the receiver.

I think I understand halfway :) Test262 doesn't currently have any tests for resizing an ArrayBuffer during traversal with the iterative Array methods like `every` (the same goes for the corresponding iterative TypeArray methods, actually). I can appreciate all the ways this can go wrong, and those hazards are probably why Test262 *does* have tests for resizing Arrays under such circumstances. I'll write up some new tests.

[...]

PR #3299, opened 2021-11-05
Add tests: resizing ArrayBuffer during iteration

jugglinmike:
The Resizable ArrayBuffer proposal allows implementations to reject any
resize operation, so the tests must accommodate that possibility.

Mitigate the complexity this entails by minimizing branches and by
deferring assertions to locations with shallow call stacks.

---

Resolves gh-3245

I tested this using V8's in-progress implementation (version 9.8.1). More on V8's behavior below.

As noted in the commit message, these tests are much more complex than what we typically accept in Test262 because they have to accommodate the proposal's optional behavior. Each test includes two invocations of `ArrayBuffer.prototype.resize`, and because each invocation may succeed or throw an error, they support a total of four paths of execution. I've simulated those by running these tests an additional three times using overridden implementations of `resize` (one which throws on first invocation, one which throws on second invocation, and one which throws on both invocations).

@marjakh a couple things about V8's behavior:

- seven of the Array.prototype methods (`every`, `filter`, `forEach`, `map`, `reduce`, `reduceRight`, and `some`) appear to be responding to changes in length rather than operating on a cached value. I don't see any bugs in those tests, but I'm wrong about that quite frequently. Given that V8's support is a work-in-progress, does that behavior seem plausible to you?
- `Array.prototype.map` and `TypedArray.prototype.map` both crash when invoked on a TypedArray whose backing buffer is resizable

      $ ~/.jsvu/v8 --harmony-rab-gsab -e 'ab=new ArrayBuffer(1, {xmaxByteLength: 2}); a=new Int8Array(ab); a.map(()=>{})'
      $
      $ ~/.jsvu/v8 --harmony-rab-gsab -e 'ab=new ArrayBuffer(1, {maxByteLength: 2}); a=new Int8Array(ab); a.map(()=>{})'
      /home/mike/.jsvu/v8: line 2: 20644 Trace/breakpoint trap   (core dumped) "/home/mike/.jsvu/engines/v8/v8" --snapshot_blob="/home/mike/.jsvu/engines/v8/snapshot_blob.bin" "$@"

### DIFF
--- a/test/built-ins/Array/prototype/every/callbackfn-resize-arraybuffer.js
+++ b/test/built-ins/Array/prototype/every/callbackfn-resize-arraybuffer.js
@@ -1,0 +1,73 @@
+// Copyright (C) 2021 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-array.prototype.every
+description: TypedArray instance buffer can be resized during iteration
+includes: [testTypedArray.js, compareArray.js]
+features: [TypedArray, resizable-arraybuffer]
+---*/
+
+// If the host chooses to throw as allowed by the specification, the observed
+// behavior will be identical to the case where `ArrayBuffer.prototype.resize`
+// has not been implemented. The following assertion prevents this test from
+// passing in runtimes which have not implemented the method.
+assert.sameValue(typeof ArrayBuffer.prototype.resize, 'function');
+
+testWithTypedArrayConstructors(function(TA) {
+  var BPE = TA.BYTES_PER_ELEMENT;
+  var buffer = new ArrayBuffer(BPE * 3, {maxByteLength: BPE * 4});
+  var sample = new TA(buffer);
+  var finalElement, expectedElements, expectedIndices, expectedArrays;
+  var elements, indices, arrays, result;
+
+  elements = [];
+  indices = [];
+  arrays = [];
+  result = Array.prototype.every.call(sample, function(element, index, array) {
+    if (elements.length === 0) {
+      try {
+        buffer.resize(2 * BPE);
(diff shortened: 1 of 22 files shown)


marjakh:
Hi,

Awesome, thanks for adding these tests! Indeed, the "every resize might or might not fail" is a significant source of complexity.

Re: "every, filter, forEach, map, reduce, reduceRight, some", why do you say they respond to changes in length rather than operating on the cached value?

E.g., %TypedArray%.prototype.every:

...
3. Let len be O.[[ArrayLength]]. <<<< Get the length here and cache it
...
5. Let k be 0.
6. Repeat, while k < len, <<<< This loop runs a fixed amount of iterations, based on the cached length
a. Let Pk be ! ToString(𝔽(k)).
b. Let kValue be ! Get(O, Pk).
c. Let testResult be ! ToBoolean(? Call(callbackfn, thisArg, « kValue, 𝔽(k), O »)).
d. If testResult is false, return false.
e. Set k to k + 1. <<< No length reloading inside the loop!
7. Return true.

-> The number of times the callback runs is not affected by the resizing. The values the callback gets as params are affected, since Get returns undefined.

Re: TA.p.map & A.p.map crashing, that's OK, I haven't implemented RAB / GSAB support for it yet.

rwaldron (approved):

marjakh:
Hmm, I think these tests are incorrect, @rwaldron , do you think they're correct?

Since this got merged already, opened https://github.com/tc39/test262/issues/3303 to track.

marjakh:
Oops, sorry, I was confused, the test are correct. The discrepancies are because I've so far only modified TypedArray.prorotype.{every, some, ...}, not yet Array.prototype.{every, some}. Sorry for the noise :)

Issue #3132, opened 2021-08-05
Need test for Intl NumberFormat v3 API


FrankYFTang:
Intl NumberFormat v3 API advanced to Stage 3 in TC39 2021 July meeting. We need tests for it.
https://github.com/tc39/proposal-intl-numberformat-v3
@sffc

jugglinmike:
I'll give it a shot!

jugglinmike:
Here's a progress update:

feature                        | status
-------------------------------|-------
formatRange                    | [in progress](https://github.com/tc39/test262/pull/3283)
Grouping Enum                  | [under review](https://github.com/tc39/test262/pull/3301)
New Rounding/Precision Options | (awaiting a volunteer)
Interpret Strings as Decimals  | (awaiting a volunteer)
Rounding Modes                 | [merged](https://github.com/tc39/test262/pull/3268)
Sign Display Negative          | [under review](https://github.com/tc39/test262/pull/3300)

jugglinmike:
I'll take "Interpret Strings as Decimals", next

jugglinmike:
I'll take "New Rounding/Precision Options", now

PR #3300, opened 2021-11-05

jugglinmike:
ECMA402: the "negative sign display" feature.

This is in service of gh-3132


sffc (approved):
LGTM except that I think this is more tests than I expect are needed for this feature :)

rwaldron (approved):

PR #3301, opened 2021-11-05

jugglinmike:
ECMA402: the "grouping enum" feature. It also includes coverage for the
formatting option as already defined by the latest version of ECMA402.

---

This is in service of gh-3132.

When adding test coverage for language proposals in "stage 3" of TC39's process, contributors to Test262 typically organize assertions within tests so that proposed semantics are encapsulated in dedicated files. This allows conforming engines to continue to benefit from coverage for standard behavior and to ignore the new behavior until they choose to opt in.

That is not always possible in this patch because the proposal includes semantics which contradict the current standard. I've included a note in the files where this was unavoidable so that it's easier for folks to update this when the proposal is standardized.

/cc @sffc


test/intl402/NumberFormat/prototype/format/useGrouping-de-DE.js
```diff
@@ -0,0 +1,29 @@
+// Copyright 2021 the V8 project authors. All rights reserved.
```
sffc, line 1: Suggestion: since you're including multiple locales, I think it would be useful to include "en-IN", which triggers the crore-style grouping separators.  You probably don't need too many other locales for this feature.

jugglinmike:
Thanks, @sffc! I've removed the tests concerning ko-KR and zh-TW. I've also added tests for en-IN, but I'm not confident about their correctness. Specifically, I'm unsure about this case:

    new Intl.NumberFormat('en-IN', {useGrouping: 'min2'}).format(100_000);

In [your description from the original issue](https://github.com/tc39/proposal-intl-numberformat-v3/issues/3), you wrote "display grouping separators when there are at least 2 digits in a group." My interpretation of that statement is that the above invocation should produce "100,000", so that's how I've written the corresponding assertions in the latest commit.

However, the description of "min2" in the ICU seems somewhat vague about what happens for values greater than 10,000:

> Display grouping using locale defaults, except do not show grouping on values smaller than 10000 (such that there is a minimum of two digits before the first separator).
>
> Note that locales may restrict grouping separators to be displayed only on 1 million or greater (for example, ee and hu) or disable grouping altogether (for example, bg currency).
>
> Locale data is used to determine whether to separate larger numbers into groups of 2 (customary in South Asia) or groups of 3 (customary in Europe and the Americas).

I think that the answer may have to do with that last statement, "Locale data is used [...]", but I don't know how to interpret that.

sffc:
Great question!  I opened an issue to discuss: #77

jugglinmike:
Cool! GitHub interpreted that reference in terms of Test262. Here's the corresponding issue in the proposal's repository: https://github.com/tc39/proposal-intl-numberformat-v3/issues/77

rwaldron (approved):

PR #3293, opened 2021-10-29
Tests for some of the normative changes made to Temporal at the October TC39 plenary

ptomato:
Best examined commit by commit; each commit message has a link back to the normative PR into the Temporal proposal.

This is what I have so far, I will open subsequent test262 PRs for the others. Some of the changes are already covered by existing tests.

PR #3304, opened 2021-11-11
Tests for some more of the normative changes made to Temporal at the October TC39 plenary

ptomato:
Here's a follow up to #3293, with tests from me and @sarahghp. More to follow soon.


justingrant (approved):
Cool!  I reviewed the strings in `round`/`total` commit and it looks good. Did a quick scan through the rest but didn't review closely.

rwaldron (approved):

FrankYFTang:
This PR has many tests are not right. The problem is when we have the following message, 
 `Uncallable ... ${notCallable} ...` and when notCallable is a Symbol. The message format itself will cause TypeError because Symbol cannot convert to String. Delete the "${notCallable}" from the message will fix it. Please fix it. there are more than hundreds of tests in this PR is ill this way.

ptomato:
I opened https://github.com/tc39/test262/pull/3311 to fix this.

PR #3307, opened 2021-11-12

jugglinmike:
ECMA402: the "interpret strings as decimals" feature.

---

The proposal text for this feature is still being refined; these tests reflect my understanding of the intention based on [this discussion in the proposal's repository](https://github.com/tc39/proposal-intl-numberformat-v3/pull/72).

@sffc would you mind taking a look?

### DIFF
--- a/test/intl402/NumberFormat/prototype/format/value-decimal-string.js
+++ b/test/intl402/NumberFormat/prototype/format/value-decimal-string.js
@@ -1,0 +1,24 @@
+// Copyright (C) 2021 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-number-format-functions
+description: >
+  Intl.NumberFormat.prototype.format converts its argument (called value) to a
+  number using ToIntlMathematicalValue.
+features: [Intl.NumberFormat-v3]
+locale: [en-US]
+---*/
+
+var nf = new Intl.NumberFormat('en-US');
+
+// The value 100,000 should only be interpreted as infinity if the input is the
+// string "Infinity".
+assert.sameValue(nf.format('100000'), '100,000');
+// The value -100,000 should only be interpreted as negative infinity if the
+// input is the string "-Infinity".
+assert.sameValue(nf.format('-100000'), '-100,000');
+
+assert.sameValue(nf.format('1.0000000000000001'), '1.0000000000000001');
+assert.sameValue(nf.format('-1.0000000000000001'), '1.0000000000000001');
+assert.sameValue(nf.format('987654321987654321'), '987,654,321,987,654,321');
+assert.sameValue(nf.format('-987654321987654321'), '-987,654,321,987,654,321');
--- a/test/intl402/NumberFormat/prototype/format/value-tonumber.js
+++ b/test/intl402/NumberFormat/prototype/format/value-tonumber.js
@@ -20,8 +20,8 @@ const toNumberResults = [
   [false, 0],
   ['42', 42],
   ['foo', NaN],
-  ['Infinity', 'Infinity'],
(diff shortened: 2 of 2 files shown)


test/intl402/NumberFormat/prototype/format/value-tonumber.js
```diff
@@ -19,7 +19,9 @@ const toNumberResults = [
   [true, 1],
   [false, 0],
   ['42', 42],
-  ['foo', NaN]
+  ['foo', NaN],
+  ['Infinity', 'Infinity'],
+  ['-Infinity', '-Infinity']
```
sffc, line 24: This doesn't test much; the test is

```

toNumberResults.forEach(pair => {
  const [value, result] = pair;
  assert.sameValue(nf.format(value), nf.format(result));
});
```

so you're only testing that `nf.format('Infinity')` is the same value as `nf.format('Infinity')`, which is an identity.

test/intl402/NumberFormat/prototype/format/value-decimal-string.js
```diff
@@ -0,0 +1,24 @@
...
+---*/
+
+var nf = new Intl.NumberFormat('en-US');
+
+// The value 100,000 should only be interpreted as infinity if the input is the
+// string "Infinity".
+assert.sameValue(nf.format('100000'), '100,000');
+// The value -100,000 should only be interpreted as negative infinity if the
+// input is the string "-Infinity".
+assert.sameValue(nf.format('-100000'), '-100,000');
```
sffc, line 19: I don't understand what invariant these tests are verifying.  Is the number 100,000 significant in the spec somewhere?

test/intl402/NumberFormat/prototype/format/value-tonumber.js
jugglinmike: Ah, that's right. I've changed the second element in each case to be the Number representation of Infinity.

test/intl402/NumberFormat/prototype/format/value-decimal-string.js
jugglinmike: This is my attempt to verify the following conditions in ToIntlMathematicalValue:

> 8. If mv is 10<sup>10000</sup> and str contains Infinity, return +∞.
> 9. If mv is -10<sup>10000</sup> and str contains Infinity, return -∞.

The intent of these assertions is to ensure that implementations reference not only the mathematical value of the input but also the original string representation. I believe these assertions would fail if an engine incorrectly implemented somethng like this:

> 8. If mv is 10<sup>10000</sup>, return +∞.
> 9. If mv is -10<sup>10000</sup>, return -∞.

Does that seem right to you?

test/intl402/NumberFormat/prototype/format/value-decimal-string.js
```diff
@@ -0,0 +1,24 @@
...
+
+assert.sameValue(nf.format('1.0000000000000001'), '1.0000000000000001');
```
FrankYFTang, line 21: is the result '1.0000000000000001'   here?
I think it should be '1' becaues the rounding w/ the default values of fractionDigits and signification digits.

test/intl402/NumberFormat/prototype/format/value-decimal-string.js
```diff
@@ -0,0 +1,24 @@
...
+assert.sameValue(nf.format('-1.0000000000000001'), '1.0000000000000001');
```
FrankYFTang, line 22: This is surely wrong, a negative number in the string should not be formatted as a posotive number here and without rounding.

sffc:
To close the loop: @FrankYFTang's comments were addressed in https://github.com/tc39/test262/pull/3335

PR #3311, opened 2021-11-17
Fix not-callable getOffsetNansecondsFor tests

ptomato:
Unfortunately, in #3304 I made a last-minute mistake when I added the
uncallable value to the assertion message, and neglected to test it;
Symbols can't be converted to strings like that, so these tests would
fail. This fixes the assertion messages.


Ms2ger (approved):

linusg:
Is anything preventing this PR from being merged? Would be nice to get these tests fixed on `main`. Thanks!

rwaldron (approved):

rwaldron:
> Is anything preventing this PR from being merged? Would be nice to get these tests fixed on main. Thanks!

Human availability... I think most folks in the US were out on holiday all last week.

PR #3317, opened 2021-11-19
Mention raw flag in test environment section (fixes #3310)

### LINKED ISSUE #3310: Conflict between "Test Environment" helper function and a testcase
test262 requires "Test Environment" to have helper functions:
https://github.com/tc39/test262/blob/main/CONTRIBUTING.md#test-environment
https://github.com/tc39/test262/blob/26f1f4567ee7e33163d961c867d689173cbb9065/CONTRIBUTING.md#L232-L240

there `assert` function is listed, and the following test redefines `assert` on the global.
https://github.com/tc39/test262/blob/26f1f4567ee7e33163d961c867d689173cbb9065/test/language/module-code/eval-gtbndng-indirect-faux-assertion.js#L34-L38

this doesn't work if the `assert` function is provided as global function that has `configurable: false` in the property descriptor.
and that's the case for SpiderMonkey's harness
https://searchfox.org/mozilla-central/rev/674b6582bafedc11fba6dc537769c88e9ed921d3/js/src/tests/test262/shell.js#12-21
```js
function assert(mustBeTrue, message) {
  if (mustBeTrue === true) {
    return;
  }

  if (message === undefined) {
    message = 'Expected true but got ' + assert._toString(mustBeTrue);
  }
}
```

there should be either of the following:
  * the "Test Environment" section specify the property descriptor of each helper
  * a rule that the testcase should not redefine/rewrite/remove properties listed in "Test Environment"

arai-a:
Clarify that helper functions aren't available if `raw` flag is specified.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -238,7 +238,7 @@ Read the [Test262 Technical Rationale Report](https://github.com/tc39/test262/wi
 
 ## Test Environment
 
-Each test case is run in a fresh JavaScript environment; in a browser, this will be a new &lt;iframe&gt;; for a console runner, this will be a new process.  The test harness code is loaded before the test is run.  The test harness defines the following helper functions:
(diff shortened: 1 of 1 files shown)


jugglinmike (approved):
Thanks!

PR #3318, opened 2021-11-19
Fix bad tests which `${notCallable}` caused TypeError for the case of Symbol

FrankYFTang:
The test itself which format the `${notCallable}`  will cause TypeError throws when the notCallable is a Symbol. Therefore we should remove that part from the debug message and the typeof should be good enough to provide useful informaiton. 

@ptomato @jugglinmike @Ms2ger @ryzokuken


Ms2ger:
I believe `${String(notCallable)}` would work, that might be nicer.

ptomato:
I'm on it: https://github.com/tc39/test262/pull/3311 But thanks!

PR #3321, opened 2021-11-19
Remove invalid tests against unspecified 'era' values

FrankYFTang:
Fix  https://github.com/tc39/test262/issues/3320 which coded not-yet-defined values of era by remove these two invalid tests

@ryzokuken @ptomato @jugglinmike @Ms2ger @sffc


sffc:
(please see discussion in the issue)

FrankYFTang:
Please accept these two tests, there are no way we can follow the current spec text (Temporal proposal) to reach out the expectation in this tests. The expectation in these two tests are from assumption out of the spec text and are NOT supported by current Temporal spec text, UTS35, CLDR nor ICU.

rwaldron (approved):

PR #3324, opened 2021-11-19
Improve documentation of harness files & raw flag

jugglinmike:
Prior to this commit, the term "harness" was used to refer both to the
like-named directory in the repository and the external projects which
consume the tests. Reserve the term "harness" to refer solely to the
files maintained within this project's `harness/` directory, and use
the term "runner" to refer to the external projects.

Prior to this commit, the term "helper" was not defined and therefore
introduced ambiguity in text that described harness files. Replace the
term with more explicit wording in terms of the harness files.

Because test runners have some leeway in how they include harness files,
the requirement that "[t]he test source code must not be modified in any
way" does not necessarily preclude the use of harness files. Update the
instructions for interpreting the `raw` frontmatter flag to explicitly
disallow harness files.

/cc @arai-a and @anba

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -179,7 +179,7 @@ negative:
 
 This key names a list of files in the `harness/` directory that will be included in the test environment prior to running the test.  Filenames **must** include the `.js` extension.
 
-When some code is used repeatedly across a group of tests, it may be appropriate to define it in a harness file. This practice increase test complexity, so it should be applied sparingly.
+When some code is used repeatedly across a group of tests, it may be appropriate to define it in a harness file. This practice increases test complexity, so it should be applied sparingly.
 
 #### author
 `author: [string]`
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -177,9 +177,9 @@ negative:
 #### includes
 `includes: [file-list]`
 
-This key names a list of helper files that will be included in the test environment prior to running the test.  Filenames **must** include the `.js` extension.
+This key names a list of files in the `harness/` directory that will be included in the test environment prior to running the test.  Filenames **must** include the `.js` extension.
 
-The helper files are found in the `harness/` directory. When some code is used repeatedly across a group of tests, a new helper function (or group of helpers) can be defined. Helpers increase test complexity, so they should be created and used sparingly.
+When some code is used repeatedly across a group of tests, it may be appropriate to define it in a harness file. This practice increase test complexity, so it should be applied sparingly.
 
 #### author
 `author: [string]`
@@ -195,8 +195,8 @@ This key is for boolean properties associated with the test.
 - **noStrict** - only run the test in "sloppy" mode
 - **module** - interpret the source text as
   [module code](https://tc39.github.io/ecma262/#sec-modules)
-- **raw** - execute the test without any modification (no helpers will be
-  available); necessary to test the behavior of directive prologue; implies
+- **raw** - execute the test without any modification (no harness files will be
+  included); necessary to test the behavior of directive prologue; implies
   `noStrict`
 - **async** - defer interpretation of test results until after the invocation
   of the global `$DONE` function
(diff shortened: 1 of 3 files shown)


INTERPRETING.md
```diff
@@ -26,8 +26,8 @@ code](https://tc39.github.io/ecma262/#sec-types-of-source-code).
 
 ### Test262-Defined Bindings
 
-The contents of the following files must be evaluated in the test realm's
-global scope prior to test execution:
+The contents of the following harness files must be evaluated in the test
```
arai-a, line 29: it may be better mentioning `raw` flag here as well, given "must" sounds like no exception.

jugglinmike: Good catch!

CONTRIBUTING.md
```diff
@@ -177,9 +177,9 @@ negative:
 #### includes
 `includes: [file-list]`
 
-This key names a list of helper files that will be included in the test environment prior to running the test.  Filenames **must** include the `.js` extension.
+This key names a list of files in the `harness/` directory that will be included in the test environment prior to running the test.  Filenames **must** include the `.js` extension.
 
-The helper files are found in the `harness/` directory. When some code is used repeatedly across a group of tests, a new helper function (or group of helpers) can be defined. Helpers increase test complexity, so they should be created and used sparingly.
+When some code is used repeatedly across a group of tests, it may be appropriate to define it in a harness file. This practice increase test complexity, so it should be applied sparingly.
```
jmdyck, line 182: s/increase/increases/

rwaldron:
@jugglinmike awaiting updates per @jmdyck's review. When ready, please rebase fixups.

CONTRIBUTING.md
```diff
@@ -195,8 +195,8 @@ This key is for boolean properties associated with the test.
 - **noStrict** - only run the test in "sloppy" mode
 - **module** - interpret the source text as
   [module code](https://tc39.github.io/ecma262/#sec-modules)
-- **raw** - execute the test without any modification (no helpers will be
-  available); necessary to test the behavior of directive prologue; implies
+- **raw** - execute the test without any modification (no harness files will be
+  included); necessary to test the behavior of directive prologue; implies
```
leobalter, line 199:
```suggestion
- **raw** - execute the test without any modification (no harness files or extra content must be
  included); necessary to test the behavior of directive prologue; implies
```
This is a nit pick, but reflects better the correct functionality of the raw flagged files. If they will be included or not, that depends on the runner, which is not controlled by this repo.

CONTRIBUTING.md
```diff
@@ -238,7 +238,9 @@ Read the [Test262 Technical Rationale Report](https://github.com/tc39/test262/wi
 
 ## Test Environment
 
-Each test case is run in a fresh JavaScript environment; in a browser, this will be a new &lt;iframe&gt;; for a console runner, this will be a new process.  The test harness code is loaded before the test is run unless the test file has the `raw` flag.  The test harness defines the following helper functions:
+Each test case is run in a fresh JavaScript environment; in a browser, this will be a new &lt;iframe&gt;; for a console runner, this will be a new process.
+
+Before the test is executed (and unless the test uses the `raw` frontmatter flag), the test runner will evaluate a number of files in the `harness/` directory. At a minimum, this procedure will define the following functions:
```
leobalter, line 243: The "at a minimum" part is slightly ambiguous. It's the minimum when raw flag is not present. I tried to rephrase the same content above to resolve this, WDYT?

```suggestion
The test runner must evaluate the required files from the `harness/` directory before the test is executed. Unless the `raw` frontmatter flag is present, this procedure will define at least the following functions:
```

CONTRIBUTING.md
```diff
@@ -281,7 +283,7 @@ $DONOTEVALUATE();
 var var = var;
 ```
 
-If the test case has the `raw` flag, this disallows the test to load any harness file including `$DONOTEVALUATE`. In this case, include a direct `throw "Test262: This statement should not be evaluated.";` statement:
+If the test case has the `raw` flag, the test runner will not load any harness file including the file which defines `$DONOTEVALUATE`. In this case, include a direct `throw "Test262: This statement should not be evaluated.";` statement:
```
leobalter, line 286:
```suggestion
If the test case has the `raw` flag, the test runner must not load any harness file including the file which defines `$DONOTEVALUATE`. In this case, include a direct `throw "Test262: This statement should not be evaluated.";` statement:
```

leobalter (commented):
I added some suggestions you can use at your discretion.

jugglinmike:
Thanks for the feedback, @leobalter! I think both of our phrasings are technically correct, but I think mine is more appropriate for the intended audience.

`CONTRIBUTING.md` is written for test authors, so they don't need to know about the responsibilities of the harness. Instead, we can tell them to take some things for granted. It's like saying "rest assured, the harness will do X."

`INTERPRETING.md`, on the other hand, is written for people writing harnesses. There, the terminology of RFC2119 appropriately conveys the obligations that Test262 is imposing.

It's a subtle distinction, but I think that using different language tailored to the two audiences will limit confusion about who is responsible for what.

rwaldron (approved):

PR #3325, opened 2021-11-20
Remove ${notCallable} in the fmt msg

FrankYFTang:
In these test notCallable could be a Symbol so we should not
reference to ${notCallable} in message format which will cause
TypeError while the message string try to convert notCallable
from Symbol to String.

@ptomato @ryzokuken @jugglinmike @Ms2ger


FrankYFTang:
I creaet this PR by the following script

find test |egrep -i temporal|egrep not-callable|xargs sed -i 's/${notCallable} //g'

FrankYFTang:
Close in favor of https://github.com/tc39/test262/pull/3311

PR #3334, opened 2021-12-04

jugglinmike:
ECMA402: the "new rounding/precision options" feature.

---

This is in service of gh-3132.

@sffc Would you mind taking a look? I'm especially interested in your thoughts on the tests for the behavior of the `roundingIncrement` option. I don't understand why the proposal supports values that differ only in order of magnitude (e.g. 5 and 50) because it seems to me that either one could be used to achieve the same result, just using different values of `maxFractionDigit`. For instance,

    {maximumFractionDigits: 2, roundingIncrement: 5}

seems equivalent to

    {maximumFractionDigits: 3, roundingIncrement: 50}

The tests reflect this interpretation, so I suspect that either some of the expected values are incorrect, or meaningful edge cases are omitted.

### DIFF
--- a/test/intl402/NumberFormat/prototype/format/format-rounding-increment-1.js
+++ b/test/intl402/NumberFormat/prototype/format/format-rounding-increment-1.js
@@ -1,0 +1,39 @@
+// Copyright 2021 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-intl.numberformat.prototype.format
+description: When set to `1`, roundingIncrement is correctly applied
+features: [Intl.NumberFormat-v3]
+includes: [testIntl.js]
+---*/
+
+var locales = [
+  new Intl.NumberFormat().resolvedOptions().locale, 'ar', 'de', 'th', 'ja'
+];
+var numberingSystems = ['arab', 'latn', 'thai', 'hanidec'];
+
+testNumberFormat(
+  locales,
+  numberingSystems,
+  {roundingIncrement: 1, maximumFractionDigits: 1},
+  {
+    '1.100': '1.1',
+    '1.125': '1.1',
+    '1.150': '1.2',
+    '1.175': '1.2',
+    '1.200': '1.2',
+  }
+);
+
+testNumberFormat(
+  locales,
+  numberingSystems,
+  {roundingIncrement: 1, maximumFractionDigits: 2},
+  {
+    '1.0100': '1.01',
+    '1.0125': '1.01',
+    '1.0150': '1.02',
+    '1.0175': '1.02',
+    '1.0200': '1.02',
+  }
+);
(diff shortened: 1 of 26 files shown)


rwaldron (approved):

rwaldron:
@sffc awaiting your review

rwaldron:
@sffc friendly nudge ;)

test/intl402/NumberFormat/prototype/format/format-rounding-increment-10.js
```diff
@@ -0,0 +1,39 @@
...
+  new Intl.NumberFormat().resolvedOptions().locale, 'ar', 'de', 'th', 'ja'
+];
+var numberingSystems = ['arab', 'latn', 'thai', 'hanidec'];
+
+testNumberFormat(
+  locales,
+  numberingSystems,
+  {roundingIncrement: 10, maximumFractionDigits: 2},
+  {
+    '1.100': '1.1',
```
sffc, line 20: As you noted, currently in the spec there is no difference between `{roundingIncrement:1, maximumFractionDigits:1}` and `{roundingIncrement:10, maximumFractionDigits:2}`. This is likely to change after we resolve the following issue:

https://github.com/tc39/proposal-intl-numberformat-v3/issues/81

In the mean time, we should move forward with these tests as written, and then update the tests after that issue has been resolved.

test/intl402/NumberFormat/prototype/format/format-rounding-increment-1.js
```diff
@@ -0,0 +1,39 @@
...
+  new Intl.NumberFormat().resolvedOptions().locale, 'ar', 'de', 'th', 'ja'
+];
+var numberingSystems = ['arab', 'latn', 'thai', 'hanidec'];
+
+testNumberFormat(
+  locales,
+  numberingSystems,
+  {roundingIncrement: 1, maximumFractionDigits: 1},
+  {
+    '1.100': '1.1',
```
sffc, line 20: Nit: Here and elsewhere: add a test for `'1.01': '1'` (soon to be `1.0`) since trailing zero behavior is important.

test/intl402/NumberFormat/prototype/format/format-rounding-priority-less-precision.js
```diff
@@ -0,0 +1,47 @@
+// Copyright 2021 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-intl.numberformat.prototype.format
+description: >
+  When roungingPriority is "lessPrecision", the constraint which produces the
```
sffc, line 6: Suggestion: Add tests for roundingPriority when all 4 settings are used at the same time (min+max fraction+significant)

sffc (commented):
LGTM with some suggestions!

rwaldron:
@sffc I'm going to merge these as-is, but create issues for the remaining suggestions.

sffc:
@romulocintra Can you take a look at my suggestions from above and apply them in a new PR?

Issue #3338, opened 2021-12-07
Add test about BigInt-Map/Set interaction


carlosame:
I could not find any test covering `BigInt`-`Map`/`Set` interaction, which was the subject of mozilla/rhino#1023. I'd suggest adding a test with a subset of Rhino's [bigint-mapset.js](https://github.com/mozilla/rhino/blob/Rhino1_7_14_RC1_Release/testsrc/jstests/es2020/bigint-mapset.js), something along the lines of:

```javascript
var lo53 = 9007199254740991 // 2^53 - 1
var big53 = BigInt("9007199254740991")
var big = BigInt("100000000000000000000000000000000000000000000000000000000000000000000000000000000001")
var d = 1e+83

// Map
var m = new Map()

m.set(lo53, 'lo53 key')
m.set(big53, 'big53 key')
m.set(big, 'bigint key')
m.set(d, 'double key')

assert.sameValue('lo53 key', m.get(lo53));
assert.sameValue('big53 key', m.get(big53));
assert.sameValue('bigint key', m.get(big));
assert.sameValue('double key', m.get(d));

// Set
var s = new Set()

s.add(lo53)

assert(s.has(lo53));
assert(!s.has(big53));

s.add(big53)
s.add(big)

assert(s.has(lo53));
assert(s.has(big53));
assert(s.has(big));
assert(!s.has(d));

s.add(d)

assert(s.has(big));
assert(s.has(d));
```

PR #3339, opened 2021-12-07
fix: Add test about BigInt-Map/Set interaction. Closes gh-3338

rwaldron:
cc @carlosame the relevant changes are in da08287672e484f2368115620543f203c2c851b9


carlosame:
Thank you @rwaldron for the quick response, although I fail to see a test that verifies that a `Map` or a `Set` can contain a given number as two different objects: a Number and a BigInt. I mean that a given Map (key) or Set may contain the number `1` but not `BigInt('1')`, and vice-versa. Perhaps that test is there but I did not see it.

The Rhino test checks numbers below and above 2**53 to prevent potential implementation-specific oddities (and the test in Rhino is also a test about `Number` as it is about `BigInt`).

rwaldron:
Ah, impedance mismatch between my eyes and brain when reading your example test code. Thanks for the heads up—I'll make some updates shortly.

rwaldron:
@carlosame https://github.com/tc39/test262/pull/3339/commits/f801cdb5ae301b6abf8acb4f5f2c3ef3389938ca

PR #3340, opened 2021-12-07
Add assertions of typeerror on ShadowRealm.prototype.importValue

legendecas:


test/built-ins/ShadowRealm/prototype/importValue/throws-if-import-value-does-not-exist.js
```diff
@@ -7,31 +7,20 @@ description: >
...
-    Let string be f.[[ExportNameString]].
-    Assert: Type(string) is String.
-    Let hasOwn be ? HasOwnProperty(exports, string).
-    If hasOwn is false, throw a TypeError exception.
+    1. Assert: exports is a module namespace exotic object.
+    2. Let f be the active function object.
+    3. Let string be f.[[ExportNameString]].
+    4. Assert: Type(string) is String.
+    5. Let hasOwn be ? HasOwnProperty(exports, string).
+    6. If hasOwn is false, throw a TypeError exception.
```
leobalter, line 23: Nit pick: AFAICT @rwaldron is avoiding the step numbers as they frequently change in the specs so either way is fine and correct enough for this frontmatter.

leobalter (dismissed):

test/built-ins/ShadowRealm/prototype/importValue/throws-if-import-value-does-not-exist.js
rwaldron: Yeah, it's "whatever". I leave them out now because I'm sick of updating the numbers when one step changes. Easier to drop in the change and move on.

test/built-ins/ShadowRealm/prototype/importValue/throws-if-import-value-does-not-exist.js
```diff
@@ -48,6 +37,11 @@ const r = new ShadowRealm();
 
 r.importValue('./import-value_FIXTURE.js', 'y')
   .then(
-    () => $DONE('Expected rejection'),
-    () => $DONE()
-  );
+    () => {
+      throw "(Expected rejection)";
```
rwaldron, line 41: Please `throw new Test262Error()` here.

rwaldron (changes requested):

test/built-ins/ShadowRealm/prototype/importValue/throws-if-import-value-does-not-exist.js
legendecas: Hmm, interesting point. I was thinking that having the step numbers may help to distinguish that these lines are spec steps rather than descriptive lines. What do you think we just number them all with `1.` just like what we have in the spec text?

legendecas:
Updated. Also applied a change suggested at https://github.com/tc39/test262/pull/3341#discussion_r764257767.

leobalter (approved):

test/built-ins/ShadowRealm/prototype/importValue/throws-if-import-value-does-not-exist.js
leobalter: I'm neutral to any of these options. the 1. might help indicating it's an ordered list

rwaldron: @legendecas honestly, you are more than welcome to include the line numbers if you'd like <3

rwaldron (approved):

rwaldron:
@legendecas as always, thanks!!

PR #3341, opened 2021-12-07
Add coverage on ShadowRealm.prototype.importValue onRejected

legendecas:
Add coverage on onRejected of PerformPromiseThen on step 17 of ShadowRealmImportValue.

I'm not sure if this is the correct way to test the HostImportModuleDynamically in test262 since the behavior is defined by hosts. But onRejected is defined in the spec, I'd believe we need to cover this.


test/built-ins/ShadowRealm/prototype/importValue/import-value_FIXTURE_throws.js
```diff
@@ -0,0 +1,4 @@
+// Copyright (C) 2021 Chengzhong Wu. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+throw new Error('foobar');
```
leobalter, line 4: All good in the tests! I don't have a good suggestion but I hope this test never changes to throwing an instance of TypeError so the importValue rejection is better captured and distinguished.

test/built-ins/ShadowRealm/prototype/importValue/throws-typeerror-import-throws.js
```diff
@@ -0,0 +1,34 @@
...
+esid: sec-realmimportvalue
+description: >
+  ShadowRealm.prototype.importValue rejects with TypeError when the imported script throws.
+info: |
+  RealmImportValue ( specifierString, exportNameString, callerRealm, evalRealm, evalContext )
+
+    ...
+    17. Return ! PerformPromiseThen(innerCapability.[[Promise]], onFulfilled, callerRealm.[[Intrinsics]].[[%ThrowTypeError%]], promiseCapability).
+
+flags: [async, module]
```
leobalter, line 13: I don't think we need the module flag for these tests.

```suggestion
flags: [async]
```

test/built-ins/ShadowRealm/prototype/importValue/throws-typeerror-import-syntax-error.js
```diff
@@ -0,0 +1,34 @@
...
+esid: sec-realmimportvalue
+description: >
+  ShadowRealm.prototype.importValue rejects with TypeError when the imported script unable to be parsed.
+info: |
+  RealmImportValue ( specifierString, exportNameString, callerRealm, evalRealm, evalContext )
+
+    ...
+    17. Return ! PerformPromiseThen(innerCapability.[[Promise]], onFulfilled, callerRealm.[[Intrinsics]].[[%ThrowTypeError%]], promiseCapability).
+
+flags: [async, module]
```
leobalter, line 13:
```suggestion
flags: [async]
```

leobalter (changes requested):

leobalter:
The module flag can be removed as the tests should run as regular script as well, similarly to the dynamic import tests.

rwaldron (changes requested):
Please rename the module fixture files so that they end in `_FIXTURE.js`. Other tools depend on the file name being consistent.

test/built-ins/ShadowRealm/prototype/importValue/throws-typeerror-import-syntax-error.js
rwaldron: @legendecas the reason for this change is that this must be tested in both non-strict mode and strict mode. The `module` flag will contradict that.

legendecas: I was following the flags here https://github.com/tc39/test262/blob/main/test/built-ins/ShadowRealm/prototype/importValue/throws-if-import-value-does-not-exist.js#L37. Do you agree we should remove it too? I'd believe the condition is similar for throws-if-import-value-does-not-exist.js.

legendecas:
@leobalter @rwaldron thanks for your suggestsions. Updated! :D

leobalter (dismissed):

test/built-ins/ShadowRealm/prototype/importValue/throws-typeerror-import-syntax-error.js
rwaldron: I'm really sorry, but I've realized that @leobalter and I were wrong: `module` is currently required for the `test262-harness` + `eshost` infra to resolve any kind of imported code. I will work on trying to update that, but for now we'll have to continue to include the `module` flag (which does follow the rules in INTERPRETING.md)

rwaldron: I can fix the other PR that was already landed, unless you get to it before me.

legendecas: @rwaldron updated!

rwaldron: @legendecas again, I apologize for the confusion and thank you again for writing these and for doing the other follow up!

rwaldron (approved):

PR #3342, opened 2021-12-09
Delete duplicate "`Array#reverse` is not a constructor" test

nicolo-ribaudo:
It's already tested by [test/built-ins/Array/prototype/reverse/not-a-constructor.js](https://github.com/tc39/test262/blob/main/test/built-ins/Array/prototype/reverse/not-a-constructor.js)


jugglinmike (approved):
Thanks! It looks like Test262 has some duplication across the "built-ins" directory. I'll follow up with a patch to address the rest.

PR #3344, opened 2021-12-10
Verify that `Temporal.*.prototype` isn't writeable

justingrant:
This PR verifies that the `prototype` property of Temporal class constructors is not writeable, not configurable, and not enumerable.

Context: https://github.com/tc39/proposal-temporal/issues/1965.

Note that `verifyProperty()` removes the property being tested, so the `prototype` prop needs to be tested before testing the constructor itself.

### DIFF
--- a/test/built-ins/Temporal/Calendar/prototype/prop-desc.js
+++ b/test/built-ins/Temporal/Calendar/prototype/prop-desc.js
@@ -1,0 +1,18 @@
+// Copyright (C) 2021 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal-calendar-prototype
+description: The "prototype" property of Temporal.Calendar
+includes: [propertyHelper.js]
+features: [Temporal]
+---*/
+
+assert.sameValue(typeof Temporal.Calendar.prototype, "object");
+assert.notSameValue(Temporal.Calendar.prototype, null);
+
+verifyProperty(Temporal.Calendar, "prototype", {
+  writable: false,
+  enumerable: false,
+  configurable: false,
+});
--- a/test/built-ins/Temporal/Duration/prototype/prop-desc.js
+++ b/test/built-ins/Temporal/Duration/prototype/prop-desc.js
@@ -1,0 +1,18 @@
+// Copyright (C) 2021 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal-duration-prototype
+description: The "prototype" property of Temporal.Duration
+includes: [propertyHelper.js]
+features: [Temporal]
+---*/
+
+assert.sameValue(typeof Temporal.Duration.prototype, "object");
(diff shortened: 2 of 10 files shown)


ptomato (approved):
:+1:

jugglinmike:
Thanks for the patch! Could you make these assertions from new files in the respective "prototype" directories? For instance, `test/built-ins/Temporal/Calendar/prototype/prop-desc.js`? That will make the distinction easier to see from the test names, improve the granularity of the tests, and avoid the tension you've identified resulting from assertion ordering.

justingrant:
> Could you make these assertions from new files in the respective "prototype" directories?

Yep!  Done. Thanks @jugglinmike!

test/built-ins/Temporal/Calendar/prototype/prop-desc.js
```diff
@@ -0,0 +1,20 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal-calendar-objects
+description: The "prototype" property of Temporal.Calendar
+includes: [propertyHelper.js]
+features: [Temporal]
+---*/
+
+const { Calendar } = Temporal;
```
Ms2ger, line 11: It's probably my fault for introducing this pattern in the first place, but I'd prefer not to propagate it, and just use `Temporal.Calendar` below.

test/built-ins/Temporal/Calendar/prototype/prop-desc.js
```diff
@@ -0,0 +1,20 @@
+// Copyright (C) 2020 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal-calendar-objects
```
Ms2ger, line 5:
```suggestion
esid: sec-temporal-calendar-prototype
```

Same for the other files

Ms2ger (approved):

test/built-ins/Temporal/Calendar/prototype/prop-desc.js
```diff
@@ -0,0 +1,20 @@
+// Copyright (C) 2020 Igalia, S.L. All rights reserved.
```
jugglinmike, line 1: Could you update this to 2021? (Sorry I missed this the first time around)

test/built-ins/Temporal/Calendar/prototype/prop-desc.js
justingrant: Done!

test/built-ins/Temporal/Calendar/prototype/prop-desc.js
justingrant: 👍 Done.

test/built-ins/Temporal/Calendar/prototype/prop-desc.js
justingrant: Yep, done!

justingrant:
All review comments are addressed, so this should be ready to merge unless there's more issues that come up.

Ms2ger (approved):

rwaldron (approved):
Thanks!

PR #3346, opened 2021-12-10
Temporal changes from October TC39 meeting, Part 3

ptomato:
Here are some more tests for normative PRs that reached consensus in the October TC39 plenary meeting, from @jessealama and me. I expect there to be one more batch of October ones, after which we'll start submitting tests for the ones that hopefully reach consensus in next week's December plenary :smile: 

Covers:
- https://github.com/tc39/proposal-temporal/pull/1796
- https://github.com/tc39/proposal-temporal/pull/1873
- https://github.com/tc39/proposal-temporal/pull/1874
- https://github.com/tc39/proposal-temporal/pull/1881
- https://github.com/tc39/proposal-temporal/pull/1889


rwaldron (approved):

FrankYFTang:
I don't think these tests work with these PR and the spec unless two more issues got address
1. you first need https://github.com/tc39/proposal-temporal/pull/1941 so ParseTemporalTimeZoneString will even consider the  TimeZoneUTCOffsetName as name
2. Step 7-a and 7-b of ParseTemporalTimeZoneString  need not to throw when the name is TimeZoneUTCOffsetName
@gibson042

PR #3347, opened 2021-12-10
Various fixes to Temporal tests

ptomato:
Here are some minor maintenance-type changes that I wrote while working on #3346, separated out of that PR for ease of review.

### DIFF
--- a/test/built-ins/Temporal/TimeZone/prototype/getNextTransition/instant-string.js
+++ b/test/built-ins/Temporal/TimeZone/prototype/getNextTransition/instant-string.js
@@ -7,11 +7,11 @@ description: Conversion of ISO date-time strings to Temporal.Instant instances
 features: [Temporal]
 ---*/
 
-const instance = new Temporal.TimeZone("America/Vancouver");
+const instance = new Temporal.TimeZone("UTC");
 
 let str = "1970-01-01T00:00";
 assert.throws(RangeError, () => instance.getNextTransition(str), "bare date-time string is not an instant");
-str = "1970-01-01T00:00[America/Vancouver]";
+str = "1970-01-01T00:00[UTC]";
 assert.throws(RangeError, () => instance.getNextTransition(str), "date-time + IANA annotation is not an instant");
 
 // The following are all valid strings so should not throw:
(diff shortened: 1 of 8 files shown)


test/built-ins/Temporal/TimeZone/prototype/getNextTransition/instant-string.js
```diff
@@ -3,22 +3,26 @@
...
+const valids = [
   "1970-01-01T00:00Z",
   "1970-01-01T00:00+01:00",
-  "1970-01-01T00:00Z[America/Vancouver]",
-  "1970-01-01T00:00+01:00[America/Vancouver]",
-].forEach((str) => instance.getNextTransition(str));
+  "1970-01-01T00:00Z[UTC]",
+  "1970-01-01T00:00+01:00[UTC]",
+];
+for (const str of valids) {
```
EarthyOrange, line 25: Question: Can `assert.compareArray` be used instead of the `for` loop and `assert.sameValue`?

ptomato: Probably there would be a way to use it, but what array did you have in mind to go on the right side of the comparison?

rwaldron: @ptomato probably something like...

```js
const valids = [
  "1970-01-01T00:00Z",
  "1970-01-01T00:00+01:00",
  "1970-01-01T00:00Z[UTC]",
  "1970-01-01T00:00+01:00[UTC]",
];
assert.compareArray(
  valids.map(valid => instance.getNextTransition(valid)), 
  [null, null, null, null]
)
```

But that's not a blocking request of any kind (I don't expect you to make that change)

rwaldron (approved):

test/built-ins/Temporal/TimeZone/prototype/getNextTransition/instant-string.js
ptomato: Thanks! Happy to change this in a follow up, but I personally find it clearer as currently written.

PR #3351, opened 2021-12-14
Add test for shadow realms wrapped function behavior on throw

jjgriego:
There was already existing coverage for when `evaluate`'d code directly throws, but not for wrapped functions; TypeError from the destination realm should be thrown whenever the a wrapped function throws, so cover that.

I signed the CLA--let me know if anything is missing : )

### DIFF
--- a/test/built-ins/ShadowRealm/prototype/evaluate/wrapped-function-throws-typeerror-on-exceptional-exit.js
+++ b/test/built-ins/ShadowRealm/prototype/evaluate/wrapped-function-throws-typeerror-on-exceptional-exit.js
@@ -1,0 +1,39 @@
+// Copyright (C) 2021 Igalia SL. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-wrapped-function-exotic-objects-call-thisargument-argumentslist
+description: >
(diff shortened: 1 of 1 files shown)


jjgriego:
cc @legendecas @leobalter for visibility

legendecas (approved):
👍

test/built-ins/ShadowRealm/prototype/evaluate/wrapped-function-throws-typeerror-on-exceptional-exit.js
```diff
@@ -0,0 +1,39 @@
...
+    }
+    return 'normal exit';
+  }`)(() => { throw Error("ahh"); }),
+  'ok',
+  'WrappedFunction throws TypeError (from the calling realm) if the wrapped callable throws any exception'
+);
+
+assert.throws(
+  TypeError,
+  () => r.evaluate('() => { throw new Error("ahh"); }')(),
```
rwaldron, line 37: I think this could probably just be `r.evaluate('() => { throw new Error("ahh"); }')`, but that's not a blocking request:

rwaldron (approved):
Thanks!

PR #3354, opened 2021-12-15
added tests for Array.prototype.group

polsevev:
Added tests for proposal Array Grouping groupBy

CC @codehag


rwaldron:
Again, very cool that you jumped right in on these :D we'll also go over this one during our video call on Friday. See you then!!

ljharb (commented):
This looks great! I compared it against my polyfill's tests, and noticed this PR isn't testing that a non-function callbackfn throws a TypeError, and that the [[Prototype]] of the returned object is null.

test/built-ins/Array/prototype/groupBy/callback-arg.js
```diff
@@ -0,0 +1,33 @@
...
+
+  6. Repeat, while k < len
+
+      a. Let Pk be ! ToString(𝔽(k)).
+      b. Let kValue be ? Get(O, Pk).
+      c. Let propertyKey be ? ToPropertyKey(? Call(callbackfn, thisArg, « kValue, 𝔽(k), O »)).
+      d. Perform ! AddValueToKeyedGroup(groups, propertyKey, kValue).
+      e. Set k to k + 1.
+  ...
+
```
rwaldron, line 20:
```suggestion
features: [Array.prototype.groupBy]
```

test/built-ins/Array/prototype/groupBy/null.js
```diff
@@ -0,0 +1,21 @@
...
+
+  3. If IsCallable(callbackfn) is false, throw a TypeError exception.
+
+  ...
+
+---*/
+
+
+assert.throws(TypeError, function() {
+  Array.prototype.groupBy.call(null,null);
```
rwaldron, line 20:
```suggestion
Array.prototype.groupBy.call(null, () => {
  throw new Test262Error();
});
```

rwaldron: And combine all of the invalid this tests into one file :)

test/built-ins/Array/prototype/groupBy/callback-arg.js
```diff
@@ -0,0 +1,33 @@
...
+---*/
+
+
+const arr = [-0,0,1,2,3];
+
+const sentinel = {};
+
+const map = arr.groupBy.call(sentinel, (n,i,testArr) => {
```
legendecas, line 28: IIUC, `LengthOfArrayLike` of the `sentinel` is 0, the callback won't be called in this case.

test/built-ins/Array/prototype/groupBy/length.js
```diff
@@ -0,0 +1,29 @@
...
+assert.sameValue(
+  Array.prototype.groupBy.length, 1,
+  'The value of `Array.prototype.groupBy.length` is `1`'
+);
+  
+verifyProperty(Array.prototype.groupBy, "length", {
+  enumerable: false,
+  writable: false,
+  configurable: true
+});
```
leobalter, line 29:
```suggestion
verifyProperty(Array.prototype.groupBy, "length", {
  value: 1,
  enumerable: false,
  writable: false,
  configurable: true
});
```

test/built-ins/Array/prototype/groupBy/name.js
```diff
@@ -0,0 +1,28 @@
...
+assert.sameValue(
+    Array.prototype.groupBy.name, 'groupBy',
+    'The value of `Array.prototype.groupBy.name` is `"groupBy"`'
+);
+  
+verifyProperty(Array.prototype.groupBy, "name", {
+  enumerable: false,
+  writable: false,
+  configurable: true
+});
```
leobalter, line 28:
```suggestion
verifyProperty(Array.prototype.groupBy, "name", {
  value: "groupBy",
  enumerable: false,
  writable: false,
  configurable: true
});
```

test/built-ins/Array/prototype/groupBy/undefined.js
```diff
@@ -0,0 +1,21 @@
...
+  3. If IsCallable(callbackfn) is false, throw a TypeError exception.
+
+  ...
+
+---*/
+
+
+assert.throws(TypeError, function() {
+  Array.prototype.groupBy.call(undefined);
+});
```
leobalter, line 21:
```suggestion
var calls = 0;

assert.throws(TypeError, function() {
  Array.prototype.groupBy.call(undefined, () => calls++);
});
assert.sameValue(calls, 0, "callbackfn should not be called");
```

test/built-ins/Array/prototype/groupBy/emptyList.js
```diff
@@ -0,0 +1,23 @@
...
+    a. Let elements be ! CreateArrayFromList(g.[[Elements]]).
+    b. Perform ! CreateDataPropertyOrThrow(obj, g.[[Key]], elements).
+
+  ...
+
+---*/
+
+const obj = [].groupBy(i=>i);
+
+assert.sameValue(Object.keys(obj).length, 0);
```
leobalter, line 23: You might also benefit testing that `obj` is different to the original array and asserting the callbackfn is not called. Something like:

```js
var calls = 0;
const original = [];
const obj = original.groupBy(() => calls++);

assert.sameValue(calls, 0, "callback function should not be called");
assert.notSameValue(original, obj, "groupBy returns a different object");
assert.sameValue(Object.keys(obj).length, 0);
```

leobalter (changes requested):
The tests are good, I added a few suggestions only.

You may also wanna test:

- the return type of groupBy
- assert this value inside the callbackfn w/ and w/o `thisArg`
- non-callback values for callbackfn
- non valid values for the this value
- tainted callbackfns (throwing an exception).
- callbackfn calls with terse arrays like `[,,,1,,2]`
- array like values for the this value, e.g. `[].groupBy.call({length: 2, '1': 'foo'}, callbackfn);`

hemanth:
This should be `group` instead of `groupBy` now.

jridgewell (commented):
https://github.com/tc39/test262/pull/3354#pullrequestreview-848251319

> the return type of groupBy

done.

> assert this value inside the callbackfn w/ and w/o thisArg

done.

> non-callback values for callbackfn

done.

> non valid values for the this value

done.

> tainted callbackfns (throwing an exception).

Do we have to?

> callbackfn calls with terse arrays like [,,,1,,2]

done.

> array like values for the this value, e.g. [].groupBy.call({length: 2, '1': 'foo'}, callbackfn);

Done.

test/built-ins/Array/prototype/group/callback-arg.js
```diff
@@ -0,0 +1,36 @@
...
+      e. Set k to k + 1.
+  ...
+features: [Array.prototype.group]
+---*/
+
+
+const arr = [-0, 0, 1, 2, 3];
+
+const sentinel = {};
+let calls = 0;
```
ptomato, line 27: `sentinel` is not used, this seems to be left over from one of the other tests?
```suggestion
let calls = 0;
```

test/built-ins/Array/prototype/group/emptyList.js
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) 2021 Ecma International.  All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-array.prototype.group
+description: Array.prototype.group populates object with correct keys and values
```
ptomato, line 6: This description doesn't quite seem to fit the test; maybe something like this?
```suggestion
description: Callback is not called and object is not populated if the array is empty
```

test/built-ins/Array/prototype/group/groupLength.js
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) 2021 Ecma International.  All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-array.prototype.group
+description: Array.prototype.group populates object with correct keys and values
```
ptomato, line 6: Same here, I'd suggest something like:
```suggestion
description: Callback can return numbers that are converted to property keys
```

test/built-ins/Array/prototype/group/sparse-array.js
```diff
@@ -0,0 +1,36 @@
...
+const obj = array.group(() => {
+  calls++;
+  return 'key';
+});
+
+assert.sameValue(calls, 3);
+
+assert.sameValue(0 in obj['key'], true);
+assert.sameValue(1 in obj['key'], true);
+assert.sameValue(2 in obj['key'], true);
```
ptomato, line 35: `assert.sameValue(..., true)` is going to produce a pretty context-less error message if it fails, something like "Expected true but got false". Plain `assert()` won't be any better (at least it's shorter) but I'd recommend adding assertion messages here:
```suggestion
assert(0 in obj['key'], 'group has a first element');
assert(1 in obj['key'], 'group has a second element');
assert(2 in obj['key'], 'group has a third element');
```

test/built-ins/Array/prototype/group/sparse-array.js
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) 2021 Ecma International.  All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-array.prototype.group
+description: Array.prototype.group populates object with correct keys and values
```
ptomato, line 6:
```suggestion
description: Array.prototype.group adds undefined to the group for sparse arrays
```

ptomato (approved):
Thanks for resuming this PR, and apologies that it took some time to get around to it. (I'll start on groupToMap as soon as possible.) I think this is basically ready to merge. Some small suggestions, but keeping in mind how long this has already taken to review, I've made them into GitHub suggestions that you can just click to accept or not, so that the review process don't cause any more delay on this PR.

I have some suggestions for additional things that could be tested, but for the same reason I wouldn't say that they should block this PR. They can be added in a second PR if you are so inclined:
- Test throwing on an arraylike with a value of its `length` property that can't be converted to an integer, or a getter that throws
- Test throwing in step 6.b if _Pk_ is a property with a getter that throws
- Test throwing if the callback returns a value that can't be converted to a property key (e.g. an object with a toString method that throws)

Thanks again for resuming this work!

ptomato:
> > tainted callbackfns (throwing an exception).
> 
> Do we have to?

Would be nice, but as with the other suggestions I don't see a need to block this PR on having that.

Another one I thought of: shortening and lengthening the array during the iteration.

jridgewell:
> Test throwing on an arraylike with a value of its length property that can't be converted to an integer, or a getter that throws

Done.

> Test throwing in step 6.b if Pk is a property with a getter that throws

Done.

> Test throwing if the callback returns a value that can't be converted to a property key (e.g. an object with a toString method that throws)

Done.

> tainted callbackfns (throwing an exception).

Done.

ptomato (approved):
Awesome, thanks!

PR #3358, opened 2021-12-18
Fix: correct Resizable ArrayBuffer/Array tests

jugglinmike:
Prior to this patch, the tests for certain Array prototype methods
incorrectly expected all elements to be "visited" even when a given
property was not present according to HasProperty.

Update the tests to reflect the fact that under such conditions, the
removed element is not visited.

@marjakh  @phoddie Since I authored the mistaken tests, it's particularly important that someone closer to an implementation reviews this patch. Would you mind taking a look?


rwaldron (approved):
Nice cleanup and repairs

phoddie:
Looks good. Changes reviewed and run.

PR #3360, opened 2021-12-27
Add tests for \p{…} with properties of strings

mathiasbynens:
This functionality is part of the RegExp `v` flag proposal: https://github.com/tc39/proposal-regexp-set-notation


harness/regExpUtils.js
```diff
@@ -31,23 +31,65 @@ function buildString(args) {
...
-  if (!regex.test(string)) {
+function printCodePoint(codePoint) {
+  const hex = codePoint
+    .toString(16)
+    .toUpperCase()
+    .padStart(6, "0");
+  return `U+${hex}`;
+}
+
+function printStringCodePoints(string) {
```
mathiasbynens, line 42: Note: I’ve extracted the `printCodePoint` and `printStringCodePoints` logic, but haven’t added these to the `defines` frontmatter above since they’re not used directly in any tests. They’re just called from the other helpers (which are in `defines`) in this file. Let me know if they need to be added to `defines` regardless.

test/built-ins/RegExp/property-escapes/generated/strings/Basic_Emoji-negative-CharacterClass.js
```diff
@@ -0,0 +1,20 @@
+// Copyright 2021 Mathias Bynens. All rights reserved.
```
mathiasbynens, line 1: Question about the location of these new tests: should they go under [Test262’s `test/built-ins/RegExp/property-escapes/` directory](https://github.com/tc39/test262/tree/main/test/built-ins/RegExp/property-escapes) under a new `strings` subdirectory (like in the current patch), or in a new sibling directory alongside `property-escapes`, or somewhere else?

rwaldron: For now the directory structure is fine. I doubt it will need to change, but if it does that's not a problem.

harness/regExpUtils.js
rwaldron: > frontmatter above since they’re not used directly in any tests.

That's perfect, thank you.

rwaldron (approved):
Thanks for covering all of the admin parts on this!

FrankYFTang:
> Note: This proposal is not yet at Stage 3, so you may want to hold off on merging the PR.

PR #3366, opened 2022-01-04
Add identifier tests for private class fields specifically

mathiasbynens:
@bathos discovered that multiple engines implement this incorrectly:
https://twitter.com/bhathos/status/1478370049303928832

With that in mind, it’s probably worth adding some test coverage.
Test generator patch: https://github.com/mathiasbynens/caniunicode/pull/6


rwaldron (approved):
Nice use of the `info:` tag!

Issue #1997, opened 2018-12-15
YAML formatting restrictions


leobalter:
We should restrict the YAML format for the test files frontmatter to improve interoperability.

Many test runners depend on quick but limited yaml parsers and it's ok to have style restrictions identified in the importing processes for Test262.

Some of them I can already tell from scratch:

- [ ] no leading or trailing spaces from inline lists. e.g.: `[ leading-space, foo, trailing-space ]`
- [ ] no leading or trailing blank lines from text blocks. e.g. the info tag cannot start with an empty line
- [ ] lists should always be inline. Always wrapped with square brackets and comma-space divided words. e.g.:

```diff
features:
-  - var
-  - foo
-  - bar
+  [var, foo, bar]
```

There are might be more items we want to identify here. We should increase the list as necessary.

cc @caitp @gsathya @psmarshall

gsathya:
Looks like JSC's parser is having trouble with the YAML as well: https://bugs.webkit.org/show_bug.cgi?id=193053

leobalter:
@gsathya yes, the issues we can list here might affect at least V8, JSC and SpiderMonkey. Turns out it Yaml parsers are more challenging than one can expect.

gsathya:
Another one: https://github.com/tc39/test262/issues/1997

littledan:
Thanks for working on this important issue, @leobalter.

kmiller68:
Would it also be possible to prohibit inline value strings that span multiple lines? That seems to cause issues for JSC's YAML parser... 

For example:
```
description: Security Throws a TypeError if start coercion detaches
  the buffer
```

Ideally, this would be formatted as one line or with "|"/">"

rkirsling:
Opened #2505 to address the specific issue @kmiller68 identified.

PR #3369, opened 2022-01-06
tools: enforce restriction on YAML negative key

ossy-szeged:
The phase field must precede the type field for negative tests
to have a consistent style and be able to parse easier.
Related to the goal of https://github.com/tc39/test262/issues/1997

Added this check to the linting script and updated tests accordingly.

### DIFF
--- a/tools/lint/test/fixtures/test/negative_invalid_phase_type_order.js
+++ b/tools/lint/test/fixtures/test/negative_invalid_phase_type_order.js
@@ -1,15 +1,15 @@
+NEGATIVE
 ^ expected errors | v input
-// Copyright (C) 2019 Leo Balter. All rights reserved.
+// Copyright (C) 2017 Mike Pennisi. All rights reserved.
 // This code is governed by the BSD license found in the LICENSE file.
 /*---
 esid: sec-assignment-operators-static-semantics-early-errors
 description: Minimal test
 negative:
-  phase: parse
   type: SyntaxError
-flags: [raw]
+  phase: parse
 ---*/
 
-throw "Test262: This statement should not be evaluated.";
+$DONOTEVALUATE();
 
 !!!
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -159,10 +159,10 @@ single line comment syntax.
 #### negative
 `negative: [dictionary containing "phase" and "type" keys]`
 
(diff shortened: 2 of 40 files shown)


rwaldron:
@ossy-szeged I just want to make sure I understand this: we're adding a new enforcement here, not because anything will necessarily break otherwise, but for tidy and consistent meta data? I'm 100% for it, I just want to make sure I understand.

tools/lint/test/fixtures/test/negative_invalid_phase_type_order.js
```diff
@@ -0,0 +1,15 @@
...
+esid: sec-assignment-operators-static-semantics-early-errors
+description: Minimal test
+negative:
+  type: SyntaxError
+  phase: parse
+---*/
+
+$DONOTEVALUATE();
+
+!!!
```
leobalter, line 15: thanks for adding a test. As long as we have the documentation and this, I'm good.

One nit pick: you might want to update line 3 here with 2022 and your name. For the other side, the more the test looks like the other files and the diff is only within the order, the current text is better.

leobalter (approved):
I'd love to advocate for test runners to not rely on the order of yaml-based metadata, but the consistency followed by linter checks from this project's end seems fair enough.

rwaldron:
@ossy-szeged I have another question, which is just out of my own curiosity: what project (if any) does the current "unenforced" format have an impact on? Are you working with, or writing, a runner that has a yaml parser that encounters issues? (Again, this is just my curiosity)

tools/lint/test/fixtures/test/negative_invalid_phase_type_order.js
ossy-szeged: You're right, it is a copy of a similar test, that's why I didn't want to update the copyright. It's OK for me as is.

ossy-szeged:
@rwaldron Of course, the main goal of this PR is to have consistent meta data. I am working on JerryScript engine ( https://github.com/jerryscript-project/jerryscript ) and started to simplify our test runner. Now we uses a test runner based on the old test262-harness-py with monkeyYaml. The idea was to check if we can replace monkeyYaml with simple regexps to reduce external dependencies. It turned out that we can do it easily. But during the development, I found this and #3367 inconsistencies in meta data and I thought we can fix them quite easily, it's just a community contribution to test262.

PR #3371, opened 2022-01-06
Remove roundingIncrement: 1.1 case which should NOT throw RangeError

FrankYFTang:
roundingIncrement: 1.1 will not throw RangeError because
1.  roundingIncrement is read by calling GetNumberOption in step 21 of https://tc39.es/proposal-intl-numberformat-v3/out/numberformat/diff.html#sec-initializenumberformat
2. GetNumberOption https://tc39.es/ecma402/#sec-getnumberoption will call DefaultNumberOption https://tc39.es/ecma402/#sec-defaultnumberoption and 
3. DefaultNumberOption will call floor() in step 3. so roundingIncrement will be 1 if the option bag has {roundingIncrement: 1.1}

### DIFF
--- a/test/intl402/NumberFormat/constructor-roundingIncrement-invalid.js
+++ b/test/intl402/NumberFormat/constructor-roundingIncrement-invalid.js
@@ -11,10 +11,6 @@ assert.throws(RangeError, function() {
 }, '0');
 
 assert.throws(RangeError, function() {
-  new Intl.NumberFormat([], {roundingIncrement: 1.1});
-}, '1.1');
-
-assert.throws(RangeError, function() {
   new Intl.NumberFormat([], {roundingIncrement: 3});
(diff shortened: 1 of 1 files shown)


FrankYFTang:
@sffc @jugglinmike

rwaldron (approved):

sffc:
Good catch!

sffc:
@FrankYFTang Rather than deleting the test, would you consider changing the expectation of the test to match the spec?

FrankYFTang:
> @FrankYFTang Rather than deleting the test, would you consider changing the expectation of the test to match the spec?

It is already merged. also the test is inside test/intl402/NumberFormat/constructor-roundingIncrement-invalid.js  and the expectation is it won't throw so that should be part of other file, not in this file.

Issue #3365, opened 2022-01-04
Intl NumberFormat v3: add a test for `'1.01': '1'` (soon to be `1.0`) since trailing zero behavior is important.


rwaldron:
Nit: Here and elsewhere: add a test for `'1.01': '1'` (soon to be `1.0`) since trailing zero behavior is important.

_Originally posted by @sffc in https://github.com/tc39/test262/pull/3334#discussion_r774629784_

Issue #3364, opened 2022-01-04
Intl NumberFormat v3: Add tests for roundingPriority when all 4 settings are used at the same time (min+max fraction+significant)


rwaldron:
Suggestion: Add tests for roundingPriority when all 4 settings are used at the same time (min+max fraction+significant)

_Originally posted by @sffc in https://github.com/tc39/test262/pull/3334#discussion_r774638051_

romulocintra:
@rwaldron  I think we can close this one -> https://github.com/tc39/test262/pull/3405

ptomato:
I understood that another PR was still needed with tests for different values of `roundingPriority` (https://github.com/tc39/test262/pull/3405#pullrequestreview-907898302)

rwaldron:
@ptomato I'll let you drive this

romulocintra:
Related PR https://github.com/tc39/test262/pull/3372

PR #3372, opened 2022-01-06
set minimumFractionDigits: 1 in format-rounding-increment- test

FrankYFTang:
We need to set minimumFractionDigits: 1 for the case of {roundingIncrement: 5, maximumFractionDigits: 1} and also adjust the test expectation. 
Otherwise, inside testNumberFormat , it will call   getPatternParts to format 1.1 and -1.1 and cause pattern mismatch (because the result will be "1" instead of "1.1" in that configuration)

https://github.com/tc39/test262/blob/main/harness/testIntl.js#L2369

### DIFF
--- a/test/intl402/NumberFormat/prototype/format/format-rounding-increment-25.js
+++ b/test/intl402/NumberFormat/prototype/format/format-rounding-increment-25.js
@@ -15,7 +15,7 @@ var numberingSystems = ['arab', 'latn', 'thai', 'hanidec'];
 testNumberFormat(
   locales,
   numberingSystems,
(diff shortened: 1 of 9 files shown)


FrankYFTang:
@sffc @jugglinmike

romulocintra:
Hi @rwaldron I think this one closes #3365

FrankYFTang:
> Hi @rwaldron I think this one closes #3364

no, it does not. It is to fix the broken test, not to improve the test.

rwaldron (approved):

PR #3373, opened 2022-01-07
Fix test of intl402/NumberFormat/prototype/formatRangeToParts/en-US.js 

FrankYFTang:
1.  add includes: [propertyHelper.js]
2. "–" => " – "
3.  Move verifyProperty to the end of the loop because it has the side effect of removing the property while testing on writable: true


FrankYFTang:
@jugglinmike

romulocintra (approved):

rwaldron:
@romulocintra thanks for additional review! Awesome!

rwaldron (approved):

PR #3375, opened 2022-01-11
Intl.NumberFormat default useGrouping value is "auto"

Constellation:
After Intl.NumberFormat v3, default useGrouping is "auto".

### DIFF
--- a/test/intl402/NumberFormat/test-option-useGrouping.js
+++ b/test/intl402/NumberFormat/test-option-useGrouping.js
@@ -1,4 +1,5 @@
 // Copyright 2012 Mozilla Corporation. All rights reserved.
+// Copyright 2022 Apple Inc. All rights reserved.
 // This code is governed by the BSD license found in the LICENSE file.
 
 /*---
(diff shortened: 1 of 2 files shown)


test/intl402/NumberFormat/test-option-useGrouping.js
```diff
@@ -14,4 +15,22 @@ author: Norbert Lindenberg
 includes: [testIntl.js]
```
rwaldron, line 15: The linter wants us to remove this line now :)

rwaldron (changes requested):

test/intl402/NumberFormat/test-option-useGrouping.js
Constellation: OK! I'll do that :D

Constellation: Ah, looks like the error is happening in testIntl.js.

Constellation: OK, fixed. And I also added NumberFormat v3 features since this change happens with that proposal.

rwaldron (approved):

PR #3376, opened 2022-01-11
Add AggregateError as feature to seal-aggregateerror.js

lahma:
Test case `seal-aggregateerror.js` cannot work without `AggregateError` being implemented.


rwaldron (approved):
Nice catch!

PR #3378, opened 2022-01-18
Test DurationFormat

romulocintra:
Add initial tests to Intl.DurationFormat proposal #3305 

Proposal: https://github.com/tc39/proposal-intl-duration-format

PR #3385, opened 2022-01-25
Update DurationFormat tests

romulocintra:
Summary:

- Amend missing feature fields in test description 
- Update code style to use `verifyProperty` instead of method for each `Writable` , `Configurable` , `Enumerable`
- Also solves comments https://github.com/tc39/test262/pull/3378#discussion_r791500096 and https://github.com/tc39/test262/pull/3378#discussion_r791499618

### DIFF
--- a/test/intl402/DurationFormat/instance/length.js
+++ b/test/intl402/DurationFormat/instance/length.js
@@ -25,8 +25,6 @@ features: [Intl.DurationFormat]
 includes: [propertyHelper.js]
 ---*/
 
-assert.sameValue(Intl.DurationFormat.length, 0);
-
 verifyProperty(Intl.DurationFormat, 'length', {
   value: 0,
   enumerable: false,
--- a/test/intl402/DurationFormat/instance/length.js
+++ b/test/intl402/DurationFormat/instance/length.js
@@ -27,6 +27,8 @@ includes: [propertyHelper.js]
 
 assert.sameValue(Intl.DurationFormat.length, 0);
 
-verifyNotEnumerable(Intl.DurationFormat, "length");
-verifyNotWritable(Intl.DurationFormat, "length");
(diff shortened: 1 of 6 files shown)


romulocintra:
cc @iamstolis

iamstolis:
> cc @iamstolis looks like it's failing on GraalJS

I have no idea what these jobs are supposed to do but it seems that `GraalJS` job is failing because `graaljs` engine provided by `esvu` is broken. I guess that the recent changes in GraalVM launchers are to blame. `js` binary downloaded by `esvu` is not standalone anymore (it requires `jsvm` dynamic library now). I will create PR against `esvu` with a fix, but this should not block this PR.

test/intl402/DurationFormat/instance/length.js
```diff
@@ -27,6 +27,8 @@ includes: [propertyHelper.js]
 
 assert.sameValue(Intl.DurationFormat.length, 0);
 
-verifyNotEnumerable(Intl.DurationFormat, "length");
-verifyNotWritable(Intl.DurationFormat, "length");
-verifyConfigurable(Intl.DurationFormat, "length");
+verifyProperty(Intl.DurationFormat, 'length', {
+  enumerable: false,
```
Ms2ger, line 31: Maybe:
```suggestion
verifyProperty(Intl.DurationFormat, 'length', {
  value: 0,
  enumerable: false,
```

romulocintra: Thanks , @Ms2ger also deleted the previous assertion for value

rwaldron:
@romulocintra there is actually no requirement for the CI tests in engines to pass—in fact, we generally anticipate that most should initially fail (depending on how new the feature is). The only things that matter are the two tools and linting tasks 👍

rwaldron (approved):

PR #3394, opened 2022-02-03
Test WrappedFunction properties

legendecas:
Add test coverages on [CopyNameAndLength](https://tc39.es/proposal-shadowrealm/#sec-copynameandlength) in [WrappedFunctionCreate](https://tc39.es/proposal-shadowrealm/#sec-wrappedfunctioncreate).

### DIFF
--- a/test/built-ins/ShadowRealm/WrappedFunction/length-throws-typeerror.js
+++ b/test/built-ins/ShadowRealm/WrappedFunction/length-throws-typeerror.js
@@ -1,0 +1,36 @@
+// Copyright (C) 2022 Chengzhong Wu. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-wrappedfunctioncreate
+description: >
+  WrappedFunctionCreate throws a TypeError from its caller realm.
+info: |
+  WrappedFunctionCreate ( callerRealm: a Realm Record, Target: a function object, )
+
+  ...
+  7. Let result be CopyNameAndLength(wrapped, Target).
+  8. If result is an Abrupt Completion, throw a TypeError exception.
+  ...
+
+features: [ShadowRealm]
+---*/
+
+assert.sameValue(
+  typeof ShadowRealm.prototype.evaluate,
+  'function',
+  'This test must fail if ShadowRealm.prototype.evaluate is not a function'
+);
+
+const r = new ShadowRealm();
+
+assert.throws(TypeError, r.evaluate(`
+function fn() {}
+Object.defineProperty(fn, 'length', {
+  get: () => {
+    throw new Error('blah');
+  },
+  enumerable: false,
+  configurable: true,
+});
+fn;
+`), 'expect a TypeError on length getter throwing');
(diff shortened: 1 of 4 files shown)


leobalter (approved):

test/built-ins/ShadowRealm/WrappedFunction/length-throws-typeerror.js
```diff
@@ -0,0 +1,36 @@
...
+function fn() {}
+Object.defineProperty(fn, 'length', {
+  get: () => {
+    throw new Error('blah');
+  },
+  enumerable: false,
+  configurable: true,
+});
+fn;
+`), 'expect a TypeError on length getter throwing');
```
rwaldron, line 36: I think this should be: 

```js
  r.evaluate(`
    function fn() {}
    Object.defineProperty(fn, 'length', {
      get: () => {
        throw new Error('blah');
      },
      enumerable: false,
      configurable: true,
    });
    fn;
  `)
}, 'expect a TypeError on length getter throwing');
```

And same for the test further down

rwaldron (changes requested):

test/built-ins/ShadowRealm/WrappedFunction/length-throws-typeerror.js
rwaldron: Oh wait... NO, I'm wrong. This is awesome.

rwaldron (approved):

test/built-ins/ShadowRealm/WrappedFunction/length-throws-typeerror.js
leobalter: Yes! This error happens by the time of the evaluation and somehow I got it as throwing when calling the fn.

leobalter: @rwaldron I believe your initial review is actually correct.

```
CopyNameAndLength ( F: a function object, Target: a function object, optional prefix: a String, optional argCount: a Number, )
1. If argCount is undefined, then set argCount to 0.
2. Let L be 0.
3. Let targetHasLength be ? HasOwnProperty(Target, "length").
4. If targetHasLength is true, then
  a. Let targetLen be ? Get(Target, "length").
``

The abrupt completion happens during the evaluation time when creating a wrapped function for `fn` and running Get(fn, "length")

In this test, `r.evaluate()` is the throwing factor. This also leads to some missing coverage:

- tests for the `HasOwnProperty(length)` part (e.g. callable proxies), including abrupt completion.
- tests for `length` in the prototype chain (mixed with the hasOwn internal)
- resolution of name/length for values evaluated from other wrapped functions (e.g. a return from function, or sending values back to the shadowRealm).

File: test/built-ins/ShadowRealm/WrappedFunction/name.js [hunk 2]
```diff
@@ -0,0 +1,80 @@
...
+  configurable: true,
+});
+fn;
+`);
+verifyProperty(wrapped, "name", {
+  value: "",
+  enumerable: false,
+  writable: false,
+  configurable: true,
+});
```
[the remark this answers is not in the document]
legendecas: Submitted https://github.com/tc39/test262/pull/3396 :)

test/built-ins/ShadowRealm/WrappedFunction/length-throws-typeerror.js
legendecas: Yeah, @rwaldron is right. Submitted https://github.com/tc39/test262/pull/3396 to fix this.

PR #3400, opened 2022-02-06
Test that `Array.prototype` methods throw on attempt to `[[Set]]` read-only `"length"`

shvaikalesh:
JSC bug: [Attempting to `[[Set]]` `JSArray`'s read-only `"length"` should throw even with current `[[Value]]`](https://bugs.webkit.org/show_bug.cgi?id=221177).


rwaldron (approved):
Spectacular

PR #3403, opened 2022-02-09
Check a variety of offset Etc/GMT timezones

jessealama:
It looks like the reference polyfill already does what we expect: these tests all pass. If we work on cleaning up the grammar for the various Etc/GMT timezones [as suggested](https://github.com/tc39/proposal-temporal/issues/1993), these tests might be handy.

### DIFF
--- a/test/intl402/Temporal/TimeZone/etc-timezone.js
+++ b/test/intl402/Temporal/TimeZone/etc-timezone.js
@@ -25,11 +25,11 @@ assert.sameValue(
   );
 });
 
-let gmtMinus15TZ = "Etc/GMT-15";
+let gmtMinus24TZ = "Etc/GMT-24";
 assert.throws(
   RangeError,
-  () => { new Temporal.TimeZone(gmtMinus15TZ); },
-  gmtMinus15TZ + " is an invalid timezone"
+  () => { new Temporal.TimeZone(gmtMinus24TZ); },
+  gmtMinus24TZ + " is an invalid timezone"
 );
 
 // "Etc/GMT-0N" is not OK (1 ≤ N ≤ 9)
--- a/test/intl402/Temporal/TimeZone/etc-timezone.js
+++ b/test/intl402/Temporal/TimeZone/etc-timezone.js
@@ -10,14 +10,14 @@ features: [Temporal]
 // "Etc/GMT-0" through "Etc/GMT-14" are OK
 
 assert.sameValue(
-  Temporal.TimeZone.from("Etc/GMT-0").toString(),
+  (new Temporal.TimeZone("Etc/GMT-0")).toString(),
   "UTC", // if the offset is -0, we say "UTC" rather than "GMT"
   "Etc/GMT-0 is a valid timezone"
 );
 
 [1,2,3,4,5,6,7,8,9,10,11,12,13,14].forEach((n) => {
   let tz = "Etc/GMT-" + n;
-  let instance = Temporal.TimeZone.from(tz);
+  let instance = new Temporal.TimeZone(tz);
   assert.sameValue(
     instance.toString(),
     tz,
@@ -25,12 +25,19 @@ assert.sameValue(
   );
 });
 
+let gmtMinus15TZ = "Etc/GMT-15";
+assert.throws(
+  RangeError,
+  () => { new Temporal.TimeZone(gmtMinus15TZ); },
+  gmtMinus15TZ + " is an invalid timezone"
+);
+
 // "Etc/GMT-0N" is not OK (1 ≤ N ≤ 9)
 [1,2,3,4,5,6,7,8,9].forEach((n) => {
   let tz = "Etc/GMT-0" + n;
   assert.throws(
     RangeError,
-    () => { Temporal.TimeZone.from(tz); },
+    () => { new Temporal.TimeZone(tz); },
     tz + " is an invalid timezone"
   );
 });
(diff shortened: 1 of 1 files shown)


jessealama:
cc @ptomato @Ms2ger @gibson042 (sorry I need to mention you here in a comment-- it seems I can't add you as a reviewer in the normal way)

linusg:
> Where do they belong? (Is this even the right repo?)

Yes, here: https://github.com/tc39/test262/tree/main/test/intl402/Temporal

This is simply so that engines/builds without Intl can chose to ignore `test/intl402/` to not get any unexpected build failures simply because they're missing non-`UTC` time zones.

test/intl402/Temporal/ZonedDateTime/etc-timezone.js
```diff
@@ -0,0 +1,61 @@
...
+  let tz = "Etc/GMT-" + n;
+  let instance = Temporal.ZonedDateTime.from("2022-02-07T12:34[" + tz + "]");
+  assert.sameValue(
+    instance.timeZone.toString(),
+    tz,
+    tz + " is a valid timezone"
+  );
+});
+
+// "Etc/GMT+0N" is not OK (0 ≤ N ≤ 9)
```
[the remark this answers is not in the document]
jessealama: good idea, done

I also added tests for `Etc/GMT-02`, etc., up to `Etc/GMT-09`.

test/intl402/Temporal/ZonedDateTime/etc-timezone.js
```diff
@@ -0,0 +1,61 @@
...
+    instance.timeZone.toString(),
+    tz,
+    tz + " is a valid timezone"
+  );
+});
+
+assert.throws(
+  RangeError,
+  () => { Temporal.ZonedDateTime.from("2022-02-07T12:34[Etc/GMT+10])"); },
+  "Etc/GMT+10 is an invalid timezone"
```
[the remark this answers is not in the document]
jessealama: good idea, done

test/intl402/Temporal/ZonedDateTime/etc-timezone.js
```diff
@@ -0,0 +1,61 @@
...
+/*---
+esid: sec-temporal.zoneddatetime
+description: Some Etc/GMT{+/-}{0}N timezones are valid, but not all
+features: [Temporal]
+---*/
+
+// "Etc/GMT-0" through "Etc/GMT-14" are OK
+
+assert.sameValue(
+  Temporal.ZonedDateTime.from("2022-02-07T12:34[Etc/GMT-0]").timeZone.toString(),
```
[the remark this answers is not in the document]
jessealama: good idea, done (and the test file was moved from the `ZonedDateTime` directory to `TimeZone`)

ptomato:
I have a few other suggestions that didn't occur to me last time I looked at it:
- The test is now in the `Temporal/TimeZone/` folder but uses `Temporal.TimeZone.from()`, we should consider using `new Temporal.TimeZone(...)` instead
- It might also be good to test that "one past the end" throws, so `Etc/GMT-15` and `Etc/GMT+13`

jessealama:
>     * The test is now in the `Temporal/TimeZone/` folder but uses `Temporal.TimeZone.from()`, we should consider using `new Temporal.TimeZone(...)` instead

done

>     * It might also be good to test that "one past the end" throws, so `Etc/GMT-15` and `Etc/GMT+13`

done

ptomato (approved):
I think the content of this test is all good now from the Temporal side of things, assuming the normative change achieves consensus. Thanks!

There are some coding style consistency issues that I'd normally nitpick, but I don't know how much we care about that in test262.

test/intl402/Temporal/TimeZone/etc-timezone.js
```diff
@@ -10,27 +10,34 @@ features: [Temporal]
 // "Etc/GMT-0" through "Etc/GMT-14" are OK
 
 assert.sameValue(
-  Temporal.TimeZone.from("Etc/GMT-0").toString(),
+  (new Temporal.TimeZone("Etc/GMT-0")).toString(),
```
gibson042, line 13: What motivates `new Temporal.TimeZone(…)` vs. `Temporal.TimeZone.from(…)`?

test/intl402/Temporal/TimeZone/etc-timezone.js
```diff
@@ -10,27 +10,34 @@ features: [Temporal]
...
     tz + " is a valid timezone"
   );
 });
 
+let gmtMinus15TZ = "Etc/GMT-15";
+assert.throws(
+  RangeError,
+  () => { new Temporal.TimeZone(gmtMinus15TZ); },
+  gmtMinus15TZ + " is an invalid timezone"
+);
```
gibson042, line 33: This test will fail if IANA ever add Etc/GMT-15.
```suggestion
let gmtMinus24TZ = "Etc/GMT-24";
assert.throws(
  RangeError,
  () => { new Temporal.TimeZone(gmtMinus24TZ); },
  gmtMinus24TZ + " is an invalid timezone"
);
```

test/intl402/Temporal/TimeZone/etc-timezone.js
```diff
@@ -40,26 +47,35 @@ assert.sameValue(
...
-// zero is handled in its own way:
+// zero is handled in its own way (say "UTC" rather than "GMT"):
 assert.sameValue(
-  Temporal.TimeZone.from("Etc/GMT+0").toString(),
+  (new Temporal.TimeZone("Etc/GMT+0")).toString(),
   "UTC", // if the offset is +0, we say "UTC" rather than "GMT"
   "Etc/GMT+0 is a valid timezone"
 );
 
 [1,2,3,4,5,6,7,8,9, 10, 11, 12].forEach((n) => {
```
gibson042, line 64: Can we correct the variable spacing around commas here?

test/intl402/Temporal/TimeZone/etc-timezone.js
```diff
@@ -40,26 +47,35 @@ assert.sameValue(
...
 });
+
+// "Etc/GMT+13" is not OK
+
+let gmtPlus13TZ = "Etc/GMT+13";
+assert.throws(
+  RangeError,
+  () => { new Temporal.TimeZone(gmtPlus13TZ); },
+  gmtPlus13TZ + " is an invalid timezone"
+);
```
gibson042, line 81: This test will fail if IANA ever add Etc/GMT+13.

```suggestion
let gmtPlus24TZ = "Etc/GMT+24";
assert.throws(
  RangeError,
  () => { new Temporal.TimeZone(gmtPlus24TZ); },
  gmtPlus24TZ + " is an invalid timezone"
);
```

gibson042 (changes requested):

test/intl402/Temporal/TimeZone/etc-timezone.js
ptomato: That was my suggestion, I'd prefer to use the constructor in the tests unless we are specifically testing functionality of from(), so that the results aren't affected if there is some other bug in from().

test/intl402/Temporal/TimeZone/etc-timezone.js
ptomato: Another of my previous review suggestions :sweat_smile: It didn't occur to me that IANA could add it in the future. I think checking the "out of bounds" case by using 24 is a good idea.

test/intl402/Temporal/TimeZone/etc-timezone.js
jessealama: done

gibson042 (approved):

ptomato (approved):

ptomato:
https://github.com/tc39/proposal-temporal/pull/2050 achieved consensus, so I'll merge this.

PR #3405, opened 2022-02-10
use all values max and min fraction and significant digits

romulocintra:
Summary : 

WIP updating https://github.com/tc39/test262/issues/3364 

Will be draft PR depending on: https://github.com/tc39/proposal-intl-numberformat-v3/issues/81 that could also affect the tests

### DIFF
--- a/test/intl402/NumberFormat/prototype/format/format-max-min-fraction-significant-digits.js
+++ b/test/intl402/NumberFormat/prototype/format/format-max-min-fraction-significant-digits.js
@@ -1,0 +1,27 @@
+// Copyright (C) 2022 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-intl.numberformat.prototype.format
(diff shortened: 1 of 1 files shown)


rwaldron:
LGTM, ping when out of draft mode

romulocintra:
> LGTM, ping when out of draft mode

Thanks for having a look at it, will leave it like this and do incremental coverage as it covers  #3364

sffc (approved):

ptomato (approved):
This test seems correct according to https://tc39.es/proposal-intl-numberformat-v3/out/numberformat/proposed.html#sec-setnfdigitoptions

According to #3364 the idea was to add tests for different values of `roundingPriority`. Is that still planned to be added in another PR?

romulocintra:
> This test seems correct according to https://tc39.es/proposal-intl-numberformat-v3/out/numberformat/proposed.html#sec-setnfdigitoptions
> 
> 
> 
> According to #3364 the idea was to add tests for different values of `roundingPriority`. Is that still planned to be added in another PR?

Yes that's the idea add those in a new PR

ptomato:
OK, thanks!

PR #3406, opened 2022-02-10
Rough outlines for new documentation/guides

rwaldron:
@sarahghp

### DIFF
--- a/guides/Getting-Started.md
+++ b/guides/Getting-Started.md
@@ -1,0 +1,19 @@
+# Getting Started With Test262
+
+
+Rough outline...
+
+
+- Why do we test?
+- What do we test? 
+  - Similar to https://github.com/tc39/test262/blob/main/CONTRIBUTING.md#acceptable-tests and https://github.com/tc39/test262/blob/main/CONTRIBUTING.md#unacceptable-tests
+- Anatomy of a test
+  - frontmatter
+    - esid
+    - description
+    - don't mention any other frontmatter
+  - test code
+    - assert api
+- Write a test
+- Run your test (links to Running-Tests.md)
+
--- a/guides/Running-Tests.md
+++ b/guides/Running-Tests.md
@@ -1,0 +1,14 @@
+# Running Tests
+
+- Explain Test262-Harness
+  - Explain role of eshost
+  - Explain role of esvu
+- How to set up Test262-Harness
+  - Borrow from https://github.com/bterlson/test262-harness#getting-started
+- Provide a bunch of common and useful examples of test262-harness commands, with explanations of what is happening and what to expect
+  - Speeding up with threads
+  - Specifying engines-specific host args
+  - filtering tests with `--features`
+  - Polyfill injection with `--prelude`
+  - Code tranformation intermediary with `--transformer`	
+  - Code preprocessing and outcome overide with `--preprocessor`


guides/Running-Tests.md
```diff
@@ -0,0 +1,14 @@
+# Running Tests
```
sarahghp, line 1: For my Temporal tests, I actually ran via the node script. Do you think it would be helpful to mention this route or do you think it's just a recipe for more misery?

guides/Getting-Started.md
```diff
@@ -0,0 +1,77 @@
...
+      - Does it need a new feature tag?
+    - Share test plan
+  - Writing our first test
+    - Git stuff
+      - Pull test272 repo
+      - Create a branch
+      - What else?
+    - Create a new file
+      - Where? Why?
+      - What do I name it? Why?
```
sarahghp, line 33: Should we add the idea of how many tests per file here? Or maybe above in the create test plan section, something about "how to decompose behavior into files"?

sarahghp (commented):
This looks really good, @rwaldron! I just have two suggestions.

guides/Getting-Started.md
```diff
@@ -0,0 +1,77 @@
...
+- Run our test
+  - Links to Running-Tests.md
+  - Shows minimal run example, ie. "If we've already read Running-Tests.md and set up a test running environment, go ahead and execute this command..."
+- What else could be tested?
+- Commit code
+- Push branch
+- Open PR
+- Ask for review
+  - Who?
+    - We need a list of people that will likely respond
```
ptomato, line 75: Remembering to update this list would be an extra hoop to jump through when somebody joins or leaves the maintainers group, how about a team in the TC39 GitHub org so you can do `@tc39/test262-maintainers`?

guides/Running-Tests.md
```diff
@@ -0,0 +1,14 @@
...
+
+- Explain Test262-Harness
```
ptomato, line 3: I'd suggest splitting up this doc into a first half that answers "I just wanna run my tests, how do I make it work?" and a second half with everything else. (For example, for #3414, I just did `npx test262-harness "test/language/expressions/logical-assignment/left-hand-side-private*"` and that was all.)

Although now that I think about it, probably the explanation of how to try more than one host belongs in the first half as well. Contributors might not think about the possibility there could be a conformance bug in v8 if they just run on Node.js by default.

guides/Running-Tests.md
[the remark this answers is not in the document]
sarahghp: I think I disagree slightly, @jugglinmike. Part of writing a test is being able to run it and so it should be included in the tutorial, at least in a lightweight way. Given that, I think it might be appropriate to write a short explanation about how to run it, perhaps linking to the `test262-harness` for more details, but covering enough that test writers aren't expected to learn a whole second project or left scratching their heads. I think that's a balance we could achieve. What do you think?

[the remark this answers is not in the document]
jugglinmike: Agreed--running the tests is definitely central to the contribution experience. It's just that Test262 intentionally doesn't offer that service, and I'm not convinced it has a perspective or audience that warrants original content on the topic. The kind guide that @rwaldron has outlined would be just as valuable (and even more relevant) to folks who arrive from the test262-harness repository.

I'd be more comfortable with something minimal--a command to install it, a command to run it, and a link to a version of this guide provided by test262-harness.

rwaldron: I'm actually not sure what node script you're referring to

guides/Getting-Started.md
```diff
@@ -0,0 +1,77 @@
...
+        - Write a checklist (we'll give a 2-3 point example to work from)
+      - Does it need a new feature tag?
+    - Share test plan
+  - Writing our first test
+    - Git stuff
+      - Pull test272 repo
+      - Create a branch
+      - What else?
+    - Create a new file
+      - Where? Why?
```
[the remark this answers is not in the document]
rwaldron: As far as I can tell, unless I missed something somewhere, there are only three places where that `Foo/constructor/` path pattern occurred: 

- test/intl402/ListFormat/constructor
- test/intl402/RelativeTimeFormat/constructor
- test/intl402/Segmenter/constructor

...I've had a "TODO" in some eng. journal of mine (somewhere) to fix them. I think I merged them and wrote myself the note and never followed up. The reason there isn't a `built-ins/X/constructor` is because `X.constructor` is `Function` and those tests are in `built-ins/Function`. One might argue that constructor-specific tests maybe belong in `built-ins/X/prototype/constructor`, however I would counter with: `X.prototype.constructor === X`, so `built-ins/X` is semantically the same and much less to type.

guides/Getting-Started.md
rwaldron: Sure 🤷

guides/Running-Tests.md
rwaldron: We could have a "TL;DR" that's basically like... 

> If you want to quickly verify that your tests can be run successfully, then the fastest and easiest way to achieve that is to run `npx test262-harness "test/path/to/your/tests"`. If you want to run your tests in different engines, to check for implementation conformance, read on...

guides/Running-Tests.md
rwaldron: Maybe the `npm run test` thing? I'm sorry, no one should have to sit through running the entire test suite with a single thread pool :| The `test` script was originally used for CI, but then we made one called `ci` and `test` became a generic "run all the tests" command (https://github.com/tc39/test262/pull/2137/).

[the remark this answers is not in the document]
rwaldron: I'll gladly put this in Test262-Harness, but we could keep short blurb described [here](https://github.com/tc39/test262/pull/3406/files#r823011869)

guides/Running-Tests.md
ptomato: Slightly related to this topic, in https://github.com/tc39/test262/pull/3430#discussion_r828247024 I realized it's not obvious that a green CI job doesn't mean the tests in your PR passed on that host... is that something that we should note here?

guides/Getting-Started.md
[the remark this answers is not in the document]
ptomato: Yeah, I didn't mean I was super bothered by the discrepancy of `constructor/`, but maybe we could solve the larger question by having a "Where to put my test?" guide document. I could try to throw an outline together during this work session time.

guides/Running-Tests.md
sarahghp: I was referring to what seems to be an invocation of the CI script from [the Temporal Polyfill package.json](https://github.com/tc39/proposal-temporal/blob/main/polyfill/package.json#L14).

It seems to me that it would be desirable to have a script that can accept file glob arguments and run from the command line, as a low-key way to run specs one is working on and as a way to let writers, especially new ones, see run new or changed tests without shunting them into another project. I thought it already existed, but I was just lucky!

(I agree that not _only_ making it possible to run the entire suite is good, but if people want to do so, by not filtering it, who am I to stop them?)

sarahghp: Ah, I see this is [covered below](https://github.com/tc39/test262/pull/3406#discussion_r809541417) — glad we all agree about running tests lo-key.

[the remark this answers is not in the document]
sarahghp: I would like to keep the blurb and whatever it takes for the experience not to be "Oh ho, you wanted to RUN the tests? well that's a WHOLE different kettle of fish." I think that's putting idealism over a welcoming experience. But luckily I think we've arrived at a resonable decision.

Issue #3413, opened 2022-02-16
Incorrect new private reference methods and accessors with logical assignment


davidot:
In 1b1097dbf6e93fb17b5a8156736563370ad4700e some tests for compound assignment of private fields and methods were added.

However I believe there are a couple of incorrect tests:
For example: `test/language/expressions/compound-assignment/left-hand-side-private-reference-method-nullish.js`
```js
class C {
  #privateMethod() {}
  compoundAssignment() {
    return this.#privateMethod ??= 1;
  }
}

const o = new C();
assert.throws(TypeError, () => o.compoundAssignment(), "PutValue throws when storing the result in a method private reference");
```

Looking at: https://tc39.es/ecma262/#sec-assignment-operators-runtime-semantics-evaluation
`||=`, `&&=` and `??=` are all special cases which short circuit when the left hand side passes a condition (like ToBoolean being `true` for `||=`)

So should I would think that it should not reach the `Putvalue(lref, rval)` call and thus does not throw in these instances.

Affected tests I found which fail this.
```
test/language/expressions/compound-assignment/left-hand-side-private-reference-method-nullish.js
test/language/expressions/compound-assignment/left-hand-side-private-reference-method-or.js
test/language/expressions/compound-assignment/left-hand-side-private-reference-readonly-accessor-property-nullish.js
test/language/expressions/compound-assignment/left-hand-side-private-reference-readonly-accessor-property-or.js
```
Since `&&=` does call `PutValue` if the left hand side is true which the methods evaluate to the following tests pass but might not be quite tested what was intended.
```
test/language/expressions/compound-assignment/left-hand-side-private-reference-method-and.js
test/language/expressions/compound-assignment/left-hand-side-private-reference-readonly-accessor-property-and.js
```

davidot:
These tests might also need to be moved as these are the only tests with `||=`, `&&=` and `??=` in the `compound-assignment` folder instead of the rest of those tests in `logical-assignment`.

linusg:
cc @ptomato

ptomato:
Thanks for the feedback. Here's what I propose to do:
- the read-only accessor property tests for `||=` and `??=` can be fixed by returning a falsey or nullish value, respectively, from the getter
- the private method tests for `||=` and `??=` are invalid and should be removed
- I'll move the `||=`, `??=`, and `&&=` tests to `test/language/expressions/logical-assigment` (thanks! I didn't spot that folder at first)

[...]

Issue #2940, opened 2021-01-27
Missing coverage: compound assignment w/ private accessor properties


marjakh:
Per this V8 bug which was not caught by test262: https://bugs.chromium.org/p/v8/issues/detail?id=11360

Added V8 tests which could be repurposed as test262 tests:
https://chromium-review.googlesource.com/c/v8/v8/+/2652495/1/test/mjsunit/regress/regress-v8-11360.js

ptomato:
I'd like to start working on this.

PR #3414, opened 2022-02-17
Fix tests for private reference with logical assignment

ptomato:
There were three things wrong with the 'and', 'or', and 'nullish' tests
that I added as part of #2940:

1. They were in the wrong folder (should be
   expressions/logical-assignment, not expressions/compound-assignment)
2. The tests for ||= and ??= on readonly accessor properties were
   incorrect. These assignments would short-circuit if the getter
   returned 1 as it previously did, so PutValue would not throw.
3. The tests for ||= and ??= on private methods were invalid, as a
   method always evaluates to true in a boolean context, and is not
   nullish, so these would always short-circuit.

I've removed the invalid private method cases, fixed the readonly
accessor cases, and added new templates to test the short-circuit
behaviour as well as the non-short-circuit behaviour.

Closes: #3413


ptomato:
cc @davidot

src/logical-assignment-private/nullish.case
```diff
@@ -0,0 +1,34 @@
+// Copyright 2022 Igalia S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+templates:
```
ptomato, line 5: This seems like it's a fairly brittle way to express what I wanted to achieve, which is "use all the templates in `default/` except for `method.template`. I didn't see a better way to do this, but let me know if there is.

davidot (commented):
Thank you for fixing the test!

The fixed tests now seem correct to me.
However the new tests could be a little more thorough.

test/language/expressions/logical-assignment/left-hand-side-private-reference-data-property-short-circuit-and.js
```diff
@@ -0,0 +1,52 @@
...
+  compoundAssignment() {
+    return this.#field &&= false;
+  }
+  fieldValue() {
+    return this.#field;
+  }
+}
+
+const o = new C();
+assert.sameValue(o.compoundAssignment(), false, "The expression should evaluate to the short-circuit value");
```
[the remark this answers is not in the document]
ptomato: I like this idea! I've done so in the new version.

davidot (commented):
Look great now! (I would approve but I don't know if that messes up the PR workflow)
Whoops I found one last problem

test/language/expressions/logical-assignment/left-hand-side-private-reference-method-short-circuit-and.js
```diff
@@ -0,0 +1,56 @@
...
+
+
+function doNotCall() {
+  throw new Test262Error("The right-hand side should not be evaluated");
+}
+
+class C {
+  #privateMethod() {}
+  compoundAssignment() {
+    return this.#privateMethod &&= doNotCall();
```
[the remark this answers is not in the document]
ptomato: Oops, I forgot to skip this one. `&&=` on a private method can't ever short-circuit.

davidot (approved):
I can't find any problems anymore, thanks again for the fixes (and new tests)

src/logical-assignment-private/nullish.case
rwaldron: Did you try using a glob expression? `template: expr`?

ptomato: I did look at that, but crafting a glob expression that signified "`default/*` except for one file in there" seemed worse.

PR #3418, opened 2022-02-22
Check `formatRange` on pairs of invalid arguments

jessealama:
Here are some tests to go along with a [recently discovered issue](https://github.com/tc39/proposal-temporal/pull/2043) in proposal-temporal.

All these tests pass, showing that the polyfill as it currently stands already does the right thing.


test/intl402/Temporal/Instant/formatrange-fails.js
```diff
@@ -0,0 +1,32 @@
+// Copyright (C) 2022 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.instant
```
ptomato, line 5: This should be a different ID, since we are testing functionality from Intl.DateTimeFormat.prototype.formatRange. (Although I'm not sure if it should be the ID from https://tc39.es/ecma402/#sec-intl.datetimeformat.prototype.formatRange where the functionality will eventually be described, or it should have a dummy ID for now)

The test file should also be moved to `test/intl402/Intl/DateTimeFormat/prototype/formatRange/` and it should have a more descriptive name like `arguments-different-types.js`.

ptomato (commented):
I think we should have a similar test for Intl.DateTimeFormat.prototype.formatRangeToParts as well.

test/intl402/Temporal/Instant/formatrange-fails.js
```diff
@@ -0,0 +1,32 @@
...
+description: formatRange fails if given arguments of different types
+features: [Temporal]
+---*/
+
+const us = new Intl.DateTimeFormat('en-US');
+
+const instances = {
+  instant: new Temporal.Instant(0n),
```
ptomato, line 13: Oh, I'd also add something like `number: Date.UTC(2000, 5, 2)` to test the combination of a Temporal object with any other value.

jessealama: good idea -- done

test/intl402/Temporal/Instant/formatrange-fails.js
jessealama: done (I settled on `sec-intl.datetimeformat.prototype.formatRange` for the ID, though if that needs to be changed, we can change it).

rwaldron (approved):

PR #3410, opened 2022-02-12
Compound assignment to private references

ptomato:
This adds generated tests for each of the compound assignment operators, assigning to each type of private reference. I used https://chromium-review.googlesource.com/c/v8/v8/+/2652495/1/test/mjsunit/regress/regress-v8-11360.js as a starting point for the tests, and expanded that to cover all the operators.

Also two minor fixes to .editorconfig and CONTRIBUTING.md.

Closes: #2940

PR #3423, opened 2022-03-03
Document rationale for some maintenance practices

jugglinmike:


docs/rationale.md
```diff
@@ -0,0 +1,144 @@
...
+expressive enough to model the semantics of a rich and evolving programming
+language like ECMAScript. This means the common and crucial task of coverage
+assessment will likely always be a challenge, but strong conventions around
+file organization can help.
+
+Tests for syntax-derived operations are organized according to the language
+grammar, with directories used to describe non-terminals. Tests for built-in
+APIs are organized according to the identifiers by which they can be accessed,
+with directories used to describe the sequence of properties that can be used
+from the global scope.
```
ptomato, line 60: I'd love to have this expanded so that it fully answers the question "Where do I put my test?" as in https://github.com/tc39/test262/pull/3406#discussion_r829390527
- For "the language grammar", including an example of the `language/statements/` directory with a link to https://tc39.es/ecma262/#sec-ecmascript-language-statements-and-declarations would help illustrate the correspondence that Rick showed in that other thread.
- The part about built-in identifiers could use an example for clarity
- We should say something about `intl402/` vs `built-ins/`, I'd guess it wouldn't be obvious to some newcomers that some hosts don't have Intl and toLocaleString
- For completeness we could mention `annexB/` and `harness/`

docs/rationale.md
```diff
@@ -0,0 +1,144 @@
...
+## Regression tests
+
+It is possible to write tests for semantics which, while not explicitly
+specified by ECMA262, are nonetheless valid according to the normative text.
+Such tests are welcome in Test262, but their fitness is not a given. Consumers
+from many constituencies value the coherence and consistency of the test suite,
+and tests which disallow arbitrary extraneous behavior can degrade those
+qualities. Because Test262 is not maintained as a repository of regression
+tests, contributions which include these kinds of tests will be weighed against
+their likelihood of identifying error in a plurality of implementations.
```
ptomato, line 80: I think it'd benefit all engines if we could get implementations to contribute their regression tests in a shareable format, maybe outside the core suite, even if they were unlikely to identify errors in other implementations. That doesn't reflect current practice, so I don't think this text needs to change, but is this something you'd be open to discussing in the future?

docs/rationale.md
```diff
@@ -0,0 +1,144 @@
...
+When testing a syntactic feature of the language, it can be tempting to write
+tests which verify that some bit of source text does *not* produce a syntax
+error. Contributors should try to push beyond verifying only the lack of a
+syntax error because almost all such tests also have observable semantics. It's
+often better for a test to assert that the expected semantics are followed,
+even when they may already be covered elsewhere.
+
+However, this is not always desirable because verifying semantics invariably
+requires inserting still more code, and that additional code may degrade the
+tests' precision for verifying syntax.
```
ptomato, line 110: I'm not sure I completely understand this text and I think it'd benefit from an example.

docs/rationale.md
```diff
@@ -0,0 +1,144 @@
...
+- it discourages contributors by requiring them to learn more
+- it frustrates implementers by making it harder to understand what's being
+  tested and what has failed
+
+One of abstraction's common motivations is its tendency to reduce maintenance
+costs by limiting duplication. TC39 has a very high standard for compatibility
+between revisions of ECMA262. This gives us a certain assurance in Test262 that
+maintainers of other test suites do not enjoy: Test262's tests are very rarely
+invalidated. The project takes advantage of this by using a more declarative,
+readable, and verbose style.
```
ptomato, line 138: I'm OK with setting a high bar for abstractions for the reasons you give here, and it is also current practice. But I don't agree philosophically with some of the drawbacks (like abstractions being less readable), or I feel that the corresponding advantages aren't really mentioned. So I'm not sure I'd feel comfortable about enshrining that as policy. I do agree for sure that there is a line to be drawn, so maybe we could phrase this differently?

E.g., I'd be in favour of removing these two paragraphs, and having the last paragraph say something like "Abstraction has other motivations, so there will always be room for it to some extent. We expect in the future to have more guidance about when it is appropriate and when it isn't. When the benefits of a specific proposal ..."

For example, I feel that the part about test262 tests rarely being invalidated holds very well for testing things that are already part of the standard, but not for developing tests for pre-Stage-3 proposals, which was my main exposure to test262 for a long time. So that may help illustrate why I'm approaching this from a different perspective.

docs/rationale.md
```diff
@@ -0,0 +1,144 @@
...
+Tests for syntax-derived operations are organized according to the language
+grammar, with directories used to describe non-terminals. Tests for built-in
+APIs are organized according to the identifiers by which they can be accessed,
+with directories used to describe the sequence of properties that can be used
+from the global scope.
+
+Directories are not generally applied beyond these limits; further
+differentiation is instead achieved through structured file names which follow
+ad-hoc conventions. This organization balances the need to group tests
+logically with the need to discover tests.
```
sarahghp, line 65: I would love to have examples here as well, even just as links to a few directories that you think serve as good examples of various ad-hoc conventions.

I have also gotten the impression that some files are traditional (`name`, `constructor`, `invoked-as-function`) and some approaches are as well: for instance, separating `throws` from successful invocations. I know these might seem implied or obvious from other sentences here, but listing them out would be helpful, even if in another details file so as not to mix them with high-level.

docs/rationale.md
```diff
@@ -0,0 +1,144 @@
...
+re-organized after being accepted.
+
+## Regression tests
+
+It is possible to write tests for semantics which, while not explicitly
+specified by ECMA262, are nonetheless valid according to the normative text.
+Such tests are welcome in Test262, but their fitness is not a given. Consumers
+from many constituencies value the coherence and consistency of the test suite,
+and tests which disallow arbitrary extraneous behavior can degrade those
+qualities. Because Test262 is not maintained as a repository of regression
```
sarahghp, line 78: Is it possible to give examples here again? (I realize I keep asking this, but I realize I am not able to determine what someone would consider an extraneous test that harms coherence and consistency.)

sarahghp (commented):
Thanks for writing this one up, @jugglinmike! I have a handful of comments but they are mostly the same, which is a request for examples and more specificity.

I suspect you are so familiar with what you mean that many things that seem opaque to me are clear to you, and examples seem like a good way to pass that clarity along.

docs/rationale.md
```diff
@@ -0,0 +1,144 @@
...
+particularly for parts of the grammar that are used in many productions
+(specifically: the destructuring assignment patterns introduced in the 6th
+edition of ECMA262).
+
+This tool makes it easy to introduce enormous numbers of tests. Introducing
+more tests is not a goal unto itself, though. Test262 prioritizes coherence in
+its coverage of the specification, and it recognizes that the value of tests,
+as measured by their likelihood to identify defects, varies.
+
+For these reasons, the maintainers urge restraint in the application of the
```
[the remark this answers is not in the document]
jugglinmike: The word "redundant" suggests an objective quality that is rarely applicable in my experience, so I think using it here would tend to minimize the consideration.

@ptomato To me, the text, "promote uniformity of coverage, particularly for parts of the grammar that are used in many productions," seems to cover gh-3410.

@sarahghp The section references destructuring assignment as one example, think I should add another?

docs/rationale.md
jugglinmike: Great ideas, all.

docs/rationale.md
jugglinmike: Sure

docs/rationale.md
jugglinmike: I included this in recognition of years-old conversation about the test262-parser-tests project, probably with @bakkot (feel free to chime in here, Kevin). Honestly, though, I'm having trouble coming up with an example where I'd recommend forgoing the validation of semantics in favor of syntactic precision.

This draft already has a lot of appeals to subjective judgement; maybe this is a case for more firm advice. How about something like, "Please always validate semantics. If you feel the code required to do so meaningfully reduces the precision of the syntax test, then consider contributing a simplified form of the test to [the test262-parser-tests project](). [...]"

What do you think?

jugglinmike:
Thanks for the feedback, folks! I wan't able to get through it all today, so although I'm sharing my progress so far, please know that I'm not ignoring the pieces that I haven't addressed.

docs/rationale.md
bakkot: The "Please always validate semantics" suggestion sounds good to me.

jugglinmike: Thanks! I've also removed some text which wasn't so relevant.

docs/rationale.md
```diff
@@ -0,0 +1,144 @@
...
+  conditional statements rather than the assertion API implemented by the
+  harness files (though this explicitness will always been desirable when
+  asserting the semantics of conditional statements and `throw` statements
+  themselves)
+- tests with file names derived from section numbers in the 5th edition of
+  ECMA262, e.g. `built-ins/Array/15.4.5-1.js`
+- tests which validate multiple behaviors, using elaborate comment blocks to
+  designate sections
+- tests which use deprecated harness functions, e.g. `verifyEnumerable`,
+  `verifyConfigurable`, and `verifyWritable`
```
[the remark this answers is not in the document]
jugglinmike: Thanks for calling this out! We can't really call something "deprecated" if we haven't documented it as such. Here's a patch to address that: https://github.com/tc39/test262/pull/3449

docs/rationale.md
```diff
@@ -0,0 +1,144 @@
...
+
+Since existing tests do not necessarily reflect the project's current
+best-practices, it's especially important for test authors to familiarize
+themselves with the contribution guidelines.
```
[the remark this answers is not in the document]
jugglinmike: Yes!

docs/rationale.md
jugglinmike: This reflects our current thinking. It's certainly not set in stone, though, and I'm particularly interested in your experience as a proposal champion. At the same time, I don't think this pull request is well-suited to making changes to the rationale. Starting a dedicated discussion based on a finalized document (perhaps using the nascent RFC process) would promote visibility and make it easier for others to recognize if/how maintenance practices changed over time.

Given that the proposed text doesn't give any concrete instruction on the topic, how do you feel about merging it as-is and following up with a discussion about improvements?

README.md
```diff
@@ -34,6 +34,9 @@ Guidance for contributing to Test262 can be found in [CONTRIBUTING.md](./CONTRIB
 
 Guidance for running Test262 and explanations of how a test file must be interpreted by a test runner is in [INTERPRETING](./INTERPRETING.md)
 
+### Rationale
+
+This project offers an explanation for many of its design decisions and maintenance practices--see [rationale.md](./docs/rationale.md).
```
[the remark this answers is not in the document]
jugglinmike: Yeah, including all this in `CONTRIBUTING.md` might make it harder to find. It also might intimidate folks who want to contribute and don't (yet) care about policy stuff.

The placement of the file itself is less substantial, but in the absence of a convention for such a document (such as those for `README`, `CONTRIBUTING`, etc.), I'm inclined to put it in `docs` to reduce clutter in the top level.

docs/rationale.md
jugglinmike: > I would love to have examples here as well, even just as links to a few directories that you think serve as good examples of various ad-hoc conventions.

Sounds good. I've recommended the tests for template literals as an example where tests are organized by a structured file naming scheme.

> I have also gotten the impression that some files are traditional (`name`, `constructor`, `invoked-as-function`) and some approaches are as well: for instance, separating `throws` from successful invocations. I know these might seem implied or obvious from other sentences here, but listing them out would be helpful, even if in another details file so as not to mix them with high-level.

Speaking personally, I don't hold any general expectations at that granularity. Although I might frequently use some names (or even suggest them to others), I don't know if I could formalize rules for which files are appropriate for any given language feature.

docs/rationale.md
jugglinmike: Sure!

docs/rationale.md
```diff
@@ -0,0 +1,180 @@
...
+
+Test262 has been maintained for many years, and the practices used to write
+tests have evolved alongside the needs of its consumers. When conventions
+change, old tests are typically updated to accommodate the new practices. That
+doesn't always happen, though, and one doesn't have to look very far to find
+examples of tests which contradict the preferred patterns.
+
+For instance:
+
+- tests which expression expectations with `throw` statements inside of
```
ljharb, line 15:
```suggestion
- tests which express expectations with `throw` statements inside of
```

docs/rationale.md
```diff
@@ -0,0 +1,180 @@
...
+qualities. Because Test262 is not maintained as a repository of regression
+tests, contributions which include these kinds of tests will be weighed against
+their likelihood of identifying error in a plurality of implementations.
+
+For example, assume that some runtime spuriously accesses the `toJSON` property
+of the value passed to
+[`String.prototype.repeat`](https://tc39.es/ecma262/#sec-string.prototype.repeat).
+While the maintainers of the engine may decide to include a regression test
+which disallows that behavior in their project, the maintainers of Test262
+would not necessarily accept such a test.
```
ljharb, line 122: i'm not sure this matches the first paragraph in the section - doing anything observable that's not in the spec for a method is definitely prohibited.

docs/rationale.md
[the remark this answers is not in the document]
ptomato: I should've made a more concrete suggestion earlier, sorry.

How about:

"To summarize, use the tools to generate tests when complete coverage would be useful for JS engine implementations."

And maybe insert an example at the end of the previous paragraph? "For example, generating tests that cover all possible input values to a function isn't likely to add much value."

docs/rationale.md
ptomato: I think Mike's point is that accessing `.toJSON` as part of `String.prototype.repeat`, while prohibited, is a sufficiently unlikely and out-there spec violation that it's not a good use of test262 bandwidth to encourage people to write tests for it.

docs/rationale.md
```diff
@@ -0,0 +1,180 @@
...
+
+## Large tests
+
+Test262 tests are typically very focused. The vast majority exercise just one
+algorithm step/grammar production, and some are even more granular that that!
```
ptomato, line 127:
```suggestion
algorithm step/grammar production, and some are even more granular than that!
```

ptomato (approved):
Thanks for resuming this!

docs/rationale.md
ljharb: I agree - right up until an implementation does it. That makes it infinitely likely (since it actually happened) and at that point imo we should be strongly inclined to test against it.

docs/rationale.md
```diff
@@ -0,0 +1,180 @@
...
+- tests with file names derived from section numbers in the 5th edition of
+  ECMA262, e.g. `built-ins/Array/15.4.5-1.js`
+- tests which validate multiple behaviors, using elaborate comment blocks to
+  designate sections
+- tests which use deprecated harness functions, e.g. `verifyEnumerable`,
+  `verifyConfigurable`, and `verifyWritable`
+
+Since existing tests do not necessarily reflect the project's current
+best-practices, it's especially important for test authors to familiarize
+themselves with [the contribution guidelines](CONTRIBUTING.md).
```
Ms2ger, line 29:
```suggestion
themselves with [the contribution guidelines](../CONTRIBUTING.md).
```

docs/rationale.md
```diff
@@ -0,0 +1,180 @@
...
+
+Some test contributors are uncomfortable splitting their work across files like
+this. It's certainly unlike the practices that are common in modern application
+development. In those settings, many tests are often grouped into the same file
+and separated by function boundaries.
+
+Test262 doesn't use the same approach as a typical application test suite in
+order to limit complexity. The guidance of "one test per file" means that
+consuming Test262 is relatively easy; there is no "test runner" API for
+consumers to implement, and interpreting results is likewise straightforward.
```
gibson042, line 137: I don't think this adequately explains the rationale for "one test per file"—what constitutes "one test", and what kind of complexity is being limited?

docs/rationale.md
```diff
@@ -0,0 +1,180 @@
...
+It also lowers the barrier to entry for new contributors since there is no API
+to learn.
+
+## Syntax tests
+
+When testing a syntactic feature of the language, it can be tempting to write
+tests which verify that some bit of source text does *not* produce a syntax
+error. Contributors should try to push beyond verifying only the lack of a
```
gibson042, line 145: It _can_ be tempting to write such tests, and this document should explain and attempt to justify the proper response to that temptation.

docs/rationale.md
jugglinmike: @ptomato's interpretation is pretty close to my intent, though it maybe under-emphasizes the part about "identifying error in a plurality of implementations." In any case, I'm confident this has been watered down enough already--the phrase "would not necessarily accept such a test" leaves plenty of room for interpretation.

jugglinmike:
@ms2ger I'd prefer it if you didn't push changes to my fork.

docs/rationale.md
```diff
@@ -0,0 +1,180 @@
...
+For practical reasons, tests are organized in a tree structure according to the
+conventions of modern file systems. Unfortunately, this structure is not
+expressive enough to model the semantics of a rich and evolving programming
+language like ECMAScript. This means the common and crucial task of coverage
+assessment will likely always be a challenge, but strong conventions around
+file organization can help.
+
+Tests for syntax-derived operations are organized according to the language
+grammar, with directories used to describe non-terminals. For example, tests
+for example, tests for [the `if`
```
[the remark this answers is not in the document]
jugglinmike: I've removed the duplication. The examples were added [at Philip's request last year](https://github.com/tc39/test262/pull/3423/commits/1dd07c411bebd86158bbefc50cc5caed6101dd07#r829403415).

docs/rationale.md
jugglinmike: Defining a "test" sounds to me much more like content for a contribution guide. This document is about explaining why Test262 looks the way it does.

docs/rationale.md
jugglinmike: I see what you're saying, and I think the first statement is just missing an operative word. Of course we want tests for syntactically-interesting forms, but the statement "it can be tempting to [verify that runtimes successfully parse some syntax]" heavily insinuates there's something wrong with the practice. I think a qualifier like "only" is critical:

> When testing a syntactic feature of the language, it can be tempting to write tests which verify only that some bit of source text does *not* produce a syntax error.

Thanks for catching this!

ptomato:
As per maintainers meeting, let's move this forward, it's been a while since the last round of comments so there's been enough time for folks to respond - any subsequent discussions please open a new PR!

PR #3426, opened 2022-03-06
Add missing feature flag array-grouping

legendecas:


test/built-ins/Array/prototype/Symbol.unscopables/value.js
```diff
@@ -25,7 +25,7 @@ info: |
     16. Assert: Each of the above calls returns true.
     17. Return unscopableList.
 includes: [propertyHelper.js]
-features: [Symbol.unscopables, array-find-from-last]
+features: [Symbol.unscopables, array-find-from-last, array-grouping]
```
ljharb, line 28: does this mean that this test won’t run at all when an engine has unscopbales, but doesn’t have both find from last and array grouping?

if so, it seems like this is a good case for multiple files, so that the feature applies to its own method name, but doesn’t gate access to the rest of the tests?

legendecas: This is a good idea! Updated.

rwaldron (approved):

Issue #3224, opened 2021-10-01
Temporal "branding" tests are imprecise


jugglinmike:
Many of the tests intended to verify the "this value" (a.k.a "receiver") of the proposal's methods invoke those methods without providing the required arguments. In some cases, this omission would cause the expected TypeError when the implementation incorrectly tolerated an invalid receiver.

[The test for Temporal.Duration.prototype.total](https://github.com/tc39/test262/blob/04ad519be979ae86e21dc157ee5647eb7635e541/test/built-ins/Temporal/Duration/prototype/total/branding.js) currently demonstrates this hazard:

```js
const total = Temporal.Duration.prototype.total;

assert.sameValue(typeof total, "function");

assert.throws(TypeError, () => total.call(undefined), "undefined");
```

If an implementation incorrectly accepted "undefined" as a receiver, it could still spuriously pass this test by throwing a TypeError for the omission of the required "options" argument (that is, by correctly implementing the subsequent step in the algorithm).

These tests should be modified to provide valid values where they are required. Although not all cases of argument omission is susceptible to this problem, the pattern is consistently misleading for human readers, so it should be addressed in all cases.

PR #3431, opened 2022-03-11
Add tests for Temporal.Instant.p*.subtract

ptomato:
This was originally @FrankYFTang's PR #3076 which was closed without being merged. I'm following up as promised. I rebased it, added assertion messages, did some reformatting, and combined with some existing tests.

### DIFF
--- a/test/built-ins/Temporal/Instant/prototype/add/branding.js
+++ b/test/built-ins/Temporal/Instant/prototype/add/branding.js
@@ -11,12 +11,14 @@ const add = Temporal.Instant.prototype.add;
 
 assert.sameValue(typeof add, "function");
 
-assert.throws(TypeError, () => add.call(undefined), "undefined");
-assert.throws(TypeError, () => add.call(null), "null");
-assert.throws(TypeError, () => add.call(true), "true");
-assert.throws(TypeError, () => add.call(""), "empty string");
-assert.throws(TypeError, () => add.call(Symbol()), "symbol");
-assert.throws(TypeError, () => add.call(1), "1");
-assert.throws(TypeError, () => add.call({}), "plain object");
-assert.throws(TypeError, () => add.call(Temporal.Instant), "Temporal.Instant");
-assert.throws(TypeError, () => add.call(Temporal.Instant.prototype), "Temporal.Instant.prototype");
+const arg = new Temporal.Duration(0, 0, 0, 0, 5);
+
+assert.throws(TypeError, () => add.call(undefined, arg), "undefined");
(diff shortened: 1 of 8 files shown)


ptomato:
I've included #3075 in here as well.

test/built-ins/Temporal/Instant/prototype/add/branding.js
```diff
@@ -11,12 +11,14 @@ const add = Temporal.Instant.prototype.add;
...
+
+assert.throws(TypeError, () => add.call(undefined, arg), "undefined");
+assert.throws(TypeError, () => add.call(null, arg), "null");
+assert.throws(TypeError, () => add.call(true, arg), "true");
+assert.throws(TypeError, () => add.call("", arg), "empty string");
+assert.throws(TypeError, () => add.call(Symbol(), arg), "symbol");
+assert.throws(TypeError, () => add.call(1, arg), "1");
+assert.throws(TypeError, () => add.call({}, arg), "plain object");
+assert.throws(TypeError, () => add.call(Temporal.Instant, arg), "Temporal.Instant");
+assert.throws(TypeError, () => add.call(Temporal.Instant.prototype, arg), "Temporal.Instant.prototype");
```
rwaldron, line 24: Nice! I see that this fixes the ambiguity between step 2 and step 3 😄

ptomato: Yeah, I figured even though #3224 is still a problem in other branding tests (which I will tackle separately), Frank's original test nicely avoided that problem, so best to preserve it.

test/built-ins/Temporal/Instant/prototype/add/result-out-of-range.js
```diff
@@ -9,8 +9,22 @@ features: [Temporal]
...
 fields.forEach((field) => {
-  assert.throws(RangeError, () => instance.add({ [field]: 1 }));
+  assert.throws(
+    RangeError,
+    () => latest.add({ [field]: 1 }),
+    `adding ${field} with result out of range (positive)`
+  );
+});
+
+const earliest = Temporal.Instant.fromEpochNanoseconds(-8640000_000_000_000_000_000n);
```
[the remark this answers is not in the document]
ptomato: Oh, I just copied that from the "latest" case without thinking about it too much. In hindsight I think I should've punctuated it as `86400_00000000_000_000_000n` to emphasize that it's 86400 seconds per day, times 10<sup>8</sup> days, times three factors of 1000 to go from seconds to nanoseconds.

PR #3441, opened 2022-03-23
update nfv3 test for roundingIncrement

romulocintra:
Update test for NumberFormat v3 according to : 

https://github.com/tc39/proposal-intl-numberformat-v3/issues/81  and https://github.com/tc39/proposal-intl-numberformat-v3/issues/68 reflecting the  [spec update on L23 and L24](https://tc39.es/proposal-intl-numberformat-v3/out/numberformat/proposed.html#sec-initializenumberformat)

### DIFF
--- a/test/intl402/NumberFormat/constructor-roundingIncrement-invalid.js
+++ b/test/intl402/NumberFormat/constructor-roundingIncrement-invalid.js
@@ -27,8 +27,6 @@ assert.throws(RangeError, function() {
   new Intl.NumberFormat([], {roundingIncrement: 5001});
 }, '5001');
 
-// InitializeNumberFormat ( numberFormat, locales, options ) (...)
-// 23. If roundingIncrement is not 1 and numberFormat.[[RoundingType]] is not fractionDigits, throw a TypeError exception.
 assert.throws(TypeError, function() {
   new Intl.NumberFormat([], {roundingIncrement: 2, roundingPriority: 'morePrecision'});
 }, '2, roundingType is "morePrecision"');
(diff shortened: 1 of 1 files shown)


test/intl402/NumberFormat/constructor-roundingIncrement-invalid.js
```diff
@@ -26,14 +27,22 @@ assert.throws(RangeError, function() {
   new Intl.NumberFormat([], {roundingIncrement: 5001});
 }, '5001');
 
-assert.throws(RangeError, function() {
+// InitializeNumberFormat ( numberFormat, locales, options ) (...)
```
ptomato, line 30: I might put these into the `info:` key in the front matter. Although I tend to find that the `info` key is less useful because it goes stale the fastest, it is current practice. You could also just phrase the comment as prose, something like "roundingIncrement is only allowed for fraction digits rounding".

Same goes for the comment below.

ptomato (dismissed):
This looks correct. What's the has/needs consensus status of the changes in https://github.com/tc39/proposal-intl-numberformat-v3/pull/85 and https://github.com/tc39/proposal-intl-numberformat-v3/pull/91?

romulocintra:
> This looks correct. What's the has/needs consensus status of the changes in [tc39/proposal-intl-numberformat-v3#85](https://github.com/tc39/proposal-intl-numberformat-v3/pull/85) and [tc39/proposal-intl-numberformat-v3#91](https://github.com/tc39/proposal-intl-numberformat-v3/pull/91)?

Consensus : 
https://github.com/tc39/proposal-intl-numberformat-v3/issues/81#issuecomment-1012509356
https://github.com/tc39/proposal-intl-numberformat-v3/issues/68#issuecomment-1071498460

rwaldron:
@ptomato I'll leave this to you to wrap up 👍

test/intl402/NumberFormat/constructor-roundingIncrement-invalid.js
romulocintra: I decided to drop the comments, the description it's self-explanatory  :)

ptomato (approved):
Thanks.

PR #3445, opened 2022-03-25
Temporal.PlainDate invalid string tests

ptomato:
This tests all entry points where a string is converted to a Temporal.PlainDate, with various invalid ISO 8601 strings that should all cause a RangeError to be thrown.


test/built-ins/Temporal/Calendar/prototype/dateAdd/argument-string-invalid.js
```diff
@@ -0,0 +1,49 @@
...
+
+const invalidStrings = [
+  "",
+  "invalid iso8601",
+  "2020-01-00",
+  "2020-01-32",
+  "2020-00-01",
+  "2020-13-01",
+  "-999999-01-01",
+  "+999999-01-01",
```
Ms2ger, line 18: Are those invalid iso strings or just failing ISODateTimeWithinLimits? If the latter, please clarify the description above

test/built-ins/Temporal/PlainDate/compare/argument-string-invalid.js
```diff
@@ -0,0 +1,54 @@
...
+  assert.throws(
+    RangeError,
+    () => Temporal.PlainDate.compare(arg, other),
+    `"${arg}" should not be a valid ISO string for a PlainDate (first argument)`
+  );
+  assert.throws(
+    RangeError,
+    () => Temporal.PlainDate.compare(arg, other),
+    `"${arg}" should not be a valid ISO string for a PlainDate (second argument)`
+  );
```
Ms2ger, line 53: That's the same thing twice

test/built-ins/Temporal/PlainTime/prototype/toZonedDateTime/argument-string-invalid.js
```diff
@@ -0,0 +1,49 @@
...
+  "2020-01-01T001",
+  "2020-01-01T01:001",
+  "2020-01-01T01:01:001",
+  "2020-01-01[u-ca=notexist]",
+];
+const instance = new Temporal.PlainTime(12, 34, 56, 987, 654, 321);
+for (const arg of invalidStrings) {
+  assert.throws(
+    RangeError,
+    () => instance.toZonedDateTime({ plainDate: arg }),
```
Ms2ger, line 46: I guess this is strictly speaking not an argument, but no need to change.

Ms2ger (approved):

test/built-ins/Temporal/Calendar/prototype/dateAdd/argument-string-invalid.js
ptomato: I guess these could go in a separate test.

ptomato: so you'd like the description to say something like "RangeError thrown if an invalid ISO string (or valid ISO string outside the supported range) is used as a PlainDate"?

Ms2ger: That would work for me, yeah.

ptomato:
I've reorganized the list of strings that should throw, and added some comments clarifying whether they are syntactically invalid or just not supported (and why).

rwaldron (approved):

PR #3448, opened 2022-03-25
Use ECMAScript version 11

jessealama:
We use BigInt syntax in a bunch of tests. This change
registers that fact with linters that use `.jshintrc`.


ptomato (approved):
Makes sense to me to keep this updated to the latest valid value. There have to be tests for things present in the latest edition :smile:

jugglinmike (approved):

PR #3449, opened 2022-03-25
Deprecate some property helpers

jugglinmike:
Document the preference for `verifyProperty` over the various other
property-related helper functions.

### DIFF
--- a/harness/propertyHelper.js
+++ b/harness/propertyHelper.js
@@ -6,13 +6,13 @@ description: |
     property descriptors.
 defines:
   - verifyProperty
-  - verifyEqualTo
-  - verifyWritable
-  - verifyNotWritable
-  - verifyEnumerable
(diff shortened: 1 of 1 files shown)


rwaldron:
This makes me very happy.

rwaldron (approved):

ptomato:
I was wondering if it might be good to rewrite the deprecated helpers in terms of the non-deprecated helper, in order to show people how to update, but I hadn't taken a close look at whether that was possible yet.

jugglinmike:
@ptomato We've been discouraging the use of these functions because their side-effects tacitly impose restrictions on the sequence in which they should be invoked. While we would rewrite their implementation, doing so would not be instructive for test authors because the preferred function is designed to abstract away the implicit sequencing requirements.

PR #3450, opened 2022-03-26
Remove configuration file for third-party tool

jugglinmike:
The `.jshintrc` file configures the JavaScript "linting" tool named
JSHint. Test262 does not depend on that tool, making the file's purpose
and validity ambiguous and potentially distracting.

---

@jessealama before reviewing gh-3448, I would have assumed that no one is using the `.jshintrc` file. Maintaining your own configuration will reduce friction for changes and risk of incompatibilities. Will that work for you?


jessealama:
I think removing `.jshintrc` tool is a valid decision. I agree that Test262 doesn't depend on JSHint, and that having `.jshintrc` is a potential source of friction. I will manage the file in a private `.gitignore`. (NB WebStorm sniffs around for a `.jshintrc` file, among other things, when working with JS codebases. WebStorm's suggestion was my initial motivation to include this file.)

That said, I think it's also not unreasonable to include the file. It's an expression of the Test262 community's code standards. Having a linter config in the repo documents those standards, if imperfectly. There may even come a time when we will need something like `.jshintrc`. For instance, if Test262 continues to grow and attract more contributors, we may need to add a linter to our CI as part of the review process.

In summary: getting rid of it is OK with me.

ljharb:
I think that the common practice in the JS ecosystem is to only include config files that the project enforces; that one should never use a linting/formatting tool on a project unless the project has it configured; and that eslint is the defacto standard (for a long time now).

test262 should definitely be using eslint, but that's a different discussion.

rwaldron (dismissed):

rwaldron:
@jugglinmike please resolve the conflict—thanks!

jugglinmike:
All set, @rwaldron!

PR #3452, opened 2022-03-28
Fixed YAML in some new test metadata

### LINKED ISSUE #3451: Incorrect YAML metadata in some tests
While trying to update the test suite for the [Boa](https://github.com/boa-dev/boa) JavaScript engine (https://github.com/boa-dev/boa/pull/1982), we have found that the metadata of a couple of new tests contains invalid YAML:

```
./test/built-ins/ShadowRealm/WrappedFunction/throws-typeerror-on-revoked-proxy.js
./test/built-ins/ShadowRealm/prototype/evaluate/throws-typeerror-wrap-throwing.js
```

The error in both cases is the same. The description contains 2 spaces and a tab character instead of 4 spaces.

Razican:
This closes #3451.

It coverts the tab character to spaces to avoid YAML parsing errors.


ptomato (approved):
Thanks for the fix! Agreed, there are not supposed to be mixed spaces and tabs there.

PR #3455, opened 2022-03-30
Undefined options check in Temporal

ptomato:
This adds tests for https://github.com/tc39/proposal-temporal/pull/2028, which achieved consensus at the March 2022 TC39 meeting.

Additionally includes a bug fix in one of the custom time zones used for testing time zone arithmetic in absence of Intl in the implementation.

### DIFF
--- a/test/harness/temporalHelpers-one-shift-time-zone.js
+++ b/test/harness/temporalHelpers-one-shift-time-zone.js
@@ -1,0 +1,80 @@
+// Copyright (C) 2022 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: >
+  Verify the time zone arithmetic used in TemporalHelpers.oneShiftTimeZone()
+  against known cases in the implementation's time zone database
+includes: [compareArray.js, temporalHelpers.js]
+features: [Temporal]
+---*/
+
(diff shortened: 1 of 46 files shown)


test/harness/temporalHelpers-one-shift-time-zone.js
```diff
@@ -0,0 +1,80 @@
...
+  against known cases in the implementation's time zone database
+includes: [compareArray.js, temporalHelpers.js]
+features: [Temporal]
+---*/
+
+function checkTimeZoneArithmetic(shiftInstant, shiftNs, realTimeZoneName, shiftWallTime) {
+  // No need to test this on hosts that don't provide an Intl object. It's
+  // sufficient that the logic is tested on at least one host.
+  if (typeof globalThis.Intl === "undefined")
+    return;
```
ptomato, line 16: I'm not a big fan of this. Any suggestions on how to organize it differently are appreciated.

rwaldron: Normally a test should fail if a requisite aspect is not supported, for example: https://github.com/tc39/test262/blob/main/test/built-ins/ShadowRealm/prototype/importValue/validates-realm-object.js#L10-L14

rwaldron (approved):

PR #3459, opened 2022-03-31
Tests for Temporal formatting the year appropriately as 4 or 6 digits

ptomato:
https://github.com/tc39/proposal-temporal/pull/2090 is a normative change
that reached consensus at the March 2022 TC39 plenary meeting. This adds
tests that verify the change made to the formatting of years between 0 and
999 inclusive in all toString and toJSON methods of Temporal types that
can output an ISO year number in their return value.


rwaldron (approved):
Thorough!!

PR #3462, opened 2022-04-01
Tests for computing PlainYearMonth addition and subtraction in correct calendar space

ptomato:
https://github.com/tc39/proposal-temporal/pull/2003 is a normative change
that reached consensus at the March 2022 TC39 plenary meeting. This adds
tests that verify the new spec text is implemented correctly, performing
arithmetic on a PlainYearMonth instance that would previously have thrown
an error if it was implemented as written.


rwaldron (approved):
These are pretty dense, but I trust you :)

ptomato:
@rwaldron Agreed, and sorry about the density! We can ask Ms2ger to review them when he's available if you want an extra pair of eyes.

If it helps, I think the code example in https://github.com/tc39/proposal-temporal/pull/2003 gives a better description of what needs to be tested — but we can't use that example directly since it relies on the `chinese` calendar provided by `Intl`, so most of the complexity is me constructing a user calendar that would exhibit the same problem.

PR #3463, opened 2022-04-01
Fix false negative in for-in test

jugglinmike:
As originally written, this test would spuriously pass when the deleted
property was incorrectly visited by enumation but correctly removed from
the object. In such cases, the accumulator string would take the form

    "aa1baundefinedca3"

And satisfy all conditions intended to highlight implementation errors.

Refactor the test to avoid false negative by using an object with a null
prototype and verifying the exact contents of the accumulator string.

### DIFF
--- a/test/language/statements/for-in/S12.6.4_A7_T2.js
+++ b/test/language/statements/for-in/S12.6.4_A7_T2.js
@@ -28,7 +28,10 @@ for (var __key in __obj){
 	
 }
 
-assert.sameValue(__accum, "aa1ca3");
+assert(
+	__accum === "aa1ca3" || __accum === "ca3aa1",
+	"Unexpected value: '" + __accum + "'"
+);
 
 // erasator is the hash map terminator
 function erasator_T_1000(hash_map, charactr){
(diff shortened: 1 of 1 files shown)


test/language/statements/for-in/S12.6.4_A7_T2.js
```diff
@@ -25,23 +28,7 @@ for (var __key in __obj){
...
-
-//////////////////////////////////////////////////////////////////////////////
-//CHECK#2
-if (__accum.indexOf("ba2")!==-1) {
-	throw new Test262Error('#2: __accum.indexOf("ba2") === -1. Actual:  __accum.indexOf("ba2") ==='+ __accum.indexOf("ba2")  );
-}
-//
-//////////////////////////////////////////////////////////////////////////////
-
+assert.sameValue(__accum, "aa1ca3");
```
IdanHo, line 31: The specification does not guarantee the iteration order of `for-in` loops, so I think this can possibly fail in valid implementations.

jugglinmike: Ah, right. Thanks

rwaldron (approved):
lol, much better.

PR #3465, opened 2022-04-04
Remove check for per-iteration detach check in TypedArray.prototype.set

syg:
This updates tests in line with the normative change in
https://github.com/tc39/ecma262/pull/2646


rwaldron:
Merged locally

iamstolis:
This PR forgot to update [BigInt version](https://github.com/tc39/test262/blob/main/test/built-ins/TypedArray/prototype/set/BigInt/array-arg-targetbuffer-detached-on-get-src-value-throws.js) of `array-arg-targetbuffer-detached-on-get-src-value-throws.js` test.

syg:
> This PR forgot to update [BigInt version](https://github.com/tc39/test262/blob/main/test/built-ins/TypedArray/prototype/set/BigInt/array-arg-targetbuffer-detached-on-get-src-value-throws.js) of array-arg-targetbuffer-detached-on-get-src-value-throws.js test.

Nice catch. I took a look and that test isn't quite what it set out to test, which is why it didn't show up on local failures and wasn't updated as part of this PR.

The BigInt version throws a `TypeError` even with the detach check removed, because the getter that does the detaching implicitly returns undefined, which [throws a `TypeError` in ToBigInt](https://tc39.es/ecma262/#sec-tobigint). So the test continues to pass.

In any case I think it's clear that that wasn't what they were intending to test, so I'll make a PR to remove the test.

PR #3466, opened 2022-04-04
Remove check for per-comparator call detach check in TypedArray.prototype.sort

syg:
This updates tests in line with the normative change in
https://github.com/tc39/ecma262/pull/2723


rwaldron:
Merged locally

PR #3472, opened 2022-04-10
Add test on handling broken promises in AsyncGenerator.prototype.return

legendecas:
Tests for https://github.com/tc39/ecma262/pull/2683.

### DIFF
--- a/test/built-ins/AsyncGeneratorPrototype/return/return-blocked-broken-promise.js
+++ b/test/built-ins/AsyncGeneratorPrototype/return/return-blocked-broken-promise.js
@@ -1,0 +1,59 @@
+// Copyright (C) 2022 Chengzhong Wu. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-asyncgenerator-prototype-return
+description: >
+  A broken promise should reject the returned promise of
+  AsyncGenerator.prototype.return when the generator is completed.
+info: |
+  AsyncGenerator.prototype.return ( value )
+  ...
+  8. If state is either suspendedStart or completed, then
+    a. Set generator.[[AsyncGeneratorState]] to awaiting-return.
+    b. Perform ! AsyncGeneratorAwaitReturn(generator).
+  ...
+
+  AsyncGeneratorAwaitReturn ( generator )
+  ...
+  6. Let promise be Completion(PromiseResolve(%Promise%, completion.[[Value]])).
+  7. If promiseCompletion is an abrupt completion, then
+    a. Set generator.[[AsyncGeneratorState]] to completed.
+    b. Perform AsyncGeneratorCompleteStep(generator, promiseCompletion, true).
+    c. Perform AsyncGeneratorDrainQueue(generator).
+    d. Return unused.
+  8. Assert: promiseCompletion.[[Type]] is normal.
+  9. Let promise be promiseCompletion.[[Value]].
+  ...
+
+flags: [async]
+features: [async-iteration]
+---*/
+
+let unblock;
+let blocking = new Promise(res => { unblock = res; });
+var g = async function*() {
+  await blocking;
+};
+
+var it = g();
+let brokenPromise = Promise.resolve(42);
+Object.defineProperty(brokenPromise, 'constructor', {
+  get: function () {
+    throw new Error('broken promise');
+  }
+});
+
+it.next();
+it.return(brokenPromise)
+  .then(
+    () => {
+      throw new Test262Error("Expected rejection");
+    },
+    err => {
+      assert.sameValue(err.message, 'broken promise');
+    }
+  )
+  .then($DONE, $DONE);
+
+unblock();
--- a/test/built-ins/AsyncGeneratorPrototype/return/return-state-completed-broken-promise.js
+++ b/test/built-ins/AsyncGeneratorPrototype/return/return-state-completed-broken-promise.js
@@ -31,9 +31,11 @@ features: [async-iteration]
 ---*/
 
 let unblock;
-let blocking = new Promise(res => { unblock = res; });
+let blocking = new Promise(resolve => { unblock = resolve; });
+let unblocked = false;
 var g = async function*() {
   await blocking;
+  unblocked = true;
 };
 
 var it = g();
@@ -51,6 +53,7 @@ it.return(brokenPromise)
       throw new Test262Error("Expected rejection");
     },
     err => {
+      assert(unblocked, false, 'return should be rejected before generator is resumed');
       assert.sameValue(err.message, 'broken promise');
     }
   )
--- a/test/built-ins/AsyncGeneratorPrototype/return/return-suspendedStart-broken-promise.js
+++ b/test/built-ins/AsyncGeneratorPrototype/return/return-suspendedStart-broken-promise.js
@@ -1,0 +1,54 @@
+// Copyright (C) 2022 Chengzhong Wu. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-asyncgenerator-prototype-return
+description: >
+  A broken promise should reject the returned promise of
+  AsyncGenerator.prototype.return.
+info: |
+  AsyncGenerator.prototype.return ( value )
+  ...
+  8. If _state_ is either ~suspendedStart~ or ~completed~, then
+    a. Set _generator_.[[AsyncGeneratorState]] to ~awaiting-return~.
+    b. Perform AsyncGeneratorAwaitReturn(_generator_).
+  ...
+
+  AsyncGeneratorAwaitReturn ( generator )
+  ...
+  6. Let promise be Completion(PromiseResolve(%Promise%, completion.[[Value]])).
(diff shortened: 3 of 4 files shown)


test/built-ins/AsyncGeneratorPrototype/return/return-suspendedYield-broken-promise-try-catch.js
```diff
@@ -0,0 +1,62 @@
...
+  ...
+
+flags: [async]
+features: [async-iteration]
+---*/
+
+var g = async function*() {
+  try {
+    yield 0;
+    return 1;
```
ptomato, line 42: Maybe for extra clarity for readers, change this to something like the following?
```suggestion
    return 'this is never returned';
```

test/built-ins/AsyncGeneratorPrototype/return/return-suspendedYield-broken-promise-try-catch.js
```diff
@@ -0,0 +1,62 @@
...
+    throw new Error('broken promise');
+  }
+});
+
+var it = g();
+it.next().then(() => {
+  return gen.return(brokenPromise)
+}).then(ret => {
+  assert.sameValue(ret.value, 2, 'returned value');
+  assert.sameValue(ret.done, true, 'returned value');
```
ptomato, line 61:
```suggestion
  assert.sameValue(ret.done, true, 'iterator is closed');
```
?

test/built-ins/AsyncGeneratorPrototype/return/return-blocked-broken-promise.js
```diff
@@ -0,0 +1,59 @@
...
+  8. Assert: promiseCompletion.[[Type]] is normal.
+  9. Let promise be promiseCompletion.[[Value]].
+  ...
+
+flags: [async]
+features: [async-iteration]
+---*/
+
+let unblock;
+let blocking = new Promise(res => { unblock = res; });
```
ptomato, line 34: Although I believe I understand how this is working now, I found this difficult to comprehend on the first few readings :smile: One thing that got me on the wrong track was that I was reading `res` as "result" rather than "resolve" and was thinking of this arrow function as a callback passed to `then()` rather than the `(resolve, reject) => {...}` passed to the Promise constructor. Maybe writing out the name like this could make it clearer?
```suggestion
let blocking = new Promise(resolve => { unblock = resolve; });
```

Although, I'm not 100% clear on why we need to block the generator in order to test `return(brokenPromise)` in the "completed" state. Would it be sufficient to do something like
```js
var g = async function*() {};

var it = g();
it.return();
it.return(brokenPromise).then(...)
```
(There's probably a good reason why not, so maybe a comment would be helpful.)

test/built-ins/AsyncGeneratorPrototype/return/return-blocked-broken-promise.js
```diff
@@ -0,0 +1,59 @@
+// Copyright (C) 2022 Chengzhong Wu. All rights reserved.
```
ptomato, line 1: I would suggest renaming this test from "blocked" to "completed", since the other tests are named after the AsyncGeneratorState that they are testing.

test/built-ins/AsyncGeneratorPrototype/return/return-blocked-broken-promise.js
```diff
@@ -0,0 +1,59 @@
...
+
+/*---
+esid: sec-asyncgenerator-prototype-return
+description: >
+  A broken promise should reject the returned promise of
+  AsyncGenerator.prototype.return when the generator is completed.
+info: |
+  AsyncGenerator.prototype.return ( value )
+  ...
+  8. If state is either suspendedStart or completed, then
```
ptomato, line 12: I'm not sure whether we prefer having ecmarkup like `_state_` and `~suspendedStart~` here in the front matter or just plain text, but it should be consistent in between these tests; currently, the suspendedStart one uses ecmarkup and the other two don't.

test/built-ins/AsyncGeneratorPrototype/return/return-suspendedYield-broken-promise-try-catch.js
```diff
@@ -0,0 +1,62 @@
+// Copyright (C) 2022 Chengzhong Wu. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-asyncgenerator-prototype-return
+description: >
+  A broken promise should resume the generator and reject with
+  the exception.
```
ptomato, line 8:
```suggestion
  the exception when the generator's state is suspendedYield.
```

test/built-ins/AsyncGeneratorPrototype/return/return-suspendedStart-broken-promise.js
```diff
@@ -0,0 +1,54 @@
+// Copyright (C) 2022 Chengzhong Wu. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-asyncgenerator-prototype-return
+description: >
+  A broken promise should reject the returned promise of
+  AsyncGenerator.prototype.return.
```
ptomato, line 8:
```suggestion
  AsyncGenerator.prototype.return when the generator's state is suspendedStart.
```

ptomato (commented):
Thanks for these tests! I have to say, this is my first time really digging into how AsyncGenerators are specified, so several of my comments might come from misunderstanding something. But hopefully I've also suggested some improvements that make it easier to comprehend by someone new to this part of the specification.

Do we need to have tests for the "executing" and "awaiting-return" states as well? Or is this enough since steps 10 and 11 of https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-asyncgenerator-prototype-return weren't touched by the normative PR?

test/built-ins/AsyncGeneratorPrototype/return/return-blocked-broken-promise.js
ptomato: Ah, that explains it! Much clearer now with the new check, thank you.

ptomato (dismissed):
Anyone else want to take a look?

legendecas:
> Do we need to have tests for the "executing" and "awaiting-return" states as well? Or is this enough since steps 10 and 11 of https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-asyncgenerator-prototype-return weren't touched by the normative PR?

I didn't find such tests. I can add those in a follow-up PR.

test/built-ins/AsyncGeneratorPrototype/return/return-suspendedYield-broken-promise-try-catch.js
```diff
@@ -0,0 +1,62 @@
...
+let brokenPromise = Promise.resolve(42);
+Object.defineProperty(brokenPromise, 'constructor', {
+  get: function () {
+    throw new Error('broken promise');
+  }
+});
+
+var it = g();
+it.next().then(() => {
+  return gen.return(brokenPromise)
```
jugglinmike, line 58:
```suggestion
  return it.return(brokenPromise);
```

test/built-ins/AsyncGeneratorPrototype/return/return-suspendedYield-broken-promise-try-catch.js
```diff
@@ -0,0 +1,62 @@
...
+flags: [async]
+features: [async-iteration]
+---*/
+
+var g = async function*() {
+  try {
+    yield;
+    return 'this is never returned';
+  } catch (err) {
+    assert.sameValue(err.message, 'broken promise');
```
jugglinmike, line 44: Since this test is scrutinizing the completion semantics of this function, we should avoid making assertions within the body. Instead, we could store the value in a mutable binding declared in a higher scope, and assert that value in the test's final validation phase.

jugglinmike (changes requested):

test/built-ins/AsyncGeneratorPrototype/return/return-suspendedYield-broken-promise-try-catch.js
legendecas: Thank you for your suggestion! Updated :)

jugglinmike (approved):
Thanks, @legendecas, this looks good to me!

We'll coordinate with the ECMA262 maintainers to merge this and the corresponding spec patch.

jugglinmike:
Thanks again, @legendecas!

test/built-ins/AsyncGeneratorPrototype/return/return-state-completed-broken-promise.js
```diff
@@ -0,0 +1,62 @@
...
+});
+
+it.next();
+it.return(brokenPromise)
+  .then(
+    () => {
+      throw new Test262Error("Expected rejection");
+    },
+    err => {
+      assert(unblocked, false, 'return should be rejected before generator is resumed');
```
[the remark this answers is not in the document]
legendecas: I think this assertion message is incorrect, and the second parameter is redundant: https://github.com/tc39/test262/pull/4486

PR #3468, opened 2022-04-08
Test improvements for Temporal toString methods with options for outputting seconds

ptomato:
This PR collects some improvements to `toString()` methods of Temporal objects, regarding the options for output of seconds and subseconds, and some improvements to tests for `smallestUnit` and `largestUnit` options in general.

Consists mostly of making similar tests consistent with each other, and adding better assertion messages.

Best reviewed commit by commit. There is more explanation in each commit message.

PR #3478, opened 2022-04-12
Temporal: Port Demitasse PlainDateTime `round` tests

jessealama:


test/built-ins/Temporal/PlainDateTime/prototype/round/round-toward-big-bang.js
```diff
@@ -0,0 +1,36 @@
+// Copyright (C) 2022 Igalia, S.L. All rights reserved.
```
Ms2ger, line 1: I think this is redundant with `test/built-ins/Temporal/PlainDateTime/prototype/round/rounding-direction.js` in #3468

test/built-ins/Temporal/PlainDateTime/prototype/round/roundingmode-ceil-basic.js
```diff
@@ -0,0 +1,30 @@
...
+  "nanosecond": [1976, 11, "M11", 18, 14, 23, 30, 123, 456, 789]
+};
+
+Object.entries(incrementOneCeil).forEach(([smallestUnit, expected]) => {
+  const [y, mon, code, d, h, min, s, ms, µs, ns] = expected;
+  TemporalHelpers.assertPlainDateTime(
+    dt.round({ smallestUnit, roundingMode: "ceil" }),
+    y, mon, code, d, h, min, s, ms, µs, ns,
+    `rounds up to ${smallestUnit} (ceil)`
+  );
```
Ms2ger, line 29: I think this should work:
```suggestion
  TemporalHelpers.assertPlainDateTime(
    dt.round({ smallestUnit, roundingMode: "ceil" }),
    ...expected,
    `rounds up to ${smallestUnit} (ceil)`
  );
```

If you think it's an improvement, please adjust all similar tests.

test/built-ins/Temporal/PlainDateTime/prototype/round/roundingmode-halfexpand-is-default.js
```diff
@@ -0,0 +1,21 @@
...
+---*/
+
+const dt = new Temporal.PlainDateTime(1976, 11, 18, 14, 23, 30, 123, 456, 789);
+
+const units = ["day", "hour", "minute", "second", "millisecond", "microsecond", "nanosecond"];
+
+units.forEach((unit) => {
+  TemporalHelpers.assertPlainDateTimesEqual(
+    dt.round({ roundingmode: "halfExpand", smallestUnit: unit }),
+    dt.round({ smallestUnit: unit }),
```
Ms2ger, line 18: I would prefer you to copy the actual expected values from `test/built-ins/Temporal/PlainDateTime/prototype/round/roundingmode-halfexpand-basic.js` rather than calling `dt.round()` twice per test

Ms2ger (changes requested):

test/built-ins/Temporal/PlainDateTime/prototype/round/roundingmode-ceil-basic.js
jessealama: Cool! I didn't know you could do that kind of destructuring. I agree that that's an improvement. I'll adopt this suggestion, both here and in the other places.

test/built-ins/Temporal/PlainDateTime/prototype/round/roundingmode-halfexpand-is-default.js
jessealama: Sure, done.

test/built-ins/Temporal/PlainDateTime/prototype/round/round-toward-big-bang.js
jessealama: Yeah -- I deleted it. I didn't know about Philip's work there. Good catch!

test/built-ins/Temporal/PlainDateTime/prototype/round/roundingmode-halfexpand-is-default.js
```diff
@@ -0,0 +1,36 @@
...
+};
+
+const expected = [1976, 11, "M11", 18, 0, 0, 0, 0, 0, 0];
+
+Object.entries(units).forEach(([unit, expected]) => {
+  TemporalHelpers.assertPlainDateTime(
+    dt.round({ roundingmode: "halfExpand", smallestUnit: unit }),
+    ...expected,
+    `halfExpand is the default (smallest unit = ${unit}, rounding mode present)`
+  );
```
Ms2ger, line 30: This should be covered by the other file, right?
```suggestion
```

Ms2ger (dismissed):

test/built-ins/Temporal/PlainDateTime/prototype/round/roundingmode-halfexpand-is-default.js
jessealama: True, good point. Having the tests next to each other in the same file makes the intention clearer, I'd say ("with and without, you get the same result"), but I can't deny that we're doing the same work here that we're doing elsewhere.

PR #3481, opened 2022-04-12
Temporal: Some more tests for PlainDateTime#with.

Ms2ger:


test/built-ins/Temporal/PlainDateTime/prototype/with/calendar-options.js
```diff
@@ -0,0 +1,27 @@
+// Copyright (C) 2022 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.plaindatetime.prototype.with
+description: Verify the result of calendar.fields() is treated correctly.
```
ptomato, line 6: It's not clear to me from the description what the test is actually testing — it seems to test that the options object is passed through to `dateFromFields` which is different from what the description says. I'd expect to do what the description says, we'd need to add another field in an overridden `fields()` method?

test/built-ins/Temporal/PlainDateTime/prototype/with/calendar-options.js
```diff
@@ -0,0 +1,27 @@
...
+  dateFromFields(fields, optionsArg) {
+    ++calledDateFromFields;
+    assert.sameValue(optionsArg, options, "should pass options object through");
+    return super.dateFromFields(fields, optionsArg);
+  }
+};
+const calendar = new Calendar();
+const plaindatetime = new Temporal.PlainDateTime(2000, 5, 2, 12, 34, 56, 987, 654, 321, calendar);
+const result = plaindatetime.with({ year: 2005 }, options);
+TemporalHelpers.assertPlainDateTime(result, 2005, 5, "M05", 2, 12, 34, 56, 987, 654, 321);
```
ptomato, line 26: An assertion message might help clarify what the test is doing.

ptomato (changes requested):
I have comments on one of these tests, the rest look good!

test/built-ins/Temporal/PlainDateTime/prototype/with/calendar-options.js
Ms2ger: Forgot to update this description after copy/pasting from another test; fixed.

ptomato (approved):

PR #3485, opened 2022-04-13
Remove check for per-iteration detach check in TypedArray.prototype.set

syg:
Note that this test currently unintentionally passes, because a
TypeError is thrown for failing to convert the undefined returned from
the getter to a BigInt. But since this test was intended to test
detaching, it's no longer valid and should still be removed.

See https://github.com/tc39/test262/pull/3465#issuecomment-1098388916


rwaldron:
Look at that! GraalJS isn't failing to install!

rwaldron (approved):

PR #3491, opened 2022-04-19
Temporal: Port some Duration tests.

Ms2ger:


ptomato (approved):
Generally I'd try to add assertion messages in more places, but I'd say that's optional. Looks good otherwise, I don't have any specific comments.

Issue #3401, opened 2022-02-07
GraalJS fails to install in CI


Ms2ger:
```
#!/bin/bash -eo pipefail
./node_modules/.bin/esvu install $hostName

esvu ❯ version 1.2.5
GraalJS ❯ Checking version...
GraalJS ❯ Installing version 22.0.0.2
GraalJS ❯ Downloading https://github.com/graalvm/graalvm-ce-builds/releases/download/vm-22.0.0.2/graalvm-ce-java11-linux-amd64-22.0.0.2.tar.gz
progress [----------------------------------------] 0% | ETA: 0s | 0/425118686progress [=---------------------------------------] 2% | ETA: 3s | 12435456/425118686progress [===-------------------------------------] 6% | ETA: 3s | 28066752/425118686progress [=====-----------------------------------] 11% | ETA: 3s | 49922048/425118686progress [=======---------------------------------] 17% | ETA: 3s | 72302592/425118686progress [=========-------------------------------] 21% | ETA: 3s | 93093888/425118686progress [===========-----------------------------] 27% | ETA: 3s | 115760256/425118686progress [=============---------------------------] 33% | ETA: 2s | 141410304/425118686progress [===============-------------------------] 37% | ETA: 2s | 160791168/425118686progress [==================----------------------] 43% | ETA: 2s | 186282112/425118686progress [====================--------------------] 51% | ETA: 2s | 217006080/425118686progress [=======================-----------------] 57% | ETA: 2s | 245678080/425118686progress [=========================---------------] 62% | ETA: 0s | 267141120/425118686progress [==========================--------------] 66% | ETA: 0s | 281018368/425118686progress [===========================-------------] 66% | ETA: 1s | 284377088/425118686progress [============================------------] 70% | ETA: 1s | 300515328/425118686progress [===============================---------] 76% | ETA: 1s | 326746112/425118686progress [================================--------] 80% | ETA: 1s | 344309760/425118686progress [==================================------] 85% | ETA: 1s | 364904448/425118686progress [====================================----] 90% | ETA: 0s | 385875968/425118686progress [======================================--] 95% | ETA: 1s | 406847488/425118686progress [========================================] 99% | ETA: 0s | 425115648/425118686progress [========================================] 100% | ETA: 0s | 425118686/425118686GraalJS ❯ Extracting /tmp/26ff3db048a58cc2edb927557bbb1bc2
[...]

PR #3492, opened 2022-04-19
CI: disable graaljs test run

rwaldron:
Ref. #3401.


ljharb (approved):

sarahghp (approved):
This looks good to me, although I wonder if we shouldn't add an explanation to this PR for why we are making the change, in case future us forgets and wonders.

Ms2ger:
I pointed to the issue.

PR #3494, opened 2022-04-20
Make minimumFractionDigits = maximumFractionDigits

FrankYFTang:
Make minimumFractionDigits equal to maximumFractionDigits while
roundingIncrement != 1 to sync with latest spec change in
https://github.com/tc39/proposal-intl-numberformat-v3/pull/85
which landed to spec text in Feb 22, 2022

@sffc @romulocintra

### DIFF
--- a/test/intl402/NumberFormat/constructor-roundingIncrement.js
+++ b/test/intl402/NumberFormat/constructor-roundingIncrement.js
@@ -39,7 +39,8 @@ for (const [value, expected] of values) {
     get roundingIncrement() {
       callOrder.push("roundingIncrement");
       return value;
-    }
+    },
+    minimumFractionDigits: 3
   });
   const resolvedOptions = nf.resolvedOptions();
   assert("roundingIncrement" in resolvedOptions, "has property for value " + value);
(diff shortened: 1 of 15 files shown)


FrankYFTang:
@ryzokuken @jugglinmike

test/intl402/NumberFormat/constructor-roundingIncrement.js
```diff
@@ -39,7 +39,8 @@ for (const [value, expected] of values) {
     get roundingIncrement() {
       callOrder.push("roundingIncrement");
       return value;
-    }
+    },
+    minimumFractionDigits: 3
```
sffc, line 43: Observation: This case, where maximumFractionDigits is not present and therefore defaults to 3, wasn't one that was considered when making the spec change. This is a little bit awkward, but I think it's fine, since in most cases roundingIncrement should have an explicit maximumFractionDigits value set (rather than relying on the implicit default).

sffc (approved):
LGTM

romulocintra (approved):
lgtm

PR #3495, opened 2022-04-21
Temporal: Port Demitasse tests for PlainDateTime toZonedDateTime

jessealama:


jessealama:
Matches [this](https://github.com/tc39/proposal-temporal/pull/2146) PR on the proposal-temporal side.

test/built-ins/Temporal/PlainDateTime/prototype/toZonedDateTime/disambiguation-invalid.js
```diff
@@ -0,0 +1,18 @@
+// Copyright (C) 2022 Igalia, S.L. All rights reserved.
```
ptomato, line 1: Consider naming this test `disambiguation-invalid-string.js` and expanding the invalid strings a bit to match https://github.com/tc39/test262/blob/main/test/built-ins/Temporal/PlainDate/prototype/add/overflow-invalid-string.js (e.g. "earlier" with a dotless i to test Unicode canonicalization, a string with an extra \0 in it)

test/built-ins/Temporal/PlainDateTime/prototype/toZonedDateTime/options-object.js
```diff
@@ -0,0 +1,23 @@
...
+esid: sec-temporal.plaindate.prototype.tozoneddatetime
+description: Empty object may be used as options
+features: [Temporal]
+---*/
+
+const dt = new Temporal.PlainDateTime(2019, 10, 29, 10, 46, 38, 271, 986, 102);
+const expected = "2019-10-29T10:46:38.271986102-03:00[America/Sao_Paulo]";
+
+assert.sameValue(
+  `${dt.toZonedDateTime("America/Sao_Paulo", {})}`,
```
ptomato, line 14: - Use "UTC" since "America/Sao_Paulo" will not be a valid time zone for non-Intl implementations
- Prefer not using `toString` to do the assertion, instead check the expected value of `.epochNanoseconds`.

(Both these comments apply to most tests in this PR. In some cases you might also need to check `.timeZone` and `.calendar`)

test/intl402/Temporal/PlainDateTime/prototype/toZonedDateTime/disambiguate-basic.js
```diff
@@ -0,0 +1,35 @@
...
+/*---
+esid: sec-temporal.plaindatetime.prototype.tozoneddatetime
+description: Basic cases for disambiguate options
+features: [Temporal]
+---*/
+
+const dt = new Temporal.PlainDateTime(2020, 3, 8, 2, 0);
+
+assert.sameValue(
+  `${dt.toZonedDateTime("America/Los_Angeles", { disambiguation: "compatible" })}`,
```
ptomato, line 13: In this case you'll need to have a custom time zone with a DST transition, so consider using `TemporalHelpers.springForwardFallBackTimeZone`.

ptomato (changes requested):

test/built-ins/Temporal/PlainDateTime/prototype/toZonedDateTime/disambiguation-invalid.js
jessealama: Good idea -- done

test/built-ins/Temporal/PlainDateTime/prototype/toZonedDateTime/options-object.js
jessealama: Good point -- I've updated the tests to use UTC or, in some cases, the spring-ahead-fall-back setup in `temporalHelpers.js`. I've tweaked things so that we don't test against the stringified object, but compare the epoch nanoseconds. That said, I'm not sure where to test the time zone and calendar, though. Any advice on where that makes sense, and where not?

test/intl402/Temporal/PlainDateTime/prototype/toZonedDateTime/disambiguate-basic.js
jessealama: Good idea -- done.

test/built-ins/Temporal/PlainDateTime/prototype/toZonedDateTime/invalid-instant.js
```diff
@@ -0,0 +1,16 @@
+// Copyright (C) 2022 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.plaindatetime.prototype.tozoneddatetime
+description: Convert to Instant that falls outside valid range
```
ptomato, line 6: This description is not quite right, it's converting to ZonedDateTime

test/built-ins/Temporal/PlainDateTime/prototype/toZonedDateTime/invalid-instant.js
```diff
@@ -0,0 +1,16 @@
...
+features: [Temporal]
+---*/
+
+const max = new Temporal.PlainDateTime(275760, 9, 13, 23, 59, 59, 999, 999, 999);
```
ptomato, line 10: I'd also test the `min` value here.

test/built-ins/Temporal/PlainDateTime/prototype/toZonedDateTime/disambiguate-basic.js
```diff
@@ -0,0 +1,26 @@
+// Copyright (C) 2022 Igalia, S.L. All rights reserved.
```
ptomato, line 1: This test seems to duplicate half of the multiple-instants.js test. I think it's probably OK to drop this one.

test/built-ins/Temporal/PlainDateTime/prototype/toZonedDateTime/basic.js
```diff
@@ -0,0 +1,14 @@
...
+/*---
+esid: sec-temporal.plaindatetime.prototype.tozoneddatetime
+description: Straightforward case of using an IANA timezone
+features: [Temporal]
+---*/
+
+const dt = new Temporal.PlainDateTime(2020, 1, 1, 0, 0);
+const zdt = dt.toZonedDateTime("UTC");
+
+assert.sameValue(zdt.epochNanoseconds, 1577836800000000000n, "works");
```
ptomato, line 13: Here's one place where I think it'd be good to test the time zone.

ptomato (approved):
Much improved, thanks. I still have a few suggestions but they're pretty minor.

test/built-ins/Temporal/PlainDateTime/prototype/toZonedDateTime/options-object.js
ptomato: I only saw one place where I'd recommend testing the time zone. I added a comment marking it in my latest review.

test/built-ins/Temporal/PlainDateTime/prototype/toZonedDateTime/basic.js
jessealama: Ah, I see that intuition now -- done.

test/built-ins/Temporal/PlainDateTime/prototype/toZonedDateTime/disambiguate-basic.js
jessealama: Good point. Dropped.

test/built-ins/Temporal/PlainDateTime/prototype/toZonedDateTime/invalid-instant.js
jessealama: Good idea, done.

test/built-ins/Temporal/PlainDateTime/prototype/toZonedDateTime/invalid-instant.js
jessealama: Oops, this was just copied over -- on auto-pilot -- from the Demitasse test description. I've updated it.

Ms2ger (approved):

PR #3497, opened 2022-04-22
Temporal: `PlainDateTime`: Port Demitasse tests for `getIsoFields`

jessealama:


jessealama:
Goes along with [this](https://github.com/tc39/proposal-temporal/pull/2148) PR on the proposal-temporal side.

ptomato (commented):
I think that https://github.com/tc39/test262/blob/main/test/built-ins/Temporal/PlainDateTime/prototype/getISOFields/field-prop-desc.js already tests that the properties are enumerable. So we can probably just drop this one.

PR #3498, opened 2022-04-22
Temporal: `PlainDateTime`: Port Demitasse tests for bounds

jessealama:
Goes along with [this](https://github.com/tc39/proposal-temporal/pull/2149) PR on the proposal-temporal side.

### DIFF
--- a/test/built-ins/Temporal/PlainDate/prototype/toPlainDateTime/limits.js
+++ b/test/built-ins/Temporal/PlainDate/prototype/toPlainDateTime/limits.js
@@ -1,0 +1,39 @@
+// Copyright (C) 2022 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.plaindate.prototype.toplaindatetime
+description: Checking limits of representable PlainDateTime
+features: [Temporal]
+includes: [temporalHelpers.js]
+---*/
+
+const midnight = new Temporal.PlainTime(0, 0);
+const firstNs = new Temporal.PlainTime(0, 0, 0, 0, 0, 1);
+const lastNs = new Temporal.PlainTime(23, 59, 59, 999, 999, 999);
+const min = new Temporal.PlainDate(-271821, 4, 19);
+const max = new Temporal.PlainDate(275760, 9, 13);
+
+assert.throws(
+  RangeError,
+  () => min.toPlainDateTime(midnight),
+  "Cannot go below representable limit for PlainDateTime"
+);
+
+TemporalHelpers.assertPlainDateTime(
+  max.toPlainDateTime(midnight),
+  275760, 9, "M09", 13, 0, 0, 0, 0, 0, 0,
+  "Midnight on maximal representable PlainDate"
+);
+
+TemporalHelpers.assertPlainDateTime(
+  min.toPlainDateTime(firstNs),
+  -271821, 4, "M04", 19, 0, 0, 0, 0, 0, 1,
+  "Computing the minimum (earliest) representable PlainDateTime"
+);
+
+TemporalHelpers.assertPlainDateTime(
+  max.toPlainDateTime(lastNs),
+  275760, 9, "M09", 13, 23, 59, 59, 999, 999, 999,
+  "Computing the maximum (latest) representable PlainDateTime"
+);
--- a/test/built-ins/Temporal/PlainDateTime/from/limits.js
+++ b/test/built-ins/Temporal/PlainDateTime/from/limits.js
@@ -1,0 +1,70 @@
+// Copyright (C) 2022 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.plaindatetime.from
+description: Checking limits of representable PlainDateTime
+features: [Temporal]
+includes: [temporalHelpers.js]
+---*/
+
+["reject", "constrain"].forEach((overflow) => {
+  assert.throws(
+    RangeError,
+    () => Temporal.PlainDateTime.from({ year: -271821, month: 4, day: 19 }, { overflow }),
+    `negative out of bounds (plain object, overflow = ${overflow})`
+  );
+  assert.throws(
+    RangeError,
+    () => Temporal.PlainDateTime.from({ year: 275760, month: 9, day: 14 }, { overflow }),
+    `positive out of bounds (plain object, overflow = ${overflow})`
+  );
+});
+
+TemporalHelpers.assertPlainDateTime(
+  Temporal.PlainDateTime.from({ year: -271821, month: 4, day: 19, nanosecond: 1 }),
+  -271821, 4, "M04", 19, 0, 0, 0, 0, 0, 1,
+  "construct from property bag (negative boundary)"
+);
+
+TemporalHelpers.assertPlainDateTime(
+  Temporal.PlainDateTime.from(
+    {
+      year: 275760,
+      month: 9,
+      day: 13,
+      hour: 23,
+      minute: 59,
+      second: 59,
+      millisecond: 999,
+      microsecond: 999,
+      nanosecond: 999
+    }
+  ),
+  275760, 9, "M09", 13, 23, 59, 59, 999, 999, 999,
+  "construct from property bag (positive boundary)"
+);
+
+assert.throws(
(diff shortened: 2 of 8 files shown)


test/built-ins/Temporal/PlainDate/prototype/toPlainDateTime/limits.js
```diff
@@ -0,0 +1,51 @@
...
+  RangeError,
+  () => min.toPlainDateTime(midnight),
+  'converting from Date and Time (1)'
+);
+
+assert.throws(
+  RangeError,
+  () => midnight.toPlainDateTime(min),
+  'converting from Date and Time (2)'
+);
```
ptomato, line 27: I think some of these tests (like this one) should go into a file in PlainTime/prototype/toPlainDateTime that mirrors this file.

test/built-ins/Temporal/PlainDate/prototype/toPlainDateTime/limits.js
```diff
@@ -0,0 +1,51 @@
+// Copyright (C) 2022 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.plaindatetime.prototype.toplaindatetime
+description: Numeric bounds check
```
ptomato, line 6: I think these tests could have a better description, like "Limits of representable range" (there's nothing particularly numeric about the check)

test/built-ins/Temporal/PlainDate/prototype/toPlainDateTime/limits.js
```diff
@@ -0,0 +1,51 @@
...
+const midnight = new Temporal.PlainTime(0, 0);
+const firstNs = new Temporal.PlainTime(0, 0, 0, 0, 0, 1);
+const lastNs = new Temporal.PlainTime(23, 59, 59, 999, 999, 999);
+const min = new Temporal.PlainDate(-271821, 4, 19);
+const max = new Temporal.PlainDate(275760, 9, 13);
+
+assert.throws(
+  RangeError,
+  () => min.toPlainDateTime(midnight),
+  'converting from Date and Time (1)'
```
ptomato, line 20: I'd recommend writing more expressive assertion messages so that we know what each assertion is testing for.

test/built-ins/Temporal/PlainDateTime/from/limits.js
```diff
@@ -0,0 +1,70 @@
+// Copyright (C) 2022 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.plaindatetime.from
+description: Numeric bounds check
```
ptomato, line 6: Here you could additionally say that we are testing that going outside the representable range throws regardless of the `overflow` option (but more concisely than I said it here :stuck_out_tongue:)

test/built-ins/Temporal/PlainDateTime/limits.js
```diff
@@ -0,0 +1,35 @@
+// Copyright (C) 2022 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.plaindatetime
+description: Numeric bounds check for constructor (string, plain object)
```
ptomato, line 6: "string, plain object" doesn't seem to apply here.

test/built-ins/Temporal/PlainDateTime/limits.js
```diff
@@ -0,0 +1,35 @@
...
+    RangeError,
+    () => min.round({ smallestUnit, roundingMode: 'floor' }),
+    `rounding beyond limit (unit = ${smallestUnit}, rounding mode = floor)`
+  );
+  assert.throws(
+    RangeError,
+    () => max.round({ smallestUnit, roundingMode: 'ceil' }),
+    `rounding beyond limit (unit = ${smallestUnit}, rounding mode = ceil)`
+  );
+});
```
ptomato, line 35: I think this duplicates the PlainDateTime/prototype/round test below.

test/built-ins/Temporal/PlainDateTime/limits.js
```diff
@@ -0,0 +1,35 @@
...
+  'negative year out of bounds'
+);
+assert.throws(
+  RangeError,
+  () => new Temporal.PlainDateTime(275760, 9, 14, 0, 0, 0, 0, 0, 0),
+  'positive year out of bounds'
+);
+
+const min = new Temporal.PlainDateTime(-271821, 4, 19, 0, 0, 0, 0, 0, 1);
+const max = new Temporal.PlainDateTime(275760, 9, 13, 23, 59, 59, 999, 999, 999);
```
ptomato, line 22: I'd use `TemporalHelpers.assertPlainDateTime` on these, as you did above in the PlainDateTime/from test.

test/built-ins/Temporal/PlainDateTime/prototype/add/limits.js
```diff
@@ -0,0 +1,19 @@
...
+
+const min = new Temporal.PlainDateTime(-271821, 4, 19, 0, 0, 0, 0, 0, 1);
+const max = new Temporal.PlainDateTime(275760, 9, 13, 23, 59, 59, 999, 999, 999);
+
+['reject', 'constrain'].forEach((overflow) => {
+  assert.throws(
+    RangeError,
+    () => max.add({ nanoseconds: 1 }, { overflow }),
+    `adding beyond limit (overflow = ${overflow})`
+  );
```
ptomato, line 18: I'd add another test here adding `{ nanoseconds: -1 }` to `min`.

test/built-ins/Temporal/PlainDateTime/prototype/subtract/limits.js
```diff
@@ -0,0 +1,19 @@
...
+
+const min = new Temporal.PlainDateTime(-271821, 4, 19, 0, 0, 0, 0, 0, 1);
+const max = new Temporal.PlainDateTime(275760, 9, 13, 23, 59, 59, 999, 999, 999);
+
+['reject', 'constrain'].forEach((overflow) => {
+  assert.throws(
+    RangeError,
+    () => min.subtract({ nanoseconds: 1 }, { overflow }),
+    `subtracting beyond limit (overflow = ${overflow})`
+  );
```
ptomato, line 18: Same here, I'd add another test subtracting `{ nanoseconds: -1 }` from `max`.

test/built-ins/Temporal/TimeZone/prototype/getPlainDateTimeFor/limits.js
```diff
@@ -0,0 +1,27 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.plaindatetime.prototype.getplaindatetimefor
+description: Numeric bounds check
+features: [Temporal]
+includes: [temporalHelpers.js]
+---*/
+
+const min = Temporal.Instant.from('-271821-04-20T00:00Z');
```
ptomato, line 11: I'd prefer to use `new Temporal.Instant(-8_640_000_000_000_000_000_000n)` and `new Temporal.TimeZone("-23:59")` to avoid coupling with from() where it's not necessary.

Also I think since that test was originally written a few years ago, it's now possible to have `new Temporal.TimeZone("-23:59.999999999")` so it might be better to use that here. The purpose of this test is to show that the representable range of PlainDateTime is enough to cover any possible combination of Instant and TimeZone, so it seems better to me to use the most extreme value down to the nanosecond.

ptomato (changes requested):
Thanks!

test/built-ins/Temporal/PlainDate/prototype/toPlainDateTime/limits.js
jessealama: Quite right -- I thought this was the right place, but indeed some of the tests don't belong here. I've moved those to the right place.

test/built-ins/Temporal/PlainDate/prototype/toPlainDateTime/limits.js
jessealama: Good spot -- I'll fix the descriptions. Indeed it's not about the limits of numbers qua numbers, but about domain limits. I'll update the descriptions.

Ms2ger (approved):

Issue #3305, opened 2021-11-11
Tests for Intl DurationFormat  API 


romulocintra:
Hi, opening this issue as a placeholder/reminder for the Intl.DurationFormat tests that advanced to Stage 3 in TC39 2021 October meeting.

https://github.com/tc39/proposal-intl-duration-format

Checklist 

- [x] Basic checks (name, length, prop-desc, slot)
- [x] Intl.DurationFormat.prototype.constructor
- [x]  Intl.DurationFormat.prototype [ @@toStringTag ]
- [x]  Intl.DurationFormat.prototype.format
- [x] Intl.DurationFormat.prototype.formatToParts 
- [x]  Intl.DurationFormat.prototype.resolvedOptions
- [ ] ...

jugglinmike:
Thanks, Romulo! If anyone is interested in writing these tests, please comment here. That will help us avoid duplicating effort.

EarthyOrange:
@jugglinmike I can help out with this. 

- Is this feature already implemented (partially?) in any engine available under esvu which I can use to write tests?
- If so then is there a flag under which it is being hidden?
- If not then should I just add (failing) tests based on the proposal?

EarthyOrange:
@jugglinmike Can you please take a look at the PR?

FrankYFTang:
intl402/DurationFormat/constructor-options-defaults.js need to be changed to sync with the changes landed in https://github.com/tc39/proposal-intl-duration-format/pull/185 @ben-allen

ryzokuken:
Can we close this? We already have the tests merged at https://github.com/tc39/test262/tree/main/test/intl402/DurationFormat.

PR #3501, opened 2022-04-25
Intl.DurationFormat update coverage

romulocintra:
Progressively adding a few more tests to Intl.Duration Format to improve ctor coverage
 
cc @sffc 

#3305


test/intl402/DurationFormat/constructor/constructor/locales-invalid.js
```diff
@@ -0,0 +1,17 @@
+// Copyright 2022 Igalia, S.L. All rights reserved.
```
ptomato, line 1: Looking at the structure of existing tests elsewhere in Intl, e.g. for DateTimeFormat, it seems like these tests should go directly in the `test/intl402/DurationFormat` directory, without the `constructor/constructor/`. See the thread https://github.com/tc39/test262/pull/3406#discussion_r809537160

test/intl402/DurationFormat/constructor/constructor/options-localeMatcher-valid.js
```diff
@@ -0,0 +1,25 @@
...
+const validOptions = [
+  undefined,
+  'lookup',
+  'best fit'
+];
+
+for (const localeMatcher of validOptions) {
+  const obj = new Intl.DurationFormat('en', { localeMatcher });
+
+  assert.sameValue(Object.getPrototypeOf(obj), Intl.DurationFormat.prototype, `proto check - ${localeMatcher}`);
```
ptomato, line 23: I'm guessing this is a stand-in for lack of a better assertion that says "constructing the object didn't throw"?
Is it possible to check this option via `obj.resolvedOptions()` instead?

ptomato, line 23: I'm guessing this is a stand-in for lack of a better assertion that says "constructing the object didn't throw"?

test/intl402/DurationFormat/constructor/supportedLocalesOf/basic.js
```diff
@@ -0,0 +1,19 @@
...
+
+assert.sameValue(typeof Intl.DurationFormat.supportedLocalesOf, "function",
+                 "supportedLocalesOf should be supported.");
+
+const defaultLocale = new Intl.DurationFormat().resolvedOptions().locale;
+const notSupported = "zxx"; // "no linguistic content"
+const requestedLocales = [defaultLocale, notSupported];
+
+const supportedLocales = Intl.DurationFormat.supportedLocalesOf(requestedLocales);
+assert.sameValue(supportedLocales.length, 1, "The length of supported locales list is not 1.");
```
ptomato, line 18: These assertion messages seem unusual compared to what I've seen elsewhere in test262. I think "The length of the supported locales list should be 1" would be more usual. Same for the message just below this one.

ptomato (commented):
Thanks!

test/intl402/DurationFormat/constructor-options-order.js
```diff
@@ -0,0 +1,41 @@
...
+    actual.push("style");
+    return undefined;
+  },
+};
+
+const expected = [
+  "localeMatcher",
+  "numberingSystem",
+  "style"
+];
```
ptomato, line 38: Optional: There is a helper for testing the order of operations of property gets, although maybe it should get a more general place than TemporalHelpers if we're going to use it here as well.

```suggestion
const options = {};
TemporalHelpers.observeProperty(actual, options, "localeMatcher", undefined);
TemporalHelpers.observeProperty(actual, options, "numberingSystem", undefined);
TemporalHelpers.observeProperty(actual, options, "style", undefined);

const expected = [
  "get localeMatcher",
  "get numberingSystem",
  "get style"
];
```

test/intl402/DurationFormat/supportedLocalesOf/basic.js
```diff
@@ -0,0 +1,19 @@
...
+assert.sameValue(typeof Intl.DurationFormat.supportedLocalesOf, "function",
+                 "supportedLocalesOf should be supported.");
+
+const defaultLocale = new Intl.DurationFormat().resolvedOptions().locale;
+const notSupported = "zxx"; // "no linguistic content"
+const requestedLocales = [defaultLocale, notSupported];
+
+const supportedLocales = Intl.DurationFormat.supportedLocalesOf(requestedLocales);
+assert.sameValue(supportedLocales.length, 1, "The length of the supported locales list should be 1");
+assert.sameValue(supportedLocales[0], defaultLocale, "The default locale is not returned in the supported list.");
```
ptomato, line 19:
```suggestion
assert.sameValue(supportedLocales[0], defaultLocale, "The default locale is returned in the supported list.");
```
?

ptomato (dismissed):
LGTM, one optional improvement (up to you to decide) and one wording nitpick

test/intl402/DurationFormat/constructor-options-order.js
romulocintra: That's nice, I'm ok with using it, no strong preference at moment, only  the "TemporalHelpers" name, would be really nice to have it as generic utils functions and documented somewhere else cause they're really handy

ptomato (approved):

PR #3503, opened 2022-04-26
Temporal: Port a few tests for Duration#{add,subtract}.

Ms2ger:


ptomato (approved):
In general I'd like to have meaningful assertion messages everywhere, and avoid coupling with from() methods. But I don't have any other comments on the tests, they all look correct to me, so no need to hold up the PR on that.

PR #3504, opened 2022-04-26
use `isConstructor` assertion in "non-constructor" tests

ljharb:
A few of the "not constructor" tests weren't using isConstructor, and thus lacked the `Reflect.construct` feature flag.

At first I tried to make a new "non-constructor" feature flag, but it would change hundreds of files - and since these tests were missing an `isConstructor` precondition assertion anyways, this seemed more expedient.

### DIFF
--- a/test/built-ins/Array/prototype/concat/create-species-non-ctor.js
+++ b/test/built-ins/Array/prototype/concat/create-species-non-ctor.js
@@ -18,8 +18,15 @@ info: |
        b. If C is null, let C be undefined.
     [...]
     9. If IsConstructor(C) is false, throw a TypeError exception.
-features: [Symbol.species]
+includes: [isConstructor.js]
+features: [Symbol.species, Reflect.construct]
 ---*/
+
+assert.sameValue(
+  isConstructor(parseInt),
+  false,
+  'precondition: isConstructor(parseInt) must return false'
+);
 
 var a = [];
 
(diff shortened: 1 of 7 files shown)


test/built-ins/Array/prototype/concat/create-species-non-ctor.js
```diff
@@ -18,9 +18,16 @@ info: |
...
     9. If IsConstructor(C) is false, throw a TypeError exception.
-features: [Symbol.species]
+includes: [isConstructor.js]
+features: [Symbol.species, Reflect.construct]
 ---*/
 
+assert.sameValue(
+  isConstructor(parseInt),
+  false,
+  'precondition: isConstructor(parseInt) must return false'
```
gibson042, line 28: Would it make more sense to replace `parseInt` with a [[Construct]]-free function that can't be overridden, ideally defined in a harness/ helper?
```js
var nonConstructorCallable = (function() {
  "use strict";
  try {
    // https://tc39.es/ecma262/#sec-createunmappedargumentsobject
    return Object.getOwnPropertyDescriptor(arguments, "callee").get;
  } catch (unused) {
    // Absent strict mode, presume a built-in `Function`
    // and use its non-writable non-configurable "prototype" property.
    // https://tc39.es/ecma262/#sec-properties-of-the-function-constructor
    return Function.prototype;
  }
})()
```

ljharb: Sure, that would also work for my purposes - but even then i still think the precondition assert is useful.

PR #3505, opened 2022-04-27
Temporal: Port some Duration.compare tests.

Ms2ger:


test/intl402/Temporal/Duration/compare/relativeto-hour.js
```diff
@@ -0,0 +1,31 @@
...
+  Temporal.Duration.compare(oneDay, hours24, { relativeTo: '2019-11-03T00:00[America/Vancouver]' }),
+  1,
+  'casts relativeTo to ZonedDateTime from string');
+assert.sameValue(
+  Temporal.Duration.compare(oneDay, hours24, {
+    relativeTo: { year: 2019, month: 11, day: 3, timeZone: 'America/Vancouver' }
+  }),
+  1,
+  'casts relativeTo to ZonedDateTime from object');
+
```
ptomato, line 31: I think we need to have these tests in the `built-ins/` suite as well. I would suggest using `TemporalHelpers.springForwardFallBackTimeZone()` to create a time zone that has a DST change, in a way that hosts without 402 can execute. For the first assertion where the duration encompasses no DST change, you could just use UTC instead of America/Montevideo.

ptomato (changes requested):

test/intl402/Temporal/Duration/compare/relativeto-hour.js
Ms2ger: Are the tests in [twenty-five-hour-day.js](https://github.com/tc39/test262/blob/28b31c0bf1960878abb36ab8597a0cae224a684d/test/built-ins/Temporal/Duration/compare/twenty-five-hour-day.js) sufficient?

ptomato: They do look like they cover somewhat the same thing. I don't know if they cover the case where we get a different answer out of Duration.compare() depending on whether the string or property bag value for `relativeTo` converts to a PlainDate or ZonedDateTime.

Ms2ger: I think you'll be able to write that test more quickly than me :)

ptomato: Fair enough, I'll include it in my next PR.

ptomato (approved):

PR #3506, opened 2022-04-28
Fix Intl.NumberFormat roundingPriority test to reflect default values for min/max digits

sffc:
@FrankYFTang @romulocintra


romulocintra (approved):

ptomato (commented):
I'm not sure I get what is going on in this change, some context in the commit message would be helpful.

FrankYFTang:
LGTM

sffc:
> I'm not sure I get what is going on in this change, some context in the commit message would be helpful.

I updated the PR description so if this gets squash-merged then it will update the commit message

ptomato:
@romulocintra and I have verified by walking through the spec steps that this test is correct according to the spec. It's still not clear to me from reading the documentation why I would expect this result though, so I opened https://github.com/tc39/proposal-intl-numberformat-v3/issues/96

PR #3510, opened 2022-05-02
durationformat.format coverage

romulocintra:
Incremental  coverage for DurationFormat.format

cc @FrankYFTang @ryzokuken @sffc


test/intl402/DurationFormat/prototype/format/en-us-style-options.js
```diff
@@ -0,0 +1,40 @@
...
+locale: [en-US]
+---*/
+
+const stylesList = ["long" , "short" , "narrow", "digital"];
+
+const assertList = {
+  "PT1H50M":  {
+    "long": "1 hour and 50 minutes",
+    "short": "1 hr, 50 min",
+    "narrow": "1h 50m",
```
romulocintra, line 17: Not sure if `hr` or `hr.`

test/intl402/DurationFormat/prototype/format/branding.js
```diff
@@ -0,0 +1,22 @@
...
+esid: sec-Intl.DurationFormat.prototype.format
+description: Verifies the branding check for the "format" function of the DurationFormat prototype object.
+features: [Intl.DurationFormat]
+---*/
+
+const format = Intl.DurationFormat.prototype.format;
+
+assert.sameValue(typeof format, "function");
+
+assert.throws(TypeError, () => format.call(undefined), "undefined");
```
ptomato, line 14: I learned from https://github.com/tc39/test262/issues/3224 that this kind of test should call the method with otherwise correct arguments.

e.g.
```suggestion
assert.throws(TypeError, () => format.call(undefined, duration), "undefined");
```
(with some `duration` defined above)

test/intl402/DurationFormat/prototype/format/en-us-style-options.js
```diff
@@ -0,0 +1,40 @@
...
+    "digital": "1:50:00",
+  },
+  "PT2H46M40S":  {
+    "long": "2 hours, 46 minutes and 40 seconds",
+    "short": "2 hr, 46 min, 40 sec",
+    "narrow": "2h 46m 40s",
+    "digital": "2:46:40",
+  },
+  // Negative durations in Intl.DurationFormat https://github.com/tc39/proposal-intl-duration-format/issues/29
```
ptomato, line 26: Is the idea that these tests will be added later? From my reading of the issue, it says that tests should verify that the output is different between positive and negative. (Or should that be a separate test that isn't dependent on locale?)

In any case, I think the tests will currently fail as written because `stylelist` includes the keys that are missing from these objects? Maybe it's better to have something like
```js
if (!(style in styletestData))
  continue;
```

test/intl402/DurationFormat/prototype/format/invalid-arguments-throws.js
```diff
@@ -0,0 +1,21 @@
...
+  3. Let record be ? ToPartialDuration(duration).
+---*/
+
+const df = new Intl.DurationFormat();
+
+assert.throws(TypeError, () => { df.format(undefined) });
+assert.throws(TypeError, () => { df.format(-12) });
+assert.throws(TypeError, () => { df.format(-12n) });
+assert.throws(TypeError, () => { df.format(1) });
+assert.throws(TypeError, () => { df.format(2n) });
```
ptomato, line 20: Assuming https://github.com/tc39/proposal-intl-duration-format/pull/92 gets merged, these will be `RangeError`s because they will convert to a string which is not a valid ISO 8601 duration string.

I'd also suggest testing things like `{}`, `null`, `true`, and `"bad string"`.

ptomato (changes requested):
Thanks!

test/intl402/DurationFormat/prototype/format/en-us-style-options.js
sffc: Do we normally put the localized strings in the Test262 suite? Adding and removing `.` is a common thing that changes across CLDR updates

romulocintra: There are some cases at relative time format : 

https://github.com/tc39/test262/blob/main/test/intl402/RelativeTimeFormat/prototype/format/en-us-numeric-auto.js#L29 

I'm happy to only check the resolvedOptions of those methods instead avoiding depending on CLDR

sffc: I was asking about the general policy of Test262 and CLDR-dependent output. I think it's useful to have at least a few tests for the expected output of the format function, but we just need to be aware that these tests are data-dependent, and the expectations could drift.

Ms2ger: This is what the `locale` key in the front matter is for

PR #3512, opened 2022-05-02
Decorators: add initial tests on class

legendecas:
This PR adds initial decorator syntax tests on:
- ClassDeclaration
- ClassExpression
- ClassElement
- FieldDefinition accessor

### DIFF
--- a/src/class-elements/field-definition-accessor-no-line-terminator.case
+++ b/src/class-elements/field-definition-accessor-no-line-terminator.case
@@ -1,0 +1,28 @@
+// Copyright (C) 2022 Chengzhong Wu. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+desc: Valid accessor FieldDefinition, ClassElementName, PropertyName Syntax
+info: |
+  FieldDefinition[Yield, Await] :
+    accessor [no LineTerminator here] ClassElementName[?Yield, ?Await] Initializer[+In, ?Yield, ?Await]opt
+
+template: default
+features: [decorators]
+---*/
+
+//- elements
+accessor
+$;
+static accessor
+$;
+
+//- assertions
+let c = new C();
+
+assert.sameValue(Object.hasOwnProperty.call(C.prototype, 'accessor'), false);
+assert.sameValue(Object.hasOwnProperty.call(C.prototype, '$'), false);
+assert.sameValue(Object.hasOwnProperty.call(C, 'accessor'), true);
+assert.sameValue(Object.hasOwnProperty.call(C, '$'), true);
+assert.sameValue(Object.hasOwnProperty.call(c, 'accessor'), true);
+assert.sameValue(Object.hasOwnProperty.call(c, '$'), true);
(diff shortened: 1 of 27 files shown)


src/class-elements/field-definition-accessor-no-line-terminator.case
```diff
@@ -0,0 +1,28 @@
...
+
+//- assertions
+let c = new C();
+
+assert.sameValue(Object.hasOwnProperty.call(C.prototype, 'accessor'), false);
+assert.sameValue(Object.hasOwnProperty.call(C.prototype, '$'), false);
+assert.sameValue(Object.hasOwnProperty.call(C, 'accessor'), true);
+assert.sameValue(Object.hasOwnProperty.call(C, '$'), true);
+assert.sameValue(Object.hasOwnProperty.call(c, 'accessor'), true);
+assert.sameValue(Object.hasOwnProperty.call(c, '$'), true);
```
rwaldron, line 28: 🥳

rwaldron: I see that this `Object.hasOwnProperty.call` exists elsewhere, but I think that was a mistake and the intention was actually to use `Object.prototype.hasOwnProperty.call`. I'm not going to block on it here, but someone should do a cleanup to use `Object.prototype.hasOwnProperty.call` or `Reflect.hasOwnProperty`.

rwaldron (approved):
This is a great initial set!!

src/class-elements/field-definition-accessor-no-line-terminator.case
legendecas: I can do it in a follow-up PR if you are ok with it.

src/decorator/decorator-call-expr-identifier-reference.case
```diff
@@ -0,0 +1,44 @@
...
+var _ = decorator;
+var \u{6F} = decorator;
+var \u2118 = decorator;
+var ZW_\u200C_NJ = decorator;
+var ZW_\u200D_J = decorator;
+var yield = decorator;
+var await = decorator;
+
+//- decorators
+@$()
```
iamstolis, line 37: This test seems to be incorrect because it does not define this (`$()`) decorator.

test/language/expressions/class/elements/field-definition-accessor-no-line-terminator.js
```diff
@@ -0,0 +1,31 @@
...
+  $;
+
+}
+
+let c = new C();
+
+assert.sameValue(Object.hasOwnProperty.call(C.prototype, 'accessor'), false);
+assert.sameValue(Object.hasOwnProperty.call(C.prototype, '$'), false);
+assert.sameValue(Object.hasOwnProperty.call(C, 'accessor'), true);
+assert.sameValue(Object.hasOwnProperty.call(C, '$'), true);
```
iamstolis, line 29: Class `C` is basically `class { accessor; $; static accessor; $; }` due to automatic semicolon insertion (and the fact `FieldDefinition` of an accessor cannot contain `LineTerminator`). So, there is not static `$` field. In other words, this line should expect `false` (not `true`).

src/decorator/decorator-call-expr-identifier-reference.case
legendecas: Thank you for raising these! I've submitted https://github.com/tc39/test262/pull/3585 to fix these.

Issue #3508, opened 2022-04-30
Intl.NumberFormat v3: Improve coverage of SetNumberFormatDigitOptions


sffc:
SetNumberFormatDigitOptions is a complicated part of the spec that we should ensure that implementations get right.

In particular:

1. Ensure that all combinations of the presence of `maximumFractionDigits`, `maximumSignificantDigits`, `roundingPriority: "morePrecision"`, `roundingPriority: "lessPrecision"`, and `notation: "compact"` are tested and that they produce the expected resolvedOptions.
2. Migrate as much of @gibson042's excellent test suite as possible; see https://github.com/tc39/ecma402/pull/575#pullrequestreview-730209467

CC @romulocintra @FrankYFTang

romulocintra:
> 1. Ensure that all combinations of the presence of `maximumFractionDigits`, `maximumSignificantDigits`, `roundingPriority: "morePrecision"`, `roundingPriority: "lessPrecision"`, and `notation: "compact"` are tested and that they produce the expected resolvedOptions.

Started a Draft [here](https://github.com/tc39/test262/pull/3515) combining some of the values :
Todo :

- [ ] Test notation 
- [ ]  Add correct results

PR #3515, opened 2022-05-02
Tests combinations max-min fractionDigits and Significant digits

romulocintra:
https://github.com/tc39/test262/issues/3508#issue-1221898118

Todo's  for other tests files:

- [ ] Combine "notation" : "compact" with all other values
- [ ] validate  resolvedOptions for each set of options

The goal's to have correct matrix results aligned with spec

cc @sffc @FrankYFTang

### DIFF
--- a/test/intl402/NumberFormat/test-option-roundingPriority-mixed-options.js
+++ b/test/intl402/NumberFormat/test-option-roundingPriority-mixed-options.js
@@ -1,0 +1,80 @@
+// Copyright (C) 2022 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+  /*---
+  esid: sec-intl.numberformat.prototype.format
+  description: Tests that the digits are determined correctly when specifying at same time «"minimumFractionDigits", "maximumFractionDigits", "minimumSignificantDigits", "maximumSignificantDigits"»
+  features: [Intl.NumberFormat-v3]
+  includes: [testIntl.js]
+  ---*/
+
+  var locales = [new Intl.NumberFormat().resolvedOptions().locale, "ar", "de", "th", "ja"];
+
+  var numberingSystems =  ['arab', 'latn', 'thai', 'hanidec'];
+
+  var expectedResults = {
+    'auto' : [
+      { '1': '1.0', '1.500': '1.5', '1.625': '1.6', '1.750': '2.0', '1.875': '2.0', '2.000': '2.0' },
+      { '1': '1.0', '1.500': '1.5', '1.625': '1.6', '1.750': '2.0', '1.875': '2.0', '2.000': '2.0' },
+      { '1': '1.0', '1.500': '1.5', '1.625': '1.6', '1.750': '2.0', '1.875': '2.0', '2.000': '2.0' },
+      { '1': '1.0', '1.500': '1.5', '1.625': '1.6', '1.750': '2.0', '1.875': '2.0', '2.000': '2.0' },
+      { '1': '1.0', '1.500': '1.5', '1.625': '1.6', '1.750': '2.0', '1.875': '2.0', '2.000': '2.0' },
+      { '1': '1.0', '1.500': '1.5', '1.625': '1.6', '1.750': '2.0', '1.875': '2.0', '2.000': '2.0' },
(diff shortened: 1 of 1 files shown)


test/intl402/NumberFormat/test-option-roundingPriority-mixed-options.js
```diff
@@ -0,0 +1,80 @@
...
+  description: Tests that the digits are determined correctly when specifying at same time «"minimumFractionDigits", "maximumFractionDigits", "minimumSignificantDigits", "maximumSignificantDigits"»
+  features: [Intl.NumberFormat-v3]
+  includes: [testIntl.js]
+  ---*/
+
+  var locales = [new Intl.NumberFormat().resolvedOptions().locale, "ar", "de", "th", "ja"];
+
+  var numberingSystems =  ['arab', 'latn', 'thai', 'hanidec'];
+
+  var expectedResults = {
```
sffc, line 15: Request: Please check the output of `resolvedOptions` as well.

ptomato (commented):
I think the test matrix approach looks good, I trust that the champions know best what the correct format outputs should be :smile: Currently they are all the same though, I assume that's because it's a draft and the real values will be entered later?

One improvement might be to use keys that explain what that particular line tests. e.g.
```js
var optionsMatrix = {
  'same-minimums': { useGrouping: false, minimumSignificantDigits: 2, minimumFractionDigits: 2 },
  'min-sd-larger': { useGrouping: false, minimumSignificantDigits: 3, minimumFractionDigits: 1 },
  ...
```
and 
```js
'auto': {
  'same-minimums': { '1': '1.0', ...
```

test/intl402/NumberFormat/test-option-roundingPriority-mixed-options.js
romulocintra: I will address this later on this same test or in a different set of tests

ptomato (approved):
I think this is basically correct — I didn't check the rounding results exhaustively, but I sampled maybe about 40% of them.

If you wanted to add some comments to explain what is going on in each entry, I'm sure that would really help implementors debug their implementations. For example, in the expected results matrix,
```js
'morePrecision': {
  // 2-21 significant digits
  'same-minimums': { ... },
  // 1-2 significant digits or 0-2 fraction digits, whichever is more precise
  'same-maximums': { ... },
  // 3-21 significant digits
  'minSd-larger-minFd': { ... },
```
and in the options matrix, for example,
```js
// max significant digits wins
'same-minimums': { ... },
// max digits equal, result with more precision wins
'same-maximums': { ... },
```
etc.

test/intl402/NumberFormat/test-option-roundingPriority-mixed-options.js
```diff
@@ -0,0 +1,94 @@
...
+      'minSd-maxSd-smaller-minFd-maxFd' :{ '1': '1', '1.500': '1.5', '1.625': '1.6', '1.750': '1.8', '1.875': '1.9', '2.000': '2' },
+      'minSd-maxSd-larger-minFd-maxFd' :{ '1': '1.00', '1.500': '1.50', '1.625': '1.625', '1.750': '1.75', '1.875': '1.875', '2.000': '2.00' },
+      'same-minimums-maxFd-larger-maxSd': { '1': '1', '1.500': '1.5', '1.625': '1.6', '1.750': '1.8', '1.875': '1.9', '2.000': '2' },
+      'same-minimums-maxFd-smaller-maxSd': { '1': '1', '1.500': '1.5', '1.625': '1.625', '1.750': '1.75', '1.875': '1.875', '2.000': '2' }
+  }
+}
+
+const optionsMatrix = {
+  'same-minimums' :{ minimumSignificantDigits: 2, minimumFractionDigits: 2},
+  'same-maximums' :{ maximumSignificantDigits: 2, maximumFractionDigits: 2},
```
ljharb, line 68:
```suggestion
  'same-maximums': { maximumSignificantDigits: 2, maximumFractionDigits: 2},
```
can the colons and braces be consistently spaced throughout?

sffc (approved):

ptomato (approved):
OK by me, I still think the comments I suggested would be useful if you have time to add them, but what you have is already fine.

Issue #3516, opened 2022-05-03
Problems of verifying era/eraYear in Temporal tests localted under test/built-ins


FrankYFTang:
All tests which part of ECMA402 should be put under test/intl402 and tests outside test/intl402 should not verify any behavior defined in ECMA402.

But in Temporal, because Temporal.*.prototype.era and eraYear are defined in Chapter 15
https://tc39.es/proposal-temporal/#sec-temporal-intl

Behavior of Temporal.*.prototype.era and eraYear should NOT be test and verify under any tests inside test/built-ins/Temporal. Testing and verifing value of Temporal.*.prototype.era and eraYear  should be moved under test/intl402/Temporal instead.

I found the following helper functions will test era and eraYear

assertPlainDate
https://github.com/tc39/test262/blob/main/harness/temporalHelpers.js#L62
assertPlainDateTime
https://github.com/tc39/test262/blob/main/harness/temporalHelpers.js#L80
assertPlainYearMonth
https://github.com/tc39/test262/blob/main/harness/temporalHelpers.js#L161

Or we dup the test and only verify era and eraYear on the tests under test/intl402/Temporal but not those test under test/built-ins/Temporal

@ptomato , @Ms2ger, @jessealama

ljharb:
I'm confused; are you saying these accessors *only* exist in Intl? If so, I think that's a mistake in the proposal; they should be defined in 262 and overridden in 402.

FrankYFTang:
> I'm confused; are you saying these accessors _only_ exist in Intl? 

Why don't you read the Temporal proposal yourself and decide is that case or not. 

>If so, I think that's a mistake in the proposal; they should be defined in 262 and overridden in 402.

My understanding is calling these getter should return undefined under 262 but may return different value under 402 as what currently stated in the Temporal.


Here are all the tests
$ find test/built-ins/Temporal/|egrep "js$"|xargs egrep "assertPlain(Date(Time)?|YearMonth)" |cut -d ':' -f1 |sort -u
test/built-ins/Temporal/Calendar/prototype/dateAdd/add-days.js
test/built-ins/Temporal/Calendar/prototype/dateAdd/add-months.js
test/built-ins/Temporal/Calendar/prototype/dateAdd/add-months-weeks.js
test/built-ins/Temporal/Calendar/prototype/dateAdd/add-weeks-days.js
test/built-ins/Temporal/Calendar/prototype/dateAdd/add-weeks.js
[...]

PR #3517, opened 2022-05-03
Move testing of era/eraYear under intl402

FrankYFTang:
Temporal.PlainDateTime.prototype.ear and earYear are not defined in first 14 chapters and only defined under Chapter 15. Therefore they should always return undefined unless it is an intl implementation.

Copy the same test to intl402 and remove the non undefined value of era/eraYear from the built-ints version of test. 

Fix https://github.com/tc39/test262/issues/3516 
Also see https://github.com/tc39/proposal-temporal/issues/2169

### DIFF
--- a/test/built-ins/Temporal/PlainDateTime/prototype/withPlainDate/argument-plaindate-calendar-noniso.js
+++ b/test/built-ins/Temporal/PlainDateTime/prototype/withPlainDate/argument-plaindate-calendar-noniso.js
@@ -26,9 +26,8 @@ const shifted = pdt.withPlainDate(pd);
 TemporalHelpers.assertPlainDateTime(
   shifted,
   2008, 9, "M09", 6, 3, 24, 30, 0, 0, 0,
-  "calendar is changed if receiver has ISO calendar (1)",
-  "the era",
-  1909
+  "calendar is changed if receiver has ISO calendar (1)"
+  // Testing of era and eraYear should only be coded under intl402
 );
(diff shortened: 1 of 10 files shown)


test/built-ins/Temporal/PlainDateTime/prototype/withPlainDate/argument-plaindate-calendar-noniso.js
```diff
@@ -26,9 +26,8 @@ const shifted = pdt.withPlainDate(pd);
 TemporalHelpers.assertPlainDateTime(
   shifted,
   2008, 9, "M09", 6, 3, 24, 30, 0, 0, 0,
-  "calendar is changed if receiver has ISO calendar (1)",
-  "the era",
-  1909
+  "calendar is changed if receiver has ISO calendar (1)"
+  // Testing of era and eraYear should only be coded under intl402
```
ptomato, line 30: In all these tests under `built-ins/` I'd recommend removing the `era` and `eraYear` properties from `cal` above (and then this comment would be redundant)

FrankYFTang: you are right, it won't work if I do not remove them, because then the implementation supporting 402 will return those but our code will assert it equals to undefined.

ptomato (approved):
Looks good, thanks.

PR #3519, opened 2022-05-04
Add missing binding

mathiasbynens:


ptomato (approved):
There is indeed a missing binding there. Thanks!

PR #3520, opened 2022-05-04
Turn unimportant tabs into spaces in S12.6.4_A7_T2.js

kmiller68:
This test currently has some tabs. Clean this up and make it use spaces instead.


ljharb (approved):
boo, wrong direction

PR #3525, opened 2022-05-12
Draft: RFC process documentation

ptomato:
This is a process allowing contributors to submit well-motivated
proposals to improve test262 using an RFC (Request For Commments)
process which is adapted from a similar process in the Rust community.


rfcs/process.md
```diff
@@ -0,0 +1,235 @@
...
+    to push back if it is made prematurely.
+  - For RFCs with lengthy discussion, the motion to FCP is usually
+    preceded by a summary comment trying to lay out the current state of
+    the discussion and major tradeoffs/points of disagreement.
+  - Before actually entering FCP, the maintainers group must sign off as
+    a whole.
+    This requires the usual "consensus minus one".
+
+- The FCP lasts ten calendar days, so that it is open for at least 5
+  business days.
```
gibson042, line 142: This claim could be invalidated by the inclusion of more than one holiday in any eight-day interval, which seems plausible. Perhaps the secondary clause should be removed or weakened?

Also, why is "ten" spelled with letters but "5" with a digit?

gibson042 (approved):

rfcs/0000-template.md
```diff
@@ -0,0 +1,104 @@
...
+
+# Prior art
+
+Discuss prior art, both the good and the bad, in relation to this
+proposal.
+A few examples of what this can include are:
+
+- For test harness proposals: Does this feature exist in other test
+  harnesses and what experience have their community had?
+  Is the feature generally considered a net positive?
```
ljharb, line 60: i have a mild preference for soft-wrapping instead of hard-wrapping; is there a reason to prefer hard-wrapping?

rfcs/process.md
```diff
@@ -0,0 +1,235 @@
...
+  or are disingenuous about the drawbacks or alternatives tend to be
+  poorly-received.
+
+- Submit a pull request.
+  As a pull request the RFC will receive design feedback from the
+  various stakeholders, and the author should be prepared to revise it
+  in response.
+
+- Now that your RFC has an open pull request, use the issue number of
+  the PR to update your 0000- prefix to that number.
```
ljharb, line 94: i've seen a few places use this convention, and it seems like it adds unnecessary churn. is there a benefit to using the number in the filename that git history doesn't provide?

legendecas: I'd find that the numbering causes a lot of unnecessary conflicts too.

ptomato: Having the number in the filename seems like a benefit if you want to be able to colloquially refer to "RFC 1234" rather than "the deepEqual RFC". Otherwise you'll have to look for your title in the crowded `rfcs/` directory. That said, this part comes from Rust and I'm not attached to it.

rfcs/0000-template.md
ptomato: I find soft-wrapped markdown inconvenient on GitHub when viewing diffs, because you often have to scroll side-to-side. Hard-wrapped is also annoying for diffs because you rewrap subsequent lines when you change something, but my standard practice is to limit that by putting a newline after each sentence as well as at the ruler.

rfcs/_process.md
```diff
@@ -0,0 +1,144 @@
...
+- [RFC postponements](#rfc-postponements)
+- [Acknowledgements](#acknowledgements)
+
+## When you need to follow this process
+
+You need to follow this process if you intend to make "substantial" changes to the test262 suite and harness.
+What constitutes a "substantial" change is evolving based on community norms, but may include the following.
+
+- Additions or changes to [INTERPRETING.md](https://github.com/tc39/test262/blob/main/INTERPRETING.md) that require consumers of the test suite to make changes to their test runners.
+- Additions to [features.txt](https://github.com/tc39/test262/blob/main/features.txt) that are not new Stage 3 proposals, or removals of feature flags.
```
ptomato, line 28: In hindsight an RFC seems a bit heavyweight for this item. Maybe this could just be a PR that the maintainers sign off on?

rfcs/_process.md
```diff
@@ -0,0 +1,144 @@
...
+- Bug fixes to the [harness files](https://github.com/tc39/test262/tree/main/harness).
+
+If you submit a pull request for a "substantial" change without going through the RFC process, it may be closed with a polite request to submit an RFC first.
+
+## Before creating an RFC
+
+Laying some groundwork ahead of the RFC can make the process smoother.
+Getting people to agree with an RFC often requires some effort towards consensus-building.
+
+The most common preparations for writing and submitting an RFC include talking the idea over on our [Matrix channel](https://matrix.to/#/#test262-maintainers:matrix.org), discussing the topic on our [discussion forum](https://es.discourse.group/c/test262/FIXME), and occasionally posting "pre-RFCs" on the forum.
```
ptomato, line 45: We should create this tag in the Discourse.

rfcs/_process.md
```diff
@@ -0,0 +1,144 @@
...
+- At some point, a maintainer will propose a "motion for final comment period" (FCP), along with a disposition for the RFC (merge, close, or postpone).
+  - This step is taken when enough of the tradeoffs have been discussed that the maintainers group is in a position to make a decision.
+    That does not require consensus among all participants in the RFC thread (which is usually impossible).
+    However, the argument supporting the disposition on the RFC needs to have already been clearly articulated.
+    Maintainers use their best judgment in taking this step, and the FCP itself ensures there is ample time and notification for stakeholders to push back if it is made prematurely.
+  - For RFCs with lengthy discussion, the motion to FCP is usually preceded by a summary comment trying to lay out the current state of the discussion and major tradeoffs/points of disagreement.
+  - Before actually entering FCP, the maintainers group must sign off as a whole.
+    This requires the usual "consensus minus one".
+
+- The FCP lasts 10 calendar days, so that it is open for at least 5 business days, absent holidays.
```
ptomato, line 87: I think the FCP was too long for the RFC we just merged, but I'm not sure if I'd say that it's too long in general.

rfcs/_process.md
```diff
@@ -0,0 +1,144 @@
...
+  It is also advertised widely, e.g. on the discussion forum and in the [TC39 General](https://matrix.to/#/#tc39-general:matrix.org) Matrix channel.
```
ptomato, line 88: We should be more specific about where to publish it. I advertised it in TC39 General and TC39 Implementers on Matrix.

rfcs/_process.md
```diff
@@ -0,0 +1,144 @@
...
+  This way all stakeholders have a chance to lodge any final objections before a decision is reached.
+
+- In most cases, the FCP period is quiet, and the RFC is either merged or closed.
```
ptomato, line 91:
```suggestion
- In most cases, the FCP is quiet, and the RFC is either merged or closed.
```

ptomato (commented):
Now that we've had one RFC land, does anyone have any feedback on the process? I've added some specific feedback from myself as review comments.

As for general feedback, I think the process is a bit on the heavy side for the harness modification that we did. I wonder if we can adjust things so that the amount of effort is proportional to the disruptiveness of the change.

rfcs/_process.md
ljharb: I agree, features should just be a PR.

rfcs/_process.md
ptomato: During the meeting we discussed that it's probably not necessary to have a tag at this time — communication via the GitHub repo and Matrix channels seems to be sufficient right now

rfcs/_process.md
ptomato: During the meeting we discussed that rather than shortening the FCP, we'd try to use the RFC process only for larger changes in the future, where this length would be more appropriate for the FCP.

ptomato:
I've updated the process document based on our discussion in this week's maintainers meeting, please have a look! (I've kept the changes in a separate commit so it's easy to see what's new.) It would be nice to be able to merge this on Monday so we can announce in our TC39 status update that it's done.

jugglinmike (approved):

ljharb (approved):

PR #3531, opened 2022-05-19
Temporal: Test @@toStringTag properties.

Ms2ger:


ptomato (approved):
Great, thanks!

I verified that tests for `Temporal[Symbol.toStringTag]` and `Temporal.Now[Symbol.toStringTag]` are already present.

PR #3532, opened 2022-05-20
Temporal: Extend toJSON/toString tests.

Ms2ger:


ptomato:
Thanks, making all the toJSON methods consistent seems good!

ptomato (approved):

PR #3537, opened 2022-05-25
Temporal: Cleanup some tests.

Ms2ger:
The options-invalid.js tests were also covered by options-wrong-type.js.

The tests for add/subtract without an options argument were also covered by options-undefined.js.


ptomato (approved):
Thanks!

PR #3543, opened 2022-05-26
Add test262-harness-dotnet to README

lahma:
I've developed a library for running and generating (currently) NUnit based test suites for test262. This is already being successfully used in two .NET projects, [Jint](https://github.com/sebastienros/jint) (JS interpreter) and [esprima-dotnet](https://github.com/sebastienros/esprima-dotnet) (JS parsing library).


ptomato (approved):
Thanks! Seems good to have links to any current projects using this.

PR #3550, opened 2022-06-02
Add draft for community responsibilities

sarahghp:
What it says on the tin! Most of this is very sketchy, so I am interested to know what folks think.


sarahghp:
@tc39/test262-maintainers

PR #3551, opened 2022-06-03
add noStrict flag to yield-as-binding-id cases

JLHwung:
The test preparation code in recently added decorator tests contains `var yield` and `function yield() {}`. However, a BindingIdentifier must not be `yield` in strict mode, so I add `noStrict` flag to these cases.

Context: https://github.com/babel/babel/pull/14638


Ms2ger:
cc @legendecas

legendecas (approved):
LGTM. Thank you for the fix!

Ms2ger (approved):

PR #3555, opened 2022-06-10
Temporal: Test custom calendar operations.

Ms2ger:


test/built-ins/Temporal/PlainDate/prototype/dayOfWeek/custom.js
```diff
@@ -0,0 +1,27 @@
...
+    ++calls;
+    assert.compareArray(args, [pd], "dayOfWeek arguments");
+    return "7";
+  }
+}
+
+const calendar = new CustomCalendar();
+const pd = new Temporal.PlainDate(1830, 8, 25, calendar);
+const result = pd.dayOfWeek;
+assert.sameValue(result, "7", "result");
```
ptomato, line 26: Dang, it seems that this is correct, but it certainly wasn't intended. I guess this is affected by tc39/proposal-temporal#2265.

ptomato (approved):
This does reflect the proposal as it currently stands, so I'll merge it, but heads up that there will be a normative change applying to these files...

PR #3556, opened 2022-06-10
Temporal: Add some Instant string tests.

Ms2ger:


ptomato (approved):
Not sure if there are more variations that we should put in here, but nothing jumped out as obviously missing. Thanks!

PR #3560, opened 2022-06-11
Temporal: Extend tests for getISOFields.

Ms2ger:


test/built-ins/Temporal/PlainDate/prototype/getISOFields/custom.js
```diff
@@ -0,0 +1,35 @@
+// Copyright (C) 2022 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.plaindate.prototype.getisofields
+description: Properties added in correct order to object returned from getISOFields
```
ptomato, line 6: This description doesn't seem to apply to this test. (and in similar tests below)

ptomato (changes requested):

PR #3572, opened 2022-06-16
Normative: Add new numbering system "tnsa"

romulocintra:
#### Summary

 Add new numbering system "tnsa"  to tests

#### Supporting details

https://github.com/unicode-org/cldr/blob/main/common/bcp47/number.xml#L95

#### Related issues

https://github.com/tc39/ecma402/pull/614


ptomato (commented):
Should there be a corresponding addition to `numberingSystemDigits`? From what I can tell, there is a 1:1 mapping with U+16AC0 through U+16AC9

harness/testIntl.js
```diff
@@ -2278,6 +2279,7 @@ var numberingSystemDigits = {
   takr: "𑛀𑛁𑛂𑛃𑛄𑛅𑛆𑛇𑛈𑛉",
   talu: "᧐᧑᧒᧓᧔᧕᧖᧗᧘᧙",
   tamldec: "௦௧௨௩௪௫௬௭௮௯",
+  tnsa: "\u16AC0\u16AC1\u16AC2\u16AC3\u16AC4\u16AC5\u16AC6\u16AC7\u16AC8\u16AC9",
```
ptomato, line 2282: It's 5 digits so it requires braces:
```suggestion
  tnsa: "\u{16AC0}\u{16AC1}\u{16AC2}\u{16AC3}\u{16AC4}\u{16AC5}\u{16AC6}\u{16AC7}\u{16AC8}\u{16AC9}",
```
(`\u16AC0` is `\u16AC` followed by `0`)

ptomato (approved):
This seems correct to me.
I'm not sure if there are any other places where we'd have to edit a table, or something. I'll check with the other maintainers, then merge it.

I see number formats in `implementation-contributed/v8/intl/{date,number}-format/check-nu-option.js` but I understand our current policy is not to update the tests in `implementation-contributed/` unless the implementations themselves do it.

PR #3538, opened 2022-05-25
Temporal: Add tests for prototype.constructor.

Ms2ger:

PR #3574, opened 2022-06-16
Temporal: Add missing feature flag for prototype.constructor tests.

rkirsling:
Fix for #3538. Temporal tests will immediately ReferenceError out without a feature flag.

@Ms2ger @ptomato


ptomato (approved):
Thanks for catching this.

PR #3575, opened 2022-06-16
Fix wrong error for a lexical redeclaration test

grassator:
The test as originally specified fails in all compatible parsers, but for the wrong reason. Below is an excerpt from V8, but all parser I tested behave the same:

```js
for (const x; false; ) {
           ^
SyntaxError: Missing initializer in const declaration
```

After the change the error is the assumed:

```js
  var x;
      ^
SyntaxError: Identifier 'x' has already been declared
```


grassator:
> Out of curiosity: how did you spot this one?

I am writing an ECMAScript parser as a hobby project and after implementing the const init early error I noticed that one of the newly passing tests didn't make sense.

PR #3577, opened 2022-06-17
Add tests for Intl.Segmenter option granularity - "word" 

romulocintra:
Increasing coverage on Intl.Segmenter when used with option granularity and value "word" related with https://github.com/tc39/ecma402/issues/656

cc @FrankYFTang  @sffc

### DIFF
--- a/test/intl402/Segmenter/constructor/constructor/options-granularity-word.js
+++ b/test/intl402/Segmenter/constructor/constructor/options-granularity-word.js
@@ -19,16 +19,15 @@ const testValuesByLength = [
   ["example.es , example.com they from era .com", 14],
   ["124.13 1231,23", 3],
   ["#test262, #TC39", 6],
+  ["my@mail.com", 3],
   ["my@mail.com sending message to my@mail.org", 13],
   ["it's a @decorator", 6],
 ];
 
 const testValuesByResult = [
   ["example.es", "example.es"],
-  ["my@mail.com", "my@mail.com"],
   ["mail.com", "mail.com"],
-  ["#test262", "#test262"],
-  ["https://example.com", "https://example.com"],
+  ["www.wikipedia.org", "www.wikipedia.org"],
 ];
 
 // test word segmentation based on word count
--- a/test/intl402/Segmenter/constructor/constructor/options-granularity-word.js
+++ b/test/intl402/Segmenter/constructor/constructor/options-granularity-word.js
@@ -1,0 +1,57 @@
+// Copyright 2022 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-Intl.Segmenter
+description: Checks handling of valid values for the granularity option "word" when string text includes URLs, email addresses, #, Dates and acronyms
+features: [Intl.Segmenter]
+---*/
+
+const testValuesByLength = [
+  ["F.C.B", 1],
+  ["F.C.BALL", 1],
+  ["F.C. BALL", 4],
+  ["Please P.T.O 123 times. 123.3242432", 10],
+  ["at 12:43 PM", 7],
+  ["in 21/03/2013.", 8],
+  ["https://example.com/tc39/test262", 9],
+  ["https://example.com/tc39/test262?id=test", 13],
+  ["example.es , example.com they from era .com", 14],
+  ["124.13 1231,23", 3],
+  ["#test262, #TC39", 6],
+  ["my@mail.com sending message to my@mail.org", 13],
+  ["it's a @decorator", 6],
+];
+
+const testValuesByResult = [
+  ["example.es", "example.es"],
+  ["my@mail.com", "my@mail.com"],
+  ["mail.com", "mail.com"],
+  ["#test262", "#test262"],
+  ["https://example.com", "https://example.com"],
+];
+
+// test word segmentation based on word count
+function testSegmenterByExpectedLengths([value, expectedSize]) {
+  const segmenter = new Intl.Segmenter("en", { granularity: "word" });
+  const segments = segmenter.segment(value);
+  let actualSize = 0;
+  for (s of segments) {
+    actualSize++;
+  }
+  assert.sameValue(
+    actualSize,
+    expectedSize,
+    `Assert segment count for the text string : ${value}"`
+  );
+}
+
+// test word segmentation based on expected results
+function testSegmenterByExpectedValues([value, expected]) {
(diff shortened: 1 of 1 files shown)


test/intl402/Segmenter/constructor/constructor/options-granularity-word.js
```diff
@@ -0,0 +1,57 @@
...
+  ["F.C. BALL", 4],
+  ["Please P.T.O 123 times. 123.3242432", 10],
+  ["at 12:43 PM", 7],
+  ["in 21/03/2013.", 8],
+  ["https://example.com/tc39/test262", 9],
+  ["https://example.com/tc39/test262?id=test", 13],
+  ["example.es , example.com they from era .com", 14],
+  ["124.13 1231,23", 3],
+  ["#test262, #TC39", 6],
+  ["my@mail.com sending message to my@mail.org", 13],
```
romulocintra, line 22: Expecting a length of 13 based on JSC output but I believe the correct answer it's 9 due the segmentation of email addresses 

```
[Log] my
[Log] @
[Log] mail.com
[Log]  
[Log] sending
[Log]  
[Log] message
[Log]  
[Log] to
[Log]  
[Log] my
[Log] @
[Log] mail.org
```

test/intl402/Segmenter/constructor/constructor/options-granularity-word.js
```diff
@@ -0,0 +1,57 @@
...
+
+const testValuesByLength = [
+  ["F.C.B", 1],
+  ["F.C.BALL", 1],
+  ["F.C. BALL", 4],
+  ["Please P.T.O 123 times. 123.3242432", 10],
+  ["at 12:43 PM", 7],
+  ["in 21/03/2013.", 8],
+  ["https://example.com/tc39/test262", 9],
+  ["https://example.com/tc39/test262?id=test", 13],
```
romulocintra, line 18: Unsure about the rules for URL segmentation and could be interpreted differently

ryzokuken (dismissed):
Tests LGTM and pass everywhere, should be fine.

test/intl402/Segmenter/constructor/constructor/options-granularity-word.js
```diff
@@ -0,0 +1,57 @@
...
+  ["my@mail.com sending message to my@mail.org", 13],
+  ["it's a @decorator", 6],
+];
+
+const testValuesByResult = [
+  ["example.es", "example.es"],
+  ["my@mail.com", "my@mail.com"],
+  ["mail.com", "mail.com"],
+  ["#test262", "#test262"],
+  ["https://example.com", "https://example.com"],
```
ptomato, line 31: These results expecting just one segment, seem to contradict the tests above this, where email addresses and URLs are broken into many segments?

ptomato: My preference would be to use the intended behaviour in all the tests. If the test fails in engines that use ICU, then that provides pressure to ICU to adjust. If we later determine that ICU isn't going to change, then we can still change the expectation in this test.

Any opinions from other maintainers on this?

rwaldron:
@ptomato I'll leave this to you to merge when you're satisfied

test/intl402/Segmenter/constructor/constructor/options-granularity-word.js
ptomato: Seems like no other opinions for the time being; I do feel strongly that before merging, these tests should be internally consistent, so we should either expect the intended results everywhere, or consistently expect the results taking the ICU shortcomings into account. I would prefer the former, could be convinced of the latter, but not a mix :smile:

sffc: All test expectations should align with the requirements from UTR 14.

Running plain ICU with no patches should get you UTR 14 behavior. I wouldn't trust browser output, although Safari is probably better than Chrome.

To get the "pure ICU" output, you can use this tool:

https://icu4c-demos.unicode.org/icu-bin/icusegments#3/en

romulocintra: Does this mean you support option 1 mentioned above?

> 1. Modify the tests to be according to the ICU results that agree with some engine implementations, even if for userland purposes they look wrong and report to ICU those use cases. [as suggested here](https://github.com/tc39/ecma402/issues/656#issuecomment-1159323276)

ryzokuken: I'm also split here. While I see @sffc's point, perhaps it's better to stick to what the user intends and fix upstream than the other way around.

However, more fundamentally, why are we testing unspecified behavior in test262? I feel that this isn't quite right, and if some behavior is permitted by the spec then it shouldn't fail test262 tests irrespective of what ICU says or even what the user expects. test262 is compliance tests and not implementation tests...

sffc: My issue is, how do you define "user expected behavior"? That is not written down anywhere, and there are likely many other cases besides just email addresses that produce surprising behavior. I think we should make the tests conform to UAX 29, which is what EMCA-402 refers to, and modify the tests whenever changes land in UAX 29, after having gone through an approval process.

> However, more fundamentally, why are we testing unspecified behavior in test262?

It is not unspecified; it is specified in UAX 29, and ECMA-402 has a normative reference to it:

https://tc39.es/ecma402/#sec-findboundary

We agreed to add these tests in the June 2022 TG2 meeting. Notes:

https://github.com/tc39/ecma402/blob/master/meetings/notes-2022-06-16.md#intlsegmenter-with-urls-email-addresses-and-acronyms-656

ryzokuken: > It is not unspecified; it is specified in UAX 29, and ECMA-402 has a normative reference to it:

> We agreed to add these tests in the June 2022 TG2 meeting.

I was not aware of these, thanks!

In this case, I feel that tests should do what the spec says (UAX 29) and only change once we have changes in the UAX spec. Any differences between what's "expected behavior" and "specified behavior" should be open tickets in the Unicode Jira and not failing tests in test262...

romulocintra: Test are agreeing with specified behaviour (UAX 29 ) and ticket was open https://unicode-org.atlassian.net/browse/CLDR-15767

romulocintra:
Closing it until we get a consensus on this topic. cc @sffc

PR #3579, opened 2022-06-20
Temporal: Improve branding tests.

### LINKED ISSUE #3224: Temporal "branding" tests are imprecise

Ms2ger:
Fixes #3224.

### DIFF
--- a/test/built-ins/Temporal/Calendar/prototype/fields/branding.js
+++ b/test/built-ins/Temporal/Calendar/prototype/fields/branding.js
@@ -11,12 +11,14 @@ const fields = Temporal.Calendar.prototype.fields;
 
 assert.sameValue(typeof fields, "function");
 
-assert.throws(TypeError, () => fields.call(undefined), "undefined");
-assert.throws(TypeError, () => fields.call(null), "null");
-assert.throws(TypeError, () => fields.call(true), "true");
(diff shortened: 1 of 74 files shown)


ptomato (approved):
I didn't review each one individually, but I think I got a representative sample.

This should close #3224, yes?

PR #3585, opened 2022-06-23
Fix decorator tests

legendecas:
Fixes comments in https://github.com/tc39/test262/pull/3512#discussion_r905015609.


ptomato (approved):
Thanks, these seem to solve the comments in the other thread.

PR #3591, opened 2022-06-27
Add proposal-temporal's Temporal.Instant tests to staging

Ms2ger:


ljharb:
hmm, i would expect @tc39/test262-staging to be a codeowner too. I wonder if the `test/staging/*` needs to be `test/staging` or `test/staging/**/*`?

ljharb:
@ptomato while it seems fine to allow it, technically only the earliest year needs to be mentioned, so they could also change to just "2018"

ptomato:
Yeah, that's fine too.

ptomato:
> hmm, i would expect @tc39/test262-staging to be a codeowner too. I wonder if the `test/staging/*` needs to be `test/staging` or `test/staging/**/*`?

Based on https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#example-of-a-codeowners-file I guess it should be `/test/staging/`. Sorry, I should've paid more attention when I merged it.

ptomato (approved):
Merging as discussed in the maintainers meeting

ljharb:
Went ahead and pushed the CODEOWNERS change in 8c64a666e10fd88c58d63d4cb02f771fd29684fe

PR #3599, opened 2022-07-05
Import SpiderMonkey Temporal tests

anba:
Temporal tests written for the SpiderMonkey implementation. Mostly
covers edge cases around mathematical operations and regression tests
for reported spec bugs.

### DIFF
--- a/test/built-ins/Temporal/Duration/prototype/round/nanoseconds-to-days-precision-exact-mathematical-values-1.js
+++ b/test/built-ins/Temporal/Duration/prototype/round/nanoseconds-to-days-precision-exact-mathematical-values-1.js
@@ -1,0 +1,77 @@
+// Copyright (C) 2022 André Bargull. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.zoneddatetime.prototype.round
+description: >
+  NanosecondsToDays computes with precise mathematical integers.
+info: |
+  NanosecondsToDays ( nanoseconds, relativeTo )
+  ...
+  14. If sign is 1, then
+    a. Repeat, while days > 0 and intermediateNs > endNs,
+      i. Set days to days - 1.
+     ii. ...
+
+  Ensure |days = days - 1| is exact and doesn't loose precision.
+features: [Temporal]
+---*/
+
+var expectedDurationDays = [
+  Number.MAX_SAFE_INTEGER + 4, // 9007199254740996
+  Number.MAX_SAFE_INTEGER + 3, // 9007199254740994
+  Number.MAX_SAFE_INTEGER + 2, // 9007199254740992
+  Number.MAX_SAFE_INTEGER + 1, // 9007199254740992
+  Number.MAX_SAFE_INTEGER + 0, // 9007199254740991
+  Number.MAX_SAFE_INTEGER - 1, // 9007199254740990
+  Number.MAX_SAFE_INTEGER - 2, // 9007199254740989
+  Number.MAX_SAFE_INTEGER - 3, // 9007199254740988
+  Number.MAX_SAFE_INTEGER - 4, // 9007199254740987
+  Number.MAX_SAFE_INTEGER - 5, // 9007199254740986
+];
+
+// Intentionally not Test262Error to ensure assertions errors are propagated.
+class StopExecution extends Error {}
+
+var cal = new class extends Temporal.Calendar {
+  #dateUntil = 0;
+
+  dateUntil(one, two, options) {
+    if (++this.#dateUntil === 1) {
+      return Temporal.Duration.from({days: Number.MAX_SAFE_INTEGER + 4});
+    }
+    return super.dateUntil(one, two, options);
+  }
+
+  #dateAdd = 0;
+
+  dateAdd(date, duration, options) {
+    // Ensure we don't add too many days which would lead to creating an invalid date.
+    if (++this.#dateAdd === 3) {
+      assert.sameValue(duration.days, Number.MAX_SAFE_INTEGER + 4);
+
+      // The added days must be larger than 5 for the |intermediateNs > endNs| condition.
+      return super.dateAdd(date, "P6D", options);
+    }
+
+    // Ensure the duration days are exact.
+    if (this.#dateAdd > 3) {
+      if (!expectedDurationDays.length) {
+        throw new StopExecution();
+      }
+      assert.sameValue(duration.days, expectedDurationDays.shift());
+
+      // Add more than 5 for the |intermediateNs > endNs| condition.
+      return super.dateAdd(date, "P6D", options);
+    }
+
+    // Otherwise call the default implementation.
+    return super.dateAdd(date, duration, options);
+  }
+}("iso8601");
+
+var zoned = new Temporal.ZonedDateTime(0n, "UTC", cal);
+var duration = Temporal.Duration.from({days: 5});
+var options = {smallestUnit: "days", relativeTo: zoned};
+
+assert.throws(StopExecution, () => duration.round(options));
--- a/test/built-ins/Temporal/Duration/prototype/round/nanoseconds-to-days-precision-exact-mathematical-values-2.js
+++ b/test/built-ins/Temporal/Duration/prototype/round/nanoseconds-to-days-precision-exact-mathematical-values-2.js
@@ -1,0 +1,45 @@
+// Copyright (C) 2022 André Bargull. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.zoneddatetime.prototype.round
+description: >
+  NanosecondsToDays computes with precise mathematical integers.
+info: |
+  NanosecondsToDays ( nanoseconds, relativeTo )
+  ...
+  17. Repeat, while done is false,
+    ...
+    c. If (nanoseconds - dayLengthNs) × sign ≥ 0, then
+      ...
+      iii. Set days to days + sign.
+
+  Ensure |days = days + sign| is exact and doesn't loose precision.
+features: [Temporal]
+---*/
+
+var cal = new class extends Temporal.Calendar {
+  #dateUntil = 0;
+
+  dateUntil(one, two, options) {
+    if (++this.#dateUntil === 1) {
+      return Temporal.Duration.from({days: Number.MAX_SAFE_INTEGER + 10});
+    }
+    return super.dateUntil(one, two, options);
+  }
+
+  #dateAdd = 0;
+
+  dateAdd(date, duration, options) {
+    if (++this.#dateAdd === 3) {
+      return super.dateAdd(date, "P1D", options);
+    }
+    return super.dateAdd(date, duration, options);
+  }
+}("iso8601");
+
+var zoned = new Temporal.ZonedDateTime(0n, "UTC", cal);
+var duration = Temporal.Duration.from({days: 5});
+var result = duration.round({smallestUnit: "days", relativeTo: zoned});
+
+assert.sameValue(result.days, Number(BigInt(Number.MAX_SAFE_INTEGER + 10) + 5n));
--- a/test/built-ins/Temporal/Duration/prototype/total/precision-exact-mathematical-values-1.js
+++ b/test/built-ins/Temporal/Duration/prototype/total/precision-exact-mathematical-values-1.js
@@ -1,0 +1,96 @@
+// Copyright (C) 2022 André Bargull. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.duration.prototype.total
+description: >
+  RoundDuration computes on exact mathematical values.
+features: [Temporal]
+---*/
+
+// Return the next Number value in direction to +Infinity.
+function nextUp(num) {
+  if (!Number.isFinite(num)) {
+    return num;
+  }
+  if (num === 0) {
+    return Number.MIN_VALUE;
+  }
+
+  var f64 = new Float64Array([num]);
+  var u64 = new BigUint64Array(f64.buffer);
+  u64[0] += (num < 0 ? -1n : 1n);
+  return f64[0];
+}
+
+// Return the next Number value in direction to -Infinity.
+function nextDown(num) {
+  if (!Number.isFinite(num)) {
+    return num;
+  }
+  if (num === 0) {
+    return -Number.MIN_VALUE;
+  }
+
+  var f64 = new Float64Array([num]);
+  var u64 = new BigUint64Array(f64.buffer);
+  u64[0] += (num < 0 ? 1n : -1n);
+  return f64[0];
+}
+
+let duration = Temporal.Duration.from({
+  hours: 4000,
+  nanoseconds: 1,
+});
+
+let total = duration.total({unit: "hours"});
+
+// From RoundDuration():
+//
+// 7. Let fractionalSeconds be nanoseconds × 10^-9 + microseconds × 10^-6 + milliseconds × 10^-3 + seconds.
+// = nanoseconds × 10^-9
+// = 1 × 10^-9
+// = 10^-9
+// = 0.000000001
+//
+// 13.a. Let fractionalHours be (fractionalSeconds / 60 + minutes) / 60 + hours.
+// = (fractionalSeconds / 60) / 60 + 4000
+// = 0.000000001 / 3600 + 4000
+//
+// 13.b. Set hours to RoundNumberToIncrement(fractionalHours, increment, roundingMode).
+// = trunc(fractionalHours)
+// = trunc(0.000000001 / 3600 + 4000)
+// = 4000
+//
+// 13.c. Set remainder to fractionalHours - hours.
+// = fractionalHours - hours
+// = 0.000000001 / 3600 + 4000 - 4000
+// = 0.000000001 / 3600
+//
+// From Temporal.Duration.prototype.total ( options ):
+//
+// 18. If unit is "hours", then let whole be roundResult.[[Hours]].
+// ...
+// 24. Return whole + roundResult.[[Remainder]].
+//
+// |whole| is 4000 and the remainder is (0.000000001 / 3600).
+//
+//   0.000000001 / 3600
+// = (1 / 10^9) / 3600
+// = (1 / 36) / 10^11
+// = 0.02777.... / 10^11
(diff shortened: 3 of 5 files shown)


Ms2ger:
Linter fails - I think you're missing the year in your copyright headers

test/built-ins/Temporal/Duration/compare/precision-exact-mathematical-values-1.js
```diff
@@ -0,0 +1,68 @@
...
+    a. Let unbalanceResult1 be ? UnbalanceDurationRelative(one.[[Years]], one.[[Months]],
+       one.[[Weeks]], one.[[Days]], "day", relativeTo).
+  ...
+  9. Let ns1 be ! TotalDurationNanoseconds(days1, one.[[Hours]], one.[[Minutes]], one.[[Seconds]],
+     one.[[Milliseconds]], one.[[Microseconds]], one.[[Nanoseconds]], shift1).
+ 10. Let ns2 be ! TotalDurationNanoseconds(days2, two.[[Hours]], two.[[Minutes]], two.[[Seconds]],
+     two.[[Milliseconds]], two.[[Microseconds]], two.[[Nanoseconds]], shift2).
+ 11. If ns1 > ns2, return 1𝔽.
+ 12. If ns1 < ns2, return -1𝔽.
+ 13. Return +0𝔽.
```
ptomato, line 23:
```suggestion
  10. Let ns2 be ! TotalDurationNanoseconds(days2, two.[[Hours]], two.[[Minutes]], two.[[Seconds]],
     two.[[Milliseconds]], two.[[Microseconds]], two.[[Nanoseconds]], shift2).
  11. If ns1 > ns2, return 1𝔽.
  12. If ns1 < ns2, return -1𝔽.
  13. Return +0𝔽.
```
(dedenting by one position seems to mess up the YAML parser)

test/built-ins/Temporal/Duration/compare/precision-exact-mathematical-values-2.js
```diff
@@ -0,0 +1,68 @@
...
+    a. Let unbalanceResult1 be ? UnbalanceDurationRelative(one.[[Years]], one.[[Months]],
+       one.[[Weeks]], one.[[Days]], "day", relativeTo).
+  ...
+  9. Let ns1 be ! TotalDurationNanoseconds(days1, one.[[Hours]], one.[[Minutes]], one.[[Seconds]],
+     one.[[Milliseconds]], one.[[Microseconds]], one.[[Nanoseconds]], shift1).
+ 10. Let ns2 be ! TotalDurationNanoseconds(days2, two.[[Hours]], two.[[Minutes]], two.[[Seconds]],
+     two.[[Milliseconds]], two.[[Microseconds]], two.[[Nanoseconds]], shift2).
+ 11. If ns1 > ns2, return 1𝔽.
+ 12. If ns1 < ns2, return -1𝔽.
+ 13. Return +0𝔽.
```
ptomato, line 23: Same here.

test/intl402/Temporal/TimeZone/non-canonical-africa.js
```diff
@@ -0,0 +1,19 @@
+// Copyright (C) 2022 André Bargull. All rights reserved.
```
ptomato, line 1: Out of curiosity what's the reason the non-canonical time zone name tests can't be in one file?

ptomato (commented):
Thanks. I expect to be able to look at these in detail next week, although I need to look at https://github.com/tc39/test262/pull/3523 first, so it will depend on when I get through with that one.

I've given a quick try to running these and have left some comments based on that.

test/built-ins/Temporal/PlainTime/prototype/add/precision-exact-mathematical-values-2.js
```diff
@@ -0,0 +1,23 @@
...
+  minutes: Number.MAX_VALUE,
+});
+
+let time = new Temporal.PlainTime(0, 0, 0, 0, 0, 0);
+
+let result = time.add(duration);
+
+TemporalHelpers.assertPlainTime(result, 10, 8, 0, 0, 0, 0);
+
+assert.sameValue("10:08:00", result.toString());
```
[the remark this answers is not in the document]
anba: Actually I've removed this line, because it was a leftover from the original test and the assertion is superseded by `TemporalHelpers.assertPlainTime`.

test/built-ins/Temporal/Duration/prototype/add/days-is-number-max-value-with-zoneddatetime.js
```diff
@@ -0,0 +1,18 @@
+// Copyright (C) 2022 André Bargull. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.duration.prototype.add
```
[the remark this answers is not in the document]
anba: Added `subtract` tests for all `add` tests under "Temporal/Duration" and "Temporal/PlainTime".

test/intl402/Temporal/TimeZone/non-canonical-africa.js
ptomato: Wouldn't all the files have to be skipped anyway if CLDR canonicalization was used?

We could also assert that the ID is equal to either the expected value for CLDR or IANA, but in practice that probably doesn't tell us anything more than your second option (only testing that the non-canonical name is allowed.)

test/intl402/Temporal/TimeZone/non-canonical-antarctica.js
```diff
@@ -0,0 +1,19 @@
...
+
+/*---
+esid: sec-temporal.timezone
+description: >
+  TimeZone constructor canonicalises its input.
+features: [Temporal]
+---*/
+
+const testCases = {
+  "Antarctica/South_Pole": "Antarctica/McMurdo",
```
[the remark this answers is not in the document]
anba: ECMA-402 assumes that all links are in `backward`, which is a [known error](https://tc39.es/archives/bugzilla/1892/). For example following the spec by the letter means that `"Etc/GMT"` is canonicalised to `"UTC"` (per [CanonicalizeTimeZoneName](https://tc39.es/ecma402/#sec-canonicalizetimezonename), step 3), whereas `"GMT"` shouldn't be canonicalised, because the `GMT` &rarr; `Etc/GMT` link is in [etcetera](https://github.com/eggert/tz/blob/bf8aa9414b0de5fc9df5f25fa12d07451510e926/etcetera#L26). 

https://github.com/tc39/ecma402/issues/121#issuecomment-323929572 proposes to follow all links in the main time zone files (i.e. excluding `backzone`), but implementers are unlikely to completely ignore `backzone` because of the possible negative cultural/political effects mentioned in https://github.com/tc39/ecma402/issues/272#issuecomment-423928522.

test/intl402/Temporal/TimeZone/non-canonical-africa.js
anba: > Wouldn't all the files have to be skipped anyway if CLDR canonicalization was used?

Not necessarily all files, for example all links in "non-canonical-brazil.js" are exactly the same in IANA and CLDR.

test/intl402/Temporal/TimeZone/non-canonical-antarctica.js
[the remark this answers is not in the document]
ptomato: If the links for Antarctica/South_Pole in `backward` and `backzone` conflict, then which one should take precedence in your opinion?

(If this is an existing web reality issue then it seems better to me that we only check that the link names _can_ be canonicalized, and not what they are canonicalized to...)

test/intl402/Temporal/TimeZone/prototype/getNextTransition/subtract-second-and-nanosecond-from-last-transition.js
```diff
@@ -0,0 +1,55 @@
...
+// From <https://github.com/eggert/tz/blob/main/europe>:
+//
+// # Zone  NAME       STDOFF  RULES FORMAT  [UNTIL]
+// Zone  Europe/Paris 0:09:21 -     LMT     1891 Mar 15  0:01
+//                    0:09:21 -     PMT     1911 Mar 11  0:01 # Paris MT
+
+let tz = new Temporal.TimeZone("Europe/Paris");
+
+let zdt = new Temporal.PlainDateTime(1800, 1, 1).toZonedDateTime(tz);
+assert.sameValue(zdt.toString(), "1800-01-01T00:00:00+00:09[Europe/Paris]");
```
ptomato, line 20: FYI, we are usually avoiding using `toString()` to make assertions about Temporal objects, though here I think it is justified for easy comparison to the TZDB excerpt above.

test/intl402/Temporal/TimeZone/prototype/getNextTransition/transition-at-instant-boundaries.js
```diff
@@ -0,0 +1,32 @@
...
+  if (next) {
+    assert(next.epochNanoseconds > min.epochNanoseconds);
+  }
+
+  // There shouldn't be any next transition after |max|.
+  next = null;
+  try {
+    next = tz.getNextTransition(max);
+  } catch (e) {
+    // This can throw a RangeError.
```
ptomato, line 28: This is no longer the case, I think?

test/intl402/Temporal/TimeZone/prototype/getOffsetNanosecondsFor/nanoseconds-subtracted-or-added-at-dst-transition.js
```diff
@@ -0,0 +1,37 @@
...
+esid: sec-temporal.timezone.prototype.getoffsetnanosecondsfor
+description: >
+  Test offset when nanoseconds are subtracted or added from DST transition.
+features: [Temporal]
+---*/
+
+// From <https://github.com/eggert/tz/blob/main/northamerica>:
+//
+// # Rule  NAME  FROM  TO    -  IN   ON       AT    SAVE  LETTER
+// Rule    CA    1962  1966  -  Oct  lastSun  2:00  0     S
```
ptomato, line 14: Small detail, but isn't this the wrong line for the data used in the test below? We need the one about the [last Sunday in April](https://github.com/eggert/tz/blob/main/northamerica#L559)

test/intl402/Temporal/TimeZone/prototype/getPreviousTransition/transition-at-instant-boundaries.js
```diff
@@ -0,0 +1,32 @@
...
+  if (prev) {
+    assert(prev.epochNanoseconds < max.epochNanoseconds);
+  }
+
+  // There shouldn't be any previous transition before |min|.
+  prev = null;
+  try {
+    prev = tz.getPreviousTransition(min);
+  } catch (e) {
+    // This can throw a RangeError.
```
ptomato, line 28: Same comment as in the corresponding getNextTransition test

ptomato (commented):
I didn't go through everything, but I did verify a big chunk of these — would you prefer to keep everything together in one PR or would you like me to commit the ones that I've done already?

test/built-ins/Temporal/ZonedDateTime/prototype/round/smallest-unit-day-daylength-too-large.js
```diff
@@ -0,0 +1,85 @@
...
+  let zoned = new Temporal.ZonedDateTime(0n, tz);
+  let result = zoned.round({ smallestUnit: "days" });
+  assert(zoned.equals(result));
+}
+{
+  let tz = new TimeZone(minInstant);
+  let zoned = new Temporal.ZonedDateTime(maxInstant - oneDay, tz);
+  let result = zoned.round({ smallestUnit: "days" });
+  assert(zoned.equals(result));
+}
```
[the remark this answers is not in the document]
anba: Yes.

test/built-ins/Temporal/ZonedDateTime/prototype/round/smallest-unit-day-daylength-zero-or-negative.js
```diff
@@ -0,0 +1,67 @@
...
+  let zoned = new Temporal.ZonedDateTime(0n, tz);
+  assert.throws(RangeError, () => zoned.round({ smallestUnit: "days" }));
+}
+
+{
+  let tz = new TimeZone(-1n);
+  let zoned = new Temporal.ZonedDateTime(0n, tz);
+  let result = zoned.round({ smallestUnit: "days" });
+  assert(zoned.equals(result));
+}
```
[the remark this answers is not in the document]
anba: Yes.

test/intl402/Temporal/TimeZone/prototype/getNextTransition/transition-at-instant-boundaries.js
anba: <https://github.com/tc39/proposal-temporal/pull/2351> will ensure no error is thrown.

test/intl402/Temporal/TimeZone/prototype/getOffsetNanosecondsFor/nanoseconds-subtracted-or-added-at-dst-transition.js
anba: Thanks for catching this error!

anba:
> I didn't go through everything, but I did verify a big chunk of these — would you prefer to keep everything together in one PR or would you like me to commit the ones that I've done already?

It's all the same to me. Let's do whatever works best for you.

test/built-ins/Temporal/ZonedDateTime/prototype/round/smallest-unit-day-daylength-too-large.js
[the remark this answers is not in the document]
ptomato: Assuming we get consensus on those normative PRs this week, I'd rather just wait a few days and commit the updated behaviour, so let's adjust these (here and below) before merging.

ptomato:
> > I didn't go through everything, but I did verify a big chunk of these — would you prefer to keep everything together in one PR or would you like me to commit the ones that I've done already?
> 
> It's all the same to me. Let's do whatever works best for you.

I've done this in #3606.

test/intl402/Temporal/TimeZone/prototype/getNextTransition/transition-at-instant-boundaries.js
ptomato: I made that change and moved this into #3610.

test/intl402/Temporal/TimeZone/prototype/getPreviousTransition/transition-at-instant-boundaries.js
ptomato: I made that change and moved this into #3610.

ptomato:
My plan for following up:
1. Separate out the tests that are affected by the recently adopted Temporal normative changes, update them to reflect the changes, and merge them at the same time as the corresponding normative PRs
2. Then, review everything else that's not number precision issues
3. Save the most difficult work for last :stuck_out_tongue_closed_eyes: 

@anba, if you have time to do the updates for (1) I'd appreciate it, otherwise I'll get around to it sometime this week or next.

test/intl402/Temporal/TimeZone/non-canonical-antarctica.js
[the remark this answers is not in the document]
anba: I've updated the tests to only check that link names are accepted.

test/built-ins/Temporal/ZonedDateTime/prototype/round/smallest-unit-day-daylength-too-large.js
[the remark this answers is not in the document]
anba: #3617 updates this test to reject negative day lengths.

test/built-ins/Temporal/ZonedDateTime/prototype/round/smallest-unit-day-daylength-zero-or-negative.js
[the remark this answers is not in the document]
anba: #3617 updates this test to reject negative day lengths.

anba:
> @anba, if you have time to do the updates for (1) I'd appreciate it, otherwise I'll get around to it sometime this week or next.

I've added the following four commits:
- The first one updates a test which uses negative day lengths.
- The other three commits address review comments from your first review.

ptomato (commented):
Thanks, I've merged another batch in, and rebased this one on top.

test/intl402/Temporal/TimeZone/non-canonical-utc.js
```diff
@@ -0,0 +1,27 @@
...
+description: >
+  TimeZone constructor canonicalises its input.
+features: [Temporal]
+---*/
+
+const testCases = {
+  "Etc/GMT": "UTC",
+  "Etc/GMT+0": "UTC",
+  "Etc/GMT-0": "UTC",
+  "Etc/GMT0": "UTC",
```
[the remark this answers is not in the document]
anba: I think it's okay to test `Etc/GMT0` in both files, because it's both a legacy name and also another alias for `UTC`.

test/built-ins/Temporal/Duration/compare/precision-exact-mathematical-values-1.js
```diff
@@ -0,0 +1,68 @@
...
+
+var cal = new class extends Temporal.Calendar {
+  dateAdd(date, duration, options) {
+    // Add one day when one week was requested.
+    if (duration.toString() === "P1W") {
+      return super.dateAdd(date, "P1D", options);
+    }
+
+    // Only expect to add one week.
+    throw new Test262Error();
```
ptomato, line 58: For future reference, better to add a message here (and elsewhere) like `new Test262Error("dateAdd called with unexpected value")` to aid in debugging implementations. Don't worry about fixing it up, I can add this before I push it.

ptomato:
Sorry it took so long. I've reviewed another batch of these and they're ready to land in https://github.com/tc39/test262/pull/3677. Once that is merged, I'll rebase this again and review the remaining 5 tests. I expect to be able to get to that early next week.

ptomato (approved):
Thanks very much for the thorough tests, as well as for bearing with me for so long while I found time to work through the arithmetic ... these are unbelievably helpful.

ptomato (approved):
Approving again after force-push

PR #3604, opened 2022-07-13
DurationFormat basic format tests

romulocintra:


test/intl402/DurationFormat/prototype/format/basic-format-en.js
```diff
@@ -0,0 +1,26 @@
...
+  days: 3,
+  hours: 4,
+  minutes: 5,
+  seconds: 6,
+  milliseconds: 7,
+  microseconds: 8,
+  nanoseconds: 9,
+};
+
+const df = new Intl.DurationFormat('pt');
```
ptomato, line 24:
```suggestion
const df = new Intl.DurationFormat('en');
```
The name of the test and the expected value indicate English. (Or this could be `en-US` as the test frontmatter indicates?)

test/intl402/DurationFormat/prototype/format/style-options-en.js
```diff
@@ -0,0 +1,37 @@
...
+  milliseconds: 7,
+  microseconds: 8,
+  nanoseconds: 9,
+};
+
+
+for (const style in testData) {
+   const df = new Intl.DurationFormat("en", {style});
+   const expected = testData[style];
+   assert.sameValue(df.format(duration), expected);
```
ptomato, line 35: Nitpick: Could you add an assertion message here that indicates which style is failing?

ptomato (approved):
I have a few comments, but generally these seem like good tests of the basic formatting functionality.

ptomato (approved):
Thanks!

PR #3605, opened 2022-07-13
DurationFormat -> add missing feature flag  

romulocintra:
meta: add a missing feature flag .  cc @iamstolis


ptomato (approved):
Thanks for fixing it so quickly.

PR #3608, opened 2022-07-21
sync to PR100 of intl-numberformat-v3

FrankYFTang:
TC39 2022-07-20 discuss and agree to take 
 https://github.com/tc39/proposal-intl-numberformat-v3/pull/100 
to remove order check


FrankYFTang:
@sffc @ryanmurakami @romulocintra

anba:
Instead of removing the tests altogether, how about rewriting them to assert that it's now valid that `start` is before the `end` parameter?

ptomato:
Agreed, same comment as on #3609. Otherwise implementations that don't update will still pass the tests.

FrankYFTang:
@anba @ptomato please take another look

FrankYFTang:
PR100 is already landed

test/intl402/NumberFormat/prototype/formatRangeToParts/x-greater-than-y-not-throws.js
```diff
@@ -0,0 +1,47 @@
...
+  "formatRangeToParts" basic tests when argument  x > y, BigInt included and covers PartitionNumberRangePattern return a object.
+info: |
+  1.1.21 PartitionNumberRangePattern( numberFormat, x, y )
+  (...)
+features: [Intl.NumberFormat-v3]
+---*/
+
+const nf = new Intl.NumberFormat();
+
+// If x > y, throw a RangeError exception.
```
sffc, line 16: should not throw

test/intl402/PluralRules/prototype/selectRange/x-greater-than-y-not-throws.js
```diff
@@ -0,0 +1,16 @@
...
+description: >	
+  "selectRange" basic tests when argument  x > y, return a string.	
+info: |	
+  1.1.6 ResolvePluralRange ( pluralRules, x, y )	
+features: [Intl.NumberFormat-v3]	
+---*/	
+
+const pr = new Intl.PluralRules();	
+
+// 1. If x > y, throw a RangeError exception.
```
sffc, line 15: should not throw

FrankYFTang:
PTAL

sffc (dismissed):

test/intl402/NumberFormat/prototype/formatRange/x-greater-than-y-not-throws.js
```diff
@@ -0,0 +1,47 @@
...
+  "formatRange" basic tests when argument  x > y, BigInt included and covers PartitionNumberRangePattern return a string.
+info: |
+  1.1.21 PartitionNumberRangePattern( numberFormat, x, y )
+  (...)
+features: [Intl.NumberFormat-v3]
+---*/
+
+const nf = new Intl.NumberFormat();
+
+// If x > y, throw a RangeError exception.
```
ptomato, line 16:
```suggestion
// If x > y, return a string.
```

ptomato (dismissed):
One comment to fix, I'll do it from here

ptomato (dismissed):

ptomato (approved):

ptomato:
The linter problem was due to trailing spaces in the YAML front matter — we should make it more robust to this, but in the meantime please watch out for trailing spaces!

PR #3609, opened 2022-07-21
Sync to PR701

FrankYFTang:
In TC39 2022-07-20, we decide to merge PR701. 
See https://github.com/tc39/ecma402/pull/701 and https://docs.google.com/presentation/d/1UUvbf3FFu9PGtrPAKPdMad9DZuVFLIvkAsAxyJZyvxM/edit 
for details.


FrankYFTang:
@ryzokuken @gibson042 @romulocintra @sffc

ryzokuken (dismissed):

gibson042 (approved):

sffc:
Wouldn't it be better if we made the tests expect that the format functions don't throw and produce some kind of output, rather than just delete them?

ptomato:
Agreed, it'd be nice to have some sort of affirmative test that the reverse order works; otherwise implementations that don't update still pass the tests.

FrankYFTang:
tests added. PTAL

sffc (approved):

ptomato (approved):

PR #3613, opened 2022-07-23
Test primitive retval of Iterator and AsyncIterator getters

### LINKED ISSUE #3489: Missing coverage: Primitive this value in %IteratorPrototype%[@@iterator]
...and possibly others, there's a handful of these functions that simply return the this value unaltered.

We've had a bug in our engine [1] where the function would additionally attempt to coerce the value to an object by mistake. Since this is a bug that wasn't caught by test262 as the test [2] only checks with an object this value, and this kind of thing is somewhat common across various built-in functions, I think testing two more scenarios would be helpful:

- null/undefined are allowed and returned without throwing
- Other primitive values are not wrapped in an object but returned as is

1: https://github.com/SerenityOS/serenity/commit/472ff7a6d44b4d32679af00927824e7f9d6bdf93
2: https://github.com/tc39/test262/blob/main/test/built-ins/IteratorPrototype/Symbol.iterator/return-val.js

Zirak:
The existing test passed an object for a `this` value. This commit checks that
primitive values are returned verbatim, without any conversion.

Resolve #3489.

### DIFF
--- a/test/built-ins/AsyncIteratorPrototype/Symbol.asyncIterator/return-val.js
+++ b/test/built-ins/AsyncIteratorPrototype/Symbol.asyncIterator/return-val.js
@@ -10,11 +10,22 @@ features: [Symbol.asyncIterator, async-iteration]
 ---*/
 
 async function* generator() {}
-var AsyncIteratorPrototype = Object.getPrototypeOf(Object.getPrototypeOf(generator.prototype))
+const AsyncIteratorPrototype = Object.getPrototypeOf(Object.getPrototypeOf(generator.prototype))
+const getAsyncIterator = AsyncIteratorPrototype[Symbol.asyncIterator];
 
-var thisValue = {};
+const thisValues = [
+    {},
+    Symbol(),
+    4,
+    4n,
+    true,
+    undefined,
+    null,
+];
 
-assert.sameValue(
-  AsyncIteratorPrototype[Symbol.asyncIterator].call(thisValue),
-  thisValue
-);
+for (const thisValue of thisValues) {
+    assert.sameValue(
+        getAsyncIterator.call(thisValue),
+        thisValue
+    );
+}
--- a/test/built-ins/IteratorPrototype/Symbol.iterator/return-val.js
+++ b/test/built-ins/IteratorPrototype/Symbol.iterator/return-val.js
@@ -1,19 +1,32 @@
 // Copyright (C) 2015 the V8 project authors. All rights reserved.
 // This code is governed by the BSD license found in the LICENSE file.
 /*---
-es6id: 25.1.2.1
-description: Descriptor for `name` property
+esid: sec-%iteratorprototype%-@@iterator
+description: Return value of @@iterator on IteratorPrototype
 info: |
+  %IteratorPrototype% [ @@iterator ] ( )
     1. Return the this value.
 features: [Symbol.iterator]
 ---*/
 
-var IteratorPrototype = Object.getPrototypeOf(
+const IteratorPrototype = Object.getPrototypeOf(
   Object.getPrototypeOf([][Symbol.iterator]())
 );
-var thisValue = {};
+const getIterator = IteratorPrototype[Symbol.iterator];
 
-assert.sameValue(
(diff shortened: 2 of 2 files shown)


test/built-ins/AsyncIteratorPrototype/Symbol.asyncIterator/return-val.js
```diff
@@ -10,11 +10,22 @@ features: [Symbol.asyncIterator, async-iteration]
...
+    true,
+    undefined,
+    null,
+];
 
-assert.sameValue(
-  AsyncIteratorPrototype[Symbol.asyncIterator].call(thisValue),
-  thisValue
-);
+for (const thisValue of thisValues) {
```
Zirak, line 26: Is this an ok way of iterating? The repository contains both `for of` and `forEach`:

https://github.com/tc39/test262/blob/6833180066733cce4789d290ddbdfcf6e309226a/test/built-ins/Temporal/TimeZone/basic.js#L25-L29
https://github.com/tc39/test262/blob/6833180066733cce4789d290ddbdfcf6e309226a/test/intl402/PluralRules/prototype/select/non-finite.js#L20-L22

It felt like newer tests leaned more towards `for of`. Is that something to preserve, or eh, whichever way works?

test/built-ins/IteratorPrototype/Symbol.iterator/return-val.js
```diff
@@ -1,19 +1,32 @@
 // Copyright (C) 2015 the V8 project authors. All rights reserved.
 // This code is governed by the BSD license found in the LICENSE file.
 /*---
-es6id: 25.1.2.1
-description: Descriptor for `name` property
+esid: sec-%iteratorprototype%-@@iterator
```
Zirak, line 4: Is this the correct way of going about this? I copied directly from the `AsyncIterator` test above, did I miss something?

test/built-ins/AsyncIteratorPrototype/Symbol.asyncIterator/return-val.js
```diff
@@ -10,11 +10,22 @@ features: [Symbol.asyncIterator, async-iteration]
 ---*/
 
 async function* generator() {}
-var AsyncIteratorPrototype = Object.getPrototypeOf(Object.getPrototypeOf(generator.prototype))
+const AsyncIteratorPrototype = Object.getPrototypeOf(Object.getPrototypeOf(generator.prototype))
```
Zirak, line 13: I took some liberty in converting the `var` to `const`; is that ok? Would it be preferable to keep them as `var`s or turn them to `let`s or whichever way is fine?

Zirak (commented):
Even on this small a PR I have a few concerns; do tell if they're non-issues. Thanks for your time!

test/built-ins/AsyncIteratorPrototype/Symbol.asyncIterator/return-val.js
ptomato: As far as I know, we don't have a policy either way, so either is fine. (If we did, I think it should be a linter rule and not dependent on reviewers remembering it, anyway)

test/built-ins/IteratorPrototype/Symbol.iterator/return-val.js
ptomato: The `esid` is supposed to be the same as the HTML anchor of the relevant section of the spec. [(docs)](https://github.com/tc39/test262/blob/main/CONTRIBUTING.md#esid) In this case, it's https://tc39.es/ecma262/#sec-%iteratorprototype%-@@iterator, so this value looks correct.

test/built-ins/AsyncIteratorPrototype/Symbol.asyncIterator/return-val.js
ptomato: I don't think this is a problem either way — personally I prefer `const` in new code but again this is something that we should codify in a linter rule if we really care about it.

ptomato (approved):
Thanks! We love it when people fill the gaps in coverage like this, much appreciated.

PR #3616, opened 2022-07-26
Tests for Temporal time zones with legacy IANA names

ptomato:
This adds tests for https://github.com/tc39/proposal-temporal/pull/2292 which is a normative change to the Temporal proposal that reached consensus at the July 2022 TC39 meeting.

### DIFF
--- a/test/intl402/Temporal/TimeZone/from/iana-legacy-names.js
+++ b/test/intl402/Temporal/TimeZone/from/iana-legacy-names.js
@@ -1,0 +1,24 @@
+// Copyright (C) 2022 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.timezone
+description: IANA legacy names must be supported
+features: [Temporal]
+---*/
+
+const legacyNames = [
+  ["Etc/GMT0", "UTC"],
+  ["GMT0", "UTC"],
+  ["GMT-0", "UTC"],
+  ["GMT+0", "UTC"],
+  ["EST5EDT", "EST5EDT"],
+  ["CST6CDT", "CST6CDT"],
+  ["MST7MDT", "MST7MDT"],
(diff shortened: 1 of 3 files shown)


test/intl402/Temporal/TimeZone/from/iana-legacy-names.js
```diff
@@ -0,0 +1,24 @@
...
+const legacyNames = [
+  ["Etc/GMT0", "UTC"],
+  ["GMT0", "UTC"],
+  ["GMT-0", "UTC"],
+  ["GMT+0", "UTC"],
+  ["EST5EDT", "EST5EDT"],
+  ["CST6CDT", "CST6CDT"],
+  ["MST7MDT", "MST7MDT"],
+  ["PST8PDT", "PST8PDT"],
+];
```
jessealama, line 19: Are these the only legacy names? The EST/CST/PST examples seem like they're just a small sampling of the possible legacy names. I took a look at one of the IANA deprecated timezone databases (`zone.tab`) and find a bunch of additional timezones (`Pacific/Pago_Pago`, `America/Grenada`, to name just two of hundreds of examples). Should these (or some of them) be included in this test case?

If this is indeed a complete list of legacy names, perhaps that could be usefully added as a comment.

anba: Yes, those are the only "legacy names". From <https://data.iana.org/time-zones/theory.html#naming>:
> Older versions of this package defined legacy names that are incompatible with the first guideline of location names, but which are still supported. These legacy names are mostly defined in the file 'etcetera'. Also, the file 'backward' defines the legacy names 'GMT0', 'GMT-0' and 'GMT+0', and the file 'northamerica' defines the legacy names 'EST5EDT', 'CST6CDT', 'MST7MDT', and 'PST8PDT'. 

These names don't follow the normal guidelines for time zone identifiers and were therefore categorised as "legacy names".

jessealama (approved):

PR #3617, opened 2022-07-27
Update tests which used negative day lengths

anba:
Negative day lengths are no longer valid after <https://github.com/tc39/proposal-temporal/pull/2261>.


ptomato (approved):
Thanks for updating!

PR #3619, opened 2022-07-28
Implement changes for "avoid mostly-redundant `await` in async `yield*`"

bakkot:
These are updates for https://github.com/tc39/ecma262/pull/2819.

I made the relevant modification to engine262 in a local branch, ran all of test262, and only found one test whose behavior needed to change, which was exercising the number of ticks implied by `yield*` in an async generator. This PR modifies that test and also adds an explicit test for that `yield*` does not perform unwrapping of the inner value.

I recommend reviewing commits individually - in particular, the first commit only updates the existing test's `info` to match the _current_ spec, which is then modified to match https://github.com/tc39/ecma262/pull/2819 in the second commit. It's hard to see the change otherwise.

cc @anba who wrote the test which is being updated.


test/language/statements/async-generator/yield-star-promise-not-unwrapped.js
```diff
@@ -0,0 +1,41 @@
...
+    return {
+      done: false,
+      value: innerPromise,
+    };
+  },
+  get return() {
+    throw new Test262Error(".return should not be accessed");
+  },
+  get throw() {
+    throw new Test262Error(".throw should not be accessed");
```
jugglinmike, line 28: In cases like this, I prefer to observe execution with "flag" variables that are then asserted at the end of the test. Reason being: the exception handling mechanisms are themselves being validated, so there's a risk that whatever bug causes this statement to be evaluated would also cause the failure not to be reported.

That said, async tests are fundamentally dependent on asynchronous exception handling semantics, so I don't feel this risk is large enough to warrant another round of review. (I'm sharing my perspective anyway in case it's relevant for future tests and in case anyone out there can explain why I should reconsider.)

test/language/statements/async-generator/yield-star-promise-not-unwrapped.js
```diff
@@ -0,0 +1,41 @@
+// Copyright (C) 2022 André Bargull. All rights reserved.
```
jugglinmike, line 1: Is this test intentionally attributed to André?

jugglinmike (changes requested):
Looks good to me! I just want to make sure the license is accurate before merging.

test/language/statements/async-generator/yield-star-promise-not-unwrapped.js
bakkot: Whoops, no, I just forgot to update it when I copied an existing test.

Ms2ger (approved):

Ms2ger (approved):

PR #3620, opened 2022-07-28
Temporal: Add tests for an empty or a function options object

Aditi-1400:


ptomato (approved):
Thanks! Of course, I didn't review each one individually, but I looked at a representative sample.

PR #3622, opened 2022-07-29
Temporal: Porting some more of proposal-temporal tests

ptomato:
Recommend reviewing commit by commit; the tests within each commit are very similar.

### DIFF
--- a/test/built-ins/Temporal/Calendar/prototype/weekOfYear/argument-propertybag-calendar-string.js
+++ b/test/built-ins/Temporal/Calendar/prototype/weekOfYear/argument-propertybag-calendar-string.js
@@ -1,0 +1,16 @@
+// Copyright (C) 2022 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.calendar.prototype.weekofyear
+description: A calendar ID is valid input for Calendar
+features: [Temporal]
+---*/
+
+const instance = new Temporal.Calendar("iso8601");
+
+const calendar = "iso8601";
+
+const arg = { year: 1976, monthCode: "M11", day: 18, calendar };
+const result = instance.weekOfYear(arg);
+assert.sameValue(result, 47, `Calendar created from string "${calendar}"`);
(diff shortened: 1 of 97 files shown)


test/built-ins/Temporal/ZonedDateTime/prototype/with/offset-property-invalid-string.js
```diff
@@ -0,0 +1,29 @@
+// Copyright (C) 2022 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.zoneddatetime.prototype.equals
```
Aditi-1400, line 5: Should be: `esid: sec-temporal.zoneddatetime.prototype.with`

Aditi-1400:
Thanks. 
I didn't find anything else, looks good to me! 👍🏽

test/built-ins/Temporal/Calendar/prototype/weekOfYear/argument-propertybag-calendar-string.js
```diff
@@ -0,0 +1,16 @@
...
+features: [Temporal]
+---*/
+
+const instance = new Temporal.Calendar("iso8601");
+
+const calendar = "iso8601";
+
+const arg = { year: 1976, monthCode: "M11", day: 18, calendar };
+const result = instance.weekOfYear(arg);
+assert.sameValue(result, 47, `Calendar created from string "${arg}"`);
```
jessealama, line 16: Perhaps we should some something more like 

````
`Calendar created from ${typeof arg} "${arg}"`
````

? In this case, I guess it's not a string.

ptomato: I think I intended it to be `Calendar created from string ${calendar}` in all these tests.

ptomato: (at least, the property bag ones)

ptomato:
Thanks for the reviews. I've pushed an update.

Aditi-1400 (approved):

PR #3623, opened 2022-07-29
Temporal: Adjust and expand tests for observable calls to ToString(calendar)

ptomato:
This implements the normative change in
https://github.com/tc39/proposal-temporal/pull/2269 which reached
consensus at the July 2022 TC39 meeting.

There was already a test for PlainDate for this topic, which needs to be
adjusted to accommodate the normative change. Tests for PlainDateTime and
ZonedDateTime did not yet exist, so add new ones based on the PlainDate
test.


ptomato:
cc @anba

Aditi-1400 (approved):

PR #3624, opened 2022-07-31
ShadowRealm: Relax requirements of globalThis available properties test

linusg:
The intention of this test is to ensure that all built-in properties of the global object are also exposed on the `ShadowRealm`'s global object, without penalizing implementations that don't have all of them implemented.

Notably, `SharedArrayBuffer` may still not be (re-)enabled in all circumstances.

cc @leobalter (original author of this test)

### DIFF
--- a/test/built-ins/ShadowRealm/prototype/evaluate/globalthis-available-properties.js
+++ b/test/built-ins/ShadowRealm/prototype/evaluate/globalthis-available-properties.js
@@ -35,7 +35,7 @@ assert.sameValue(
 
 const r = new ShadowRealm();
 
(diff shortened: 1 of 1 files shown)


Ms2ger (approved):

leobalter:
Thanks for quoting me. I no longer participate in Test262 maintenance, but I appreciate the mention as I'm a co-champion of the ShadowRealm proposal. 

This test has no blockers for the proposal itself, so acceptance of the test strategy goes to the discretion of the Test262 maintainers. It all seems be fine anyway as the PR is now merged.

PR #3627, opened 2022-08-01
Temporal: Add test for Temporal.Duration.prototype.total

Aditi-1400:
This adds test for https://github.com/tc39/proposal-temporal/pull/2298 which is a normative change to the Temporal proposal that reached consensus at the July 2022 TC39 meeting.


ptomato (approved):
Thanks!

PR #3632, opened 2022-08-03
Rename Array#groupBy and Array#groupByToMap to Array#group and Array#groupToMap

Constellation:
TC39 decided to rename them to avoid web-compat issue[1].
This change follows to it.

[1]: https://github.com/tc39/proposal-array-grouping/pull/39


ptomato (approved):
Thanks! This change indeed matches that rename, and I verified there are not yet any other instances of it.

PR #3633, opened 2022-08-04
Add exhaustive test for String.prototype.substr with various number inputs

anba:
Also see: https://github.com/tc39/ecma262/pull/2844


anba:
I wrote a SpiderMonkey patch which had an incorrect change to `String.prototype.substr`. Test262 didn't uncover the bug, only some other unrelated Firefox test triggered it. So I thought it's time to add a more exhaustive test for `substr`. :smile:

test/annexB/built-ins/String/prototype/substr/start-and-length-as-numbers.js
```diff
@@ -0,0 +1,92 @@
...
+  // Step 10.
+  let intEnd = Math.min(intStart + intLength, size);
+
+  // |intStart <= intEnd <= size| now holds.
+  assert(intStart <= intEnd && intEnd <= size);
+
+  // Step 11.
+  //
+  // Call `substring` and check the result is correct.
+  let result = string.substring(intStart, intEnd);
```
ljharb, line 57: we should probably not assume that a substr implementation does not delegate to substring in some way?

anba: I think that's okay, because even if it does, the next lines will verify the correct result was computed.

test/annexB/built-ins/String/prototype/substr/start-and-length-as-numbers.js
```diff
@@ -0,0 +1,92 @@
...
+  if (intStart === -Infinity) {
+    intStart = 0;
+  } else if (intStart < 0) {
+    intStart = Math.max(size + intStart, 0);
+  } else {
+    intStart = Math.min(intStart, size)
+  }
+
+  // |0 <= intStart <= size| now holds.
+  assert(0 <= intStart && intStart <= size);
```
ptomato, line 37: I have a slight preference for turning these comments into assertion messages, e.g.
```js
assert(0 <= intStart && intStart <= size, "0 <= intStart <= size now holds");
```
That would've made it easier for me to debug when I was confused about why the implementation was different than in the spec. But it's a personal preference, I wouldn't block merging the PR on that.

ptomato (approved):
Took me a while to realize that the reference implementation of StringSubstr is that from tc39/ecma262#2844, not the one currently in ECMA-262. Once I realized that, I think this is correct.

I'd be OK with merging it now instead of waiting for tc39/ecma262#2844 to merge. Realistically, there can't be any other outcome to that than reverting the accidental normative change, because it is definitely web reality. However, we might want to update the comments in this reference implementation to match the step numbers in case the result there turns out to be worded slightly differently.

ptomato:
I approved this a while ago, no other reviews after mine, and Jordan's comment has been addressed; I think we should merge it.

PR #3636, opened 2022-08-08
Meta: Link to engine262 instructions

gibson042:


ljharb (approved):

ptomato (approved):
This is probably going to be quite helpful for authors of proposals that don't have any implementations yet and aren't polyfillable!

Issue #3476, opened 2022-04-11
Consider alternatives to deepEqual


ptomato:
We briefly discussed in the maintainers' meeting of 2022-04-07, the `deepEqual` file in the harness has some problems in its semantics and we should at least encourage new tests not to use it, while working out a better way to provide that functionality.

I've taken a look at what uses it and identified three categories:
1. Comparing two arrays with primitive values in them ­— it seems to me this can be replaced wholesale by `compareArray`, which is clearer and more fit for purpose anyway.
2. Comparing two objects with primitive values as their keys, where order of assignment doesn't matter.
3. Comparing two arrays with non-primitive values in them.

Here are the places where it's currently used:
- [Error/constructor.js](https://github.com/tc39/test262/blob/9055521fc7ea8d44a34fd4fe137f2c25c60655ee/test/built-ins/Error/constructor.js) - (1) - I've changed this already in https://github.com/tc39/test262/pull/3475
- [RegExp/match-indices/indices-array-non-unicode-match.js](https://github.com/tc39/test262/blob/9055521fc7ea8d44a34fd4fe137f2c25c60655ee/test/built-ins/RegExp/match-indices/indices-array-non-unicode-match.js), [indices-array-unicode-match.js](https://github.com/tc39/test262/blob/9055521fc7ea8d44a34fd4fe137f2c25c60655ee/test/built-ins/RegExp/match-indices/indices-array-unicode-match.js) - (3)
- [Temporal/Calendar/prototype/mergeFields/basic.js](https://github.com/tc39/test262/blob/9055521fc7ea8d44a34fd4fe137f2c25c60655ee/test/built-ins/Temporal/Calendar/prototype/mergeFields/basic.js), [iso8601-calendar-month-monthCode.js](https://github.com/tc39/test262/blob/9055521fc7ea8d44a34fd4fe137f2c25c60655ee/test/built-ins/Temporal/Calendar/prototype/mergeFields/iso8601-calendar-month-monthCode.js), [non-string-properties.js](https://github.com/tc39/test262/blob/9055521fc7ea8d44a34fd4fe137f2c25c60655ee/test/built-ins/Temporal/Calendar/prototype/mergeFields/non-string-properties.js) - (2)
[...]

PR #3638, opened 2022-08-11
Temporal: Add intl in staging directory

Aditi-1400:


ptomato (approved):
We might want to refine these in the future with a [`locale`](https://github.com/tc39/test262/blob/main/CONTRIBUTING.md#locale) tag in the frontmatter, especially if some test runners end up not having the locales available to run them. We'll also probably need to [remove `deepEqual`](https://github.com/tc39/test262/issues/3476) before porting the tests into the main tree.

But, that can all be improved while these tests are in staging! I've verified that these are executable under the test262 harness, so merging now.

FrankYFTang:
Could you explain what the directory "test/staging/Intl402/Temporal/old" intend for? I understand what the staging means, but what does the "old" means?

ptomato:
Folders in staging don't have any requirements for the name and can be organized in any way. In this case it means the "old" format, with tests grouped together in one file and not yet converted into individual test files.

Issue #3594, opened 2022-06-30
DurationFormat's constructor-options-defaults.js assumes incorrect default values


IdanHo:
The default values used for the options tested are from the "digital default" column of Table 1, but as the name suggests, these defaults are only used when the DurationFormat style is "digital", since the default style is "long", these do not apply.

ptomato:
Closed by #3640. Thanks for reporting it!

PR #3640, opened 2022-08-13
Fix DurationFormat default tests

FrankYFTang:
1. add the test for "seconds"
2. since the default value for "style" is "long", baseStyle for GetDurationUnitOptions is "long" and therefore the 
last argument in testOption should be "long"
3. the valid values for "days" does not contains  "numeric", and "2-digit". remove them. 
4. the valid values for "milliseconds", "microseconds" and "nanoseconds" does not contains "2-digit". remove it.

See https://tc39.es/proposal-intl-duration-format/#table-duration-components about the valid value
notice the last colum is for "Digital Default" while the baseStyle is "digital" but the set up does not set it that way, the default value for "style" is "long" as in
```
13. Let style be ? GetOption(options, "style", "string", « "long", "short", "narrow", "digital" », "long").
```
of https://tc39.es/proposal-intl-duration-format/#sec-Intl.DurationFormat

@sffc @romulocintra  @ryzokuken

### DIFF
--- a/test/intl402/DurationFormat/constructor-options-defaults.js
+++ b/test/intl402/DurationFormat/constructor-options-defaults.js
@@ -20,13 +20,14 @@ features: [Intl.DurationFormat]
 includes: [testIntl.js]
 ---*/
 
-testOption( Intl.DurationFormat, "years", "string", ["long", "short", "narrow"], "narrow");
-testOption( Intl.DurationFormat, "months", "string", ["long", "short", "narrow"], "narrow");
-testOption( Intl.DurationFormat, "weeks", "string", ["long", "short", "narrow"], "narrow");
-testOption( Intl.DurationFormat, "days", "string", ["long", "short", "narrow", "numeric", "2-digit"], "numeric");
-testOption( Intl.DurationFormat, "hours", "string", ["long", "short", "narrow", "numeric", "2-digit"], "numeric");
-testOption( Intl.DurationFormat, "minutes", "string", ["long", "short", "narrow", "numeric", "2-digit"], "numeric");
-testOption( Intl.DurationFormat, "milliseconds", "string", ["long", "short", "narrow", "numeric", "2-digit"], "numeric");
-testOption( Intl.DurationFormat, "microseconds", "string", ["long", "short", "narrow", "numeric", "2-digit"], "numeric");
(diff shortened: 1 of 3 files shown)


ptomato:
@FrankYFTang Is this an implementation of a normative PR or a correction to tests that were previously incorrect?

FrankYFTang:
> a correction to tests that were previously incorrect?

It is a correction to tests that were previously incorrect to the current spec text. 
As I mention,
First, it does not include the test of "seconds" (This should be obvious)
Second, the test include the value for  of "numeric", "2-digit" for "days" but in https://tc39.es/proposal-intl-duration-format/#sec-intl-durationformat-abstracts it clearly show "numeric", "2-digit"  are not valid values for "days"
Third, the test include the value for  of  "2-digit" for  "milliseconds",  "microseconds", and "nanoseconds" but in https://tc39.es/proposal-intl-duration-format/#sec-intl-durationformat-abstracts it clearly shows "2-digit"  is not a valid value for "milliseconds",  "microseconds", and "nanoseconds"
Fourth, it valid the default for all these values are "narrow" but that is incorrect, the default is "long" because the default for "style" is "long", and that became the default for all the fieds. 

@ryzokuken

FrankYFTang:
ping

FrankYFTang:
ping @ryzokuken @romulocintra @sffc

sffc (dismissed):

romulocintra (dismissed):

ryzokuken (dismissed):

romulocintra:
Closes https://github.com/tc39/test262/issues/3594

trflynn89:
Now that the default style is `short` as of https://github.com/tc39/proposal-intl-duration-format/commit/b28949490bd0bea23e09e31bdeb8847e996482f4, the fallback used in these tests should be updated to be `short` instead of `long`, right?

FrankYFTang:
> Now that the default style is `short` as of [tc39/proposal-intl-duration-format@b289494](https://github.com/tc39/proposal-intl-duration-format/commit/b28949490bd0bea23e09e31bdeb8847e996482f4), the fallback used in these tests should be updated to be `short` instead of `long`, right?

yes

ptomato (approved):
Thanks, these changes all seem to check out with the proposal text.

PR #3645, opened 2022-08-25
Temporal: Add calendar strings to instant string tests

Aditi-1400:
This adds tests for the normative change: https://github.com/tc39/proposal-temporal/pull/2345

These tests ensure that while parsing an Instant string, calendar annotation is ignored.


ptomato (approved):
Thanks!

Issue #3253, opened 2021-10-08
Please move Temporal tests with IANA timezone other than UTC under intl402


FrankYFTang:
Temporal tests which depends on IANA TimeZone other than UTC should be moved from builtin-ins to intl402 directory

 find Temporal/|xargs egrep -s "Asia|America|Europe|Africa"|cut -d':' -f1|sort -u

Temporal/Duration/compare/timezone-string-datetime.js
Temporal/Duration/prototype/add/relativeto-string-datetime.js
Temporal/Duration/prototype/add/timezone-string-datetime.js
Temporal/Duration/prototype/round/relativeto-string-datetime.js
Temporal/Duration/prototype/round/timezone-string-datetime.js
Temporal/Duration/prototype/subtract/relativeto-string-datetime.js
Temporal/Duration/prototype/subtract/timezone-string-datetime.js
Temporal/Duration/prototype/total/relativeto-string-datetime.js
Temporal/Duration/prototype/total/timezone-string-datetime.js
Temporal/Instant/compare/instant-string.js
Temporal/Instant/from/instant-string.js
Temporal/Instant/prototype/equals/instant-string.js
Temporal/Instant/prototype/since/instant-string.js
Temporal/Instant/prototype/toString/timezone-string-datetime.js
Temporal/Instant/prototype/toZonedDateTimeISO/timezone-string-datetime.js
Temporal/Instant/prototype/toZonedDateTime/timezone-string-datetime.js
Temporal/Instant/prototype/until/instant-string.js
Temporal/Now/plainDateISO/timezone-string-datetime.js
Temporal/Now/plainDateTimeISO/timezone-string-datetime.js
Temporal/Now/plainDateTime/timezone-string-datetime.js
Temporal/Now/plainDate/timezone-string-datetime.js
Temporal/Now/plainTimeISO/timezone-string-datetime.js
Temporal/Now/zonedDateTimeISO/timezone-string-datetime.js
Temporal/Now/zonedDateTime/timezone-string-datetime.js
Temporal/PlainDate/prototype/toZonedDateTime/basic.js
Temporal/PlainDate/prototype/toZonedDateTime/timezone-string-datetime.js
Temporal/PlainDateTime/prototype/toZonedDateTime/timezone-string-datetime.js
Temporal/PlainTime/prototype/toZonedDateTime/basic.js
Temporal/PlainTime/prototype/toZonedDateTime/timezone-string-datetime.js
Temporal/TimeZone/from/argument-primitive.js
[...]

PR #3646, opened 2022-08-25
Move/split non UTC part of tz tests to intl402

FrankYFTang:
Temporal tests which depends on IANA TimeZone other than UTC should be moved from builtin-ins to intl402 directory

find Temporal/|xargs egrep -s "Asia|America|Europe|Africa"|cut -d':' -f1|sort -u
```
Temporal/Instant/from/argument-string.js
Temporal/PlainDate/from/argument-string.js
Temporal/PlainDateTime/from/argument-string-timezone.js
Temporal/PlainDateTime/prototype/toPlainTime/basic.js
```


FrankYFTang:
address https://github.com/tc39/test262/issues/3253 @ptomato

ptomato (changes requested):
IMO it's not correct for these tests to exist only in intl402. In all of these cases, the tests verify that the time zone annotation must be ignored. So these tests don't depend on a time zone database.

PR #3651, opened 2022-08-31
test: AssignmentExpression where LeftHandSideExpression is CallExpresion

### LINKED ISSUE #3650: Question - assignment of CallExpression
consider the following code
```js
function $(arg) { return arg }
$(window) = $(window).test
```
using chrome/nodejs this code throws during execution, parsing works.

based on the spec
https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#sec-assignment-operators

image
image

so the syntax of a CallExpression on the left side should be parsable.

what is the expected behavior here?

based on discussion in https://github.com/acornjs/acorn/issues/1145

Meir017:
closes #3650
a missing scenario from https://tc39.es/ecma262/#sec-assignment-operators-static-semantics-early-errors

### DIFF
--- a/test/language/expressions/assignment/target-callexpression.js
+++ b/test/language/expressions/assignment/target-callexpression.js
@@ -1,0 +1,22 @@
+// Copyright (c) 2012 Ecma International.  All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-assignment-operators-static-semantics-early-errors
+description: >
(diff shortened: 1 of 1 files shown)


anba:
This is already covered by <https://github.com/tc39/test262/blob/main/test/language/expressions/assignmenttargettype/direct-callexpression-arguments.js>.

Meir017:
@anba I think that test should include `esid: sec-assignment-operators-static-semantics-early-errors` in it

PR #3353, opened 2021-12-15
add tests for array.prototype.groupToMap

polsevev:
Added tests for proposal Array Grouping 

CC @codehag

PR #3652, opened 2022-09-01
Add group/groupToMap tests

jridgewell:
Forking from #3354 and #3353, preserving the original commits and adding my own on top.


jridgewell:
Got write access to update the original PRs.

PR #3659, opened 2022-09-13
RangeError when useGrouping is string 'undefined'

FrankYFTang:
per https://github.com/tc39/proposal-intl-numberformat-v3/pull/107

### DIFF
--- a/test/intl402/NumberFormat/test-option-useGrouping-extended.js
+++ b/test/intl402/NumberFormat/test-option-useGrouping-extended.js
@@ -35,6 +35,8 @@ assert.sameValue(render({notation: 'compact', useGrouping: false}), false, 'comp
 assert.sameValue(render({notation: 'compact', useGrouping: null}), false, 'compact, null');
 assert.sameValue(render({notation: 'compact', useGrouping: 'min2'}), 'min2', 'compact, "min2"');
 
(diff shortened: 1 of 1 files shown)


FrankYFTang:
@romulocintra @sffc @ryzokuken

test/intl402/NumberFormat/test-option-useGrouping-extended.js
```diff
@@ -35,6 +35,8 @@ assert.sameValue(render({notation: 'compact', useGrouping: false}), false, 'comp
 assert.sameValue(render({notation: 'compact', useGrouping: null}), false, 'compact, null');
 assert.sameValue(render({notation: 'compact', useGrouping: 'min2'}), 'min2', 'compact, "min2"');
 
-assert.sameValue(render({useGrouping: 'undefined'}), 'auto', 'use fallback value');
+assert.throws(RangeError, function () {
```
romulocintra, line 38: Thanks for catching this @FrankYFTang  I would rather suggest deleting this assertion and add "undefined" string at https://github.com/tc39/test262/blob/main/test/intl402/NumberFormat/test-option-useGrouping.js#L34

ptomato:
Do I understand correctly that we should merge either #3660 or this PR, but not both?

FrankYFTang:
Close this PR and yield to  #3660

PR #3660, opened 2022-09-13
Update:RangeError when useGrouping is string 'undefined'

romulocintra:
Applied changes related with https://github.com/tc39/test262/pull/3659


ptomato:
Do I understand correctly that we should merge either #3659 or this PR, but not both?

ptomato (approved):
OK, thanks!

PR #3663, opened 2022-09-14
Update RegExp `\p{…}` tests per Unicode 15

mathiasbynens:
Upstream commit: https://github.com/mathiasbynens/unicode-property-escapes-tests/commit/859db0d6df0664244f6b53ca8da1015daa3be572

Issue: https://github.com/tc39/ecma262/issues/2904


trflynn89:
If it helps for reviewers, I've verified that these all pass with [SerenityOS's implementation](https://github.com/SerenityOS/serenity/tree/master/Userland/Libraries/LibJS) when we upgrade to Unicode 15:
https://github.com/SerenityOS/serenity/pull/15230

PR #3664, opened 2022-09-14
Fix features in tests from #3354

ljharb:
Rather than add `arrow-function`, i just changed all the arrows to normal functions, since nothing was relying on any arrow-specific differences.


ptomato (approved):
I'd say that `array-grouping` should automatically imply `Map` since you can't implement the proposal if your implementation doesn't support Map, but I certainly won't object if that makes it easier to run test262 on packages.

ljharb:
You can implement `Array.prototype.group` without `Map`, though.

PR #3668, opened 2022-09-15
Fix limit test

FrankYFTang:
The doAsserts is not correct for case like "P100000000000000000000000000Y" If we test "P100000000000000000000000000Y" because the filed in Duration is only required to  "A float64-representable integer is an integer that is exactly representable as a Number value. That is, for a float64-representable integer x, it must hold that ℝ(𝔽(x)) = x." per  https://tc39.es/proposal-temporal/#sec-properties-of-temporal-duration-instances
If there is a positive numeric error, the string will be like "P100000000000000000000001234Y" but we ma also have a negative numeric error, which will have the string be something like "P99999999999999987584860160Y" 
.  
Even the test() only test the first 10 digits ( log(10^10)/log(2) = 33 bits) of the formated the string, this will still not match. In this PR, I change the test to test  

"P100000000000001000000000000Y" instead (Notice the "1" after 13 '0' right to the "P1") , then a negative numerical error will not change the start of the string  from "P1000000000" to "P9999999999"

@anba @ptomato @ryzokuken @romulocintra @Aditi-1400

### DIFF
--- a/test/staging/Temporal/Duration/old/limits.js
+++ b/test/staging/Temporal/Duration/old/limits.js
@@ -78,7 +78,7 @@ function test(ix, prefix, suffix, infix = "") {
   doAsserts(new Temporal.Duration(...Array(ix).fill(0), 1e+26, ...Array(9 - ix).fill(0)));
   doAsserts(Temporal.Duration.from({ [units[ix]]: 1e+26 }));
   if (!infix)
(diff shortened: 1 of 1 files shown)


ptomato (approved):
OK, thanks. We will definitely need to make a more robust test for this when we move it from staging into the main tree, but avoiding the problem you mentioned for now makes sense to me.

PR #3655, opened 2022-09-07
Temporal: Add Temporal calendar and timezone case insensitive tests

Aditi-1400:
This PR adds tests for the normative change: https://github.com/tc39/proposal-temporal/pull/2395

PR #3670, opened 2022-09-17
Split withCalendar/calendar-case-insensitive.js to intl402

FrankYFTang:
Tests depends on calendar other than "iso8601" in AvailableCalendars() is not required to supported by ECMA262 and therefore should be moved into intl402 directory.

@Aditi-1400 @ptomato @romulocintra


Ms2ger:
Let's not move the test but make it check "IsO8601" instead, like the other tests in #3655.

FrankYFTang:
@Ms2ger  how about having both?

Ms2ger (approved):
Excellent!

PR #3671, opened 2022-09-19
Fix resizable ArrayBuffer subarray tests

syg:
This is from the normative change in
https://github.com/tc39/proposal-resizablearraybuffer/pull/93, which got consensus in the March 2022 TC39.

### DIFF
--- a/test/staging/ArrayBuffer/resizable/subarray-parameter-conversion-grows.js
+++ b/test/staging/ArrayBuffer/resizable/subarray-parameter-conversion-grows.js
@@ -127,7 +127,8 @@ for (let ctor of ctors) {
       return 0;
     }
   };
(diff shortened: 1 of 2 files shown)


syg:
@ptomato PTAL, thanks.

ptomato (approved):
Both seem correct to me, thanks. The one in the "grows" test does implement a change due to that normative PR which I missed the first time around. The one in the "shrinks" test seems to me that it should pass either way, by my reading, but maybe I missed something — it's been a couple of weeks since I last looked at these files.

syg:
Thanks for the review! At the time you converted the V8 tests, the V8 tests were also incorrect. We had forgotten to implement the normative change in https://github.com/tc39/proposal-resizablearraybuffer/pull/93 until recently, which is why these were surfaced.

PR #3673, opened 2022-09-21
CI: Update CircleCI Docker images.

Ms2ger:
CircleCI reports: "You’re using a [deprecated Docker convenience image.](https://discuss.circleci.com/t/legacy-convenience-image-deprecation/41034) Upgrade to a next-gen Docker convenience image."

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,7 +39,7 @@ execution_steps: &execution_steps
 jobs:
   "Test262: verify tools; build & lint tests (Python 2)":
     docker:
-      - image: circleci/python:2.7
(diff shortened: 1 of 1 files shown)


ptomato (approved):
Would be nice to have some context for this change :smile: 
But I found some: https://circleci.com/blog/announcing-our-next-generation-convenience-images-smaller-faster-more-deterministic/
This looks like they have done some work on making the images spin up faster. Sign me up!

ptomato:
I also wonder if we should switch the Python 3 image to one in the 3.10 series but that certainly doesn't have to be done in this PR.

PR #3674, opened 2022-09-26
Temporal: Add tests for observable operations in MoveRelativeDate

ptomato:
See https://github.com/tc39/proposal-temporal/pull/2267 which eliminated some unnecessary lookups of the calendar's dateAdd method from the MoveRelativeDate AO, which is called in the calendar types' since() and until() methods, Duration.p.round(), and Duration.p.total().

This adds tests for the order of all observable operations for these 10 methods, not just the lookups of dateAdd on the calendar. (These methods needed to have their order of observable operations tested anyway.)

They heavily use the TemporalHelpers.calendarObserver and TemporalHelpers.timeZoneObserver added in the first two commits of this PR.

https://github.com/tc39/proposal-temporal/pull/2267 included changes to several code paths, not all of which are reachable through every method tested here; only Duration.p.round() can trigger the full set.

https://github.com/tc39/proposal-temporal/pull/2267 reached consensus at the July 2022 TC39 plenary meeting.

I'd recommend reviewing it commit by commit. The last commit is going to be difficult to review exhaustively because it'll require a lot of clicking back and forth between [RoundDuration](https://tc39.es/proposal-temporal/#sec-temporal-roundduration), [UnbalanceDurationRelative](https://tc39.es/proposal-temporal/#sec-temporal-unbalancedurationrelative), and [BalanceDurationRelative](https://tc39.es/proposal-temporal/#sec-temporal-balancedurationrelative) in the Temporal spec text. (At least, that's what it took to write the tests.) Hopefully the comments I added about what algorithm steps incur what observable operations will help, but still I'd recommend just reviewing a representative sample. It's a lot, and it'll be easy for implementors to verify if I did get something wrong.

### DIFF
--- a/harness/temporalHelpers.js
+++ b/harness/temporalHelpers.js
@@ -1636,6 +1636,44 @@ var TemporalHelpers = {
   },
 
   /*
+   * timeZoneObserver:
+   * A custom calendar that behaves exactly like the UTC time zone but tracks
+   * calls to any of its methods, and Get/Has operations on its properties, by
+   * appending messages to an array. This is for the purpose of testing order of
+   * operations that are observable from user code. objectName is used in the
+   * log. methodOverrides is an optional object containing properties with the
+   * same name as Temporal.TimeZone methods. If the property value is a function
+   * it will be called with the proper arguments instead of the UTC method.
+   * Otherwise, the property value will be returned directly.
+   */
+  timeZoneObserver(calls, objectName, methodOverrides = {}) {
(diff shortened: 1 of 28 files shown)


Aditi-1400 (commented):
I think we should add these tests for `Temporal.Duration.compare` as well.

ptomato:
Glad you caught the case I missed! I've added `Temporal/Duration/compare/order-of-operations.js`.

Ms2ger (approved):

PR #3683, opened 2022-10-08
Temporal: tests for IETF annotations standard

ptomato:
This is another giant pull request, I recommend reviewing it commit by commit again.

Most of the tests in each commit are the same, so I'd recommend reviewing a representative sample in each commit; maybe pay special attention to Calendar.dateUntil, ZonedDateTime.from, ZonedDateTime.equals, and the compare() functions, since those are the ones where I sometimes had to do something out of the ordinary compared to the rest of the tests.

These tests are for the normative changes in https://github.com/tc39/proposal-temporal/pull/2397 which reached consensus at the September 2022 TC39 meeting.


Aditi-1400 (approved):
Thanks for these! :)

Ms2ger (approved):

Issue #3684, opened 2022-10-11
Temporal/Duration/prototype/round/order-of-operations.js has an incorrect assertion


anba:
The test expects that `UnbalanceDurationRelative` performs property lookups, even though [UnbalanceDurationRelative, step 1](https://tc39.es/proposal-temporal/#sec-temporal-unbalancedurationrelative) returns early when `years = months = weeks = days = 0`:
https://github.com/tc39/test262/blob/eb67a7689f422cf14776f78191650d0d78c2d2e4/test/built-ins/Temporal/Duration/prototype/round/order-of-operations.js#L205-L207

The [spec polyfill](https://github.com/tc39/proposal-temporal/blob/3a3a46c15c662977db824e020133be748d6b3317/polyfill/lib/ecmascript.mjs#L2730) doesn't seem to implement the early return from step 1.

@ptomato

ptomato:
Thanks for catching this. The same mistake happens in the following stanza with day-to-week balancing as well. The fix is in https://github.com/tc39/test262/pull/3686.

PR #3686, opened 2022-10-11
Temporal: Fix expectation for observable operations in Duration.p.round()

ptomato:
I had missed that there's an early return step from [UnbalanceDurationRelative](https://tc39.es/proposal-temporal/#sec-temporal-unbalancedurationrelative) if years, months, weeks, and days are all 0.

Closes: #3684


ptomato:
cc @anba

Aditi-1400 (approved):
Thanks for this! 👍

Ms2ger (approved):

PR #3690, opened 2022-10-18
Add missing includes to toSorted/toReversed this-value-invalid.js tests

linusg:
Otherwise $DETACHBUFFER() is not defined in those files.


linusg:
cc @nicolo-ribaudo @catamorphism @Ms2ger

Ms2ger (approved):
Thanks

PR #3694, opened 2022-10-19
Temporal: Check for absence of toString call in Calendar.id and TimeZone.id

ptomato:
This is a test for the normative change https://github.com/tc39/proposal-temporal/pull/2411 which has not yet received consensus in TC39, so review it with that in mind and don't merge it yet.


gibson042 (approved):

ptomato:
The corresponding normative change achieved consensus today in TC39.

PR #3697, opened 2022-10-24
Add `/iu` Unicode case folding tests

mathiasbynens:
Ref. https://bugs.chromium.org/p/v8/issues/detail?id=13377


ptomato (dismissed):
Thanks! After reading the Chromium bug and digging into the CompileToCharSet operation and CaseFolding.txt, I think I understand what's going on here, and I concur. These six code points all have an "F" for Full case mapping in CaseFolding.txt, and so they should not be considered in the Canonicalize operation.

This is fine as-is, but for maximum usefulness, would you consider putting a test in the main tree e.g. somewhere in `test/built-ins/RegExp/`? (Or otherwise in staging so that it can be moved into the main tree eventually) I don't know how many other engines regularly execute the `implementation-contributed/` tests.

ptomato (approved):
Re-approving

ptomato:
I'll merge this and open a new PR to add this to the main tree as well.

mathiasbynens:
> I'll merge this and open a new PR to add this to the main tree as well.

Cheers!

PR #3698, opened 2022-10-24
Temporal: ISO strings improvements

ptomato:
Here are some refactors, and expansion of test coverage, that I did while I was working on something else.


Aditi-1400 (approved):
These changes look good. Thanks :)
Q. Will we be creating a separate PRs for other collections of valid/invalid strings like PlainDate strings etc.?

ptomato:
Indeed, I updated these when I came across them as part of other work. So, I expect PRs for other collections of strings will happen at some point.
Thanks for the review!

gibson042 (approved):

PR #3699, opened 2022-10-24
Temporal: Tests for normative changes around date-only strings

ptomato:
As per the discussion in
https://github.com/tc39/proposal-temporal/issues/2379#issuecomment-1248557100 and the PR https://github.com/tc39/proposal-temporal/pull/2428, which is to be presented for consensus to TC39 in the upcoming plenary meeting, UTC offsets and the Z designator should be disallowed after any date-only strings (YYYY-MM-DD, YYYY-MM, and MM-DD). They should only be allowed to follow a time component. Z remains disallowed in any string being parsed into a Plain type.

Annotations become allowed after any ISO string, even YYYY-MM and MM-DD where they were previously disallowed.

### DIFF
--- a/harness/temporalHelpers.js
+++ b/harness/temporalHelpers.js
@@ -1845,11 +1845,14 @@ var TemporalHelpers = {
     plainTimeStringsAmbiguous() {
       const ambiguousStrings = [
         "2021-12",  // ambiguity between YYYY-MM and HHMM-UU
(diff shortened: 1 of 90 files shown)


Aditi-1400 (approved):

ptomato:
The corresponding normative change achieved consensus in the TC39 plenary meeting today.

Ms2ger (approved):

PR #3700, opened 2022-10-24
Fix toSpliced mutate-while-iterating test

syg:
The `arr[0] = 17` throws in strict mode. I'm not sure what that line is supposed to be testing. This can be either fixed per this PR or by adding a `flags: [noStrict]`.

### DIFF
--- a/test/built-ins/Array/prototype/toSpliced/mutate-while-iterating.js
+++ b/test/built-ins/Array/prototype/toSpliced/mutate-while-iterating.js
@@ -29,10 +29,14 @@ includes: [compareArray.js]
 ---*/
 
 var arr = [0, 1, 2, 3];
+var zerothElementStorage = arr[0];
 Object.defineProperty(arr, "0", {
   get() {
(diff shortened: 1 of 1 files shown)


ptomato (dismissed):
This solution makes sense to me. But I suspect this line might be testing that the "17" value is no longer read anywhere, in which case a better solution might be to add a setter to the `0` property, or use a Proxy instead of an accessor so you can observe the `0` Get operation but not block it. @nicolo-ribaudo @catamorphism @acutmore would you mind confirming before I merge this?

syg:
>  But I suspect this line might be testing that the "17" value is no longer read anywhere, in which case a better solution might be to add a setter to the 0 property, or use a Proxy instead of an accessor so you can observe the 0 Get operation but not block it.

That also sounds reasonable, adding a setter is easier than using a Proxy, let me update.

acutmore (approved):
LGTM thanks!

Ms2ger (approved):

Issue #3649, opened 2022-08-30
What is the purpose of test/staging/?


linusg:
#3591 quietly introduced the `test/staging/` directory without explaining its purpose. Since then more tests have been added there, but I'm failing to find any documentation about it.

- Are test262 runners expected to include or exclude tests in this directory by default?
- Are tests in this directory expected to pass in an engine with a correct implementation of the latest (draft) spec these tests are targeting?
- What are the requirements for someone writing tests to add them to staging, and when can they be moved out of staging?

We're seeing an unusal high failure rate (75% not passing overall) for that directory, so some clarification would be greatly appreciated!

ljharb:
The intention is for it to be for implementation-contributed tests that don’t yet meet the bar for being full test262 tests. That way, implementations are less likely to duplicate work.

My hope is that implementations don’t *run* those tests (except for the ones they’ve contributed), but instead help migrate the tests to the proper place, and run them from there.

cc @syg for a better explanation (which we should probably add to the readme)

syg:
> The intention is for it to be for implementation-contributed tests that don’t yet meet the bar for being full test262 tests. That way, implementations are less likely to duplicate work.

The intention is to get tests that are **correct and runnable** running across implementations ASAP, without having to massage them to conform to test262 directory structures, frontmatter style, and so forth. Eventually they should be massaged and merged into the main test262 suite.

> My hope is that implementations don’t run those tests (except for the ones they’ve contributed), but instead help migrate the tests to the proper place, and run them from there.

No, the intention is _explicitly_ for all implementations to run tests in `tests/staging/` to ensure interop as early as possible.

To @linusg's specific questions:

- Are test262 runners expected to include or exclude tests in this directory by default?

Expected to include.

- Are tests in this directory expected to pass in an engine with a correct implementation of the latest (draft) spec these tests are targeting?

Yes.

- What are the requirements for someone writing tests to add them to staging, and when can they be moved out of staging?

Requirement for adding a test to staging is that it

1. Tests a late-stage proposal or an existing language feature
2. Is correct
3. Is runnable
4. Is reviewed by _someone_ who can land tests in `tests/staging/`, not necessarily a test262 maintainer

Requirement for moving a test out of staging is the same as merging tests into the main test262 suite today, e.g. being in the right directory and having the right frontmatter and so forth.

syg:
[...]

PR #3707, opened 2022-10-29
Document staging folder

ptomato:
This intends to make explicit in the contributor and interpreter docs, the intention of the staging folder and how implementations should handle it in their test262 interpreters.

Also adds a note about the `test/intl402/` folder since there wasn't one previously.

Closes: #3649

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -246,6 +246,29 @@ This key identifies the section number from the portion of the ES6 standard that
 
 Read the [Test262 Technical Rationale Report](https://github.com/tc39/test262/wiki/Test262-Technical-Rationale-Report,-October-2017#specification-reference-ids) for reasoning behind deprecation.
 
+## Staging
+
+There is a `test/staging/` folder, containing tests that are subject to fewer requirements, in order to get tests running across more than one implementation as early as possible, and promote interoperability of in-progress features.
+
+Tests in staging are not required to be split up into one test per file, or to conform to any particular style as long as they are runnable.
+In particular, mechanically-converted tests from implementations' private test suites are encouraged, as sharing them with other implementations promotes interoperability.
+
+These are the requirements for adding a test to staging:
+
+1. Just as with a test outside of staging, it must test an existing language feature, a Stage 3 TC39 proposal, or a normative pull request.
+2. It is correct.
+3. It is runnable using the usual test262 runner.
+4. It is reviewed by someone with permission to land tests in the staging folder, not necessarily a test262 maintainer.
+5. If it requires the implementation to support ECMA-402, it must be in `test/staging/intl402/` or a subfolder thereof.
+
+Tests are intended to live temporarily in staging.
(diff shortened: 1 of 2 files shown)


gibson042 (approved):

linusg (approved):
Perfect, thanks!

CONTRIBUTING.md
```diff
@@ -246,6 +246,36 @@ This key identifies the section number from the portion of the ES6 standard that
...
+2. It is correct.
+3. It is runnable using the usual test262 runner.
+4. It is reviewed by someone with permission to land tests in the
+   staging folder, not necessarily a test262 maintainer.
+5. If it requires the implementation to support ECMA-402, it must be in
+   `test/staging/intl402/` or a subfolder thereof.
+
+Tests are intended to live temporarily in staging.
+Once the feature they test has stabilized, contributors are encouraged
+to refine them to meet the requirements of the main test262 suite.
```
ljharb, line 274: i'd prefer to avoid hard-wrapping, if possible, but not a blocker.

ljharb (dismissed):
LGTM, once @syg approves

CONTRIBUTING.md
```diff
@@ -246,6 +246,36 @@ This key identifies the section number from the portion of the ES6 standard that
...
+
+Tests in staging are not required to be split up into one test per file,
+or to conform to any particular style as long as they are runnable.
+In particular, mechanically-converted tests from implementations'
+private test suites are encouraged, as sharing them with other
+implementations promotes interoperability.
+
+These are the requirements for adding a test to staging:
+
+1. It tests a late-stage TC39 proposal or an existing language feature.
```
Ms2ger, line 264: Consider clarifying "late-stage"

Ms2ger (approved):

CONTRIBUTING.md
ptomato: I don't mind changing this, but it should really be a lint rule. The existing text in the files is a mix of wrapping styles, and you can't see the difference when it's rendered anyway. I think we should pick a style ([here is a comprehensive blog post about pros and cons of different wrapping styles](https://mtsknn.fi/blog/4-1-wrapping-styles-for-markdown-prose-and-code-comments/)) and just automate it.

CONTRIBUTING.md
ptomato: I realized "late-stage" is no different from the rule about what tests are [accepted](https://github.com/tc39/test262/blob/main/CONTRIBUTING.md#acceptable-tests) in the main tree, so I rephrased this.

ptomato:
Updated. I've also added a paragraph about implementations being able to designate someone(s) who can land tests in staging independently.

syg (approved):
This is great, thanks for documenting!

CONTRIBUTING.md
ljharb: Totally fine with doing that (i prefer a sentence per line, since "semantic linebreaks" isn't automateable)

Ms2ger (approved):

PR #3712, opened 2022-11-04
Temporal: Improve tests for observable order of operations

ptomato:
Add more tests for user-observable order of operations, and improve the ones that were already there. I'm planning to present a normative change to the proposal that will reorder some of these operations, so it's important to have coverage in place first, for the current state of the proposal.

I'd recommend reviewing this commit-by-commit. Most of the changes within each individual commit are quite similar to each other, so it should be sufficient to check a representative sample.

One thing to pay special attention to while reviewing, is that PlainYearMonth.subtract subtracts its duration from the end of the month, while PlainYearMonth.add adds its duration to the beginning of the month. This results in an extra call to the calendar's `daysInMonth` method.

### DIFF
--- a/test/built-ins/Temporal/PlainDate/prototype/toString/order-of-operations.js
+++ b/test/built-ins/Temporal/PlainDate/prototype/toString/order-of-operations.js
@@ -1,0 +1,31 @@
+// Copyright (C) 2022 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.plaindate.prototype.tostring
+description: Properties on an object passed to toString() are accessed in the correct order
+includes: [compareArray.js, temporalHelpers.js]
+features: [Temporal]
+---*/
+
+const expected = [
+  "get options.calendarName",
+  "get options.calendarName.toString",
+  "call options.calendarName.toString",
+  "get this.calendar[Symbol.toPrimitive]",
+  "get this.calendar.toString",
+  "call this.calendar.toString",
+];
+const actual = [];
+
+const calendar = TemporalHelpers.calendarObserver(actual, "this.calendar");
+const instance = new Temporal.PlainDate(2000, 5, 2, calendar);
+// clear observable operations that occurred during the constructor call
+actual.splice(0);
+
+const options = TemporalHelpers.propertyBagObserver(actual, {
+  calendarName: "auto",
+}, "options");
+
+instance.toString(options);
+assert.compareArray(actual, expected, "order of operations");
--- a/test/built-ins/Temporal/ZonedDateTime/prototype/with/order-of-operations.js
+++ b/test/built-ins/Temporal/ZonedDateTime/prototype/with/order-of-operations.js
@@ -1,0 +1,137 @@
+// Copyright (C) 2022 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.zoneddatetime.prototype.with
+description: Properties on objects passed to with() are accessed in the correct order
+includes: [compareArray.js, temporalHelpers.js]
+features: [Temporal]
+---*/
+
+const expected = [
+  // RejectObjectWithCalendarOrTimeZone
+  "get fields.calendar",
+  "get fields.timeZone",
+  // CalendarFields / PrepareTemporalFields
+  "get this.calendar.fields",
+  "call this.calendar.fields",
+  "get fields.day",
+  "get fields.day.valueOf",
+  "call fields.day.valueOf",
+  "get fields.hour",
+  "get fields.hour.valueOf",
+  "call fields.hour.valueOf",
+  "get fields.microsecond",
+  "get fields.microsecond.valueOf",
+  "call fields.microsecond.valueOf",
+  "get fields.millisecond",
+  "get fields.millisecond.valueOf",
+  "call fields.millisecond.valueOf",
+  "get fields.minute",
+  "get fields.minute.valueOf",
+  "call fields.minute.valueOf",
+  "get fields.month",
+  "get fields.month.valueOf",
+  "call fields.month.valueOf",
+  "get fields.monthCode",
+  "get fields.monthCode.toString",
+  "call fields.monthCode.toString",
+  "get fields.nanosecond",
+  "get fields.nanosecond.valueOf",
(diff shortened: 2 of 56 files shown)


test/built-ins/Temporal/PlainDate/prototype/toString/order-of-operations.js
```diff
@@ -0,0 +1,31 @@
...
+description: Properties on an object passed to toString() are accessed in the correct order
+includes: [compareArray.js, temporalHelpers.js]
+features: [Temporal]
+---*/
+
+const expected = [
+  "get options.calendarName",
+  "get options.calendarName.toString",
+  "call options.calendarName.toString",
+  "get this.calendar[Symbol.toPrimitive]",
```
cjtenny, line 15: Not to be a pain in the butt, but if we want to test every possible observation, we could ensure calendar's Symbol.toPrimitive method is called as well, and that there's a toString call on the result. In general, for my education on test262, what's the threshold for how in-depth to go here? To me checking all these calls could be pretty excessive, but I don't know what bar to set for test262.

test/built-ins/Temporal/PlainDate/prototype/toString/order-of-operations.js
```diff
@@ -0,0 +1,31 @@
...
+  "get this.calendar.toString",
+  "call this.calendar.toString",
+];
+const actual = [];
+
+const calendar = TemporalHelpers.calendarObserver(actual, "this.calendar");
+const instance = new Temporal.PlainDate(2000, 5, 2, calendar);
+// clear observable operations that occurred during the constructor call
+actual.splice(0, actual.length);
```
cjtenny, line 24: FYI, the second parameter is optional and will have this behavior if omitted - with less room for error. Could be nice to prefer actual.splice(0), or perhaps this is worth moving out of TemporalHelpers into a more broad observer helper?

test/built-ins/Temporal/ZonedDateTime/prototype/with/order-of-operations.js
```diff
@@ -0,0 +1,138 @@
...
+  "get this.timeZone.getOffsetNanosecondsFor",  // ZonedDateTime.p.nanosecond
+  "call this.timeZone.getOffsetNanosecondsFor",
+  "get this.timeZone.getOffsetNanosecondsFor",  // ZonedDateTime.p.second
+  "call this.timeZone.getOffsetNanosecondsFor",
+  "get this.timeZone.getOffsetNanosecondsFor",
+  "call this.timeZone.getOffsetNanosecondsFor",
+  "get this.calendar.year",
+  "call this.calendar.year",
+  "get this.timeZone.getOffsetNanosecondsFor",  // ZonedDateTime.p.offset
+  "call this.timeZone.getOffsetNanosecondsFor",
```
cjtenny, line 88: I didn't understand how this could result from PrepareTemporalFields - nor could I replicate it manually tracing through user callable sites of the proposal text. What's going on here, and why are there so many calls expected to getOffsetNansecondsFor - am I missing something in the proposal text or is this an oops? Running it against the polyfill and engines shows it passes, so I'm guessing I'm just missing it and pardon the perhaps silly question.

cjtenny (commented):
Took a stab at reviewing this in preparation for #3726 , only ran into some snafus on the last commit. Pending that, looks fairly complete and nicely cleaned up, as far as I can tell. Thanks!

test/built-ins/Temporal/PlainDate/prototype/toString/order-of-operations.js
ptomato: There could certainly be different opinions on this, but to me that doesn't add a lot of value. Get @@toPrimitive followed by Get toString and Call toString, on an object with no `[Symbol.toPrimitive]` property, most likely indicates that ToString is used correctly here.

To be entirely exhaustive here you'd have to check many more cases than that (e.g., valueOf but no toString, @@toPrimitive returns an object with @@toPrimitive up to an arbitrary depth), and I don't think it provides any clear benefit to implementors or web compatibility.

test/built-ins/Temporal/PlainDate/prototype/toString/order-of-operations.js
ptomato: Sure, I'll add another commit to clean up existing uses of this.

As for an observer helper, I'm hoping that tests like these will help build a strong case for such a helper in the future via the RFC process.

test/built-ins/Temporal/ZonedDateTime/prototype/with/order-of-operations.js
ptomato: It's tricky. This occurs in step 14 of [Temporal.ZonedDateTime.prototype.with](https://tc39.es/proposal-temporal/#sec-temporal.zoneddatetime.prototype.with), "PrepareTemporalFields(_zonedDateTime_, _fieldNames_, ...)". That performs a Get on the ZonedDateTime object for each of the properties named in _fieldNames_. The property getters (e.g., [year](https://tc39.es/proposal-temporal/#sec-get-temporal.zoneddatetime.prototype.year)) each call [BuiltinTimeZoneGetPlainDateTimeFor](https://tc39.es/proposal-temporal/#sec-temporal-builtintimezonegetplaindatetimefor), which produces the getOffsetNanosecondsFor call.

Converting the exact time to a PlainDateTime so many times is a bit over the top, but it is what is specified. I think it can be addressed in https://github.com/tc39/proposal-temporal/issues/2289.

test/built-ins/Temporal/PlainDate/prototype/toString/order-of-operations.js
cjtenny: Thank you, that's a good explanation. As someone new to test262, in my tests for Array.fromAsync, I find myself wondering how deep down the rabbit hole of checking iterator protocol calls to go - I'll try to use this as a guideline to remove some of the extra-deep checks :)

test/built-ins/Temporal/ZonedDateTime/prototype/with/order-of-operations.js
cjtenny: Thank you very much! I forgot that "Get" wouldn't just call user code, but would also call the built-in getters. That was the missing link in my trace.

cjtenny:
Looks great to me, thanks for the detailed answers!

Ms2ger (approved):

PR #3713, opened 2022-11-04
Update Intl tests to recognize microsecond and nanosecond as sanctioned

trflynn89:
These were added to the list of sanctioned units in a normative change:
https://github.com/tc39/ecma402/commit/f627573


ptomato (dismissed):
Thanks. Both of these changes look correct to me, based on the recent changes. However, it looks like only the microsecond/nanosecond change has been presented. The DurationFormat change looks like it was merged prematurely without TC39 consensus, so I expect it'll be presented at the upcoming November/December meeting.

We can just wait to merge this for a few weeks until that happens, or if you prefer to split the changes up into two PRs, I can merge the microsecond/nanosecond one immediately.

trflynn89:
Sure! I changed this PR to only include the microsecond/nanosecond change, and moved the DurationFormat change to #3714.

ptomato (approved):
Thanks. Then I will go ahead and merge this.

PR #3715, opened 2022-11-08
Add missed changes for Symbols as WeakMap keys proposal.

Ms2ger:
These are the remaining substantial changes from #2890.

Fixes #2890.


ptomato (approved):
Thanks. I did realize that #2890 was a later version of Rick's branch than the one that I had gotten to from the commits linked in #2850, but I didn't go back and check the completeness against my branch. Good catch!

Issue #3703, opened 2022-10-26
GraalJS fails to install in CI


Ms2ger:
https://app.circleci.com/pipelines/github/tc39/test262/4468/workflows/69b41b54-d4ff-4b23-beb7-f3bd63554ffe/jobs/41781
cc @devsnek 
```
esvu ❯ version 1.2.10
GraalJS ❯ Checking version...
GraalJS ❯ Installing version 22.3.0
GraalJS ❯ Downloading https://github.com/oracle/graaljs/releases/download/vm-22.3.0/graaljs-22.3.0-linux-amd64.tar.gz
progress [----------------------------------------] 0% | ETA: 0s | 0/59452349progress [==========------------------------------] 24% | ETA: 0s | 14352384/59452349progress [===========-----------------------------] 28% | ETA: 0s | 16777216/59452349progress [====================--------------------] 50% | ETA: 0s | 30310400/59452349progress [=============================-----------] 71% | ETA: 0s | 42401792/59452349progress [========================================] 100% | ETA: 0s | 59452349/59452349GraalJS ❯ Extracting /tmp/esvu-9ea49255fad1beb5fb67a9a903903de2
GraalJS ❯ Installing /tmp/esvu-9ea49255fad1beb5fb67a9a903903de2-extracted
GraalJS ❯ Testing...
esvu ✖ Error: Command failed with exit code 255: /home/circleci/.esvu/bin/graaljs -e print("42");
Could not load JVM.
    at makeError (/home/circleci/test262/node_modules/execa/lib/error.js:59:11)
    at handlePromise (/home/circleci/test262/node_modules/execa/index.js:114:26)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async GraalJSInstaller.test (/home/circleci/test262/node_modules/esvu/src/engines/graaljs.js:79:8)
    at async Function.install (/home/circleci/test262/node_modules/esvu/src/installer.js:110:5)
    at async installEngine (/home/circleci/test262/node_modules/esvu/src/bin.js:121:3)
    at async main (/home/circleci/test262/node_modules/esvu/src/bin.js:168:7) {
  shortMessage: 'Command failed with exit code 255: /home/circleci/.esvu/bin/graaljs -e print("42");',
  command: '/home/circleci/.esvu/bin/graaljs -e print("42");',
  exitCode: 255,
  signal: undefined,
  signalDescription: undefined,
  stdout: '',
  stderr: 'Could not load JVM.',
  failed: true,
[...]

PR #3716, opened 2022-11-08
CI: Disable GraalJS.

Ms2ger:
Ref. #3703.


ptomato (approved):
I opened https://github.com/devsnek/esvu/issues/38 in case they missed the ping on #3703, let's wait a few days for a response and then go ahead with this.

ptomato (approved):

PR #3718, opened 2022-11-10
Temporal: Test throwing on time-zone-to-calendar conversion and vice versa

ptomato:
This contains tests for the normative PR https://github.com/tc39/proposal-temporal/pull/2433, which is to be presented for consensus to TC39 in the upcoming plenary meeting. That PR changes ToTemporalCalendar to throw when it encounters a Temporal.TimeZone instance, and ToTemporalTimeZone to throw when it encounters a Temporal.Calendar instance.


ptomato:
The corresponding normative change achieved consensus at the TC39 plenary meeting today.

cjtenny (approved):
Matches the normative changes and thorough, I think that's everywhere!

Ms2ger (approved):

Issue #3380, opened 2022-01-21
built-ins/TypedArray/prototype/map/callbackfn-resize.js seems wrong


marjakh:
In this part:

```
  result = sample.map(function(element, index, array) {
    if (elements.length === 0) {
      try {
        buffer.resize(2 * BPE);
        finalElement = undefined;
        finalResult = NaNvalue;
        ...
      } catch (_) {
        ...
      }
    }

    elements.push(element);
    indices.push(index);
    arrays.push(array);

    return index; <<<< here
  });
  ```

The callback function which we pass as a parameter to map() always returns `index`, however, we expect the result to be:

```
assert.compareArray(result, [0, 1, finalResult], 'result (shrink)');
```

This doesn't seem right. Shouldn't we do:
```
assert.compareArray(result, [0, 1, 2], 'result (shrink)');
```
instead?

cc @syg

syg:
My reading of that test agrees with @marjakh's reading.

Since `result` is the result of calling `TypedArray.prototype.map`, it's a new TA backed by a different buffer. The comparison to `[0, 1, finalResult]` makes me think the test writer maybe was trying to test `sample` after it shrunk? But that doesn't really make sense either since I don't know where `1` and `2` would come from...

marjakh:
cc @sarahghp

rwaldron:
@marjakh lets also add @Ms2ger @romulocintra @ryzokuken

PR #3721, opened 2022-11-15
test/built-ins/TypedArray/prototype/map/callbackfn-resize.js test result is incorrect

Constellation:
%TypedArray%.prototype.map's iteration count does not change even though the source view is resized during the iteration. And since result is just returning index, then result array should be [0, 1, 2].


Ms2ger:
@jugglinmike I think this is a test you wrote - could you take a look?

syg:
This is the same issue of https://github.com/tc39/test262/issues/3380. @Constellation is right.

jugglinmike:
Will do!

jugglinmike (approved):
Thanks!

PR #3722, opened 2022-11-15
Update DateTimeFormat proleptic Gregorian testing to include era

gibson042:
Ref https://github.com/tc39/ecma402/issues/723


ptomato (approved):
ECMA-402 currently specifies the strings "BC" and "AD", but it makes sense to me to make the test robust for a change to BCE/CE nomenclature in the future.

If I understand correctly, we don't actually need to wait for https://github.com/tc39/ecma402/issues/723 to achieve consensus in order to merge this? The test as written already reflects web reality, and disagrees with the spec text. This PR just improves the test, right?

sffc (approved):
Thank you!

gibson042:
https://github.com/tc39/ecma402/pull/729 achieved consensus at the TC39 plenary of 2023-01-31.

Issue #3480, opened 2022-04-12
CI suggestion: Drop Python 2 build & lint tests


jessealama:
In our CircleCI setup, we run some build & lint tools both with Python 2 and with Python 3. The scripts that get run are exactly the same: the Python 3 build runs the same code that the Python 2 does, just with a different Python binary. Of course, the dependencies (what gets installed by pip) are surely different, but the code we wrote, in this repository, is run as-is by both Python 2 and Python 3. I wonder whether there's any additional information to be gained, at this stage in the whole Python 2-to-3 evolution, when comparing the two runs. I wager not; I would expect that either both succeed or both fail. If that's right, I would suggest that we can just drop the Python 2 CI step.

ljharb:
Are there any implementations that rely on Python 2 to run tests?

ptomato:
I asked in the TC39 Implementers chat and the only response was "kill it" devsnek. I think we can probably drop the Python 2 step. I can open a PR and see if anyone complains.

PR #3727, opened 2022-11-23
Remove Python 2 CI job

ptomato:
The Python 2 and 3 jobs run essentially the same thing twice. I asked for a sign from implementations if any of them were relying on Python 2 for their test262 runners, but didn't hear of any that were.

The Python tools are changed so rarely that I think it's OK to drop this CI job. If a future change breaks Python 2 for someone, we can accept PRs to fix it on a best-effort basis, or reinstate this CI job if it's really necessary.

Closes: #3480

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -146,7 +146,7 @@ workflows:
   version: 2
   Tools:
     jobs:
-      - "Test262: verify tools; build & lint tests (Python 3)"
+      - "Test262: verify tools; build & lint tests"
   Tests execution:
     jobs:
       - "ChakraCore: New or modified tests execution"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,33 +37,6 @@ execution_steps: &execution_steps
     - run: hostPath=$HOME/.esvu/bin/$hostPath npm run ci
 
 jobs:
-  "Test262: verify tools; build & lint tests (Python 2)":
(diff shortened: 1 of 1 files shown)


ptomato:
(Note, we'd have to remove the Python 2 job from the branch protection settings of `main` before merging this)

.circleci/config.yml
```diff
@@ -37,9 +37,9 @@ execution_steps: &execution_steps
     - run: hostPath=$HOME/.esvu/bin/$hostPath npm run ci
 
 jobs:
-  "Test262: verify tools; build & lint tests (Python 2)":
+  "Test262: verify tools; build & lint tests (Python 3)":
```
gibson042, line 40: We might as well drop the now-unnecessary refinement.
```suggestion
  "Test262: verify tools; build & lint tests":
```

gibson042 (approved):

.circleci/config.yml
```diff
@@ -170,7 +146,6 @@ workflows:
   version: 2
   Tools:
     jobs:
-      - "Test262: verify tools; build & lint tests (Python 2)"
       - "Test262: verify tools; build & lint tests (Python 3)"
```
Ms2ger, line 149:
```suggestion
      - "Test262: verify tools; build & lint tests"
```

Ms2ger (approved):

PR #3728, opened 2022-11-23
Implementation of RFC: Async Helpers (#3724)

cjtenny:
This implements the `harness/asyncHelpers.js` file proposed in @ptomato 's [RFC](https://github.com/tc39/test262/pull/3724), with its functions `asyncTest(testFunc)` and `assert.throwsAsync(expectedErrorConstructor, funcOrThenable, message)`. Also included are tests for both, including tests mirroring `assert.throws` and a number of additional throwsAsync-specific tests.

Per @jugglinmike 's comments in the maintainers call, never uses await; always explicitly invokes as a thenable.

Implements the original intention of accepting a function or thenable for throwsAsync, happy to modify if there's some general agreement on changing that.


harness/asyncHelpers.js
```diff
@@ -0,0 +1,122 @@
...
+  }
+  try {
+    return testFunc().then(
+      function () {
+        $DONE();
+      },
+      function (error) {
+        $DONE(error);
+      }
+    );
```
ljharb, line 25: what happens if this is trying to test Promise.prototype.then itself?

harness/asyncHelpers.js
```diff
@@ -0,0 +1,122 @@
...
+  } catch (syncError) {
+    $DONE(syncError);
+    return Promise.resolve();
```
ljharb, line 28: why not make the containing function `async` rather than needing to always return a Promise?

harness/asyncHelpers.js
```diff
@@ -0,0 +1,122 @@
...
+      if (
+        innerThenable === null ||
+        typeof innerThenable !== "object" ||
+        typeof innerThenable.then !== "function"
+      ) {
+        message +=
+          "Expected to obtain an inner promise that would reject with a" +
+          expectedErrorConstructor.name +
+          " but result was not a thenable";
+        return Promise.reject(new Test262Error(message));
```
ljharb, line 55: same here; why not throw this inside an async function?

harness/asyncHelpers.js
cjtenny: The goal is to ensure that if it's called without the async flag, it will still execute to throw the error on line 11 - I believe top-level calls to async functions might not execute in most runners before the implementation returns, leading to silent failures if an author omits the 'async' flag. But I could well be wrong!

harness/asyncHelpers.js
cjtenny: You're right, this one could and should be inside an async function - I see no reason it couldn't, anyway. I'll refactor, thanks! I think my brain was stuck on the asyncTest logic and I forgot to re-evaluate here.

harness/asyncHelpers.js
cjtenny: Good question. Looking at those tests, many of them wouldn't benefit from this function. But also, Promise.prototype.then seems core enough to async functionality that it seems there's a bit of a circular dependency to using both; perhaps that's one of the places it makes sense to continue to use custom logic around manually invoking $DONE? Very open to other ideas, if you have suggestions for how to improve this case.

Either way, I think it remains useful for simplifying tests of other async features not implementing core async interfaces.

harness/asyncHelpers.js
ljharb: You're correct, but i'd expect that to be handled by the unhandled rejection handler, rather than a confusing sync exception from a promise-returning function (ie, zalgo)

cjtenny: The RFC itself [makes the case](https://github.com/tc39/test262/pull/3724/files#diff-d4e487dcd72bb7cd7462c9709f50abd806977f296306b5f4eb77e3fc011ede51R112) for summoning Zalgo here, and test262 runners either don't handle unhandled rejections or handle them without clear messages. Engine262 and JSC both pass a test with an unhandled rejection; v8 will fail it with no output. Node gives an UnhandledPromiseRejection, as does SpiderMonkey. Because of these reasons, I think it's actually worth the top-level Zalgo - hopefully we only would expect this to be called top level. I'd argue it's at least moderately easier to review and catch than the possible mistakes that arise from not having it.

harness/asyncHelpers.js
cjtenny: Thinking further about it: This only depends on the test function returning a thenable. There are a few options:
1. Do not use asyncTest when testing Promise.prototype.then. Instead, write your own logic for calling $DONE.
2. In your test for Promise.prototype.then, return a custom thenable rather than a Promise. This seems reasonable for a test of Promises anyway.
3. If testing a polyfilled Promise.prototype.then, one would expect the regular tests for Promise.prototype.then to uncover bugs, or the consistency tests for this harnenss to catch them. This itself is not a test for Promise.prototype.then, and doesn't reference or depend on it in any way, as it only references a thenable.

I'll push a quick update removing the references to Promise.resolve and instead returning a simple thenable, so that now the implementation doesn't reference Promise at all.

Does that cover all the cases you were thinking of, @ljharb ?

test/harness/asyncHelpers-asyncTest-func-throws-sync.js
```diff
@@ -0,0 +1,16 @@
...
+    The 'asyncTest' helper reports synchronous errors via $DONE.
+includes: [asyncHelpers.js]
+---*/
+var called = false;
+function $DONE(error) {
+  called = true;
+  assert(error instanceof Test262Error);
+}
+asyncTest(function () {
+  throw new Test262Error("Should not be rethrown");
```
ptomato, line 14: Maybe we should make this a string or something like that, so it's easy to assert that it's the exact same thrown value that is passed to `$DONE`?

test/harness/asyncHelpers-asyncTest-return-not-thenable.js
```diff
@@ -0,0 +1,29 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+description: |
+    The 'asyncTest' helper rejects test functions that do not return a thenable.
+includes: [asyncHelpers.js, compareArray.js]
+---*/
+
+const doneValues = [];
+function $DONE(error) {
+  doneValues.push(error instanceof TypeError);
```
ptomato, line 11: The RFC doesn't say what kind of error this is, but I'd guess it should be a `Test262Error` in parallel with `assert.throwsAsync()`.

test/harness/asyncHelpers-throwsAsync-funcOrThenable-throws-sync.js
```diff
@@ -0,0 +1,63 @@
...
+  await checkRejects({
+    then: function () {
+      throw new Error();
+    },
+  });
+  await checkRejects({
+    then: function () {
+      throw new Test262Error();
+    },
+  });
```
ptomato, line 48: This needs to be updated since we changed `throwsAsync` to just take a function and not a thenable. Also suggest renaming this test.

test/harness/asyncHelpers-throwsAsync-invalid-funcOrThenable.js
```diff
@@ -0,0 +1,71 @@
...
+  await checkRejects(null);
+  await checkRejects({});
+  await checkRejects("string");
+  await checkRejects(10);
+  await checkRejects();
+  await checkRejects({ then: null });
+  await checkRejects({ then: {} });
+  await checkRejects({ then: "string" });
+  await checkRejects({ then: 10 });
+  await checkRejects({ then: undefined });
```
ptomato, line 42: These are still correct, but can probably be removed as they're no longer relevant if we don't accept a thenable. Also suggest renaming this test.

test/harness/asyncHelpers-throwsAsync-same-realm.js
```diff
@@ -0,0 +1,37 @@
+// Copyright (C) 2022 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+description: |
+    Thenables that reject with instances of the realm specified constructor function
+    satisfy the assertion, without cross realms collisions.
```
ptomato, line 6: I was a bit confused by this description, it sounds like the opposite of what the test actually does.

test/language/statements/class/decorator/syntax/valid/decorator-call-expr-identifier-reference-yield.js
```diff
@@ -1,15 +1,19 @@
 // This file was procedurally generated from the following sources:
 // - src/decorator/decorator-call-expr-identifier-reference-yield.case
-// - src/decorator/syntax/valid/cls-decl-decorators-valid-syntax.template
+// - src/decorator/syntax/valid/cls-element-decorators-valid-syntax.template
```
ptomato, line 3: This doesn't seem right. I remember somewhat recently there was a case where two test generators raced to produce the same file, which I believe was fixed. Is it possible that this branch is old enough that it doesn't have that fix, or maybe there's another similar case that wasn't caught yet?

ptomato (changes requested):
Now that the RFC is merged I went over this in a bit more detail, and found a few things that should be fixed or clarified. Also it's slightly odd that the CI didn't run, so it would be good to find out why that is.

test/language/statements/class/decorator/syntax/valid/decorator-call-expr-identifier-reference-yield.js
cjtenny: Yep, that was the case! A rebase, clean, and re-build fixed it. Thanks for catching it.

test/harness/asyncHelpers-throwsAsync-invalid-funcOrThenable.js
cjtenny: I realized this test also no longer correctly tests things because it doesn't properly check calls to $DONE - oops. I'll keep one example of a thenable, just to assert that we no longer take thenables.

test/harness/asyncHelpers-throwsAsync-same-realm.js
cjtenny: I copied it directly from assert-throws-same-realm.js. I *think* the description just needs to be updated in both, so I'll do that. But I'm not very familiar with realms yet.

cjtenny:
Rebased and addressed feedback; feedback addressed in separate commit, nothing else changed in older commits. I'm sorry I dropped the ball on this and hadn't responded sooner.

CONTRIBUTING.md
```diff
@@ -344,52 +344,43 @@ Consumers that violate the spec by throwing exceptions for parsing errors at run
...
+/*---
+... (other frontmatter) ...
+flags: [async]
+includes: [asyncHelpers.js]
+---*/
 
-}).then($DONE, $DONE);
+asyncTest(async function() {
+  assert.sameValue(true, await someTestCode(1), "someTestCode(1) should return true");
+;
```
ptomato, line 360:
```suggestion
```

CONTRIBUTING.md
```diff
@@ -344,52 +344,43 @@ Consumers that violate the spec by throwing exceptions for parsing errors at run
...
 
-Function `checkAssertions` implicitly returns `undefined` if the expected condition is observed.  The return value of function `checkAssertions` is then used to asynchronously invoke the first function of the final `then` call, resulting in a call to `$DONE(undefined)`, which signals a passing test.
+For more complicated asynchronous testing, such as testing Promise or other core asynchronous functionality, the runner expects that the global `$DONE()` function will be called **exactly once** to signal test completion.
 
-If the expected condition is not observed, function `checkAssertions` throws a `Test262Error`.  This is caught by the Promise and then used to asynchronously invoke the second function in the call -- which is also `$DONE` -- resulting in a call to `$DONE(error_object)`, which signals a failing test.
+ * If the argument to `$DONE` is omitted, is `undefined`, or is any other falsy value, the test is considered to have passed.
+
+ * If the argument to `$DONE` is a truthy value, the test is considered to have failed and the argument is displayed as the failure reason.
 
 ### Checking Exception Type and Message in Asynchronous Tests
```
ptomato, line 370:
```suggestion
### Checking Exception Type in Asynchronous Tests
```
We no longer have an example of how to check the exception message, but also I think we don't need one because I don't think any tests I've seen do so. Nor should they, since the contents of exception messages are not part of the requirements for a JS implementation.

test/harness/asyncHelpers-asyncTest-return-not-thenable.js
ptomato: I'm not sure I get where this TypeError is thrown, it seems to me that it should be a Test262Error according to the following code:
```js
innerThenable = func();
if (
  innerThenable === null ||
  typeof innerThenable !== "object" ||
  typeof innerThenable.then !== "function"
) {
  message += /* etc */;
}
```
it seems to me that e.g. `asyncTest(() => 42)` assigns 42 to `innerThenable`, so the condition should trigger here. But the test passes, so I'm not sure what's going on.

ptomato (approved):
Just some minor loose ends!

Before merging would you mind debugging why the CI is not running? It looks like you may have to authorize the CircleCI app on your fork or something.

test/harness/asyncHelpers-asyncTest-return-not-thenable.js
cjtenny: > The RFC doesn't say what kind of error this is, but I'd guess it should be a `Test262Error` in parallel with `assert.throwsAsync()`.

Oops, I forgot to respond to this rather than just mark the issue resolved, sorry! That is very much my mistake. Thank you for bringing it up again. It comes from [this line](https://github.com/tc39/test262/blob/main/rfcs/async-helpers.md?plain=1#L103) - this occurs when calling testFunc synchronously throws. In this case, because testFunc is not a function, trying to call it throws a TypeError. The test passes as a result, and if we want to change it to throw a Test262Error I'd support that - but we should change the RFC text to match.

ptomato: I see where I was confused now; I was looking at the code for `assert.throwsAsync()`, but this is `asyncTest()`.

PR #3729, opened 2022-11-24
add a test for Generator

SkyCrystal:
Hello! This is my first pull request here.
I found a bug in quickjs, and it should fail this test.
Hopes that everything goes right.


test/built-ins/GeneratorPrototype/return/try-finally-set-property-within-try.js
```diff
@@ -0,0 +1,28 @@
+// Copyright (C) 2022 Bo Pang. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 25.3.1.3
+description: >
+    When a generator is paused within a `try` block of a `try..finally`
+    statement, `return` should interrupt control flow as if a `return`
+    statement had appeared at that location in the function body.
+    And the `finally` block runs first.
```
gibson042, line 9:
```suggestion
description: >
    When a generator is paused within a `try` block of a `try..finally`
    statement, `return` should interrupt control flow as if a `return`
    statement had appeared at that location in the function body.
    The `finally` block is still evaluated, and may override the return value.
```

test/built-ins/GeneratorPrototype/return/try-finally-set-property-within-try.js
```diff
@@ -0,0 +1,28 @@
+// Copyright (C) 2022 Bo Pang. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 25.3.1.3
```
gibson042, line 4: `es6id` is deprecated in favor of `esid`, see https://github.com/tc39/test262/blob/main/CONTRIBUTING.md#frontmatter . Also note that 25.3.1.3 doesn't even exist in the current edition of ECMAScript, but if it did exist then it would be a subsection of [Abstract Operations For DataView Objects](https://tc39.es/ecma262/multipage/structured-data.html#sec-abstract-operations-for-dataview-objects), which doesn't correspond with the content of this test.
```suggestion
esid: sec-generator.prototype.return
```

test/built-ins/GeneratorPrototype/return/try-finally-set-property-within-try.js
```diff
@@ -0,0 +1,28 @@
...
+    obj.foo=yield;
+  } finally {
+    return 1;
+  }
+}
+var iter = g();
+var result;
+
+iter.next();
+let ret=iter.return(45);
```
gibson042, line 26: Note the linting failures: https://app.circleci.com/pipelines/github/tc39/test262/4541/workflows/a2e87bc5-695b-4e54-b68e-2ebd6ca03653/jobs/42470#shared-improved-time-tooltips
```suggestion
result = iter.return(45);
```

test/built-ins/GeneratorPrototype/return/try-finally-set-property-within-try.js
```diff
@@ -0,0 +1,28 @@
...
+assert.sameValue(obj.foo,undefined,'`obj.foo` did not set');
+assert.sameValue(ret,1,'return in finally executes first');
```
gibson042, line 28: This assertion is [failing](https://app.circleci.com/pipelines/github/tc39/test262/4541/workflows/73bdd566-06d5-4713-8a55-5837484b80a8/jobs/42468), because the shape of the result of `iter.return(45)` is not a primitive but rather an [_IteratorResult_ object](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-iteratorresult-interface) with `value` and `done` properties per [_Iterator_ Interface Optional Properties](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#table-iterator-interface-optional-properties).

gibson042 (changes requested):
Thanks for the contribution!

Is there special significance to "set property" as opposed to "evaluate code"? If not, I would expect this file to more closely track [try-finally-within-try.js](https://github.com/tc39/test262/blob/83a46bfe0e79aed8274a1b9f4beb0a2efa0b3533/test/built-ins/GeneratorPrototype/return/try-finally-within-try.js), employing the same structure and making assertions only against `inTry`, `inFinally`, `unreachable` and `result` variables (e.g., replace `obj.foo=yield` with `unreachable += yield`).

test/built-ins/GeneratorPrototype/return/try-finally-set-property-within-try.js
SkyCrystal: Oh sorry, I forgot to write that.Thanks a lot!

SkyCrystal:
Yes there is some different between set property and evaluate code.
When generateing bytecode, compiler may parse `obj.foo =` first and push `obj` into stack.
So after we call `return` the `obj` may remains in stack.
quickjs throws a InternalError while compiling this code.

gibson042:
> Yes there is some different between set property and evaluate code. When generateing bytecode, compiler may parse `obj.foo =` first and push `obj` into stack. So after we call `return` the `obj` may remains in stack. quickjs throws a InternalError while compiling this code.

Is that speculation, or a statement of fact? And what would be the relevant difference between `obj.foo = yield` and `unreachable += yield`?

SkyCrystal:
> Is that speculation, or a statement of fact? And what would be the relevant difference between obj.foo = yield and unreachable += yield?

It's a statement of fact in [quickjs](https://github.com/bellard/quickjs), and is very easy to reproduce. Just try to compile this function.

>When generateing bytecode, compiler may parse obj.foo = first and push obj into stack.
So after we call return the obj may remains in stack.

Quickjs works as I said, and generates a wrong bytecode, which fails its check in `compute_stack_size`.

To get this error, you need a `obj.foo =` to push obj into stack first, a `yield` to generate two different branch( The `return` and the `next`, and only `next` really evaulates `obj.foo = yield` to use the `foo` in stack.) And at last, a `try-finally` block is needed, to let these different branch meet with each other.

This is a bytecode generated by quickjs:


```
error_case.js:2: function: *foo
  stack_size: 4
  opcodes:
;; function* foo() {
;;     try {
;;         test.f= yield 123;
;;     }
;;     finally {
;;             return 1;
;;     }
;; }

        initial_yield
        catch 38
        get_var test                 //push test in stack first
        push_i8 123
        yield
        if_false8 23
        nip                          //even if the set property won't be evaluated.
        gosub 44
        return_async
   23:  put_field f                  //sp[-2].f=sp[-1],sp-=2
        drop
        undefined
        gosub 44
        drop
        goto8 46
   38:  gosub 44
        throw
   44:  push_1 1                     //two branches meet at here.
        return_async
   46:  undefined
        return_async
```
InternalError: unconsistent stack size: 2 3 (pc=44)
```
```

test/built-ins/GeneratorPrototype/return/try-finally-set-property-within-try.js
```diff
@@ -0,0 +1,28 @@
...
+description: >
+    When a generator is paused within a `try` block of a `try..finally`
+    statement, `return` should interrupt control flow as if a `return`
+    statement had appeared at that location in the function body.
+    The `finally` block is still evaluated, and may override the return value.
+features: [generators]
+---*/
+
+
+var obj={foo:undefined};
```
leobalter, line 14: As a suggestion to avoid false positives, replace the undefined value with something more specific like:

```suggestion
var obj={foo:'not modified'};
```

test/built-ins/GeneratorPrototype/return/try-finally-set-property-within-try.js
```diff
@@ -0,0 +1,28 @@
...
+    return 1;
+  }
+}
+var iter = g();
+var result;
+
+iter.next();
+result = iter.return(45).value;
+assert.sameValue(obj.foo,undefined,'`obj.foo` did not set');
+assert.sameValue(result,1,'return in finally executes first');
```
leobalter, line 28:
```suggestion
assert.sameValue(obj.foo,'not modified','`obj.foo` did not set');
assert.sameValue(result,1,'finally block supersedes return value');
```

I updated the message above to reflect the actual behavior, PTAL.

test/built-ins/GeneratorPrototype/return/try-finally-set-property-within-try.js
```diff
@@ -0,0 +1,28 @@
...
+    statement had appeared at that location in the function body.
+    The `finally` block is still evaluated, and may override the return value.
+features: [generators]
+---*/
+
+
+var obj={foo:undefined};
+function* g() {
+  try {
+    obj.foo=yield;
```
gibson042, line 17: Style nit:
```suggestion
    obj.foo = yield;
```

test/built-ins/GeneratorPrototype/return/try-finally-set-property-within-try.js
gibson042: I agree with @leobalter , and would also add spaces to match the prevailing style.
```suggestion
assert.sameValue(obj.foo, 'not modified', '`obj.foo` was not set');
assert.sameValue(result, 1, 'finally block supersedes return value');
```

ptomato (approved):
Thanks for picking this up again.

LGTM. @gibson042 does this version address your concerns?

test/built-ins/GeneratorPrototype/return/try-finally-set-property-within-try.js
```diff
@@ -0,0 +1,28 @@
...
+    return 1;
+  }
+}
+var iter = g();
+var result;
+
+iter.next();
+result = iter.return(45).value;
+assert.sameValue(obj.foo, 'not modified', '`obj.foo` was not set');
+assert.sameValue(result, 1, 'finally block supersedes return value');
```
gibson042, line 28:
```suggestion
var obj = { foo: 'not modified' };
function* g() {
  try {
    obj.foo = yield;
  } finally {
    return 1;
  }
}
var iter = g();
var result;

iter.next();
result = iter.return(45).value;
assert.sameValue(obj.foo, 'not modified', '`obj.foo` must not be set');
assert.sameValue(result, 1, 'finally block must supersede return value');
```

gibson042 (approved):

PR #3730, opened 2022-11-25
CI: Integrate ESMeta

doehyunbaek:
Hi, this PR contains changes that integrate ESMeta's feature to test test262 changes. This PR heavily borrows from changes made at https://github.com/tc39/ecma262/pull/2926, which integrates ESMeta's type checker into ecma262's CI.

ESMeta is able to derive an interpreter from ECMAScript spec and execute JavaScript with it. With this change, we can confirm new or modified test files behave according to ECMAScript spec.

I could not test the changes by actually running CircleCI pipeline as that was not simply possible, but instead tested with local CircleCI cli. 

As I am not very experienced with both test262 and CircleCI, if there are any deficiencies please let me know. Thanks in advance.

cc @jhnaldo

### DIFF
--- a/.github/workflows/esmeta-test262.yml
+++ b/.github/workflows/esmeta-test262.yml
@@ -24,9 +24,8 @@ jobs:
           git clone --branch v0.1.0-rc8 --depth 1 https://github.com/es-meta/esmeta.git "${ESMETA_HOME}"
           cd "${ESMETA_HOME}" && git submodule update --init --depth 1
       - name: build esmeta
-        run: |
-          cd "${ESMETA_HOME}"
-          sbt assembly
+        working-directory: ${{ env.ESMETA_HOME }}
+        run: sbt assembly
       - name: run test262
         run: |
           git fetch origin main
--- a/.github/workflows/esmeta-test262.yml
+++ b/.github/workflows/esmeta-test262.yml
@@ -38,4 +38,3 @@ jobs:
           echo New or modified test files:
           echo "$paths"
           "${ESMETA_HOME}"/bin/esmeta test262-test -status -test262dir=$(pwd) $paths
-          
--- a/.github/workflows/esmeta-test262.yml
+++ b/.github/workflows/esmeta-test262.yml
@@ -1,0 +1,41 @@
+name: 'esmeta test262'
+
+on: [pull_request]
+
+jobs:
+  esmeta-test262:
+    name: 'esmeta test262'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout 
+        uses: actions/checkout@v3
+      - name: Setup JDK
+        uses: actions/setup-java@v3
+        with:
+          distribution: temurin
+          java-version: 17
+      - name: Add to PATH
+        shell: bash
+        run: |
+          echo "ESMETA_HOME=$(pwd)/vendor/esmeta" >> $GITHUB_ENV
+      - name: download esmeta
+        run: |
+          mkdir -p "${ESMETA_HOME}"
+          git clone --branch v0.1.0-rc8 --depth 1 https://github.com/es-meta/esmeta.git "${ESMETA_HOME}"
+          cd "${ESMETA_HOME}" && git submodule update --init --depth 1
+      - name: build esmeta
+        run: |
+          cd "${ESMETA_HOME}"
+          sbt assembly
+      - name: run test262
+        run: |
+          git fetch origin main
+          paths=$(git diff --diff-filter ACMR --name-only origin/main.. -- test/)
+          if [ "$paths" == "" ]; then
+            echo No test files added or modified. Exiting.
+            exit 0
+          fi
+          echo New or modified test files:
+          echo "$paths"
+          "${ESMETA_HOME}"/bin/esmeta test262-test -status -test262dir=$(pwd) $paths
+          


.circleci/config.yml
```diff
@@ -166,6 +166,30 @@ jobs:
...
+    working_directory: ~/test262
+    environment:
+      hostType: esmeta
+      ESMETA_HOME: /app/esmeta
+    steps:
+      - checkout
+      - run: 
+          name: "Make directory"
+          command: mkdir -p "${ESMETA_HOME}" 
+      - run:
```
ljharb, line 181: if possible we should run this on github actions, and NOT circle CI.

doehyunbaek: That's totally possible. I'll work on an alternative PR running this on github actions.

ljharb: Great! Any reason not to update this one?

doehyunbaek: Nope! I meant to work on a alternative version of this pr. I misworded.

I will force push the required commit.

.github/workflows/esmeta-test262.yml
```diff
@@ -0,0 +1,41 @@
...
+          git clone --branch v0.1.0-rc8 --depth 1 https://github.com/es-meta/esmeta.git "${ESMETA_HOME}"
+          cd "${ESMETA_HOME}" && git submodule update --init --depth 1
+      - name: build esmeta
+        run: |
+          cd "${ESMETA_HOME}"
+          sbt assembly
+      - name: run test262
+        run: |
+          git fetch origin main
+          paths=$(git diff --diff-filter ACMR --name-only origin/main.. -- test/)
```
ptomato, line 33: If possible, it might be better to use
```suggestion
          paths=$(npm run diff)
```
in order to make sure that this script runs the same set of tests that the other jobs do. But if there's a reason not to do that (maybe we'd have to run npm install first?) then it's fine as is.

ptomato (approved):
Great!! I think we should merge this.

I think in the future it would save time to use a Docker image with ESMeta already built at the correct revision, but that certainly shouldn't block this PR.

doehyunbaek:
> I think in the future it would save time to use a Docker image with ESMeta already built at the correct revision, but that certainly shouldn't block this PR.

Agreed! That would cut off 2 minutes from this workflow.

I discussed with @jhnaldo about preparing a docker image and as we were not building one at our existing CI pipeline, we thought it was appropriate to open the PR without the docker image. We are certainly considering the option and I think we might build a docker image alongside our CI in the future.

.github/workflows/esmeta-test262.yml
doehyunbaek: I wll check if it's possible later today. Thanks for great suggestion.

doehyunbaek: Hi, I tested out the suggested change [here](https://github.com/doehyunbaek/test262/actions/runs/3551724196/jobs/5966141508) and It did not work very well.

Actually, ci_test.sh which other jobs use, do not use `npm run diff` as well, as can be seen below.

https://github.com/tc39/test262/blob/1ebd34b53b59a5cb84bdfece8931e40c462747e0/tools/scripts/ci_test.sh#L3-L4

I could make it work by changing the value of npm run diff according to one used by ci_test.sh and this workflow, and make npm run output silent(https://github.com/doehyunbaek/test262/commit/75ae1577729f8f4a5a8890583168b99a25ce0c9e, https://github.com/doehyunbaek/test262/actions/runs/3551711422). 

If you think this is a good path forward, I will push the required changes. Also if you think ci_test.sh should also be modified to use `npm run diff`, let me know.

ptomato: Oh, sorry for the bad suggestion. No need to dive into this more, I think we can just merge the PR as is. Thanks for checking it out.

.github/workflows/esmeta-test262.yml
```diff
@@ -0,0 +1,41 @@
...
+          if [ "$paths" == "" ]; then
+            echo No test files added or modified. Exiting.
+            exit 0
+          fi
+          echo New or modified test files:
+          echo "$paths"
+          "${ESMETA_HOME}"/bin/esmeta test262-test -status -test262dir=$(pwd) $paths
+
```
ljharb, line 41:
```suggestion

```

.github/workflows/esmeta-test262.yml
```diff
@@ -0,0 +1,41 @@
...
+      - name: Add to PATH
+        shell: bash
+        run: |
+          echo "ESMETA_HOME=$(pwd)/vendor/esmeta" >> $GITHUB_ENV
+      - name: download esmeta
+        run: |
+          mkdir -p "${ESMETA_HOME}"
+          git clone --branch v0.1.0-rc8 --depth 1 https://github.com/es-meta/esmeta.git "${ESMETA_HOME}"
+          cd "${ESMETA_HOME}" && git submodule update --init --depth 1
+      - name: build esmeta
```
ljharb, line 26:
```suggestion
      - name: build esmeta
         working-directory: ${{ env.ESMETA_HOME }}
```
i think this works and would avoid the need for the cd?

.github/workflows/esmeta-test262.yml
ptomato: If we wanted to do this in the future, I think we'd have to do something like `git checkout -b main origin/main`.

.github/workflows/esmeta-test262.yml
```diff
@@ -37,5 +37,4 @@ jobs:
           fi
           echo New or modified test files:
           echo "$paths"
-          "${ESMETA_HOME}"/bin/esmeta test262-test -status -test262dir=$(pwd) $paths
-          
\ No newline at end of file
+          "${ESMETA_HOME}"/bin/esmeta test262-test -status -test262dir=$(pwd) $paths
```
ljharb, line 40:
```suggestion
          "${ESMETA_HOME}"/bin/esmeta test262-test -status -test262dir=$(pwd) $paths

```
nit: to be clear, there should always be a trailing newline, but no trailing whitespace :-)

doehyunbaek:
I applied the suggested changes and if there's anything more to be done. Let me know.

ljharb (approved):

ptomato (approved):
:+1: Thanks!

PR #3731, opened 2022-11-28
Update Intl.NumberFormat test with recent reorder of option evaluation

trflynn89:
The order of evaluation for "notation" and "roundingIncrement" changed
in https://github.com/tc39/proposal-intl-numberformat-v3/commit/a260aa3.


ptomato (commented):
Thanks! This looks correct according to tc39/proposal-intl-numberformat-v3#102, but as far as I can tell that is still awaiting consensus in TG1.

Ms2ger (approved):

Issue #3524, opened 2022-05-09
Object.hasOwnProperty.call


legendecas:
I see that this `Object.hasOwnProperty.call` exists elsewhere, but I think that was a mistake and the intention was actually to use `Object.prototype.hasOwnProperty.call`. I'm not going to block on it here, but someone should do a cleanup to use `Object.prototype.hasOwnProperty.call` or `Reflect.hasOwnProperty`.

_Originally posted by @rwaldron in https://github.com/tc39/test262/pull/3512#discussion_r866315540_

creampnx-x:
Totally agree with you
https://github.com/airbnb/javascript#objects--prototype-builtins

PR #3733, opened 2022-11-29
Replace Object.hasOwnProperty.call

ptomato:
Replaces existing instances of Object.hasOwnProperty.call with Object.prototype.hasOwnProperty.call, as per #3524.

In affected lines, updates assertion messages and uses assert() instead of assert.sameValue() when we are not specifically testing the boolean-ness of the value, for clarity.

The third commit, though huge, is just the result of running the script to rebuild the generated tests. It doesn't need to be reviewed.

Closes: #3524


ljharb (approved):
rubberstamp

Ms2ger:
Lint is unhappy

ptomato:
It turns out that's what it was - a pre-existing bug where two sibling test templates shared the same `path` key meaning they were overwriting each others' tests. I've added a fix for this but I'll assume it's OK to just push it to fix the build?

PR #3696, opened 2022-10-21
Temporal: Add tests for new yearOfWeek API

ptomato:
To be presented for consensus in the November/December TC39 meeting. This adds tests for a 'yearOfWeek' getter to PlainDate, PlainDateTime, and ZonedDateTime, for use alongside 'weekOfYear', and tests for a corresponding method to Calendar.

The tests are basically the existing tests of 'weekOfYear' adapted.

Temporal issue: https://github.com/tc39/proposal-temporal/issues/2405

PR #3735, opened 2022-11-30
Temporal: Reconcile yearOfWeek tests with changes to YYYY-MM-DD+UU fmt

ptomato:
This reconciles changes made in two normative PRs to the Temporal proposal which each had their own separate PRs for tests. The yearOfWeek tests were written before the changes to the ISO 8601 grammar which disallowed the YYYY-MM-DD+UU format. This brings the yearOfWeek tests in line with the tests for the other calendar methods.

See: #3696, #3699

### DIFF
--- a/test/built-ins/Temporal/Calendar/prototype/yearOfWeek/argument-string-date-with-utc-offset.js
+++ b/test/built-ins/Temporal/Calendar/prototype/yearOfWeek/argument-string-date-with-utc-offset.js
@@ -1,0 +1,45 @@
+// Copyright (C) 2022 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.calendar.prototype.yearofweek
+description: UTC offset not valid with format that does not include a time
+features: [Temporal]
+---*/
+
+const instance = new Temporal.Calendar("iso8601");
+
+const validStrings = [
+  "2000-05-02T00+00:00",
+  "2000-05-02T00+00:00[UTC]",
+  "2000-05-02T00+00:00[!UTC]",
+  "2000-05-02T00-02:30[America/St_Johns]",
+];
+
+for (const arg of validStrings) {
+  const result = instance.yearOfWeek(arg);
+
+  assert.sameValue(
+    result,
+    2000,
+    `"${arg}" is a valid UTC offset with time for PlainDate`
+  );
+}
+
+const invalidStrings = [
+  "2022-09-15Z",
+  "2022-09-15Z[UTC]",
+  "2022-09-15Z[Europe/Vienna]",
+  "2022-09-15+00:00",
+  "2022-09-15+00:00[UTC]",
+  "2022-09-15-02:30",
+  "2022-09-15-02:30[America/St_Johns]",
+];
+
+for (const arg of invalidStrings) {
+  assert.throws(
+    RangeError,
+    () => instance.yearOfWeek(arg),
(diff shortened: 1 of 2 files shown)


test/built-ins/Temporal/Calendar/prototype/yearOfWeek/argument-string-date-with-utc-offset.js
```diff
@@ -0,0 +1,45 @@
...
+  "2022-09-15+00:00[UTC]",
+  "2022-09-15-02:30",
+  "2022-09-15-02:30[America/St_Johns]",
+];
+
+for (const arg of invalidStrings) {
+  assert.throws(
+    RangeError,
+    () => instance.yearOfWeek(arg),
+    `"${arg}" UTC offset without time is not valid for PlainDate`
```
gibson042, line 43: Is this just testing that Temporal.Calendar.prototype.yearOfWeek internally converts its argument to a PlainDate? If so, then I'd rather see that made clear and a standard suite of valid and invalid values presented to it, because most if not all other Calendar prototype methods are _also_ expected to perform such conversion, as are various other operations:
* Temporal.PlainDate.{from,compare}
* Temporal.PlainDate.prototype.{equals,since,until}
* Temporal.PlainTime.prototype.to{Plain,Zoned}DateTime
* the various `withPlainDate` methods
* {Balance,Unbalance}DurationRelative
* RoundDuration

test/built-ins/Temporal/Calendar/prototype/yearOfWeek/argument-string-date-with-utc-offset.js
```diff
@@ -0,0 +1,45 @@
+// Copyright (C) 2022 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.calendar.prototype.yearofweek
+description: UTC offset not valid with format that does not include a time
+features: [Temporal]
+---*/
+
+const instance = new Temporal.Calendar("iso8601");
```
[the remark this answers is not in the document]
ptomato: I'd prefer to keep this as is, for now ­— there are 63 other `*date-with-utc-offset.js` files and in some of them the `instance` is a different Temporal object. This makes it easier to keep them in sync. (Once Temporal gets closer to stage 4, I plan to write generated test templates for things like this.)

test/built-ins/Temporal/Calendar/prototype/yearOfWeek/argument-string-date-with-utc-offset.js
ptomato: By standard suite of valid and invalid values do you mean something like the arrays in https://github.com/tc39/test262/blob/e7895c53d295760d7c2b89743f8311fa60480392/harness/temporalHelpers.js#L1800? I agree that would be better, but I'd prefer to keep that out of scope for this PR. I'd just like this to be a quick fix for the discrepancy between #3696 and #3699.

ptomato:
Thanks for the review. I don't disagree with the comments but I'd like to keep them out of scope for this PR so that we can quickly get `main` back in sync with the current consensus and get the other Temporal PRs merged in. Holler if you disagree but otherwise I'll plan to merge this tomorrow.

test/built-ins/Temporal/Calendar/prototype/yearOfWeek/argument-string-date-with-utc-offset.js
gibson042: I do mean that, and agree on the scope of this PR. :+1:

PR #3737, opened 2022-12-01
Fix typos in String.prototype.isWellFormed/String.prototype.toWellFormed

trflynn89:
1. `String.prototype.toWellFormed`'s name is "toWellFormed".

2. `asserts` does not exist in the test harness, and `assert.throws` expects the error type first.


Ms2ger (approved):

ljharb (commented):
whoops, thanks

PR #3740, opened 2022-12-02
Improve NumberFormat and DateTimeFormat tests for OrdinaryHasInstance

Ms2ger:
This ensures that the `OrdinaryHasInstance` is actually reached in the test. It also fixes some issues with the preamble.

### DIFF
--- a/test/intl402/DateTimeFormat/constructor-no-instanceof.js
+++ b/test/intl402/DateTimeFormat/constructor-no-instanceof.js
@@ -2,11 +2,18 @@
 // This code is governed by the BSD license found in the LICENSE file.
 
 /*---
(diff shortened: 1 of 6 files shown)


ptomato (approved):
:+1:

I was initially confused by this about where the TypeError comes from. It is thrown from RequireInternalSlot, because UnwrapDateTimeFormat/UnwrapNumberFormat return undefined due to the absence of an %Intl%.[[FallbackSymbol]] property. This way we know that OrdinaryHasInstance has been called, and does not hit Symbol.hasInstance.

@ryzokuken As the original author of these, any comments?

ptomato:
No comments, so I'll merge this.

PR #3741, opened 2022-12-02
Intl: Add substantive tests for Array.prototype.toLocaleString

gibson042:
Several implementations get this method wrong, including [V8](https://bugs.chromium.org/p/v8/issues/detail?id=13564) and [GraalJS](https://github.com/oracle/graal/issues/5562).
```sh
$ npm run test:diff:v8

> test262@5.0.0 test:diff:v8
> test262-harness -t 8 --hostType=d8 --hostPath=v8 $(npm run --silent diff)

FAIL test/intl402/Array/prototype/toLocaleString/invoke-element-tolocalestring.js (strict mode)
  must invoke element toLocaleString with two arguments when provided no arguments Expected SameValue(«0», «2») to be true

FAIL test/intl402/Array/prototype/toLocaleString/invoke-element-tolocalestring.js (default)
  must invoke element toLocaleString with two arguments when provided no arguments Expected SameValue(«0», «2») to be true

Ran 2 tests
0 passed
2 failed
```

### DIFF
--- a/test/built-ins/Array/prototype/toLocaleString/invoke-element-tolocalestring.js
+++ b/test/built-ins/Array/prototype/toLocaleString/invoke-element-tolocalestring.js
@@ -1,0 +1,63 @@
+// Copyright (C) 2022 Richard Gibson. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-array.prototype.tolocalestring
+description: >
+    The toLocaleString method of each non-undefined non-null element
+    must be called with no arguments.
+info: |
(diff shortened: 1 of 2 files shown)


ptomato (approved):
LGTM, this matches my reading of the spec.

Should we have a corresponding test in `built-ins/Array/prototype/toLocaleString` checking that `toLocaleString` of each element is invoked with 0 arguments? I guess it'd have to exit early if the Intl object is present.

gibson042:
Good idea; done.

ptomato (approved):
New test LGTM as well.

PR #3742, opened 2022-12-02
Update test case style to match current patterns

gibson042:
In recent tests, I usually see a blank line after the copyright notice and `esid` as the first frontmatter key.


ptomato (approved):
+1, this matches what I've seen recently, and is also the convention that I follow.

PR #3744, opened 2022-12-02
remove invalid element decorator yield reference tests

JLHwung:
This PR fixes a regression introduced at https://github.com/tc39/test262/pull/3733.

All parts of ClassBody is in strict mode, where the `yield` reference is invalid, therefore these cases should not be generated from the class element template.

Previously they were not generated due to the path conflict.


ptomato (approved):
Thanks for tracking this down. I'm prepared to take your word for it :smile:

@legendecas, as the author of these tests, any feedback?

PR #3751, opened 2022-12-10
Support running Temporal tests in Node 19 (cont'd for staging tests)

justingrant:
Following up from #3750, this PR supports running Temporal "staging" tests (https://github.com/tc39/test262/tree/main/test/staging) under Node 19.

For context, `Intl.DateTimeFormat` implementations in Node 19 have ICU format changes:
- replace some spaces with non-breaking-space unicode characters
- change German date separators (these may be bugs)
- months in German are shown with a leading zero, but only in Year/Month formats (also may be a bug)

This PR makes affected tests more resilient to `Intl.DateTimeFormat` output changes between Node versions. They'll continue to work even if the behavior (e.g. in German) turns out to be a bug and is fixed in the platform.

### DIFF
--- a/test/staging/Intl402/Temporal/old/date-time-format.js
+++ b/test/staging/Intl402/Temporal/old/date-time-format.js
@@ -8,8 +8,36 @@ includes: [deepEqual.js]
 features: [Temporal]
 ---*/
 
-// should return an Array
+// Tolerate implementation variance by expecting consistency without being prescriptive.
+// TODO: can we change tests to be less reliant on CLDR formats while still testing that
+// Temporal and Intl are behaving as expected?
+const usDayPeriodSpace =
+  new Intl.DateTimeFormat("en-US", { timeStyle: "short" })
+    .formatToParts(0)
+    .find((part, i, parts) => part.type === "literal" && parts[i + 1].type === "dayPeriod")?.value || "";
+const usDateRangeSeparator = new Intl.DateTimeFormat("en-US", { dateStyle: "short" })
+  .formatRangeToParts(1 * 86400 * 1000, 366 * 86400 * 1000)
+  .find((part) => part.type === "literal" && part.source === "shared").value;
+const deDateRangeSeparator = new Intl.DateTimeFormat("de-AT", { dateStyle: "short" })
+  .formatRangeToParts(1 * 86400 * 1000, 366 * 86400 * 1000)
+  .find((part) => part.type === "literal" && part.source === "shared").value;
 
+// Workarounds for https:// unicode-org.atlassian.net/browse/CLDR-16243
+const deMonthDayRangeSeparator = new Intl.DateTimeFormat("de-AT", { month: "numeric", day: "numeric" })
+  .formatRangeToParts(1 * 86400 * 1000, 90 * 86400 * 1000)
+  .find((part) => part.type === "literal" && part.source === "shared").value;
+const deMonthYearSeparator = new Intl.DateTimeFormat("de-AT", { year: "numeric", month: "numeric" })
+  .formatToParts(0)
+  .find((part) => part.type === "literal").value;
+const deMonthInYearMonthHasLeadingZero = new Intl.DateTimeFormat("de-AT", {
+  year: "numeric",
+  month: "numeric"
+})
+  .formatToParts(new Date(2000, 3, 1))
+  .find((part) => part.type === "month")
+  .value.startsWith("0");
+
+// should return an Array
 assert(Array.isArray(Intl.DateTimeFormat.supportedLocalesOf()));
 var onlyOnce = (value) => {
   var obj = {
@@ -52,19 +80,19 @@ var start = new Date("1922-12-30");
 var end = new Date("1991-12-26");
 
 // should work for Instant
-assert.sameValue(us.format(Temporal.Instant.from(t1)), "11/18/1976, 9:23:30 AM");
+assert.sameValue(us.format(Temporal.Instant.from(t1)), `11/18/1976, 9:23:30${usDayPeriodSpace}AM`);
 assert.sameValue(at.format(Temporal.Instant.from(t1)), "18.11.1976, 15:23:30");
 
 // should work for ZonedDateTime
-assert.sameValue(us2.format(Temporal.ZonedDateTime.from(t1)), "11/18/1976, 2:23:30 PM UTC");
+assert.sameValue(us2.format(Temporal.ZonedDateTime.from(t1)), `11/18/1976, 2:23:30${usDayPeriodSpace}PM UTC`);
 assert.sameValue(at2.format(Temporal.ZonedDateTime.from(t1)), "18.11.1976, 14:23:30 UTC");
 
 // should work for DateTime
-assert.sameValue(us.format(Temporal.PlainDateTime.from(t1)), "11/18/1976, 2:23:30 PM");
+assert.sameValue(us.format(Temporal.PlainDateTime.from(t1)), `11/18/1976, 2:23:30${usDayPeriodSpace}PM`);
 assert.sameValue(at.format(Temporal.PlainDateTime.from(t1)), "18.11.1976, 14:23:30");
 
 // should work for Time
-assert.sameValue(us.format(Temporal.PlainTime.from(t1)), "2:23:30 PM");
+assert.sameValue(us.format(Temporal.PlainTime.from(t1)), `2:23:30${usDayPeriodSpace}PM`);
 assert.sameValue(at.format(Temporal.PlainTime.from(t1)), "14:23:30");
 
 // should work for Date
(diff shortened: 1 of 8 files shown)


test/staging/Intl402/Temporal/old/date-time-format.js
```diff
@@ -616,42 +639,56 @@ assert.deepEqual(at.formatToParts(end), [
...
 //should work for MonthDay 
 var date1 = Temporal.PlainDate.from(t1);
 var date2 = Temporal.PlainDate.from(t2);
-assert.sameValue(us.formatRange(date2.withCalendar(usCalendar).toPlainMonthDay(), date1.withCalendar(usCalendar).toPlainMonthDay()), "2/20 \u2013 11/18");
-assert.sameValue(at.formatRange(date2.withCalendar(atCalendar).toPlainMonthDay(), date1.withCalendar(atCalendar).toPlainMonthDay()), "20.02. \u2013 18.11.");
+assert.sameValue(
+  us.formatRange(date2.withCalendar(usCalendar).toPlainMonthDay(), date1.withCalendar(usCalendar).toPlainMonthDay()),
+  `2/20${dateRangeSeparator}11/18`
+);
+// In Node 19.2.0, apparently the german Month/Day format uses regular spaces, but all other formats and locales seem to use thin spaces ('\u{2009}');
```
gibson042, line 679: There seems to be a lot of variability with de-AT; perhaps it would make sense to pick a different non-English locale if these comparisons truly are important?

gibson042 (changes requested):

test/staging/Intl402/Temporal/old/date-time-format.js
justingrant: After quite a bit of digging, the problem looks like a regression in CLDR 42's German date formats. Frustratingly, the Node 19 build used by GitHub Actions picks up ICU 72 (which includes CLDR 42 with the problems) but `brew install node` on my local machine currently includes ICU 71/CLDR 41. This explains why I only saw it in CI and why this problem has been so frustrating to track down!

I filed a bug in CLDR's JIRA to fix the formats: https://unicode-org.atlassian.net/browse/CLDR-16243

Now that I understand the actual root cause, I can remove a lot of the German-specific code because only month/year formats are affected, not all date formats. This should clean up this PR quite a bit. I'll do that tomorrow and will ping you to review again once that's done.

Thanks for pushing me to dig into this!

test/staging/Intl402/Temporal/old/date-time-format.js
```diff
@@ -9,7 +9,23 @@ includes: [deepEqual.js]
...
+)[5].value;
+// May be an Intl bug: the de-AT date separator changed from "." to "/" in Node 19
+const germanDateSeparator = new Intl.DateTimeFormat("de-AT", { dateStyle: "short", calendar: "gregory" }).formatToParts(
+  0
+)[1].value;
+// Probably an Intl bug: the month and year format seems to use a different separator character (in Node 19.2.0)
+// than other date formats.
+const germanMonthYearSeparator = new Intl.DateTimeFormat("de-AT", { year: "numeric", month: "numeric" }).formatToParts(
+  0
+)[1].value;
```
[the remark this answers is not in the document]
justingrant: Good questions!  I think your changes are good ones, I'll include them in the next push of this PR, hopefully tomorrow.  Also (see discussion below about the German formats) it's likely that much of the German-specific code in this PR will go away, which should further simplify things and shrink the diff.

I agree that it would be good to further reduce the dependency of these tests on CLDR string formats. But I don't think I'll have time to do that work myself. For this PR, I was kinda hoping to do the minimum changes required to get Test262 to pass in both Node 18 and Node 19 so we don't block anyone's testing as folks start upgrading their CIs to v19.

If OK with you, I'd like to leave the "refactor to reduce CLDR format dependency" as an exercise for the future. Perhaps whoever ports these "staging" tests into the mainstream Test262 could also look into refactoring them? Does this sound OK?

justingrant:
@gibson042 - I made the changes you recommended, rebased to main, and cleaned up commits by running prettier first and then layering my changes on top so the diff will be easier to review.

test/staging/Intl402/Temporal/old/date-time-format.js
```diff
@@ -518,6 +537,15 @@ assert.deepEqual(us.formatToParts(t.withCalendar(usCalendar).toPlainYearMonth())
...
 ]);
+
+// Work around variability in month format caused by https:// unicode-org.atlassian.net/browse/CLDR-16243
+var deMonthInYearMonthHasLeadingZero = new Intl.DateTimeFormat("de-AT", {
+  year: "numeric",
+  month: "numeric"
+})
+  .formatToParts(new Date(2000, 3, 1))
+  .find((part) => part.type === "month")
+  .value.startsWith("0");
```
gibson042, line 548: This appears to be unused.
```suggestion
```

test/staging/Intl402/Temporal/old/date-time-format.js
```diff
@@ -680,10 +714,11 @@ assert.sameValue(
...
 
 // should not break legacy Date
-assert.sameValue(us.formatRange(start, end), "12/29/1922 \u2013 12/25/1991");
-assert.sameValue(at.formatRange(start, end), "30.12.1922 \u2013 26.12.1991");
+assert.sameValue(us.formatRange(start, end), `12/29/1922${usDateRangeSeparator}12/25/1991`);
+assert.sameValue(at.formatRange(start, end), `30.12.1922${deDateRangeSeparator}26.12.1991`);
 
-// should throw a TypeError when called with dissimilar types", () => assert.throws(TypeError, () => us.formatRange(Temporal.Instant.from(t1), Temporal.PlainDateTime.from(t2))));
+// should throw a TypeError when called with dissimilar types",
+//  () => assert.throws(TypeError, () => us.formatRange(Temporal.Instant.from(t1), Temporal.PlainDateTime.from(t2))));
```
gibson042, line 721: Looks like this is actually tested below.
```suggestion
```

test/staging/Intl402/Temporal/old/instant-toLocaleString.js
```diff
@@ -7,6 +7,12 @@ description: Instant.toLocaleString()
 features: [Temporal]
 ---*/
 
+// Tolerate implementation variance by expecting consistency without being prescriptive.
+const usDayPeriodSpace =
+  new Intl.DateTimeFormat("en-US", { timeStyle: "short" })
```
gibson042, line 12: Shouldn't this be used in the first assertion, or are we looking at a difference between "en-US" and "en"?
```diff
-assert.sameValue(`${instant.toLocaleString("en", { timeZone: "America/New_York" })}`, "11/18/1976, 9:23:30 AM");
+assert.sameValue(`${instant.toLocaleString("en", { timeZone: "America/New_York" })}`, `11/18/1976, 9:23:30${usDayPeriodSpace}AM`);
```

test/staging/Intl402/Temporal/old/time-toLocaleString.js
```diff
@@ -7,6 +7,12 @@ description: time.toLocaleString()
 features: [Temporal]
 ---*/
 
+// Tolerate implementation variance by expecting consistency without being prescriptive.
+const usDayPeriodSpace =
+  new Intl.DateTimeFormat("en-US", { timeStyle: "short" })
```
gibson042, line 12: Shouldn't this be used in the assertions, or are we looking at a difference between "en-US" and "en"?

gibson042 (approved):

test/staging/Intl402/Temporal/old/instant-toLocaleString.js
justingrant: Good catch. It looks like there's a mix in these tests between "en" & "de" vs. "en-US" and "de-AT", which I assume was an oversight from whoever originally wrote these tests. I'll just normalize them all into "en-US" and "de-AT" throughout all these files, and I'll put that into a separate commit to keep the noise down.

test/staging/Intl402/Temporal/old/date-time-format.js
justingrant: Looks like the original port from demitasse tests forgot to remove this commented line. I'll remove it.

test/staging/Intl402/Temporal/old/date-time-format.js
justingrant: Whoops, looks like I lost this check during the last rebase. The next line should have used this value in a conditional:
```js
assert.deepEqual(at.formatToParts(t.withCalendar(atCalendar).toPlainYearMonth()), [
  {
    type: "month",
    value: deMonthInYearMonthHasLeadingZero ? "02" : "2"
  },
```

Will be fixed in the next push. 

It's actually lucky you found this, because it looks like something went wrong on the proposal-temporal side of this PR and it wasn't picking up the right Test262 subproject commit. Otherwise the proposal-temporal PR should have failed. Investigating that now.  Thanks!

test/staging/Intl402/Temporal/old/date-time-format.js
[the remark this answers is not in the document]
justingrant: Yep, done!

test/staging/Intl402/Temporal/old/date-time-format.js
justingrant: I ended up leaving German as the locale because AFAICT would be a lot of work to change it throughout all these tests, and it's relatively straightforward to write tests that ignore the German issues. I added TODO comments that encourage future contributors to consider tests that have less dependence on CLDR formats.

justingrant:
@gibson042 All code review issues should be resolved, but there's been enough change that it may be worth one more review before merging.

I apologize for all the commit notification spam from this PR. I wasn't able to figure out how to get my local install of Node 19 to use ICU 72 like GH Actions does, so I had to use CI to validate all changes. Grrrrr.

gibson042 (approved):

PR #3754, opened 2022-12-20
Add tests for the asyncItems argument to Array.fromAsync.

Ms2ger:
Depends on #3724 (RFC) and #3728.


test/built-ins/Array/fromAsync/asyncitems-array-add-to-empty.js
```diff
@@ -0,0 +1,19 @@
...
+flags: [async]
+features: [Array.fromAsync]
+---*/
+
+asyncTest(async function () {
+  const items = [];
+  const promise = Array.fromAsync(items);
+  items.push(7);
+  const result = await promise;
+  assert.compareArray(result, []);
```
ptomato, line 18: Should this be `[7]` or is it intentionally omitted? I've been poring over the spec and don't see where an empty array is treated differently than an array with elements in it.

ptomato: Or is it something like, the first iterator `next()` happens before control returns to JS, while subsequent ones happen in following ticks? This and the other similar tests below could probably use a comment explaining what's going on.

test/built-ins/Array/fromAsync/asyncitems-arraylike-promise.js
```diff
@@ -0,0 +1,30 @@
...
+---*/
+
+asyncTest(async function () {
+  const actual = [];
+  const items = {};
+  TemporalHelpers.observeProperty(actual, items, Symbol.asyncIterator, undefined, "items");
+  TemporalHelpers.observeProperty(actual, items, Symbol.iterator, undefined, "items");
+  TemporalHelpers.observeProperty(actual, items, "length", 2, "items");
+  TemporalHelpers.observeProperty(actual, items, 0, Promise.resolve(2), "items");
+  TemporalHelpers.observeProperty(actual, items, 1, Promise.resolve(1), "items");
```
ptomato, line 20: Well... perhaps this is a sign we need to separate out these observer APIs into a separate helper file; they certainly aren't specific to Temporal. But I'd say that's out of scope for this PR.

I think it could be simplified to this:
```js
const items = TemporalHelpers.propertyBagObserver(actual, {
  length: 2,
  0: Promise.resolve(2),
  1: Promise.resolve(1),
}, "items");
```
(and similar for some of the tests below with repeated calls to TemporalHelpers.observeProperty)

test/built-ins/Array/fromAsync/asyncitems-asynciterator-sync.js
```diff
@@ -0,0 +1,32 @@
+// Copyright (C) 2022 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-array.fromasync
+description: >
+  Array.fromAsync handles a sync iterator returned from @@iterator
```
ptomato, line 7:
```suggestion
  Array.fromAsync handles a sync iterator returned from @@asyncIterator
```

test/built-ins/Array/fromAsync/asyncitems-function.js
```diff
@@ -0,0 +1,20 @@
+// Copyright (C) 2022 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-array.fromasync
+description: >
+  Array.fromAsync treats a BigInt as an array-like
```
ptomato, line 7:
```suggestion
  Array.fromAsync treats a function as an array-like
```

test/built-ins/Array/fromAsync/asyncitems-iterator-async.js
```diff
@@ -0,0 +1,35 @@
...
+
+asyncTest(async function () {
+  async function * asyncGen() {
+    for (let i = 0; i < 4; i++) {
+      yield Promise.resolve(i * 2);
+    }
+  }
+
+
+  return; // check what this should do
```
ptomato, line 21: Leftover from debugging?

test/built-ins/Array/fromAsync/asyncitems-bigint.js
```diff
@@ -0,0 +1,20 @@
+// Copyright (C) 2022 Igalia, S.L. All rights reserved.
```
ptomato, line 1: I'd suggest maybe putting all the builtins-as-arraylikes in the same file?

test/built-ins/Array/fromAsync/asyncitems-null-undefined.js
```diff
@@ -0,0 +1,18 @@
+// Copyright (C) 2022 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-array.fromasync
+description: >
+  Array.fromAsync rejects with a TypeError if the asyncItems argument is null or undefined
+info: |
+  3.c. Let usingAsyncIterator be ? GetMethod(asyncItems, @@asyncIterator).
```
ptomato, line 9: Fine with not having the chapter and verse, but if it's present then I believe the place the error is thrown is:
```suggestion
  3.k.ii. Let arrayLike be ! ToObject(asyncItems).
```
(and I believe that ! should be a ? in the proposal)

ptomato (commented):
Thanks!

Ms2ger (commented):
Will review your other comments later

test/built-ins/Array/fromAsync/asyncitems-arraylike-promise.js
Ms2ger: This causes issues when comparing the output, because they're arrays of `TemporalHelpers.toPrimitiveObserver` objects.

test/built-ins/Array/fromAsync/asyncitems-iterator-async.js
Ms2ger: Turns out this scenario doesn't do anything interesting (besides timing out)

test/built-ins/Array/fromAsync/asyncitems-bigint.js
Ms2ger: I'd rather not, since they're relying on messing with global state.

test/built-ins/Array/fromAsync/asyncitems-arraylike-promise.js
ptomato: Fair enough.

test/built-ins/Array/fromAsync/asyncitems-array-add-to-empty.js
ptomato: Thanks for the explanation, this helps a lot.

test/built-ins/Array/fromAsync/asyncitems-iterator-async.js
ptomato: So, should we just not have this file?

test/built-ins/Array/fromAsync/asyncitems-null-undefined.js
ptomato: Up to you. On a second look it was actually pretty obvious, so I must have just had a blind spot for it on the previous review.

ptomato (approved):
Thanks.

Optionally, consider adding a comment like `// See explanation in asyncitems-array-add-to-empty.js` to the other tests that rely on the first `next()` call having happened by the time we await the promise, like in `asyncitems-array-add-to-singleton.js`?

test/built-ins/Array/fromAsync/asyncitems-iterator-async.js
Ms2ger: Huh, I was sure I deleted it

PR #3755, opened 2022-12-20
Improve is/toWellFormedString coverage.

Ms2ger:


ljharb (approved):
Confirmed, I added these same tests in my polyfills and they passed :-)

also thanks for catching those other mistakes

PR #3745, opened 2022-12-05
Temporal operation NanosecondsToDays: Test RangeErrors reachable in Duration and ZonedDateTime

cjtenny:
Corresponds to https://github.com/tc39/proposal-temporal/pull/2387 and follow-up https://github.com/tc39/proposal-temporal/pull/2444. Many thanks again to @anba for the inspiration and help.

PR #3757, opened 2022-12-22
Temporal operation NanosecondsToDays: Test infinite or arbitrary-length observable loops in Duration and ZonedDateTime

cjtenny:
Extends previous coverage developed by @anba to all other callers that can reach NanosecondsToDays with a user-provided `relativeTo` and can therefore trigger the indefinite-length loops; refactors @anba's code for the original loops to try to emphasize a few things - though perhaps at the expense of making some other values/interactions less clear, feedback welcome please!

Loop 1:
1) The loop is of arbitrary but constrained length), and only one call to `relativeTo.calendar.dateUntil` determines the length of the loop.
2) The loop will call several observable methods ^ # of times
3) Show that the loop occurs the user-provided number of times for different values

Loop 2:
1) The loop is infinite, and can continue as long as `relativeTo.timeZone.getPossibleInstantsFor` returns values that allow the loop to continue.
2) The loop continues calling ^ until it returns values that allow the loop to terminate.
3) Show that the loop can persist as long as ^ for several different loop lengths.

Based upon #3745 - uses changes made to temporalHelpers - will update when that's merged so there's only one commit here.

### DIFF
--- a/harness/temporalHelpers.js
+++ b/harness/temporalHelpers.js
@@ -1383,6 +1383,20 @@ var TemporalHelpers = {
   },
 
   /*
+   * observeMethod(calls, object, propertyName, value):
+   *
+   * Defines an own property @object.@propertyName with value @value, that
+   * will log any calls of @value to the array @calls.
+   */
+  observeMethod(calls, object, propertyName, objectName = "") {
+    const method = object[propertyName];
+    object[propertyName] = function () {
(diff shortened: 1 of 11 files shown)


Aditi-1400 (approved):

ptomato (approved):
Thanks.

PR #3759, opened 2023-01-04
Temporal: Update Temporal test262 as per the normative changes

Aditi-1400:
Updates Temporal tests as per the normative change in PR: https://github.com/tc39/proposal-temporal/pull/2438


ptomato (approved):
Thanks!

PR #3760, opened 2023-01-05
Add a feature for `with`

raskad:
I'd like to propose adding a feature for tests containing the `with` language feature.

The use case that I have is pretty simple; in the [boa](https://github.com/boa-dev/boa) engine we currently do not have the `with` statement implemented and would like to ignore all tests containing it. This makes it a bit easier to see what features we have fully implemented right now, as some tests outside of the `test/language/statements/with` folder rely on the `with` statement and skew the results. It also makes it a bit easier to look for tests to fix in our overview.

I'd imagine that some engines may never want to implement `with` due to it's legacy nature. Making it easy to ignore while running the tests may also be nice for them.

I have added `with` to the feature list and added the feature to all tests that utilize it (that I could find using basic text search).

### DIFF
--- a/features.txt
+++ b/features.txt
@@ -383,6 +383,7 @@ Uint32Array
 Uint8ClampedArray
 WeakMap
 WeakSet
+with
 __proto__
 __getter__
 __setter__
--- a/test/built-ins/Function/StrictFunction_reservedwords_with.js
+++ b/test/built-ins/Function/StrictFunction_reservedwords_with.js
@@ -14,6 +14,7 @@ info: |
          If strict is true, the Early Error rules for UniqueFormalParameters:FormalParameters are applied.
       ...
     ...
+features: [with]
 ---*/
 
 assert.throws(SyntaxError, function() {
--- a/test/built-ins/Proxy/has/call-with.js
+++ b/test/built-ins/Proxy/has/call-with.js
@@ -11,7 +11,7 @@ info: |
     9. Let booleanTrapResult be ToBoolean(Call(trap, handler, «target, P»)).
     ...
 flags: [noStrict]
-features: [Proxy]
+features: [Proxy, with]
 ---*/
 
 var _handler, _target, _prop;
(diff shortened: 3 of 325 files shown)


ptomato:
Thanks! A few notes:
- The CI failure is due to some of the edited tests being generated code. You'll have to find their sources in the `src/` folder and edit those as well.
- This is a reversal of how feature flags are usually used - to mark features that are so new that engines might not have implemented them yet, rather than to mark features that have always been present in the language but are so obsolete that engines might not want to implement them. It's debatable whether we want test262 to facilitate _less_ interoperability between engines ... but I can see this being useful regardless, for new engines that put `with` at a low priority relative to other features. I'm curious what other maintainers think about this.
- I didn't look closely at all the tests you marked but if there are any using a `with` statement, that aren't testing the `with` statement specifically, then those should probably be rewritten not to use `with`.
- Using acorn to get an AST of each test file and walking the tree to see if any WithStatement nodes are present would probably be more precise than a text search, but I think this is probably fine. Walking an AST is the approach I'd take if we ever wanted to automate this, though.

ljharb:
While it's possible to have a spec-compliant engine that doesn't implement, say, optional chaining (because the spec it complies with might be one prior to that feature's introduction), It's not possible to have a spec-compliant engine that doesn't implement `with`.

While I'm sympathetic to the annoyance of an in-progress implementation needing to filter out test failures, I have no interest in supporting *in any way* an engine that is intending to be willfully noncompliant by failing to ever implement `with`.

Note that you can use a custom preprocessor and filter out tests by various attributes, like here: https://github.com/es-shims/all/blob/main/package.json#L31 / https://github.com/es-shims/all/blob/main/test/preprocessor.js

raskad:
This is only a quality of life improvement for active engine development. If there is that much to be discussed, it is probably not worth it.

We could of course just filter the tests in our tester with any method, but I thought a feature may benefit others too.

I will close this based on the discussion.

PR #3761, opened 2023-01-05
Temporal: Update custom Calendar/TimeZone tests to cover rejection of invalid output

gibson042:
Temporal issue: https://github.com/tc39/proposal-temporal/issues/2443
Temporal pull request: https://github.com/tc39/proposal-temporal/pull/2456


test/built-ins/Temporal/PlainDate/prototype/day/validate-calendar-value.js
```diff
@@ -8,18 +8,23 @@ features: [Temporal]
...
   [Symbol("foo"), TypeError],
   [7n, TypeError],
-  [{}, RangeError],
+  [{}, TypeError],
+  [true, TypeError],
+  [7.1, RangeError],
+  ["7", TypeError],
+  ["7.5", TypeError],
+  [{valueOf() { return 7; }}, TypeError],
 ];
```
gibson042, line 28: This list and variations thereof is repeated **a lot**. It would be nice to consolidate them such that the coverage of Temporal (and non-Temporal!) tests could clearly be evaluated:
```js
const badResults = COMMON_VALUES.filter(({ value }) => !Number.isInteger(value));
badResults.forEach(({ label, value }) => {
  const expectedError = typeof value === "number" ? RangeError : TypeError;
  const calendar = new class extends Temporal.Calendar {
    day: () => result
  };
  const instance = new Temporal.PlainDate(…, calendar);
  assert.throws(expectedError, () => instance.day,
    `Calendar instance day() output should be rejected: ${label}`);
});
```

where `COMMON_VALUES` contains inputs (and metadata about them) for covering [Type Conversion operations](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-type-conversion) and related operations:
* at least one instance of each primitive type
* `null`
* both boolean values
* registered and unregistered symbols with empty and nonempty descriptions
* a built-in symbol
* edge-case Number and BigInt values [`NaN`, ±Infinity, ±0, ±(2**{8,16,32,53,64} ± {0,1})]
* positive and negative non-integer Numbers that round differently [e.g., ±({0,1,2} + {0.1,0.5,0.9})]
* decimal string representations of the preceding two categories
* empty string
* objects with combinations of {toString,valueOf,Symbol.toPrimitive} methods that return each value of the preceding points

ptomato: That seems like it'd be useful to have, but a _much_ larger effort than this PR.

ptomato (approved):
For full coverage, we should also have similar tests for the various prototypes' `era` and `eraYear` properties in `intl402/`.

test/built-ins/Temporal/PlainDate/prototype/day/validate-calendar-value.js
gibson042: For future reference, [harness/typeCoercion.js](https://github.com/tc39/test262/blob/2ac0d3c47b1e5cdf4319eec13cc53043f8427a97/harness/typeCoercion.js) seems to be similarly motivated (although it does not provide the kind of interface I'm seeking).

gibson042:
@ptomato Done.

ptomato:
This change achieved consensus at the 2023-01-31 TC39 plenary meeting.

PR #3676, opened 2022-09-27
Update expected results for CLDR 42

anba:
Draft until CLDR 42 is published.

PR #3762, opened 2023-01-10
Fix more 402 Temporal tests broken by ICU72/CLDR42

justingrant:
Fixes 4 more Temporal-related tests that break in ICU 72 / CLDR 42.

Follows up on https://github.com/tc39/test262/pull/3676#issuecomment-1302685505 and #3751.

### DIFF
--- a/test/intl402/DateTimeFormat/prototype/format/temporal-objects-resolved-time-zone.js
+++ b/test/intl402/DateTimeFormat/prototype/format/temporal-objects-resolved-time-zone.js
@@ -8,7 +8,15 @@ locale: [en]
 features: [Temporal]
 ---*/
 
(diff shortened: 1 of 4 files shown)


ptomato (approved):
Thanks for catching and fixing these!

I'll merge this right away in order to fix the breakage, but I think we can probably improve these tests more in the future. I'll write down my notes here so I can find them later:

- The point of the tests is that the formatter gives the same date as the PlainDateTime despite the formatter being constructed with a time zone with a large offset.
- Probably it'd be sufficient to ensure that the correct numbers were in the string.
- For `formatToParts`/`formatRangeToParts`, we can verify this directly.
- For `format/`formatRange`, maybe asserting something like `result.includes('8/4/2021')` or comparing with a regexp `/8[^0-9]+4[^0-9]+2021/

PR #3765, opened 2023-01-11
Final batch of CLDR backwards-compatibility fixes

justingrant:
Following up on #3751 and #3762, this commit makes a few tests work on both CLDR 42 and CLDR 41. Previously these tests were tied to a specific CLDR 42 format.


Ms2ger (approved):
Ugh.

PR #3772, opened 2023-01-18
Update tointlmathematicalvalue tests

romulocintra:
Added tests for [Limit exponent of intl mathematical value #98
](https://github.com/tc39/proposal-intl-numberformat-v3/pull/128)

### DIFF
--- a/test/intl402/NumberFormat/constructor-max-and-min-exponent-limits.js
+++ b/test/intl402/NumberFormat/constructor-max-and-min-exponent-limits.js
@@ -1,0 +1,19 @@
+// Copyright 2023 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-tointlmathematicalvalue
+description: Tests exponent limits
+features: [Intl.NumberFormat-v3]
+---*/
+
+const nf = Intl.NumberFormat();
+
+assert.sameValue(nf.format(-100e309), "-∞", "Should output -∞");
+assert.sameValue(nf.format(-100e-309), "-0", "Should output  -0");
+assert.sameValue(nf.format(100e309), "∞", "Should output  ∞");
+assert.sameValue(nf.format(100e-309), "0", "Should output  0");
+assert.sameValue(nf.format(Number.MAX_VALUE * -2), "-∞", "Should output  -∞");
+assert.sameValue(nf.format(Number.MAX_VALUE * 2), "∞", "Should output ∞");
+assert.sameValue(nf.format(Number.MIN_VALUE * -2), "-0", "Should output -0");
+assert.sameValue(nf.format(Number.MIN_VALUE * 2), "0", "Should output  0");


test/intl402/NumberFormat/constructor-max-and-min-exponent-limits.js
```diff
@@ -0,0 +1,19 @@
...
+
+/*---
+esid: sec-tointlmathematicalvalue
+description: Tests exponent limits
+features: [Intl.NumberFormat-v3]
+---*/
+
+const nf = Intl.NumberFormat();
+
+assert.sameValue(nf.format(-100e309), "-∞", "Should output -∞");
```
sffc, line 12: Request: please test values that are _slightly below_, _exactly equal to_, and _slightly above_ the cutoff points.

test/intl402/NumberFormat/constructor-max-and-min-exponent-limits.js
```diff
@@ -0,0 +1,19 @@
...
+assert.sameValue(nf.format(-100e-309), "-0", "Should output  -0");
+assert.sameValue(nf.format(100e309), "∞", "Should output  ∞");
```
sffc, line 14: Suggestion: pending the final resolution on https://github.com/tc39/proposal-intl-numberformat-v3/pull/128, we should offer two valid outputs: the `"∞"` output as well as the long form `1,000,000,...,000` output.

sffc (changes requested):
Thanks, @romulocintra!

test/intl402/NumberFormat/constructor-max-and-min-exponent-limits.js
sffc: This comment is obsolete now. There is one deterministic result (technically two due to float64 rounding).

test/intl402/NumberFormat/constructor-max-and-min-exponent-limits.js
```diff
@@ -0,0 +1,33 @@
...
+assert.sameValue(nf.format(100e310), "∞", "Should output  ∞");
+
+assert.sameValue(nf.format(100e-308), "0", "Should output  0");
+assert.sameValue(nf.format(100e-309), "0", "Should output  0");
+assert.sameValue(nf.format(100e-310), "0", "Should output  0");
+
+assert.sameValue(nf.format(Number.MAX_VALUE * -2), "-∞", "Should output  -∞");
+assert.sameValue(nf.format(Number.MAX_VALUE * 2), "∞", "Should output ∞");
+assert.sameValue(nf.format(Number.MIN_VALUE * -2), "-0", "Should output -0");
+assert.sameValue(nf.format(Number.MIN_VALUE * 2), "0", "Should output  0");
```
ptomato, line 33:
```suggestion
assert.sameValue(nf.format(Number.MIN_VALUE / -2), "-0", "Should output -0");
assert.sameValue(nf.format(Number.MIN_VALUE / 2), "0", "Should output  0");
```
?

(I'd expect that `Number.MIN_VALUE * 2` would format 1e-323, for example)

sffc: `Number.MIN_VALUE / 2` is equivalent to `0`

ptomato: Ah, right. But `Number.MIN_VALUE * 2` still shouldn't output 0?

sffc: Well by default the NumberFormat is configured with maximumFractionDigits: 3, so all these values will round to 0. In order to test this we need to configure the NumberFormat with maximumSignificantDigits instead.

PR #3773, opened 2023-01-18
ArrayBuffer.prototype.detached tests

ljharb:
See https://tc39.es/proposal-arraybuffer-transfer/

### DIFF
--- a/test/built-ins/ArrayBuffer/prototype/detached/detached-buffer-resizable.js
+++ b/test/built-ins/ArrayBuffer/prototype/detached/detached-buffer-resizable.js
@@ -1,0 +1,36 @@
+// Copyright (C) 2023 Jordan Harband. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-get-arraybuffer.prototype.detached
+description: Return a boolean indicating if the ArrayBuffer is detached
+info: |
+  get ArrayBuffer.prototype.detached
+
+  1. Let O be the this value.
+  2. Perform ? RequireInternalSlot(O, [[ArrayBufferData]]).
+  3. If IsSharedArrayBuffer(O) is true, throw a TypeError exception.
+  4. Return IsDetachedBuffer(O).
+includes: [detachArrayBuffer.js]
+features: [ArrayBuffer, arraybuffer-transfer, resizable-arraybuffer]
+---*/
+
+var ab1 = new ArrayBuffer(0, { maxByteLength: 0 });
+assert.sameValue(ab1.detached, false, 'Resizable ArrayBuffer with maxByteLength of 0 is not detached');
+
+$DETACHBUFFER(ab1);
+
+assert.sameValue(ab1.detached, true, 'Resizable ArrayBuffer with maxByteLength of 0 is now detached');
+
+var ab2 = new ArrayBuffer(0, { maxByteLength: 23 });
+assert.sameValue(ab2.detached, false, 'Resizable ArrayBuffer with maxByteLength of 23 is not detached');
+
+$DETACHBUFFER(ab2);
+
+assert.sameValue(ab2.detached, true, 'Resizable ArrayBuffer with maxByteLength of 23 is now detached');
+
+var ab3 = new ArrayBuffer(42, { maxByteLength: 42 });
+assert.sameValue(ab3.detached, false, 'Resizable ArrayBuffer with maxByteLength of 42 is not detached');
+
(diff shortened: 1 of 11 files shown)


test/built-ins/ArrayBuffer/prototype/detached/this-is-sharedarraybuffer-resizable.js
```diff
@@ -0,0 +1,29 @@
...
+var detached = Object.getOwnPropertyDescriptor(
+  ArrayBuffer.prototype, "detached"
+);
+
+var getter = detached.get;
+var sab = new SharedArrayBuffer(4);
+
+assert.sameValue(typeof getter, "function");
+
+Object.defineProperties(sab, { maxByteLength: maxByteLength });
```
jugglinmike, line 25: This seems like a mistake--`maxByteLength` isn't defined, and defining any property won't make the ArrayBuffer resizable.

test/built-ins/ArrayBuffer/prototype/detached/this-is-sharedarraybuffer-resizable.js
```diff
@@ -0,0 +1,29 @@
...
+
+assert.throws(TypeError, function() {
+  sab.detached;
```
jugglinmike, line 28: We'll need something like `getter.call` in here since at the moment, the proposal doesn't define `detached` on the SharedArrayBuffer prototype.

features.txt
```diff
@@ -252,6 +252,10 @@ resizable-arraybuffer
 # https://github.com/tc39/proposal-arraybuffer-transfer
 arraybuffer-transfer
 
+# ArrayBuffer detached
+# https://tc39.es/proposal-arraybuffer-transfer/
+arraybuffer-detached
```
jugglinmike, line 257: Could we re-use the flag `arraybuffer-transfer` for these tests? `arraybuffer-detached` may be incorrectly assumed to mean "tests involving detached ArrayBuffers." `arraybuffer-prototype-detached` would be more clear in that sense, but as a general rule, I prefer to use a single feature flag for a given proposal. It's an easier policy to apply regardless of proposal size, and more granularity doesn't buy much in most cases: folks that want to filter for certain APIs can lean on the directory structure.

jugglinmike (changes requested):

test/built-ins/ArrayBuffer/prototype/detached/detached-buffer-resizable.js
```diff
@@ -0,0 +1,41 @@
...
+$DETACHBUFFER(ab2);
+
+assert.sameValue(ab2.detached, true);
+
+var ab3 = new ArrayBuffer(42, { maxByteLength: 42 });
+assert.sameValue(ab3.detached, false);
+
+$DETACHBUFFER(ab3);
+
+assert.sameValue(ab3.detached, true);
```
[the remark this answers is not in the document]
ljharb: Nothing does, really - I just copied them from the byteLength tests, so it's just three different lengths. I'll add messages tho.

features.txt
ljharb: We could, but they're three separate methods, and there's no guarantee they'll be implemented independently.

syg: Please reuse the same `arraybuffer-transfer` flag. There is a 99.9% chance they will be implemented together, since implementations implement proposals at once, not cherrypick individual methods out of proposals...

syg (approved):
Thanks for writing the tests!

lgtm % changing the feature tag to `arraybuffer-detached`

features.txt
ljharb: I don't agree that's the main point of the flags _now_, but either way I've updated to reuse the transfer flag.

Ms2ger:
```
/home/circleci/test262/test/built-ins/ArrayBuffer/prototype/detached/invoked-as-accessor.js: INCLUDES - Unused include: "detachArrayBuffer.js"
/home/circleci/test262/test/built-ins/ArrayBuffer/prototype/detached/invoked-as-func.js: INCLUDES - Unused include: "detachArrayBuffer.js"
/home/circleci/test262/test/built-ins/ArrayBuffer/prototype/detached/length.js: INCLUDES - Unused include: "detachArrayBuffer.js"
/home/circleci/test262/test/built-ins/ArrayBuffer/prototype/detached/this-is-sharedarraybuffer.js: POSIX - File must end with a newline
```

Ms2ger (approved):

PR #3774, opened 2023-01-18
Update ArrayBuffer.prototype.transfer tests

syg:
This PR updates the existing transfer tests as well as add a new feature flag for it, since the proposal was split from resizable buffers.

### DIFF
--- a/features.txt
+++ b/features.txt
@@ -248,6 +248,10 @@ json-modules
 # https://github.com/tc39/proposal-resizablearraybuffer
 resizable-arraybuffer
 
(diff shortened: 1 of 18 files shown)


ljharb (approved):

ljharb:
(related to #3773)

ptomato (approved):
Thanks for re-tagging these :smile: 

#3773 adds tests for `.detached`, these are for `.transfer()`, so this is a note to myself that tests for `.fix()` are still pending.

PR #3775, opened 2023-01-29
Test Array.prototype.findLast* length clamping

gibson042:
* [Array.prototype.findLast](https://tc39.es/ecma262/multipage/indexed-collections.html#sec-array.prototype.findlast)
* [Array.prototype.findLastIndex](https://tc39.es/ecma262/multipage/indexed-collections.html#sec-array.prototype.findlastindex)

TypedArray methods are excluded because their brand-checking inherently prevents this kind of trickery.

### DIFF
--- a/test/built-ins/Array/prototype/findLast/maximum-index.js
+++ b/test/built-ins/Array/prototype/findLast/maximum-index.js
@@ -1,0 +1,36 @@
+// Copyright (C) 2023 Richard Gibson. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-array.prototype.findlast
+description: >
+  Support the maximum valid integer index.
+info: |
+  Array.prototype.findLast ( predicate [ , thisArg ] )
+
+  1. Let O be ? ToObject(this value).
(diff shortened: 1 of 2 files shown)


test/built-ins/Array/prototype/findLast/maximum-index.js
```diff
@@ -0,0 +1,36 @@
...
+---*/
+
+var tooBigLength = Number.MAX_VALUE;
+var maxExpectedIndex = 9007199254740990;
+var arrayLike = { length: tooBigLength };
+var calledWithIndex = [];
+
+Array.prototype.findLast.call(arrayLike, function(_value, index) {
+  calledWithIndex.push(index);
+  return true;
```
ljharb, line 32: with the `return true`, it would always be only called once, right? or is the test that it's "not zero calls"?

gibson042: > or is the test that it's "not zero calls"?

Correct, and "zero calls" is the current nonconforming behavior of XS.

ljharb (approved):
Verified against my implementations.

Ms2ger (approved):

PR #3777, opened 2023-01-30
Swap call order of NumberFormat's notation and roundingIncrement options

trflynn89:
The order in which these options are read was changed in commit: https://github.com/tc39/proposal-intl-numberformat-v3/commit/29acfc6c3fc9f9183f83ef49e74da747134d63c8


ptomato (approved):
Agreed, thanks!

I can't find the notes of when that pull request was presented to TG2, but since it's incorporated in the stage 4 PR for NumberFormat V3 (https://github.com/tc39/ecma402/pull/753) I assume it's official.

PR #3781, opened 2023-02-07
Temporal: Test copying options object in since() and until()

ptomato:
In since() and until() methods, we copy the options object with CopyDataProperties. Previously, its properties could be read in more than one place (the method itself, and the calendar method), triggering user code each time. Now, we pass around null-prototype objects with only data properties.

See https://github.com/tc39/proposal-temporal/pull/2447


Ms2ger (approved):

ptomato:
Thanks for the review!

PR #3782, opened 2023-02-07
Feature flags cleanup

ptomato:
- Move flags to bottom section that are for proposals already in a published standard
- Remove unused flags
- More guidance for people adding new flags

### DIFF
--- a/features.txt
+++ b/features.txt
@@ -7,234 +7,33 @@
 #
 # https://github.com/tc39/process-document
 
-# Missing checks in Proxy internal methods
-# https://github.com/tc39/ecma262/pull/666
-proxy-missing-checks
-
-# Promise.allSettled
-# https://github.com/tc39/proposal-promise-allSettled
-Promise.allSettled
-
 # Hashbang Grammar
 # https://github.com/tc39/proposal-hashbang
 hashbang
-
-# Object.fromEntries
-# https://github.com/tc39/proposal-object-from-entries
-Object.fromEntries
-
-# BigInt
-# https://github.com/tc39/proposal-bigint
-BigInt
-
-# Class Fields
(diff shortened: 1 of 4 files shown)


features.txt
```diff
@@ -7,235 +7,34 @@
 #
 # https://github.com/tc39/process-document
 
-# Missing checks in Proxy internal methods
-# https://github.com/tc39/ecma262/pull/666
```
ljharb, line 11: can we keep this PR link on the moved item?

features.txt
```diff
@@ -5,6 +5,10 @@
 # dedicated feature flag so that consumers may more easily omit them as
 # necessary.
 #
+# Generally, feature flags should be per proposal, not per individual API
+# that a proposal adds. To filter for specific APIs, use the directory
+# structure.
```
ljharb, line 10: this is unfortunate advice imo because test262harness makes it much easier to exclude feature flags than to target "everything, minus specific directories", but there may be alternative solutions there.

ljharb (commented):
In general I think we should keep the commented URLs near the feature flag, if we can

features.txt
ptomato: Sure — I found there was a bug in the linter that that triggered, but I've added a commit fixing it.

I assume you mean just the URLs to individual pull requests, that aren't easily discoverable given the name of the feature? I think the flags which correspond to whole proposals themselves are already pretty discoverable via tc39/proposals, so in the interest of avoiding clutter I didn't add those everywhere.

features.txt
ptomato: This is nearly the same discussion as on https://github.com/tc39/test262/pull/3760 — I'd be in favour of adding exclude paths to test262-harness.

Ms2ger (approved):

PR #3783, opened 2023-02-08
Push into calls array in calendar and timeZone observers.

12wrigja:
These functions are currently pushing values into `actual` arrays that just happen to be defined in the test and end up in the global scope (e.g. https://github.com/tc39/test262/blob/780b7b806eb6c73122f82ec8dbbe7f74bb7475bf/test/built-ins/Temporal/PlainYearMonth/prototype/add/order-of-operations.js#L77). Other pieces of the observers push their recorded calls into the `calls` array, and in all current cases this array is the same as the previous `actual` array.


Ms2ger (approved):
Oops. Thank you!

Ms2ger (approved):

PR #3785, opened 2023-02-10
Temporal: Remove failing tests for CalendarFields and CalendarMergeFields

Aditi-1400:
This PR removes the tests failing for PR https://github.com/tc39/proposal-temporal/pull/2467 and improves some of the old tests in staging as per the change.
It's better to remove these tests rather than modifying, since [the change](https://github.com/tc39/proposal-temporal/pull/2316/files) for which these tests were added is opposite (at least for the polyfill) to the change we did by PR 2467. 

The commit which adds these tests: https://github.com/tc39/test262/commit/01f73e96ca5ca21a04add782783c0a794224eb3d


ptomato (approved):
Thanks!

Issue #3725, opened 2022-11-17
Tests for Array.fromAsync 


Aditi-1400:
[Proposal readme](https://github.com/tc39/proposal-array-from-async#readme)
[Proposal Spec Text](https://tc39.es/proposal-array-from-async/)

## Testing plan

- ### Standard tests: https://github.com/tc39/test262/pull/3747
    - [x] Property descriptor
    - [x] ~~Brand check~~ - _I think this covered by "`this` value is not a constructor" below_
    - [x] Function length
    - [x] Function name
    - [x] Not a constructor
    - [x] Prototype and extensibility ~~([blocked](https://github.com/tc39/ecma262/pull/2942#issuecomment-1304291563)?)~~

- ### Basic Functionality
    - Success cases - https://github.com/tc39/test262/pull/3809
        - [x] Creates promise
        - [x] Create new array/arrayLike in promise (with length = length property)

- ### Arguments
    - **Input** - https://github.com/tc39/test262/pull/3754
        - *Invalid input values:*
            - [x] `undefined`
            - [x] `null`
            - [x] Symbol
            - [x] Boolean
            - [x] Number + BigInt
            - [x] function
            - [x] @@asyncIterator/@@iterator throws on getting
            - [x] @@asyncIterator/@@iterator returns non-callable value
            - [x] @@iterator returns async iterator (seems like it shouldn't work?)
            - [x] asyncIterator throws on step 
            - [x] nonconforming object (arraylike without length, missing keys) - https://github.com/tc39/test262/pull/3809

        - *Valid inputs:*
            - [x] String
            - [x] Sync-iterable with no promises
            - [x] Sync-iterable with promises
            - [x] Async-iterable
            - [x] Non-iterable with no promises
            - [x] Non-iterable with promises
            - [x] @@asyncIterator returns sync iterator
        - *Tests for input*
            - [ ] Input array with boundary values

            - Covered by polyfill tests: - https://github.com/tc39/test262/pull/3791
                - [x] Output promise rejects if iteration of input fails
                - [x] Output promise rejects if element access fails
                - [x] ~~Result promise rejects if input length exceeds safe integers~~ seems impractical, tests time out
                - [x] Awaits each input value once without mapping callback (with promises)
                - [x] Result promise rejects if element's "then" method fails (with promises)
                - [x] Result promise rejects if thenable element rejects (with promises)
                - [x] Does not await input values (Async iterable)
[...]

PR #3790, opened 2023-02-24
Tests for mapfn and thisArg arguments to Array.fromAsync

ptomato:
- normal case with synchronous and asynchronous mapfn
- a non-callable value is passed as mapfn
- behaviour of various values of thisArg in strict and sloppy mode
- mapfn result is awaited once per iteration
- iterator is closed when mapfn throws

See #3725

PR #3791, opened 2023-03-02
Add some Array.from async tests

Aditi-1400:
This PR adds some more tests for Array.fromAsync proposal.
It includes most of the tests defined in https://github.com/es-shims/array-from-async/blob/main/test.mjs

See:https://github.com/tc39/test262/issues/3725


ptomato:
A couple things to check before further review:
- The tests should be in the folder `test/built-ins/Array/fromAsync/`, currently it's `.../fromasync`
- We can use the `asyncTest()` helper in these tests
- Some of them include `compareArray.js` when not using it (see linter output)
- Some of them seem to be duplicated with tests in #3790

test/built-ins/Array/fromAsync/non-iterable-sync-mapped-callback-err.js
```diff
@@ -0,0 +1,29 @@
+// Copyright (C) 2023 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-array.fromasync
+description: Non iterable input with non-promise thenables works.
```
ptomato, line 6: This description doesn't seem to match what's being tested.

test/built-ins/Array/fromAsync/sync-iterable-with-thenable-async-mapped-callback-await.js
ptomato, line 1: This file might be duplicated by mapfn-result-awaited-once-per-iteration.js. (Although I think that's the only case I tested, the other tests you have for awaiting the result once per iteration are not duplicated.)

I think the name and description are slightly off; it's a sync iterable without thenables.

ptomato (changes requested):
Thanks! Looks good. I had a few minor comments here and there, mostly about test descriptions.

I've updated the checklist in #3725 assuming this PR will get merged.

test/built-ins/Array/fromAsync/non-iterable-sync-mapped-callback-err.js
```diff
@@ -0,0 +1,29 @@
...
+includes: [compareArray.js, asyncHelpers.js]
```
ptomato, line 7: The linter says compareArray isn't used here.

test/built-ins/Array/fromAsync/non-iterable-with-thenable-async-mapped-awaits-once.js
```diff
@@ -0,0 +1,29 @@
+// Copyright (C) 2023 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-array.fromasync
+description: >
+  Non-iterable input with thenables awaits each input once without mapping callback
+includes: [compareArray.js, asyncHelpers.js]
```
ptomato, line 8: The linter says compareArray isn't used here.

ptomato (approved):
Thanks!

PR #3792, opened 2023-03-07
Temporal: Add Temporal era/eraYear tests

Aditi-1400:
Adds tests for Temporal PR: https://github.com/tc39/proposal-temporal/pull/2472

This PR makes PrepareTemporalFields not validate that either both or none of era/eraYear are present. In all cases (except when fields is to be converted to a PlainTime, in which case it doesn't matter), that is already done immediately afterwards anyway by one of the dateFromFields/yearMonthFromFields/monthDayFromFields Calendar methods, which is where we think this validation belongs.

See issue: https://github.com/tc39/proposal-temporal/issues/2465


ptomato (approved):
Thanks!

PR #3796, opened 2023-03-15
Add tests for ArrayBuffer.prototype.transferToFixedLength

syg:
These are copied from ArrayBuffer/prototype/transfer/ with expectations adjusted.

See https://tc39.es/proposal-arraybuffer-transfer/

### DIFF
--- a/test/built-ins/ArrayBuffer/prototype/transferToFixedLength/new-length-excessive.js
+++ b/test/built-ins/ArrayBuffer/prototype/transferToFixedLength/new-length-excessive.js
@@ -3,20 +3,8 @@
 /*---
 esid: sec-arraybuffer.prototype.transfertofixedlength
 description: >
-  Throws a RangeError the newLength value is too large to create a new
-  ArrayBuffer.
-info: |
-  ArrayBuffer.prototype.transferToFixedLength ( [ newLength ] )
-
-  1. Let O be the this value.
-  2. Perform ? RequireInternalSlot(O, [[ArrayBufferData]]).
-  3. If IsSharedArrayBuffer(O) is true, throw a TypeError exception.
-  4. If IsDetachedBuffer(O) is true, throw a TypeError exception.
-  5. If newLength is undefined, let newByteLength be
-     O.[[ArrayBufferByteLength]].
-  6. Else, let newByteLength be ? ToIntegerOrInfinity(newLength).
-  7. Let new be ? Construct(%ArrayBuffer%, « 𝔽(newByteLength) »).
-  [...]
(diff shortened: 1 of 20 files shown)


test/built-ins/ArrayBuffer/prototype/transferToFixedLength/new-length-excessive.js
```diff
@@ -0,0 +1,28 @@
...
+
+  1. Let O be the this value.
+  2. Perform ? RequireInternalSlot(O, [[ArrayBufferData]]).
+  3. If IsSharedArrayBuffer(O) is true, throw a TypeError exception.
+  4. If IsDetachedBuffer(O) is true, throw a TypeError exception.
+  5. If newLength is undefined, let newByteLength be
+     O.[[ArrayBufferByteLength]].
+  6. Else, let newByteLength be ? ToIntegerOrInfinity(newLength).
+  7. Let new be ? Construct(%ArrayBuffer%, « 𝔽(newByteLength) »).
+  [...]
```
ljharb, line 19: these aren't the spec steps for transferToFixedLength: https://tc39.es/proposal-arraybuffer-transfer/ only has two steps, and they call into https://tc39.es/proposal-arraybuffer-transfer/, and the byte length is ToIndex'd which clamps it - so it shouldn't be throwing, i think?

ljharb: (that suggests that the transfer tests also need updating)

ljharb: aha, thanks for clarifying, you're right

ljharb (approved):
confirmed, everything looks good :-D

test/built-ins/ArrayBuffer/prototype/transferToFixedLength/new-length-non-number.js
ptomato, line 1: This test looks like a good case that moving [`toPrimitiveObserver()`](https://github.com/tc39/test262/blob/main/harness/temporalHelpers.js#L1819-L1842) out of TemporalHelpers and into a general helper file would be useful for other proposals. Either way, that's out of scope for this PR, just noting it down for myself for future reference.

ptomato (approved):

test/built-ins/ArrayBuffer/prototype/transferToFixedLength/new-length-excessive.js
syg: I've removed the spec steps from both of those tests and changed the description to talk about clamping in ToIndex.

syg:
Can this be merged?

Issue #3746, opened 2022-12-06
Change Intl Locale Info API test to getter


FrankYFTang:
see https://github.com/tc39/proposal-intl-locale-info/pull/65 discussed during TC39 Nov 2022 meeting.

Impacted tests

We need to change the following test262 tests with this PR
```
'intl402/Locale/prototype/calendars/branding': [FAIL],
'intl402/Locale/prototype/calendars/name': [FAIL],
'intl402/Locale/prototype/calendars/output-array': [FAIL],
'intl402/Locale/prototype/calendars/prop-desc': [FAIL],
'intl402/Locale/prototype/collations/branding': [FAIL],
'intl402/Locale/prototype/collations/name': [FAIL],
'intl402/Locale/prototype/collations/output-array': [FAIL],
'intl402/Locale/prototype/collations/output-array-values': [FAIL],
'intl402/Locale/prototype/collations/prop-desc': [FAIL],
'intl402/Locale/prototype/hourCycles/branding': [FAIL],
'intl402/Locale/prototype/hourCycles/name': [FAIL],
'intl402/Locale/prototype/hourCycles/output-array': [FAIL],
'intl402/Locale/prototype/hourCycles/output-array-values': [FAIL],
'intl402/Locale/prototype/hourCycles/prop-desc': [FAIL],
'intl402/Locale/prototype/numberingSystems/branding': [FAIL],
'intl402/Locale/prototype/numberingSystems/name': [FAIL],
'intl402/Locale/prototype/numberingSystems/output-array': [FAIL],
'intl402/Locale/prototype/numberingSystems/prop-desc': [FAIL],
'intl402/Locale/prototype/textInfo/branding': [FAIL],
'intl402/Locale/prototype/textInfo/name': [FAIL],
'intl402/Locale/prototype/textInfo/output-object': [FAIL],
'intl402/Locale/prototype/textInfo/output-object-keys': [FAIL],
'intl402/Locale/prototype/textInfo/prop-desc': [FAIL],
'intl402/Locale/prototype/timeZones/branding': [FAIL],
'intl402/Locale/prototype/timeZones/name': [FAIL],
'intl402/Locale/prototype/timeZones/output-array': [FAIL],
'intl402/Locale/prototype/timeZones/output-array-sorted': [FAIL],
'intl402/Locale/prototype/timeZones/output-undefined': [FAIL],
'intl402/Locale/prototype/timeZones/prop-desc': [FAIL],
[...]

PR #3797, opened 2023-03-21
Sync Intl Locale API tests w/ PR67

FrankYFTang:
change the tests to sync with 

https://github.com/tc39/proposal-intl-locale-info/pull/67


FrankYFTang:
@ryzokuken @sffc @romulocintra

gibson042 (approved):

ptomato:
@FrankYFTang I believe this is ready, and the normative change has consensus and has been merged. It looks like some `output-array-values.js` tests have been added in the meantime to some of the folders that this PR removes, though. Could you look at whether those need to be moved those to the new folders?

CanadaHonk:
This fixes #3746.

FrankYFTang:
rebased PTAL

Ms2ger (approved):

PR #3799, opened 2023-03-22
Add `Array.prototype.includes` and `exponentiation` features

jedel1043:
While trying to use the `features` metadata to classify tests per ECMAScript edition, I realized there weren't features for the `Array.prototype.includes` function, nor the `**` operator. Seeing that even ES6 features  like `class`, `let` or `const` have their own features, it'd be more consistent to also have the same labeling for ES7 features.   

This PR introduces the features `Array.prototype.includes` and `exponentiation`, and adds them to the corresponding tests.


ljharb (approved):
seems good to me, thanks

PR #3800, opened 2023-03-23
Update decorator member expr private id cases

### LINKED ISSUE #3794: Invalid test for decorators
https://github.com/tc39/test262/blob/main/test/language/expressions/class/decorator/syntax/class-valid/decorator-member-expr-private-identifier.js

I'm quite sure that this test is invalid, and that `@#x` is not a valid decorator (cc @pzuraq). I noticed it because Babel's implementation is throwing a syntax error for that test.

JLHwung:
Per https://github.com/tc39/ecma262/pull/2417/files#diff-181371b08d71216599b0acccbaabd03c306da6de142ea6275c2135810999805aR24159-R24162, updated the production of DecoratorMemberExpression and test cases.

Fixes #3794

### DIFF
--- a/test/language/expressions/class/decorator/syntax/class-valid/decorator-member-expr-private-identifier.js
+++ b/test/language/expressions/class/decorator/syntax/class-valid/decorator-member-expr-private-identifier.js
@@ -22,7 +22,9 @@ info: |
 
 
     DecoratorMemberExpression[Yield, Await] :
-      PrivateIdentifier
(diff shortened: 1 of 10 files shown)


nicolo-ribaudo:
Fixes https://github.com/tc39/test262/issues/3794

ptomato (approved):
Thanks, looks like a correct representation of the current state of the proposal.

PR #3802, opened 2023-03-24
Round epoch time getters towards negative infinity

anba:
Update tests to match <https://github.com/tc39/proposal-temporal/pull/2424>.


ptomato (approved):
Thanks, you are correct about this. I had looked at these tests before and mistakenly thought that they already tested for the new behaviour; with your changes they're correct.

PR #3804, opened 2023-03-27
Update expected test results for CLDR 43

anba:
intl402/Locale/constructor-non-iana-canon.js
- Likely subtags entry for "hyw" -> "hyw-Armn-AM" was added to supplemental/likelySubtags.xml

intl402/Locale/prototype/minimize/removing-likely-subtags-first-adds-likely-subtags.js
- Likely subtags entry for "aae" -> "aae-Latn-IT" was added to supplemental/likelySubtags.xml
- Likely subtags entry for "pap" was changed from "pap_Latn_AW" to "pap_Latn_CW"

### DIFF
--- a/test/intl402/Locale/constructor-non-iana-canon.js
+++ b/test/intl402/Locale/constructor-non-iana-canon.js
@@ -86,6 +86,7 @@ var testData = [
     {
         tag: "hy-arevmda",
         canonical: "hyw",
+        maximized: "hyw-Armn-AM",
     },
 ];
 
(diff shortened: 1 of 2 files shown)


ptomato (approved):
Thanks @anba, I don't have much expertise in what's changing here but I basically trust you to make accurate CLDR updates, so +1 from me. Maybe @gibson042 could take a look before we merge this.

anba:
We don't need to merge right away, because this will only start to trigger failures when implementations update to ICU 73. The ICU release is scheduled for [mid April](https://icu.unicode.org/download/73). (I hit these failures when testing with the ICU 73 release candidate.)

ptomato:
The CLDR release has happened, so let's merge this.

Issue #3756, opened 2022-12-20
Tests for regular expression modifiers


cjtenny:
Explainer: https://github.com/tc39/proposal-regexp-modifiers
Spec text: https://tc39.es/proposal-regexp-modifiers/

# regexp-modifiers testing plan

* Syntax errors: Throw both in parsing and when constructed with new Regexp("...")
  * Basic regular expression flags (n.b. source text refers to matched text for "regular expression flags" production of grammar)
    * [x] Source text contains other code points than `i`, `m`, `s`
    * [x] Source text contains combining codepoints alongside `i`, `m`, `s`
    * [x] Source text contains other non-display codepoints alongside `i`, `m`, `s`
    * [x] Source text contains ZWNJ, ZWJ, ZWNBSP alongside `i`, `m`, `s` (I think this is right? https://tc39.es/ecma262/#sec-unicode-format-control-characters)
    * [x] Source text contains `i`, `m`, and/or `s` more than once
    * [x] Source text in a case-ignoring context contains code points that case fold to `i`, `m`, `s` e.g. `I`, `M`, `S`
    * [x] Source text contains code points outside the basic latin range that, were they canonicalized by a unicode-mapping regex, would map to e.g. `i`, `m`, or `s` (e.g. ſ (U+017F) would map to `s`, U+0130 to `i`) (ref. https://www.unicode.org/Public/12.1.0/ucd/CaseFolding.txt)
      * [x] (e.g. `/foo(?\u{017F}:bar)/u` is a syntax error, `/foo(?s:bar)/u` is not)
  * Arithmetic regular expression flags
    * [x] First or second source text exhibits any of the 'basic regular expression flags' errors
    * [x] Both source texts are empty
    * [x] Code point matched by first flags is also contained in source text matched by second flags
    * [x] Various forms of (?ims-ims) - no colon - is a syntax error
    * [x] Source text cannot use unicode escape sequences to express code points `i`, `m`, `s`
* Valid syntax
  * [x] Basic regular expression flags parse correctly
  * [x] Source text with any valid combination of flags or arithmetic flags - reasonable to enumerate
* Behavior
  * [x] Disabling flag in subexpression behaves correctly when corresponding top-level flag is and isn't already set
  * [x] Enabling flag in subexpression behaves correctly when corresponding top-level flag is and isn't already set
  * [x] Constructing a RegExp from a literal but changing flags by an argument to the RegExp constructor does (or does not) correctly change behavior of a subexpression that enables or removes flags.
[...]

PR #3807, opened 2023-03-31
 proposal-regexp-modifiers: Tests for syntax errors

guijemont:
This covers the syntax error cases from https://github.com/tc39/test262/issues/3756.
This is an updated version of https://github.com/tc39/test262/pull/3758 addressing review comments.


guijemont:
The main issue why this is still a work in progress, is that the templated test generator has some issues.
In particular, we want to use it with templates such as:
```javascript
/(?/*{ subpattern-add-modifiers }*/-/*{ subpattern-remove-modifiers }*/:a)//*{ global-modifiers }*/;
```

But the generator cannot replace `/*{ global-modifiers }*/` as it seems to consider that `//*{ global-modifiers }*/;` is a comment. In otherwords, the generator needs to learn about how to handle comments that immediately follow a regex literal.

guijemont:
> The main issue why this is still a draft, is that the templated test generator has some issues. In particular, we want to use it with templates such as:
> 
> ```js
> /(?/*{ subpattern-add-modifiers }*/-/*{ subpattern-remove-modifiers }*/:a)//*{ global-modifiers }*/;
> ```
> 
> But the generator cannot replace `/*{ global-modifiers }*/` as it seems to consider that `//*{ global-modifiers }*/;` is a comment. In otherwords, the generator needs to learn about how to handle comments that immediately follow a regex literal.

I created issue https://github.com/tc39/test262/issues/3808 to report the problem.

rbuckton (commented):
This looks good as far as invalid syntax coverage is concerned, at least once #3808 is resolved.

ptomato:
Closing in favour of #4016.

PR #3809, opened 2023-03-31
Array.fromAsync various remaining coverage

ptomato:
This contains a few more tests for Array.fromAsync, in addition to what has already been merged and what is under review at #3791.

This covers the following items from the testing plan at #3725:

- Success cases
  - Creates promise
  - Create new array/arraylike in promise (with length = length property)
- Input
  - Invalid input values - nonconforming object (arraylike without length, missing keys)
  - Covered by polyfill tests
    - Result promise rejects if length access fails (non-iterable input)
    - Unaffected by globalThis.Symbol mutation (non-iterable)
- this-value
  - this-value is a constructor
  - this-value is not a constructor
  - If this is a constructor, and items doesn't have a Symbol.iterator, returns a new instance of this
  - Iterator closed when property creation on this fails
  - Returned promise rejects when ^
- Other tests
  - Error is thrown for all CreateDataProperty fails
  - Non-writable properties are overwritten by CreateDataProperty
  - Input with missing values

### DIFF
--- a/test/built-ins/Array/fromAsync/this-constructor-operations.js
+++ b/test/built-ins/Array/fromAsync/this-constructor-operations.js
@@ -1,0 +1,74 @@
+// Copyright (C) 2023 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-array.fromasync
+description: >
+  Order of user-observable operations on a custom this-value and its instances
+includes: [compareArray.js, asyncHelpers.js]
+flags: [async]
+features: [Array.fromAsync]
+---*/
+
+function formatPropertyName(propertyKey, objectName = "") {
+  switch (typeof propertyKey) {
+    case "symbol":
+      if (Symbol.keyFor(propertyKey) !== undefined) {
+        return `${objectName}[Symbol.for('${Symbol.keyFor(propertyKey)}')]`;
+      } else if (propertyKey.description.startsWith('Symbol.')) {
+        return `${objectName}[${propertyKey.description}]`;
+      } else {
+        return `${objectName}[Symbol('${propertyKey.description}')]`
+      }
+    case "string":
+      if (propertyKey !== String(Number(propertyKey)))
+        return objectName ? `${objectName}.${propertyKey}` : propertyKey;
+      // fall through
+    default:
+      // integer or string integer-index
+      return `${objectName}[${propertyKey}]`;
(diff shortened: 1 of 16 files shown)


test/built-ins/Array/fromAsync/this-constructor-with-bad-length-setter.js
```diff
@@ -0,0 +1,34 @@
...
+  3.k.viii. Perform ? Set(_A_, *"length"*, 𝔽(_len_), *true*)
+includes: [asyncHelpers.js]
+flags: [async]
+features: [Array.fromAsync]
+---*/
+
+asyncTest(async function () {
+  class MyArray {
+    set length(v) {
+      assert.sameValue(v, 3, "setter should be called with length 3");
```
Ms2ger, line 21: Hm, if this fails, won't it also throw a `Test262Error`? It seems potentially confusing to use it for both assertion failures and testing control flow.

test/built-ins/Array/fromAsync/this-constructor-with-bad-length-setter.js
```diff
@@ -0,0 +1,34 @@
...
+      throw new Test262Error("setter of length property throws")
```
Ms2ger, line 22: Do we have a test that checks that `"length"` is only set once?

Ms2ger (approved):

test/built-ins/Array/fromAsync/this-constructor-with-bad-length-setter.js
ptomato: Good point. There are other tests that already confirm the value with which the length property is set, so I'll just remove this line.

test/built-ins/Array/fromAsync/this-constructor-operations.js
```diff
@@ -0,0 +1,74 @@
...
+      }
+    });
+  }
+
+  let result = await Array.fromAsync.call(MyArray, [1, 2]);
+  assert.compareArray(expectedCalls, actualCalls, "order of operations for array argument");
+
+  actualCalls.splice(0);  // reset
+
+  const expectedCallsForArrayLike = [
```
Ms2ger, line 61:
```suggestion
  // Note https://github.com/tc39/proposal-array-from-async/issues/35
  const expectedCallsForArrayLike = [
```

Ms2ger (approved):

PR #3813, opened 2023-04-07
Temporal: Remove support for nested property bags from TimeZone/Calendar protocol, and change validation

ptomato:
Normative PR: tc39/proposal-temporal#2485 which reached consensus in the January/February 2023 TC39 meeting.

The required changes to tests involve a lot of search-and-replace so I'd recommend reviewing this PR commit by commit and probably skimming over the boring parts. The good news is that the tests become much simpler overall, reflecting a simplification of the TimeZone and Calendar protocols in Temporal.


test/intl402/Temporal/ZonedDateTime/compare/argument-propertybag-timezone-string-datetime.js
```diff
@@ -0,0 +1,28 @@
+// Copyright (C) 2022 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.zoneddatetime.compare
+description: Conversion of ISO date-time strings to Temporal.TimeZone instances (with Intl time zones)
```
justingrant, line 6: Minor nit: Intl is not required to use IANA time zones, so maybe this title (and others like it, if any) should say "IANA" instead of "Intl" ?

test/built-ins/Temporal/PlainYearMonth/compare/argument-propertybag-calendar-leap-second.js
```diff
@@ -0,0 +1,24 @@
+// Copyright (C) 2022 Igalia, S.L. All rights reserved.
```
justingrant, line 1: I saw a few brand-new tests in this PR that seem unrelated to the nested property bags. That seems fine to me, just wanted to confirm that this was intended.

test/built-ins/Temporal/Instant/prototype/toZonedDateTimeISO/timezone-wrong-type.js
```diff
@@ -27,6 +26,8 @@ for (const [timeZone, description] of rangeErrorTests) {
 
 const typeErrorTests = [
   [Symbol(), "symbol"],
+  [{}, "object not implementing time zone protocol"],
```
justingrant, line 29: Should there be a `null` case here too? I'm not adding the same comment to all the other tests like this, if that's OK.

test/built-ins/Temporal/PlainDate/prototype/toString/calendarname-undefined.js
```diff
@@ -14,12 +14,35 @@ info: |
...
 const tests = [
   [[], "2000-05-02", "built-in ISO"],
-  [[{ id: "custom" }], "2000-05-02[u-ca=custom]", "custom"],
-  [[{ id: "iso8601" }], "2000-05-02", "custom with iso8601 id"],
-  [[{ id: "ISO8601" }], "2000-05-02[u-ca=ISO8601]", "custom with caps id"],
-  [[{ id: "\u0131so8601" }], "2000-05-02[u-ca=\u0131so8601]", "custom with dotless i id"],
+  [[{ id: "custom", ...calendarMethods }], "2000-05-02[u-ca=custom]", "custom"],
+  [[{ id: "iso8601", ...calendarMethods }], "2000-05-02", "custom with iso8601 id"],
+  [[{ id: "ISO8601", ...calendarMethods }], "2000-05-02[u-ca=ISO8601]", "custom with caps id"],
+  [[{ id: "\u0131so8601", ...calendarMethods }], "2000-05-02[u-ca=\u0131so8601]", "custom with dotless i id"],
```
justingrant, line 45: We allow non-ASCII calendar IDs? @sffc is this OK?

justingrant (approved):
Wow, amazing to get this done. Only a few minor nits & questions.

test/built-ins/Temporal/Instant/prototype/toZonedDateTimeISO/timezone-wrong-type.js
ptomato: `null` is tested above in the list of RangeError cases - it falls through to the non-Object case in [ToTemporalTimeZoneSlotValue](file:///var/home/ptomato/workspace/proposal-temporal/out/index.html#sec-temporal-totemporaltimezoneslotvalue). (The Type() operation in the spec, thankfully, isn't the same as `typeof` in JS code — Type(**null**) is Null, not Object.)

justingrant: Whoops, sorry for not looking up. :-)

test/intl402/Temporal/ZonedDateTime/compare/argument-propertybag-timezone-string-datetime.js
ptomato: I'll tack on another commit that replaces "Intl time zone" with "IANA time zone" everywhere.

test/built-ins/Temporal/PlainYearMonth/compare/argument-propertybag-calendar-leap-second.js
ptomato: Yeah, the story behind this one is that it should've been here but for some reason was missing from PlainYearMonth.compare. (Probably an oversight on my part.) It would've previously had a nested property bag in it, which I would've removed in this PR.

test/built-ins/Temporal/PlainDate/prototype/toString/calendarname-undefined.js
ptomato: I don't think we allow them as such - but we can't really stop someone from doing something like
```js
get id() {
  return "🍕";
}
```
I believe we previously chose not to prevent `toString` from being run on a date with such a calendar. I'll try to look for the issue or meeting note where this was discussed. (My feeling is that if you're going to use a malformed calendar ID string, you'll get a malformed ISO string, but it's probably useful to still attempt to serialize a Temporal object to a string if we can at all do so.)

Ms2ger (approved):
I trust Philip and Justin here.

test/built-ins/Temporal/PlainDate/prototype/toString/calendarname-undefined.js
ptomato: See https://github.com/tc39/proposal-temporal/issues/1815
I didn't link a discussion when I closed it, but in the meeting notes I see that we did a bugtracker sweep in the champions meeting just before I closed it, so I assume that's why.

ptomato:
Thanks for the reviews!

Issue #3808, opened 2023-03-31
[generation] Support comments directly following regular expression litterals


guijemont:
The generation tool does not properly substitute in cases such as:
```javascript
/foo//*{ bar }*/
```
Though something like this works:
```javascript
/foo/ /*{ bar }*/
```

It seems that [`find_comments()`](https://github.com/tc39/test262/blob/ff405fd0df81ff52c3bcb5b2be3e872983fb0402/tools/generation/lib/util/find_comments.py#L4) needs to learn about regexp literals and handle them properly.

See https://github.com/tc39/test262/pull/3807#issuecomment-1492571532

guijemont:
FWIW [this branch](https://github.com/guijemont/test262/tree/issue-3808) contains my (broken) attempt at fixing this, unfortunately I don't have more time to spend on this right now.

gibson042:
I'm fine with fixing `/foo//*{ bar }*/`, but don't want to support syntactically invalid constructs like `/(?/*{ foo }*/-/*{ bar }*/:a)/`. I'd rather introduce a new style of procedural test generation with corresponding frontmatter key, even if that style remains inverted (in which templates are referenced by cases) like e.g.
```js
/*---
esid: …
…
placeholder-prefix: ' %%'
placeholder-suffix: '%% '
---*/

/(? %%subpattern-add-modifiers%% - %%subpattern-remove-modifiers%% :a)/ %%global-modifiers%% ;
```
for accepting existing .case files like
```js
//- subpattern-add-modifiers
i
//- subpattern-remove-modifiers
//- global-modifiers
gu
```
to produce content like `/(?i-:a)/gu;`.

ptomato:
I'm not sure I understand the difference - isn't `/(? %%foo%% - %%bar%% :a)/` just as syntactically invalid as `/(?/*{ foo }*/-/*{ bar }*/:a)/`?

ptomato:
After poking at this bug, I think I understand better why you are asking for this. It's simple enough to support `/foo//*{ bar }*/` but without extra work to support placeholder comments inside RegExp literals, we actually _lose_ the existing support for `/(?/*{ foo }*/-/*{ bar }*/:a)/`.

[...]

PR #3817, opened 2023-04-14
Fix test generation placeholders inside and adjacent to RegExp literals

ptomato:
Several improvements to the test generation tool. The first commit is based on @guijemont's work and fixes the case where a placeholder comment is immediately adjacent to the end of a RegExp literal.

The second commit adds a feature to test template files which can be used as an escape hatch in case the `/*` and `*/` delimiters of placeholder comments would be misparsed. (This is one way that it could work, I'm not attached to it.)

Closes: #3808


jugglinmike (commented):
The first commit looks like it unblocks gh-3807, so I'm all for it!

I'm less enthusiastic about the second commit, though. Addressing needs through configuration certainly makes the tool more powerful, but it also makes things more daunting for new contributors--even for those who have no need for it (presumably the majority of potential users). This implementation's distinct edge case (the restriction on interpolated values), if documented, will steepen the learning curve a little further still.

I'd be more inclined to collaborate on the design if there was a motivating use case. Without that, its hard for me to justify the complexity and to envision the scenarios where test authors would need this ability.

ptomato:
@jugglinmike I see your point, in fact I shared your skepticism about the configurable delimiters feature when Richard suggested it. [This comment](https://github.com/tc39/test262/issues/3808#issuecomment-1509221736) explains why I changed my mind. But like I said in the description, I'm not attached to the design of this feature, so I'd be happy to go with something completely different.

But more immediately: the first commit actually doesn't unblock #3807 unless I add some more processing for comments inside regexp literals. (The first commit fixes `/foo//*{ bar }*/` but breaks `/foo/*{ bar }*/baz/`. Fixing the latter made me come around to Richard's suggestion of configurable delimiters.) I'd be happy to make a separate PR with [83614a1](https://github.com/tc39/test262/pull/3817/commits/83614a1eabc6af65841299bdb92c5811606811fb) and a second commit that fixes comments inside regexp literals, while we iterate on the configurable delimiters, or decide not to do them.

PR #3818, opened 2023-04-18
Temporal: Add test for non-iso calendars for Temporal.PlainMonthDay.from

Aditi-1400:
Adds test for normative PR : https://github.com/tc39/proposal-temporal/pull/2460
This pull request was approved in March Plenary.


ptomato (approved):
Thanks!

PR #3823, opened 2023-05-03
Temporal: Tests for multiple calendar annotations with at least one critical

ptomato:
As per IETF review, an IXDTF string (ISO 8601 with annotations) is no longer valid if it contains more than one u-ca annotation and at least one of the annotations is marked critical.

Removes tests where such a string was assumed to be valid, and adds new ones with a few variations on invalid strings.

Corresponds to a normative change: tc39/proposal-temporal#2572.


ptomato:
The corresponding change to the proposal received consensus today at the May 2023 TC39 meeting. Review of the tests appreciated!

Ms2ger (approved):

ptomato:
Thanks for the reviews.

PR #3831, opened 2023-05-19
Test for Unicode Escape in Literals

veera-sivarajan:
This PR adds tests to check if literals (`true`, `false`, and `null`) containing Unicode escape sequence throws a syntax error as mentioned in NOTE 1 in the [spec](https://tc39.es/ecma262/#sec-keywords-and-reserved-words).


ptomato (approved):
Thanks for contributing this! Sorry the review took such a long time. This seems correct!

PR #3832, opened 2023-05-22
Fix phase on some negative module tests

jedel1043:
A couple of negative tests had an incorrect phase tagged, since their errors happen at linking time, which is part of the "resolution" phase. However, they were marked as happening at runtime.

This PR fixes those tests and additionally adds a `$DONOTEVALUATE` marker to them.

### DIFF
--- a/test/language/module-code/instn-iee-err-ambiguous-as.js
+++ b/test/language/module-code/instn-iee-err-ambiguous-as.js
@@ -31,9 +31,11 @@ info: |
                   SameValue(resolution.[[BindingName]],
                   starResolution.[[BindingName]]) is false, return "ambiguous".
 negative:
-  phase: runtime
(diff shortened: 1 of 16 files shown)


jugglinmike (approved):
Thanks for the fix!

Out of curiosity: do you have an environment that was failing the erroneous version of these tests? Or did you spot the mistake through careful reading?

jedel1043:
> Or did you spot the mistake through careful reading?
I wish 😅

> Out of curiosity: do you have an environment that was failing the erroneous version of these tests?

I'm part of the team developing Boa, an experimental JS engine written in Rust. We found out about the bugs when we were finishing our implementation of the ES modules spec
(https://github.com/boa-dev/boa/pull/2955)

PR #3837, opened 2023-05-29
Tests for proposal-canonical-tz Stage 3 (will be merged into Temporal)

justingrant:
This PR includes tests for the Time Zone Canonicalization proposal ([proposal-canonical-tz](https://github.com/tc39/proposal-canonical-tz)) that reached Stage 3 at the July 2023 TC39 meeting. 

At the same meeting, it was decided that this proposal should be merged into Temporal, so these tests are now part of tc39/proposal-temporal#2633.

### Test roadmap

- [x] DONE Add lightweight "Demitasse" tests in https://github.com/tc39/proposal-canonical-tz/blob/main/polyfill/test/canonicaltz.mjs that cover this proposal's full surface area.
- [x] DONE Fix 15 existing Test262 tests that were broken by this proposal, because they assumed that time zone identifiers are always canonicalized.
- [x] DONE: Migrate `Temporal.TimeZone.p.equals` (the only new API in this proposal) Demitasse tests to Test262.
- [x] DONE: Migrate `Temporal.TimeZone` Demitasse tests to Test262.
- [x] DONE: Migrate `Temporal.ZonedDateTime` Demitasse tests to Test262.
- [x] DONE: Migrate `Intl.DateTimeFormat` Demitasse tests to Test262.


test/intl402/DateTimeFormat/timezone-canonicalized.js
```diff
@@ -1,35 +0,0 @@
...
-author: Norbert Lindenberg
----*/
-
-// canonicalization specified in conformance clause
-var additionalTimeZoneNames = {
-    "Etc/GMT": "UTC",
-    "Greenwich": "UTC",
-    "PRC": "Asia/Shanghai",
-    "AmErIcA/LoS_aNgElEs": "America/Los_Angeles",
-    "etc/gmt+7": "Etc/GMT+7"
```
ptomato, line 18: Aren't these still valid? Or are they rolled into a different test now?

test/intl402/Temporal/TimeZone/from/timezone-case-insensitive.js
```diff
@@ -3,10 +3,615 @@
...
+  'America/Bogota',
+  'America/Boise',
+  'America/Cambridge_Bay',
+  'America/Campo_Grande',
+  'America/Cancun',
+  'America/Caracas',
+  'America/Cayenne',
+  'America/Chicago',
+  'America/Chihuahua',
+  // 'America/Ciudad_Juarez',
```
ptomato, line 61: What is special about this one?

test/intl402/Temporal/TimeZone/from/timezone-case-insensitive.js
```diff
@@ -3,10 +3,615 @@
...
-description: Time zone names are case insensitive
-features: [Temporal]
+description: Time zone identifiers are case-normalized
+features: [Temporal, time-zone-canonicalization]
 ---*/
 
-const timeZone = 'eTc/gMt+1';
-const result = Temporal.TimeZone.from(timeZone);
-assert.sameValue(result.id, 'Etc/GMT+1', `Time zone created from string "${timeZone}"`);
+const timeZoneIdentifiers = [
```
ptomato, line 10: This list should probably be alphabetical so you know where to look if you need to check whether a particular one is included.

test/intl402/Temporal/TimeZone/from/timezone-case-insensitive.js
```diff
@@ -3,10 +3,615 @@
...
+  'US/Pacific',
+  'US/Pacific-New',
+  'US/Samoa',
+  'UTC',
+  'Universal',
+  'W-SU',
+  'Zulu'
+];
+
+const ids = [...new Set([...timeZoneIdentifiers, ...Intl.supportedValuesOf('timeZone')])];
```
ptomato, line 610: Not sure I understand the purpose of this — it's a bit puzzling to me why we need to use both the hardcoded list of identifiers and the `supportedValuesOf` list. Maybe an explanatory comment would be helpful?

test/intl402/Temporal/TimeZone/prototype/equals/timezone-case-insensitive.js
```diff
@@ -0,0 +1,618 @@
+// Copyright (C) 2022 Igalia, S.L. All rights reserved.
```
ptomato, line 1: Same remarks for this file as with the other similar one above.

test/intl402/Temporal/ZonedDateTime/from/timezone-case-insensitive.js
```diff
@@ -0,0 +1,619 @@
+// Copyright (C) 2022 Igalia, S.L. All rights reserved.
```
ptomato, line 1: Ditto.

test/built-ins/Temporal/TimeZone/prototype/equals/argument-object.js
```diff
@@ -0,0 +1,56 @@
+// Copyright (C) 2020 Igalia, S.L. All rights reserved.
```
ptomato, line 1: The copyright should say 2023 and belong to you, since you contributed them (or, in the general case, to your employer if your contract with them requires it) — even if they are similar to other tests elsewhere in the suite.

ptomato (approved):
Thanks! This matches my understanding of the proposal.

Looks like there's a merge conflict in `features.txt`, so it needs a rebase.

test/built-ins/Temporal/TimeZone/prototype/equals/argument-object.js
justingrant: OK, will do. Didn't even notice the copyright, lol. Too busy copy/pasting boilerplate.

test/intl402/Temporal/TimeZone/from/timezone-case-insensitive.js
justingrant: OK will add one. `supportedValuesOf` was included to handle the case of new IDs being added before these tests are updated to include those new IDs. The reason why `supportedValuesOf` is insufficient is because it only returns canonical IDs. If you also want to test non-canonical IDs ("non-primary time zone identifiers") in official spec terminology) then you need to hard-code a list in the test.

test/intl402/Temporal/TimeZone/from/timezone-case-insensitive.js
justingrant: OK, will alphabetize.

justingrant (commented):
Thanks for the review!  I'll make the changes you recommend. Will try to get these changed before I present in 15 hours. :-)

test/intl402/Temporal/TimeZone/from/timezone-case-insensitive.js
justingrant: `America/Ciudad_Juarez` was just added to TZDB this year, and CLDR hasn't caught up yet. So this test would never pass. We can uncomment it if a later Node version includes this ID.  I'll add a comment to explain.

ptomato:
> Thanks for the review! I'll make the changes you recommend. Will try to get these changed before I present in 15 hours. :-)

I don't think it's necessary to rush! Having test262 tests merged is a requirement for stage 4, but not 3. Anyway the comments are just minor :smile:

test/intl402/DateTimeFormat/timezone-canonicalized.js
justingrant: This API no longer canonicalizes identifiers, so the first three tests don't apply anymore. The latter two are covered by the lowercase/uppercase test. Well, I guess "AmErIcA/LoS_aNgElEs" isn't if you're splitting hairs, but I didn't really think that a combo of upper and lower case was important to test if we're already testing all lower and all upper case for each ID. Holler and un-resolve if you disagree!

test/intl402/Temporal/TimeZone/from/timezone-case-insensitive.js
justingrant: I realized that the list is already sorted, but it's separated into two parts: Zone names from TZDB and Link names from TZDB. That seems like a useful separation in case tests fail for only Links or only Zones. So I decided to leave the two groups as-is, but I added comments in front of each section to clarify what are Zones and what are Links.

test/intl402/Temporal/ZonedDateTime/from/timezone-case-insensitive.js
justingrant: I added my own copyright line for all new files I created here in this PR. There were a lot more than just the few you flagged.

justingrant:
All review feedback is resolved in the latest fixup commits.

I'll autosquash them if this proposal achieves Stage 3 later tonight (my time), but figured that it'd be helpful to keep them separate for now so they're easier to review.

justingrant:
@ptomato - let me know if you want to review the fixup commits that I made in response to your feedback. Otherwise I'll just rebase and autosquash them. Thanks!

justingrant:
Marked as ready for review because this proposal reached Stage 3. Also made a few updates:
* Autosquashed older changes from reviews
* Reverted changes to features.txt because this proposal is being merged into Temporal
* Added cases to ensure handling of case-sensitive comparisons of custom time zone IDs
* Updated OP with link to Temporal PR (tc39/proposal-temporal#2633) that will include these tests.

PR #3839, opened 2023-06-06
Don't require Inflect for linting files

anba:
Allow to lint files even when the "inflect" module isn't available.

### DIFF
--- a/tools/lint/lint.py
+++ b/tools/lint/lint.py
@@ -5,7 +5,6 @@
 from __future__ import print_function
 
 import argparse
-import inflect
 import os
 try:
     from pip._internal import main as pip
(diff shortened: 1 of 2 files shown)


anba:
PEP 668 made `pip install` a bit more complicated [1], so it's seems nicer to make the "inflect" use optional.

[1] <https://www.omgubuntu.co.uk/2023/04/pip-install-error-externally-managed-environment-fix>

ptomato (approved):
Thanks for catching this. Seems fine. Although, if installing this dependency is problematic, I would also be fine with just dropping it. It's used only for formatting 2 messages, e.g. "Linting 1 file"/"Linting 2 files", and it'd be fine to rephrase that as "Linting 2 file(s)" or "Files to lint: 2" or something like that.

anba:
> Although, if installing this dependency is problematic, I would also be fine with just dropping it. 

Okay. I've added a separate commit (08705a11348cd69b3b288718e68665f4e3e70886) to remove the dependency.

jugglinmike:
Thanks, @anba! I'd welcome this simplification even if it wasn't motivated by an issue with the contribution experience, but this is a bit of a Chesterton's fence situation. @rwaldron [this commit doesn't explain the rationale for the Python module](https://github.com/tc39/test262/commit/50d840aceab7afffc7dcbf9732e8d65a7cadb459); can you recall any reason we shouldn't remove it?

anba:
Ping?

Issue #3829, opened 2023-05-18
Tests for import attributes


nicolo-ribaudo:
Repo: https://github.com/tc39/proposal-import-attributes/
Spec text: https://tc39.es/proposal-import-attributes/

Import assertions have recently been changed to be "import attributes", with the following changes:
1. the keyword is now `with` instead of `assert`
2. the property in the dynamic import options bag is now `with`
3. there are fewer restrictions about what a host can do with the given attributes

Additionally, the old syntax is still specified but marked as "normative optional, [deprecated](https://tc39.es/proposal-import-attributes/#sec-conformance)", and there is an optional fallback to lookup the `assert` property in the dynamic import options bag if `with` is undefined.

I don't think we can test (3), but for (1) and (2) we can copy the existing tests replacing `assert` with `with`. Additionally, we may need a test for dynamic import with both `with` and `assert` defined in the options bag.

JonnyTizz:
Hi! I was just browsing issues and happened across this one.

It looks like the related PR was split, and the split parts were all merged successfully. Does this mean this issue is resolved and can be closed?

Thanks!

PR #3843, opened 2023-06-07
Add tests for import attributes

nicolo-ribaudo:
https://github.com/tc39/proposal-import-attributes
The proposal supersedes import assertions, and is at stage 3 conditional on ecma-262 editors reviewing its spec.

Closes https://github.com/tc39/test262/issues/3829

TODO (Maybe in a separate PR, to keep this one only about renaming and copying tests):
- [ ] Test using both variants at the same time, both in dynamic import and in static import


src/dynamic-import/import-attributes-trailing-comma-secon.case
linusg, line 1: `secon` -> `second`

test/language/module-code/import-attributes/import-assertion-trlng-comma.js
```diff
@@ -0,0 +1,31 @@
...
+    AssertionKey : StringLiteral , AssertEntries
+
+  AssertionKey:
+    IdentifierName
+    StringLiteral
+features: [import-attributes, globalThis]
+flags: [module]
+---*/
+
+import x from './import-assertion-1_FIXTURE.js' with {test262:'',};
```
nicolo-ribaudo, line 26: With the new proposal, unknown import attributes should throw rather than being ignored. Should we require runner to define a no-op `test262` attribute that we can use for testing purposes?

test/language/module-code/import-attributes/eval-gtbndng-indirect-faux-assertion.js
```diff
@@ -0,0 +1,70 @@
...
+Object.defineProperty(globalThis, 'assert', {
+  get: function() {
+    callCount += 1;
+  }
+});
+
+import x from './import-assertion-1_FIXTURE.js'
+with
+{test262:''};
+
```
mgaudet, line 43: This test currently is sort of splinched; should define `with` and ensure `callCount === `0 right?

test/language/module-code/import-attributes/import-assertion-trlng-comma.js
mgaudet: Not sure how the runner would add a new attribute -- the set of expected attributes isn't user-extensible is my understanding. 

For this test case, it seems like `type: 'js'` would be sufficient?

test/language/expressions/dynamic-import/import-attributes/2nd-param-get-assert-error.js
```diff
@@ -0,0 +1,36 @@
...
+        b. Let assertionsObj be Get(options, "assert").
+        c. IfAbruptRejectPromise(assertionsObj, promiseCapability).
+    [...]
+features: [dynamic-import, import-attributes]
+flags: [async]
+---*/
+
+var thrown = new Test262Error();
+var options = {
+  get assert() {
```
mgaudet, line 25: s/assert/with/

test/language/module-code/import-attributes/import-assertion-newlines.js
```diff
@@ -0,0 +1,64 @@
...

{
+
+

test262
+
+

:with
```
mgaudet, line 35: This `with` doesn't seem to belong here.

test/language/module-code/import-attributes/import-assertion-newlines.js
```diff
@@ -0,0 +1,64 @@
...
+
+

''
+
+

};
+import './import-assertion-2_FIXTURE.js' assert
```
mgaudet, line 40: s/assert//

test/language/module-code/import-attributes/import-assertion-newlines.js
```diff
@@ -0,0 +1,64 @@
...
:
+
+

''
+
+

};
+export * from './import-assertion-3_FIXTURE.js' assert
```
mgaudet, line 51: s/assert/with/ right?

test/language/module-code/import-attributes/import-assertion-trlng-comma.js
nicolo-ribaudo: > Not sure how the runner would add a new attribute -- the set of expected attributes isn't user-extensible is my understanding.

It's defined at the embedder level.

> For this test case, it seems like type: 'js' would be sufficient?

In this specific case it would, however:
- the `type` attribute is not defined in ECMA-262. It's defined by the JSON modules proposal, but only with `type: "json"`
- we need multiple attributes for some tests, for example for testing the order properties in dynamic import are accessed.

mgaudet: So hosts will have to implement 'test262-only' attributes?

ptomato: That seems like it'd be prohibitive for runners that aren't a JS engine's shell...

ptomato:
I don't think this was supposed to be auto-closed by #3921!

nicolo-ribaudo:
Yes it was :) This PR was split in separate PRs to make it easier to review.

Issue #3196, opened 2021-09-13
At what point is it acceptable to use new syntax or interfaces?


gibson042:
(prompted by https://github.com/tc39/test262/pull/3192#discussion_r707369883)

To what extent are non-original or "new" features appropriate in testing unrelated older functionality? For example, should an intentionally limited implementation like Hermes or QuickJS or XS or a new in-progress implementation be unable to even run tests covering ES6 functionality because it does not yet support class private fields or String.prototype.matchAll?

Ideally, this is resolved by updating CONTRIBUTING.md with appropriate guidance.

ljharb:
My hope would be that tests can run in the oldest possible engine, and don't assume new features, because then test262 tests can also be run against polyfills *in* those older engines.

However, I believe historically this hasn't been a priority for test authoring.

leobalter:
There is an abstract line of what is old enough, I feel it's subjective and I'd say things dating from the last 2 years are just fine?

The other workaround is flagging things using the features metadata.

There is a lot of maintenance required for targeting multiple versions of EcmaScript, and this also means how we deal with incoming PRs. 

Relieving the rules here will result in avoiding asking PR authors to modify their code to not use modern ES syntax/api. I agree we should avoid edgy features whenever necessary but some are still ok, IMO, e.g. object method definitions, symbols, async functions, optional catch binding.

ljharb:
If the intention is to be able to test polyfills, "time" is insufficient; it'd be "whatever polyfills need to run in", which for all of mine includes IE 6 (ie, ES3 syntax).

leobalter:
Some rationale: It's rare to have bug fixes for tests of features released in ES more than 1 or 2 years. Way more often we have small spec changes, such as PRs to ECMA-262, changing bits of the normative text and adapting tests here.

The compatibility for older ES versions is pretty much compromised, because it flows between what was set in the older ES version to features of that older version that had their adaptation. I wonder if some/most of these implementations are just shiping modern ES but using a subset of the current features. This is likely what happens with XS opting out of Intl API or cross realms functionality.

It's pretty complex to have all the subset implementations and coordinate with the PRs volunteers to make everything compatible. Requiring contributors to use old features only creates a burden in maintenance.

jugglinmike:
Thanks for raising this, @gibson042. You're not the first to wonder; we really need to nail down a policy.

[...]

PR #3845, opened 2023-06-08
Replace arrow function with regular function in harness/sta.js

linusg:
All functions in `assert.js` and `sta.js` (which are the two files required for every test) are already regular functions, except this one.

### DIFF
--- a/harness/sta.js
+++ b/harness/sta.js
@@ -18,7 +18,7 @@ Test262Error.prototype.toString = function () {
   return "Test262Error: " + this.message;
 };
 
-Test262Error.thrower = (message) => {
+Test262Error.thrower = function (message) {
   throw new Test262Error(message);
 };
 


Ms2ger (approved):
This is okay with me, but please hold off on landing for a bit in case anyone wants to discuss this further.

linusg:
Prior art/additional context: #3000. The primary reason why I want to keep core harness files as simple as possible is to shrink the language subset required to start running test262.

gibson042 (approved):
I agree with the "keep core harness files as simple as possible" rationale and would support this anyway based on consistency, but note that #3196 remains unresolved.

ljharb (approved):

ljharb:
In general imo, syntax features should only be used when strictly necessary.

ptomato (approved):
+1 in favour of this.

jugglinmike (approved):
Thanks!

Issue #3840, opened 2023-06-06
Missing Intl.DisplayNames tests for RangeError in .of function


sffc:
According to https://tc39.es/ecma402/#sec-canonicalcodefordisplaynames, there should be RangeErrors thrown when invalid values are passed to Intl.DisplayNames.prototype.of, but I'm not currently getting these exceptions in Chrome, at least not for `type: "language"`:

```javascript
new Intl.DisplayNames("en", { type: "language" }).of("en-GB-fonipa-u-ca-hebrew")
// Expected: RangeError
// Actual: 'British English (FONIPA)'
```

Note that the language is defined to be a string matching `unicode_language_id` from UTS 35: https://unicode.org/reports/tr35/#Canonical_Unicode_Locale_Identifiers

I looked for Test262 to enforce this RangeError but couldn't find any. Please fix.

CC @FrankYFTang @ben-allen @ryzokuken @zbraniecki @romulocintra

PR #3846, opened 2023-06-08
Added tests for structurally valid language tags in DisplayNames.prototype.of()

ben-allen:
Previously there was no test coverage for this function in cases where `type` is 'language', i.e. ` new Intl.DisplayNames(undefined, {type: 'language'}).of('en');`

Related to #3840

### DIFF
--- a/test/intl402/DisplayNames/prototype/of/type-language-valid.js
+++ b/test/intl402/DisplayNames/prototype/of/type-language-valid.js
@@ -1,0 +1,74 @@
+// Copyright (C) 2023 Igalia S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-Intl.DisplayNames.prototype.of
+description: Returns string value for valid `language` codes
+features: [Intl.DisplayNames]
+---*/
+
+
+/*
+unicode_language_id = "root"    // not allowed in ES
+                      | (unicode_language_subtag
+                        (sep unicode_script_subtag)?
+                      | unicode_script_subtag)
+                      (sep unicode_region_subtag)?
+                      (sep unicode_variant_subtag)*
+ */
+
+// unicode_language_subtag = alpha{2,3} | alpha{5,8} ;
+
+var languages = [ { subtag: 'ab', description: '2 letter language_subtag' }, 
+                  { subtag: 'cde', description: '3 letter language_subtag' },
+                  { subtag: 'zzzzzzzz', description: '8 letter language_subtag'} ];
+
+// unicode_script_subtag = alpha{4} ;
+var scripts = [  {subtag: 'abcd', description: '4 letter script_subtag' },
+                 {subtag: '', description: ''} ];
+
+// unicode_region_subtag = (alpha{2} | digit{3}) ;
+var regions = [ {subtag: 'ab', description: '2 letter region_subtag' },
+                {subtag: '123', description: '3 digit region_subtag'},
+                {subtag: '', description: ''} ];
+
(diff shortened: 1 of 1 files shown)


gibson042 (approved):
I wonder if these could be structured or extended to be more clearly comprehensive.
```js
// unicode_language_subtag = alpha{2,3} | alpha{5,8}
const languages = {
  'two-character language': 'ab',
  'three-character language': 'abc',
  'five-character language': 'abcde',
  'eight-character language': 'abcdefgh',
};
// unicode_script_subtag = alpha{4}
const scripts = {
  '': '',
  'script': 'abcd',
};
// unicode_region_subtag = (alpha{2} | digit{3})
const regions = {
  '': '',
  'two-character region': 'ab',
  'three-digit region': '123',
};
// unicode_variant_subtag = (alphanum{5,8} | digit alphanum{3})
const variants = {
  '': '',
  'five-character variant': 'abcde',
  'five-digit variant': '12345',
  'five-character leading-digit variant': '1bcde',
  'five-character trailing-digit variant': 'a2345',
  'eight-character variant': 'abcdefgh',
  'eight-digit variant': '12345678',
  'eight-character leading-digit variant': '1bcdefgh',
  'eight-character trailing-digit variant': 'a2345678',
  'four-character leading-digit variant': '1bcd',
  'four-digit variant': '1234',
};

const notEmpty = subtag => subtag !== '';

const displayNames = new Intl.DisplayNames(undefined, {type: 'language'});

for (const [langLabel, lang] of Object.entries(languages)) {
  for (const [scriptLabel, script] of Object.entries(scripts)) {
    for (const [regionLabel, region] of Object.entries(regions)) {
      for (const [variantLabel, variant] of Object.entries(variants)) {
        const langId = [lang, script, region, variant].filter(notEmpty).join('-');
        const label = [langLabel, scriptLabel, regionLabel, variantLabel].filter(notEmpty).join('; ');
        assert.sameValue(typeof displayNames.of(langId), 'string', label + ': ' + langId);

        // variant is arbitrarily repeatable
        if (variantLabel) {
          const extLabel = label + '; multiple distinct variants';
          let extLangId = langId;
          for (const extraVariant of Object.values(variants).filter(notEmpty)) {
            if (extraVariant !== variant) {
              extLangId += '-' + extraVariant;
            }
          }
          assert.sameValue(typeof displayNames.of(extLangId), 'string', extLabel + ': ' + extLangId);
        }
      }
    }
  }
}
```
(or a more verbose equivalent producing the same 504 assertions using older syntax)

ben-allen:
Ah, thank you! I had been erring hard on the side of "write tests that are not abstract _at all_". Good to see that loops & functions aren't verboten.

Mind if I update the PR with a version of the above minus the ES6 features?

gibson042:
There are existing examples, some of which are very similar to what you're doing here.
* https://github.com/tc39/test262/blob/main/test/intl402/DateTimeFormat/timezone-invalid.js
* https://github.com/tc39/test262/blob/main/test/intl402/Intl/supportedValuesOf/units.js
* https://github.com/tc39/test262/blob/main/test/intl402/language-tags-invalid.js
* https://github.com/tc39/test262/blob/main/test/intl402/Locale/constructor-tag-tostring.js
* https://github.com/tc39/test262/blob/main/test/intl402/Locale/likely-subtags.js
* https://github.com/tc39/test262/blob/main/test/intl402/NumberFormat/numbering-system-options.js
* etc.

ben-allen:
Updated to ES5 version of Richard's suggestion above.

ptomato (approved):
Thanks!
I think the table-style test is an improvement.

PR #3847, opened 2023-06-09
Temporal tests for "stop coercing non-string primitives to strings" (tc39/proposal-temporal#2574)

justingrant:
Test changes to support https://github.com/tc39/proposal-temporal/pull/2574.

### DIFF
--- a/test/built-ins/Temporal/Calendar/argument-wrong-type.js
+++ b/test/built-ins/Temporal/Calendar/argument-wrong-type.js
@@ -1,0 +1,30 @@
+// Copyright (C) 2023 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.calendar
+description: RangeError thrown when constructor invoked with the wrong type
+features: [Temporal]
+---*/
+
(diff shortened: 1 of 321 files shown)


justingrant:
The PR for these tests, https://github.com/tc39/proposal-temporal/pull/2574, was approved in the July 2023 TC39 meeting, so taking this PR out of draft state.

ptomato (approved):
I reviewed a representative sample of the repetitive tests; obviously not all 300+ of them. I also reviewed the changes to staging. Thanks for the comprehensive work, @justingrant!

I have a minor thing I'd like to change about how the tests are now organized, I think the current division into files no longer makes sense after this change. But that's a "me problem", so I'll open an issue for it and probably make that change myself in a few weeks, as I don't see any reason to hold up this PR! I'd like to ensure that JS engines have access to the newly-correct tests as soon as possible.

(Details: I'd fold the `calendar-number` tests into `calendar-wrong-type` since there is no longer any special behaviour of numbers to be tested; and I'd split the string tests out of `calendar-wrong-type` into something like `calendar-invalid-iso-string` because that now throws a different error. But still, I'll fix this myself later.)

PR #3848, opened 2023-06-11
Work around Node 20 / ICU 73.1 calendar bug

justingrant:
These tests accompany a Temporal PR that works around a bug in ICU 73.1 (which ships with Node 20). These test changes enable tests to pass in both the buggy state and after the bug is eventually fixed.


ptomato (approved):
I'd also be in favour of simply removing those 'Year 1 snapshot' tests from staging ­— they seem too locale-formatting-specific for test262 and I think it's unlikely we'd convert them to actual test262 tests in that form.

Issue #3828, opened 2023-05-16
Tests for Float16Array


bakkot:
The [Float16Array proposal](https://github.com/tc39/proposal-float16array) is now stage 3 (🎉). That means it's time for tests! Here's the things which occur to me offhand; feel free to edit this issue to add more.

- the usual tests around property descriptors etc
- rounding behavior (for Float16Array instances as well as for the DataView methods and for f16round), including at least the following cases:
  - 0, NaN, Infinity
  - the largest finite double which is precisely representable as a float16: 65504
  - the smallest integral double which is not precisely representable as a float16: 2049
  - the smallest double which is precisely representable as a non-subnormal float16: 0.00006103515625
  - the largest double which is precisely representable as a subnormal float16: 0.000060975551605224609375
  - the smallest double which is precisely representable precisely as a float16: 0.000000059604644775390625
  - the largest double which rounds to 0 when cast to float16: 0.0000000298023223876953125
  - the smallest double which does not round to 0 when cast to float16, i.e. the next double after the one in the previous bullet point: 0.00000002980232238769531911744490042422139897126953655970282852649688720703125
  - the smallest double which rounds to Infinity when cast to float16: $b^{emax} \times \lparen b - \frac{1}{2} b ^ {1 - p} \rparen$ i.e. $2^{15} \times \lparen 2 - \frac{1}{2} 2 ^ {1 - 11} \rparen =$ 65520
  - the largest double which does not round to Infinity when cast to float16 (i.e. the next double below 65520): 65519.9999999999927240423858165740966796875
  - the smallest double which rounds to a non-subnormal float16: 0.0000610053539276123046875
  - the largest double which rounds to a subnormal float16, i.e. the next double below the one in the previous bullet point: 0.0000610053539276122979112364219655972874534199945628643035888671875
  - some values which are precisely halfway between two adjacent float16 values, to test roundTiesToEven, including
    - a subnormal which rounds up
    - a subnormal which rounds down
    - a number with magnitude between 1 and 2047 which rounds up
    - a number with magnitude between 1 and 2047 which rounds down
    - a number with magnitude above 2047 which rounds up
    - a number with magnitude above 2047 which rounds down
  - all of those again but negative

Incidentally [float.exposed](https://float.exposed/) is a great tool for exploring the representation of these values, ~~though it has some edge-case bugs when converting between representations~~ (edit: no longer true; I reported the issue and the owner fixed it).

bakkot:
I've started working on this.

PR #3849, opened 2023-06-11
add Float16Array to all generic TypedArray tests

bakkot:
First step of #3828.

I've updated `harness/testTypedArray.js` so that it includes Float16Array iff it is available as a global, so as to not render all tests using it useless for any implementation which has not yet implemented Float16Array. I've also updated a bunch of tests which weren't using that helper to use it, so that they wouldn't require separate updates for Float16Array.

In principle these should behave identically on existing implementations as prior to this PR.

I'll have a follow-up adding Float16Array-specific tests.

### DIFF
--- a/harness/testTypedArray.js
+++ b/harness/testTypedArray.js
@@ -4,22 +4,23 @@
 description: |
     Collection of functions used to assert the correctness of TypedArray objects.
 defines:
-  - typedArrayConstructors
   - floatArrayConstructors
   - intArrayConstructors
+  - typedArrayConstructors
   - TypedArray
   - testWithTypedArrayConstructors
+  - nonAtomicsFriendlyTypedArrayConstructors
   - testWithAtomicsFriendlyTypedArrayConstructors
   - testWithNonAtomicsFriendlyTypedArrayConstructors
   - testTypedArrayConversions
 ---*/
 
-/**
- * Array containing every typed array constructor.
- */
-var typedArrayConstructors = [
+var floatArrayConstructors = [
   Float64Array,
-  Float32Array,
+  Float32Array
+];
+
+var intArrayConstructors = [
   Int32Array,
   Int16Array,
   Int8Array,
@@ -29,8 +30,17 @@ var typedArrayConstructors = [
   Uint8ClampedArray
 ];
 
-var floatArrayConstructors = typedArrayConstructors.slice(0, 2);
-var intArrayConstructors = typedArrayConstructors.slice(2, 7);
+// Float16Array is a newer feature
+// adding it to this list unconditionally would cause implementations lacking it to fail every test which uses it
+if (typeof Float16Array !== 'undefined') {
+  floatArrayConstructors.push(Float16Array);
+}
+
+/**
+ * Array containing every non-bigint typed array constructor.
+ */
+
+var typedArrayConstructors = floatArrayConstructors.concat(intArrayConstructors);
 
 /**
  * The %TypedArray% intrinsic constructor function.
@@ -63,6 +73,7 @@ function testWithTypedArrayConstructors(f, selected) {
   }
 }
 
+var nonAtomicsFriendlyTypedArrayConstructors = floatArrayConstructors.concat([Uint8ClampedArray]);
 /**
  * Calls the provided function for every non-"Atomics Friendly" typed array constructor.
  *
@@ -70,11 +81,7 @@ function testWithTypedArrayConstructors(f, selected) {
  * @param {Array} selected - An optional Array with filtered typed arrays
  */
 function testWithNonAtomicsFriendlyTypedArrayConstructors(f) {
-  testWithTypedArrayConstructors(f, [
-    Float64Array,
-    Float32Array,
-    Uint8ClampedArray
-  ]);
+  testWithTypedArrayConstructors(f, nonAtomicsFriendlyTypedArrayConstructors);
 }
 
 /**
--- a/harness/testTypedArray.js
+++ b/harness/testTypedArray.js
@@ -127,3 +127,31 @@ function testTypedArrayConversions(byteConversionValues, fn) {
     });
   });
 }
+
+/**
+ * Checks if the given argument is one of the float-based TypedArray constructors.
+ *
+ * @param {constructor} ctor - the value to check
+ * @returns {boolean}
+ */
+function isFloatTypedArrayConstructor(arg) {
+  return floatArrayConstructors.indexOf(arg) !== -1;
+}
+
+/**
+ * Determines the precision of the given float-based TypedArray constructor.
+ *
+ * @param {constructor} ctor - the value to check
+ * @returns {string} "half", "single", or "double" for Float16Array, Float32Array, and Float64Array respectively.
+ */
+function floatTypedArrayConstructorPrecision(FA) {
+  if (typeof Float16Array !== "undefined" && FA === Float16Array) {
+    return "half";
+  } else if (FA === Float32Array) {
+    return "single";
+  } else if (FA === Float64Array) {
+    return "double";
+  } else {
+    throw new Error("Malformed test - floatTypedArrayConstructorPrecision called with non-float TypedArray");
+  }
+}
--- a/test/built-ins/TypedArray/prototype/fill/fill-values-conversion-operations-consistent-nan.js
+++ b/test/built-ins/TypedArray/prototype/fill/fill-values-conversion-operations-consistent-nan.js
@@ -73,12 +73,7 @@ features: [TypedArray]
 ---*/
 
 testWithTypedArrayConstructors(function(FA) {
-  var precision =
-    FA === Float32Array
-    ? "single"
-    : (typeof Float16Array !== "undefined" && TA === Float16Array)
-    ? "half"
-    : "double";
+  var precision = floatTypedArrayConstructorPrecision(FA);
   var samples = new FA(3);
   var controls, idx, aNaN;
 
--- a/test/built-ins/TypedArray/prototype/fill/fill-values-conversion-operations-consistent-nan.js
+++ b/test/built-ins/TypedArray/prototype/fill/fill-values-conversion-operations-consistent-nan.js
@@ -73,7 +73,12 @@ features: [TypedArray]
 ---*/
 
 testWithTypedArrayConstructors(function(FA) {
-  var precision = FA === Float32Array ? "single" : "double";
+  var precision =
+    FA === Float32Array
+    ? "single"
+    : (typeof Float16Array !== "undefined" && TA === Float16Array)
+    ? "half"
+    : "double";
   var samples = new FA(3);
   var controls, idx, aNaN;
 
@@ -98,4 +103,4 @@ testWithTypedArrayConstructors(function(FA) {
       );
     }
   }
-}, [Float32Array, Float64Array]);
+}, floatArrayConstructors);
--- a/test/built-ins/TypedArray/prototype/filter/callbackfn-resize.js
+++ b/test/built-ins/TypedArray/prototype/filter/callbackfn-resize.js
@@ -15,7 +15,7 @@ assert.sameValue(typeof ArrayBuffer.prototype.resize, 'function');
 
 testWithTypedArrayConstructors(function(TA) {
   var BPE = TA.BYTES_PER_ELEMENT;
-  var NaNvalue = (TA === Float32Array || TA === Float64Array) ? NaN : 0;
+  var NaNvalue = (TA === Float32Array || TA === Float64Array || (typeof Float16Array !== 'undefined' && TA === Float16Array)) ? NaN : 0;
   var buffer = new ArrayBuffer(BPE * 3, {maxByteLength: BPE * 4});
   var sample = new TA(buffer);
   var finalElement, finalResult, expectedElements, expectedIndices, expectedArrays;
--- a/test/built-ins/TypedArray/prototype/filter/callbackfn-resize.js
+++ b/test/built-ins/TypedArray/prototype/filter/callbackfn-resize.js
@@ -15,7 +15,7 @@ assert.sameValue(typeof ArrayBuffer.prototype.resize, 'function');
 
 testWithTypedArrayConstructors(function(TA) {
   var BPE = TA.BYTES_PER_ELEMENT;
-  var NaNvalue = (TA === Float32Array || TA === Float64Array || (typeof Float16Array !== 'undefined' && TA === Float16Array)) ? NaN : 0;
+  var NaNvalue = isFloatTypedArrayConstructor(TA) ? NaN : 0;
   var buffer = new ArrayBuffer(BPE * 3, {maxByteLength: BPE * 4});
   var sample = new TA(buffer);
   var finalElement, finalResult, expectedElements, expectedIndices, expectedArrays;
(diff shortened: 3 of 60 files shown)


ljharb:
The harness doesn't include the bigint arrays, how are those tested?

bakkot:
A separate harness: [testBigIntTypedArray.js](https://github.com/tc39/test262/blob/c4642dd714175b5d27939c920abc6059c9fddb06/harness/testBigIntTypedArray.js)

But they kind of have to be, since otherwise all consumers need to work with both Number and BigInt, which is pretty annoying; might as well just split them up. That's not true for Float16Array, since it's just another Number-containing TA kind.

ljharb:
If the current split is by number type, then this approach seems fine.

jugglinmike:
I like how this approach minimizes maintenance effort, but I've been wondering if the effect on consumers is acceptable.

As far as I know (which admittedly isn't very far these days), Test262 doesn't make assertions conditionally anywhere else. I think there's value in a test suite that makes the same statement about conformance for everyone (e.g. when comparing results across implementations), so I don't want to give that up unintentionally.

The only alternative comes to mind is to duplicate the tests. I'm not thrilled about that idea, either, but we might mitigate the worst of the maintenance hassle with a new rule for Test262's linter.

bakkot:
I think in practice it still amounts to the same assertion - ["`Atomics.wait` throws for every numeric typed array kind other than Int32Array"](https://github.com/tc39/test262/blob/2f0193d4cf583091f7717b7851270a75a78826e4/test/built-ins/Atomics/wait/validate-arraytype-before-timeout-coercion.js), for example. it doesn't make sense for an implementation to fail the assertion because of an engine not supporting a new Typed Array kind.

If we were to duplicate the tests, I think the way to do it would be to migrate all of them to templates, and generate one test per typed array kind. (The current template system isn't ideally suited to this kind of thing, but it would suffice.) That way it would have sufficient granularity for an engine to fail only the Float16Array variants, without needing the logic itself to be duplicated in source. Buuuuuut that migration would be a lot of work; I wouldn't want to block adding these tests on it.

(FWIW, @syg [said](https://matrixlogs.bakkot.com/TC39_Test262_Maintainers/2023-06-13#L16) the approach in this PR would work for V8.)

ljharb:
fwiw I think it's very reasonable to land this now, with a future roadmap item to change these tests to be generated per-typed-array.

bakkot:
Yes, I'd really like to see this landed as-is, unless we hear from a consumer that it's a big problem for them. As a consumer (in several projects) this works fine for me. Sounded like it was fine for V8 too.

test/built-ins/TypedArrayConstructors/ctors/object-arg/returns.js
```diff
@@ -35,7 +35,7 @@ testWithTypedArrayConstructors(function(TA) {
   assert.sameValue(typedArray.constructor, TA);
   assert.sameValue(Object.getPrototypeOf(typedArray), TA.prototype);
 
-  if (TA === Float32Array || TA === Float64Array) {
+  if (TA === Float32Array || TA === Float64Array || (typeof Float16Array !== 'undefined' && TA === Float16Array)) {
```
Ms2ger, line 38: I've seen this condition a bit too often in this PR; let's add a helper function to `testTypedArray.js` (or would we prefer `floatArrayConstructors.indexOf...`?)

test/built-ins/TypedArrayConstructors/internals/DefineOwnProperty/conversion-operation-consistent-nan.js
```diff
@@ -69,7 +69,12 @@ features: [TypedArray]
 ---*/
 
 testWithTypedArrayConstructors(function(FA) {
-  var precision = FA === Float32Array ? "single" : "double";
+  var precision =
+    FA === Float32Array
+    ? "single"
+    : (typeof Float16Array !== "undefined" && TA === Float16Array)
+    ? "half"
+    : "double";
```
Ms2ger, line 77: This one would perhaps be good to factor out as well.

Ms2ger (commented):
It's somewhat unfortunate that the "one file = one result" approach pushes us in this direction, but within that context, this seems fine.

test/built-ins/TypedArray/prototype/fill/fill-values-conversion-operations-consistent-nan.js
```diff
@@ -73,7 +73,12 @@ features: [TypedArray]
 ---*/
 
 testWithTypedArrayConstructors(function(FA) {
-  var precision = FA === Float32Array ? "single" : "double";
+  var precision =
+    FA === Float32Array
+    ? "single"
+    : (typeof Float16Array !== "undefined" && TA === Float16Array)
```
ljharb, line 79:
```suggestion
    : FA === Float16Array
```
TA → FA, but also when could this ever be `Float16Array` when the global isn't present?

test/built-ins/TypedArray/prototype/filter/callbackfn-resize.js
```diff
@@ -15,7 +15,7 @@ assert.sameValue(typeof ArrayBuffer.prototype.resize, 'function');
 
 testWithTypedArrayConstructors(function(TA) {
   var BPE = TA.BYTES_PER_ELEMENT;
-  var NaNvalue = (TA === Float32Array || TA === Float64Array) ? NaN : 0;
+  var NaNvalue = (TA === Float32Array || TA === Float64Array || (typeof Float16Array !== 'undefined' && TA === Float16Array)) ? NaN : 0;
```
ljharb, line 18:
```suggestion
  var NaNvalue = (TA === Float32Array || TA === Float64Array || TA === Float16Array) ? NaN : 0;
```

test/built-ins/TypedArray/prototype/map/callbackfn-resize.js
```diff
@@ -15,7 +15,7 @@ assert.sameValue(typeof ArrayBuffer.prototype.resize, 'function');
 
 testWithTypedArrayConstructors(function(TA) {
   var BPE = TA.BYTES_PER_ELEMENT;
-  var NaNvalue = (TA === Float32Array || TA === Float64Array) ? NaN : 0;
+  var NaNvalue = (TA === Float32Array || TA === Float64Array || (typeof Float16Array !== 'undefined' && TA === Float16Array)) ? NaN : 0;
```
ljharb, line 18: same here, etc

ljharb (commented):
seems like there's a lot of places the conditional global check isn't needed, if it's conditionally added to `typedArrayConstructors`?

test/built-ins/TypedArray/prototype/fill/fill-values-conversion-operations-consistent-nan.js
Ms2ger: We want to avoid
```
Uncaught ReferenceError: Float16Array is not defined
```

ljharb: ahh duh, not sure how i missed that (－‸ლ)

ljharb:
```suggestion
    : (typeof Float16Array !== "undefined" && FA === Float16Array)
```

test/built-ins/TypedArrayConstructors/internals/DefineOwnProperty/conversion-operation-consistent-nan.js
Ms2ger: Especially because that would fix the TA/FA confusion here at the same time

bakkot:
@Ms2ger Done. I also updated one test which was defining but not using `precision`; seemed like it probably meant to use it.

test/built-ins/TypedArray/prototype/set/typedarray-arg-set-values-same-buffer-other-type.js
```diff
@@ -44,6 +44,11 @@ var expected = {
 };
 
 testWithTypedArrayConstructors(function(TA) {
+  if (typeof Float16Array !== 'undefined' && TA === Float16Array) {
+    // TODO(bakkot) work out bit patterns here
+    return;
+  }
```
Ms2ger, line 50: Do you want to do that in this PR?

test/built-ins/TypedArrayConstructors/from/nan-conversion.js
```diff
@@ -46,13 +43,4 @@ testWithTypedArrayConstructors(function(TA) {
...
-  Int8Array,
-  Int32Array,
-  Int16Array,
-  Int8Array,
-  Uint32Array,
-  Uint16Array,
-  Uint8Array,
-  Uint8ClampedArray
-]);
+floatArrayConstructors);
```
Ms2ger, line 46: Those don't look like floats

Ms2ger (approved):
Modulo comments below, this looks good to me.

test/built-ins/TypedArray/prototype/set/typedarray-arg-set-values-same-buffer-other-type.js
bakkot: Suppose I might as well. Done.

test/built-ins/TypedArrayConstructors/from/nan-conversion.js
bakkot: Whoops, fixed.

bakkot:
@Ms2ger I'm not sure what you want me to do with that error message. `test/built-ins/Atomics/add/validate-arraytype-before-index-coercion.js` does not currently have the `TypedArray` feature flag, and this PR doesn't affect it either way. It's just a straight refactoring.

Are you saying that I should fix the tests which currently lack that flag as part of this PR?

Ms2ger:
It seems like the linter only runs on changed files, so your changes caused it to catch preexisting bugs :facepalm: 

In order of preference, you could:
1. Fix the linter and any fallout :innocent: 
2. Add the flags in a separate PR to be landed first (I promise speedy review)
3. Add the flags here

bakkot:
Done, but I'm not really looking forward to fixing the resulting merge conflicts... They're all trivial, it's just that it's a lot of them and I don't know of an easy way to make `git` address them itself.

I'll get to it eventually though if no one else does.

Ms2ger:
I rebased, thanks!

bakkot:
I think this is ready then?

harness/testTypedArray.js
```diff
@@ -29,8 +30,17 @@ var typedArrayConstructors = [
   Uint8ClampedArray
 ];
 
-var floatArrayConstructors = typedArrayConstructors.slice(0, 2);
-var intArrayConstructors = typedArrayConstructors.slice(2, 7);
```
iamstolis, line 33: `intArrayConstructors` did not contain `Uint8ClampedArray` but it is included now. This seems to be an intended change. Unfortunately, some `Atomics` tests (like `built-ins/Atomics/add/good-views.js`) still use `intArrayConstructors`. In other words, these tests are incorrect now. They should be rewritten to use `testWithAtomicsFriendlyTypedArrayConstructors()`, I guess.

bakkot: Oh, that was not an intended change, I just didn't realize that the `.slice` above stopped before the end of the list.

In fact the `.slice` was previously also omitting `Uint8Array`, which seems wrong. I've left it in, but opened https://github.com/tc39/test262/pull/3984 to remove Uint8ClampedArray.

iamstolis: I think that the original idea was to include `Uint8ClampedArray` in `intArrayConstructors`. The author just haven't realized that the second argument in `typedArrayConstructors.slice(2, 7);` is not the length but an (excluded) end index. That's why `Uint8Array` and `Uint8ClampedArray` were missing by accident. I find it correct to include `Uint8ClampedArray` in this array (it is a typed array with int elements). Some tests may skip `Uint8ClampedArray` unnecessarily when it is not included. That's why I though that the change was intentional and that just the problematic Atomics tests (that use `intArrayConstructors` without excluding `Uint8ClampedArray`) should be fixed/updated.

iamstolis: The following tests started to fail (when run against `graal-js` engine). So, this is the correct list, hopefully.
```
Unexpectedly failed 'built-ins/Atomics/add/bad-range.js'
Unexpectedly failed 'built-ins/Atomics/add/good-views.js'
Unexpectedly failed 'built-ins/Atomics/and/bad-range.js'
Unexpectedly failed 'built-ins/Atomics/and/good-views.js'
Unexpectedly failed 'built-ins/Atomics/compareExchange/bad-range.js'
Unexpectedly failed 'built-ins/Atomics/compareExchange/good-views.js'
Unexpectedly failed 'built-ins/Atomics/exchange/bad-range.js'
Unexpectedly failed 'built-ins/Atomics/exchange/good-views.js'
Unexpectedly failed 'built-ins/Atomics/load/bad-range.js'
Unexpectedly failed 'built-ins/Atomics/load/good-views.js'
Unexpectedly failed 'built-ins/Atomics/or/bad-range.js'
Unexpectedly failed 'built-ins/Atomics/or/good-views.js'
Unexpectedly failed 'built-ins/Atomics/store/bad-range.js'
Unexpectedly failed 'built-ins/Atomics/store/good-views.js'
Unexpectedly failed 'built-ins/Atomics/sub/bad-range.js'
Unexpectedly failed 'built-ins/Atomics/sub/good-views.js'
Unexpectedly failed 'built-ins/Atomics/xor/bad-range.js'
Unexpectedly failed 'built-ins/Atomics/xor/good-views.js'
```

bakkot: @iamstolis Thanks, see https://github.com/tc39/test262/pull/3984.

iamstolis: Thanks, I run the tests from the current revision (`62294b592e332bc16da6114e461fd7a8ed90daab`) of https://github.com/tc39/test262/pull/3984 against `graal-js` and I haven't seen any unexpected failures.

harness/testTypedArray.js
```diff
@@ -29,8 +30,17 @@ var typedArrayConstructors = [
...
+// adding it to this list unconditionally would cause implementations lacking it to fail every test which uses it
+if (typeof Float16Array !== 'undefined') {
+  floatArrayConstructors.push(Float16Array);
+}
+
+/**
+ * Array containing every non-bigint typed array constructor.
+ */
+
+var typedArrayConstructors = floatArrayConstructors.concat(intArrayConstructors);
```
iamstolis, line 43: This PR included `Float16Array` in `typedArrayConstructors`. Unfortunately, it failed to provide `byteConversionValues.expected.Float16` (in `harness/byteConversionValues.js`). So, the tests that use `testTypedArrayConversions(byteConversionValues, ...)` fail when `Float16Array` is enabled. This seems to affect the following tests:
```
built-ins/TypedArray/prototype/fill/fill-values-conversion-operations.js
built-ins/TypedArray/prototype/map/return-new-typedarray-conversion-operation.js
built-ins/TypedArray/prototype/set/array-arg-src-tonumber-value-conversions.js
built-ins/TypedArray/prototype/set/typedarray-arg-set-values-diff-buffer-other-type-conversions-sab.js
built-ins/TypedArray/prototype/set/typedarray-arg-set-values-diff-buffer-other-type-conversions.js
built-ins/TypedArrayConstructors/ctors/object-arg/conversion-operation.js
built-ins/TypedArrayConstructors/internals/DefineOwnProperty/conversion-operation.js
built-ins/TypedArrayConstructors/internals/Set/conversion-operation.js
```

ptomato: @iamstolis Please open an issue - if no-one happens to have time to fix this right when they see it, it'll likely get lost out of their notifications.

iamstolis: OK, I have created #3989

PR #3855, opened 2023-06-15
Add missing iterator-helpers feature flag

syg:


syg:
This is needed because the test was updated on line 29: `assert.sameValue(toString.call(strIter), '[object Iterator]');`

ptomato (approved):

saifulrizal (approved):

saifulrizal (changes requested):
Nice

PR #3858, opened 2023-06-20
Temporal Issue 2532

guijemont:
Add tests for the new PrepareTemporalFields behavior, that has some observable changes where many functions throw when encountering a calendar the fields() method of which returns duplicates, certain values that create duplicates internally, `__proto__` or `constructor`.

See https://github.com/tc39/proposal-temporal/issues/1315


guijemont:
I've left the changes in many individual commits in the hope that it makes review easier, but it probably makes more sense to land this as one commit.

ptomato (approved):
I reviewed a representative sample of this, but it looks good.

PR #3861, opened 2023-06-23
Update temporalHelpers.js

justingrant:
* Fix `SpecificOffsetTimeZone` helper class so that it correctly implements the time zone protocol. Previously, some tests were passing but not actually exercising the expected codepaths.
* Add `assertDateDuration` function, which makes it shorter to assert durations that only contain date components.

### DIFF
--- a/harness/temporalHelpers.js
+++ b/harness/temporalHelpers.js
@@ -57,6 +57,26 @@ var TemporalHelpers = {
   },
 
   /*
+   * assertDateDuration(duration, years, months, weeks, days, [, description]):
+   *
+   * Shorthand for asserting that each date field of a Temporal.Duration is
+   * equal to an expected value.
+   */
+  assertDateDuration(duration, years, months, weeks, days, description = "") {
+    assert(duration instanceof Temporal.Duration, `${description} instanceof`);
+    assert.sameValue(duration.years, years, `${description} years result`);
+    assert.sameValue(duration.months, months, `${description} months result`);
+    assert.sameValue(duration.weeks, weeks, `${description} weeks result`);
+    assert.sameValue(duration.days, days, `${description} days result`);
+    assert.sameValue(duration.hours, 0, `${description} hours result should be zero`);
+    assert.sameValue(duration.minutes, 0, `${description} minutes result should be zero`);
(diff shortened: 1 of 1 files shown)


harness/temporalHelpers.js
```diff
@@ -56,6 +56,20 @@ var TemporalHelpers = {
...
+   * Shorthand for asserting that each date field of a Temporal.Duration is
+   * equal to an expected value.
+   */
+  assertDateDuration(duration, years, months, weeks, days, description = "") {
+    assert(duration instanceof Temporal.Duration, `${description} instanceof`);
+    assert.sameValue(duration.years, years, `${description} years result`);
+    assert.sameValue(duration.months, months, `${description} months result`);
+    assert.sameValue(duration.weeks, weeks, `${description} weeks result`);
+    assert.sameValue(duration.days, days, `${description} days result`);
+  },
```
Ms2ger, line 71: I would expect this to also test that the other fields are zero.

justingrant: Fixed! Also rebased to get rid of the failing Graals action.

ptomato (approved):
> Fix SpecificOffsetTimeZone helper class so that it correctly implements the time zone protocol. Previously, some tests were passing but not actually exercising the expected codepaths.

I definitely intended the `[]` as a stub, because it hadn't been needed so far. Now that it's needed for the time zone offset tests, seems great to implement it :smile: 

Thanks!

harness/temporalHelpers.js
```diff
@@ -1784,8 +1804,14 @@ var TemporalHelpers = {
         return this._offsetValue;
       }
 
-      getPossibleInstantsFor() {
-        return [];
+      getPossibleInstantsFor(dt) {
+        if (typeof this._offsetValue !== 'number' || Math.abs(this._offsetValue) >= 86400e9 || isNaN(this._offsetValue)) return [];
```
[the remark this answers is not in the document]
justingrant: If I remember correctly, I did it this way so that tests would fail in the right place in cases where the offset was invalid: either not a number or out of range. Otherwise, the failures would happen in the next line here in temporalHelpers, instead of actually exercising Temporal implementations' ability to reject invalid offset time zone IDs.

PR #3862, opened 2023-06-23
Tests for proposal-temporal#2607: Limit offset time zones to minutes precision

justingrant:
Tests for Temporal normative PR tc39/proposal-temporal#2607: Limit offset time zones to minutes precision. EDIT: this PR reached consensus in the July 2023 TC39 meeting.

~~This is a draft PR until the corresponding Temporal PR is presented at the upcoming July 2023 TC39 plenary meeting.~~


test/built-ins/Temporal/Duration/compare/relativeto-sub-minute-offset.js
```diff
@@ -3,19 +3,25 @@
...
 const duration2 = new Temporal.Duration(0, 1);
 
-let relativeTo = "2000-01-01T00:00+00:45[+00:44:30.123456789]";
-assert.sameValue(Temporal.Duration.compare(duration1, duration2, { relativeTo }), 0, "rounded HH:MM is accepted in string");
+let result;
+let relativeTo;
 
-relativeTo = "2000-01-01T00:00+00:44:30[+00:44:30.123456789]";
-assert.throws(RangeError, () => Temporal.Duration.compare(duration1, duration2, { relativeTo }), "no other rounding is accepted for offset");
+const action = (relativeTo) => Temporal.Duration.compare(duration1, duration2, { relativeTo });
```
ptomato, line 16: There are differing opinions among the test262 maintainers about whether this degree of deduplication is desirable in tests — I don't have a strong opinion but others might.

gibson042: I'm generally on the side that favors abstraction, but `action(relativeTo)` strikes me as confoundingly opaque, and the test cases don't make clear what this file is covering. I'd suggest clarifying that and varying _only_ relevant details across cases... IIUC, maybe something more like this?
```
const P31D = new Temporal.Duration(0, 0, 0, 31);
const P1M = new Temporal.Duration(0, 1);
const compareRelativeTo = relativeTo => Temporal.Duration.compare(P31D, P1M, { relativeTo });

let result;

result = compareRelativeTo("1970-01-01T00:00+00:45:00.000000000[+00:45]");
assert.sameValue(result, 0, "UTC offset with zeroed seconds must be accepted (string)");

result = compareRelativeTo({ year: 1970, month: 1, day: 1, offset: "+00:45:00.000000000", timeZone: "+00:45" });
assert.sameValue(result, 0, "UTC offset with zeroed seconds must be accepted (property bag)");

assert.throws(RangeError, () => compareRelativeTo("1970-01-01T00:00+00:45:01[+00:45]"),
  "mismatching seconds between UTC offset and offset time zone must be rejected");
```

justingrant: If I remember correctly, there were 5 test files that were identical except for the method being tested. They used the same object and string values to test against, and the same messages.

Furthermore, because all of them rely on the same underlying parts of the spec, it'd be likely for all these 5 files to fail as a unit; either they'd all pass or they'd all fail.

So my thinking was that a generic `action()` would be helpful, to direct readers' attention away from the specific methods and towards the more generalized patterns that all 5 files have in common, so they'd be thinking "oh, my problem is with relativeTo handling, not `compare` or `add`".

But also happy to differentiate these files if you guys think it's needed. I'm just not convinced that the current state is bad, and it might even be better. 

What do you think? Not a big deal for me either way.

> test cases don't make clear what this file is covering

Could you explain a bit more about this?  If these were long or complex files then I'd agree, but all of these files seem to have very clear descriptions for each test, and the files themselves are short enough to read the whole thing without scrolling. To me they seemed pretty clear, but I might not correctly understand the mindset of an implementer or polyfill author using these tests.

ptomato: I'm fine with it as is, I take your point about the 5 files passing or failing together. But I didn't have a strong opinion in the first place.

gibson042: My opinions here are also not strong, but to respond:

> If I remember correctly, there were 5 test files that were identical except for the method being tested. They used the same object and string values to test against, and the same messages.
> 
> Furthermore, because all of them rely on the same underlying parts of the spec, it'd be likely for all these 5 files to fail as a unit; either they'd all pass or they'd all fail.
> 
> So my thinking was that a generic `action()` would be helpful, to direct readers' attention away from the specific methods and towards the more generalized patterns that all 5 files have in common, so they'd be thinking "oh, my problem is with relativeTo handling, not `compare` or `add`".

Hmm, I wonder if it would make sense to define them [procedurally](https://github.com/tc39/test262/blob/main/CONTRIBUTING.md#procedurally-generated-tests) (e.g., **test/built-ins/Temporal/Duration/relativeto-sub-minute-offset.template** with <b>*.case</b> files for generating **test/built-ins/Temporal/Duration/relativeto-sub-minute-offset-{compare,prototype-add,prototype-round,…}.js**)? And if not, then cross-referencing comments in these files would be helpful.

> > test cases don't make clear what this file is covering
> 
> Could you explain a bit more about this? If these were long or complex files then I'd agree, but all of these files seem to have very clear descriptions for each test, and the files themselves are short enough to read the whole thing without scrolling. To me they seemed pretty clear, but I might not correctly understand the mindset of an implementer or polyfill author using these tests.

Taking this file as an example, the description of "_**relativeTo string accepts trailing zeroes in sub-minute UTC offset**_" (which would probably be better as something like "_**relativeTo sub-minute UTC offset is allowed if zero**_") seems to suggest that the content should verify acceptance of UTC offsets like "-00:45:00.000000000" and "-004500.000000000" in both strings and property bags, and should verify rejection of UTC offsets like "-00:44:30" _regardless_ of the corresponding time zone (i.e., the reason for rejection should be the presence of nonzero sub-minute data rather than a mismatch between UTC offset and time zone), all which should be made clear by assertion messages like "_{basic,extended}-format UTC offset with zeroed seconds and nanoseconds must be accepted ({string,property bag})_" and "_{basic,extended}-format UTC offset with nonzero seconds must be rejected ({string,property bag})_" as suggested above. And for completeness, I'd probably split the rejection case into further variants like "no time zone" **1970-01-01T00:00-00:44:30**, "offset time zone" **1970-01-01T00:00-00:44:30[-00:45]**, "named time zone" **1970-01-01T00:00-00:44:30[Africa/Monrovia]**, and "custom time zone" `{ year: 1970, month: 1, day: 1, offset: "-00:44:30", timeZone: new FixedOffsetTimeZone("-00:44:30", "MMT") }` (which come to think of it probably also implies a worthwhile positive case like "custom time zone" `{ …, offset: "+00:01:00.000000000", timeZone: new FixedOffsetTimeZone("+00:01", "plus_one_min") }`).

justingrant: @gibson042 - I don't disagree with your feedback but I also think the code as-is seems good enough for the time being, and I've got a lot of other PRs to catch up with. So if OK with you, I'll leave this PR as-is, but if you wanted to make changes that'd be OK with me!

ptomato: @gibson042 I agree with Justin, I'd prefer not to hold this up while rewriting it to be procedurally-generated.

ptomato:
There was apparently a blip in CircleCI where it failed to pull the branch from GitHub. I re-ran the required job.

PR #3867, opened 2023-07-05
add tests for Promise.withResolvers proposal

peetklecha:
This shouldn't be merged yet as Promise.withResolvers is still only Stage 2.


test/built-ins/Promise/withResolvers/result.js
```diff
@@ -0,0 +1,18 @@
+// Copyright (C) 2023 Peter Klecha. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Promise.withResolvers result is an object with keys "promise", "reject", and "resolve"
+esid: sec-promise.withresolvers
+features: [promise-with-resolvers]
```
ljharb, line 7:
```suggestion
features: [promise-with-resolvers, Object.hasOwn]
```
unless you want to use `.hasOwnProperty` below

peetklecha: thanks for this!

peetklecha:
Now Stage 3.

test/built-ins/Promise/withResolvers/result.js
```diff
@@ -0,0 +1,18 @@
...
+description: Promise.withResolvers result is an object with keys "promise", "reject", and "resolve"
+esid: sec-promise.withresolvers
+features: [promise-with-resolvers, Object.hasOwn]
+---*/
+
+
+var instance = Promise.withResolvers();
+
+assert.sameValue(typeof instance, "object");
+assert.notSameValue(instance, null);
```
ljharb, line 14:
```suggestion
assert.notSameValue(instance, null);
assert(instance intanceof Object);
```

ljharb (approved):

features.txt
```diff
@@ -108,6 +108,10 @@ json-parse-with-source
 # https://github.com/tc39/proposal-iterator-helpers
 iterator-helpers
 
+# Promise.withResolvers
+# https://github.com/tc30/proposal-promise-with-resolvers
```
trflynn89, line 112: Noticed a typo here - the link has `tc30` instead of `tc39`

test/built-ins/Promise/withResolvers/result.js
```diff
@@ -0,0 +1,19 @@
...
+assert(instance intanceof Object);
```
trflynn89, line 15: Typo here as well: `intanceof` should be `instanceof`.

ljharb: whoops
```suggestion
assert(instance instanceof Object);
```

features.txt
ljharb:
```suggestion
# https://github.com/tc39/proposal-promise-with-resolvers
```

test/built-ins/Promise/withResolvers/ctx-non-ctor.js
```diff
@@ -0,0 +1,12 @@
+// Copyright (C) 2023 Peter Klecha. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Promise.withResolvers errors when the reciever is not a constructor
```
Ms2ger, line 5:
```suggestion
description: Promise.withResolvers errors when the receiver is not a constructor
```

test/built-ins/Promise/withResolvers/ctx-ctor.js
```diff
@@ -0,0 +1,16 @@
+// Copyright (C) 2023 Peter Klecha. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Promise.withResolvers produces instances of the reciever
```
linusg, line 5:
```suggestion
description: Promise.withResolvers produces instances of the receiver
```

test/built-ins/Promise/withResolvers/ctx-non-object.js
```diff
@@ -0,0 +1,32 @@
...
+    Promise.withResolvers.call('string');
+  });
+  
+  assert.throws(TypeError, function() {
+    Promise.withResolvers.call(true);
+  });
+  
+  assert.throws(TypeError, function() {
+    Promise.withResolvers.call(Symbol());
+  });
```
linusg, line 32: indentation is a little off here :)

test/built-ins/Promise/withResolvers/result.js
```diff
@@ -0,0 +1,19 @@
...
+assert(Object.hasOwn(instance, "promise"));
+assert(Object.hasOwn(instance, "resolve"));
+assert(Object.hasOwn(instance, "reject"));
```
linusg, line 18: Would be nice to also test the property attributes here, using `verifyProperty()` from `propertyHelper.js`:

https://github.com/tc39/test262/blob/9d7eb8018e3b74ae9bcb12fbe8db0dfce4ef6adf/test/built-ins/BigInt/length.js#L27-L32

linusg: (you can omit `value` as it's already tested elsewhere - https://github.com/tc39/test262/blob/9d7eb8018e3b74ae9bcb12fbe8db0dfce4ef6adf/harness/propertyHelper.js#L67-L71)

ptomato:
Thanks for contributing this! Would you mind opening a tracking issue? An example is #3756. You don't have to write the whole testing plan, someone else can help with that, but we should use one to verify that we're covering everything in the proposal.

I haven't looked too closely at this PR yet so I might be missing something, but it seems like there are some cases particularly with subclassing `Promise` that aren't yet tested.

We also need to add the standard boilerplate tests for built-in APIs. For an example, see `length.js`, `name.js`, `prop-desc.js`, `not-a-constructor.js`, `builtin.js` in https://github.com/tc39/test262/tree/main/test/built-ins/Array/fromAsync.

We don't have to block reviewing and merging this PR on complete coverage, the other tests can come in following PRs, but we should keep track in the tracking issue of what's been merged.

syg:
Could we have this merged soon for shipping in Chrome?

peetklecha:
sorry for the delay with this.

- i've now addressed all the comments i've recieved
- there's a failing test from circle/ci which wasn't there before and which i don't understand, any help with it would be appreciated
- per @ptomato 's request i've created an issue to track full test coverage of Promise.withResolvers: #3910

ptomato:
Thanks, @peetklecha. The failing CI job is #3891, we really should remove the job until that's fixed. I'll merge this now, and add some suggestions to the tracking issue you opened.

PR #3868, opened 2023-07-08
Replace checks in old tests with assert.sameValue

CanadaHonk:
Many old tests use custom checks with if and throwing a custom error instead of using `assert.sameValue`, this replaces most cases.


ptomato:
@CanadaHonk, this is amazing! Thanks. I think it'll be quite difficult to review effectively by hand, so my preference would be that we just rely on the ESMeta checks, and possibly ask an implementation to run the test suite from your branch. I'm curious what the other maintainers think though.

It looks like there are a few failures in the ESMeta job in the last run, so would you be able to take a look at those before proceeding?

linusg (changes requested):
I ran test262 with LibJS against this branch and got the following results:

```
Diff Tests:
    test/built-ins/Array/prototype/sort/bug_596_2.js       ✅ -> ❌
    test/language/expressions/unary-minus/S11.4.7_A3_T1.js ✅ -> ❌
    test/language/expressions/unary-minus/S11.4.7_A3_T4.js ✅ -> ❌
    test/language/line-terminators/between-tokens-cr.js    ✅ -> ❌
    test/language/line-terminators/comment-multi-cr.js     ✅ -> ❌
```

- `test/built-ins/Array/prototype/sort/bug_596_2.js` has a syntax error (`assert.sameValue(array.hasOwnProperty('0'), true, '#2: array.hasOwnProperty('0')');`)
- `test/language/expressions/unary-minus/S11.4.7_A3_T1.js` and `test/language/expressions/unary-minus/S11.4.7_A3_T4.js` compare `-0` with `0` which worked before with `===` but no longer does with `assert.sameValue()`
- `test/language/line-terminators/between-tokens-cr.js`and `test/language/line-terminators/comment-multi-cr.js` got their line terminators messed up

CanadaHonk:
Yeah, this still has some issues with a few tests hence draft ;) Will fix today. Thanks both for feedback :)

CanadaHonk:
Fixed comments from Linus (thanks for details), will check results of engines with new changes.

ptomato:
@CanadaHonk It's been a while, so sorry for the delay, but could you check the ESMeta results again? Since this is a long and repetitive PR we should make sure that we're using all of the possibilities for automation that are at our disposal!

Here are the failures reported by the job:
```
[Test262Error] 9 tests are failed:
built-ins/String/S15.5.1.1_A1_T16.js - Invalid operation
built-ins/String/S15.5.1.1_A1_T17.js - Invalid operation
built-ins/String/S9.8.1_A10.js - Invalid operation
built-ins/String/S9.8.1_A7.js - Invalid operation
built-ins/String/S9.8.1_A8.js - Invalid operation
built-ins/String/S9.8.1_A9_T2.js - Invalid operation
built-ins/String/prototype/lastIndexOf/S15.5.4.8_A1_T7.js - [InterpreterError] return not undefined: comp[~throw~](#2203)
built-ins/String/prototype/slice/S15.5.4.13_A1_T15.js - Invalid operation
built-ins/String/prototype/substring/S15.5.4.15_A1_T15.js - Invalid operation
```

PR #3798, opened 2023-03-21
Split DurationFormat style tests

FrankYFTang:
For better parallization and reporting
Mostly split the tests only. But also remove the output of unit for ":" in "digital" style. 

@sffc @ryzokuken @romulocintra

PR #3771, opened 2023-01-18
Coverage DurationFormat formatToParts

romulocintra:
Test formatToParts output https://github.com/tc39/proposal-intl-duration-format/pull/126

PR #3870, opened 2023-07-13
Correct DurationFormat test

FrankYFTang:
The expectation values for the formatToParts are incorrect.

A. The listFormat is consturcted with
  6. Perform ! CreateDataPropertyOrThrow(lfOpts, "type", "unit"). so it should not have " and " in the string. Comparing with (new Intl.ListFormat("en", {type: "unit"})).formatToParts(["a", "b", "c"]) and value in the format d and the expected values in the DurationFormat/prototype/format directory B. The localized unit values for "short" (and therefore default)
   are incorrect. Comparing with the expected values in the
  DurationFormat/prototype/format directory


FrankYFTang:
@ryzokuken @romulocintra @ptomato  @sffc 
These tests were added in https://github.com/tc39/test262/pull/3771 and I split them into inidivisual one in 

https://github.com/tc39/test262/pull/3798

but the expected values were set incorrectly in 3771

PR #3872, opened 2023-07-17
Move all tests using import assertions to files named as such

nicolo-ribaudo:
Extracted from https://github.com/tc39/test262/pull/3843. I need this commit as a common base for two PR:
- add import attributes tests
- update import assertions tests that are not valid anymore


nicolo-ribaudo:
CI is failing because gralljs 23 changed the name of their release artifacts (now the folder is prefixed as `graal-23.0.0` and not `vm-23.0.0`), and esvu does not support it yet.

https://github.com/oracle/graaljs/releases/tag/graal-23.0.0

linusg:
GraalJS should not even run as part of CI anymore (https://github.com/tc39/test262/commit/a85807714b9ad57da896be8fcc782bfc6d443ad0), you might need to rebase the PR

ptomato (approved):
I had a question about some of the tests that look like they would be common to both import assertions and import attributes, such as the trailing comma in `import('foo',)`. @nicolo-ribaudo explained to me that we'll get duplicate tests for these (1) because they are generated anyway and (2) because there is no way to express "import-assertions OR import-attributes" in the feature flags.

Since this is blocking other PRs, I'll go ahead and merge it.

PR #3874, opened 2023-07-18
Add tests for AsyncContext (Stage 2)

andreubotella:
See https://github.com/tc39/proposal-async-context.

### DIFF
--- a/test/built-ins/AsyncContext/Variable/prototype/run/restores-context-snapshot-after-callback.js
+++ b/test/built-ins/AsyncContext/Variable/prototype/run/restores-context-snapshot-after-callback.js
@@ -1,0 +1,46 @@
+// Copyright (C) 2023 Igalia, S. L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-asynccontext-variable.prototype.run
+description: >
+  Switches back into the previous context snapshot after calling the callback.
+info: |
+  AsyncContext.Variable.prototype.run ( value, func, ...args )
+
+  3. Let previousContextMapping be AsyncContextSnapshot().
+  ...
+  10. Let result be Completion(Call(func, undefined, args)).
+  11. AsyncContextSwap(previousContextMapping).
+  12. Return result.
+
+  AsyncContextSwap ( snapshotMapping )
+
+  1. Let agentRecord be the surrounding agent's Agent Record.
+  ...
+  3. Set agentRecord.[[AsyncContextMapping]] to snapshotMapping.
+
+  AsyncContextSnapshot ( )
+
+  1. Let agentRecord be the surrounding agent's Agent Record.
+  2. Return agentRecord.[[AsyncContextMapping]].
+features: [AsyncContext]
+---*/
+
+const asyncVar = new AsyncContext.Variable();
+
+asyncVar.run("foo", () => {
+  asyncVar.run("42", () => { });
+
+  assert.sameValue(
+    asyncVar.get(),
+    "foo",
+    'The value of `asyncVar.get()` inside the first `run` and after the second one should be `"foo"`'
+  );
+});
+
+assert.sameValue(
+  asyncVar.get(),
+  undefined,
+  'The value of `asyncVar.get()` after the first `run` should be `undefined`'
+);
--- a/test/built-ins/AsyncContext/context-propagation/generators.js
+++ b/test/built-ins/AsyncContext/context-propagation/generators.js
@@ -1,0 +1,35 @@
+// Copyright (C) 2023 Igalia, S. L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+flags: [async]
+features: [AsyncContext]
+includes: [asyncHelpers.js]
+---*/
+
+// TODO: This test tests the agreed behavior in
+// https://github.com/tc39/proposal-async-context/issues/18, but the proposed
+// spec text does not yet implement that behavior.
+
+const context = new AsyncContext.Variable();
+
+async function* gen(context) {
+  await Promise.resolve();
+  assert.sameValue(context.get(), "init");
+  yield;
+  await Promise.resolve();
+  assert.sameValue(context.get(), "init");
+}
+
+asyncTest(async () => {
+  let g;
+  context.run('init', () => {
+    g = gen(context);
+  });
+  await context.run('first iter', async () => {
+    await g.next();
+  });
+  await context.run('second iter', async () => {
+    await g.next();
+  });
+});
--- a/test/built-ins/AsyncContext/Snapshot/constructor.js
+++ b/test/built-ins/AsyncContext/Snapshot/constructor.js
@@ -1,0 +1,16 @@
+// Copyright (C) 2023 Igalia, S. L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-asynccontext-snapshot-constructor
+description: >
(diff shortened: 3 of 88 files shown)


andreubotella:
The tests in `built-ins/AsyncContext/context-propagation/*` are really tests that the changes to the `HostMakeJobCallback` and `HostCallJobCallback` host hooks are indeed reflected in the built-in APIs that use those hooks. Maybe they should be moved to those APIs' folders.

Also, `built-ins/AsyncContext/context-propagation/generators.js` currently tests a behavior that was agreed to in tc39/proposal-async-context#18 but that hasn't yet been changed in the spec text.

test/built-ins/AsyncContext/context-propagation/generators-yield-inside-run.js
```diff
@@ -0,0 +1,33 @@
...
+features: [AsyncContext]
+---*/
+
+// TODO: This test tests the behavior in
+// https://github.com/tc39/proposal-async-context/pull/61
+
+const asyncVar = new AsyncContext.Variable();
+
+function* gen() {
+  assert.sameValue(asyncVar.get(), "init");
```
caitp, line 15: This assert is not reachable in the "init" run, because the generator is in the SUSPENDED-START state until the "first-iter" run.

We can either remove it, or add a generator.next() call to the "init" run invocation.

test/built-ins/AsyncContext/context-propagation/generators.js
```diff
@@ -0,0 +1,29 @@
...
+features: [AsyncContext]
+---*/
+
+// TODO: This test tests the behavior in
+// https://github.com/tc39/proposal-async-context/pull/61
+
+const asyncVar = new AsyncContext.Variable();
+
+function* gen() {
+  assert.sameValue(asyncVar.get(), "init");
```
caitp, line 15: ditto

test/built-ins/AsyncContext/context-propagation/async-generators-yield-inside-run.js
```diff
@@ -0,0 +1,43 @@
...
+---*/
+
+// TODO: This test tests the behavior in
+// https://github.com/tc39/proposal-async-context/pull/61
+
+const asyncVar = new AsyncContext.Variable();
+
+async function* gen() {
+  await Promise.resolve();
+  assert.sameValue(asyncVar.get(), "init");
```
caitp, line 18: Same here, async generators are initially in a SUSPEND-START state, so we need to resume the generator with .next() before the first await or this assert can be reached.

test/built-ins/AsyncContext/context-propagation/async-generators.js
```diff
@@ -0,0 +1,35 @@
...
+---*/
+
+// TODO: This test tests the behavior in
+// https://github.com/tc39/proposal-async-context/pull/61
+
+const asyncVar = new AsyncContext.Variable();
+
+async function* gen() {
+  await Promise.resolve();
+  assert.sameValue(asyncVar.get(), "init");
```
caitp, line 18: ditto

caitp (commented):
disclaimer: I _may_ be missing some adjustments to generators in https://tc39.es/proposal-async-context/, but at a glance I wasn't able to find anything yet.

test/built-ins/AsyncContext/context-propagation/generators.js
caitp: hmm. maybe I'm wrong here

jridgewell (commented):
I've only reviewed the Snapshot cases so far, but a few more cases we could add:

- `snapshot.run()` uses `undefined` this context
- `snapshot.run()` restores previous context after abrupt throw

test/built-ins/AsyncContext/Variable/prototype/run/restores-context-snapshot-after-callback.js
jridgewell, line 1: I know this is dumb, but can you add a test for run restoring the state after a snapshot has been taken? It's one of the bugs that can happen with a forking snapshot design as in https://github.com/tc39/proposal-async-context/tree/master/src:

```js
const asyncVar = new AsyncContext.Variable();

asyncVar.run("foo", () => {
  asyncVar.run("42", () => {
    // Taking a snapshot forces a fork when we return, the snapshot makes the AC immutable.
    new AsyncContext.Snapshot();
  });

  assert.sameValue(
    asyncVar.get(),
    "foo",
    'The value of `asyncVar.get()` inside the first `run` and after the second one should be `"foo"`'
  );
});

assert.sameValue(
  asyncVar.get(),
  undefined,
  'The value of `asyncVar.get()` after the first `run` should be `undefined`'
);
```

jridgewell, line 1: Another variant of this is to ensure that the `run` does not wait for a return promise to resolve (a bug Vercel had in a prototype PR):

```js
asyncVar.run("foo", () => {
  asyncVar.run("42", () => {
    return new Promise(() => {});
  });

  assert.sameValue(
    asyncVar.get(),
    "foo",
    'The value of `asyncVar.get()` inside the first `run` and after the second one should be `"foo"`'
  );
});


assert.sameValue(
  asyncVar.get(),
  undefined,
  'The value of `asyncVar.get()` after the first `run` should be `undefined`'
);
```

andreubotella: I wonder if the bug was that the `run` implementation assumed that if the callback returns a promise, that promise must be awaited in the caller?

jridgewell: Basically. But not every promise will be awaited by the caller.

```js
class Variable {
  async run(val, fn) {
    const before = __storage__;
    try {
      __storage__ = new Map(before);
      __storage__.set(this, val);
      return await fn();
    } finally {
      __storage__ = before;
    }
  }
}
```

The `await fn()` there prevented the finally block from running, meaning code synchronously run after `run()` call in the caller inherited the ambient state.

test/built-ins/AsyncContext/Variable/proc-desc.js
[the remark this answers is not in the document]
andreubotella: My bad, copy-paste failure. As I was fixing this, I also copied some related tests from elsewhere.

jridgewell (approved):

test/built-ins/AsyncContext/Snapshot/wrap/wrapped-function/length.js
```diff
@@ -0,0 +1,83 @@
...
+
+setAndAssertLength(Number.NaN, 0);
+
+setAndAssertLength(7.2, 7);
+
+setAndAssertLength(7.9, 7);
+
+setAndAssertLength(undefined, 0);
+
+setAndAssertLength("42", 42);
```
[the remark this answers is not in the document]
andreubotella: I don't really care one way or the other, but `Function.prototype.bind` checks whether the length is a number before calling `ToIntegerOrInfinity()`, and if it isn't one it uses 0 rather than converting to a number.

Also, now that I look at it in some more detail, I realize that `bind` also checks `HasOwnProperty` before doing `Get`, whereas the `wrap` spec only does `Get`, so this might affect proxies. When we settle on this (one way or the other), we should also test it.

In any case, spec-wise the ShadowRealm proposal adds a [`CopyNameAndLength`](https://tc39.es/proposal-shadowrealm/#sec-copynameandlength) AO to share the behavior between `bind` and ShadowRealm wrapped functions. If we decide to have the same behavior as `bind`, it'd be better to just use that if we can.

PR #3877, opened 2023-07-21
Temporal: Test rejecting ISO strings with subminute offsets

justingrant:
This PR verifies that ISO strings with sub-minute offsets cannot be parsed into time zone identifiers. This was a change introduced in the recently-merged tc39/proposal-temporal#2607, but tests for this case were missing from #3862 (the tests for that PR).

I noticed in codecov results on an unrelated PR that this case wasn't being tested, so fixing that mistake now.


Ms2ger (approved):

ptomato (approved):
Thanks!

PR #3881, opened 2023-07-22
Test that non-string Temporal.TimeZone IDs will throw

justingrant:
This PR fills a test hole found by Codecov: we weren't checking for the case of a custom time zone whose `id` property returns a non-string value.

Note that the `id` property is only read in a few places, all on the ZonedDateTime prototype: `until`, `since`, `equals`, `toString`, `toLocaleString`, `toJSON`, and `timeZoneId`.


Ms2ger (approved):
Thank you!

justingrant:
> Note that the same test gap seems to exist for custom calendars too, so a future PR will be needed to fill that gap.

FYI @ptomato: there may be a test gap for non-string custom calendar IDs.  Where is the best place to record that this gap exists?  Issue here in Test262 repo?  In proposal-temporal repo? Somewhere else? I probably won't be able to PR that, but at least wanted to record that the gap (probably) exists.

Ms2ger:
I created one here

PR #3882, opened 2023-07-24
use proper isConstructor check for recently added ArrayBuffer/SharedArrayBuffer tests

ljharb:
This also allows these tests to be filtered out by "features".


Ms2ger (changes requested):
Adding the `isConstructor` checks is fine, but could you explain why you're removing the `'prototype'` checks? They don't seem to be obviously superseded by `isConstructor`.

ljharb:
They’re not particularly useful; isConstructor is sufficient, and we don’t check the prototype anywhere else.

Ms2ger (approved):

PR #2669, opened 2020-06-24
Address various test bugs/issues

anba:
b78c6d24be0ba1bfc3a9903d59ccf59b194dce2a
- Per the latest changes to the PR, change the expected value to `undefined` when no fractional second digits are present.
- https://github.com/tc39/ecma402/pull/347#issuecomment-516380558

02ab047287227078ec560a9ea07e3b796a320a68
- See https://github.com/tc39/ecma402/pull/418/commits/e905e0fca2bd4a1dbdb78f5b59da6c6cc9da4aee

3921e70041490f21fdf537fd60a635e95eeb5caf
a77c4564679eab7e25fccf7043208ff55a9a96dd
- See https://github.com/tc39/proposal-promise-any/commit/e4efffb52c35c3b14b39fa9cde3fa30afcbc543d

75aad9c8d88022389080c26972b9fb74c6f6fe5c
- https://github.com/tc39/ecma402/pull/438

PR #3883, opened 2023-07-25
Normative: Add new numbering system "kawi" and "nagm"

FrankYFTang:
Adding tests for https://github.com/tc39/ecma402/pull/714


Ms2ger:
Is this something that needs a feature flag? I suppose that would be painful.

anba:
We didn't add any feature flags for past updates to the list of numbering systems: #3572, #2669, #2554

Ms2ger:
Sounds good. @anba, do you want to check the code points? :)

anba:
All code points seem to match:
- <https://github.com/unicode-org/cldr/blob/706d94d95c855bb281a1829488c511f99e689ecf/common/supplemental/numberingSystems.xml#L49>
- and <https://github.com/unicode-org/cldr/blob/706d94d95c855bb281a1829488c511f99e689ecf/common/supplemental/numberingSystems.xml#L71>

Ms2ger (approved):

PR #3892, opened 2023-08-10
Test more ISO strings with sub-minute offsets

justingrant:
Extends the tests in https://github.com/tc39/test262/pull/3877 with more comprehensive tests of sub-minute offsets in ISO strings, including trailing zeroes and sub-second offsets.


justingrant:
The failure above seems to be a build error for the XS engine, not anything related to this PR.

> /home/circleci/.esvu/bin/xs: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by /home/circleci/.esvu/bin/xs)

Ms2ger (approved):

PR #3893, opened 2023-08-10
Disallow sub-minute annotations in Instant ISO strings

justingrant:
Adds tests that verify that Instant ISO strings are syntactically invalid if they include offset time zone annotations with sub-minute units.

Also removes a few staging test cases that, due to a polyfill bug, were incorrectly passing.


justingrant:
The CI failure with the XS engine is a build problem unrelated to this PR.

Ms2ger (approved):

PR #3894, opened 2023-08-11
Tests for audit of Temporal user code calls, part 1

ptomato:
tc39/proposal-temporal#2519 is a very large pull request and I'm planning to land it in parts. This PR contains tests for the first part. Each commit here corresponds to the same-named commit in the proposal-temporal normative PR.


test/built-ins/Temporal/PlainYearMonth/from/observable-get-overflow-argument-string.js
```diff
@@ -0,0 +1,34 @@
...
+TemporalHelpers.assertPlainYearMonth(result, 2021, 5, "M05");
+
+actual.splice(0);  // empty it for the next check
+const failureExpected = [
+  "ownKeys options",
+  "getOwnPropertyDescriptor options.overflow",
+  "get options.overflow",
+];
+
+assert.throws(TypeError, () => Temporal.PlainYearMonth.from(7, options));
```
Ms2ger, line 33: For all of these, perhaps `-argument-primitive.js` is a better name

test/built-ins/Temporal/PlainYearMonth/prototype/with/calendar-arguments.js
```diff
@@ -12,15 +12,18 @@ includes: [temporalHelpers.js]
...
 class CustomCalendar extends Temporal.Calendar {
   constructor() {
     super("iso8601");
   }
   yearMonthFromFields(...args) {
     assert.sameValue(args.length, 2, "args.length");
     assert.sameValue(typeof args[0], "object", "args[0]");
-    assert.sameValue(args[1], options, "args[1]");
+    assert.notSameValue(args[1], options, "args[1] is a copy of options");
+    assert.sameValue(args[1].extra, "property", "All properties are copied");
```
Ms2ger, line 26: Consider whether you want to verify the prototype for these

ptomato: Added, and found that ZonedDateTime didn't have a test like this one yet, so added that as well.

ptomato:
Thanks for the review!

PR #3897, opened 2023-08-18
Tests for audit of Temporal user code calls, part 2

ptomato:
Continuing on from #3894, https://github.com/tc39/proposal-temporal/pull/2519 is a very large pull request and I'm planning to land it in parts. This PR contains tests for the second part. Each commit here corresponds to the same-named commit in the proposal-temporal normative PR.

Recommend reviewing commit-by-commit. Note, in the commit titled "Temporal: Don't observably iterate array in built-in calendar's fields()" there are a large number of tests that are basically copy-pasted except for one, as noted in the commit message, so double-check that one when reviewing.

### DIFF
--- a/harness/temporalHelpers.js
+++ b/harness/temporalHelpers.js
@@ -1948,6 +1948,30 @@ var TemporalHelpers = {
   },
 
   /*
+   * A custom time zone that does not allow any of its methods to be called, for
+   * the purpose of asserting that a particular operation does not call into
+   * user code.
+   */
+  timeZoneThrowEverything() {
+    class TimeZoneThrowEverything extends Temporal.TimeZone {
+      constructor() {
+        super("UTC");
+      }
+      getOffsetNanosecondsFor() {
+        TemporalHelpers.assertUnreachable("getOffsetNanosecondsFor should not be called");
+      }
+      getPossibleInstantsFor() {
+        TemporalHelpers.assertUnreachable("getPossibleInstantsFor should not be called");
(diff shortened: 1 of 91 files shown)


cjtenny:
Looks great to me. Covers all the different entry points for each operation. Really thorough - discussed some feedback directly while working through the commits, and you fixed the spec bug I found in the no-op round operations before I got to comment it on the review :P only extant feedback:

For the tests in 'Don't observably iterate array in built-in calendar's fields()', do we have any tests already that cover that an otherwise unmodified `Temporal.Calendar.from('iso8601')` in the same use case *will* cause `Array.prototype[Symbol.iterator]` to be invoked?

ptomato:
> For the tests in 'Don't observably iterate array in built-in calendar's fields()', do we have any tests already that cover that an otherwise unmodified `Temporal.Calendar.from('iso8601')` in the same use case _will_ cause `Array.prototype[Symbol.iterator]` to be invoked?

I made various modifications to the reference polyfill and re-ran the tests, to check this. I think the answer is, we do not have any tests for this directly. (If you modify CalendarFields to skip the iteration when _calendar_ has the correct internal slot and Get(_calendar_, **"fields"**) is the same value as %Temporal.Calendar.prototype.fields%, then no tests fail.) However, we do have tests that will fail if an implementation makes the mistake that would generally be the cause of this situation, that is, treating an unmodified instance of a built-in calendar the same as a string calendar.

PR #3898, opened 2023-08-21
verify that non-integral values throw

ben-allen:
Update to reflect recent normative change https://github.com/tc39/proposal-intl-duration-format/pull/158 making ToIntegerWithoutRounding throw on NaN instead of treating it as 0.


ben-allen:
closed: realized it's redundant

PR #3901, opened 2023-08-22
Test coverage for Intl.DisplayNames.prototype.of()

ben-allen:
The .of() method previously only had tests for the types "language", "calendar", and "dateTimeField". Added tests for "region", "script", "currency".


ptomato:
@ben-allen What is keeping this a draft?

PR #3902, opened 2023-08-22
Accept "mo" narrow style for month

Constellation:
macOS system ICU is shipping new CLDR, but it has many overrides on the top of it to make the formatted output suitable for the system. For narrow "month" representation, CLDR uses "m", but AppleICU uses "mo" to distinguish between "month" and "minute". This patch accepts this since it is done based on different readability decision, and ECMA-402 allows it.


Constellation:
@ptomato

ljharb (approved):

anba:
#3903 fixed this issue in a different way by computing the expected pattern. So this PR is actually no longer needed.

ptomato:
OK, thanks!

PR #3903, opened 2023-08-23
Improve Intl.DurationFormat tests

anba:
d26a15712fd1e06e7496d640e838a447fb2b5b78
- Adds a simplified implementation of [PartitionDurationFormatPattern](https://tc39.es/proposal-intl-duration-format/#sec-partitiondurationformatpattern) to generate the expected results
- The implementation is partially based on the updates from <https://github.com/tc39/proposal-intl-duration-format/pull/166> and the [SpiderMonkey patches](https://bugzilla.mozilla.org/show_bug.cgi?id=1648139)

275c0cabfa4d2453cc02366eeec88baf7b51524c
- `Intl.DurationFormat` currently requires exact mathematical computations. Test for this until there's an agreement to change this.
- Also see <https://github.com/tc39/proposal-intl-duration-format/issues/157>.

2e7a522c372a08d8a54b950a0ec2358af100f917
- In <https://github.com/tc39/proposal-intl-duration-format/pull/166> I've mentioned a strange formatting case. This adds a test case for it.

89ecd6d12ec0ed53dec3599538037600df20f831
- Add missing coverage for negative durations.
- Also see: https://github.com/tc39/proposal-intl-duration-format/issues/169


ben-allen:
Can I just say that this is truly incredible? Thank you so much for all your work here!

harness/testIntl.js
```diff
@@ -2502,3 +2504,212 @@ function isCanonicalizedStructurallyValidTimeZoneName(timeZone) {
...
+        };
+      }
+
+      let nf = new Intl.NumberFormat(locale, nfOpts);
+      let formatted = nf.formatToParts(value);
+
+      // Add |numberFormatUnit| to the formatted number.
+      let list = [];
+      for (let {value, type} of formatted) {
+        list.push({value, type, unit: numberFormatUnit});
```
FrankYFTang, line 2645: The order stated in the proposal is {type, value, unit} not {value, type, unit}

anba: Good catch!

anba: Fixed.

FrankYFTang (approved):
LGTM thanks for working on this.

Ms2ger (approved):

PR #3904, opened 2023-08-23
testing for invalid combinations of numeric and non-numeric styles

ben-allen:


ptomato:
@ben-allen What is keeping this a draft? (Also is there an issue in the bug tracker that gives context?)

PR #3905, opened 2023-08-23
add test for regexp matching with zero-length quantified matches

bakkot:
This behavior is pretty subtle, and apparently JSC's two regexp engines are inconsistent here. So it seems worth testing.


msaboff (approved):
LGTM

Thanks.

Ms2ger (approved):

PR #3906, opened 2023-08-24
update iterator helpers tests that passed even in noncompliant impls

michaelficarra:
These were the only 2 iterator helper tests that were passing even in implementations that had not yet added support for iterator helpers.


ljharb (approved):

ptomato (approved):
Good catch.

PR #3907, opened 2023-08-30
Fixup class names in class decorator private identifier tests.

dminor:
For the statement level test, the inner class name is not initialized at the time the decorators evaluate, resulting in a ReferenceError that the declaration can not be accessed prior to initialization.

Similar, non-decorator code, like:
    class C {
        static dec() {}
        static {
            this.x = C.dec();
            class C {}
        }
    }
also results in a ReferenceError.

For the expression level test, the var C is undefined at the time the decorators are evaluated, resulting in TypeError while trying to access a member of undefined.

Similar, non-decorator code, like:
    var C = class {
        static f() {};
        static {
            this.x = C.f();
        }
    }
also results in a TypeError.

### DIFF
--- a/src/decorator/syntax/class-valid/cls-decl-decorators-valid-syntax.template
+++ b/src/decorator/syntax/class-valid/cls-decl-decorators-valid-syntax.template
@@ -25,6 +25,6 @@ features: [class, decorators]
 class C {
   /*{ elements }*/
   static {
-    /*{ decorators }*/ class C {}
+    /*{ decorators }*/ class D {}
   }
 }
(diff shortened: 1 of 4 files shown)


ptomato:
Hi @dminor, thanks for the fix. These tests are autogenerated and so I'd recommend changing the test templates that they come from, instead:
- `src/decorator/syntax/class-valid/cls-decl-decorators-valid-syntax.template`
- `src/decorator/syntax/class-valid/cls-expr-decorators-valid-syntax.template`

Ms2ger:
I'm guessing Philip's "instead" was somewhat confusing - you'll *also* need to commit the changes to the generated tests

ptomato:
I agreed to review this. Sorry it took so long to get to it. I expect to have time this week.

ptomato (approved):
Makes sense, thanks for the fix and for the patient wait for review.

PR #3908, opened 2023-09-05
Remove pointless 'string check' from isEnumerable() property helper

linusg:
As far as I can tell this does nothing that `hasOwnProperty()` doesn't already take care of. This is essentially a revert of 866d7f8d8e4601618d4a94fb8fa7b4219c69bcf6, the purpose of which I don't understand.

### DIFF
--- a/harness/propertyHelper.js
+++ b/harness/propertyHelper.js
@@ -113,23 +113,10 @@ function isConfigurable(obj, name) {
 }
 
 function isEnumerable(obj, name) {
-  var stringCheck = false;
-
-  if (typeof name === "string") {
-    for (var x in obj) {
-      if (x === name) {
-        stringCheck = true;
-        break;
-      }
-    }
-  } else {
-    // skip it if name is not string, works for Symbol names.
-    stringCheck = true;
-  }
-
-  return stringCheck &&
+  return (
     Object.prototype.hasOwnProperty.call(obj, name) &&
-    Object.prototype.propertyIsEnumerable.call(obj, name);
+    Object.prototype.propertyIsEnumerable.call(obj, name)
+  );
 }
 
 function isSameValue(a, b) {


linusg:
CI failure unrelated (`/home/circleci/.esvu/bin/xs: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by /home/circleci/.esvu/bin/xs)`)

anba:
Implementations often have different code paths for `for-in` when compared to `Object.prototype.propertyIsEnumerable`. 

For similar reasons `isConfigurable` uses `delete` instead of just calling `Object.getOwnPropertyDescriptor` and `isWritable` performs `obj[name] = newValue`.

linusg:
Of course, but can the `for-in` code path not be sufficiently tested separately? I'm unsure why it needs to be run as part of this helper function, assuming a working implementation of `hasOwnProperty` and `propertyIsEnumerable`.

anba:
The same is true for using `delete` in `isConfigurable`: Assuming a working implementation of `Object.getOwnPropertyDescriptor`, there's no need to actually modify the input object through `delete`. 

But past experience has shown that implementations don't always implement every object internal methods correctly, especially for exotic objects like arguments objects or typed arrays. Therefore the property helper functions are testing both introspection methods like `Object.getOwnPropertyDescriptor` and actual operations like `delete` or `for-in`.

linusg:
I see, thanks for the explanation! That leaves me in a tricky situation, my actual motivation for this PR is similar to this one: https://github.com/tc39/test262/pull/3845
The one `for-in` loop causing the entire harness file to not parse in engines that don't implement them (yet) is how I found it in the first place, which in turn breaks all 5.1k tests including the helper.

Let's leave it up to the maintainers to decide what to do with this :)

anba:
Work-in-progress implementations which don't implement `for-in`, but still like to run test262 should probably just provide their own implementation of the `isConfigurable` function. Either by using a modified "harness/propertyHelper.js" file or alternatively executing `const isConfigurable = function() { ... }` before loading "harness/propertyHelper.js" (if the implementation happens to implement `const` already). I've used both approaches in the past when working on <https://github.com/anba/es6draft> during the development of the ES6 spec.

ljharb:
It’s fine to prefer avoiding arrows, because there was a time when arrows didn’t exist. I’m not aware of any time where for-in didn’t exist.

Ms2ger:
I would be inclined against merging this for the reasons upthread.

PR #3912, opened 2023-09-09
Add tests for DurationFormat PR 172 and 167

FrankYFTang:
Test the order of resolvedOptions in default setting Also add test to check the reading order of options based on the property returned in the default setting resolvedOptions.

https://github.com/tc39/proposal-intl-duration-format/pull/167 

https://github.com/tc39/proposal-intl-duration-format/pull/172 

https://tc39.es/proposal-intl-duration-format/#sec-Intl.DurationFormat.prototype.resolvedOptions

These two PRs are presenting to TC39 2023-09 meeting

### DIFF
--- a/test/intl402/DurationFormat/constructor-option-read-order.js
+++ b/test/intl402/DurationFormat/constructor-option-read-order.js
@@ -1,0 +1,47 @@
+// Copyright 2023 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-Intl.DurationFormat
+description: Checks the order of option read.
+features: [Intl.DurationFormat]
+includes: [compareArray.js]
+---*/
+
+let optionKeys =  Object.keys((new Intl.DurationFormat()).resolvedOptions());
+let opt = {};
+let readKeys = new Array();
+// For each item returned by resolvedOptions of default, add a getter
+// to track the reading order.
+optionKeys.forEach((property) =>
+    Object.defineProperty(opt, property, {
+        get() {
+            readKeys[readKeys.length] = property;
+            return undefined;
+        },
+    }));
+let p = new Intl.DurationFormat(undefined, opt);
+assert.compareArray(
+    readKeys,
+    ['numberingSystem',
+    'style',
+    'years',
+    'yearsDisplay',
+    'months',
+    'monthsDisplay',
+    'weeks',
+    'weeksDisplay',
+    'days',
+    'daysDisplay',
+    'hours',
+    'hoursDisplay',
+    'minutes',
+    'minutesDisplay',
+    'seconds',
+    'secondsDisplay',
+    'milliseconds',
+    'millisecondsDisplay',
+    'microseconds',
+    'microsecondsDisplay',
+    'nanoseconds',
+    'nanosecondsDisplay']);
--- a/test/intl402/DurationFormat/prototype/resolvedOptions/return-keys-order-default.js
+++ b/test/intl402/DurationFormat/prototype/resolvedOptions/return-keys-order-default.js
@@ -1,0 +1,34 @@
+// Copyright 2023 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-intl.durationformat.prototype.resolvedoptions
+description: order of property keys for the object returned by resolvedOptions()
+features: [Intl.DurationFormat]
+includes: [compareArray.js]
+---*/
+
+assert.compareArray(
+    Object.keys((new Intl.DurationFormat()).resolvedOptions()),
+    ['locale',
+    'numberingSystem',
+    'style',
+    'years',
+    'yearsDisplay',
+    'months',
+    'monthsDisplay',
+    'weeks',
+    'weeksDisplay',
+    'days',
+    'daysDisplay',
+    'hours',
+    'hoursDisplay',
+    'minutes',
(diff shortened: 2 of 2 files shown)


test/intl402/DurationFormat/constructor-option-read-order.js
```diff
@@ -0,0 +1,46 @@
+/// Copyright 2023 the V8 project authors. All rights reserved.
```
anba, line 1:
```suggestion
// Copyright 2023 the V8 project authors. All rights reserved.
```

anba: #3979 recently fixed some triple leading slashes, so maybe there are some tools which don't handle this case correctly.

test/intl402/DurationFormat/constructor-option-read-order.js
```diff
@@ -0,0 +1,46 @@
...
+// to track the reading order.
+optionKeys.forEach((property) =>
+    Object.defineProperty(opt, property, {
+        get() {
+            readKeys[readKeys.length] = property;
+            return undefined;
+        },
+    }));
+let p = new Intl.DurationFormat(undefined, opt);
+assert.sameValue(
```
anba, line 23: Nit: This could use `assert.compareArray`. (Requires `includes: [compareArray.js]` in the frontmatter comment.)

test/intl402/DurationFormat/prototype/resolvedOptions/return-keys-order-default.js
```diff
@@ -0,0 +1,33 @@
+// Copyright 2023 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-intl.durationformat.prototype.resolvedoptions
+description: order of property keys for the object returned by resolvedOptions()
+features: [Intl.DurationFormat]
+---*/
+
+assert.sameValue(
```
anba, line 9: `assert.compareArray` can also be used here.

anba (approved):
Tests pass with an implementation of the latest spec draft. 

`assert.compareArray` could be used instead of `assert.sameValue`, but that's not a blocker preventing to land this PR.

test/intl402/DurationFormat/constructor-option-read-order.js
Ms2ger: Yeah, please do change this.

FrankYFTang: done

test/intl402/DurationFormat/prototype/resolvedOptions/return-keys-order-default.js
FrankYFTang: done

FrankYFTang:
ping

test/intl402/DurationFormat/constructor-option-read-order.js
```diff
@@ -0,0 +1,47 @@
...
+    Object.defineProperty(opt, property, {
+        get() {
+            readKeys[readKeys.length] = property;
+            return undefined;
+        },
+    }));
+let p = new Intl.DurationFormat(undefined, opt);
+assert.compareArray(
+    readKeys,
+    ['numberingSystem',
```
ptomato, line 26: `localeMatcher` is read before `numberingSystem` but isn't in `resolvedOptions()`, I think? Probably better to use a Proxy that pushes all Get operations to the array, rather than just the ones in `resolvedOptions()`.

test/intl402/DurationFormat/prototype/resolvedOptions/return-keys-order-default.js
```diff
@@ -0,0 +1,34 @@
...
+    'minutes',
+    'minutesDisplay',
+    'seconds',
+    'secondsDisplay',
+    'milliseconds',
+    'millisecondsDisplay',
+    'microseconds',
+    'microsecondsDisplay',
+    'nanoseconds',
+    'nanosecondsDisplay']);
```
ptomato, line 34: `fractionalSeconds` is added last if not undefined, I think? Consider testing the order again when constructing the object with e.g. `new Intl.DurationFormat('en', { fractionalDigits: 0 })`.

ptomato (approved):
Thanks, seems correct but it seems like there's a couple of opportunities to cover more observable effects of the PRs. Can be done in a followup if necessary.

ptomato:
@FrankYFTang Do you want to address the above comments or postpone them to a followup?

test/intl402/DurationFormat/constructor-option-read-order.js
FrankYFTang: I will address that later

test/intl402/DurationFormat/prototype/resolvedOptions/return-keys-order-default.js
FrankYFTang: I will address that later

FrankYFTang:
> @FrankYFTang Do you want to address the above comments or postpone them to a followup?

I will address that later

test/intl402/DurationFormat/constructor-option-read-order.js
ptomato: → #4168

test/intl402/DurationFormat/prototype/resolvedOptions/return-keys-order-default.js
ptomato: → #4168

PR #3914, opened 2023-09-10
Test that `HasVarDeclaration` accounts for bindings created via `eval()`

shvaikalesh:
JSC bug: [Implement `HasVarDeclaration` abstract operation](https://bugs.webkit.org/show_bug.cgi?id=261179).

### DIFF
--- a/test/annexB/language/eval-code/direct/script-decl-lex-collision-in-sloppy-mode.js
+++ b/test/annexB/language/eval-code/direct/script-decl-lex-collision-in-sloppy-mode.js
@@ -2,28 +2,20 @@
 // This code is governed by the BSD license found in the LICENSE file.
 /*---
 esid: sec-globaldeclarationinstantiation
-description: Let binding collision with existing var declaration that was created via eval().
+description: Let binding collision with existing var declaration that was created for hoisted function.
 info: |
   [...]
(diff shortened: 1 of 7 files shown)


shvaikalesh:
It appears (CI results) like `$262.evalScript()` doesn't actually perform https://tc39.es/ecma262/#sec-runtime-semantics-scriptevaluation for SpiderMonkey?

The following snippet

```html
<!doctype html>
<meta charset="utf-8">
<script>eval("var foo");</script>
<script>let foo = 1;</script>
```

correctly throws `SyntaxError` in Firefox yet not in Chrome.

shvaikalesh:
On the CI, both V8 and SM fail existing https://github.com/tc39/test262/blob/main/test/language/global-code/script-decl-lex-var.js test (JSC passes).

So this implies `eshost` implements `$262.evalScript` incorrectly, and we can totally merge these tests.

shvaikalesh:
Thank you for review @anba!

After two-year break in contributing to `test262`, I've forgot that tests are running in strict mode as well 🤣

Accounted for comments in [bcee172](https://github.com/tc39/test262/pull/3914/commits/bcee172259c2de039b13f7f4ff0d511a65d1124b).

Ms2ger (approved):

PR #3915, opened 2023-09-11
Replace "imp" module with "importlib" in parseTestRecord

anba:
The "imp" module is deprecated in Python 3.4 and will be removed in Python 3.12. Use it only if the current Python version is too old to use the "importlib" module.

### DIFF
--- a/tools/packaging/parseTestRecord.py
+++ b/tools/packaging/parseTestRecord.py
@@ -7,7 +7,7 @@ from __future__ import print_function
 
 import os
 import re
-import imp
+import sys
 
 # Matches trailing whitespace and any following blank lines.
 _BLANK_LINES = r"([ \t]*[\r\n]{1,2})*"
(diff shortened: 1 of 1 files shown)


anba:
Release notes for Python 3.12 mentioning that the "imp" module is removed:
- https://docs.python.org/3.12/whatsnew/3.12.html#removed

ptomato (approved):
Thanks for fixing this up and sorry the review took so long.

I generally don't like having multiple code paths for the same thing, one of which is not tested. So I'd be fine with just requiring `importlib` outright, (Python 3.4  was released 9½ years ago) but let's see if any other maintainers know of a reason to keep compatibility with old Pythons.

ptomato:
We discussed this in the maintainers meeting of 2023-11-15. We won't remove `imp` at this time, so I'll merge this as-is.

@anba Could I ask how you ran into this? We don't have a clear idea of how anyone (other than the test262 CI) uses these tools.

anba:
> @anba Could I ask how you ran into this? We don't have a clear idea of how anyone (other than the test262 CI) uses these tools.

SpiderMonkey's [test262 update script](https://searchfox.org/mozilla-central/rev/7e60ac48dc5b3e3f06c1baf2a9a6e0352bd85c01/js/src/tests/test262-update.py#64-91) loads "parseTestRecord.py" to support environments which don't have PyYAML installed.

PR #3916, opened 2023-09-11
Update test/intl402 and test/language to use verifyProperty function

anba:
190d71c78aff0703947cb77a31665bbaaad1568a
- Change `verifyProperty` to also test the object value in addition to the descriptor value.
- This helps to verify `[[Get]]` is correctly implemented in addition to `[[GetOwnProperty]]`.

1392ed70acdbe5e6cb121c606caa2de2726ab7dd
- Validate that the descriptor passed to `verifyProperty` has the correct properties.
- And fix some tests which had typos in the descriptor properties.

b9a68616cff5f5e16d851c2ab737c621f2f75b1d
- It looks like these tests were never updated to use `verify(Not){Enumerable,Writable,Configurable}` functions.

All other commits:
- Replace `verify(Not){Enumerable,Writable,Configurable}` with `verifyProperty`.

### DIFF
--- a/harness/propertyHelper.js
+++ b/harness/propertyHelper.js
@@ -45,6 +45,19 @@ function verifyProperty(obj, name, desc, options) {
     return true;
   }
 
+  var names = Object.getOwnPropertyNames(desc);
+  for (var i = 0; i < names.length; i++) {
+    assert(
+      names[i] === "value" ||
+        names[i] === "writable" ||
+        names[i] === "enumerable" ||
+        names[i] === "configurable" ||
+        names[i] === "get" ||
+        names[i] === "set",
+      "Invalid descriptor field: " + names[i],
+    );
+  }
+
   assert(
     Object.prototype.hasOwnProperty.call(obj, name),
     "obj should have an own property " + nameStr
--- a/harness/propertyHelper.js
+++ b/harness/propertyHelper.js
@@ -68,6 +68,9 @@ function verifyProperty(obj, name, desc, options) {
     if (!isSameValue(desc.value, originalDesc.value)) {
       failures.push("descriptor value should be " + desc.value);
     }
+    if (!isSameValue(desc.value, obj[name])) {
+      failures.push(failures, "object value should be " + desc.value);
+    }
   }
 
   if (Object.prototype.hasOwnProperty.call(desc, 'enumerable')) {
--- a/test/language/arguments-object/10.6-11-b-1.js
+++ b/test/language/arguments-object/10.6-11-b-1.js
@@ -7,6 +7,7 @@ description: >
     Arguments Object has index property '0' as its own property, it
     shoulde be writable, enumerable, configurable and does not invoke
     the setter defined on Object.prototype[0] (Step 11.b)
+includes: [propertyHelper.js]
 ---*/
 
             var data = "data";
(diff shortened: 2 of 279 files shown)


test/language/arguments-object/10.6-11-b-1.js
```diff
@@ -26,26 +27,11 @@ description: >
 
             var argObj = (function () { return arguments })(1);
 
-            var verifyValue = false;
-            verifyValue = (argObj[0] === 1);
+verifyProperty(argObj, "0", {
```
Ms2ger, line 30: What's up with the indentation in these files? Feel free to also de-indent the code before this. Same elsewhere in this commit.

test/language/arguments-object/10.6-13-a-1.js
```diff
@@ -19,23 +20,11 @@ flags: [noStrict]
...
-                if (p === "callee" && argObj.hasOwnProperty("callee")) {
-                    verifyEnumerable = true;
-                }
-            }
-
-            var verifyConfigurable = false;
-            delete argObj.callee;
-            verifyConfigurable = argObj.hasOwnProperty("callee");
 
 assert(verifyValue, 'verifyValue !== true');
```
Ms2ger, line 24: Use `value: true`

test/language/statements/function/13.2-17-1.js
```diff
@@ -31,23 +32,12 @@ description: >
             var verifyValue = false;
             verifyValue = typeof fun.prototype.constructor === "function";
```
Ms2ger, line 33: Turn this into an `assert.sameValue`, please

Ms2ger (approved):

harness/propertyHelper.js
```diff
@@ -45,6 +45,19 @@ function verifyProperty(obj, name, desc, options) {
...
 
+  var names = Object.getOwnPropertyNames(desc);
+  for (var i = 0; i < names.length; i++) {
+    assert(
+      names[i] === "value" ||
+        names[i] === "writable" ||
+        names[i] === "enumerable" ||
+        names[i] === "configurable" ||
+        names[i] === "get" ||
+        names[i] === "set",
```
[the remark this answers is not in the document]
anba: No yet. (I have some WIP patches to add accessor property support to `verifyProperty`, though.)

We still need to handle get/set already, because some tests are first passing an accessor property descriptor object to `Object.defineProperty` and then use the same descriptor object with `verifyProperty`.

test/language/arguments-object/10.6-13-a-1.js
anba: `verifyValue` is the result of string comparison of the `typeof` expression, so it's not possible to check the value using `verifyProperty`.

I'll just replace the `verifyValue` assertion with:
```js
assert.sameValue(typeof argObj.callee, "function");
```

test/language/arguments-object/10.6-11-b-1.js
anba: I didn't bother to reformat all files when I replaced `runTestCase` in #415 :smile:

test/language/arguments-object/10.6-13-a-1.js
Ms2ger: Oh right, I confused myself. Thanks

anba: Fixed.

test/language/statements/function/13.2-17-1.js
anba: Fixed.

test/language/arguments-object/10.6-11-b-1.js
anba: I've added 6070e8127bb79e8e5096436422abee8c12324d04 to remove excess whitespace in a couple of more files.

Ms2ger (approved):

PR #3921, opened 2023-09-13
Add tests for import attributes using `with`

nicolo-ribaudo:
The first commit just copy-pastes the existing files for `assert` tests: when reviewing, you should exclude that commit from the diff ([diff of the remaining commits here](https://github.com/tc39/test262/pull/3921/files/dedf968aee1d2d485037a95c69afd5f6758b406e..408f5be83407223cb0963049a1cf75b4961c55fc)).

If there are comments with suggested changes that would also apply to the `assert` tests, I'd prefer to defer resolving them and then I'll open a PR updating both the `with` and `assert` tests at the same time.

Closes #3843


test/language/module-code/import-attributes/allow-nlt-before-with.js
```diff
@@ -0,0 +1,33 @@
...
+    [no LineTerminator here] assert
+
+negative:
+  phase: resolution
+  type: SyntaxError
+features: [import-attributes, globalThis]
+flags: [module, raw]
+---*/
+
+$DONOTEVALUATE();
```
nicolo-ribaudo, line 27: Lint error:
```
/home/circleci/test262/test/language/module-code/import-attributes/allow-nlt-before-with.js: NEGATIVE - Negative tests of type "early" must include a `throw` statement
```

isn't `$DONOTEVALUATE()` enough?

test/language/module-code/import-assertions/README.md
```diff
@@ -0,0 +1,2 @@
+Make sure to keep the tests in this folder aligned
+with the tests in the import-attributes folder.
```
nicolo-ribaudo: What is the correct way of adding README files?

Linting complains:
```
/home/circleci/test262/test/language/import/import-assertions/README.md: FRONTMATTER - No valid YAML-formatted frontmatter
/home/circleci/test262/test/language/import/import-assertions/README.md: LICENSE - Invalid Copyright header
```

test/language/module-code/import-attributes/allow-nlt-before-with.js
Ms2ger: Not in "raw" files, presumably because they don't have access to `$DONOTEVALUATE`:
```suggestion
throw "Test262: This statement should not be evaluated.";
```

test/language/module-code/import-attributes/import-attribute-newlines.js
```diff
@@ -0,0 +1,71 @@
...
+  type: SyntaxError
+features: [import-attributes, globalThis]
+flags: [module]
+---*/
+
+$DONOTEVALUATE();
+
+import "./ensure-linking-error_FIXTURE.js";
+
+import x from './import-attribute-1_FIXTURE.js' assert
```
Jon4t4n, line 36: Should this be `with` instead of `assert`?

test/language/import/import-attributes/json-idempotency.js
```diff
@@ -0,0 +1,21 @@
...
+---*/
+
+import viaStaticImport1 from './json-idempotency_FIXTURE.json' with { type: 'json' };
+import {default as viaStaticImport2} from './json-idempotency_FIXTURE.json' with { type: 'json' };
+import './json-idempotency-indirect_FIXTURE.js';
+
+assert.sameValue(viaStaticImport1, viaStaticImport2);
+assert.sameValue(globalThis.viaSecondModule, viaStaticImport1);
+
+import('./json-idempotency_FIXTURE.json', { assert: { type: 'json' } })
```
Jon4t4n, line 17: Should this be `with` instead of `assert`?

test/language/module-code/import-attributes/ensure-linking-error_FIXTURE.js
```diff
@@ -0,0 +1,9 @@
+// Copyright (C) 2022 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+// When imported, this file will ensure that a linking error happens by
+// importing a non-existent binding.
+// It can be used to with that there is a linking error, which means
```
ptomato, line 6:
```suggestion
// It can be used to assert that there is a linking error, which means
```

ptomato (approved):
Thanks, and sorry for the long review time.

For future reference it'd be better to squash "Include changes from PR 3932" into "Copy all import assertions tests as a basis for import attributes tests", and the last two commits into "Update import attributes tests to use the with keyword".

In order not to delay this any longer, I'll do the squashing and fix up that one search-replace mistake, and merge it right away.

PR #3923, opened 2023-09-14
Temporal: Audit of user code calls, part 3

ptomato:
Next part of #3897. These tests should provide coverage for https://github.com/tc39/proposal-temporal/pull/2671. Best reviewed commit by commit.


test/built-ins/Temporal/PlainDate/prototype/since/order-of-operations.js
```diff
@@ -86,8 +86,8 @@ const instance = new Temporal.PlainDate(2000, 5, 2, ownCalendar);
 
 const otherDatePropertyBag = TemporalHelpers.propertyBagObserver(actual, {
   year: 2001,
-  month: 5,
-  monthCode: 'M05',
+  month: 6,
+  monthCode: "M06",
```
cjtenny, line 90: Which fast path is this avoiding?

cjtenny (approved):
Looks great to me, with one issue and one question. I think the resolution to the former is simple and wouldn't need re-review - either re-adding the tests or not.

I did find some paths that weren't covered, but that's no reason to block this set of tests and make a 6000 line pair of PRs any longer - I'll add that further coverage in a separate PR since I wrote some of them already while reasoning if the paths were viable.

6000 lines between temporal and tests for part 3, how do you do it? :face_with_spiral_eyes:

Ms2ger (approved):
I'll defer to Cam here

test/built-ins/Temporal/Duration/prototype/add/nanoseconds-to-days-loop-indefinitely-2.js
```diff
@@ -1,71 +0,0 @@
...
-    substitutions
-  );
-  substitutions.length = count;
-  let i = 0;
-  for (i = 0; i < substitutions.length; i++) {
-    // (this value)
-    substitutions[i] = [dayInstant];
-  }
-  // Record calls in calls[]
-  TemporalHelpers.observeMethod(calls, timeZone, "getPossibleInstantsFor");
```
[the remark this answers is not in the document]
ptomato: I think this was just a mistake where I messed up the `git mv`. Glad you caught it. I've re-added the tests, without the `-2` in the name.

test/built-ins/Temporal/PlainDate/prototype/since/order-of-operations.js
ptomato: You can see it a bit farther down in the same commit, where I added a new test case with the previous value of `otherDatePropertyBag`. In RoundDuration with smallestUnit = year, if there are no excess months/weeks then the dateUntil call will be skipped in DifferenceDate. This change adds an excess month.

PR #3925, opened 2023-09-15
Update two Array.fromAsync `this-constructor` tests to test the correct behavior 

rmahdav:
Based on [this fix](https://github.com/tc39/proposal-array-from-async/pull/41) on Array.fromAsync spec to avoid calling constructor twice, the following two tests needed to be updated:
built-ins/Array/fromAsync/this-constructor
built-ins/Array/fromAsync/this-constructor-operations


ljharb:
Duplicated by #3926.

PR #3926, opened 2023-09-15
Update two constructor tests of Array.fromAsync

rmahdav:
Based on [this fix](https://github.com/tc39/proposal-array-from-async/pull/41) on Array.fromAsync spec to avoid calling constructor twice, the following two tests needed to be updated:
built-ins/Array/fromAsync/this-constructor
built-ins/Array/fromAsync/this-constructor-operations


test/built-ins/Array/fromAsync/this-constructor-operations.js
```diff
@@ -60,7 +60,6 @@ asyncTest(async function () {
 
   // Note https://github.com/tc39/proposal-array-from-async/issues/35
```
ptomato, line 61: We can delete this comment now, too, since the fix makes it obsolete.

ptomato (approved):
Thanks for the contribution! This looks good. Would you mind removing the comment as well before we merge it in?

test/built-ins/Array/fromAsync/this-constructor-operations.js
rmahdav: Done. Thanks for pointing that out.

ptomato (approved):
Thanks!

PR #3929, opened 2023-09-22
Fix harness tests for verifyProperty

anba:
I didn't run the harness tests when working on #3916, which introduced these three bugs:

- `Object.getOwnPropertyNames(desc)` should be called only after verifying `desc` is not `null`.
- I didn't resolve a merge conflict with other local changes properly, which caused `failures.push(failures, ...`.
- And the "verifyProperty-value-error.js" test need to be updated to account for the new failure message.

### DIFF
--- a/harness/propertyHelper.js
+++ b/harness/propertyHelper.js
@@ -45,19 +45,6 @@ function verifyProperty(obj, name, desc, options) {
     return true;
   }
 
-  var names = Object.getOwnPropertyNames(desc);
-  for (var i = 0; i < names.length; i++) {
-    assert(
-      names[i] === "value" ||
-        names[i] === "writable" ||
-        names[i] === "enumerable" ||
-        names[i] === "configurable" ||
-        names[i] === "get" ||
-        names[i] === "set",
-      "Invalid descriptor field: " + names[i],
-    );
-  }
-
   assert(
     Object.prototype.hasOwnProperty.call(obj, name),
     "obj should have an own property " + nameStr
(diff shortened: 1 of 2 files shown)


harness/propertyHelper.js
```diff
@@ -45,19 +45,6 @@ function verifyProperty(obj, name, desc, options) {
...
-        names[i] === "enumerable" ||
-        names[i] === "configurable" ||
-        names[i] === "get" ||
-        names[i] === "set",
-      "Invalid descriptor field: " + names[i],
-    );
-  }
-
   assert(
     Object.prototype.hasOwnProperty.call(obj, name),
```
gibson042, line 49: Isn't this already covered by above use of `Object.getOwnPropertyDescriptor` without potentially double-triggering a proxy trap?
```suggestion
```

harness/propertyHelper.js
```diff
@@ -45,19 +45,6 @@ function verifyProperty(obj, name, desc, options) {
     return true;
   }
```
gibson042, line 46:
```suggestion
  } else if (originalDesc === undefined) {
    assert.sameValue(
      originalDesc,
      desc,
      "obj['" + nameStr + "'] descriptor should be " + String(desc)
    );
  }
```
or probably better:
```js
  // If either the actual descriptor or expected descriptor is undefined,
  // the other must be as well.
  if (originalDesc === undefined || desc === undefined) {
    assert.sameValue(
      originalDesc,
      desc,
      "obj['" + nameStr + "'] descriptor should be " + String(desc)
    );
    return true;
  }
```

harness/propertyHelper.js
```diff
@@ -75,14 +62,27 @@ function verifyProperty(obj, name, desc, options) {
...
   var failures = [];
 
   if (Object.prototype.hasOwnProperty.call(desc, 'value')) {
     if (!isSameValue(desc.value, originalDesc.value)) {
       failures.push("descriptor value should be " + desc.value);
     }
     if (!isSameValue(desc.value, obj[name])) {
-      failures.push(failures, "object value should be " + desc.value);
+      failures.push("object value should be " + desc.value);
     }
```
gibson042, line 86: This whole block seems unnecessary and potentially a source of incorrect failures, but I guess that's just the state of `verifyProperty`.

gibson042 (approved):

harness/propertyHelper.js
anba: Implementations can have different code paths for `Object.getOwnPropertyDescriptor` and `Object.prototype.hasOwnPropert`, so the `verifyProperty` function tries to ensure that all MOP operations are consistent. Proxies probably shouldn't be used with `verifyProperty` anyway.

PR #3930, opened 2023-09-22
Use correct variables in time zone case insensitive test

anba:
Also remove `US/Pacific-New` which has been removed from tz data since release 2020b.


ljharb (approved):

ptomato (approved):
Thanks!

PR #3931, opened 2023-09-24
Increase coverage of String.prototype.replace $xy replacement patterns

gibson042:
* $xy is a valid capture index
* $xy is not a valid capture index but $x is
* neither $xy nor $x is a valid capture index

Ref https://github.com/tc39/ecma262/issues/1426
Ref https://github.com/tc39/ecma262/pull/3157

### DIFF
--- a/test/built-ins/String/prototype/replace/regexp-capture-by-index.js
+++ b/test/built-ins/String/prototype/replace/regexp-capture-by-index.js
@@ -78,7 +78,7 @@ info: |
   6. Return result.
 ---*/
 
-var str = "foo-x-bar";
+var str = 'foo-x-bar';
 
 var re0 = /x/;
 var re1 = /(x)/;
--- a/test/built-ins/String/prototype/replace/regexp-capture-by-index.js
+++ b/test/built-ins/String/prototype/replace/regexp-capture-by-index.js
@@ -83,6 +83,7 @@ var str = 'foo-x-bar';
 var x = 'x';
 var re0 = /x/;
 var re1 = /(x)/;
+var re1x = /(x)($^)?/;
 var re10 = /((((((((((x))))))))))/;
 
 assert.sameValue(str.replace(x, '|$0|'), 'foo-|$0|-bar',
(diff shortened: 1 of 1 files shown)


ljharb (approved):
Confirmed these tests pass my ES2024 GetSubstitution implementation in `es-abstract`, and fail on <= ES2023 (as expected)

ljharb:
update; https://github.com/tc39/ecma262/pull/3157#issuecomment-1734728798 should get some coverage too

gibson042:
@ljharb coverage expanded

ljharb (approved):

ptomato:
This normative change got consensus conditionally in the 2023-09 TC39 plenary meeting. "If there is a mismatch between engines, the presenter should bring the change back to TC39 to discuss next steps." So I guess before we land this, the proponents of the normative change should perform that investigation.

gibson042:
Done. The only implementation with failures is engine262, which does not implement the single-digit fallback:
* `$10` is not a capture index (but `$1` is) in /(x)/
* `$10` is not a capture index (but `$1` is) in /(x)($^)?/
* `$10` before `0` is not a capture index (but `$1` is) in /(x)/
* `$10` before `0` is not a capture index (but `$1` is) in /(x)($^)?/
* `$20` is not a capture index (but `$2` is a failed capture index) in /(x)($^)?/
* `$20` is not a capture index (but `$2` is) in /((((((((((x))))))))))/
* `$20` before `0` is not a capture index (but `$2` is a failed capture index) in /(x)($^)?/
* `$20` before `0` is not a capture index (but `$2` is) in /((((((((((x))))))))))/

```sh
$ eshost -se '"foo-x-bar".replace(/(x)/, "|$10|")'
#### ChakraCore, GraalJS, Hermes, JavaScriptCore, Moddable XS, QuickJS, SpiderMonkey, V8
foo-|x0|-bar

#### engine262
foo-|$10|-bar
```

ptomato:
Thanks. So does that mean there are no longer any blockers for this?

gibson042:
Yes, that is my understanding.

ljharb:
It's easy to squash it at landing time, but I wanted more reviews before landing it.

linusg (approved):
LGTM, also passes with my [literal implementation of `GetSubstitution`](https://codeberg.org/kiesel-js/kiesel/src/commit/76e0de0719cf14727ba9d574a68638d431dd4d87/src/builtins/string.zig#L87-L310).

PR #3919, opened 2023-09-12
Update import attributes `assert` tests to current semantics

nicolo-ribaudo:
This was part of https://github.com/tc39/test262/pull/3843, but:
- I rewrote it from scratch to make sure that it was correct
- It just updates `assert` semantics, it does not add `with` tests yet (I'll do that in a follow-up)

There is some complexity related to tests with invalid attributes, that should throw when linking, so I have a bunch of tests to verify valid _syntax_ that will throw then during resolution.

Issue #3891, opened 2023-08-09
XS engine failing to install on CircleCI


ptomato:
```
esvu ✖ Error: Command failed with exit code 1: /home/circleci/.esvu/bin/xs -e print("42");
/home/circleci/.esvu/bin/xs: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by /home/circleci/.esvu/bin/xs)
/home/circleci/.esvu/bin/xs: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by /home/circleci/.esvu/bin/xs)
/home/circleci/.esvu/bin/xs: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /home/circleci/.esvu/bin/xs)
/home/circleci/.esvu/bin/xs: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.35' not found (required by /home/circleci/.esvu/bin/xs)
/home/circleci/.esvu/bin/xs: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by /home/circleci/.esvu/bin/xs)
```
First guess, the CI environment has a glibc that's too old for XS's versioned symbols to link against?

PR #3932, opened 2023-09-25
Fix import assertions tests

anba:
- Correct file paths in "ensure-linking-error_FIXTURE.js"
- Remove no longer valid assertion in "import-assertion-empty.js".

See also #3919.

### DIFF
--- a/test/language/module-code/import-assertions/ensure-linking-error_FIXTURE.js
+++ b/test/language/module-code/import-assertions/ensure-linking-error_FIXTURE.js
@@ -6,4 +6,4 @@
 // It can be used to assert that there is a linking error, which means
 // that there are no parsing errors.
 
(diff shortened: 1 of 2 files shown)


nicolo-ribaudo (approved):

nicolo-ribaudo:
@anba Not sure if you saw it, but if you need `with` tests you can find them at https://github.com/tc39/test262/pull/3921 (I will apply this patch also to that PR).

anba:
> @anba Not sure if you saw it, but if you need `with` tests you can find them at #3921 (I will apply this patch also to that PR).

I don't actually work on implementing import assertions, I only noticed these test failures when updating SpiderMonkey's test262 copy. :smile:

ptomato:
Unfortunately, the auto-merge won't work because of #3891.

ptomato (approved):
Re-approving

PR #3934, opened 2023-09-28
Only lint js and json files

Ms2ger:
CheckFileName is the only one that could arguably be valid for other files, but it doesn't seem worth the complexity to run it.


ptomato (approved):
Seems like a good idea to me.

Another option might be to skip only `README.md` files, if anyone was concerned about other files finding their way into the tree. But I don't know if this is actually a problem.

PR #3938, opened 2023-10-04
added missing features: line to metadata for two tests

ben-allen:
Previously merged 3890  was missing features: line, this corrects that.


ptomato (approved):
Thanks for following up so quickly. LGTM.

PR #3939, opened 2023-10-04
Temporal: Audit of user code calls, part 4 (& last)

ptomato:
Next part following on from #3923. These tests should provide coverage for tc39/proposal-temporal#2519. Two commits, best reviewed separately.

Unlike the other parts of the user code calls audit, this part didn't add any new fast paths; all it does is ensure that a TimeZone or Calendar method is only looked up once during the course of any Temporal method call. Therefore, no new tests are added (except for a few methods which were missing an `order-of-operations.js` entirely). This PR only adjusts the expectations of existing tests.


cjtenny (approved):
I promise I reviewed this thoroughly despite having no comments - all my comments in reviewing this came with thoroughly reviewing the normative changes accompanying it, and the tests do a great job covering that. Looks great to me :)

ptomato:
Thanks for the review!

PR #3940, opened 2023-10-09
Allow @tc39-staging to modify features.txt

syg:
This came up while working on the V8 test262 2-way sync bot for `staging/`. It is likely that tests added to staging/ will be for a new feature that requires a new entry in `features.txt`, so the staging group should have write access to that file.

cc @rmahdav

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,2 +1,3 @@
 * @tc39/test262-maintainers
+/features.txt @tc39/test262-staging @tc39/test262-maintainers
 /test/staging/ @tc39/test262-staging @tc39/test262-maintainers


ljharb (approved):
makes sense to me

syg:
An alternative here is to just... not lint it. I guess the downside is typos are more likely? Does anyone recall the rationale for why there's an list of allowed features in the first place?

ljharb:
I would assume it's because it was a common problem that people would typo feature names, and, that having a list ensures that there's a central place for discovery.

Ms2ger (approved):
This is fine by me, but maybe give it a little longer before merging

ptomato (approved):

ptomato:
This has the unexpected (to me, at least) side effect that the tc39-staging team will get pinged for a review every time someone modifies `features.txt`. Seems harmless, but just mentioning it in case anyone wonders why it's happening.

syg:
> This has the unexpected (to me, at least) side effect that the tc39-staging team will get pinged for a review every time someone modifies features.txt. Seems harmless, but just mentioning it in case anyone wonders why it's happening.

Seems like a positive side effect to me tbh. I appreciate being pinged on when new feature tests land.

PR #3942, opened 2023-10-10
Fix symbol-tag-generators-builtin.js

anba:
Two issues:
1. There was a typo, the first `gen` should have been `genFn`.
2. And the last `toString` result should be `'[object Generator]'`, because `Object.getPrototypeOf(gen) == genFn.prototype`. And the `@@toStringTag` property of `genFn.prototype` is inherited from `Generator.prototype[@@toStringTag]`.

That also means the "iterator-helpers" flag isn't needed for this test.


ljharb (approved):
Whoops, thanks

PR #3943, opened 2023-10-10
Allow linter to use alternate features.txt file

LiviuArln:
V8 explores 2-way-sync-ing of tests and needs to be able to introduce new features and tests first in V8 repo before exporting them to test262 github repo. Linting these tests requires the linter to become a bit more flexible with where it looks for the features file.


Ms2ger:
Isn't this what #3940 is for?

LiviuArln:
#3940 is related to this. We plan to have a V8 side copy of features.txt that we will 2-way-sync with the one here. This change is to allow us to use _our_ copy of features.txt file when linting _our_ test changes. Once we successfully ran the linter in V8 we are sure we can export to this repo both the tests and any updates to the features.txt file we might have needed.

ptomato (approved):
The explanation makes sense to me. This seems like something that wouldn't be used on our end but if it facilitates this exchange of tests I'm all for it!

LiviuArln:
Thanks! I guess I only need to wait for the failed check approval now. Or, how does this work?

ljharb:
Just needs to be rebased; and a maintainer can approve CI and land it.

PR #3946, opened 2023-10-16
Update likely subtags tests for CLDR 44

anba:
CLDR 44 changed likely-subtags mappings for "und" language tags, so we have to update these two tests.


anba:
As with previous CLDR updates, this PR shouldn't be merged before CLDR 44 is officially released.

Ms2ger (approved):
Land whenever ready

anba:
CLDR 44 / ICU 74 have been released, so this PR can now be merged.

PR #3952, opened 2023-10-27
Added tests for invalid combinations of DurationFormat style options

ben-allen:
Use of "long", "short", or "narrow" styles after units with "numeric" or "2-digit" styles should throw.


anba (approved):
LGTM and I've verified all tests pass with an implementation of the latest spec draft.

Ms2ger (approved):
I'm happy to trust anba here

PR #3954, opened 2023-11-03
Temporal: Change overflow behaviour of PYM/PMD.toPlainDate

ptomato:
These tests cover the proposed normative change:
https://github.com/tc39/proposal-temporal/issues/2706


Ms2ger (approved):

ptomato:
The normative change reached consensus in the November 2023 TC39 meeting, so this can be merged now.

PR #3955, opened 2023-11-03
Add test for async module not blocking sync siblings

nicolo-ribaudo:
cc @guybedford @codehag for reviews

Currently Chrome, Deno and Node.js pass this test, while Safari, Firefox and Bun fail it. This seems to suggest a JS engine difference (V8 passes, JSC/SM fail), but given that this is related to modules I'm not sure whether the bug lives in the engine or in the embedder. Maybe it's worth duplicating this test in WPT?

### DIFF
--- a/test/language/module-code/top-level-await/async-module-does-not-block-sibling-modules.js
+++ b/test/language/module-code/top-level-await/async-module-does-not-block-sibling-modules.js
@@ -13,3 +13,4 @@ features: [top-level-await]
 import "./async-module-tla_FIXTURE.js";
 import { check } from "./async-module-sync_FIXTURE.js";
 assert.sameValue(check, false);
+$DONE();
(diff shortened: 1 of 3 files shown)


nicolo-ribaudo:
Oh well, all engines are passing this test 🙃
I'll open a PR in wpt too.

ptomato:
> Oh well, all engines are passing this test 🙃

I'm not sure if you meant that you had verified them separately, or were looking at the CI results — if the latter, then not necessarily all engines are passing the test. It just means they can execute the test without crashing, whether it passes or not.

nicolo-ribaudo:
Oh yes, I only saw that CI was green 😓

guybedford (approved):

nicolo-ribaudo:
Both V8 and SM pass this test here, but Firefox does not. I think having this test both here and in WPT has value.

Ms2ger (approved):

PR #3956, opened 2023-11-06
Temporal: Tests for balancing units after rounding

ptomato:
This PR contains tests that cover normative changes to Temporal that reached consensus in May 2023.
Normative PR: https://github.com/tc39/proposal-temporal/pull/2571


cjtenny (approved):
These look good to me; they cover the changes in https://github.com/tc39/proposal-temporal/pull/2571/ and the bug that I found in the spec change didn't affect the tests - tests were correct for desired behavior, spec text had just one oops that's been fixed!

ptomato:
Thanks for the review!

PR #3958, opened 2023-11-09
Remove locales with explicit u-hc Unicode extension

anba:
The tests are passing an explicit `hour12` option, which disables any `hourCycle` option and any `u-hc` Unicode extension. Therefore passing input locales with `u-hc` is invalid.

```js
let locale = "en-u-hc-h24";
let hcDefault = new Intl.DateTimeFormat(locale, { hour: "numeric" }).resolvedOptions().hourCycle;

assert.sameValue(hcDefault, "h24", "hour-cycle through Unicode extension");

let hc24 = new Intl.DateTimeFormat(locale, { hour: "numeric", hour12: false }).resolvedOptions().hourCycle;

// Incorrect assertion, because |hc24| uses |hour12|, which disables any
// hour-cycle option and the hour-cycle is determined from the locale "en".
// That means |hc24| will be "h23".
assert.sameValue(hc24, hcDefault);
```

### DIFF
--- a/test/intl402/DateTimeFormat/prototype/resolvedOptions/hourCycle-default.js
+++ b/test/intl402/DateTimeFormat/prototype/resolvedOptions/hourCycle-default.js
@@ -21,25 +21,33 @@ info: |
         c. If hc is null, set hc to dataLocaleData.[[hourCycle]].
   27. Set dateTimeFormat.[[HourCycle]] to hc.
 
-locale: [en, fr, it, ja, zh, ko, ar, hi, en-u-hc-h24]
+locale: [en, fr, it, ja, zh, ko, ar, hi]
 ---*/
 
-let locales = ["en", "fr", "it", "ja", "ja-u-hc-h11", "zh", "ko", "ar", "hi", "en-u-hc-h24"];
+let locales = ["en", "fr", "it", "ja", "zh", "ko", "ar", "hi"];
 
-locales.forEach(function(locale) {
-  let hcDefault = new Intl.DateTimeFormat(locale, { hour: "numeric" }).resolvedOptions().hourCycle;
+for (let locale of locales) {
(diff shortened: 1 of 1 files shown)


linusg (commented):
Given:

```
// no locale has "h24" as a default. see https://github.com/tc39/ecma402/pull/758#issue-1622377292
```

This comment / code path no longer seems relevant:

```
// however, "h24" can be set via locale extension.
```

ptomato:
OK. @ben-allen explained this to me and I think I understand it.

With that understanding (which I may still have gotten wrong), Linus's comment makes sense to me. Maybe the loop body should look something like this?
```js
let hcDefault = new Intl.DateTimeFormat(locale, { hour: "numeric" }).resolvedOptions().hourCycle;
let hcHour12 = new Intl.DateTimeFormat(locale, { hour: "numeric", hour12: true }).resolvedOptions().hourCycle;

assert(hcHour12 === "h11" || hcHour12 === "h12", "Expected hourCycle for hour12: true to be in [\"h11\", \"h12\"]");
if (hcDefault === "h11" || hcDefault === "h12") {
  assert.sameValue(hcHour12, hcDefault);
}

// no locale has "h24" as a default. see https://github.com/tc39/ecma402/pull/758#issue-1622377292
let hcNotHour12 = new Intl.DateTimeFormat(locale, { hour: "numeric", hour12: false }).resolvedOptions().hourCycle;
assert.sameValue(hcNotHour12, "h23", "Expected hourCycle for hour12: false to be \"h23\"");
```

ptomato (changes requested):
(see previous comment, which I neglected to send as a review comment)

ptomato (approved):

ptomato:
OK, fair enough, that seems fine for now.

PR #3959, opened 2023-11-10
Temporal: Remove conditional lookups

ptomato:
These tests cover https://github.com/tc39/proposal-temporal/issues/2724

### DIFF
--- a/test/built-ins/Temporal/Duration/compare/order-of-operations.js
+++ b/test/built-ins/Temporal/Duration/compare/order-of-operations.js
@@ -109,7 +109,6 @@ const expectedOpsForPlainRelativeTo = expected.concat([
   "has options.relativeTo.calendar.year",
   "has options.relativeTo.calendar.yearMonthFromFields",
   "has options.relativeTo.calendar.yearOfWeek",
-  // lookup
   "get options.relativeTo.calendar.dateFromFields",
   "get options.relativeTo.calendar.fields",
   "call options.relativeTo.calendar.fields",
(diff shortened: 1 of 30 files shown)


test/built-ins/Temporal/Duration/compare/order-of-operations.js
```diff
@@ -264,6 +265,8 @@ const expectedOpsForZonedRelativeTo = expected.concat([
   "get options.relativeTo.timeZone.getPossibleInstantsFor",
   "call options.relativeTo.timeZone.getPossibleInstantsFor",
   "call options.relativeTo.timeZone.getOffsetNanosecondsFor",
+  // lookup
+  "get options.relativeTo.calendar.dateAdd",
```
gibson042, line 269: I wonder if expectedOpsFor{Plain,Zoned}RelativeTo could be DRYed out to better highlight actual differences.

test/built-ins/Temporal/Duration/compare/order-of-operations.js
```diff
@@ -134,6 +134,8 @@ const expectedOpsForPlainRelativeTo = expected.concat([
   "get options.relativeTo.year.valueOf",
   "call options.relativeTo.year.valueOf",
   "call options.relativeTo.calendar.dateFromFields",
+  // lookup
```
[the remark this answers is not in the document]
ptomato: Usually the lookup is in the algorithm steps for the method itself, rather than in an AO, so that's why it's not labeled with an AO. (I prefer not to put the actual step numbers unless it's really needed for clarity, since they bitrot so easily.) I can try to label these with things like `// lookup in Duration.compare` instead.

test/built-ins/Temporal/Duration/compare/order-of-operations.js
ptomato: I've extracted the common "prefix" of both of them.

PR #3957, opened 2023-11-06
Temporal.Duration normalization tests, part 1 of 3

ptomato:
~~(stacked on top of #3956, will rebase as appropriate)~~

This PR contains tests that cover a Temporal normative change that reached consensus in July 2023.
Normative PR: https://github.com/tc39/proposal-temporal/pull/2722

PR #3961, opened 2023-11-16
Temporal.Duration normalization tests, part 2 of 3

ptomato:
~~(stacked on top of #3957, will rebase as appropriate)~~

This PR contains tests that cover a Temporal normative change that reached consensus in July 2023.
Normative PR: tc39/proposal-temporal#2727


ptomato:
Added extra tests for the cases that @anba identified, in a fixup commit so it's easy to see what changed. I will squash it before merging.

Ms2ger (approved):

ptomato:
Thanks for the review. Since yesterday I added some tests from @nicolo-ribaudo as well as some tests from @anba, which I've reviewed. So I think this is ready. I'll merge it at the same time as the proposal-temporal PR.

PR #3816, opened 2023-04-12
Set union tests

frehner:
Hello! First time code contributor here, so please feel free to correct anything and everything that I'm doing. I appreciate your time. ❤️ 

---

This is the very beginnings of helping with #3738 . I'm hoping that I can be helpful here, but there are still quite a bit that I'm missing and unsure on how to do. 

If it helps, there's also an associated engine262 PR here https://github.com/engine262/engine262/pull/212 where I've done a basic implementation of the spec and have been using it to validate these tests so far. 

In any case, I figured this was a good place to stop and get feedback before moving on further. I want to make sure that I'm not doing something wrong or using patterns that are incorrect. 

Thanks again!

PR #3962, opened 2023-11-19
Consolidate set methods features

bakkot:
After https://github.com/tc39/test262/pull/3816 we accidentally had two "set-methods" features, whoops.


ljharb (approved):

Ms2ger (approved):
Thanks

PR #3963, opened 2023-11-19
Test for assignment with target wrapped in two sets of parentheses

gibson042:
Currently failing on XS: https://github.com/Moddable-OpenSource/moddable/issues/1256


Ms2ger:
Should we test this for other cases as well? `((x)) = "double-wrapped"`, like you mentioned in the issue? Prefix operators?

Issue #3738, opened 2022-12-01
Tests for Set methods


bakkot:
The [Set methods proposal](https://github.com/tc39/proposal-set-methods) is now stage 3 (🎉 ). That means it's time for tests! Here's the things which occur to me offhand; feel free to edit this issue to add more.

- for all seven methods
  - [ ] basic interactions between native Sets
    - [ ] including extremal cases, i.e. when either the receiver or the argument is empty and when the receiver and the argument are the same value
  - [ ] works when the argument is a userland class which implements `size`/`has`/`keys`
    - [ ] calls the appropriate methods on the userland class in the appropriate order, which for some algorithms varies depending on the relative sizes of the receiver vs the argument
  - [ ] works when the argument is a Map (treating it a Set of its keys)
  - [ ] throws if the argument lacks `size` or `has` or `keys`
    - [ ] throws if the argument's `size` is, or coerces to, `NaN`
    - [ ] throws if the argument's `has` or `keys` is non-callable
    - [ ] as a special case, throws if passed an Array
  - [ ] works when the receiver is a Set subclass
    - [ ] but does not read the receiver's `size`, `has`, or `keys` properties, nor those of `Set.prototype`
  - [ ] throws when the receiver is not a Set
- for each of `union`, `intersection`, `difference`, `symmetricDifference`
  - [ ] produces a native Set
    - [ ] even if the receiver is a subclass
    - [ ] even if Symbol.species is overridden; Symbol.species is not inspected at all)
  - [ ] does not invoke `Set.prototype.add` to construct the result (testing this requires mutating `Set.prototype`)
  - [ ] order of the result, both when the receiver is smaller and larger than the argument (which affects the algorithm used in some cases)
    - [ ] order of the result when the argument is a userland class which mutates the receiver during execution of the algorithm

This list previously included a special case for a sorting step in `intersection`, but that step was removed in https://github.com/tc39/proposal-set-methods/pull/94. The new behavior should now be covered by the "order of the result, both when the receiver is smaller and larger than the argument" case above.

frehner:
Hello @bakkot thanks for this issue! I'm a first time contributor to test262, in the process of writing some tests for these features (see #3816 ) and am also cross-referencing this todo list as I look to add more tests. 

Forgive my ignorance, but I had a question about:

> does not invoke Set.prototype.add to construct the result (testing this requires mutating Set.prototype)

[...]

PR #3966, opened 2023-11-26
More set methods tests

bakkot:
Followup to https://github.com/tc39/test262/pull/3816. Fixes https://github.com/tc39/test262/issues/3738. cc @syg

Basically just tweaked some of the tests from #3816, copy-pasted them for each of the other methods, and tweaked as appropriate.

### Reviewed:

- [x] difference
- [x] intersection
- [x] isDisjointFrom
- [x] isSubsetOf
- [x] isSupersetOf
- [x] symmetricDifference
- [x] union


rmahdav:
It would be appreciated if these tests are reviewed and landed as soon as possible, so we will be able to ship set methods on Chrome. Thanks! @tc39/test262-maintainers

test/built-ins/Set/prototype/difference/set-like-array.js
```diff
@@ -0,0 +1,26 @@
+// Copyright (C) 2023 Kevin Gibbons, Anthony Frehner and Kevin Gibbons. All rights reserved.
```
Ms2ger, line 1: I appreciate @bakkot as much as the next guy, but isn't this a bit over the top?

Ms2ger (commented):
I've reviewed the `difference` folder; all seems good in there and I'd be happy to land those. One thing I don't think I've seen is a check for the +0/-0 step.

I don't know when I'll have time for the rest of this PR.

test/built-ins/Set/prototype/difference/set-like-array.js
bakkot: Lol whoops, messed up my regex.

bakkot:
> One thing I don't think I've seen is a check for the +0/-0 step.

Nice catch, added. I had it for the other three non-predicate methods; don't know why I missed it for `difference`.

test/built-ins/Set/prototype/intersection/set-like-class-order.js
```diff
@@ -0,0 +1,140 @@
...
+/*---
+esid: sec-set.prototype.intersection
+description: Set.prototype.intersection calls a Set-like class's methods in order
+features: [set-methods]
+includes: [compareArray.js]
+---*/
+
+let observedOrder = [];
+
+function observableIterator() {
```
ptomato, line 12: Here's where I wish we had already expanded the Observable utilities in `harness/temporalHelpers.js` into their own, more general, helper file ... oh well, next time :smile:

test/built-ins/Set/prototype/intersection/size-is-a-number.js
```diff
@@ -0,0 +1,72 @@
...
+);
+
+s2.size = "string";
+assert.throws(
+  TypeError,
+  function () {
+    s1.intersection(s2);
+  },
+  "GetSetRecord throws an error when size is a non-numeric string"
+);
```
ptomato, line 72: Should we test what happens with `s2.size = "2"` or `s2.size = true`? Maybe in a separate file from this one. (Glad we will Stop Coercing Things from now on!)

ptomato (commented):
I've made it most of the way through the `intersection/` folder; here are the comments I have so far. Thanks for working on this!

test/built-ins/Set/prototype/intersection/combines-same-sets.js
```diff
@@ -0,0 +1,16 @@
...
+includes: [compareArray.js]
+---*/
+
+const s1 = new Set([1, 2]);
+const s2 = new Set([1, 2]);
+const expected = [1, 2];
+const combined = s1.intersection(s2);
+
+assert.compareArray([...combined], expected);
+assert.sameValue(combined instanceof Set, true, "The returned object is a Set");
```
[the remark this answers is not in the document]
bakkot: Done (for all of them).

test/built-ins/Set/prototype/intersection/require-internal-slot.js
```diff
@@ -0,0 +1,25 @@
...
+assert.throws(TypeError, () => intersection.call(null), "null");
+assert.throws(TypeError, () => intersection.call(true), "true");
+assert.throws(TypeError, () => intersection.call(""), "empty string");
+assert.throws(TypeError, () => intersection.call(Symbol()), "symbol");
+assert.throws(TypeError, () => intersection.call(1), "1");
+assert.throws(TypeError, () => intersection.call(1n), "1n");
+assert.throws(TypeError, () => intersection.call({}), "plain object");
+assert.throws(TypeError, () => intersection.call([]), "array");
+assert.throws(TypeError, () => intersection.call(new Map()), "map");
+assert.throws(TypeError, () => intersection.call(Set.prototype), "Set.prototype");
```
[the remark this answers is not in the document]
bakkot: Done (for all of them).

test/built-ins/Set/prototype/intersection/size-is-a-number.js
ptomato: OK, works for me.

test/built-ins/Set/prototype/difference/require-internal-slot.js
```diff
@@ -0,0 +1,27 @@
...
+
+assert.sameValue(typeof difference, "function");
+
+const realSet = new Set([]);
+
+assert.throws(TypeError, () => difference.call(undefined, realSet), "undefined");
+assert.throws(TypeError, () => difference.call(null, realSet), "null");
+assert.throws(TypeError, () => difference.call(true, realSet), "true");
+assert.throws(TypeError, () => difference.call("", realSet), "empty string");
+assert.throws(TypeError, () => difference.call(Symbol(, realSet)), "symbol");
```
ptomato, line 21: Looks like something not quite right in the codegen, same thing happened in other files
```suggestion
assert.throws(TypeError, () => difference.call(Symbol(), realSet), "symbol");
```

test/built-ins/Set/prototype/difference/require-internal-slot.js
```diff
@@ -0,0 +1,27 @@
...
+assert.throws(TypeError, () => difference.call(1, realSet), "1");
+assert.throws(TypeError, () => difference.call(1n, realSet), "1n");
+assert.throws(TypeError, () => difference.call({}, realSet), "plain object");
+assert.throws(TypeError, () => difference.call([], realSet), "array");
+assert.throws(TypeError, () => difference.call(new Map(, realSet)), "map");
```
ptomato, line 26:
```suggestion
assert.throws(TypeError, () => difference.call(new Map(), realSet), "map");
```

ptomato (commented):
Finished reviewing the `intersection` tests. Thanks for working on this!

test/built-ins/Set/prototype/difference/require-internal-slot.js
bakkot: Whoops, thanks for the catch. No codegen, just copy-paste.

test/built-ins/Set/prototype/difference/add-not-called.js
```diff
@@ -0,0 +1,27 @@
...
+/*---
+esid: sec-set.prototype.difference
+description: Set.prototype.difference should not call Set.prototype.add
+features: [set-methods]
+includes: [compareArray.js]
+---*/
+
+const s1 = new Set([1, 2]);
+const s2 = new Set([2, 3]);
+const expected = [2];
```
ljharb, line 12:
```suggestion
const expected = [1];
```
right? since 2 is in both sets, "difference" shouldn't include it.

bakkot: Yup, good catch.

ljharb (commented):
I've confirmed all the non-`class` tests in my polyfills for:
 - difference
 - intersection
 - isDisjointFrom
 
Also, isDisjointFrom is missing a test that sets with `0` and `-0` are disjoint from each other (the -0 test on other methods caught a bug, so i checked for it). It's also possible that https://tc39.es/proposal-set-methods/#sec-set.prototype.isdisjointfrom itself is missing an explicit `If nextValue is -0𝔽, set nextValue to +0𝔽.` step after 6.c.ii.1?

I'll post this review now, and will continue reviewing the remaining 4 methods, but since this might be a spec bug I wanted to submit it asap.

test/built-ins/Set/prototype/difference/converts-negative-zero.js
```diff
@@ -0,0 +1,28 @@
...
+info: |
+    7.b.ii If nextValue is -0𝔽, set nextValue to +0𝔽.
+features: [set-methods]
+includes: [compareArray.js]
+---*/
+
+const setlikeWithMinusZero = {
+  size: 1,
+  has: function () {
+    throw new Test262Error("Set.prototype.difference should not call its argument's has method when this.size > arg.size");
```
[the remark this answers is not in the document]
bakkot: I would expect a polyfill to check its receiver's `.size` (or rather to invoke the intrinsic `Set.prototype.size` getter on its receiver) and then branch as in step 6 of `Set.prototype.difference`, so I would in fact expect a polyfill to avoid this line.

test/built-ins/Set/prototype/isDisjointFrom/isDisjointFrom.js
```diff
@@ -0,0 +1,20 @@
+// Copyright (C) 2023 Anthony Frehner and Kevin Gibbons. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
```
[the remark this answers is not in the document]
bakkot: It does not appear to be a duplicate to me?

bakkot:
> t's also possible that https://tc39.es/proposal-set-methods/#sec-set.prototype.isdisjointfrom itself is missing an explicit If nextValue is -0𝔽, set nextValue to +0𝔽. step after 6.c.ii.1?

That's covered by SetDataHas using SameValueZero.

test/built-ins/Set/prototype/difference/converts-negative-zero.js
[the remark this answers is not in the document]
ljharb: Right, but one of the branches seems to require access to the internal data, which a polyfill can’t get.

test/built-ins/Set/prototype/isDisjointFrom/isDisjointFrom.js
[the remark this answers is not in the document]
ljharb: oops, you’re right

test/built-ins/Set/prototype/difference/converts-negative-zero.js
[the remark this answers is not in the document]
bakkot: I'm confused. Both branches in this method, and indeed all of the branches in all of the methods, require reading the `[[SetData]]` from the receiver. Obviously polyfills can't do that directly, but they can use the set prototype intrinsics. That's what I'd expect them to do. There is nothing particularly special about this branch in that regard.

Can you say in more detail exactly why you think this branch requires access to internal slots in a way the other branch does not?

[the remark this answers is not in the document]
ljharb: yep, you're right, it was just that the optimization I'd been using wasn't correct. Sorry for the noise!

ljharb:
Yep, you're right; it was my incorrect optimization. I think the test cases should still be added but I agree the proposal is correct as-is :-)

test/built-ins/Set/prototype/isSupersetOf/set-like-array.js
```diff
@@ -0,0 +1,21 @@
...
+s2.has = function (v) {
+  if (v === 1) return true;
+  if (v === 2) return true;
+  throw new Test262Error("Set.prototype.isSupersetOf should only call its argument's has method with contents of this");
+};
+s2.keys = function () {
+  throw new Test262Error("Set.prototype.isSupersetOf should not call its argument's keys iterator when this.size ≤ arg.size");
+};
+
+assert.sameValue(s1.isSupersetOf(s2), true);
```
ljharb, line 21:
```suggestion
assert.sameValue(s1.isSupersetOf(s2), false);
```
since s1 has 2 items and s2 claims to have 3?

bakkot: Yup, good catch, will fix.

ljharb (approved):
rest of the methods look good

bakkot:
@ljharb Fixed the isSupersetOf/set-like-array.js bug and added the `-0` tests for isSupersetOf and isDisjointFrom, thanks for the catches.

test/built-ins/Set/prototype/isDisjointFrom/set-like-class-mutation.js
```diff
@@ -0,0 +1,38 @@
...
+
+const baseSet = new Set(["a", "b", "c"]);
+
+const evilSetLike = {
+  size: 3,
+  has(v) {
+    if (v === "a") {
+      baseSet.delete("b");
+      baseSet.delete("c");
+      baseSet.add("b");
```
ptomato, line 18: I'm probably overlooking something, but I don't understand what's being tested by deleting and adding `"b"`, as you end up with the same `baseSet` as if you'd only deleted `"c"`. Did you mean for the added element to be `"d"` or something?

ljharb: i assumed it's to test the semantics of visiting after deletion and addition?

ptomato: Me too, but with `"b"` you wouldn't be able to tell the difference if, say, the implementation made a copy of `baseSet`'s initial [[SetData]] and iterated over the correct number of elements. Adding an element that's not `"b"` would make that implementation fail the test.

bakkot: Yeah, from the spec's point of view, deleting and re-adding does not leave you with the same data structure - there's "~empty~" tombstones to mark deleted elements. So the point is just to ensure that the behavior is consistent even if that happens while the method is running.

bakkot: I'm fine with also adding `d` here if you'd like me to though. There's definitely more cases which could be tested, it's just not what I was setting out to test with this.

ptomato: Not a blocker for me, either way. (Some of the other methods' `set-like-class-mutation.js` tests do add elements that weren't in the original [[SetData]], so I think it's unlikely that implementations would be failing this anyway)

test/built-ins/Set/prototype/isDisjointFrom/set-like-class-order.js
```diff
@@ -0,0 +1,195 @@
...
+  const s2 = new MySetLike();
+  const result = s1.isDisjointFrom(s2);
+
+  const expectedOrder = [
+    "getting size",
+    "ToNumber(size)",
+    "getting has",
+    "getting keys",
+    "calling keys",
+    "getting next",
```
ptomato, line 97: I don't see these in the spec - this is still the upper branch (step 5, _thisSize_ ≤ _otherRec_.[[Size]]), yes?

test/built-ins/Set/prototype/isDisjointFrom/set-like-class-order.js
```diff
@@ -0,0 +1,195 @@
...
```
ptomato, line 121: Same.

test/built-ins/Set/prototype/isSubsetOf/set-like-class-order.js
```diff
@@ -0,0 +1,121 @@
...
+  const s1 = new Set(["d", "a"]);
+  const s2 = new MySetLike();
+  const result = s1.isSubsetOf(s2);
+
+  const expectedOrder = [
+    "getting size",
+    "ToNumber(size)",
+    "getting has",
+    "getting keys",
+    // one calls to has
```
ptomato, line 50:
```suggestion
    // one call to has
```

test/built-ins/Set/prototype/symmetricDifference/set-like-class-order.js
```diff
@@ -0,0 +1,166 @@
...
+
+// this is smaller than argument
+{
+  observedOrder = [];
+
+  const s1 = new Set(["a", "d"]);
+  const s2 = new MySetLike();
+  const combined = s1.symmetricDifference(s2);
+
+  const expectedOrder = [
```
ptomato, line 68: Optional nitpick: You could extract `expectedOrder` to the outer scope and use it for all of the cases, that way it's clearer to the reader that they are the same.

test/built-ins/Set/prototype/union/set-like-class-order.js
```diff
@@ -76,8 +57,110 @@ class MySetLike {
...
+// this is smaller than argument
+{
+  observedOrder = [];
 
-assert.compareArray(observedOrder, expectedOrder);
+  const s1 = new Set(["a", "d"]);
+  const s2 = new MySetLike();
+  const combined = s1.union(s2);
+
+  const expectedOrder = [
```
ptomato, line 68: Same.

ptomato (changes requested):
I've reviewed all the tests that weren't already reviewed by someone else, specifically `*class*.js`.

There are a couple of things in `isDisjointFrom/set-like-class-order.js` which I think might be mistakes. Other than that, feel free to ignore the rest of the comments, they're not blocking as far as I'm concerned.

test/built-ins/Set/prototype/isDisjointFrom/set-like-class-order.js
bakkot: Good catch, fixed.

test/built-ins/Set/prototype/isDisjointFrom/set-like-class-order.js
bakkot: Fixed.

test/built-ins/Set/prototype/isSubsetOf/set-like-class-order.js
bakkot: Fixed.

test/built-ins/Set/prototype/symmetricDifference/set-like-class-order.js
bakkot: Good idea, done.

test/built-ins/Set/prototype/union/set-like-class-order.js
bakkot: Done.

ptomato (approved):

ptomato:
Well that was a wild ride! Thanks to everyone involved. Based on our chat conversation I'm merging this now.

PR #3970, opened 2023-12-04
update constructor and Symbol.toStringTag properties on Iterator.prototype

michaelficarra:
Updates tests for these two properties to match https://github.com/tc39/proposal-iterator-helpers/pull/287.

### DIFF
--- a/test/built-ins/Iterator/prototype/Symbol.toStringTag/prop-desc.js
+++ b/test/built-ins/Iterator/prototype/Symbol.toStringTag/prop-desc.js
@@ -1,6 +1,7 @@
 // Copyright (C) 2023 Michael Ficarra. All rights reserved.
 // This code is governed by the BSD license found in the LICENSE file.
 /*---
+esid: sec-iteratorprototype-@@tostringtag
 description: Property descriptor
 info: |
   `Iterator.prototype[@@toStringTag]` is an accessor property with attributes { [[Enumerable]]: *false*, [[Configurable]]: *true* }
--- a/test/built-ins/Iterator/prototype/Symbol.toStringTag/weird-setter.js
+++ b/test/built-ins/Iterator/prototype/Symbol.toStringTag/weird-setter.js
@@ -23,11 +23,13 @@ let { get, set } = Object.getOwnPropertyDescriptor(Iterator.prototype, Symbol.to
 assert.sameValue(Iterator.prototype[Symbol.toStringTag], 'Iterator');
 assert.sameValue(get.call(), 'Iterator');
 
-// 1. Let _O_ be ? RequireObjectCoercible(*this* value).
+// 1. If _this_ is not an Object, then
+//   1. Throw a *TypeError* exception.
 assert.throws(() => set.call(undefined, ''));
 assert.throws(() => set.call(null, ''));
+assert.throws(() => set.call(true, ''));
 
-// 1. If _O_ is _home_, then
+// 1. If _this_ is _home_, then
 //   1. NOTE: Throwing here emulates assignment to a non-writable data property on the _home_ object in strict mode code.
 //   1. Throw a *TypeError* exception.
 assert.throws(() => set.call(IteratorPrototype, ''));
@@ -37,7 +39,7 @@ assert.sameValue(Iterator.prototype[Symbol.toStringTag], 'Iterator');
 assert.sameValue(get.call(), 'Iterator');
 
 // 1. If _desc_ is *undefined*, then
-//   1. Perform ? CreateDataPropertyOrThrow(_O_, _p_, _v_).
+//   1. Perform ? CreateDataPropertyOrThrow(_this_, _p_, _v_).
 let o = {};
 set.call(o, sentinel);
 assert.sameValue(o[Symbol.toStringTag], sentinel);
@@ -46,7 +48,7 @@ assert.sameValue(Iterator.prototype[Symbol.toStringTag], 'Iterator');
 assert.sameValue(get.call(), 'Iterator');
 
 // 1. Else,
-//   1. Perform ? Set(_O_, _p_, _v_, *true*).
+//   1. Perform ? Set(_this_, _p_, _v_, *true*).
 GeneratorPrototype[Symbol.toStringTag] = sentinel;
 assert.sameValue(GeneratorPrototype[Symbol.toStringTag], sentinel);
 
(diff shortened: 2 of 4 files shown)


test/built-ins/Iterator/prototype/Symbol.toStringTag/prop-desc.js
```diff
@@ -1,20 +1,17 @@
 // Copyright (C) 2023 Michael Ficarra. All rights reserved.
 // This code is governed by the BSD license found in the LICENSE file.
 /*---
-es6id: 25.1.2.1
```
linusg, line 4: https://github.com/tc39/test262/blob/main/CONTRIBUTING.md#esid says `esid` is required for all new tests; I'm not sure how strictly that's enforced but might be nice to update this instead of removing it entirely

michaelficarra: Added.

test/built-ins/Iterator/prototype/Symbol.toStringTag/prop-desc.js
```diff
@@ -1,20 +1,18 @@
...
-  configurable: true,
-});
+verifyConfigurable(Iterator.prototype, Symbol.toStringTag);
+verifyNotEnumerable(Iterator.prototype, Symbol.toStringTag);
+
+let desc = Object.getOwnPropertyDescriptor(Iterator.prototype, Symbol.toStringTag);
+assert.sameValue(typeof desc.get, 'function');
+assert.sameValue(typeof desc.set, 'function');
+assert.sameValue(desc.value, undefined);
+assert.sameValue(desc.writable, undefined);
```
ljharb, line 18: do we really want to have tests for the kind of property when we're hoping to change it?

iow, could we test that it Gets the right value and that writing to a subclass when it's frozen works?

michaelficarra: I don't think the test262 tests are going to impact our ability to do that, just like the current tests didn't impact our ability to do the accessors. We just revert this PR when we make the change.

ptomato: For me it depends on when we expect to know whether the future change will happen or not. If we expect it relatively soon, I'd prefer to avoid the churn in order to save maintainer time. If it could be indefinitely long, I'd say we should merge this as is.

test/built-ins/Iterator/prototype/Symbol.toStringTag/weird-setter.js
```diff
@@ -0,0 +1,57 @@
...
+assert.sameValue(o[Symbol.toStringTag], sentinel);
+
+assert.sameValue(Iterator.prototype[Symbol.toStringTag], 'Iterator');
+assert.sameValue(get.call(), 'Iterator');
+
+// 1. Else,
+//   1. Perform ? Set(_this_, _p_, _v_, *true*).
+Object.freeze(IteratorPrototype);
+GeneratorPrototype[Symbol.toStringTag] = sentinel;
+assert.sameValue(GeneratorPrototype[Symbol.toStringTag], sentinel);
```
syg, line 54: `GeneratorPrototype`'s [[Prototype]] has a non-writable own data property named `Symbol.toStringTag`, so this line seems wrong.

h/t @rmahdav

test/built-ins/Iterator/prototype/constructor/weird-setter.js
```diff
@@ -0,0 +1,57 @@
...
+assert.sameValue(o.constructor, sentinel);
+
+assert.sameValue(Iterator.prototype.constructor, Iterator);
+assert.sameValue(get.call(), Iterator);
+
+// 1. Else,
+//   1. Perform ? Set(_this_, _p_, _v_, *true*).
+Object.freeze(IteratorPrototype);
+GeneratorPrototype.constructor = sentinel;
+assert.sameValue(GeneratorPrototype.constructor, sentinel);
```
syg, line 54: `GeneratorPrototype`'s [[Prototype]] has a non-writable own data property `constructor`, so this line seems wrong.

h/t @rmahdav

syg (approved):

ptomato (approved):
Coming back to this after the winter break, I don't see a revert of this on the agenda among the other iterator topics for February 2024, so I think we should assume it's not changing in the short term and should merge it.

test/built-ins/Iterator/prototype/Symbol.toStringTag/prop-desc.js
```diff
@@ -1,20 +1,18 @@
...
 features: [iterator-helpers]
 includes: [propertyHelper.js]
 ---*/
-verifyProperty(Iterator.prototype, Symbol.toStringTag, {
-  value: 'Iterator',
-  writable: true,
-  enumerable: false,
-  configurable: true,
-});
+verifyConfigurable(Iterator.prototype, Symbol.toStringTag);
```
iamstolis, line 11: This test and the corresponding one for `constructor` property are broken. `verifyConfigurable()` calls `isConfigurable()` that [deletes](https://github.com/tc39/test262/blob/main/harness/propertyHelper.js#L122) the property. So, the subsequent parts of this test fail because the property is not there anymore.

If you want to use `verifyConfigurable()` then it should be the last statement of the test.

test/built-ins/Iterator/prototype/Symbol.toStringTag/weird-setter.js
```diff
@@ -0,0 +1,57 @@
...
+let sentinel = 'a';
+
+let { get, set } = Object.getOwnPropertyDescriptor(Iterator.prototype, Symbol.toStringTag);
+
+assert.sameValue(Iterator.prototype[Symbol.toStringTag], 'Iterator');
+assert.sameValue(get.call(), 'Iterator');
+
+// 1. If _this_ is not an Object, then
+//   1. Throw a *TypeError* exception.
+assert.throws(() => set.call(undefined, ''));
```
iamstolis, line 27: This test and the corresponding one for `constructor` property are broken. The first argument of `assert.throws()` [should be](https://github.com/tc39/test262/blob/main/harness/assert.js#L71) the expected error (i.e. `TypeError` in this case).

ptomato:
@iamstolis Thanks for catching these, would you mind opening an issue, otherwise the comments will probably get lost :smile:

iamstolis:
> @iamstolis Thanks for catching these, would you mind opening an issue, otherwise the comments will probably get lost 😄

Done: #3993.

PR #3971, opened 2023-12-06
Add tests for Stage 3 Decorator Metadata

rbuckton:
This adds tests for the Stage 3 [Decorator Metadata](https://github.com/tc39/proposal-decorator-metadata) proposal, per the specification PR at https://github.com/pzuraq/ecma262/pull/10.

/cc: @pzuraq


test/language/expressions/class/decorator/metadata/metadata-prop-desc.js
```diff
@@ -0,0 +1,25 @@
...
+var contextObj;
+function dec(_, context) {
+  contextObj = context;
+}
+
+let C = @dec class C {};
+assert.sameValue(typeof contextObj.metadata, "object");
+verifyNotEnumerable(contextObj, "metadata");
+verifyWritable(contextObj, "metadata");
+verifyConfigurable(contextObj, "metadata");
```
ljharb, line 25: verifyProperty

test/language/statements/class/decorator/metadata/metadata-attached-when-class-is-decorated.js
```diff
@@ -0,0 +1,43 @@
...
+features: [decorators, decorator-metadata]
+---*/
+
+function dec() {}
+
+@dec class C1 {}
+assert.sameValue(typeof C1[Symbol.metadata], "object");
+
+class C2 { @dec method() {} }
+assert.sameValue(typeof C2[Symbol.metadata], "object");
```
ljharb, line 40: null?

test/language/statements/class/decorator/metadata/metadata-prop-desc.js
```diff
@@ -0,0 +1,25 @@
...
+var contextObj;
+function dec(_, context) {
+  contextObj = context;
+}
+
+@dec class C {}
+assert.sameValue(typeof contextObj.metadata, "object");
+verifyNotEnumerable(contextObj, "metadata");
+verifyWritable(contextObj, "metadata");
+verifyConfigurable(contextObj, "metadata");
```
ljharb, line 25: verifyProperty

test/built-ins/Symbol/metadata/prop-desc.js
```diff
@@ -0,0 +1,17 @@
...
+    This property has the attributes { [[Writable]]: false, [[Enumerable]]:
+    false, [[Configurable]]: false }.
+includes: [propertyHelper.js]
+features: [decorator-metadata]
+---*/
+
+assert.sameValue(typeof Symbol.metadata, 'symbol');
+verifyNotEnumerable(Symbol, 'metadata');
+verifyNotWritable(Symbol, 'metadata');
+verifyNotConfigurable(Symbol, 'metadata');
```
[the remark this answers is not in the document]
rbuckton: Testing that the value of `Symbol.metadata` is `Symbol.metadata` is a bit reductive as it would only be testing that `Object.getOwnPropertyDescriptor` works. I'll make the switch to `verifyProperty`, though.

ljharb (approved):
seems good! would be nice to have another review before merging.

test/language/expressions/class/decorator/metadata/metadata-inheritance-not-affected-when-base-metadata-replaced.js
```diff
@@ -0,0 +1,46 @@
+// Copyright (C) 2023 Ron Buckton. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-runtime-semantics-classdefinitionevaluation
+description: >
+    Metadata on a derived class inherits from the metadata of the declared super class.
```
ptomato, line 7: Suggestion, the description could be a bit more specific here, more like the title of the test: maybe something like "Metadata on a derived class inherits from the original metadata of the declared super class, unaffected by overwriting"

test/language/expressions/class/decorator/metadata/metadata-inheritance.js
```diff
@@ -0,0 +1,43 @@
+// Copyright (C) 2023 Ron Buckton. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-runtime-semantics-classdefinitionevaluation
+description: >
+    Metadata is only attached when a class or class element is decorated.
```
ptomato, line 7: This description seems copied from somewhere else, maybe something like "Metadata inherits from the metadata of the super class"?

test/language/expressions/class/decorator/metadata/metadata-is-plain-object.js
```diff
@@ -0,0 +1,32 @@
...
+      c. Set metadataObj to OrdinaryObjectCreate(metadataParent).
+    [...]
+features: [decorators, decorator-metadata]
+---*/
+
+function dec() {}
+
+let C = @dec class C {};
+const metadata = C[Symbol.metadata];
+assert(Object.isExtensible(metadata));
```
ptomato, line 32:
```suggestion
assert(Object.isExtensible(metadata), "Metadata is not frozen");
```
(I suggest using assertion messages at least in plain `assert()`, otherwise if it fails you get "Expected false to be true" or something like that)

ptomato: Should we additionally verify that the object is a plain object, something like this?
```js
assert.sameValue(Object.getPrototypeOf(metadata), Object.prototype, "Metadata's prototype is Object.prototype");
```

test/language/expressions/class/decorator/metadata/context-metadata-prop-desc.js
```diff
@@ -0,0 +1,27 @@
...
+
+var contextObj;
+function dec(_, context) {
+  contextObj = context;
+}
+
+void @dec class C {};
+assert.sameValue(typeof contextObj.metadata, "object");
+verifyProperty(contextObj, 'metadata', {
+  enumerable: false,
```
ptomato, line 24: By my reading of the spec, this seems like it should be true? In CreateDecoratorContextObject we create the property with CreateDataPropertyOrThrow, which calls [CreateDataProperty](https://tc39.es/ecma262/#sec-createdataproperty), which creates enumerable properties.

ptomato (changes requested):
Thanks for doing this!

The only blocker would be the context metadata property descriptor (if I'm correct about that.) I had a few other suggestions for improvements elsewhere.

Note that all the comments on `expressions/class/decorator/metadata/` apply to the corresponding tests in `statements/class/decorator/metadata/` as well.

PR #3972, opened 2023-12-18
Update single-line-html-close.js

lgalfaso:
Fix comment with the description of the syntax


ptomato (approved):
Thanks, yes this matches the spec :smile:

PR #3973, opened 2023-12-22
Add 402 staging tests for Islamic and Persian calendars

justingrant:
Tests the changes in https://github.com/tc39/proposal-temporal/pull/2743.

### DIFF
--- a/test/staging/Intl402/Temporal/old/islamic-calendars.js
+++ b/test/staging/Intl402/Temporal/old/islamic-calendars.js
@@ -1,0 +1,91 @@
+// Copyright (C) 2023 Justin Grant. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal-intl
+description: Islamic calendars (note there are 6 variants)
+features: [Temporal]
+---*/
+
+// Test data obtained from ICU
+
+const tests = [
+  {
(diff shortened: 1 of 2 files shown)


ptomato (commented):
These should probably cover `dayOfYear` and `monthsInYear` as well.

Do you have a reference for where the test data comes from? I imagine that'd be helpful to implementors who are not passing these tests. I've tried looking at various Islamic calendar websites, but it's hard to tell which variant they are using.

justingrant:
> These should probably cover `dayOfYear` and `monthsInYear` as well.

OK, I pushed an update that adds both. Note that `monthsInYear` is trivial because all non-lunisolar calendars have a constant number of months in every year. Both of these calendars have 12 months per year.

> Do you have a reference for where the test data comes from? I imagine that'd be helpful to implementors who are not passing these tests. I've tried looking at various Islamic calendar websites, but it's hard to tell which variant they are using.

I just pulled the data from Intl.DateTimeFormat in Chrome, which in turn gets its data from ICU. If ICU is wrong, then figuring that out is beyond my ability to figure it out, because calendar implementations (especially those like the islamic calendar that involve lunar measurements and/or complex calculations) are beyond my knowledge.

ptomato (approved):
OK, thanks. I've made a note in the files that the data comes from ICU.

PR #3978, opened 2024-01-02
Add tests for precise results in Duration.p.total and ZonedDateTime.p.hoursInDay

anba:
The existing tests didn't cover some edge cases where implementations have to compute the exact result of `numerator / denominator`, where at least one of `numerator` and `denominator` can't be exactly represented by an IEEE-754 double precision floating point value.

"precision-exact-mathematical-values-5.js" gets added in #3961, so the new tests from this commit start at "precision-exact-mathematical-values-6.js".


ptomato (approved):
Thanks as always for the number precision tests.

I've checked these against the state of the Temporal proposal including tc39/proposal-temporal#2727, which is not merged yet. So I'll close this PR and have absorbed it into #3961.

PR #3980, opened 2024-01-04
Add tests for Source Phase Imports

legendecas:
Add tests for [Source Phase Imports](https://github.com/tc39/proposal-source-phase-imports), [ecma262 PR](https://github.com/tc39/ecma262/pull/3094).


test/language/expressions/dynamic-import/usage/nested-arrow-assignment-expression-import-source-returns-thenable.js
```diff
@@ -0,0 +1,32 @@
...
+---*/
+
+let f = () => import.source('./dynamic-import-module_FIXTURE.js').then(imported => {
+
+  assert.sameValue(typeof imported, 'object');
+
+});
+
+
+f().then($DONE, $DONE).catch($DONE);
```
nicolo-ribaudo, line 32: The proposal mandates that importing the source of a JS module always throws, so the `assert` is never reached here. There is no way to test that `imported` in an object in test262, and it should instead be done in WPT where there is a known valid module source (i.e. wasm).

src/dynamic-import/import-source-script-code-valid.case
```diff
@@ -0,0 +1,16 @@
...
+/*---
+desc: import() can be used in script code
+template: syntax/valid
+features: [source-phase-imports]
+---*/
+
+//- setup
+// This is still valid in script code, and should not be valid for module code
+// https://tc39.github.io/ecma262/#sec-scripts-static-semantics-lexicallydeclarednames
+var smoosh; function smoosh() {}
```
nicolo-ribaudo, line 13: What do we need this for?

src/dynamic-import/import-source-script-code-valid.case
```diff
@@ -0,0 +1,16 @@
+// Copyright (C) 2024 Chengzhong Wu. All rights reserved.
+// Copyright (C) 2018 Leo Balter. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+desc: import() can be used in script code
```
nicolo-ribaudo, line 5:
```suggestion
desc: import.source() can be used in script code
```

src/dynamic-import/import-source-specifier-tostring.case
```diff
@@ -0,0 +1,46 @@
+// Copyright (C) 2024 Chengzhong Wu. All rights reserved.
```
nicolo-ribaudo, line 1: This file doesn't test `import.source`.

test/language/module-code/abstract-module-source/length.js
```diff
@@ -0,0 +1,28 @@
...
+  28.1.1.1 %AbstractModuleSource% ( )
+
+  The "length" property of this function is +0𝔽.
+
+includes: [propertyHelper.js]
+features: [source-phase-imports]
+flags: [module]
+---*/
+
+import source moduleSource from './name.js';
```
nicolo-ribaudo, line 17: This line throws a linking error, because JS modules do not have a source. We might ask hosts to expose `AbstractModuleSource` as part of the `$262` harness (`$262.AbstractModuleSource`). There is no other way within ECMA-262 to reach it.

nicolo-ribaudo (commented):
We should also add a test to verify that the `import source` declaration syntax is supported:

```js
// test.js
import source x from "x";
import "./ensure-linking-error_FIXTURE.js";
```
throws in the `resolution` phase and not in the `parse` phase (e.g. https://github.com/tc39/test262/blob/main/INTERPRETING.md#negative)

where that fixture file is https://github.com/tc39/test262/blob/main/test/language/module-code/import-attributes/ensure-linking-error_FIXTURE.js

--

We should also test that Source Text Module Records do not have a source representation:
```js
assertRejects(import("./a-js-file_FIXTURE.js"), ReferenceError);
```

```js
// negative:
// - phase: resolution
// - type: ReferenceError
import source s from "./a-js-file_FIXTURE.js";
```

nicolo-ribaudo:
cc @guybedford

src/dynamic-import/import-source-script-code-valid.case
legendecas: This is used to ensure that this file is been parsed as a script rather than a module. The file verifies that an `import.source()` call can be used in script code.

test/language/expressions/dynamic-import/usage/nested-arrow-assignment-expression-import-source-returns-thenable.js
legendecas: Updated with negative verifications.

test/language/module-code/abstract-module-source/length.js
legendecas: Thanks for the suggestion. Updated!

src/dynamic-import/import-source-script-code-valid.case
legendecas: This case is adapted from https://github.com/tc39/test262/blob/main/src/dynamic-import/script-code-valid.case. I believe the intention here is to disallow this file to be accidentally evaluated as a module code and pass the test as a false positive.

guybedford (commented):
Appreciated for your patience with the review. Finally had a look through and it's seeming really thorough to me.

Perhaps we should move `abstract-module-source` and its prototype tests to test/built-ins?

The invalid and valid tests could also include a test that `import.source` cannot be assigned. See some of the import.meta tests for examples. We could also ensure `import.source.x`, `typeof import.source` etc remains invalid.

I would note that we are still aiming to specify import.source for ECMAScript SourceText module records, so that reducing the number of tests that would need to change / be removed when that happens would be useful. Rather than the suite of these failures in the catch tests I think just a few would suffice if they will be changed anyway.

Finally for the static import syntax tests, it might be beneficial to also include some of the valid cases of the `source` and `from` identifier names being imported for the `source` import phase per the various discussions we had around these syntax cases.

legendecas:
@guybedford thanks for the suggestions, updated!

legendecas:
This PR doesn't include import-attributes with source phase imports. Once it is normatively specified, the test can be added as well.

guybedford (approved):

test/language/module-code/source-phase-import/import-source-newlines.js
```diff
@@ -0,0 +1,31 @@
...
+negative:
+  phase: resolution
+  type: SyntaxError
+features: [source-phase-imports]
+flags: [module]
+---*/
+
+$DONOTEVALUATE();
+
+import source x from '';
```
nicolo-ribaudo, line 25: Why do we need this first one?

test/language/module-code/source-phase-import/import-source-binding-name.js
```diff
@@ -0,0 +1,22 @@
...
+negative:
+  phase: resolution
+  type: SyntaxError
+features: [source-phase-imports]
+flags: [module]
+---*/
+
+$DONOTEVALUATE();
+
+import source source from '';
```
nicolo-ribaudo, line 20: There is no requirement that an empty specifier will fail resolving. Can we use something `from '<do not resolve>'`, and explicitly document it in https://github.com/tc39/test262/blob/main/INTERPRETING.md?

legendecas: Updated with two special specifier:

- `<do not resolve>`: which should be resolved with a TypeError throw completion.
- `<abstract module source>`: which should be resolved with an instance of %AbstractModuleSource%`, as there is not alternative ModuleSource in the ecma262 at the moment.

INTERPRETING.md
```diff
@@ -160,6 +161,12 @@ located at `test/language/import/nested/dep.js`.
 
 Files bearing a name ending in `.json` are intended to be interpreted as JSON.
 
+Implementers should resolve a specifier of `<do not resolve>` with a TypeError
+throw completion.
+
+Implementers should resolve a specifier of `<abstract module source>` with an
+instance of `%AbstractModuleSource%`.
```
nicolo-ribaudo, line 168:
```suggestion
Implementers should resolve the specifier `<abstract module source>` to a module that provides a valid [Module Source](https://tc39.es/proposal-source-phase-imports/#sec-module-source-objects), such as a [WebAssembly module](https://webassembly.github.io/esm-integration/js-api/index.html#webassembly-module-record).
```

And also explicitly tag those tests with a flag, such as `source-phase-imports, source-phase-imports-source` with a comment that only engines that can provide at least one module type that has a source can run `source-phase-imports-source`.

legendecas: Updated!

INTERPRETING.md
```diff
@@ -160,6 +161,12 @@ located at `test/language/import/nested/dep.js`.
 
 Files bearing a name ending in `.json` are intended to be interpreted as JSON.
 
+Implementers should resolve the specifier `<module source>` to a module that
+provides a valid [Module Source](https://tc39.es/proposal-source-phase-imports/#sec-module-source-objects),
+such as a [WebAssembly module](https://webassembly.github.io/esm-integration/js-api/index.html#webassembly-module-record).
+Tests use `<module source>` specifier should be guarded with a feature flag
```
nicolo-ribaudo, line 167:
```suggestion
Tests use `<module source>` specifier are guarded with a feature flag
```
Given that this is guidance for who uses the tests and not for who writes them.

nicolo-ribaudo (approved):
This looks good now, thank you! :)

Ms2ger (approved):

INTERPRETING.md
```diff
@@ -107,6 +107,7 @@ properties of the global scope prior to test execution.
     - **`sleep`** - a function that takes a millisecond argument and
         sleeps the execution for approximately that duration.
     - **`monotonicNow`** - a function that returns a value that conforms to [`DOMHighResTimeStamp`][] and is produced in such a way that its semantics conform to **[Monotonic Clock][]**.
+    - **`AbstractModuleSource`** - a reference to the `%AbstractModuleSource%` constructor which does not appear as a property of the global object.
```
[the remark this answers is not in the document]
legendecas: The constructor function should be a per realm value. It is not exposed via the realm global object, so we need hosts to expose it via test262 facilities in order to test it.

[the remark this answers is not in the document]
legendecas: Ah, nice catch! I think it should be `$262.AbstractModuleSource`.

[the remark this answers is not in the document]
ptomato: Could you open an issue for this? Otherwise it'll get forgotten.

PR #3982, opened 2024-01-08
add the TypedArray feature flag to some tests which were missing it

bakkot:
Per https://github.com/tc39/test262/pull/3849#issuecomment-1881159459.


ljharb (approved):

ptomato (approved):
I didn't check exhaustively if it was missing from more places, but if this gets #3849 to work then :+1:

Issue #3786, opened 2023-02-15
Missing guidelines/patterns for locale-sensitive testing


gibson042:
Nearly all ECMA-402 APIs (and some in ECMA-262) have implementation- and locale-dependent ("ILD") behavior, and those details change over time (e.g. with new browser customizations or CLDR/ICU releases). Relying on "golden output" is therefore brittle, but we also want to achieve good coverage.

There are options such as looking at stable substrings or other parts of output (such as month names), comparative testing across related inputs (e.g. that each setting for an input option produces a distinct output), general [metamorphic testing](https://www.hillelwayne.com/post/metamorphic-testing/), and broader decisions to make such as the set of locales and locale options to utilize (and whether or not to always utilize all of them).

Guidance should be added to CONTRIBUTING.md in the form of both explanations and examples. A proposed first step in establishing such guidance is to start with a relatively simple ILD API, and [Intl.RelativeTimeFormat](https://tc39.es/ecma402/#relativetimeformat-objects) seems like a good candidate because it supports a small number of options (just **numberingSystem**, **style** "long" vs. "short" vs. "narrow", and **numeric** "always" vs. "auto") and its formatted output includes both numeric and textual parts (the former of which depends upon and can therefore be compared against Intl.NumberFormat functionality).

littledan:
How does shifting the tests to wpt help?

ptomato:
I guess the context for this question was [my comment in this WPT pull request](https://github.com/web-platform-tests/wpt/pull/41760#issuecomment-1703256661)? I don't think we need to shift tests to WPT as such, but since we are trying to find new strategies to test ECMA-402 without relying on golden output, I'd say that if people specifically _want_ tests that rely on golden output, test262 is not the right place for that — because those tests do not accurately measure conformance with ECMA-262 or ECMA-402. Stability in ILD strings seems like it's mainly important to engines that run in browsers, so WPT seems like it could be the right place, but that's ultimately up to the people with merge privileges in WPT.

annevk:
How is it not relevant to other engines?

ptomato:
One place where compatibility issues arise with ILD string changes is from serializing and rehydrating web pages. As an example of an engine where that isn't a concern: I'm involved with the embedded SpiderMonkey in the GNOME desktop, for example. We are not serializing web pages there, but just displaying the ILD strings directly to users, in which case fixes that make the ILD output more consistent with what users in that locale expect are usually welcome.

annevk:
[...]

PR #3983, opened 2024-01-09
Update Intl.DurationFormat tests

anba:
Update `Intl.DurationFormat` tests to match the current draft spec and add tests to cover:
- https://github.com/tc39/proposal-intl-duration-format/pull/183
- https://github.com/tc39/proposal-intl-duration-format/pull/184

### DIFF
--- a/harness/testIntl.js
+++ b/harness/testIntl.js
@@ -2509,12 +2509,73 @@ function isCanonicalizedStructurallyValidTimeZoneName(timeZone) {
 /**
  * @description Simplified PartitionDurationFormatPattern implementation which
  * only supports the "en" locale.
+ * @param {Object} durationFormat the duration format object
  * @param {Object} duration the duration record
- * @param {String} style the duration format style
  * @result {Array} an array with formatted duration parts
  */
 
-function partitionDurationFormatPattern(duration, style = "short") {
+function partitionDurationFormatPattern(durationFormat, duration) {
+  function durationToFractional(duration, exponent) {
+    let {
+      seconds = 0,
+      milliseconds = 0,
+      microseconds = 0,
+      nanoseconds = 0,
+    } = duration;
+
+    // Directly return the duration amount when no sub-seconds are present.
+    switch (exponent) {
+      case 9: {
+        if (milliseconds === 0 && microseconds === 0 && nanoseconds === 0) {
+          return seconds;
+        }
+        break;
+      }
+      case 6: {
+        if (microseconds === 0 && nanoseconds === 0) {
+          return milliseconds;
+        }
+        break;
+      }
+      case 3: {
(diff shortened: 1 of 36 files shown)


ptomato (commented):
For future reference, I'm not at all a fan of the approach of essentially putting a polyfill into the `testIntl` helper file and comparing the results from the polyfill with the results from the implementation. It makes these tests opaque, difficult to understand what's being tested, and unclear whether the polyfill or the implementation is at fault when the tests fail.

However, the tests were already set up that way, so in the absence of clear recommendations about what to do otherwise (#3786) I would say that changing it is out of scope for this PR.

Nonetheless I would suggest that numeric-hour-with-zero-minutes-and-non-zero-seconds.js was clearer using NumberFormat and ListFormat directly, as it previously did, so maybe we should drop the commit ' Use formatDurationFormatPattern for "/numeric-hour-with-zero-minutes-and-non-zero-seconds"'.

Could @ryzokuken, @romulocintra, or @ben-allen take a look at the polyfill changes please?

anba:
> Nonetheless I would suggest that numeric-hour-with-zero-minutes-and-non-zero-seconds.js was clearer using NumberFormat and ListFormat directly, as it previously did, so maybe we should drop the commit ' Use formatDurationFormatPattern for "/numeric-hour-with-zero-minutes-and-non-zero-seconds"'.

This isn't actually possible, because of <https://github.com/tc39/proposal-intl-duration-format/pull/180>. Before that PR, the test expected the string `"1, 03"`, so the `hours` and `seconds` duration units were number-formatted and then concatenated through the list formatter. After the PR, the expected string is `"1:00:03"`, so the list formatter is no longer used at all. The only alternative here is to hardcode the string `"1:00:03"` as the expected result and don't use a number nor a list formatter.

ben-allen (approved):
oh my goodness, thank you for catching this. I'm actively embarassed...

ben-allen (approved):
It's all great -- thank you for all of your work here.

Thoughts in general:
* I'm wondering if it's worthwhile to have a TG2 discussion on the behaviour when seconds (or the relevant subsecond unit) has its display option set to `"auto"` and has a non-zero fractional component that's nevertheless not displayed due to use of the `fractionalDigits` option. I can see arguments for changing this behaviour, but also arguments for leaving it as-is
* I share @ptomato's skepticism about the use of the `partitionDurationFormatPattern` quasi-polyfill. That said, given that these tests are grea and given that DurationFormat is nearly done, I'm wondering if it would be useful for me to extend `partitionDurationFormatPattern` to handle unit separators other than : and numbering systems other than `latn`, in the interest of testing locales other than `"en"`.

Ms2ger (approved):

PR #3985, opened 2024-01-11
Fix an error in expected value of difference set method 

rmahdav:
The difference between two sets is `[1]`.


ljharb (approved):
good catch

ptomato (approved):
I agree.

bakkot (approved):

PR #3987, opened 2024-01-11
Do not expect specific error type in `import-attributes/allow-nlt-before-with.js`

nicolo-ribaudo:
Hosts are free to throw whatever error they want during module loading: there is no guarantee that the error thrown due to the import with `type: "foo"` is a SyntaxError (it could also be, for example, an `Error` caused by a network failure). Technically this is also true when loading just JS, but not specifying `type` reduces variance and in practice all test262 runners will not fail when loading JS modules that exist :)

### DIFF
--- a/test/language/module-code/import-attributes/allow-nlt-before-with.js
+++ b/test/language/module-code/import-attributes/allow-nlt-before-with.js
@@ -19,7 +19,6 @@ info: |
 
 negative:
   phase: resolution
-  type: SyntaxError
 features: [import-attributes, globalThis]
 flags: [module, raw]
 ---*/
--- a/test/language/module-code/import-attributes/allow-nlt-before-with.js
+++ b/test/language/module-code/import-attributes/allow-nlt-before-with.js
@@ -19,6 +19,7 @@ info: |
 
 negative:
   phase: resolution
+  type: SyntaxError
 features: [import-attributes, globalThis]
 flags: [module, raw]
 ---*/
@@ -29,4 +30,4 @@ import "./ensure-linking-error_FIXTURE.js";
 
 import * as x from './import-attribute-1_FIXTURE.js'
 with
-{ type: 'foo' };
+{};


nicolo-ribaudo:
Mhh, regarding the linting error -- how can I say "this test must throw _something_"? It's guaranteed that it will throw due to line 28, but the error type is unknown due to lines 30-32.

ptomato:
Is my understanding here correct? Thinking of HTML hosts as an example, `type` is a supported import attribute, so the SyntaxError will not be thrown in InnerModuleLoading step 2.d.i.1. Instead a TypeError will be thrown in onSingleFetchComplete step 2, failing the test.
After this change, the TypeError in onSingleFetchComplete step 2 is not thrown, and instead the line `import "./ensure-linking-error_FIXTURE.js";` throws the SyntaxError.

However, how do we distinguish between the SyntaxError from `import "./ensure-linking-error_FIXTURE.js";` and the SyntaxError that would be thrown if the import declaration were instead parsed as an import declaration + `with`-statement, which would be a SyntaxError in module code?

nicolo-ribaudo:
> Is my understanding here correct? Thinking of HTML hosts as an example, type is a supported import attribute, so the SyntaxError will not be thrown in InnerModuleLoading step 2.d.i.1. Instead a TypeError will be thrown in onSingleFetchComplete step 2, failing the test.
After this change, the TypeError in onSingleFetchComplete step 2 is not thrown, and instead the line import "./ensure-linking-error_FIXTURE.js"; throws the SyntaxError.

This is all correct. Before this PR, hosts _might_ throw an error due to the import attributes (either because they don't support `type`, or `

Before this PR, there where three possible cases:
- the host does not support the `type` attribute, so InnerModuleLoading throws a SyntaxError
- the host doesn't know how to handle that specific type, so it throws _some error_ in HostLoadImporterModule
- loading works, so we ensure a SyntaxError error in `ensure-linking-error_FIXTURE.js`

This PR removes the first two cases, ensuring a consistent error.

If a host doesn't parse this properly, it would throw in the parse phase rather than in the resolution phase.

ptomato (approved):
OK.
I was going to ask, why don't we remove the `ensure-linking-error_FIXTURE.js` import, but then we don't actually have a way to say "this test is only for the parsing phase" other than causing an intentional error in the resolution phase. So, fine with me.

Issue #3387, opened 2022-01-30
Tests for  Async-from-Sync Iterator changes


mhofman:
https://github.com/tc39/ecma262/pull/2600 reached consensus at the last plenary. It updates  Async-from-Sync Iterator objects to close the wrapped iterator in cases where the async wrapper is considered to be done, but the sync iterator isn't.

I included some examples in the presentation of where this behavior shows up, and there were a few relevant cases in the linked issue. I plan to adapt those into test262 cases, but I may not have time immediately, so creating this issue to keep track of the tests needed.

ioannad:
@mhofman I plan to be working on these tests, here is the test plan I have so far:

## Testing plan

- For the changes to `%AsyncFromSyncIteratorPrototype%.throw` (steps 8.a-8.h), check every codepath when closing the iterator for a null iterator throw method. Using sync iterators with null `throw` methods, and:
    - A return method that does something observable (like a finally). 
        - [ ] Test that the return method was called and the iterator closed before returning a new `TypeError`.
    - A return method which throws. 
        - [ ] Test this case returns a new `TypeError`.
    - Should not define a return method. 
        - Covered by existing test262 test [throw-null.js](https://github.com/tc39/test262/blob/main/test/built-ins/AsyncFromSyncIteratorPrototype/throw/throw-null.js) (IteratorClose throws a TypeError). 
        - [ ] Modify existing test to assert we ended with a new `TypeError`.

- For the changes to `AsyncFromSyncIteratorContinuation` (steps 6-6.a.), the new codepaths are reached only when `closeOnRejection` is`true`, and `done` is `false`. All tests here must ensure the iterator was closed, so the test sync iterators should have `return` methods that do something observable (like a finally).
    - Test that valueWrapper threw and the iterator was closed in the following scenarios:
        - Entering via `%AsyncFromSyncIteratorPrototype%.next` :
            - [ ] Using a sync iterator whose `next` method throws.
            - [ ] Using a sync iterator whose `next` method is null.
        - Entering via `%AsyncFromSyncIteratorPrototype%.throw` :
            - [ ] Using a sync iterator whose `throw` method is an abrupt completion. 
            - [ ] Using a sync iterator whose `throw` method is defined (gettable) but returns null.

[...]

PR #3990, opened 2024-01-15
Test that the sync iterator that returns a rejected promise closes.

ioannad:
This adds tests for the normative changes in https://github.com/tc39/ecma262/pull/2600 for the last block of new steps in  AsyncFromSyncIteratorContinuation.

These tests have been tested with [this PR](https://github.com/engine262/engine262/pull/230) of engine262, which implements the normative changes.

3rd of 3 PRs that should finish addressing #3387 .


ptomato (approved):
Looks correct to me. I was thinking of suggesting using `assert.throwsAsync()` but ultimately I think this is fine because it's clearer with the for-await loop.

Would be nice to get some eyes on it from some of the people involved in the normative PR, e.g. @bakkot, @mhofman

ptomato (approved):

PR #3991, opened 2024-01-19
Temporal assertion message formatting

ptomato:
Some improvements to the formatting of assertion messages for the various Temporal types. This should make assertion failures easier to understand.

### DIFF
--- a/harness/temporalHelpers.js
+++ b/harness/temporalHelpers.js
@@ -61,17 +61,18 @@ var TemporalHelpers = {
    * an expected value.
    */
   assertDuration(duration, years, months, weeks, days, hours, minutes, seconds, milliseconds, microseconds, nanoseconds, description = "") {
-    assert(duration instanceof Temporal.Duration, `${description} instanceof`);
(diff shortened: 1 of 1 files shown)


harness/temporalHelpers.js
```diff
@@ -101,7 +103,7 @@ var TemporalHelpers = {
    * the corresponding field in another Temporal.Duration.
    */
   assertDurationsEqual(actual, expected, description = "") {
-    assert(expected instanceof Temporal.Duration, `${description} expected value should be a Temporal.Duration`);
+    assert(expected instanceof Temporal.Duration, `${description}: expected value should be a Temporal.Duration`);
```
Ms2ger, line 106: Why not do the prefix dance here? Also e.g. `assertInstantsEqual`

Ms2ger (approved):

harness/temporalHelpers.js
ptomato: Thanks for catching that.  I think I just overlooked them.

PR #3996, opened 2024-01-25
fix broken Iterator.prototype constructor/Symbol.toStringTag tests

### LINKED ISSUE #3993: Tests of Iterator.prototype.constructor/toStringTag are broken
#3970 updated the following tests (in order to reflect a recent normative change)
```
built-ins/Iterator/prototype/Symbol.toStringTag/prop-desc.js
built-ins/Iterator/prototype/Symbol.toStringTag/weird-setter.js
built-ins/Iterator/prototype/constructor/prop-desc.js
built-ins/Iterator/prototype/constructor/weird-setter.js
```

Unfortunately, these tests are broken now.

`prop-desc.js` tests use `verifyConfigurable()` that deletes the tested property. So, the subsequent assertions fail. `verifyConfigurable()` must be the last statement of the test (if used at all).

`weird-setter.js` tests pass too few arguments to `assert.throws()` i.e. they fail to specify the type of the expected error.

michaelficarra:
Fixes #3993.

### DIFF
--- a/test/built-ins/Iterator/prototype/Symbol.toStringTag/prop-desc.js
+++ b/test/built-ins/Iterator/prototype/Symbol.toStringTag/prop-desc.js
@@ -8,11 +8,10 @@ info: |
 features: [iterator-helpers]
 includes: [propertyHelper.js]
 ---*/
-verifyConfigurable(Iterator.prototype, Symbol.toStringTag);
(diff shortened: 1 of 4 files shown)


ljharb (approved):

ptomato (approved):
Thanks!

PR #3999, opened 2024-02-01
Temporal: Prevent arbitrary loops in NormalizedTimeDurationToDays

ptomato:
Adapts the tests that checked arbitrarily long loops, to now check that an exception is thrown if the loop would happen.

Also updates some step numbers in related tests.

Tests accompany the already-approved normative change https://github.com/tc39/proposal-temporal/pull/2612


Aditi-1400 (approved):
👍

PR #4001, opened 2024-02-03
Fix comment in allow-nlt-before-with.js

lgalfaso:
The description of the test states that `WithClause` can be preceded by a line terminator, but the syntax states the opposite.

### DIFF
--- a/test/language/module-code/import-attributes/allow-nlt-before-with.js
+++ b/test/language/module-code/import-attributes/allow-nlt-before-with.js
@@ -7,7 +7,7 @@ description: >
 esid: sec-modules
 info: |
   ImportDeclaration:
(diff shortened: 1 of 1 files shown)


nicolo-ribaudo (changes requested):
There is no restriction on line terminators in the proposal spec: https://tc39.es/proposal-import-attributes/#sec-imports. Could you update the syntax comment instead?

lgalfaso:
Updated the syntax comment.

nicolo-ribaudo (approved):
Thank you!

nicolo-ribaudo:
CI failure is unrelated.

ptomato (approved):
Thanks!

PR #4002, opened 2024-02-06
Add tests for `-->` in the first line

nicolo-ribaudo:
Tests for https://github.com/tc39/ecma262/pull/3244, which reached consensus today


ljharb (approved):

nicolo-ribaudo:
@leobalter suggested to add a `throw` to make sure that the test isn't passing just because _everything_ is considered to be a comment. This is also done in all the other HTML comments tests.

Ms2ger (approved):

woess:
These tests should not have `flags: [raw]` since they're using `Test262Error`.

ptomato:
Thanks for opening the issue! I was afraid it would get lost here.

PR #4003, opened 2024-02-07
Add tests for resizable array buffer views

anba:
Covers spec compliance issues in at least one of JSC, V8, or SM.

### DIFF
--- a/test/built-ins/TypedArray/prototype/values/make-in-bounds-after-exhausted.js
+++ b/test/built-ins/TypedArray/prototype/values/make-in-bounds-after-exhausted.js
@@ -1,0 +1,55 @@
+// Copyright (C) 2024 André Bargull. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-%typedarray%.prototype.values
+description: >
+  Iterator is still exhausted when typedarray is changed to in-bounds.
+features: [TypedArray, resizable-arraybuffer]
+---*/
+
+let rab = new ArrayBuffer(3, {maxByteLength: 5});
+let ta = new Int8Array(rab);
+
+// Ensure the TypedArray is correctly initialised.
+assert.sameValue(ta.length, 3);
+assert.sameValue(ta.byteOffset, 0);
+
+ta[0] = 11;
+ta[1] = 22;
+ta[2] = 33;
+
+let it = ta.values();
+let r;
+
+// Fetch the first value.
+r = it.next();
+assert.sameValue(r.done, false);
+assert.sameValue(r.value, 11);
+
+// Resize buffer to zero.
+rab.resize(0);
+
+// TypedArray is now out-of-bounds.
+assert.sameValue(ta.length, 0);
+assert.sameValue(ta.byteOffset, 0);
+
+// Resize buffer to zero.
+rab.resize(0);
+
+// Attempt to fetch the next value. This exhausts the iterator.
+r = it.next();
+assert.sameValue(r.done, true);
+assert.sameValue(r.value, undefined);
+
+// Resize buffer so the typed array is again in-bounds.
+rab.resize(5);
+
+// TypedArray is now in-bounds.
+assert.sameValue(ta.length, 5);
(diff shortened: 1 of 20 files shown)


test/built-ins/TypedArray/prototype/values/make-in-bounds-after-exhausted.js
```diff
@@ -0,0 +1,55 @@
+// Copyright (C) 2024 André Bargull. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-%typedarray%.prototype.values
+description: >
+  Iterator is still exhausted when typedarray is changed to in-bounds.
+features: [TypedArray, resizable-arraybuffer]
```
ioannad, line 8: Just a suggestion, to indicate which step is being tested.

```suggestion
info: |
  %TypedArray%.prototype.values ( )

  ...
  3. Return CreateArrayIterator(O, VALUE).
  
features: [TypedArray, resizable-arraybuffer]
```

test/built-ins/TypedArray/prototype/values/make-out-of-bounds-after-exhausted.js
```diff
@@ -0,0 +1,49 @@
+// Copyright (C) 2024 André Bargull. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-%typedarray%.prototype.values
+description: >
+  Calling next on an out-of-bounds typedarray throws no error when iterator exhausted.
+features: [TypedArray, resizable-arraybuffer]
```
ioannad, line 8: Same as previous comment, just a suggestion.

```suggestion
info: |
  %TypedArray%.prototype.values ( )

  ...
  3. Return CreateArrayIterator(O, VALUE).
  
features: [TypedArray, resizable-arraybuffer]
```

test/built-ins/TypedArray/prototype/with/index-validated-against-current-length.js
```diff
@@ -0,0 +1,53 @@
...
+esid: sec-%typedarray%.prototype.with
+description: >
+  The index is validated against the current length.
+info: |
+  %TypedArray%.prototype.with ( index, value )
+
+  ...
+  8. Else, let numericValue be ? ToNumber(value).
+  9. If IsValidIntegerIndex(O, 𝔽(actualIndex)) is false, throw a RangeError exception.
+  ...
```
ioannad, line 14: Not adding more steps seems to suggest that a RangeError should be thrown, but the description says that the index should be validated against the current length, which at this step in this test should be the resized length, 5. Eventually the result should be created and set with the initial length, 2, as the tests confirms.

```suggestion
  1. Let O be the this value.
  2. Let taRecord be ? ValidateTypedArray(O, SEQ-CST).
  3. Let len be TypedArrayLength(taRecord).
  ...
  8. Else, let numericValue be ? ToNumber(value).
  ...
  10. Let A be ? TypedArrayCreateSameType(O, « 𝔽(len) »).
  ...
  13. Return A.
```

test/built-ins/TypedArray/prototype/values/make-in-bounds-after-exhausted.js
anba: I didn't add spec step references here, because the relevant steps are part of [CreateArrayIterator](https://tc39.es/ecma262/#sec-createarrayiterator), but it doesn't make sense to refer to `CreateArrayIterator`, because the current spec text is incorrect and has to be rewritten: https://github.com/tc39/ecma262/issues/3136

ptomato: OK, makes sense not to add the reference here, then.

test/built-ins/TypedArray/prototype/values/make-out-of-bounds-after-exhausted.js
ptomato: Likewise, assuming this should stay as is for the time being.

PR #4004, opened 2024-02-07
Temporal: Tests for date difference algorithm

ptomato:
This adds tests specifically for every kind of case that changes due to the tweak to the date difference algorithm: differences from a longer month to a shorter month, when the months are adjacent, in the same year but not adjacent, and in different years.

Also adds tests for a case that does *not* change, but would trip on an incorrectly implemented algorithm: when the intermediate months value falls at the end of February.

There was incidental coverage of the change to the date difference algorithm in other tests. Those are adjusted, as well.

Normative change: https://github.com/tc39/proposal-temporal/pull/2759
Consensus in February 2024


Ms2ger (approved):

ptomato:
Thanks for the review.

PR #4005, opened 2024-02-09
Add test with non-IANA time zone names

anba:
Time zone identifiers supported by ICU, but which aren't valid IANA identifiers.


anba:
Non-IANA time zone ids should be rejected to avoid bug reports like: <https://bugzilla.mozilla.org/show_bug.cgi?id=1879554>.

ptomato (approved):
Makes sense. Would you like to write the same test additionally for `new Temporal.TimeZone()` and `Temporal.TimeZone.from()`?

anba:
> Would you like to write the same test additionally for `new Temporal.TimeZone()` and `Temporal.TimeZone.from()`?

Sure. I've pushed 826de800a8f4e42401b36d12518afbf4213c9d45 to cover `Temporal.TimeZone`, too.

Ms2ger (approved):

PR #4009, opened 2024-02-15
Temporal: Tests for week-numbering algorithm

Aditi-1400:
Adds and updates tests for PR: https://github.com/tc39/proposal-temporal/pull/2756

### DIFF
--- a/test/built-ins/Temporal/Calendar/prototype/weekOfYear/custom-calendar-weekofyear.js
+++ b/test/built-ins/Temporal/Calendar/prototype/weekOfYear/custom-calendar-weekofyear.js
@@ -1,0 +1,23 @@
+// Copyright (C) 2024 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.calendar.prototype.weekofyear
+description: >
+  Temporal.Calendar.prototype.weekOfYear returns undefined for all 
+  custom calendars where weekOfYear() returns undefined.
+features: [Temporal]
+---*/
+
+class CustomCalendar extends Temporal.Calendar {
+  constructor() {
+    super("iso8601");
+  }
+  weekOfYear() {
+    return undefined;
+  }
+}
+
+const calendar = new CustomCalendar();
+const customCalendarDate = { month: 1, day: 1, year: 2024, calendar};
+assert.sameValue(calendar.weekOfYear({...customCalendarDate}), undefined);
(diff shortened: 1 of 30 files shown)


test/built-ins/Temporal/ZonedDateTime/prototype/yearOfWeek/non-iso-week-of-year.js
ptomato, line 1: It seems like this one should move to `test/intl402/Temporal/ZonedDateTime/prototype/yearOfWeek/non-iso-week-of-year.js` to match the other similar ones, and also because it will fail if an implementation doesn't include support for non-ISO calendars.

test/built-ins/Temporal/ZonedDateTime/prototype/weekOfYear/non-iso-week-of-year.js
ptomato, line 1: Same, `test/intl402/Temporal/ZonedDateTime/prototype/weekOfYear/non-iso-week-of-year.js`

ptomato (changes requested):
Thanks! All these tests look good, with one comment about needing to move a few of them.

However I think we need to cover a few more things in order to make sure that engines implement every aspect of this normative change:

- In `test/built-ins/`, validation that a custom calendar may return undefined from `yearOfWeek` and `weekOfYear` and that a PlainDate/PlainDateTime/ZonedDateTime with this custom calendar will have the `yearOfWeek`/`weekOfYear` property with the value of undefined
- In `test/intl402/`, a sample case where `iso8601` and `gregory` week numbers disagree according to CLDR

I'm happy to merge this after moving those two tests and add the additional coverage in a follow up, or also happy to re-review this with the additional tests added. Let me know what your preference is.

Aditi-1400:
> Thanks! All these tests look good, with one comment about needing to move a few of them.
> 
> However I think we need to cover a few more things in order to make sure that engines implement every aspect of this normative change:
> 
> * In `test/built-ins/`, validation that a custom calendar may return undefined from `yearOfWeek` and `weekOfYear` and that a PlainDate/PlainDateTime/ZonedDateTime with this custom calendar will have the `yearOfWeek`/`weekOfYear` property with the value of undefined
> * In `test/intl402/`, a sample case where `iso8601` and `gregory` week numbers disagree according to CLDR
> 
> I'm happy to merge this after moving those two tests and add the additional coverage in a follow up, or also happy to re-review this with the additional tests added. Let me know what your preference is.

I've updated the PR to add the tests you requested.

ptomato (approved):
Thanks for the update, all looks good!

PR #4012, opened 2024-02-20
Temporal: Tests for correct intermediate value in ZonedDateTime difference/rounding

ptomato:
These test cases ensure that DST disambiguation does not take place on intermediate values that are not the start or the end of the calculation.

Note that NormalizedTimeDurationToDays is no longer called inside Temporal.Duration.prototype.add/subtract, so a few tests can be deleted.

Other tests need to be adjusted because NormalizedTimeDurationToDays is no longer called inside Temporal.ZonedDateTime.prototype.since/until via DifferenceZonedDateTime, although it is still called as part of rounding.

In addition, new tests for the now-fixed edge case are added.

See https://github.com/tc39/proposal-temporal/pull/2760, a normative change approved in the TC39 plenary of 2024-02.


test/built-ins/Temporal/Duration/prototype/add/differencezoneddatetime-inconsistent-custom-calendar.js
```diff
@@ -0,0 +1,34 @@
...
+    }
+  })("iso8601");
+  
+  const duration1 = new Temporal.Duration(0, 0, /* weeks = */ 7);
+  const duration2 = new Temporal.Duration(0, 0, 0, /* days = */ 1);
+
+  const relativeTo = new Temporal.ZonedDateTime(0n, "UTC", cal);
+  
+  assert.throws(RangeError, () => duration1.add(duration2, { relativeTo }),
+    "Calendar calculation where more than one day correction is needed should cause RangeError");
```
Ms2ger, line 33: Weird indentation here and in several similar tests

Ms2ger (approved):

ptomato:
I'm expecting there may be more comments on the normative PR, so this may get updated as well.

ptomato:
I don't expect any more comments on the normative PR, so this is ready as far as I'm concerned. The two tests named `dst-month-day-boundary.js` are new ones that I've added in case anyone wants to take a look. They are straightforward regression tests for the bug that @arshaw found during review of that normative PR.

I'll plan to merge this Monday unless anyone hollers.

PR #4014, opened 2024-02-23
Add test for ECMA402 PR846

FrankYFTang:
the calenar name should be canonicalized
Spec change https://github.com/tc39/ecma402/pull/846


ptomato (approved):
Thanks, this seems like it tests exactly that normative change.

PR #3960, opened 2023-11-15
Add tests for the RegExp Modifiers proposal

rbuckton:
This adds tests for the Stage 3 [RegExp Modifiers](https://github.com/tc39/proposal-regexp-modifiers) proposal.

PR #4016, opened 2024-02-29
Regexp-modifiers syntax error tests

ioannad:
Adds syntax tests for the RegExp modifiers proposal.
 
Based on PR #3807 which had generated these tests from templates, but was stuck due to issue #3808.

Since these tests are a shipping blocker for v8, and #3808 seems not straight-forward to fix, I modified the generated tests and removed the templates.

Additional tests [suggested](https://github.com/tc39/test262/pull/3960#issuecomment-1966827213,) by @rbuckton are now added.

This should cover all the syntax error tests of the testing plan in #3756, and along with #3960 complete the tests for RegExp-modifiers.


ptomato (approved):
I only reviewed the new changes, but I previously reviewed #3758/#3807 and so have @gibson042 and @rbuckton, so I'm confident this can be merged.

What I will do before merging is squashing some of the commits.

PR #4017, opened 2024-03-12
add more Float16Array cases, and tests for Math.f16round

bakkot:
Followup to https://github.com/tc39/test262/pull/3849. Fixes https://github.com/tc39/test262/issues/3828. (~~Edit: actually I need the dataview methods also, which I'll try to get to soon. Doesn't invalidate these, it's just that more tests are needed.~~ Done.)

The structure of the TypedArray tests is such that you can't easily add tests for just one kind of typed array, so the values I'm adding need to be included in the "expected" lists for each of the typed array types. I automated that rather than doing all those by hand, but I did confirm the three major browser engines all agree (for the non-float16 cases; for the float16 ones I used conversion code I wrote and checked against other sources).

For the get/setFloat16 tests, I just copied the get/setFloat32 tests and modified as appropriate.


harness/byteConversionValues.js
```diff
@@ -139,7 +169,22 @@ var byteConversionValues = {
...
+      0,   // 2.9802322387695312e-8
+      0,   // 2.980232238769532e-8
+      0,   // 8.940696716308594e-8
+      0,   // 1.4901161193847656e-7
+      0,   // 1.490116119384766e-7
+      224, // 65504
+      240, // 65520
+      239, // 65519.99999999999
+      0,   // 0.000061005353927612305
+      0    // 0.0000610053539276123
```
ptomato, line 187: I didn't check how this is specified, but presumably these are equal to truncate(value) modulo 256, with the appropriate sign modifications for signed int8?

test/built-ins/DataView/prototype/getFloat16/length.js
```diff
@@ -0,0 +1,16 @@
...
+  DataView.prototype.getFloat16.length is 1.
+features: [Float16Array]
+includes: [propertyHelper.js]
+---*/
+
+assert.sameValue(DataView.prototype.getFloat16.length, 1);
+
+verifyNotEnumerable(DataView.prototype.getFloat16, "length");
+verifyNotWritable(DataView.prototype.getFloat16, "length");
+verifyConfigurable(DataView.prototype.getFloat16, "length");
```
ptomato, line 16: We're trying to prefer `verifyProperty` over the individual functions in new tests (command applies to other tests of property descriptors as well)
```suggestion
verifyProperty(DataView.prototype.getFloat16, "length", {
  value: 1,
  enumerable: false,
  writable: false,
  configurable: true,
});
```

test/built-ins/DataView/prototype/getFloat16/not-a-constructor.js
```diff
@@ -0,0 +1,21 @@
...
+
+assert.sameValue(
+  isConstructor(DataView.prototype.getFloat16),
+  false,
+  'isConstructor(DataView.prototype.getFloat16) must return false'
+);
+
+assert.throws(TypeError, () => {
+  let dv = new DataView(new ArrayBuffer(16)); new dv.getFloat16(0, 0);
+}, '`let dv = new DataView(new ArrayBuffer(16)); new dv.getFloat16(0, 0)` throws TypeError');
```
ptomato, line 20: Nitpick: I know the other not-a-constructor tests have these autogenerated assert messages, but let's not add more :stuck_out_tongue_closed_eyes:
Could I suggest "DataView.prototype.getFloat16 should not be a constructor" and "Constructing DataView.prototype.getFloat16 should throw TypeError"?

test/built-ins/DataView/prototype/getFloat16/resizable-buffer.js
```diff
@@ -0,0 +1,41 @@
...
+esid: sec-dataview.prototype.getfloat16
+description: Throws a TypeError if buffer is out-of-bounds
+features: [Float16Array, DataView, ArrayBuffer, resizable-arraybuffer]
+---*/
+
+assert.sameValue(
+  typeof ArrayBuffer.prototype.resize,
+  'function',
+  'implements ArrayBuffer.prototype.resize'
+);
```
ptomato, line 13: Is this necessary if we have the feature flag?

ptomato (commented):
I started reviewing this but didn't finish. Here's what I have so far.

test/built-ins/DataView/prototype/getFloat16/length.js
bakkot: I just copied these over from `getFloat32`. Would you prefer to use `verifyProperty` even if it means inconsistency between `getFloat16` and all the other dataview methods? Or I can update the other dataview methods as part of this PR and achieve consistency that way, alternatively.

test/built-ins/DataView/prototype/getFloat16/resizable-buffer.js
bakkot: I just copied this from the existing `getFloat32` test. I don't think it's necessary but I would prefer the new tests match existing ones.

test/built-ins/DataView/prototype/getFloat16/not-a-constructor.js
bakkot: Sure. Would you also like me to update the tests I copied this from?

syg (approved):
Correctness lgtm. I don't have much of an opinion on the style of these tests.

test/built-ins/DataView/prototype/getFloat16/length.js
ptomato: I mean, if you feel like doing the extra cleanup work to update the other dataview methods, great! That would be very kind of you and much appreciated. But it's not a requirement as far as I'm concerned.

I do prefer that we don't introduce new tests using APIs that we're trying to phase out, even if it means temporary inconsistency. If we preferred consistency, it'd mean more and more work for the person eventually doing the cleanup.

ptomato (approved):
I finished the rest of the review. All good with me, and many thanks for doing this. As for `verifyProperty`, I'd prefer if we could use `verifyProperty` in the new tests but I wouldn't block on it.

test/built-ins/DataView/prototype/getFloat16/length.js
bakkot: Opened https://github.com/tc39/test262/pull/4023. I'll update whichever PR lands second to match the other.

test/built-ins/DataView/prototype/getFloat16/not-a-constructor.js
bakkot: Opened https://github.com/tc39/test262/pull/4024.

ptomato:
@bakkot Many thanks for #4023 and #4024. I guess this is ready to land after rebasing, then?

bakkot:
@ptomato Should be. I rebased and re-ran my scripts from those PRs to make the same changes here.

ptomato (approved):
It's time then! Really appreciate the extra cleanup work as well.

harness/byteConversionValues.js
```diff
@@ -354,93 +459,138 @@ var byteConversionValues = {
...
+      -4294967296,             // -4294967296
+      Infinity,                // Infinity
+      -Infinity,               // -Infinity
+      0,                       // 0
+      2049,                    // 2049
+      2051,                    // 2051
+      0.00006103515625,        // 0.00006103515625
+      0.00006097555160522461,  // 0.00006097555160522461
+      5.960464477539063e-8,    // 5.960464477539063e-8
+      2.9802322387695312e-8,   // 2.9802322387695312e-8
```
ljharb, line 584: the eslint "no precision loss" rule complains about this value, is that a concern?

harness/byteConversionValues.js
```diff
@@ -354,93 +459,138 @@ var byteConversionValues = {
...
+      2.9802322387695312e-8,   // 2.980232238769532e-8
```
ljharb, line 585:
```suggestion
      2.980232238769532e-8,   // 2.980232238769532e-8
```

harness/byteConversionValues.js
```diff
@@ -51,7 +51,22 @@ var byteConversionValues = {
...
     Infinity,
     -Infinity,
-    0
+    0,
+    2049,                         // an integer which rounds down under ties-to-even when cast to float16
+    2051,                         // an integer which rounds up under ties-to-even when cast to float16
+    0.00006103515625,             // smallest normal float16
+    0.00006097555160522461,       // largest subnormal float16
+    5.960464477539063e-8,         // smallest float16
+    2.9802322387695312e-8,        // largest double which rounds to 0 when cast to float16
```
ljharb, line 60: the eslint "no precision loss" rule complains about this value, is that a concern?

harness/byteConversionValues.js
```diff
@@ -483,7 +633,22 @@ var byteConversionValues = {
...
       Infinity,    // Infinity
       -Infinity,   // -Infinity
-      0
+      0,           // 0
+      2049,                    // 2049
+      2051,                    // 2051
+      0.00006103515625,        // 0.00006103515625
+      0.00006097555160522461,  // 0.00006097555160522461
+      5.960464477539063e-8,    // 5.960464477539063e-8
+      2.9802322387695312e-8,   // 2.9802322387695312e-8
```
ljharb, line 642: the eslint "no precision loss" rule complains about this value, is that a concern?

harness/byteConversionValues.js
bakkot: No, `Float32Array.of(2.980232238769532e-8)[0] === 2.9802322387695312e-8`.

harness/byteConversionValues.js
bakkot: No.

harness/byteConversionValues.js
ljharb: ah ok, the comment confused me, thanks

test/built-ins/Math/f16round/not-a-constructor.js
```diff
@@ -0,0 +1,16 @@
...
+description: >
+  Math.f16round does not implement [[Construct]], is not new-able
+includes: [isConstructor.js]
+features: [Float16Array, Reflect.construct]
+---*/
+
+assert(!isConstructor(Math.f16round), "Math.f16round is not a constructor");
+
+assert.throws(TypeError, function () {
+  new Math.fround();
```
rmg-x, line 15: I just noticed this while working on the Ladybird browser project, isn't this supposed to be `new Math.f16round()` instead of `new Math.fround()`?

Ms2ger: Yes, good catch. Do you want to make a PR?

rmg-x: Sure! If someone else beats me to it, no issue either way.

PR #4019, opened 2024-03-13
Two tests previously assumed 'en' as locale for DurationFormat, assumption now made explicit

ben-allen:
Two tests compared output against hardcoded strings formatted as in 'en', despite the locale for the DurationFormat being set to `*undefined*`. updated to explicitly give 'en' as locale for DurationFormat


Ms2ger (changes requested):
I think you should also add the locale tag to the front matter

ben-allen:
> Also in the other file

whoops! fixed now

Ms2ger (approved):

PR #4022, opened 2024-03-22
add a test for decorator context names

mansona:
This was taken from the babel test suite https://github.com/babel/babel/blob/87ec4fc3d2cf44c30ef007f7da47d7177e07ffec/packages/babel-plugin-proposal-decorators/test/fixtures/2023-11-accessors/context-name/input.js#L1


test/language/statements/class/decorator/context-name.js
```diff
@@ -0,0 +1,31 @@
...
+
+  @dec static accessor "b"
+  @dec static accessor ["c"];
+
+  @dec static accessor 0;
+  @dec static accessor [1];
+
+  // TODO these two throw an error with SyntaxError: Unexpected token ILLEGAL but they pass on babel's test suite. should we remove them?
+  @dec static accessor 2n;
+  @dec static accessor [3n];
```
mansona, line 26: These are currently failing when I use the babel-test262-runner to test them but are working upstream in babel so I don't know if I should keep these two cases 🤔 thoughts?

test/language/statements/class/decorator/context-name.js
```diff
@@ -0,0 +1,31 @@
...
+
+  @dec static accessor [f()];
+}
+
+assert.compareArray(logs, ["computing f", "calling toPrimitive", "a", "#a", "b", "c", "0", "1", "2", "3", "f()"], 'did it work>');
```
ptomato, line 31: I'd suggest an assertion message that will give a hint about why the test is failing, instead of "did it work" :smile: 

How about something like "order of observable operations in evaluating class definition"?

ptomato (approved):
Thanks! I looked at these and they seem correct according to the current text of the decorators proposal. I don't really have a good sense of whether we already have coverage for this stuff so it'd be helpful if we could get a look from someone who's been involved with decorator tests before.

About the bigint property names, the question is, are they correct according to the spec. As far as I can tell, they are.

Here's the relevant grammar, with [Yield, Await] elided:
- _ClassElement_ : _DecoratorList_<sub>opt</sub> `static` _FieldDefinition_ `;`
- _FieldDefinition_ : `accessor` [no _LineTerminator_ here] _ClassElementName_ _Initializer_<sub>opt</sub>
- _ClassElementName_ : _PropertyName_
- _PropertyName_ : _LiteralPropertyName_
- _LiteralPropertyName_ : _NumericLiteral_
- _NumericLiteral_ : _DecimalBigIntegerLiteral_

(similar for _PropertyName_ : _ComputedPropertyName_ but with a longer path through all of the expression precedence elements)

So I don't think the SyntaxError is correct there.

test/language/statements/class/decorator/context-name.js
mansona: ha yes sorry that was a placeholder text to see if I could interact with the test harness properly 😂 I'll fix it now

mansona:
Thanks for the review @ptomato 🎉 

I've updated the message and removed my TODO item for removing the seemingly valid syntax 👍 is there something that I need to do to the frontmatter to mark it that it's currently expected to fail? 

> I don't really have a good sense of whether we already have coverage for this stuff

So I have 2 thoughts on this. If adding this test has identified a legitimate error that should have passed then this is clearly adding coverage for something that wasn't already covered 😂 But looking at the files that are already in the suite for decorators it seems that there are only `syntax` tests that have the comment `This file was procedurally generated from the following sources` so I'm assuming there are no hand-crafted tests that have any assertions, or at least that's how it seems from my investigations.

But anyway, I'm adding these tests with the help of @pzuraq so I'm sure she'll be the best person to comment on this 😂

nicolo-ribaudo:
Decorators coverage right now is something like 0.1%, so thank you very much!

If you plan to submit more tests, it would be incredibly useful if you could write a "testing plan" (like https://github.com/tc39/test262/issues/3725) so that it's easier to tell what's covered and what's not.

Ms2ger:
Please add a copyright header as well

ptomato:
> is there something that I need to do to the frontmatter to mark it that it's currently expected to fail?

@mansona There are no expected-to-fail tests, so nothing to do on your end. (The way it works is, a hypothetical JS implementation that correctly implements the spec plus all stage-3 proposals must pass all test262 tests. Implementations maintain their own expected-to-fail lists separately.)

mansona:
> Please add a copyright header as well

So I was wondering about that one 🤔 As I said in the description I copied the test from the babel test suite, so should I set the copyright header to the original author? Is there any guidance on that?

ptomato:
> As I said in the description I copied the test from the babel test suite, so should I set the copyright header to the original author? Is there any guidance on that?

Copyrights are annoying to deal with, sorry. Ideally the Babel test file would have its own copyright header, but it doesn't. You'd have to ask the Babel project if the copyright line in [babel/LICENSE](https://github.com/babel/babel/blob/87ec4fc3d2cf44c30ef007f7da47d7177e07ffec/LICENSE) ("Copyright (c) 2014-present Sebastian McKenzie and other contributors") applies to everything in the babel repo. If not, then you'd have to talk to the authors of that file figure out who owns the copyright.

(Public service announcement: Open source projects, please require license and copyright headers in every file in the repo! There are tools such as [REUSE](https://reuse.software/) that you can drop into your CI and make this easy, or you can roll your own like test262 does :smile:)

nicolo-ribaudo:
Hey, could you use "The Babel Team"? Unfortunately different packages in our repo have different author/copyright owner, so the best thing to do is to look for the author field in the relevant package.json (https://github.com/babel/babel/blob/dd699a8509f02efd7e5a45d49cf24b757a59e4a2/packages/babel-plugin-proposal-decorators/package.json#L4C14-L4C30)

pzuraq:
@nicolo-ribaudo I've made a test plan in this issue: https://github.com/tc39/test262/issues/4042

PR #4026, opened 2024-03-25
DurationFormat: changed metadata for 'en'/'en-US' tests to correctly indicate locale

ben-allen:
The metadata for several Intl.DurationFormat tests list give "en-US" as the locale when the actual locale used for the tests is just "en". This PR changes the metadata to match the tests.


ptomato (approved):
Thanks for catching this discrepancy.

PR #4027, opened 2024-03-25
DurationFormat: test formatting for units using mixed styles

ben-allen:
Tests formatting for units using a mixture of `"long"`, `"short"`, and `"narrow"` styles.  Compares output from format method to output produced through using `NumberFormat` and `ListFormat` as they are used in `DurationFormat` spec.

Considers both how `Intl.NumberFormat` formats units depending on unit styles and also how `Intl.ListFormat` formats lists depending on the base style.


ptomato (approved):
Thanks!

PR #4028, opened 2024-03-26
Correct Intl.DateTimeFormat and Intl.DisplayNames locale metadata

ben-allen:
Corrects metadata in `DateTimeFormat` and `DisplayNames` which inappropriately identified tests using 'en' locale as `locale: [en-US]`


ptomato (approved):
More of the same as yesterday, I believe?

PR #3888, opened 2023-07-28
Migrate resizable buffer tests out of staging

syg:
This PR migrates all resizable buffer tests out of staging into the main test suite. Tests that multiple similar methods at once are split up to to multiple files to be consistent with the existing directory structure.

No tests have been removed.

`test/built-ins/ArrayBuffer/prototype/resize/coerced-new-length-detach.js` is a new test that tests the normative detach timing change from the July 2023 TC39.

PR #4030, opened 2024-03-26
Collection of helper constants and functions for testing RABs

ioannad:
These are the parts of the code in resizable array buffer staging tests that are heavily repeated. In order to somewhat compact the migration of RAB staging tests (see PR #3888).

### DIFF
--- a/harness/resizableArrayBufferUtils.js
+++ b/harness/resizableArrayBufferUtils.js
@@ -139,9 +139,9 @@ function TestIterationAndResize(ta, expected, rab, resize_after, new_byte_length
     }
   }
   let len = expected.length;
-  assert(len == values.length);
+  assert.sameValue(len, values.length, "TestIterationAndResize: length of list of iterated values");
   for (let i = 0; i < len; i++) {
-    assert.compareArray(values[i], expected[i]);
+    assert.compareArray(values[i], expected[i], "TestIterationAndResize: list of iterated values");
   }
-  assert(resized);
+  assert(resized, "TestIterationAndResize: resize condition should have been hit");
 }
--- a/harness/resizableArrayBufferUtils.js
+++ b/harness/resizableArrayBufferUtils.js
@@ -138,6 +138,10 @@ function TestIterationAndResize(ta, expected, rab, resize_after, new_byte_length
       resized = true;
     }
   }
-  assert.compareArray(values.flat(), expected.flat());
+  let len = expected.length;
+  assert(len == values.length);
+  for (let i = 0; i < len; i++) {
+    assert.compareArray(values[i], expected[i]);
+  }
   assert(resized);
 }
--- a/harness/resizableArrayBufferUtils.js
+++ b/harness/resizableArrayBufferUtils.js
@@ -15,6 +15,7 @@ defines:
   - CreateRabForTest
   - CollectValuesAndResize
   - TestIterationAndResize
+features: [BigInt]
 ---*/
 
 class MyUint8Array extends Uint8Array {
@@ -40,18 +41,36 @@ const builtinCtors = [
   BigInt64Array
 ];
 
+// BigInt and Float16Array are newer features adding them above unconditionally
+// would cause implementations lacking it to fail every test which uses it.
+if (typeof Float16Array !== 'undefined') {
+  builtinCtors.push(Float16Array);
+}
+
+if (typeof BigInt !== 'undefined') {
+    builtinCtors.push(BigUint64Array);
+    builtinCtors.push(BigInt64Array);
+}
+
 const floatCtors = [
   Float32Array,
   Float64Array,
   MyFloat32Array
 ];
 
+if (typeof Float16Array !== 'undefined') {
+  floatCtors.push(Float16Array);
+}
+
 const ctors = [
   ...builtinCtors,
   MyUint8Array,
-  MyFloat32Array,
-  MyBigInt64Array
+  MyFloat32Array
 ];
+
+if (typeof BigInt !== 'undefined') {
+    ctors.push(MyBigInt64Array);
+}
 
 function CreateResizableArrayBuffer(byteLength, maxByteLength) {
   return new ArrayBuffer(byteLength, { maxByteLength: maxByteLength });
--- a/harness/resizableArrayBufferUtils.js
+++ b/harness/resizableArrayBufferUtils.js
@@ -106,7 +106,7 @@ function TestIterationAndResize(ta, expected, rab, resize_after, new_byte_length
   let values = [];
   let resized = false;
   for (const value of ta) {
-    if (value instanceof Array) {
+    if (Array.isArray(value)) {
       values.push([
         value[0],
         Number(value[1])
--- a/harness/resizableArrayBufferUtils.js
+++ b/harness/resizableArrayBufferUtils.js
@@ -141,7 +141,7 @@ function TestIterationAndResize(ta, expected, rab, resize_after, new_byte_length
   let len = expected.length;
   assert.sameValue(len, values.length, "TestIterationAndResize: length of list of iterated values");
   for (let i = 0; i < len; i++) {
-    assert.compareArray(values[i], expected[i], "TestIterationAndResize: list of iterated values");
+    assert.compareArray([].concat(values[i]), expected[i], "TestIterationAndResize: list of iterated values");
   }
   assert(resized, "TestIterationAndResize: resize condition should have been hit");
 }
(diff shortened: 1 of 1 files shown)


harness/resizableArrayBufferUtils.js
```diff
@@ -0,0 +1,125 @@
...
+  Uint16Array,
+  Int16Array,
+  Uint32Array,
+  Int32Array,
+  Float32Array,
+  Float64Array,
+  Uint8ClampedArray,
+  BigUint64Array,
+  BigInt64Array
+];
```
ljharb, line 42: what about in envs that don't support the bigint arrays? what about Float16Array? i think these need to be conditionally added

harness/resizableArrayBufferUtils.js
```diff
@@ -0,0 +1,125 @@
...
+function Convert(item) {
+  if (typeof item == 'bigint') {
+    return Number(item);
+  }
+  return item;
+}
+
+function ToNumbers(array) {
+  let result = [];
+  for (let item of array) {
```
ljharb, line 78: probably best not to use for..of here since that's not related to testing RABs?

harness/resizableArrayBufferUtils.js
```diff
@@ -0,0 +1,125 @@
...
+    rab.resize(resizeTo);
+  }
+  return true;
+}
+
+function TestIterationAndResize(ta, expected, rab, resize_after, new_byte_length) {
+  let values = [];
+  let resized = false;
+  for (const value of ta) {
+    if (value instanceof Array) {
```
ljharb, line 110:
```suggestion
    if (Array.isArray(value)) {
```
(instanceof Array should basically never be used)

harness/resizableArrayBufferUtils.js
```diff
@@ -0,0 +1,125 @@
...
+      ]);
+    } else {
+      values.push(Number(value));
+    }
+    if (!resized && values.length == resize_after) {
+      rab.resize(new_byte_length);
+      resized = true;
+    }
+  }
+  assert.compareArray(values.flat(), expected.flat());
```
ljharb, line 123: what about envs that don't support `.flat`?

ioannad: I addressed all the comments except this, as I am not sure what's the best way to proceed. One idea would be to define a new comparison function that doesn't need to use .flat here. Do you think there is a simpler way to take care of this?

ljharb: i guess that makes me wonder why they need to be flattened at all. what's the diff if you just compare them directly?

ptomato: compareArray doesn't do deep comparison, it does SameValue elementwise

ljharb: ahhh gotcha. so what we need here is `zip(values, expected).forEach((v, e) => assert.compareArray(v, e))` or something :-/

probably best to do just a normal for loop then, here?

harness/resizableArrayBufferUtils.js
```diff
@@ -0,0 +1,147 @@
...
+      rab.resize(new_byte_length);
+      resized = true;
+    }
+  }
+  let len = expected.length;
+  assert(len == values.length);
+  for (let i = 0; i < len; i++) {
+    assert.compareArray(values[i], expected[i]);
+  }
+  assert(resized);
```
ptomato, line 146: Assertions that aren't in the main test file can be difficult to track down when debugging, so we should give these some context:
```suggestion
  assert.sameValue(len, values.length, "TestIterationAndResize: length of list of iterated values");
  for (let i = 0; i < len; i++) {
    assert.compareArray(values[i], expected[i], "TestIterationAndResize: list of iterated values");
  }
  assert(resized, "TestIterationAndResize: resize condition should have been hit");
```

harness/resizableArrayBufferUtils.js
```diff
@@ -0,0 +1,147 @@
...
+  } else {
+    values.push(n);
+  }
+  if (values.length == resizeAfter) {
+    rab.resize(resizeTo);
+  }
+  return true;
+}
+
+function TestIterationAndResize(ta, expected, rab, resize_after, new_byte_length) {
```
ptomato, line 123: When I was trying to figure out when `Array.isArray(value)` could be true, I was confused because I thought `ta` stood for "typed array". But really, it's an array of keys or entries. Maybe it means `testArray`?

Small nitpick, consider changing `resize_after` and `new_byte_length` to `resizeAfter` and `newByteLength` to match the casing convention used elsewhere in the file.

harness/resizableArrayBufferUtils.js
```diff
@@ -0,0 +1,147 @@
...
+  if (typeof n == 'bigint') {
+    values.push(Number(n));
+  } else {
+    values.push(n);
+  }
+  if (values.length == resizeAfter) {
+    rab.resize(resizeTo);
+  }
+  return true;
+}
```
ptomato, line 121: This references variables that are not defined. It looks like in the original PR it was structured like this:
```js
let values;
let rab;
let resizeAfter;
let resizeTo;
function CollectValuesAndResize(n) {
  ...
}
...
assert(ArrayEveryHelper(arr, CollectValuesAndResize));
```
So I think we'll need to pass those as parameters here, and restructure the calling code to accommodate that.

For example, at first glance it doesn't seem necessary to call `ArrayEveryHelper` if `CollectValuesAndResize` can only return true or throw. So, we could change `CollectValuesAndResize` to create and return `values`, instead of capturing the local variable and pushing elements into it.

harness/resizableArrayBufferUtils.js
```diff
@@ -0,0 +1,147 @@
...
+    }
+    if (!resized && values.length == resize_after) {
+      rab.resize(new_byte_length);
+      resized = true;
+    }
+  }
+  let len = expected.length;
+  assert(len == values.length);
+  for (let i = 0; i < len; i++) {
+    assert.compareArray(values[i], expected[i]);
```
ptomato, line 144: I'm not sure `values[i]` always going to be an array, because of `values.push(Number(value))` on line 134.

It looks like `TestIterationAndResize` is called in the original PR sometimes with `ta` being an array of keys, and sometimes an array of entries. As far as I can tell, that's the reason for the `Array.isArray()` check above.

The previous `assert.compareArray(values.flat(), expected.flat())` nicely accommodated both cases. I'm not sure what's best to do here. Maybe we should consider splitting up this function, into `TestIterationKeysAndResize`/`TestIterationEntriesAndResize`?

ljharb:
```suggestion
    assert.compareArray([].concat(values[i]), expected[i]);
```

harness/resizableArrayBufferUtils.js
```diff
@@ -0,0 +1,151 @@
...
+  }
+  return true;
+}
+
+function TestIterationAndResize(ta, expected, rab, resizeAfter, newByteLength) {
+  let values = [];
+  let resized = false;
+  var arrayValues = false;
+
+  for (const value of ta) {
```
ioannad, line 128: @ljharb I switched this back to a for..of loop because some of the ta's (testArrays) are actually iterators. IIUC for..of was made exactly for this sort of usage, where it could be an array or an iterator being looped over. WDYT?

ioannad:
@ptomato @ljharb I think this is now ready for another round of reviews. I think everything is addressed except the name of the argument `ta` to `TestIterationAndResize`. Indeed "test array" seems to be the intended meaning as you said Philip, but it could also be a leftover from initial tests where the typed array - named also `ta` in the test files - was used instead of the derived arrays and iterators that are used there now.

ptomato (approved):
I think this is ready to move ahead.

The comment about for-of gave me a different idea to solve the `ta` naming; how about calling it `iterable`?

ioannad:
@ljharb just a friendly ping since my previous comment is now marked outdated:

WDYT about [leaving the for-of loop in `TestIterationAndResize`](https://github.com/tc39/test262/pull/4030/files/172b8c16f9a54d5b3cfbf8439993cb95d6b76815#r1571188646)? If you are ok with keeping the for-of loop after all, then I think all the other review comments have been addressed and this is ready to get merged.

ljharb:
I certainly won't block on it, it's just nice if tests don't rely on syntax that isn't necessary to run the test :-)

ljharb (approved):

ptomato:
OK, let's move forward with this. If there's an opportunity to get rid of the for-of by not using iterables elsewhere, I'd be happy to merge it in the future, but the RAB tests have been waiting for long enough :smile:

PR #4033, opened 2024-03-28
added test for formatting mixture of short and numeric styles

ben-allen:
Verifies that:

1. `numeric`-style minutes are not padded to 2 digits when the style for `hours` is neither `numeric` nor `2-digit`.
2. The correct `ListFormat` style (`"short"`) is used when no `DurationFormat` base style defined.

### DIFF
--- a/test/intl402/DurationFormat/prototype/format/mixed-short-and-numeric.js
+++ b/test/intl402/DurationFormat/prototype/format/mixed-short-and-numeric.js
@@ -1,0 +1,34 @@
+// Copyright (C) 2024 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-Intl.DurationFormat.prototype.format
+description: Checks that durations containing a mixture of numericlike and non-numericlike styles are formatted using the "short" style when DurationFormat base style is *undefined*.
+info: |
+    PartitionDurationFormatPattern ( durationFormat, duration )
+
+      12. Let listStyle be durationFormat.[[Style]].
+      (...)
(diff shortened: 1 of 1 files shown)


test/intl402/DurationFormat/prototype/format/mixed-short-and-numeric.js
```diff
@@ -0,0 +1,34 @@
...
+const timeSeparator = ":";
+
+
+let d = {days: 5, hours: 1, minutes: 2, seconds: 3};
+let dfOpts = {minutes: "numeric", seconds: "numeric"};
+
+let expectedList = [];
+expectedList.push(new Intl.NumberFormat(locale, {style: "unit", unit: "day", unitDisplay: "short"}).format(d.days));
+expectedList.push(new Intl.NumberFormat(locale, {style: "unit", unit: "hour", unitDisplay: "short"}).format(d.hours));
+expectedList.push(d.minutes + timeSeparator + new Intl.NumberFormat(locale, {minimumIntegerDigits: 2}).format(d.seconds));
```
ptomato, line 28: `d.minutes` should be formatted too?
```suggestion
expectedList.push(new Intl.NumberFormat(locale).format(d.minutes) + timeSeparator + new Intl.NumberFormat(locale, {minimumIntegerDigits: 2}).format(d.seconds));
```

Otherwise you're assuming that
```js
new Intl.NumberFormat(locale).format(num) === `${num}`
```
which is probably true in English.
Except for that, this test seems nicely generalizable to multiple locales.

ptomato (approved):
Intl.DurationFormat and PartitionDurationFormatPattern are surprisingly complicated to trace through, but I think I got it.

PR #4034, opened 2024-03-28
test formatting when microseconds or nanoseconds are the first numeric-styled unit

ben-allen:
If `microseconds` are the first `numeric`-styled unit, they should be displayed as a fractional component of milliseconds. Likewise, if `nanoseconds` is the first `numeric`-styled unit, they should be displayed as a fractional component of microseconds.

### DIFF
--- a/test/intl402/DurationFormat/prototype/format/fractions-of-subsecond-units-en.js
+++ b/test/intl402/DurationFormat/prototype/format/fractions-of-subsecond-units-en.js
@@ -1,0 +1,62 @@
+// Copyright (C) 2024 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-Intl.DurationFormat.prototype.format
+description: Checks that fractional milliseconds and microseconds are formatted correctly when microseconds or nanoseconds are the first units with "numeric" style.
+info: |
+    PartitionDurationFormatPattern ( durationFormat, duration )
+    (...)
+    9.
+      (...)
+      g. If unit is "seconds", "milliseconds", or "microseconds", then
+        i. If unit is "seconds", then
+            1. Let nextStyle be durationFormat.[[MillisecondsStyle]].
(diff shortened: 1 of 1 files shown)


ptomato (commented):
I didn't get all the way through this one, but I have a couple of questions:
1. If I'm understanding correctly, [GetDurationUnitOptions](https://tc39.es/proposal-intl-duration-format/#sec-getdurationunitoptions) step 4 basically forces `{microseconds: 'numeric'}` and `{nanoseconds: 'numeric'}` to `{microseconds: 'fractional'}` and `{nanoseconds: 'fractional'}` respectively. Do we need test coverage that these things produce the same output?
2. `.format(d.microseconds.toString() + decimalSeparator + d.nanoseconds.toString()))` seems OK for this particular input but it might be better to add `.padStart(3, '0')` to be robust against someone editing the values in `d`.

ptomato (changes requested):
Finished going through it, and I don't have any additional comments besides what I wrote the other day.

ben-allen:
Added the padding for the fractional components -- good catch!

With regard to testing for `"fractional"` style: The `"fractional"` style is never directly used by users, who only have `"numeric"` available, and so we don't need to have tests for it. The internal-use-only `"fractional"` style exists for a couple of different reasons:

1. The `"numeric"` style means two dramatically different things, depending on the unit to which it's applied. If the unit is hours, minutes, or seconds, it means something like "format this as on a digital clock", when it's a subsecond unit it means "make this a fractional component of the previously displayed unit". Using the name `"fractional"` simplifies the logic for formatting numeric units, particularly once https://github.com/tc39/proposal-intl-duration-format/pull/188 lands
2. A hypothetical DFv2 could allow users to style durations with fractional components of minutes or hours as well as seconds, milliseconds, and microseconds. In this case making `"fractional"` an actual user-specifiable style rather than just something for internal use would be necessary.

ptomato (approved):
Okay. `fractional` not being a user-facing option was what I was missing!

PR #4035, opened 2024-03-29
Relax whitespace requirements in Intl.DateTimeFormat.p.format timeStyle

ptomato:
This test uses formatToParts() to find out what kind of whitespace is used before the dayPeriod component, and expect that whitespace consistently.

I've observed an inconsistency between format() and formatToParts() on Node 20.7.0, so this method isn't reliable. An example:

    const dtf = new Intl.DateTimeFormat('en-US', {timeStyle: 'short'});
    dtf.formatToParts(0)[3].value.charCodeAt(0)  // 0x202F, narrow NBSP
    dtf.format(0)[4].charCodeAt(0)  // 0x0020, space

This is conformant with the spec.

So, change the test to just allow any kind of whitespace. It'd be better to rewrite the test to rely less on particular output, but I'm not prepared to do that right now.

### DIFF
--- a/test/intl402/DateTimeFormat/prototype/format/timedatestyle-en.js
+++ b/test/intl402/DateTimeFormat/prototype/format/timedatestyle-en.js
@@ -8,14 +8,6 @@ features: [Intl.DateTimeFormat-datetimestyle, Array.prototype.includes]
 locale: [en-US]
 ---*/
 
-// Tolerate implementation variance by expecting consistency without being prescriptive.
-// TODO: can we change tests to be less reliant on CLDR formats while still testing that
-// Temporal and Intl are behaving as expected?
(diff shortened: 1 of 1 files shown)


sffc:
This is a known issue. https://issues.chromium.org/issues/40893567

Ms2ger:
I'm not convinced this is conformant, given that both APIs call into PartitionDateTimePattern. At least, without evidence to the contrary, I would expect that PartitionDateTimePattern returns the same value for the same inputs

ptomato:
That's a very good point, I'm convinced. Closing.

sffc:
Yeah, the spec requires format and formatToParts to be the same. Browsers that don't currently do this are not conformant with that part of the spec.

PR #4046, opened 2024-04-08
Allow global var-via-eval be declared

syg:
See https://github.com/tc39/ecma262/pull/3226

cc @shvaikalesh who originally authored these tests.


ptomato (approved):
Looks like this accurately covers the proposal, both `var` and `function` declarations inside `eval`, and both `let` and `const` bindings shadowing. Thanks!

ptomato:
Proposal is stage 3 now, assuming no other reviews are forthcoming.

PR #4048, opened 2024-04-09
Reduce locale dependence in non-iso-calendars.js

ptomato:
This is a staging test that was tightly coupled to particular 'golden' output from toLocaleString. Rewrite it so that it uses formatToParts instead, and tests for the presence of particular parts, and their values.

Also removes some dead logging code from this test file.

### DIFF
--- a/test/staging/Intl402/Temporal/old/non-iso-calendars.js
+++ b/test/staging/Intl402/Temporal/old/non-iso-calendars.js
@@ -330,8 +330,6 @@ var fromWithCases = {
     year1: RangeError
   }
 };
-var logPerf = false;
-var totalNow = 0;
 for (var [id, tests] of Object.entries(fromWithCases)) {
   var dates = {
     year2000: Temporal.PlainDate.from("2000-01-01"),
@@ -341,7 +339,6 @@ for (var [id, tests] of Object.entries(fromWithCases)) {
     var values = tests[name];
     var errorExpected = values === RangeError;
     if ((id === "chinese" || id === "dangi") && hasOutdatedChineseIcuData ) {
-      var now = globalThis.performance ? globalThis.performance.now() : Date.now();
       if (errorExpected) {
         assert.throws(RangeError, () => {
           var inCal = date.withCalendar(id);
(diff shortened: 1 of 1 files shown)


test/staging/Intl402/Temporal/old/non-iso-calendars.js
```diff
@@ -16,70 +36,280 @@ var testChineseData = new Date("2001-02-01T00:00Z").toLocaleString("en-US-u-ca-c
 });
 var hasOutdatedChineseIcuData = !testChineseData.endsWith("2001");
 
+const calendars = [
```
Ms2ger, line 39: Can you use `Object.keys`  on `expected` instead? This runs the risk of people adding expected data but forgetting to update this list as well.

ptomato: Good point, I also added another commit to do this elsewhere in the file.

test/staging/Intl402/Temporal/old/non-iso-calendars.js
```diff
@@ -17,69 +37,257 @@ var testChineseData = new Date("2001-02-01T00:00Z").toLocaleString("en-US-u-ca-c
...
-var year2000Content = getLocalizedDates("2000-01-01T00:00Z");
-var year1Content = getLocalizedDates("0001-01-01T00:00Z");
-var year2000Snapshot = "iso8601: 1/1/2000\n" + "buddhist: 1/1/2543 BE\n" + "chinese: 11/25/1999\n" + "coptic: 4/22/1716 ERA1\n" + "dangi: 11/25/1999\n" + "ethioaa: 4/22/7492 ERA0\n" + "ethiopic: 4/22/1992 ERA1\n" + "gregory: 1/1/2000\n" + "hebrew: 23 Tevet 5760\n" + "indian: 10/11/1921 Saka\n" + "islamic: 9/25/1420 AH\n" + "islamic-umalqura: 9/24/1420 AH\n" + "islamic-tbla: 9/25/1420 AH\n" + "islamic-civil: 9/24/1420 AH\n" + "islamic-rgsa: 9/25/1420 AH\n" + "islamicc: 9/24/1420 AH\n" + "japanese: 1/1/12 H\n" + "persian: 10/11/1378 AP\n" + "roc: 1/1/89 Minguo";
-assert.sameValue(year2000Content, year2000Snapshot);
+function compareFormatToPartsSnapshot(isoString, expected) {
+  const date = new Date(isoString);
+  Object.keys(expected).forEach(calendar => {
+    const formatter = new Intl.DateTimeFormat(`en-US-u-ca-${calendar}`, { timeZone: "UTC" });
+    const actualComponents = formatter.formatToParts(date);
+    const expectedComponents = expected[calendar];
```
Ms2ger, line 45: Sorry for making you go around in circles, but it occurs to me that `Object.entries` would be even better

Ms2ger (approved):
Thanks!

test/staging/Intl402/Temporal/old/non-iso-calendars.js
ptomato: Oh yeah I didn't think about it too hard. `entries` would be better.

PR #4049, opened 2024-04-10
add tests for Math.sumPrecise

bakkot:
The [proposal for Math.sumPrecise](https://github.com/tc39/proposal-math-sum) is now stage 2.7. I've written tests which exercise
- the basic property descriptor
- the behavior for iterables, and non-iterable arguments
- rejection of non-Number values
- edge cases with `-0`, `NaN`, and ±`Infinity`
- the actual full-precision summation, using cases I wrote or ran into in my own implementation

I don't have a principled approach for what cases to include for the last bullet point. If implementations submit more, I'm inclined to include anything that actually tripped up any implementation in a way which was not covered by an existing test.


ljharb (approved):
Matches the tests in my polyfill.

ptomato:
I looked through the tests at a high level and they seem correct.

I'm working on documentation for how to write a testing plan and this proposal seemed the right size to try it out on, so I've written that up in https://github.com/tc39/test262/issues/4054. I'd be interested in your opinion; too detailed, not detailed enough?

test/built-ins/Math/sumPrecise/prop-desc.js
```diff
@@ -0,0 +1,14 @@
...
+description: checks the "sumPrecise" property of Math
+includes: [propertyHelper.js]
+features: [Math.sumPrecise]
+---*/
+
+verifyProperty(Math, "sumPrecise", {
+  writable: false,
+  enumerable: false,
+  configurable: true
+});
```
linusg, line 14: I don't think this is correct:

> Every other [data property](https://tc39.es/ecma262/#sec-object-type) described in clauses [19](https://tc39.es/ecma262/#sec-global-object) through [28](https://tc39.es/ecma262/#sec-reflection) and in Annex [B.2](https://tc39.es/ecma262/#sec-additional-built-in-properties) has the attributes { [[Writable]]: true, [[Enumerable]]: false, [[Configurable]]: true } unless otherwise specified.

linusg (changes requested):

linusg:
I'd also like to see a test case for whatever input values trigger this (non-standard?) `RangeError` in the polyfill: https://github.com/tc39/proposal-math-sum/blob/3c56f3df9f89d862515617c7e421f51b28bb71f8/polyfill/polyfill.mjs#L88

bakkot:
That `RangeError` is standard (step 7.b.ii.1), but is unreachable without providing 2**53 inputs (since each input can increment the overflow by at most one), which is impractical to test.

linusg:
I see, thanks!

test/built-ins/Math/sumPrecise/prop-desc.js
ljharb: good catch; my tests check enumerability and configurability but not writability.

bakkot:
I removed the "awaiting author" label because there's no outstanding feedback for me afaik.

ptomato:
Landing this is fine with me. NB, I did not confirm each numerical result in the numerical tests, I'm prepared to trust @bakkot on that.

I haven't looked at whether it covers every item in the testing plan #4054, but in general I don't think that should block PRs from landing.

bakkot:
> I haven't looked at whether it covers every item in the testing plan https://github.com/tc39/test262/issues/4054, but in general I don't think that should block PRs from landing.

It definitely doesn't, though I think it covers a majority.

---

You can check the numeric results against Python with the following script if you'd like:

```py
import sys
import re
import math

if len(sys.argv) != 2:
  print('Usage: python check-sum-tests.py /path/to/sumPrecise/sum.js')
  sys.exit(1)

with open(sys.argv[1]) as f:
  file = f.read()

for line in file.splitlines():
  if 'Math.sumPrecise(' not in line:
    continue
  match = re.search("Math\.sumPrecise\(\[([^]]*)\]\), ([^)]*)\)", line)

  vals = [float(x.strip()) for x in match.group(1).split(',')]
  assertedResult = float(match.group(2))
  try:
    actualResult = math.fsum(vals)
  except OverflowError:
    # python's fsum can't handle intermediate overflow
    # we can work around it by summing the large parts seperately,
    # with those parts shrunk by a factor of 64 so the result doesn't overflow
    # as long as there is a reasonable number of items this works fine
    # order doesn't matter because this is fsum
    # and shrinking then growing by a factor of 64 doesn't lose precision because
    # the values are large
    large = [x / 64 for x in vals if abs(x) > 1]
    small = [x for x in vals if abs(x) <= 1]
    actualResult = 64 * math.fsum(large) + math.fsum(small)
  print(actualResult == assertedResult)
```

Of course, my tests were against an implementation derived from the same logic as `fsum`, so that won't catch errors between `fsum` and the spec. But it's at least enough to confirm I didn't make any copy/paste errors. (Also I checked the results against [an entirely different implementation](https://github.com/tc39/proposal-math-sum/blob/3c56f3df9f89d862515617c7e421f51b28bb71f8/polyfill/full-precision-with-bigints.mjs) and they matched.)

test/built-ins/Math/sumPrecise/sum.js
```diff
@@ -0,0 +1,37 @@
...
+features: [Math.sumPrecise]
+---*/
+
+assert.sameValue(Math.sumPrecise([1, 2, 3]), 6);
+assert.sameValue(Math.sumPrecise([1e308]), 1e308);
+assert.sameValue(Math.sumPrecise([1e308, -1e308]), 0);
+assert.sameValue(Math.sumPrecise([0.1]), 0.1);
+assert.sameValue(Math.sumPrecise([0.1, 0.1]), 0.2);
+assert.sameValue(Math.sumPrecise([0.1, -0.1]), 0);
+assert.sameValue(Math.sumPrecise([1e308, 1e308, 0.1, 0.1, 1e30, 0.1, -1e30, -1e308, -1e308]), 0.30000000000000004);
```
Jack-Works, line 16: Hi! I wonder why this is `0.30000000000000004`, not `0.3`.

I found 1e308 and 1e30 cancel out, and it is 0.1+0.1+0.1

I hit this when I try to replace algorithm with a big decimal library which provides lossless computation.

sno2: I would think both are valid results given the underlying addition algorithm is unspecified. There's similar discussion here: https://github.com/tc39/proposal-math-sum/issues/17

bakkot: > Hi! I wonder why this is 0.30000000000000004, not 0.3.
> I found 1e308 and 1e30 cancel out, and it is 0.1+0.1+0.1

`0.1` in JS represents the real number 0.1000000000000000055511151231257827021181583404541015625. Adding that number to itself three times gives 0.3000000000000000166533453693773481063544750213623046875. That number is exactly equidistant between the real number represented by `0.3` and the real number represented by`0.30000000000000004`:

```js
let abs = b => b >= 0n ? b : -b;
let addend = BigInt(0.1.toFixed(100).slice(2));
let sum = addend + addend + addend;
abs(sum - BigInt(0.3.toFixed(100).slice(2))) === abs(sum - BigInt(0.30000000000000004.toFixed(100).slice(2))) // true
```
and in JS when converting a real number to a Number value, [in the case of ties we choose the one with the even significand](https://tc39.es/ecma262/multipage/ecmascript-data-types-and-values.html#sec-ecmascript-language-types-number-type):

> If two values of the set are equally close, then the one with an even significand is chosen

which in this case means picking `0.30000000000000004`:
- `(new DataView(Float64Array.of(0.30000000000000004).buffer)).getUint8(0) & 1` // 0
- `(new DataView(Float64Array.of(0.3).buffer)).getUint8(0) & 1` // 1

(above code assumes you're running on little-endian hardware, which you almost certainly are.)
 
> I would think both are valid results given the underlying addition algorithm is unspecified.

No, the specification is unambiguous.

Jack-Works: Thank you, I get it. If I run `.toPrecision(100)` on the input numbers, I can now use the big decimal library to get the correct result.

PR #4052, opened 2024-04-11
Update test for o[p] = f()

rkirsling:
Update S11.13.1_A7_T3.js now that consensus has been reached on https://github.com/tc39/ecma262/pull/3307.

This ends up basically the same as [target-member-computed-reference-null.js](https://github.com/tc39/test262/blob/main/test/language/expressions/assignment/target-member-computed-reference-null.js), just with a valid base object.


ptomato (approved):
LGTM.

For full coverage of the normative PR, do we also need a similar test for `super[prop()]`?

rkirsling:
Good point! Updated.

ptomato (approved):
:+1:

Nice work picking a different name for `S11.13.1_A7_T3.js`, as well.

ptomato:
The ESMeta test is failing because of a chicken and egg problem: tc39/ecma262#3307 isn't merged into the spec yet. In other such cases recently, I've always merged the test262 PR first.

PR #4055, opened 2024-04-13
Add missing DummyError function to target-super-computed-reference.js

lahma:
https://github.com/tc39/test262/pull/4052 introduced a new test file throwing `DummyError`, but the test file is missing its defintion.


rkirsling:
Oh geez, thanks for that!

ptomato (approved):
:+1: can't believe I missed that.

Issue #3945, opened 2023-10-13
built-ins/RegExp/property-escapes/generated need to be updated for Unicode 15.1


anba:
Similar to previous issues like #3663, the files in <https://github.com/tc39/test262/tree/main/test/built-ins/RegExp/property-escapes/generated> have to be regenerated for Unicode 15.1.

mathiasbynens:
Ref. https://github.com/tc39/ecma262/issues/3073.

PR #4058, opened 2024-04-15
Update RegExp `\p{…}` tests per Unicode 15.1.0

mathiasbynens:
Issue: #3945


ptomato (approved):
Thanks!
I propose rubber-stamping this. Unless I hear any objections I'll merge it soon.

PS. Have you considered moving those scripts into the test262 repo? Or are they used for other purposes as well?

ptomato:
PS. I'm not sure whether the existing tests already timed out on engine262 or what... maybe worth checking.

PR #4060, opened 2024-04-18
Add additional tests for tc39/ecma262#3307.

rkirsling:
Ensure that the following cases are covered:
- https://github.com/tc39/ecma262/issues/3295#issuecomment-2058110314
- https://github.com/tc39/ecma262/issues/2659#issue-1131245695


bakkot (approved):
Thanks!

ptomato (approved):
:+1:

ptomato:
Like the previous PR, I'm assuming that the ESMeta job fails because it's working from a version of ECMA-262 without the normative change.

PR #4061, opened 2024-04-18
Update NumberFormat.prototype.useGrouping test for CLDR 45

trflynn89:
In CLDR 45, in the en-IN locale, the compact thousands symbol changed from 'T' to 'K' after a survey was conducted in India:

https://github.com/unicode-org/cldr/commit/b8d447297556ee01575da2b5b710554af34b3482


Ms2ger:
I wonder if we should allow both

ptomato (approved):
This has been open for quite a long time, so implementations have had time to adapt to CLDR 45. In general I prefer not to test against hardcoded internationalization output, but given the age of this PR I'll suggest that we just land it. If that causes problems for implementations that haven't adapted to the CLDR change, we can then do something more complicated to allow both variations.

PR #4062, opened 2024-04-22
Bug fix for DurationFormat prototype/format/fractions-of-subsecond-units-en.js 

ben-allen:
Uncommented erroneously commented line, added roundingMode: 'trunc' to NumberFormat options when fractional digits are formatted to match behaviour of DurationFormat -- doesn't make a difference with these test values, but could with others

### DIFF
--- a/test/intl402/DurationFormat/prototype/format/fractions-of-subsecond-units-en.js
+++ b/test/intl402/DurationFormat/prototype/format/fractions-of-subsecond-units-en.js
@@ -42,19 +42,19 @@ let dfOpts = {microseconds: "numeric"};
 
 let expectedList = [];
 expectedList.push(new Intl.NumberFormat(locale, {style: "unit", unit: "second", unitDisplay: "short"}).format(d.seconds));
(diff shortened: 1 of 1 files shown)


Ms2ger (approved):

FrankYFTang:
@ben-allen this particular test is still not according to the spec for lacking other necessary fix

"numeric" is not a valid option value for "milliseconds", "microseconds" and "nanoseconds" based on the listed values in 4th "Values" column in 
https://tc39.es/proposal-intl-duration-format/#table-durationformat

My implementation will throw RangeError for this test because in step 17-f  of 1.2.1 Intl.DurationFormat ( [ locales [ , options ] ] )


f. Let unitOptions be ? [GetDurationUnitOptions](https://tc39.es/proposal-intl-duration-format/#sec-getdurationunitoptions)(unit, options, style, valueList, digitalBase, prevStyle).

when unit is dfOpts = {microseconds: "numeric"};
or 
dfOpts = {nanoseconds: "numeric"};

PR #4064, opened 2024-04-24
Fix mixed-short-and-numeric.js

FrankYFTang:
The option for the listFormat need to include {type: "unit"}

see https://tc39.es/proposal-intl-duration-format/#sec-partitiondurationformatpattern 

@ben-allen @anba @ryzokuken @Ms2ger


ben-allen:
Ah, thanks, this looks fixed now. I've put up https://github.com/tc39/test262/pull/4065, which fixes another test with this problem.

ben-allen (approved):
LGTM

Ms2ger (approved):

PR #4069, opened 2024-05-01
add Promise.try tests

ljharb:
https://github.com/tc39/proposal-promise-try/issues/15

### DIFF
--- a/test/built-ins/Promise/try/args.js
+++ b/test/built-ins/Promise/try/args.js
@@ -1,0 +1,22 @@
+// Copyright (C) 2024 Jordan Harband. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Promise.try forwards arguments
+esid: sec-promise.try
+features: [promise-try]
+flags: [async]
+includes: [asyncHelpers.js, compareArray.js]
+---*/
+
+var sentinel = { sentinel: true };
+
+asyncTest(
+  Promise.try(function () {
+    assert.compareArray(
+      Array.prototype.slice.call(arguments),
+      [1, 2, Test262Error, sentinel]
+    );
+  }, 1, 2, Test262Error, sentinel)
+);
+
--- a/test/built-ins/Promise/try/promise.js
+++ b/test/built-ins/Promise/try/promise.js
@@ -1,0 +1,12 @@
+// Copyright (C) 2024 Jordan Harband. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+description: Promise.try return value is a Promise
+features: [promise-try]
+---*/
+
+var instance = Promise.try(function () {});
+
+assert.sameValue(instance.constructor, Promise);
+assert.sameValue(instance instanceof Promise, true);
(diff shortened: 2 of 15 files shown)


test/built-ins/Promise/try/promise.js
```diff
@@ -0,0 +1,12 @@
...
+
+/*---
+description: Promise.try return value is a Promise
+features: [promise-try]
+---*/
+
+var instance = Promise.try(function () {});
+
+assert.sameValue(instance.promise.constructor, Promise);
+assert.sameValue(instance.promise instanceof Promise, true);
```
ckknight, line 12: Because `Promise.try`'s return value is a `Promise`, `instance` should already be a `Promise`, therefore the `.promise` property access seems like a mistake
```suggestion
assert.sameValue(instance.constructor, Promise);
assert.sameValue(instance instanceof Promise, true);
```

ljharb: oh whoops, you're right, thanks

test/built-ins/Promise/try/throws.js
```diff
@@ -0,0 +1,27 @@
...
+    }
+  )
+  .then(
+    function() {
+      $DONE();
+    },
+    function() {
+      $DONE('The promise should not be rejected.');
+    }
+  )
```
gibson042, line 27:
```suggestion
features: [promise-try, async]
includes: [asyncHelpers.js]
---*/


assert.throwsAsync(
  Test262Error,
  function () {
    return Promise.try(function () { throw new Test262Error(); });
  },
  "error thrown from callback must become a rejection"
);
```

gibson042 (approved):
LGTM modulo better use of helpers.

test/built-ins/Promise/try/args.js
```diff
@@ -0,0 +1,32 @@
...
+  assert.sameValue(v, sentinel);
+})
+.then(
+  function(v) {
+    $DONE();
+  },
+  function() {
+    $DONE('The promise should not be rejected.');
+  }
+)
```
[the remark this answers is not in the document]
ljharb: ha, thanks :-) i just looked at existing Promise helper tests, I didn't read our own docs.

test/built-ins/Promise/try/args.js
jridgewell, line 1: It looks like this case is the only path that hits the resolve path's return value, but this test case is specific to args. Can we add one specific to the resolve value?

ljharb: @jridgewell https://github.com/tc39/test262/pull/4069/files#diff-35a7b2f0585b3e54bcebe06320ebb3945195ae65444ee86b2bb99833e245df5cR23 should already cover that, unless you want it as a separate test?

jridgewell: Yah, the return value should be separated into its own test, it's not related to args passing.

ptomato:
Is this waiting for another review? Happy to oblige but it looks like it's already got plenty...

OK with me to just fix the [linter errors](https://app.circleci.com/pipelines/github/tc39/test262/5875/workflows/09504474-5604-43d5-9659-36e78abf173c/jobs/53353) and merge it. (looks like it needs 3x `flags: [async]` instead of `features: [async]`)

ljharb:
It should have both i think; i'll fix that and then land it :-)

PR #4080, opened 2024-05-14
Add test for skipped initial days in Temporal.Calendar.prototype.yearMonthFromFields

anba:
The polyfill doesn't pass this test:
```js
let pym = Temporal.PlainYearMonth.from({calendar: "japanese", era: "heisei", eraYear: 1, month: 1});
console.log(pym.getISOFields().isoDay);  // Prints 1
console.log(pym.era);  // Prints "showa"
```

Is this a bug in the polyfill or did I misinterpret this step from [Temporal.Calendar.prototype.yearMonthFromFields](https://tc39.es/proposal-temporal/#sec-temporal.calendar.prototype.yearmonthfromfields)?
> 7.c Let firstDayIndex be the 1-based index of the first day of the month described by fields (i.e., 1 unless the month's first day is skipped by this calendar.)

### DIFF
--- a/test/intl402/Temporal/Calendar/prototype/yearMonthFromFields/reference-day-japanese.js
+++ b/test/intl402/Temporal/Calendar/prototype/yearMonthFromFields/reference-day-japanese.js
@@ -1,0 +1,47 @@
+// Copyright (C) 2024 André Bargull. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.calendar.prototype.yearmonthfromfields
+description: >
+  Reference ISO day is chosen to be the first day of the calendar month.
+info: |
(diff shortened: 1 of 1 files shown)


ptomato:
I think that might indeed be a misinterpretation of that step. That January spans two era/eraYear pairs, Showa 64 and Heisei 1. It's not two separate Januaries, so the PlainYearMonth correctly resolves to the era/eraYear pair in effect at the beginning of the month.

_firstDayIndex_ is only intended to not be 1 if the calendar skips days at the beginning of the month outright, which none of the current ICU calendars do. (The only occurrence I'm aware of is in the switch from Julian to Gregorian calendar in some European regions, where they skipped days at the beginning of the month rather than in the middle.)

See also #1315.

ptomato:
Superseded by https://github.com/tc39/test262/pull/4461.

PR #4081, opened 2024-05-15
Temporal: Adjust tests for removal of options parameter from Temporal.Duration.p.add/subtract

ptomato:
This corresponds to a yet-to-be-presented potential normative change, to simplify the Temporal proposal.

This PR is split into two commits, for ease of review. The first one is just a blanket removal of everything that tested the options parameter, and doesn't need to be reviewed very attentively. The second commit contains the actual tests.

See tc39/proposal-temporal#2825


gibson042 (approved):
I'm not _certain_ that we're not losing coverage with the removed files, but I did spot check a handful and it looks good.

ptomato:
This normative change achieved consensus at the TC39 plenary of 2024-06-12.

PR #4087, opened 2024-05-17
`Promise.try`: avoid unneeded syntax

ljharb:
Followup to #4086

### DIFF
--- a/harness/asyncHelpers.js
+++ b/harness/asyncHelpers.js
@@ -28,84 +28,86 @@ function asyncTest(testFunc) {
   }
 }
 
-assert.throwsAsync = async function (expectedErrorConstructor, func, message) {
-  var innerThenable;
-  if (message === undefined) {
-    message = "";
-  } else {
-    message += " ";
-  }
-  if (typeof func === "function") {
-    try {
-      innerThenable = func();
-      if (
-        innerThenable === null ||
-        typeof innerThenable !== "object" ||
-        typeof innerThenable.then !== "function"
-      ) {
-        message +=
-          "Expected to obtain an inner promise that would reject with a" +
-          expectedErrorConstructor.name +
-          " but result was not a thenable";
-        throw new Test262Error(message);
-      }
-    } catch (thrown) {
-      message +=
-        "Expected a " +
-        expectedErrorConstructor.name +
-        " to be thrown asynchronously but an exception was thrown synchronously while obtaining the inner promise";
-      throw new Test262Error(message);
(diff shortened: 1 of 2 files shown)


gibson042 (approved):
[`throwsAsync`](https://github.com/tc39/test262/blob/57c7d06bf1829f0deec2a2b92b8f5b4a3cdd76f0/harness/asyncHelpers.js#L31) itself is defined as an `async` function, so this is probably moot unless you also update that to be non-async and replace each of its `throw`s to instead return a rejected promise.

ljharb:
oof, ok fair enough. i'll do that also.

harness/asyncHelpers.js
```diff
@@ -28,84 +28,88 @@ function asyncTest(testFunc) {
...
-        expectedErrorConstructor.name +
-        " to be thrown asynchronously but a " +
-        thrown.constructor.name +
-        " was thrown synchronously";
+      throw new Test262Error(message);
     }
-    throw new Test262Error(message);
-  }
+
+    resolve();
```
gibson042, line 113: This line will never be reached; it is preceded by `try { return …; } catch (thrown) { …; throw new Test262Error(…); }`. I think you actually want to capture both `resolve` and `reject` in the `new Promise` callback, and replace `return innerThenable.then(…)` with `innerThenable.then(…).then(resolve, reject)`.
```patch
@@ -29,7 +29,7 @@ function asyncTest(testFunc) {
 }
 
 assert.throwsAsync = function (expectedErrorConstructor, func, message) {
-  return new Promise(function (resolve) {
+  return new Promise(function (resolve, reject) {
     var innerThenable;
     if (message === undefined) {
       message = "";
@@ -64,7 +64,7 @@ assert.throwsAsync = function (expectedErrorConstructor, func, message) {
     }
 
     try {
-      return innerThenable.then(
+      innerThenable.then(
         function () {
           message +=
             "Expected a " +
@@ -92,7 +92,7 @@ assert.throwsAsync = function (expectedErrorConstructor, func, message) {
           }
         }
-      );
+      ).then(resolve, reject);
     } catch (thrown) {
       if (typeof thrown !== "object" || thrown === null) {
         message +=
@@ -109,7 +109,5 @@ assert.throwsAsync = function (expectedErrorConstructor, func, message) {
       }
     }
-
-    resolve();
   });
 };

```

gibson042 (changes requested):

gibson042 (approved):

gibson042:
LGTM!

PR #4090, opened 2024-05-18
Add missing `return` statement to `throwsAsync` callback

trflynn89:
This callback is expected to return a thenable.


ljharb (approved):
whoops, thanks

PR #4091, opened 2024-05-19
Cleanup asyncHelpers

gibson042:
**CONTRIBUTING.md**
* Only mention `$DONE` after `asyncTest`.
* Document that `asyncTest` is not intended to be called multiple times per file.

**harness/asyncHelpers.js**
* Add doc comments.
* Refactor `assert.throwsAsync` to fail fast and isolate error conditions.
* Clean up `assert.throwsAsync` failure messages (remove mention of an "inner" promise/thenable).
* Reduce the size of `assert.throwsAsync` for better comprehensibility in one screenful.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -209,8 +209,9 @@ This key is for boolean properties associated with the test.
 - **raw** - execute the test without any modification (no harness files will be
   included); necessary to test the behavior of directive prologue; implies
   `noStrict`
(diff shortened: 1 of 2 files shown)


ptomato (approved):
Thanks for the clear split into commits, that was very helpful when reviewing.

The only thing I have reservations about is renaming `innerThenable` to `p` in the final commit — I don't think that improves the readability. Everything else looks good.

PR #4093, opened 2024-05-24
Temporal: fix messages on `Duration/out-of-range` tests

zzzzBov:
Fixed test messages that incorrectly referenced the `days` parameter to say `weeks` instead.


ptomato (approved):
Thanks!

PR #4098, opened 2024-05-29
Allow alternative era names

anba:
Temporal doesn't specify concrete era names, so tests shouldn't assert for example that the era code of the current Gregorian era is `"ce"`. We still want to validate the era names somehow however, so allow alternative era names using the era codes from the "Intl era and monthCode" proposal.

### DIFF
--- a/harness/temporalHelpers.js
+++ b/harness/temporalHelpers.js
@@ -55,6 +55,100 @@ var TemporalHelpers = {
   ],
 
   /*
+   * List of known calendar eras and their possible aliases.
+   *
+   * https://tc39.es/proposal-intl-era-monthcode/#table-eras
+   */
+  CalendarEras: {
+    buddhist: [
+      { era: "buddhist", aliases: ["be"] },
+    ],
+    chinese: [
+      { era: "chinese" },
+    ],
+    coptic: [
+      { era: "coptic" },
+      { era: "coptic-inverse" },
+    ],
+    dangi: [
+      { era: "dangi" },
+    ],
+    ethiopic: [
+      { era: "ethiopic", aliases: ["incar"] },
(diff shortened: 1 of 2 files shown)


harness/temporalHelpers.js
```diff
@@ -54,6 +54,101 @@ var TemporalHelpers = {
...
+    }
+    assert.sameValue(typeof eraName, "string");
+
+    // Don't attempt to canonicalize era names of user-defined calendars.
+    let calendar = temporalObject.getCalendar();
+    if (!(calendar instanceof Temporal.Calendar)) {
+      return eraName;
+    }
+
+    let calendarId = calendar.id;
```
anba, line 141: Some tests assert that the `id` property of user-defined calendars is only accessed a specific number of times, so we can't directly use `temporalObject.calendarId`, but instead first have to ensure that `calendar` is indeed a `Temporal.Calendar` instance.

harness/temporalHelpers.js
```diff
@@ -54,6 +54,101 @@ var TemporalHelpers = {
...
+   */
+  canonicalizeEraInCalendar(temporalObject, eraName) {
+    if (eraName === undefined) {
+      return undefined;
+    }
+    assert.sameValue(typeof eraName, "string");
+
+    // Don't attempt to canonicalize era names of user-defined calendars.
+    let calendar = temporalObject.getCalendar();
+    if (!(calendar instanceof Temporal.Calendar)) {
```
Ms2ger, line 137: These no longer exist

anba:
Updated the helper function to work with string-only calendar identifiers and also updated the new test "intl402/Temporal/ZonedDateTime/construct-non-utc-non-iso.js" to call the helper function.

ptomato (approved):
Thanks!

PR #4101, opened 2024-05-30
Allow different approximations in the observational Islamic calendar

anba:
The "islamic" calendar id refers to the observational Islamic calendar and different implementation may use different algorithms to approximate when the moon's crescent can be observed. Change the test data to allow the approximations used in ICU4X in addition to the ones used in ICU4C.

Additionally allow implementations to canonicalise the calendar id.

### DIFF
--- a/test/staging/Intl402/Temporal/old/islamic-calendars.js
+++ b/test/staging/Intl402/Temporal/old/islamic-calendars.js
@@ -12,10 +12,23 @@ features: [Temporal]
 const tests = [
   {
     calendar: "islamic",
-    inLeapYear: false,
-    daysInYear: 354,
-    daysInMonth12: 29,
-    isoDate: "2023-07-18",
+    choices: [
+      // Approximations of the observational Islamic calendar as computed by ICU4C.
+      {
+        inLeapYear: false,
+        daysInYear: 354,
+        daysInMonth12: 29,
+        isoDate: "2023-07-18",
+      },
+
+      // Approximations of the observational Islamic calendar as computed by ICU4X.
+      {
+        inLeapYear: true,
+        daysInYear: 355,
+        daysInMonth12: 30,
+        isoDate: "2023-07-19",
+      }
+    ],
   },
   {
     calendar: "islamic-umalqura",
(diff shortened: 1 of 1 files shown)


test/staging/Intl402/Temporal/old/islamic-calendars.js
```diff
@@ -67,15 +97,29 @@ const tests = [
...
+  const { calendar, choices = [test] } = test;
   const year = 1445;
   const date = Temporal.PlainDate.from({ year, month: 1, day: 1, calendar });
   const isoFields = date.getISOFields();
-  assert.sameValue(date.calendarId, calendar);
+  if (calendar !== "islamicc") {
+    assert.sameValue(date.calendarId, calendar);
+  } else {
+    // TODO: Steps to canonicalize the calendar identifier are still missing.
+    // https://github.com/tc39/ecma402/issues/828
```
Ms2ger, line 108: I think this is done now

anba: Rebased and updated the test.

Ms2ger (approved):
Seems reasonable

ptomato:
How soon would it be possible to determine whether ICU4C or ICU4X's calculation is correct? I mean, this is a date in the past, so there should be one correct answer to the question "what was the Gregorian date on 1 Muharram 1445?" for each one of the supported astronomical Islamic calendars. In other words, when results from ICU4C and ICU4X differ for dates that have already happened, at least one of them is objectively wrong.

If it's not feasible in the short term to determine which is correct, then sure, we could adopt this as a temporary measure to avoid cluttering test failure results with something that is not directly relevant to whether ECMA-402 is implemented correctly. I wouldn't want to keep it permanently though.

anba:
This is possibly caused by <https://github.com/unicode-org/icu4x/issues/4982>. 

From <https://en.wikipedia.org/wiki/Hijrah#Aftermath_and_legacy>:
> Burnaby states that: "Historians in general assert that Muhammad fled from Mecca at the commencement of the third month of the Arabian year, Rabi 'u-l-awwal. They do not agree as to the precise day. According to Ibn-Ishak, it was on the first or second day of the month;"

Which is an excerpt from <https://archive.org/details/elementsofjewish00burnuoft>. Part II, Chapter 1.6, starting on p. 371 has more details: <https://archive.org/details/elementsofjewish00burnuoft/elementsofjewish00burnuoft/page/370/mode/2up>

ptomato:
Subscribed to that issue, thanks.

ptomato:
I guess there's been no movement on the ICU4X issue in 6 months, so it makes sense to support both calculations for now.

PR #4102, opened 2024-05-30
Adjust Temporal staging test of the Japanese calendar for alternative implementations

anba:
- Move tests for before-start-of-era into a different file until there's a better specification how to handle this case &rarr; <https://github.com/tc39/proposal-temporal/issues/2865>.
- Use helper function from #4098 to allow different era names.


ptomato (approved):
OK, thanks!

PR #4103, opened 2024-05-30
Add decorators tests

pzuraq:


pzuraq:
I've completed writing the tests, now I just need to actually get them to run properly and debug any issues (could be typos, etc).

nicolo-ribaudo:
https://github.com/babel/babel-test262-runner/ could be useful to run the tests

pzuraq:
Yes, I've been using that repo for testing in general. The previous commit for fields ([a3b4282](https://github.com/tc39/test262/pull/4103/commits/a3b42825b1ee1f47823a418a7465d133c20e202a)) was passing, at that point I kinda had all the details figured out for the overall structure so I scaffolded out all of the tests themselves and made sure I had written all of the ones we needed.

Next is just running them again to see if they work and fixing issues. I'm sure there are minor issues, definitely, but could also be some larger ones.

@nicolo-ribaudo one quick question, the generators in `make` throw an error if the folders don't exist for the output of a template/case file, so I've just been making them manually. I couldn't figure out the command to run to not throw, is there a way to do that easily?

ptomato:
@pzuraq We discussed this in yesterday's test262 meeting. We plan to start reviewing it after tackling one other large PR that's been waiting longer.

In the meantime, would you mind splitting the commit on this branch into two commits, one with any hand-written changes and one with autogenerated changes? We'll focus on the former.

pzuraq:
Ok, sounds good! I can get that done likely by the end of the month or early next month.

ptomato:
@pzuraq I went ahead and investigated which tests were autogenerated — it looks like they all are! :smile:

A few of them don't have corresponding templates in `src/`, maybe you forgot to add some files?
- `src/decorator/field-context-kind.case`
- `src/decorator/field-initializers-interleave.case`
- `src/decorator/field-instance-initializer-order.case`
- `src/decorator/field-received-value.case`
- `src/decorator/field-static-initializer-order.case`
- `src/decorator/decorator-test-123.case`

pzuraq:
@ptomato ah those files were mistakenly added actually, they were work in progress files as I was working out the way the generator system worked. Removing them and rebasing now.

fisker:
The diff is quite big, I didn't read all, but it will good to add [this case](https://github.com/tc39/proposal-decorators/issues/556#issue-3062870719) to tests if missing.

```js
class A {
  @decorators[0]
  method() {}
}
```

test/language/expressions/class/decorator/accessor/with-init/private/instance/decorator-order-application-field-and-accessor.js
```diff
@@ -0,0 +1,51 @@
...
+        }
+      }
+    }
+  }
+}
+
+
+var C = class {
+  @pushOrd(1, 3, 6)
+  @pushOrd(0, 4, 5)
```
o-, line 34: According to https://arai-a.github.io/ecma262-compare/?pr=2417&id=sec-initializefieldoraccessor the extraInitializers run in the order they were added by https://arai-a.github.io/ecma262-compare/?pr=2417&id=sec-createaddinitializerfunction to the context. So according to the spec this should be 0,1,4,3,5,6.

At least unless I am missing something. The polyfill claims that it should 1,2,3,4,5,6.

test/language/expressions/class/decorator/class/class-deco-returns-proxy.js
```diff
@@ -0,0 +1,33 @@
...
+
+
+var C = @dec
+ class {
+  
+}
+
+assert(C !== originalClass, 'values are not the same');
+assert(C instanceof originalClass, 'value is instance of class');
+assert.sameValue(new C(), 123);
```
o-, line 33: Shouldn't this fail with a TypeError if the 'construct' trap returns a non-object?

test/language/expressions/class/decorator/accessor/with-init/private/instance/decorator-order-application-field-and-accessor.js
pzuraq: The ordering here is correct. There are 3 phases:

1. Decorator application
2. Field initialization
3. Extra initializers

Phase two reverses the ordering of application from outermost -> innermost, because logically we are "setting" the field. The initial value is first created and then set, then the outer decorator gets the initial value and gets to mutate it, then the inner decorator gets the value, then it is set on the internal storage slot.

There is an issue on the decorators repo discussing this, and it was discussed at a Plenary because the change was made in stage 3 so needed consensus to merge.

test/language/expressions/class/decorator/class/class-deco-returns-proxy.js
pzuraq: Yes, but a Proxy is an object?

pzuraq: ah I see, sorry I was confused (still waking up).

Yes, this should error because `construct` needs to return an object, that's Proxy's behavior. Test should be updated to do that.

I don't have bandwidth to work on this at the moment but happy to take PRs to the branch or give access.

test/language/expressions/class/decorator/accessor/with-init/private/instance/decorator-order-application-field-and-accessor.js
o-: Right, application is inside out. So we first apply `@pushOrd(0, 4, 5)`, then `@pushOrd(1, 3, 6)`. According to https://arai-a.github.io/ecma262-compare/?pr=2417&id=sec-applydecoratorsanddefinemethod the extraInitializers are shared between all decorators of the same method. So the first one to run pushes a `() => ord.push(4)` and the second one `() => ord.push(3)` to the same extraInitializers list. And that is also the final order we are supposed to run them. So according to the spec I think this example should create 0,1,4,3,5,6.

test/language/expressions/class/decorator/class/class-deco-returns-proxy.js
o-: I don't have time to work on the tests either. I just stumble upon these issues since I am evaluating the coherence of spec vs. proposed V8 patch vs. babel.

o-: (btw. the line 32 fails too because of the proxy)

test/language/expressions/class/decorator/accessor/with-init/private/instance/decorator-order-application-field-and-accessor.js
o-: Sorry, I got confused. The 5,6 are the extra initializers.

So 3,4 are the normal initializers. And what you are saying is that they are supposed to run outside in. I guess that part I have not found in the spec yet.

o-: Ah, got it. `15.7.6 ApplyDecoratorsToElementDefinition [...] 8. If IsCallable(initializer) is true, prepend initializer to elementRecord.[[Initializers]].`

o-: However considering that the `[[Initializers]]` field comes from [ClassFieldDefinitionEvaluation](https://arai-a.github.io/ecma262-compare/?pr=2417&id=sec-runtime-semantics-classfielddefinitionevaluation) which already adds the actual initializer to that list, wouldn't the prepending cause the original initializer to run last?

I.e., I think according to the current spec text we are supposed to see 0,1,3,4,2,5,6.

pzuraq: This prepends the initialzer to the `[[Initializers]]` list:

> l. If kind is field, then
    i. If [IsCallable](https://arai-a.github.io/ecma262-compare/?pr=2417&id=sec-iscallable&secAll=true)(newValue) is true, prepend newValue to elementRecord.[[Initializers]].

Then we have initialization

>    1. [Assert](https://arai-a.github.io/ecma262-compare/snapshot.html?pr=2417#assert): elementRecord.[[Kind]] is field or accessor.
>    2. If elementRecord has a [[BackingStorageKey]] field, let fieldName be elementRecord.[[BackingStorageKey]].
>    3. Else, let fieldName be elementRecord.[[Key]].
>    4. Let initValue be undefined.
>    5. For each element initializer of elementRecord.[[Initializers]], do
>           a. Set initValue to ? [Call](https://arai-a.github.io/ecma262-compare/snapshot.html?pr=2417#sec-call)(initializer, receiver, « initValue »).
>    6. If fieldName is a [Private Name](https://arai-a.github.io/ecma262-compare/snapshot.html?pr=2417#sec-private-names), then
>           a. Perform ? [PrivateFieldAdd](https://arai-a.github.io/ecma262-compare/snapshot.html?pr=2417#sec-privatefieldadd)(receiver, fieldName, initValue).
>    7. Else,
>           a. [Assert](https://arai-a.github.io/ecma262-compare/snapshot.html?pr=2417#assert): [IsPropertyKey](https://arai-a.github.io/ecma262-compare/snapshot.html?pr=2417#sec-ispropertykey)(fieldName) is true.
>           b. Perform ? [CreateDataPropertyOrThrow](https://arai-a.github.io/ecma262-compare/snapshot.html?pr=2417#sec-createdatapropertyorthrow)(receiver, fieldName, initValue).
>    8. For each element initializer of elementRecord.[[ExtraInitializers]], do
>           a. Perform ? [Call](https://arai-a.github.io/ecma262-compare/snapshot.html?pr=2417#sec-call)(initializer, receiver).
>    9. Return unused.

Note that `[[Initializers]]` is used before `[[ExtraInitializers]]`, which holds actual initializers added by `addInitializers` to an element definition. Only Fields and Accessors have `[[Initializers]]`.

o-: Agreed. But my new worry is that when we call `prepend newValue to elementRecord.[[Initializers]]`, that list already contains the initializer `CreateFieldInitializerFunction(this, #element, ord.push(2))` from `ClassFieldDefinitionEvaluation`. This means prepending the initializers from the decorators will move the actual initializer to the end of the `[[Initializers]]` list. Resulting in `0,1,3,4,2,5,6`.

pzuraq: hmm, yes ok, you are correct. The prepended decorators are in the correct order, but the original initializer does need to be the first item in the list. This was likely introduced when the order was reversed, when initializers we appended to the list this behavior would have made sense.

test/language/expressions/class/decorator/class/class-deco-returns-proxy.js
nicolo-ribaudo: (it's still very helpful to get these comments, as we can probably find somebody else from the community to take over this PR)

test/language/expressions/class/decorator/class/error/class-deco-invalid-return-arrow.js
```diff
@@ -0,0 +1,26 @@
...
+
+function evaluate() {
+  var C = class {
+    @dec
+
+    #element() {}
+  }
+}
+
+assert.throws(TypeError, evaluate);
```
o-, line 25: Something is wrong here as we are not actually decorating a class as the description suggests.

test/language/expressions/class/decorator/accessor/with-init/private/instance/decorator-order-application-field-and-accessor.js
o-: Ok, I created an issue to track this https://github.com/tc39/proposal-decorators/issues/569

PR #4104, opened 2024-06-03
Remove current contents of implementation-contributed

dminor:
I believe we have consensus from the other implementers that the current contents of implementation-contributed are no longer useful. In preparation for upstreaming the SpiderMonkey `non262` tests, I'd like to remove the current contents.


ljharb (approved):
🔥

Ms2ger (approved):

ptomato:
At least this part seems to be non-controversial, let's land it

PR #4110, opened 2024-06-19
Fix a test using `console.log` instead of `print`

CanadaHonk:


linusg (approved):

ljharb (approved):

ptomato (changes requested):
This test probably shouldn't be using `print()` either — nothing will be listening to the output of `print()` unless the test has the `async` flag, and I believe it would be legal for implementations' test runners to fail the test on unexpected calls to `print()`.

If I understand correctly what this is supposed to test, I'd say let's just use `() => {}`.

test/staging/explicit-resource-management/disposable-stack-adopt-and-defer.js
```diff
@@ -49,7 +49,7 @@ assert.compareArray(valuesNormal, [44, 43, 42]);
 function TestDisposableStackDeferOnDisposedStack() {
   let stack = new DisposableStack();
   stack.dispose();
-  stack.defer(() => console.log(42));
+  stack.defer(() => print(42));
```
leobalter, line 52:
```suggestion
  stack.defer(() => { throw new TypeError("Code should not reach here"); });
  throw new Error("Code should not reach here");
```

It's been a while I use Test262, but I believe this might resolve the issue better and confirm no run runs post .defer as well.

leobalter: Given

``` 
3. If disposableStack.[[DisposableState]] is disposed, throw a ReferenceError exception.
4. If IsCallable(onDispose) is false, throw a TypeError exception.
``` 

Line 52 can also use a simple `stack.defer('')`;` so it can observe the ReferenceError in the correct order.

Ms2ger:
Hey @CanadaHonk - will you have time to look at the comments here?

Issue #4054, opened 2024-04-12
Math.sumPrecise testing plan


ptomato:
Boilerplate
  - [ ] Property descriptor
  - [ ] Function length
  - [ ] Function name
  - [ ] Is not a constructor
  - [ ] prototype is Function.prototype
  - [ ] is extensible
Test each of the following for both an array and a non-array iterable:
  - Success cases
    - [ ] Sum of finite numbers (same sign, mixed signs)
    - [ ] Sum of one number is that number (negative, -0, 0, positive, NaN, ∞, -∞)
    - [ ] Sum of empty is -0
    - [ ] Sum of finite numbers with infinite result (∞, -∞)
    - [ ] Sum of multiple numbers including any NaN is NaN
    - [ ] Sum including any ∞ and no NaN or -∞ is ∞
    - [ ] Sum including any -∞ and no NaN or ∞ is -∞
    - [ ] Sum of ∞ and -∞ is NaN
    - [ ] Sum of multiple -0 is -0
    - [ ] Sum of multiple 0 is 0
    - [ ] Sum of -0 and 0 is 0
  - Failure cases
    - [ ] rejects non-Number values in iterable (Boolean, String, Symbol, BigInt, Object, null, undefined)
Boilerplate for failure to conform to iterable protocol
  - [ ] @@iterator getter throws
  - [ ] @@iterator is not callable
  - [ ] @@iterator() throws
  - [ ] @@iterator() returns a primitive
  - [ ] iterator's `next` getter throws
  - [ ] `next` is not callable
  - [ ] `next()` throws
  - [ ] `next()` returns a primitive
  - [ ] iterator result's `done` getter throws
  - [ ] iterator result's `value` getter throws
  - [ ] original error is thrown when iterator's `return` getter throws
  - [ ] original error is thrown when `return` is not undefined and not callable
  - [ ] original error is thrown when `return()` throws
  - [ ] original error is thrown when `return()` returns a primitive
Other edge cases
  - [ ] Not passing an argument throws TypeError
  - [ ] summing each type of primitive throws TypeError
  - [ ] summing a non-iterable object throws TypeError
  - iterable yields more than 2⁵³ items (probably not testable)
  - [ ] A case where sumPrecise disagrees with naive sum via reduce
  - [ ] Order of observable operations

ptomato:
I'm trying to put together guidance on writing a testing plan, so I thought I'd use this proposal as a sample.

[...]

PR #4113, opened 2024-06-20
Add testing plan guide to documentation.

ioannad:
A guide to help with writing testing plans for ECMAScript proposals in stage 2.7.

### DIFF
--- a/docs/testing-plan-guide.md
+++ b/docs/testing-plan-guide.md
@@ -1,0 +1,107 @@
+# Testing plan guide
+
+#### What is this document
+
+[ECMAScript proposals](https://github.com/tc39/proposals/) that reach Stage 2.7 in the [TC39 Process](https://tc39.es/process-document/) need to have sufficient tests integrated to the Test262 test suite. To track the added tests, ensure that they are complete, and to get an idea of the amount of work needed, it's often helpful to have a testing plan outlining what exactly should be tested.
+
+This document is a guide intended to help proposal authors and test262 contributors create such testing plans. It includes detailed instructions for contributors less familiar with test262 testing or less familiar with composing a testing plan, as well as a [checklist](./#testing-plan-checklist) that may also be helpful to contributors more experienced with Test262 tests and testing plans.
+
+In the end of this document, in addition to the [checklist](./#testing-plan-checklist), there are also some [links](./#see-also) with further material.
+
+
+## Getting started creating a testing plan for an ECMAScript proposal.
+
+If you are looking for an ECMAScript proposal to write a testing plan for, we try to have [open issues asking for testing plans](https://github.com/tc39/test262/issues?q=is%3Aissue+is%3Aopen+testing+plan), for proposals in Stage 2.7 or 3. You can also see the [TC39 document](https://github.com/tc39/proposals/) tracking the status of all the open proposals, in case we missed something, in which case please create the missing testing plan issue.
+
+Before you start it's wise to communicate with proposal champion(s) in case they already have a testing plan and/or tests ready.
+
+Understand the proposal text:
+
+- read the explainer in the proposal repo's README, and
+- read the slides from any TC39 plenary presentation.
+
+If you're unfamiliar with ECMAScript spec text, read this guide on [how to read the ECMAScript Specification](https://timothygu.me/es-howto/).
+
+Read the proposal's rendered specification.
+
+Finally, list all the separate tests that need to be written, as explained in the following sections.
+
+### Tests for new APIs
+
+#### Boilerplate tests
+
+For new APIs, you can start by listing some standard tests, such as the following.
+
+- Property descriptor of the object/method/whatever. [Example](https://github.com/tc39/test262/blob/main/test/built-ins/String/prototype/at/prop-desc.js)
+- Brand check for methods, what happens when calling it with a `this` that is not what what it expects, via `Function.prototype.apply` or `.call`. [Example](https://github.com/tc39/test262/blob/main/test/intl402/ListFormat/prototype/format/branding.js)
+- Object prototype and extensibility. [Example](https://github.com/tc39/test262/blob/main/test/built-ins/Temporal/Instant/builtin.js)
+- Function boilerplate:
+  + Function length. [Example](https://github.com/tc39/test262/blob/main/test/built-ins/String/prototype/at/length.js)
+  + Function name. [Example](https://github.com/tc39/test262/blob/main/test/built-ins/String/prototype/at/name.js)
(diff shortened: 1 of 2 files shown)


docs/testing-plan-guide.md
```diff
@@ -0,0 +1,69 @@
...
+- Test both the syntax and how it's evaluated, so double the tests.
+- Check [CONTRIBUTING.md](https://github.com/tc39/test262/blob/main/CONTRIBUTING.md) for special frontmatter flags when testing syntax errors.
+- Look into the ECMA-262 sections named [Runtime Semantics: Evaluation](https://tc39.es/ecma262/#sec-array-initializer-runtime-semantics-evaluation).
+
+## Tips and tricks
+
+Check what engines test, in case you missed something. Some engine's test locations in their codebase:
+
+- v8 `v8/test/mjsunit/`
+- SpiderMonkey: `js/src/tests/non262/`
```
ptomato, line 58: These tests are the ones currently being moved into test262... maybe we should point to https://searchfox.org/mozilla-central/source/js/src/jit-test/tests instead? I don't know if that's the appropriate place to look.

ptomato (commented):
Looks like a good first draft to me. Before it's ready for publicizing there are a few things I'd suggest:

- We should make the document more approachable with a couple of introductory paragraphs, explaining what it is and who it is intended for.
- We should have a section about diminishing returns and advice on where to draw the line. See, for example, the discussion at https://github.com/tc39/test262/issues/4054 and in Mike's testing plan guide for WPT.
- We should have a section on what steps to carry out, concretely. Things like, open a testing plan issue on the test262 repo. (Examples of good existing testing plan issues.) CC the proposal champions (if that is not you yourself). Send a PR to tc39/proposals to link the testing plan issue.

docs/testing-plan-guide.md
ioannad: > These tests are the ones currently being moved into test262... maybe we should point to https://searchfox.org/mozilla-central/source/js/src/jit-test/tests instead? I don't know if that's the appropriate place to look.

IIUC .../jit-test/tests are focused more on JIT compiler behaviour, however it is indeed the case that more general tests are added there as well. Adding this link to the guide with this "warning".

docs/testing-plan-guide.md
```diff
@@ -0,0 +1,107 @@
+# Testing plan guide
+
+#### What is this document
+
+[ECMAScript proposals](https://github.com/tc39/proposals/) that reach Stage 2.7 in the [TC39 Process](https://tc39.es/process-document/) need to have sufficient tests integrated to the Test262 test suite. To track the added tests, ensure that they are complete, and to get an ideao of the amount of work needed, it's often helpful to have a testing plan outlining what exactly should be tested.
```
ptomato, line 5:
```suggestion
[ECMAScript proposals](https://github.com/tc39/proposals/) that reach Stage 2.7 in the [TC39 Process](https://tc39.es/process-document/) need to have sufficient tests integrated to the Test262 test suite. To track the added tests, ensure that they are complete, and to get an idea of the amount of work needed, it's often helpful to have a testing plan outlining what exactly should be tested.
```

docs/testing-plan-guide.md
```diff
@@ -0,0 +1,107 @@
...
+#### Failure cases
+
+List failure cases, when an exception is thrown:
+
+- Check new spec text for every place where a `?` appears, indicating that what follows may throw an exception.
+- List other potential unexpected edge cases. At the rendered spec, use the `U` key to see who calls the related parts of the code, and check for unexpected entry points.
+
+#### Side-effects and algorithm step order
+
+Finally, consider testing the order of the algorithm steps, if there are steps that introduce side effects. However, it's best to not overdo it here, the goal is not to have 100% coverage, but to help implementations correctly implement the new functionality.
```
ptomato, line 59: "the goal is not to have 100% coverage" — I'd make this more specific. We do want what most people understand as "100% coverage", but we don't want endless tests for every single permutation of user code calls, for example asserting the user calls for some ridiculous object like `{ toString() { return { toString() { return { toString() { ... } } } } } }` in a ToString algorithm step. Although this is already somewhat explained below, in "Completing the testing plan".

Maybe something like, "However, consider that exhaustively testing the order of side effects can lead to diminishing returns."

docs/testing-plan-guide.md
```diff
@@ -0,0 +1,107 @@
...
+
+A testing plan for new syntax is similar to that for a new API but additionally:
+
+- Test both the syntax and how it's evaluated, so double the tests.
+- Check [CONTRIBUTING.md](https://github.com/tc39/test262/blob/main/CONTRIBUTING.md) for special frontmatter flags when testing syntax errors.
+- Look into the ECMA-262 sections named [Runtime Semantics: Evaluation](https://tc39.es/ecma262/#sec-array-initializer-runtime-semantics-evaluation).
+
+## Tips and tricks
+
+Check what engines test, in case you missed something. Some engine's test locations in their codebase:
```
ptomato, line 71:
```suggestion
Check what engines test, in case you missed something. Some engines' test locations in their codebase:
```

ptomato (approved):
This seems like really good guidance. Any other maintainers want to take a look before merging?

docs/testing-plan-guide.md
```diff
@@ -0,0 +1,107 @@
...
+
+- `undefined`, `null`, symbols
+- strings, "", "\b", "\n"
+- Booleans
+- number, BigInt, NaN, ±0, ±∞
+- Objects, Arrays
+
+## Completing the testing plan
+
+As important as it is to have a complete testing plan, it is equally important to avoid diminishing returns and when to stop adding cases to be tested. Especially when new specification algorithms call on existing algorithms, it may be temping to dive in the existing algorithms to explore the different paths they may take. However it is best to focus on writing more surface tests for these cases, and only dive in the nested algorithms when the proposal is expecting these to be relevant to the new additions.
```
Ms2ger, line 89:
```suggestion
As important as it is to have a complete testing plan, it is equally important to avoid diminishing returns and to understand when to stop adding cases to be tested. Especially when new specification algorithms call on existing algorithms, it may be temping to dive in the existing algorithms to explore the different paths they may take. However it is best to focus on writing more surface tests for these cases, and only dive in the nested algorithms when the proposal is expecting these to be relevant to the new additions.
```
or something like it

Ms2ger (approved):
Do we want to link this from somewhere?

ptomato:
I added a blurb in `CONTRIBUTING.md` that will hopefully catch proposal authors' notice.

gibson042 (approved):

Ms2ger (approved):

ljharb (approved):

PR #4119, opened 2024-06-27
Temporal API removals

ptomato:
These are the test changes for the remaining API removals which reached TC39 consensus in June 2024: Temporal.Calendar, Temporal.TimeZone, the getCalendar methods, Temporal.ZonedDateTime.prototype.getTimeZone, and the getISOFields methods.

I'm well aware that this PR is thousands of files. I've tried to organize it logically into commits, so I recommend reviewing it commit by commit.

The bulk of the PR is the wholesale removal of 2250 files in the commit "Temporal: Remove tests directly relating to custom calendar and time zones". I would recommend skipping that one; you can verify mechanically that it's just file deletions.

That still leaves 1000 files to review. 200 of those are in "Temporal: Fix tests that unnecessarily create TimeZone or Calendar instances" which is a trivial, nearly mechanical replacement. Of the remaining 800, over half are spread out over the smaller commits, so shouldn't be too hard to review.

The only really tedious review will be of the commit "Temporal: Test adjustments for removing calendar and time zone objects", with 350 files, which consists of all of the various rewrites that we have to do, no way around it, that weren't easily organizable into other categories. However, a lot of these are deletions as well, so hopefully that's not too bad.

I'm open to suggestions for other ways to split this up to make reviews easier.


justingrant (approved):
Wow, amazing. I spot-checked a few files and they looked good. Thanks so much for persevering through this huge effort!

Ms2ger (approved):
lgtm

PR #4121, opened 2024-06-28
Avoid duplicate binding import-source-binding-name.js

nicolo-ribaudo:
The current test contained a parser error due to the duplicate binding. This splits the test in two files, so that we actually successfully parse the file.

cc @legendecas

Ref https://github.com/babel/babel/pull/16596

### DIFF
--- a/test/language/module-code/source-phase-import/import-source-binding-name-2.js
+++ b/test/language/module-code/source-phase-import/import-source-binding-name-2.js
@@ -1,0 +1,23 @@
+// Copyright (C) 2024 Chengzhong Wu. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+description: >
+  ImportBinding in ImportDeclaration may be 'source' and 'from'
+esid: sec-modules
+info: |
+  ImportDeclaration:
+    import source ImportedBinding FromClause ;
+
+negative:
+  phase: resolution
+  type: SyntaxError
+features: [source-phase-imports]
+flags: [module]
+---*/
+
+$DONOTEVALUATE();
+
+import "../resources/ensure-linking-error_FIXTURE.js";
+
+import source source from '<do not resolve>';
+import source from from '<do not resolve>';
--- a/test/language/module-code/source-phase-import/import-source-binding-name.js
+++ b/test/language/module-code/source-phase-import/import-source-binding-name.js
@@ -19,5 +19,5 @@ $DONOTEVALUATE();
 
 import "../resources/ensure-linking-error_FIXTURE.js";
 
-import source from from '<do not resolve>';
+import source from '<do not resolve>';
 import from from '<do not resolve>';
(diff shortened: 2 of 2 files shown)


test/language/module-code/source-phase-import/import-source-binding-name-2.js
```diff
@@ -0,0 +1,23 @@
...
+features: [source-phase-imports]
+flags: [module]
+---*/
+
+$DONOTEVALUATE();
+
+import "../resources/ensure-linking-error_FIXTURE.js";
+
+import source source from '<do not resolve>';
+import source from from '<do not resolve>';
```
legendecas, line 23: I think both statements declare `source`

test/language/module-code/source-phase-import/import-source-binding-name.js
```diff
@@ -19,7 +19,5 @@ $DONOTEVALUATE();
 
 import "../resources/ensure-linking-error_FIXTURE.js";
 
-import source source from '<do not resolve>';
 import source from from '<do not resolve>';
-import source from '<do not resolve>';
 import from from '<do not resolve>';
```
legendecas, line 23: Both statements declare `from`

legendecas (commented):
Ahh, thanks!

test/language/module-code/source-phase-import/import-source-binding-name-2.js
nicolo-ribaudo: Whops, right

legendecas (approved):
thanks~

Ms2ger (approved):

woess:
@nicolo-ribaudo I thought the whole point of these tests was that the `ImportBinding` may be `source` and `from`, so why do they expect a `SyntaxError` then? Isn't this a bug?

nicolo-ribaudo:
It's `ensure-linking-error_FIXTURE.js` that causes the syntax error. The key bit here is that the error happens in the "resolution" (i.e. linking) phase, and not in the parse phase.

woess:
@nicolo-ribaudo My bad, I missed that part; thanks for the clarification.

Although I still think there's a problem with these tests since AFAIK, the correct evaluation order would be to try to load '<do not resolve>' first before linking `ensure-linking-error_FIXTURE.js`, so you wouldn't even get to the syntax error. This is also the behavior I see from node, btw:
```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '<do not resolve>' imported from …/language/module-code/source-phase-import/import-source-binding-name.js
```

So I think the tests should be changed to point to an existing file, so that the resolution phase will throw the expected SyntaxError.

woess:
I've opened https://github.com/tc39/test262/issues/4193 to track this issue.

legendecas:
Yes, I found the trick didn't work in my V8 CL as well: https://chromium-review.googlesource.com/c/v8/v8/+/5783137. The failure to read the file happens before the linking failure, which leads to a host defined error, instead of SyntaxError.

legendecas:
The discussion at https://github.com/tc39/test262/pull/3980#discussion_r1646096110 was not documented in INTERPRETING.md. I will add a note in the doc.

PR #4130, opened 2024-07-03
Test coverage for canonicalization of calendars

ptomato:
This functionality in `Intl.DateTimeFormat` was specified several years ago in ECMA-402 but lacked coverage.

Anba recently brought Temporal's calendar canonicalization in line with ECMA-402. Add test coverage for that as well and fix one erroneous test.

### DIFF
--- a/test/built-ins/Temporal/PlainDate/calendar-case-insensitive.js
+++ b/test/built-ins/Temporal/PlainDate/calendar-case-insensitive.js
@@ -7,7 +7,14 @@ description: Calendar names are case-insensitive
 features: [Temporal]
 ---*/
 
-const arg = "iSo8601";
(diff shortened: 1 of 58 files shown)


justingrant (approved):
👍

Ms2ger (approved):
CanonicalizeUValue also ASCII-lowercases. You might consider testing that as well (including Turkish dotted capital I, if you like). Followup is fine.

ptomato:
Lowercasing was tested in most cases already, but Turkish dotted capital I was not. I figured if I punt it to a followup, I'd never do it, so I've added it to this PR. Let me know what you think.

Ms2ger (approved):
Thanks!

PR #4131, opened 2024-07-05
Remove unused harness/timer.js

linusg:
As far as I can tell the last reference to this file was removed in a2b45dd33bb5d646f7b7cea53946ad0f643b4a7b. There is a second implementation that is still in use:

https://github.com/tc39/test262/blob/3a7a72aef5009eb22117231d40f9a5a66a9a595a/harness/atomicsHelper.js#L39-L72


ptomato (approved):
Thanks for noticing the unused code!

I'll leave this unmerged for the time being and ask if any downstream consumer's test harness was using it.

PR #4133, opened 2024-07-08
Tests for later base64 changes

bakkot:
Tests for https://github.com/tc39/proposal-arraybuffer-base64/pull/60 and https://github.com/tc39/proposal-arraybuffer-base64/pull/58, which got consensus after the proposal had achieved stage 3 based on user and implementer feedback.

Based on https://github.com/tc39/test262/pull/3994 (which incidentally is ready).

### DIFF
--- a/test/built-ins/Uint8Array/prototype/setFromBase64/writes-up-to-error.js
+++ b/test/built-ins/Uint8Array/prototype/setFromBase64/writes-up-to-error.js
@@ -1,0 +1,25 @@
+// Copyright (C) 2024 Kevin Gibbons. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-uint8array.prototype.setfrombase64
+description: Uint8Array.prototype.setFromBase64 decodes and writes chunks which occur prior to bad data
+features: [uint8array-base64, TypedArray]
+---*/
+
+var target = new Uint8Array([255, 255, 255, 255, 255]);
+assert.throws(SyntaxError, function() {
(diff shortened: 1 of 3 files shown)


test/built-ins/Uint8Array/prototype/setFromBase64/writes-up-to-error.js
```diff
@@ -0,0 +1,25 @@
...
+assert.throws(SyntaxError, function() {
+  target.setFromBase64("MjYyZg", { lastChunkHandling: "strict" });
+}, "padding omitted with lastChunkHandling: strict");
+assert.compareArray(target, [50, 54, 50, 255, 255], "decoding from MjYyZg should only write the valid chunks");
+
+target = new Uint8Array([255, 255, 255, 255, 255]);
+assert.throws(SyntaxError, function() {
+  target.setFromBase64("MjYyZg===");
+}, "extra characters after padding");
+assert.compareArray(target, [50, 54, 50, 255, 255], "decoding from MjYyZg=== should not write the last chunk because it has extra padding ");
```
ljharb, line 25:
```suggestion
assert.compareArray(target, [50, 54, 50, 255, 255], "decoding from MjYyZg=== should not write the last chunk because it has extra padding");
```

ljharb (commented):
The "writes up to error" tests fail on my polyfill, so I'll review those after trying to fix my bug, but otherwise they look OK. omit-padding tests pass for me.

ljharb (approved):
confirmed, once i updated my implementation to match the updated spec, tests all pass.

PR #4125, opened 2024-07-01
RAB: Integrate staging tests for the .filter method

ioannad:
of Array.prototype and TypedArray.prototype

This is part of PR https://github.com/tc39/test262/pull/3888 to make reviewing easier. Includes changes to use the helper ./harness/resizableArrayBufferUtils.js

PR #4134, opened 2024-07-08
RAB: Integrate staging tests for the .forEach method

ioannad:

This one also did not have any ..-shrink-.. tests under Array/prototype, neither in #3888 nor in the staging tests themselves.


ptomato (approved):
Like one of the previous PRs, this one is missing `Array/prototype/forEach/resizable-buffer-shrink-mid-iteration.js` - no need to block the PR but let's write it down for later.

ioannad:
> Like one of the previous PRs, this one is missing `Array/prototype/forEach/resizable-buffer-shrink-mid-iteration.js` - no need to block the PR but let's write it down for later.

@ptomato I think I misunderstood your [previous comment](https://github.com/tc39/test262/pull/4125#pullrequestreview-2155080845) about the missing `Array/prototype/forEach/resizable-buffer-shrink-mid-iteration.js` tests. Although there were no staging tests for this, I could try to add some in the obvious ways, parallel to the TypedArray "shrink" tests.

ptomato (approved):

ptomato:
What I understood from the shrink test and reading the spec:
- `forEach` skips elements that return false from [[HasProperty]]
- [[HasProperty]] returns false if the whole typed array is out of bounds (resized smaller than fixed length), or if the index is out of bounds

PR #4139, opened 2024-07-09
RAB: Adds missing Array.prototype.filter 'shrink' tests plus typesetting

ioannad:
Also renames `ResizeBufferMidIteration` to `ResizeMidIteration`.


ptomato (approved):
Thanks!
The added shrink test for `filter` seems like the same explanation as the one you already added for `forEach`, re. HasProperty and out-of-bounds.

PR #4141, opened 2024-07-09
Update firstDay tests to latest spec after PR 79

FrankYFTang:
Spec Changes merged
https://github.com/tc39/proposal-intl-locale-info/pull/79


FrankYFTang:
@anba @ben-allen @sffc

FrankYFTang:
@trflynn89

FrankYFTang:
@Ms2ger

ptomato (approved):
OK, this matches my understanding of the change in the 2023-11 meeting.

test/intl402/Locale/constructor-options-firstDayOfWeek-valid.js
```diff
@@ -50,6 +47,24 @@ const validFirstDayOfWeekOptions = [
   [6, "en-u-fw-sat"],
   [7, "en-u-fw-sun"],
   [0, "en-u-fw-sun"],
+  [true, "en-u-fw-true"],
```
trflynn89, line 50: @FrankYFTang is this one correct? IIRC `true` is dropped from canonical locales. And that's how v8, SpiderMonkey, and LibJS all behave:
```js
> new Intl.Locale('en-u-fw-true').toString()
"en-u-fw" 

> new Intl.Locale('en-u-fw-false').toString()
"en-u-fw-false"
```

ptomato: I didn't see that in [CanonicalizeUValue](https://tc39.es/ecma402/#sec-canonicalizeuvalue), where should I be looking for that behvaiour?

trflynn89: It would fall under step 2, which defers to TR-35.

It's in TR-35: https://unicode.org/reports/tr35/#Canonical_Unicode_Locale_Identifiers
> A unicode_locale_id has canonical syntax when:
> * Any type or tfield value "true" is removed.

FrankYFTang: ah... that is true

PR #4148, opened 2024-07-11
Update the expctation of constructor-options-firstDayOfWeek-valid.js

FrankYFTang:
true will be dropped inside UTS35
See https://github.com/tc39/test262/pull/4141#pullrequestreview-2171902863 for more info


FrankYFTang:
@ trflynn89 @ptomato

ptomato (approved):
Thanks, I looked it up this morning and I agree.

PR #4149, opened 2024-07-12
Add missing test for `Math.pow(1, NaN)`

magic-akari:
In some implementations, the `pow` function may return 1, which contradicts the specification. This addition ensures that the test case covers this scenario.

### DIFF
--- a/test/built-ins/Math/pow/applying-the-exp-operator_A1.js
+++ b/test/built-ins/Math/pow/applying-the-exp-operator_A1.js
@@ -18,7 +18,7 @@ base[5] = 0.000000000000001;
 base[6] = 1.7976931348623157E308; //largest finite number
 base[7] = +Infinity;
 base[8] = NaN;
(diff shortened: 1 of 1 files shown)


ljharb (approved):
seems fine, but i'm not sure why it would be relevant to rust

test/built-ins/Math/pow/applying-the-exp-operator_A1.js
```diff
@@ -18,7 +18,8 @@ base[5] = 0.000000000000001;
 base[6] = 1.7976931348623157E308; //largest finite number
 base[7] = +Infinity;
 base[8] = NaN;
-var basenum = 9;
+base[9] = 1; // In some implementations, the `pow` function might incorrectly return 1, deviating from the ECMA specification.
```
Ms2ger, line 21: I don't think the comment is needed
```suggestion
base[9] = 1;
```

Ms2ger (approved):

PR #4160, opened 2024-07-22
 RAB: Integrate staging tests for the .reverse method

ioannad:

### DIFF
--- a/test/built-ins/Array/prototype/reverse/resizable-buffer.js
+++ b/test/built-ins/Array/prototype/reverse/resizable-buffer.js
@@ -4,227 +4,161 @@
 /*---
 esid: sec-array.prototype.reverse
 description: >
-  Array.p.reverse behaves correctly when receiver is backed by
-  resizable buffer
-includes: [compareArray.js]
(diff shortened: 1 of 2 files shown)


ptomato (approved):
Looks good. :+1:

I was wondering if there's a possibility to test mid-iteration growing and shrinking. But I think there is not. There's no callback or argument conversion which we could use to resize the ArrayBuffer. There are property accesses on the TypedArray but you cannot redefine numeric property keys on a TypedArray. You could redefine `length` but that's not mid-iteration, it'd be essentially the same as what we're already testing.

ioannad:
> I was wondering if there's a possibility to test mid-iteration growing and shrinking. But I think there is not. There's no callback or argument conversion which we could use to resize the ArrayBuffer. There are property accesses on the TypedArray but you cannot redefine numeric property keys on a TypedArray. You could redefine `length` but that's not mid-iteration, it'd be essentially the same as what we're already testing.

@ptomato do you think we should open an issue to figure out how to write such tests?

ptomato:
I'm not sure it's possible.

PR #4162, opened 2024-07-23
Remove unneeded /u flag within harness/nativeFunctionMatcher

p-bakker:

Relates to ##3032


ptomato (commented):
Thanks for the PR!

I can't confirm if the `u` flag is really unnecessary here because I don't know all the subtleties of Unicode-aware regexp parsing. Is there anyone else who could?

p-bakker:
The selection of \uXXXX-es in both regexes from which I removed the `u` flag are all non-astral-plain characters, so the `u` flag does nothing in this case

ptomato (approved):
Thanks.

PR #4163, opened 2024-07-23
Don't use const in for (of) loops so harness code can be used in environments that don't support that

p-bakker:

Relates to ##3032


ptomato (approved):
Thanks!

PR #4164, opened 2024-07-23
Prevent harness code not loading when async/generator/asyncGenerator not supported

p-bakker:

Relates to ##3032

### DIFF
--- a/harness/hidden-constructors.js
+++ b/harness/hidden-constructors.js
@@ -10,6 +10,18 @@ defines:
   - GeneratorFunction
 ---*/
 
-var AsyncFunction = Object.getPrototypeOf(async function () {}).constructor;
-var AsyncGeneratorFunction = Object.getPrototypeOf(async function* () {}).constructor;
-var GeneratorFunction = Object.getPrototypeOf(function* () {}).constructor;
+var AsyncFunction;
+var AsyncGeneratorFunction;
+var GeneratorFunction;
+
+try {
+  AsyncFunction = Object.getPrototypeOf(new Function('async function () {}')).constructor;
+} catch(e) {}
+
+try {
+  Object.getPrototypeOf(new Function('async function* () {}')).constructor;
+} catch(e) {}
+
+try {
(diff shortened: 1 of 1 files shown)


harness/hidden-constructors.js
```diff
@@ -10,6 +10,18 @@ defines:
...
+var AsyncFunction;
+var AsyncGeneratorFunction;
+var GeneratorFunction;
+
+try {
+  AsyncFunction = Object.getPrototypeOf(new Function('async function () {}')).constructor;
+} catch(e) {}
+
+try {
+  Object.getPrototypeOf(new Function('async function* () {}')).constructor;
```
ptomato, line 22:
```suggestion
  AsyncGeneratorFunction = Object.getPrototypeOf(new Function('async function* () {}')).constructor;
```

harness/hidden-constructors.js
```diff
@@ -10,6 +10,18 @@ defines:
...
+} catch(e) {}
+
+try {
+  Object.getPrototypeOf(new Function('function* () {}')).constructor;
```
ptomato, line 26:
```suggestion
  GeneratorFunction = Object.getPrototypeOf(new Function('function* () {}')).constructor;
```

ptomato (changes requested):
Thanks for the PR!

I'm less sure about this one, because this harness file is only used in 4 tests which are all guaranteed to fail anyway if async or generators aren't supported. So doesn't really matter if they fail in the harness file.

Or is the problem that you get a SyntaxError instead of a runtime error? (In that case, could we use the `new Function('...')` without swallowing the exception with `catch(e) {}`? I tend not to like unconditionally dropping exceptions...)

ljharb (approved):

p-bakker:
@ptomato the issue is any error: if for example only one of the three flavors isn't supported, that shouldn't stop/prevent the other two variables from being instantiated.

I don't see how that can be achieved without catching potential exceptions

ptomato:
Got it, thanks. I understand now.

PR #4165, opened 2024-07-23
Make resizableArrayBufferUtils not depend on syntax that may not be supported

p-bakker:

Relates to ##3032

### DIFF
--- a/harness/resizableArrayBufferUtils.js
+++ b/harness/resizableArrayBufferUtils.js
@@ -18,14 +18,10 @@ defines:
 features: [BigInt]
 ---*/
 
-class MyUint8Array extends Uint8Array {
-}
-
-class MyFloat32Array extends Float32Array {
-}
-
-class MyBigInt64Array extends BigInt64Array {
-}
+// Using new Function()(); instead of just 'class x extends Y' as to not bomb out when `class` isn't supported
+const MyUint8Array = new Function('return class MyUint8Array extends Uint8Array {}')();
+const MyFloat32Array = new Function('return class MyFloat32Array extends Float32Array {}')();
+const MyBigInt64Array = new Function('return class MyBigInt64Array extends BigInt64Array {}')();
 
 const builtinCtors = [
   Uint8Array,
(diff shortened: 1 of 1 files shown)


harness/resizableArrayBufferUtils.js
```diff
@@ -18,15 +18,25 @@ defines:
 features: [BigInt]
 ---*/
 
-class MyUint8Array extends Uint8Array {
-}
+// Helper to extend built-ins without using 'class x extends Y' syntax, which might not be supported
+function subClass(name, superType) {
+   var c = function (length) {
+     var array = new superType(length);
```
ljharb, line 24: this is not the same for a number of reasons; I'm all for not requiring `class` syntax, but there shouldn't be an engine anywhere that lacks `class` that does have RABs?

p-bakker: Rhino support ArrayBuffer and (most) TypedArray flavors, but not yet RABs, but I could see that being added way before we get around to supporting `class`

Although my subClass approach may not be the exact same as `class X extends Y` (and I'm curious how exactly), is the difference such that it changes the behaviour of the test here? 

Another approach I could take is checking if `class` is supported and if not, exclude the MyXxxxxArray variants from being created and added

ljharb: That seems very confusing; why would you add things newer than ES2015 but not add `class`? That doesn't make any sense to me.

`class extends` isn't always simple syntax sugar for ES5 inheritance. It'd be better to use `Function` inside a try/catch with the proper syntax, I think.

p-bakker: > why would you add things newer than ES2015 but not add class? That doesn't make any sense to me.

Confusing indeed, but Rhino follows a different path than most regular JavaScript engines. And all the TypedArrays are just host objects, rather easy to implement, which implementing `class` and everything that comes with it is in a whole different ballpark

> It'd be better to use Function inside a try/catch with the proper syntax, I think

You mean something like:
```
var MyFloat32Array;
try {
  MyFloat32Array = new Function('class MyFloat32Array extends Float32Array {}');
} catch (e) {}
``` 
AFAICT that also doesn't yield a conrrectly extended Float32Array

ljharb: if you `return class` then it should.

p-bakker: Ah, you mean like this? 

```
var x;

try {
    x = new Function('return class x extends Array{}')();
} catch (e) {}
```

p-bakker: Applied the changes, please review

ptomato (approved):
Thanks. This time I made sure the resizable ArrayBuffer tests do still run on implementations that support them :sweat_smile:

harness/resizableArrayBufferUtils.js
```diff
@@ -18,14 +18,10 @@ defines:
...
-
-class MyFloat32Array extends Float32Array {
-}
-
-class MyBigInt64Array extends BigInt64Array {
-}
+// Using new Function()(); instead of just 'class x extends Y' as to not bomb out when `class` isn't supported
+const MyUint8Array = new Function('return class MyUint8Array extends Uint8Array {}')();
+const MyFloat32Array = new Function('return class MyFloat32Array extends Float32Array {}')();
+const MyBigInt64Array = new Function('return class MyBigInt64Array extends BigInt64Array {}')();
```
ljharb, line 24: is it ok that these exceptions aren't caught?

p-bakker: Sigh, no, they are not. I'm sorry, juggling too many things at once and trying to rush it through, shouldn't have done that...

Will send an additional PR tomorrow... Sorry again for keeping you all occupied...

Issue #4118, opened 2024-06-26
Automatically add a results table for CI runs on pull requests


ptomato:
I've noticed several times that the meaning of the CI results is confusing for contributors:

![image](https://github.com/tc39/test262/assets/459371/723d0476-4be7-41e3-ae56-fd5db0f2b73a)

This looks like it is saying "all engines passed these tests." What it really means is "all engines installed correctly and executed these tests without crashing, regardless of pass or fail."

![image](https://github.com/tc39/test262/assets/459371/e85308ab-319b-41f1-b352-c1e25cf2f95a)

Likewise, this looks like it means "ChakraCore failed these tests, all other engines passed." What it really means is "ChakraCore failed to install."

I think we could reduce this confusion by automatically posting individual test results to a table in a PR comment. For example, the table for #4052 might look like this:

| Test name | engine262 | ESMeta | JSC | SpiderMonkey | V8 | V8 --harmony | XS |
| --- | --- | --- | --- | --- | --- | --- | --- |
| language/expressions/assignment/target-member-computed-reference.js | :x: |  ✅ | :x: | ✅ | :x: | ✅ | :x: |
| language/expressions/assignment/target-super-computed-reference.js | ✅ | :x: | ✅ | :x: | ✅ | :x: | ✅ |

(pass/fail status is entirely made up and not a comment on any engine)

[`ci_test.sh`](https://github.com/tc39/test262/blob/main/tools/scripts/ci_test.sh) already runs only the added and modified tests on each engine, and just discards the results; we could easily consume those results to build the table.

ljharb:
That would be great. What would be _even better_ if the table only included results that were *different* from the last successful run on main.

PR #4166, opened 2024-07-24
Revert "Prevent harness code not loading when async/generator/asyncGenerator not supported"

anba:
Reverts tc39/test262#4164 because it added incorrect code:

```js
js> Object.getPrototypeOf(new Function('async function () {}')).constructor;
typein:3:16 SyntaxError: function statement requires a name:
typein:3:16 async function () {}
typein:3:16 ...............^
Stack:
  @typein:1:23
```

### DIFF
--- a/harness/hidden-constructors.js
+++ b/harness/hidden-constructors.js
@@ -10,18 +10,6 @@ defines:
   - GeneratorFunction
 ---*/
 
-var AsyncFunction;
-var AsyncGeneratorFunction;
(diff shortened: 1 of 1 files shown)


ptomato:
Thanks, @anba. Alternatively, #4167 should fix the problem.

Sorry about that. I guess #4118 would have prevented this.

ptomato:
Hopefully fixed now.

PR #4169, opened 2024-07-25
Follow-up to #4165

p-bakker:
Added catch to achieve the entire goal of the previous PR to prevent the harness from loading if 'class' syntax isn't supported

Hopefully this is now properly sorted


ptomato (approved):
I think it is, thanks.

PR #4170, opened 2024-07-25
Add tool to write multi-module tests in a single file

nicolo-ribaudo:
I am writing tests for the [`import defer`](https://github.com/tc39/proposal-defer-import-eval) proposal, and all of them rely on having multiple files.

I have noticed that having all the modules in a single file significantly improves the test reading experience:
- on GitHub PR diffs, I can click on "show more" to expand the diff and see what are the other files
- when a file has changes, I can easily see what it would affect in other files
- I can easily send a single test to other people for review, rather than having to send multiple files

I was using a custom tool just for me, but it would be great to have it committed in test262 similarly to how we already have .template/.case tests.

How do you maintainers feel about this? Is it ok if the script runs on Node.js rather than python? Should I integrate it with CI?


ptomato:
I like this idea. It makes a lot of sense for several files that are conceptually one test to also be able to live in the same place, and the Markdown format seems convenient for sharing. Writing the tool in JS is also fine, I think that is more appropriate going forward than continuing to write test262 development tools in Python.

Are there any existing multi-file tests where the fixtures are shared between more than one test?

Small comment, I'd prefer if it didn't append `__generated` to the folder name.

This would need to be integrated with CI the same way the case/template generator is.

ljharb:
I'm less enthused - this seems like a lot of overhead and complexity just to avoid making multiple files inside a "fixtures" directory or similar.

There's tons of easy ways to send multiple files for review - codesandbox, gist, etc - so I'm not sure why that's much of a benefit.

PR #4171, opened 2024-07-25
Temporal: Fix Japanese era test

ptomato:
This mistakenly used a previous name for the helper function. Also adds a couple of debugging messages.

(Regression from #4102)


ptomato:
cc @anba

Ms2ger (approved):

PR #4174, opened 2024-07-29
 RAB: Integrate staging tests for the .slice method

ioannad:


ptomato (commented):
Should we also have `coerced-start-shrink.js` for the Array.prototype.slice?

We could also consider testing resizing during coercion of the `end` argument. Currently these tests only test what happens when coercion of the `start` argument resizes the buffer.

ioannad:
> Should we also have `coerced-start-shrink.js` for the Array.prototype.slice?
> 
> We could also consider testing resizing during coercion of the `end` argument. Currently these tests only test what happens when coercion of the `start` argument resizes the buffer.

Added the tests you suggested, and an additional `resizable-buffer.js` test file for `Array.p.slice`.

ptomato (approved):
Thanks!

PR #4177, opened 2024-07-30
 RAB: Integrate staging tests for the .subarray method

ioannad:
of TypedArray.prototype


ptomato (commented):
Similar to my comment on `slice()`, should we test what happens when the buffer is resized during coercion of the `end` argument?

ioannad:
@ptomato there were already some tests for 'end' in the tests, so I changed the test file names to indicate this, but I also addded more tests for 'end'.

ptomato (approved):
Thanks!

PR #4178, opened 2024-07-30
RAB: Integrate staging tests for the .toLocaleString method

ioannad:


test/built-ins/Array/prototype/toLocaleString/resizable-buffer.js
```diff
@@ -0,0 +1,82 @@
...
+  // Orig. array: [0, 2, 4, 6]
+  //              [0, 2, 4, 6] << fixedLength
+  //                    [4, 6] << fixedLengthWithOffset
+  //              [0, 2, 4, 6, ...] << lengthTracking
+  //                    [4, 6, ...] << lengthTrackingWithOffset
+
+  assert.sameValue(Array.prototype.toLocaleString.call(fixedLength), '0,2,4,6');
+  assert.sameValue(Array.prototype.toLocaleString.call(fixedLengthWithOffset), '4,6');
+  assert.sameValue(Array.prototype.toLocaleString.call(lengthTracking), '0,2,4,6');
+  assert.sameValue(Array.prototype.toLocaleString.call(lengthTrackingWithOffset), '4,6');
```
ptomato, line 35: As per https://tc39.es/ecma262/#sec-array.prototype.tolocalestring, we are assuming the separator is `","` even though it's implementation-defined.

In fact, the spec text itself suggests using `", "` with a space. Engine262 does this, while all other engines that I tested use `","` without a space.

I think we might have to write these tests to be a bit more agnostic to the localized output.

Note you can get the separator by using `[undefined,undefined].toLocaleString()`, and we should also call `toLocaleString` with the argument `"en-u-nu-latn"` to ensure that we get the right digit representations regardless of the locale in which the tests are executed.

test/built-ins/Array/prototype/toLocaleString/user-provided-tolocalestring-shrink.js
```diff
@@ -0,0 +1,66 @@
+// Copyright 2023 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-array.prototype.tolocalestringb
```
ptomato, line 5:
```suggestion
esid: sec-array.prototype.tolocalestring
```

ptomato (changes requested):

test/built-ins/Array/prototype/toLocaleString/resizable-buffer.js
ioannad: Done.

test/built-ins/Array/prototype/toLocaleString/resizable-buffer.js
```diff
@@ -0,0 +1,98 @@
...
+    let result = '';
+    if (len > 1) {
+      for (let i=0; i < len - 1 ; i++) {
+        result += list[i] + comma;
+      }
+    }
+    if (len > 0) {
+      result += list[len-1];
+    }
+    return result;
```
ptomato, line 34: There's an array method for this :smile: 
```suggestion
    return list.join(comma);
```

ptomato (approved):
LGTM after shortening `listToString`

test/built-ins/Array/prototype/toLocaleString/resizable-buffer.js
ptomato: I was unsure about whether to still pass `en-u-nu-latn` to each invocation of toLocaleString like I suggested above. In the end I think it's OK not to; it would clutter up the tests quite a bit and I am sure there are a bunch of other existing tests that would fail if the default locale wasn't `en`.

PR #4181, opened 2024-07-30
RAB: fix reduce tests

lahma:
Introduced in #4156 , `Array.prototype.reduce` had the correct signature for callback function, but typed array did not.

`reduce` callback has call signature `«accumulator, kValue, 𝔽(k), O»` earlier the collected values contained the accumalator instead of the actual value.


ioannad (approved):
LGTM, thanks for catching this! 

Just mentioning issue https://github.com/tc39/test262/issues/4118 - a table like this perhaps could have helped caught these errors earlier.

ptomato (approved):
Maintainer stamp

Issue #4161, opened 2024-07-22
Add metadata about the ECMAScript edition that each feature was introduced in


jedel1043:
It's pretty common for testers such as [test262.fyi](https://test262.fyi) to have a way to filter tests per ECMAScript edition, and some people like to analyze the conformance of engines per ES version.

Currently some runners have opted for using "features.txt" to extract this information, since features have a clear edition where they were introduced, and this could be used to know which minimum edition a test supports. However, there's some duplicate effort around this, since all runners have to maintain a separate mapping from the feature to the corresponding edition that introduced it:

- [boa_tester](https://github.com/boa-dev/boa/blob/816e1bb6481a76494b195bdd831e79162873c9b2/tests/tester/src/edition.rs)
- [test262.fyi](https://github.com/test262-fyi/test262.fyi/blob/81861b7aeeb518326147ab2826d4adc184f14686/site/generate.mjs#L242-L458)

Seeing this, it would be nice to remove this duplicate effort by moving this metadata to here.

### Possible points of discussion

- [ ] Unified file (features list + metadata) vs two files (features file + metadata file).
- [ ] File representation
  - Keep using .txt
  - Transition to yaml
  - Some other format?
- [ ] Edition representation
  - Numeric (ES6 -> 6, ES12 -> 12, ...)
  - String (ES6 -> "ES6", ES12 -> "ES12", ...)
  - Other
- [ ] Sentinel value for features that haven't been published
  - Big number (255/99/65535)
  - Negative number (-1)
  - Some string? ("N/A", "ESNext", "undefined", "None", ...)

ljharb:
I think a unified file would be nicer, but that might interfere with how existing runners interface with the existing one.

(also, we should be using years, post-ES6, not edition numbers)

jedel1043:
> I think a unified file would be nicer, but that might interfere with how existing runners interface with the existing one.

This is possible to do using comments. The file already has some comments, so we could probably leverage that to add metadata in a backwards-compatible way.

jedel1043:
> (also, we should be using years, post-ES6, not edition numbers)

That's fair. I'm just using the edition identifier used by the two example runners, but it would also be understandable to use the more "official" identifier instead.

PawelJastrzebski:
Added concept of `toml` file in the PR.

ljharb:
I'm not sure toml is a good choice solely because it's not natively parseable in node and other platforms (altho i love toml!)

[...]

PR #4182, opened 2024-07-30
ECMAScript features edition - Toml file proposal

PawelJastrzebski:
Issue: https://github.com/tc39/test262/issues/4161

TODO
- edition year was not verified

DONE 
- all features present in current order with urls (if present)
- preserved existing comments
- valid toml format

### DIFF
--- a/features.toml
+++ b/features.toml
@@ -1,0 +1,783 @@
+## Proposed language features
+#
+# This project accepts tests for language proposals that have reached stage 3
+# in TC39's standardization process. Those tests should be annotated with a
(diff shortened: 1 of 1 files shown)


ptomato:
I made this a draft because it's still under discussion in #4161. What should the next steps be?

PawelJastrzebski:
Someone would have to decide in what form the metadata would be stored. All the files are just for preview.

ptomato:
Reading the discussion in the issue, it seems like JSON is probably the only feasible format.

PR #4183, opened 2024-07-31
Run full test suite for non-PR builds

lahma:
I'm proposing to run full suite of tests when running against `main` branch. This will give full listing of currently unsupported tests by engine. Also updating to latest NPM dependencies.

Test will still be green on CI like for PRs when errors are present, see https://github.com/tc39/test262/commit/1ebd34b53b59a5cb84bdfece8931e40c462747e0 for original reasoning not erroring.

I would argue that it would be better if the PRs would fail and maintainers would need to read the actual error output to figure out whether problems are correct. The test262-harness has a flag for this `--error-for-failures` which would break PR builds if there's a problem. Of course this would create noise for unsupported features per engine.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,8 +13,8 @@
   "private": true,
   "homepage": "https://github.com/tc39/test262#readme",
   "devDependencies": {
-    "esvu": "^1.2.11",
-    "test262-harness": "^8.0.0"
+    "esvu": "^1.2.16",
+    "test262-harness": "^10.0.0"
   },
   "scripts": {
     "ci": "./tools/scripts/ci_test.sh",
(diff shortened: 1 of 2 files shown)


ptomato:
See also #4118. I think that would be better than having the PRs fail, because most of the PRs submitted to Test262 are for not-yet-supported features.

I wouldn't be opposed to running the whole suite on `main`, but two things:

1. How is it different from test262.fyi? The website seems like it puts that information in a form that's easier to consume than clicking through to CircleCI and scrolling through the log.
2. I'd like to avoid adding more CircleCI stuff, as CircleCI seems like it's been less reliable than GitHub Actions. How about having this as a separate GitHub action, if we decide to have it?

lahma:
I was not aware of https://test262.fyi/ , thank you for pointing that out. After considering this again, it's such two-edged sword and problem. The linked idea of commented result table sounds like a better approach and would bring possible problems to the front an center. I'm closing this in light of this new information.

PR #4185, opened 2024-08-01
Add tests for nullable quantifiers in RegExps

Aurele-Barriere:
The JavaScript regex quantifiers (star, plus, counted repetition) have unique semantics when it comes to matching the empty string.
Optional iterations of quantifiers that match the empty string are forbidden.
This is documented in [ECMAScript](https://262.ecma-international.org/15.0/index.html#sec-runtime-semantics-repeatmatcher-abstract-operation) (point 2.b)
and the PLDI2024 paper [Linear Matching of JavaScript Regular Expressions](https://dl.acm.org/doi/10.1145/3656431).

In many other regex languages however, the quantifier semantics are different.
Instead of only checking at the end of an iteration, it is forbidden to visit any part of the regex twice without having consumed a character.

In some rare cases, the two semantics give different results.
For instance, matching /(a?b??)*/ on string "ab" matches "ab" in JavaScript (two iterations of the star, each matching one character), but only "a" in other languages.
See for instance https://regex101.com/r/iuVSat/1

I don't believe this JavaScript-specific behavior has been documented in Test262. I suggest adding a test case for this.
Many classical algorithms for regex matching (like NFA simulation or lazy DFA) typically do not implement the JavaScript quantifier semantics.
This led to a bug in the linear engine of V8 (which uses NFA simulation):
[Bug Report](https://issues.chromium.org/issues/42204037), [Relevant test case in V8](https://chromium-review.googlesource.com/c/v8/v8/+/4661659), [V8 fix](https://chromium-review.googlesource.com/c/v8/v8/+/4755530), [Explanation of the fix](https://dl.acm.org/doi/10.1145/3656431) (section 4.1).

### DIFF
--- a/test/built-ins/RegExp/nullable-quantifier.js
+++ b/test/built-ins/RegExp/nullable-quantifier.js
@@ -1,0 +1,21 @@
+// Copyright (C) 2024 Aurèle Barrière. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-runtime-semantics-repeatmatcher-abstract-operation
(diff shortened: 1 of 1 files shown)


ptomato (approved):
Thanks for the test and for the comprehensive explanation.

PR #4194, opened 2024-08-13
Remove source phase resolution time negative error types

legendecas:
The source phase import binding name tests can be tested more easily with the resolution time negative error types removed, as they are host defined errors.

Refs: https://github.com/tc39/test262/pull/4121#issuecomment-2286041289


test/language/module-code/source-phase-import/import-source-binding-name.js
```diff
@@ -8,16 +8,9 @@ info: |
...
 flags: [module]
 ---*/
 
-$DONOTEVALUATE();
-
-import "../resources/ensure-linking-error_FIXTURE.js";
-
-import source from '<do not resolve>';
-import from from '<do not resolve>';
+import source from '<module source>';
```
woess, line 15: I think we should avoid using `<module source>` in non-source imports.

nicolo-ribaudo:
Instead of doing this, can we just remove `type: SyntaxError`? We just need to assert that we are getting an error at resolution time, regardless or where it comes from, since we are just checking that it does not throw when parsing.

legendecas:
> Instead of doing this, can we just remove type: SyntaxError?

Updated. Thanks for suggestion, sounds like a good alternative to me.

nicolo-ribaudo (approved):
Looks good, if `Error` is something valid to write there to mean "any error"

Ms2ger:
> Looks good, if `Error` is something valid to write there to mean "any error"

Hm, I'm not sure that it is.
> type - If an error is thrown, it is implicitly converted to a string. In order for the test to pass, this value must match the name of the error constructor.

nicolo-ribaudo:
Can we just omit `type`?

legendecas:
> Can we just omit `type`?

No, the linter complains about missing `type`:

https://github.com/tc39/test262/blob/75fddbd694e76ea53410560cf4138ffff900f923/tools/lint/lib/checks/negative.py#L26

nicolo-ribaudo:
Then what about completely moving those tests to `_FIXTURE` files, and add an entrypoit that does
```js
import("./the-actual-entrypoint_FIXTURE.js").then(() => $DONE(new Error("It should have failed")), () => $DONE());
```

legendecas:
Updated and verified with d8.

test/language/module-code/source-phase-import/import-source.js
```diff
@@ -0,0 +1,34 @@
...
+function assertImportSourceResolutionFailure(specifier) {
+  // Import the module and assert that the promise is rejected with a host
+  // defined error during the resolution phase.
+  // Note that this is not a `import.source`.
+  return import(specifier).then(
+    () => {
+      throw new Test262Error(`${specifier}: Promise should be rejected`);
+    },
+    error => {
+      print(error)
```
nicolo-ribaudo, line 22: This should be removed

test/language/module-code/source-phase-import/import-source-binding-name_FIXTURE.js
```diff
@@ -0,0 +1,20 @@
...
+// this file does not raise SyntaxError. Note that a SyntaxError could also
+// be raised when the imported module does not have a source phase
+// representation (see sec-source-text-module-record-initialize-environment, 7.c.ii).
+//
+// esid: sec-modules
+// info:
+//   ImportDeclaration:
+//     import source ImportedBinding FromClause ;
+
+$DONOTEVALUATE();
```
nicolo-ribaudo, line 15: We should remove this. If it ever gets called, it'll throw an error, trigger the `.catch` block in `import-source.js` with an error that's not a SyntaxError, and make the test pass.

nicolo-ribaudo (approved):

Ms2ger (approved):

PR #2957, opened 2021-02-28
Test TypedArray's [[Set]] internal method with altered receiver

shvaikalesh:
⚠️ **don't merge yet** ⚠️

Ref. https://github.com/tc39/ecma262/pull/1556

PR #4200, opened 2024-08-16
Revert "Revert "Test TypedArray's [[Set]] internal method with altered receiver""

Ms2ger:
Reverts tc39/test262#2975

In #2957, @rwaldron requested a new PR to be made with these tests; that never seems to have happened.


ptomato:
Very impressive archeology :smile:

Ms2ger (commented):
I can't approve, but I reviewed these pretty carefully and I think they're correct

ptomato (approved):
I'll rubberstamp it then.

PR #4205, opened 2024-08-21
fix: Use Object.getPrototypeOf to get prototype in `__proto__-fn-name.js`

aapoalas:
As best as I can tell, this test is accidentally and very confusingly relying on the Annex B `__proto__` getter when testing the spec-defined `__proto__` key literal special case.

As an example: Node.js passes this test as it is currently, but Deno does not as Deno does not implement the `__proto__` getters and setters.


test/language/expressions/object/__proto__-fn-name.js
```diff
@@ -23,4 +23,4 @@ o = {
   __proto__: function() {}
 };
 
-assert(o.__proto__.name !== '__proto__');
+assert(Object.getPrototypeOf(o).name !== '__proto__');
```
ljharb, line 26: this test file is explicitly about `__proto__`; if deno doesn't implement them then deno shouldn't be running this test.

ljharb (changes requested):

test/language/expressions/object/__proto__-fn-name.js
aapoalas: The test is not about the `__proto__` getter, it is checking that the anonymous function in `__proto__: function () {}` does pass through `NamedEvaluation`.

ljharb: the filename is "`__proto__-fn-name.js`", maybe i'm confused. do you mean that this specific line doesn't care about dunder proto?

aapoalas: Though ... it seems I'm wrong after all. I'm a bit confused: The `es6id` is `B.3.1` so that's obviously Annex B but the description doesn't mention it.

I'm sorry, you're right: This is probably just my misunderstanding.

aapoalas: Ooor maybe it isn't again :D The `B.3.1` points to the ES6 specification's definition of the [`__proto__` property names in object initializers](https://262.ecma-international.org/6.0/#sec-__proto__-property-names-in-object-initializers). This is now appears [here](https://tc39.es/ecma262/#sec-runtime-semantics-propertydefinitionevaluation) in the definition of PropertyDefinitionEvaluation. The definition has changed somewhat but the effect has not:

> If IsAnonymousFunctionDefinition(AssignmentExpression) is true and isProtoSetter is false, then
>     a. Let propValue be ? NamedEvaluation of AssignmentExpression with argument propKey.

So this is still living ECMAScript spec, now outside of Annex B and hence shouldn't really be dependent on the `__proto__` getter as far as I can tell.

aapoalas: > the filename is "`__proto__-fn-name.js`", maybe i'm confused. do you mean that this specific line doesn't care about dunder proto?

Yeah so the filename refers not to the `__proto__` getter and setter properties that are still defined in Annex B, but to the special casing of the `__proto__` property name in object literals. It's effectively testing this difference:

```js
const a = {
  foo: function() {},
  __proto__: function() {},
};

assert(a.foo.name === "foo"); // Anonymous function assignment to property name in object literal becomes a named function.
// But for the __proto__ property name it doesn't become a named function, as this is not a property assignment but a prototype setting!
// So let's test the same but change the expectation:
// assert(a.__proto__.name !== "foo");
// Oops! Now we're accessing the prototype of `a` using the `__proto__` getter, which may or may not be supported. Whether or not it is is entirely immaterial on the name of the function that we set as the prototype of `a`.
// So, get a's prototype the "proper" way:
assert(Object.getPrototypeOf(a).name !== "foo");

```

ljharb: I think you're right.

There are engines, however, that have `__proto__` and not `Object.getPrototypeOf`. perhaps this?
```suggestion
assert((Object.getPrototypeOf ? Object.getPrototypeOf(o) : o.__proto__).name !== '__proto__');
```

aapoalas: That is fine by me, but it would be the very first time such an invocation appears in the tests. Currently `Object.getPrototypeOf(` appears 1601 times in the repository, and `.__proto__` appears 53 times. If we filter out the explicit tests for `Object.prototype.__proto__` getter-setter property, `.__proto__` appears 29 times. Finally only 17 hits come if we filter out files that explicitly test `__proto__` (which are testing the object literal property key). (Filtering out explicit `Object.getPrototypeOf` tests still leaves 1500 matches.)

Based on this it seems like checking the prototype using the `__proto__` getter seems to be vanishingly rare: Looking at the test files in the above 17 matches seems to contain 3 or 4 files where the getter is actually being used, and those are TypedArray tests so relatively modern ECMAScript.

As such: Do you believe prototype checking should be done in this sort of conditional way to protect engines that do not implement `Object.getPrototypeOf` in this test's case? If so, should all prototype access be done in such a way?

ljharb: ideally yes, but you're right that it should be done in a holistic way or not at all.

ljharb (approved):

test/language/expressions/object/__proto__-fn-name.js
ptomato: I'm happy to have it done on a best-effort basis, FWIW.

aapoalas: I was thinking the same: I guess the test harness would need to implement a function along these lines:
```js
function getPrototypeOf(target) {
  if (Object !== null && typeof Object === "object" && typeof Object.getPrototypeOf === "function") {
    return Object.getPrototypeOf(target);
  } else if (Object !== null && typeof Object === "object" && Object.prototype !== null && typeof Object.prototype === "object" && "__proto__" in Object.prototype) {
    return target.__proto__;
  } else if (typeof $262.getPrototypeOfHook === "function") {
    return $262.getPrototypeOfHook(target);
  } else {
    throw new Error("getPrototypeOf not supported");
  }
}
```
The `getPrototypeOfHook` there is... I don't know who would really benefit from it but _maybe_ some engine might offer a limited test-time-only way to get the prototype of an object but specifically _wouldn't_ want to implement `Object.getPrototypeOf` for some reason.

Ms2ger:
Please replace the es6id by the relevant esid

aapoalas:
> Please replace the es6id by the relevant esid

According to some documentation I found it's a deprecated: Should I update it or just remove it?

Ms2ger:
Replace by the key `esid`

Ms2ger:
And yes, if the quoted spec text changed, please fix that too.

aapoalas:
> Replace by the key `esid`
> And yes, if the quoted spec text changed, please fix that too.

Oops, sorry this slipped my mind somehow. I've now fixed the esid to point to PropertyDefinitionEvaluation as it should, and copied over the relevant steps from the spec text.

Ms2ger (approved):
Thanks!

PR #4206, opened 2024-08-21
Temporal: Further coverage and tweaks for removing Calendar/TimeZone objs

ptomato:
Tweak some tests to provide coverage of new execution paths in the spec, such as calling GetOptionsObject inside ToTemporal___; add a few new tests for things that weren't covered before, such as rounding a PlainDateTime at the edge of the range; and tweak the tests verifying when the properties of the options bag are read, which I made a mistake in #4119.

See: https://github.com/tc39/proposal-temporal/pull/2925


Ms2ger (approved):

ptomato:
Thanks for the review!

PR #4209, opened 2024-08-27
Add more tests for receiver check in IntegerIndexedElementSet

anba:
Additional tests to cover some spec-compliance issues I've encountered while implementing tc39/ecma262#1556 in SpiderMonkey.

### DIFF
--- a/test/built-ins/TypedArray/prototype/set/array-arg-value-conversion-resizes-array-buffer.js
+++ b/test/built-ins/TypedArray/prototype/set/array-arg-value-conversion-resizes-array-buffer.js
@@ -1,0 +1,66 @@
+// Copyright (C) 2024 André Bargull. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-settypedarrayfromarraylike
+description: >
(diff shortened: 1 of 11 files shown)


Ms2ger:
Did you mean to push all three commits to this PR? I'm happy to approve the first one, at least, but would need to spend some more time reviewing the others.

anba:
Yes, that was intentional, because at least for SpiderMonkey there are [shared code paths](https://searchfox.org/mozilla-central/rev/490a1df802d8872f996f8ef4093d54e3c854c8f9/js/src/vm/NativeObject.h#1838-1850) for normal property operations and environment operations, so when I've updated the property operations for out-of-bounds typed array indices, I also had to worry about environment operations. Similar story about the new test for `TypedArray.prototype.set`: I've also updated the [TypedArraySetElement](https://tc39.es/ecma262/#sec-typedarraysetelement) implementation in SM, which led to checking which spec operations actually call `TypedArraySetElement` and writing some small tests to check implementations agree on the spec semantics. This uncovered a spec violation in JSC (72933dfc0be2394956cf29689336dd7678d089c0).

Issue #3665, opened 2022-09-14
Let's have exhaustive feature lists!


ljharb:
Opening this per https://github.com/tc39/test262/pull/3353#issuecomment-1247158634.

I think that in order to work towards userland packages being able to run test262 themselves, we need to do a better job ensuring that feature lists are exhaustive, and that unnecessary features aren't used (arrow functions when normal ones suffice, for example), without putting a burden on contributors.

I don't think a precommit script is a good idea specifically - either we need a way for *maintainers* to easily clean things up later, or, we need a CI check/formatter that helps contributors do the right thing without having to type it all out.

ptomato:
Personally I'd like to see this be treated as opt-in; I'm thinking along the lines of contributors should not have to pay attention to features that have been stage 4 for a while, so we shouldn't do this in CI, nor necessarily use a script to always automatically transform all arrow functions into normal functions, for example. But we should provide scripts to make it easy for userland packages to adjust whatever features they need on whatever subset of tests they need to do it on, and generally accept such PRs from userland packages without complaint.

In other words, a best-effort supported use case where that we facilitate where possible while not requiring additional effort from [constituents higher on our list](https://github.com/tc39/test262/issues/3433).

ljharb:
I'm totally fine with avoiding burden!

I'm not sure how practical it is tho for consumers to run scripts to mutate test262 on disk in order to run it.

ptomato:
I agree, I'm saying we should be willing to commit changes like this without necessarily making it a priority to make them preemptively.

ljharb:
I agree it doesn’t need to block merging, especially if we have a manual tool to detect and/or autofix them later.

anba:
Past discussions which are probably relevant here: #3196, #2862, #472 (links to [this meeting note](https://github.com/tc39/notes/blob/main/meetings/2015-09/sept-24.md#8-test262-update)), #569. I think there are more, for example in review comments when the reviewer encouraged to use `const` instead of `var` or other other way around, i.e. to use `var` instead of `const` to target older engines.

ptomato:
Related to this, it also occurred to me that for features that long ago reached Stage 4, we could organize the features list according to what platforms are likely to want to run the tests. For example, we could remove all the individual typed array feature flags (`Uint8Array`, `Int8Array`, ...), and just use `TypedArray` as a catch-all feature flag, if there aren't any platforms where we want to run the tests that only implement a subset of the typed arrays.

ljharb:
[...]

PR #4216, opened 2024-09-02
Improve coverage for reference type in various contexts

anba:
Missing coverage encountered while implementing
<https://github.com/tc39/ecma262/pull/3307> in SpiderMonkey.

Ensure environment lookups are performed in the correct order:
- keyed-destructuring-property-reference-target-evaluation-order-with-bindings.js

Ensure `delete super[elem]` steps are correctly performed:
- delete/super-property-topropertykey.js
- delete/super-property-uninitialized-this.js

Ensure ToPropertyKey for computed property names in object literals correctly performed:
- object/computed-property-name-topropertykey-before-value-evaluation.js

Ensure `GetSuperBase` is executed before `ToPropertKey`:
- super/prop-expr-getsuperbase-before-topropertykey-*

Ensure `GetThisBinding` is executed first:
- super/prop-expr-uninitialized-this-*

### DIFF
--- a/test/language/destructuring/binding/keyed-destructuring-property-reference-target-evaluation-order-with-bindings.js
+++ b/test/language/destructuring/binding/keyed-destructuring-property-reference-target-evaluation-order-with-bindings.js
@@ -1,0 +1,88 @@
+// Copyright (C) 2024 André Bargull. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-destructuring-binding-patterns-runtime-semantics-propertybindinginitialization
+description: >
+  Ensure correct evaluation order for binding lookups when destructuring target is var-binding.
+info: |
+  14.3.3.1 Runtime Semantics: PropertyBindingInitialization
+
+    BindingProperty : PropertyName : BindingElement
+
+    1. Let P be ? Evaluation of PropertyName.
+    2. Perform ? KeyedBindingInitialization of BindingElement with arguments value, environment, and P.
+    ...
+
+  14.3.3.3 Runtime Semantics: KeyedBindingInitialization
+
+    SingleNameBinding : BindingIdentifier Initializer_opt
+
+    1. Let bindingId be the StringValue of BindingIdentifier.
+    2. Let lhs be ? ResolveBinding(bindingId, environment).
+    3. Let v be ? GetV(value, propertyName).
+    4. If Initializer is present and v is undefined, then
+      ...
+      b. Else,
+        i. Let defaultValue be ? Evaluation of Initializer.
+        ii. Set v to ? GetValue(defaultValue).
+    ...
+    6. Return ? InitializeReferencedBinding(lhs, v).
+
+  9.4.2 ResolveBinding ( name [ , env ] )
+
+    ...
+    4. Return ? GetIdentifierReference(env, name, strict).
+
+  9.1.2.1 GetIdentifierReference ( env, name, strict )
+
+    ...
+    2. Let exists be ? env.HasBinding(name).
+    ...
+
+includes: [compareArray.js]
+features: [Proxy]
+flags: [noStrict]
+---*/
+
+var log = [];
+
+var sourceKey = {
+  toString: () => {
+    log.push("sourceKey");
+    return "p";
+  }
+};
+
+var source = {
+  get p() {
+    log.push("get source");
+    return undefined;
+  }
+};
+
+var env = new Proxy({}, {
+  has(t, pk) {
+    log.push("binding::" + pk);
+    return false;
+  }
+});
+
+var defaultValue = 0;
+
+var varTarget;
+
+with (env) {
+  var {
+    [sourceKey]: varTarget = defaultValue
+  } = source;
+}
+
+assert.compareArray(log, [
+  "binding::source",
+  "binding::sourceKey",
+  "sourceKey",
+  "binding::varTarget",
+  "get source",
+  "binding::defaultValue",
+]);
--- a/test/language/expressions/assignment/destructuring/keyed-destructuring-property-reference-target-evaluation-order-with-bindings.js
+++ b/test/language/expressions/assignment/destructuring/keyed-destructuring-property-reference-target-evaluation-order-with-bindings.js
@@ -1,0 +1,90 @@
+// Copyright (C) 2024 André Bargull. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-runtime-semantics-propertydestructuringassignmentevaluation
+description: >
+  Ensure correct evaluation order for binding lookups when destructuring target is var-binding.
+info: |
+  13.15.5.3 Runtime Semantics: PropertyDestructuringAssignmentEvaluation
+
+    AssignmentProperty : PropertyName : AssignmentElement
+
+    1. Let name be ? Evaluation of PropertyName.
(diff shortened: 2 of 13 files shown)


test/language/destructuring/binding/keyed-destructuring-property-reference-target-evaluation-order-with-bindings.js
```diff
@@ -0,0 +1,100 @@
...
+    4. Return ? GetIdentifierReference(env, name, strict).
+
+  9.1.2.1 GetIdentifierReference ( env, name, strict )
+
+    ...
+    2. Let exists be ? env.HasBinding(name).
+    ...
+
+includes: [compareArray.js]
+features: [Proxy]
```
linusg, line 44:
```suggestion
features: [Proxy, destructuring-assignment]
```

(line 88)

test/language/expressions/assignment/destructuring/keyed-destructuring-property-reference-target-evaluation-order-with-bindings.js
```diff
@@ -0,0 +1,90 @@
...
+    3. If Initializer is present and v is undefined, then
+      ...
+      b. Else,
+        i. Let defaultValue be ? Evaluation of Initializer.
+        ii. Let rhsValue be ? GetValue(defaultValue).
+    ...
+    6. Return ? PutValue(lRef, rhsValue).
+
+includes: [compareArray.js]
+features: [Proxy]
```
linusg, line 33:
```suggestion
features: [Proxy, destructuring-assignment]
```

test/language/expressions/super/prop-expr-uninitialized-this-getvalue.js
```diff
@@ -0,0 +1,31 @@
...
+info: |
+  13.3.7.1 Runtime Semantics: Evaluation
+
+    SuperProperty : super [ Expression ]
+
+    ...
+    2. Let actualThis be ? env.GetThisBinding().
+    3. Let propertyNameReference be ? Evaluation of Expression.
+    ...
+---*/
```
linusg, line 17:
```suggestion
features: [class]
---*/
```

test/language/expressions/super/prop-expr-uninitialized-this-putvalue-increment.js
```diff
@@ -0,0 +1,31 @@
...
+info: |
+  13.3.7.1 Runtime Semantics: Evaluation
+
+    SuperProperty : super [ Expression ]
+
+    ...
+    2. Let actualThis be ? env.GetThisBinding().
+    3. Let propertyNameReference be ? Evaluation of Expression.
+    ...
+---*/
```
linusg, line 17:
```suggestion
features: [class]
---*/
```

test/language/expressions/super/prop-expr-uninitialized-this-putvalue.js
```diff
@@ -0,0 +1,31 @@
...
+info: |
+  13.3.7.1 Runtime Semantics: Evaluation
+
+    SuperProperty : super [ Expression ]
+
+    ...
+    2. Let actualThis be ? env.GetThisBinding().
+    3. Let propertyNameReference be ? Evaluation of Expression.
+    ...
+---*/
```
linusg, line 17:
```suggestion
features: [class]
---*/
```

anba:
@linusg Do you need these flags because your implementation doesn't support these features? I'm asking because other files in the same directories are already omitting some flags. Unless it's necessary for some actual implementation, I'd prefer to omit flags for language features which were added years ago.

linusg:
Personally I've never used the flags, IMO evergreen engines don't need them. A failure is a failure and filtering doesn't change that.

The comments are just for completeness, feel free to disregard them if the maintainers agree that we don't need to be strict about flags 👍

ljharb:
We in fact do want every test to have flags for every feature, even ones added many years ago.

anba:
No, that's not what's documented in [features.txt](https://github.com/tc39/test262/blob/main/features.txt).

ljharb:
ok, but that's the current policy of the maintainers (and the existence of `class` and other <= ES2015 features should make it clear that the sentence at the top is nonexhaustive). I'll add to my list to update the wording in that document.

anba:
Is this documented anywhere? For example <https://github.com/tc39/test262/issues/3196#issuecomment-1587063544> and <https://github.com/tc39/test262/issues/3665#issuecomment-1247214982> don't state that it's a strict requirement to name each and every feature.

ljharb:
It's indeed not a strict requirement - but we still _want_ everything to have them.

ptomato:
I should nuance my position on the policy - All other things being equal I'd prefer to have exhaustive flags, but I also don't think we should be putting that burden on people who contribute tests, so I'd much rather pursue that via codemods.

Ms2ger (approved):
These are nasty. Thanks.

test/language/expressions/super/prop-expr-getsuperbase-before-topropertykey-putvalue-compound-assign.js
```diff
@@ -0,0 +1,60 @@
...
+var obj = {
+  __proto__: proto,
+  m() {
+    return super[key] += 1;
+  }
+};
+
+var key = {
+  toString() {
+    Object.setPrototypeOf(obj, proto2);
```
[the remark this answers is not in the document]
anba: I think I'd prefer to not assert this, because otherwise V8 will have to skip this test, because V8 evaluates `obj[key] += value` as `obj[key] = obj[key] + value`, which evaluates `ToPropertyKey(key)` twice. And if we don't assert how many times `ToPropertyKey(key)` is evaluated, V8 is more likely to not skip this test, so there's maybe more incentive for V8 to fix their implementation to evaluate `GetSuperBase` before `ToPropertyKey`.

test/language/destructuring/binding/keyed-destructuring-property-reference-target-evaluation-order-with-bindings.js
```diff
@@ -0,0 +1,100 @@
...
+---*/
+
+var log = [];
+
+var targetKey = {
+  toString: () => {
+    log.push("targetKey");
+    return "q";
+  }
+};
```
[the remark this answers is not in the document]
anba: Fixed. Thanks for spotting this!

test/language/destructuring/binding/keyed-destructuring-property-reference-target-evaluation-order-with-bindings.js
```diff
@@ -0,0 +1,100 @@
...
+
+var target = {
+  set q(v) {
+    log.push("set target:");
+  },
+};
+
+var env = new Proxy({}, {
+  has(t, pk) {
+    log.push("binding::" + pk);
```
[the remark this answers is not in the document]
anba: Fixed.

test/language/destructuring/binding/keyed-destructuring-property-reference-target-evaluation-order-with-bindings.js
```diff
@@ -0,0 +1,100 @@
...
+    log.push("get source");
+    return undefined;
+  }
+};
+
+var target = {
+  set q(v) {
+    log.push("set target:");
+  },
+};
```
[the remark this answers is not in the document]
anba: Fixed.

PR #4221, opened 2024-09-10
Add test coverage for correct binding object MOP executions in with-environment

anba:
Adds coverage for <https://bugzilla.mozilla.org/show_bug.cgi?id=1166408>.


ptomato (approved):
Nice, thanks for covering this gap.

PR #4225, opened 2024-09-12
Add Unicode 16 identifier tests

mathiasbynens:
Issue: https://github.com/tc39/ecma262/issues/3425


Ms2ger (approved):
rubber stamp for whenever you're ready

mathiasbynens:
This is ready to go. The esmeta check is failing but I believe this is expected since no engine implements Unicode 16 yet.

PR #3866, opened 2023-06-28
Add tests for the Explicit Resource Management proposal

rbuckton:
This adds tests for the [Explicit Resource Management proposal](https://github.com/tc39/proposal-explicit-resource-management) based on the spec text at https://tc39.es/proposal-explicit-resource-management as well as several pending pull requests in the proposal repo.

PR #4230, opened 2024-09-18
Function branding boilerplate tests from PR #3866

ioannad:
This PR splits some of the boilerplate tests from PR #3866  to make reviewing all those files a little easier.

This includes branding boilerplate tests for methods, as for example here: https://github.com/tc39/test262/blob/main/test/intl402/ListFormat/prototype/format/branding.js


ptomato (approved):
Thanks for splitting these out! Makes it easier to review in chunks :smile:

PR #4232, opened 2024-09-20
Temporal: Add test coverage for canonicalizing era codes in from()

ptomato:
There was a coverage gap for converting calendar fields to a date when the provided era was an alias.

See https://github.com/tc39/proposal-temporal/pull/2940#discussion_r1767508659


gibson042 (approved):
Thanks, LGTM!

PR #4233, opened 2024-09-20
Remove test for old behavior for Atomics.pause

syg:
Syncs up with current spec text, which allows negative integers for the parameter


ptomato (approved):
Thanks! :+1:

Issue #3920, opened 2023-09-12
Dependency of async-gc helper on cleanupSome functionality


ptomato:
The [async-gc helper](https://github.com/tc39/test262/blob/main/harness/async-gc.js) uses `FinalizationRegistry.prototype.cleanupSome`. Probably at the time the helper was written (2019), this feature was part of a stage 3 proposal, but it was split off, demoted to stage 2 in the meantime, and stalled. It's probably not desirable to have part of the test harness depend on a stage 2 proposal which may only slowly (if ever) advance to stage 3.

Outside of the tests for `FinalizationRegistry.prototype.cleanupSome` itself, the helper is used in 3 tests:
- https://github.com/tc39/test262/blob/main/test/built-ins/FinalizationRegistry/gc-has-one-chance-to-call-cleanupCallback-for-object.js
- https://github.com/tc39/test262/blob/main/test/built-ins/FinalizationRegistry/prototype/unregister/unregister-cleaned-up-object-cell.js
- https://github.com/tc39/test262/blob/main/test/built-ins/WeakRef/prototype/deref/gc-cleanup-not-prevented-with-wr-deref.js

It's not clear to me whether any of the tests that use this harness helper are actually regularly executed. [test262.fyi results](https://test262.fyi/#built-ins/FinalizationRegistry/prototype/cleanupSome) show only passing results for SpiderMonkey, Serenity LibJS, and GraalVM. It could be that any test using this helper is in practice a coverage gap for most implementations due to being skipped, or marked as expected-fail.

I don't know about the official implementation status of cleanupSome, but as far as I can tell from looking for its presence in a recently updated esvu installation:
- engine262: feature flag exists but does not define method
- JavaScriptCore: no
- QuickJS: no
- Serenity LibJS: yes
- SpiderMonkey: yes
- V8: behind flag
- XS: yes

The resolution of this issue should be to create a PR removing `cleanupSome` from the test corpus, since it is stage 2, and seek feedback from implementors on whether they would miss it. Also, investigate whether the functionality being tested in those remaining 3 tests can also be tested without `cleanupSome`, and whether any changes to the interpreting requirements would be necessary to do so, and include such changes in the PR if possible.

ptomato:
The `cleanupSome` proposal has been withdrawn as of the 2023-09 TC39 meeting.

ptomato:
I have some news on this. XS has removed `cleanupSome` since I wrote the OP.

[...]

PR #4239, opened 2024-09-25
Remove cleanupSome

ptomato:
The cleanupSome proposal has been withdrawn. This PR removes the tests for it.

It also removes the async-gc helper which was implemented using cleanupSome and the three WeakRef and FinalizationRegistry test files which used it. In these tests, cleanupSome() would have optionally emptied the FinalizationRegistry synchronously and called the finalization callbacks. Without cleanupSome(), there is no way to observe a FinalizationRegistry being emptied using "pure" ECMAScript: FinalizationRegistry doesn't get cleaned up until the microtask queue is empty, and without external APIs such as setTimeout() there is no way to pause execution until after the microtask queue is empty.

Non-deterministic tests are not that useful anyway; these tests are probably better left to implementations' internal testing strategies.

Closes: #3920


ljharb (approved):

gibson042 (approved):

ptomato:
Thanks for the quick reviews!

PR #4240, opened 2024-09-25
Add names for expected outcomes in compareArray.js

lahma:
Just a small quality of life improvement. I usually can't wrap my head around which was the expected and which one was the actual outcome so a textual clue helps.


ljharb (approved):

ptomato (approved):
Excellent. This trips me up so often too.

PR #4242, opened 2024-09-27
Fix compareArray.js harness tests

linusg:
It looks like #4240 was not tested, cc @lahma


Ms2ger (approved):

lahma:
Apologies, I was running only the actual test suite, not the the tests for the test suite itself, live and learn 😞

PR #4251, opened 2024-10-09
Avoid using destructuring in tco-fn-realm.js

### LINKED ISSUE #4218: `test/built-ins/Proxy/revocable/tco-fn-realm.js` is missing `destructuring-assignment` feature flag
https://github.com/tc39/test262/blob/fc49c9ce163dcf501bb0457f0521900cd744cbda/test/built-ins/Proxy/revocable/tco-fn-realm.js#L9

https://github.com/tc39/test262/blob/fc49c9ce163dcf501bb0457f0521900cd744cbda/test/built-ins/Proxy/revocable/tco-fn-realm.js#L15

tannal:
Fixes #4218

### DIFF
--- a/test/built-ins/Array/prototype/toString/non-callable-join-string-tag.js
+++ b/test/built-ins/Array/prototype/toString/non-callable-join-string-tag.js
@@ -27,13 +27,15 @@ assert.sameValue(Array.prototype.toString.call({ join: {} }), "[object Object]")
 
 let revokeOnGet = false;
 const proxyTarget = [];
-const { proxy, revoke } = Proxy.revocable(proxyTarget, {
+var proxyObj = Proxy.revocable(proxyTarget, {
     get: (target, key, receiver) => {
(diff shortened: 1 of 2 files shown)


anba:
There's another file where this pattern shows up: https://github.com/tc39/test262/blob/main/test/built-ins/Array/prototype/toString/non-callable-join-string-tag.js

Using the RegExp `\}\s*=\s*Proxy\.revocable\(` to search for that pattern: https://searchfox.org/mozilla-central/search?q=%5C%7D%5Cs*%3D%5Cs*Proxy%5C.revocable%5C%28&path=js%2Fsrc%2Ftests%2Ftest262%2F&case=false&regexp=true

Ms2ger (approved):
I don't object to this change, but I wonder what it is about this particular test that makes you want to avoid destructuring in it

ptomato (approved):
:+1:

PR #4253, opened 2024-10-09
harness/deepEqual.js: Improve formatting and align with base assert

gibson042:
Consolidates handling of identity-bearing values via a lazy multi-use indication.

<details open><summary>comparison</summary>

```js
const v = [];
v.push(v);
const values = [
  '',
  Symbol('foo'),
  /./gui,
  [-0, 0, 0n],
  ((a, b) => [a, b, a, b, {}])({}, {}),
  new Date(0),
  { [Symbol.toStringTag]: 'errors', error: Error(), typeError: TypeError('foo') },
  function fn(){},
  new Set([v, {}, {}]),
  Object.create(null),
];
for (const x of values) v.push(x, x, 'NL');
String(assert.deepEqual.format(v)).replace(/ "NL",/g, '\n');
```

<details><summary>before</summary>

Note the frequently missing "as #*N*" and the total lack of indication for referentially-identical symbols/regular expressions/dates/etc., not to mention the inability to differentiate signed zeros from each other or numbers from bigints.

```
[[Ref: #1], "", "",
 Symbol(foo), Symbol(foo),
 /./giu, /./giu,
 [0, 0, 0], [Ref: #2],
 [Object {  }, Object {  }, [Ref: #4], [Ref: #5], Object {  }], [Ref: #3],
 Date "1970-01-01T00:00:00.000Z", Date "1970-01-01T00:00:00.000Z",
 errors { error: Object {  }, typeError: Object {  } }, [Ref: #7],
 [Function: fn], [Function: fn],
 Set {[Ref: #1], Object {  }, Object {  }}, [Ref: #10],
 [Object: null prototype] {  }, [Ref: #13], "NL"] as #1
```

</details>
<details><summary>after</summary>

```
[ref #1, "", "",
 Symbol(foo) as #2, ref #2,
 /./giu as #3, ref #3,
 [-0, 0, 0n] as #4, ref #4,
 [Object {} as #6, Object {} as #7, ref #6, ref #7, Object {}] as #5, ref #5,
 Date("1970-01-01T00:00:00.000Z") as #9, ref #9,
 errors {error: error Error(""), typeError: error TypeError("foo")} as #10, ref #10,
 function fn as #13, ref #13,
 Set {ref #1, Object {}, Object {}} as #14, ref #14,
 [Object: null prototype] {} as #17, ref #17, "NL"] as #1
```

</details>
</details>

### DIFF
--- a/harness/deepEqual.js
+++ b/harness/deepEqual.js
@@ -85,6 +85,8 @@ assert.deepEqual.format = function(value, seen) {
     return acceptMappers;
   }
 
+  let format = assert.deepEqual.format;
+
   if (typeof value === 'function') {
     return lazyOutput`function${value.name ? ` ${String(value.name)}` : ''}`;
   }
@@ -92,31 +94,31 @@ assert.deepEqual.format = function(value, seen) {
     return lazyOutput`${value}`;
   }
   if (Array.isArray ? Array.isArray(value) : value instanceof Array) {
-    return lazyOutput`[${value.map(value => assert.deepEqual.format(value, seen))}]`(join);
+    return lazyOutput`[${value.map(value => format(value, seen))}]`(join);
   }
   if (value instanceof Date) {
-    return lazyOutput`Date(${assert.deepEqual.format(value.toISOString(), seen)})`;
+    return lazyOutput`Date(${format(value.toISOString(), seen)})`;
   }
   if (value instanceof Error) {
-    return lazyOutput`error ${value.name || 'Error'}(${assert.deepEqual.format(value.message, seen)})`;
+    return lazyOutput`error ${value.name || 'Error'}(${format(value.message, seen)})`;
   }
   if (value instanceof RegExp) {
     return lazyOutput`${value}`;
   }
   if (typeof Map !== "undefined" && value instanceof Map) {
-    return lazyOutput`Map {${Array.from(value).map(pair => `${assert.deepEqual.format(pair[0], seen)} => ${assert.deepEqual.format(pair[1], seen)}`)}}`(join);
+    let contents = Array.from(value).map(pair => `${format(pair[0], seen)} => ${format(pair[1], seen)}`);
+    return lazyOutput`Map {${contents}}`(join);
   }
   if (typeof Set !== "undefined" && value instanceof Set) {
-    return lazyOutput`Set {${Array.from(value).map(value => assert.deepEqual.format(value, seen))}}`(join);
+    let contents = Array.from(value).map(value => format(value, seen));
+    return lazyOutput`Set {${contents}}`(join);
   }
 
   let tag = Symbol.toStringTag && Symbol.toStringTag in value
     ? value[Symbol.toStringTag]
-    : 'Object';
-  if (tag === 'Object' && Object.getPrototypeOf(value) === null) {
-    tag = '[Object: null prototype]';
-  }
-  return lazyOutput`${tag ? `${tag} ` : ''}{${Object.keys(value).map(key => `${key.toString()}: ${assert.deepEqual.format(value[key], seen)}`)}}`(String, join);
+    : Object.getPrototypeOf(value) === null ? '[Object: null prototype]' : 'Object';
+  let contents = Object.keys(value).map(key => `${String(key)}: ${format(value[key], seen)}`);
+  return lazyOutput`${tag ? `${tag} ` : ''}{${contents}}`(String, join);
 };
 
 assert.deepEqual._compare = (function () {
--- a/harness/deepEqual.js
+++ b/harness/deepEqual.js
@@ -14,19 +14,9 @@ assert.deepEqual = function(actual, expected, message) {
   );
 };
 
-assert.deepEqual.format = (function () {
-  function lazyStringFactory(strings, ...subs) {
-    return function(...mappers) {
-      assert(mappers.length === subs.length, 'mappers must be paired with substitutions');
-      let toString = () => {
-        let mappedSubs = subs.map((sub, i) => mappers[i](sub));
-        return strings.map((str, i) => `${i === 0 ? '' : mappedSubs[i - 1]}${str}`).join('');
-      };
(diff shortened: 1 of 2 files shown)


ptomato:
I had a look at the code, but before I figure out how the template tag works, a before/after comparison of the output would be helpful, as well as a bit more explanation of the motivation :smile:

By the way, is #3476 on your radar? I've been considering the use of deepEqual in new tests discouraged.

gibson042:
> before I figure out how the template tag works

Ah, I wonder if I should try to add some inline documentation. But basically, ``` lazyStringFactory`…` ``` returns a function that must be called with an argument for each substitution value in the template and returns an object with only a `toString` method. Calling that method (i.e., trying to use the returned object as a string as in `assert.deepEqual`) passes each substitution value to its mapper and uses the results for simple interpolation. The reason for this is that the `usage` objects are expected to be updated as object graph traversal that follows formatting of one value discovers reuse of the value being formatted, so we want to delay rendering of those usage objects until after the outermost `format` call has returned.

> a before/after comparison of the output would be helpful

Sure; added above.

> as well as a bit more explanation of the motivation 😄

It started as fixing issues like the extra closing brace in formatted map output and handling `-0` to align with `assert._toString` (for e.g. `assert.sameValue`), but grew a little bit to better present my results for https://github.com/tc39/ecma262/issues/3388 :
```
// { "byteLength": "Symbol(Symbol.toPrimitive) returns too big for index", "maxByteLength": "getter throws", "newTarget": "get prototype throws" }
 expected  [["byteLength", "get", Symbol(Symbol.toPrimitive) as #3], ["byteLength", ref #3, "call"], ErrorShape {ctor: function RangeError}]
   actual  [["byteLength", "get", Symbol(Symbol.toPrimitive) as #3], ["byteLength", ref #3, "call"], ["options", "get", "maxByteLength"], error GetMaxByteLengthError("")]
```

> By the way, is #3476 on your radar? I've been considering the use of deepEqual in new tests discouraged.

Hmm, it really does feel like the right tool for comparing logs as above.

harness/deepEqual.js
```diff
@@ -15,59 +15,76 @@ assert.deepEqual = function(actual, expected, message) {
 };
 
 assert.deepEqual.format = (function () {
+  function lazyStringFactory(strings, ...subs) {
```
Ms2ger, line 18: Please do add some docs here

Ms2ger (approved):
Seems okay to me

harness/deepEqual.js
```diff
@@ -15,59 +15,76 @@ assert.deepEqual = function(actual, expected, message) {
...
+    if (usage) {
+      usage.used = true;
+      return `ref #${usage.id}`;
+    }
+    usage = { id: ++seen.counter, used: false };
+    seen.map.set(value, usage);
+
+    if (typeof value === 'function') {
+      return lazyStringFactory`function${value.name ? ` ${String(value.name)}` : ''}${usage}`(String, renderUsage);
+    } else if (typeof value !== 'object') {
```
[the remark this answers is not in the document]
gibson042: Changed back to that style; `else if` ended up too dense to be readable.

harness/deepEqual.js
gibson042: Done, with corresponding refactoring.

Ms2ger (approved):

harness/deepEqual.js
```diff
@@ -14,56 +14,117 @@ assert.deepEqual = function(actual, expected, message) {
   );
 };
 
+let getOwnPropertyDescriptor = Object.getOwnPropertyDescriptor;
+let join = arr => arr.join(', ');
```
ljharb, line 18:
```suggestion
function join(arr) { return arr.join(', '); }
```

harness/deepEqual.js
```diff
@@ -14,56 +14,117 @@ assert.deepEqual = function(actual, expected, message) {
   );
 };
 
+let getOwnPropertyDescriptor = Object.getOwnPropertyDescriptor;
```
ljharb, line 17:
```suggestion
var getOwnPropertyDescriptor = Object.getOwnPropertyDescriptor;
```

ljharb (commented):
in general, let's please avoid newer syntax unless it's required for the test (including spread, arrows, let/const, etc)

using tagged template literals also seems like it will permanently prevent the tests from running on an engine that doesn't have template literal support. is there a way to achieve the same goal without using backticks?

gibson042:
> in general, let's please avoid newer syntax unless it's required for the test (including spread, arrows, let/const, etc)
> 
> using tagged template literals also seems like it will permanently prevent the tests from running on an engine that doesn't have template literal support. is there a way to achieve the same goal without using backticks?

I very much agree with this, but [deepEqual.js](https://github.com/tc39/test262/blob/main/harness/deepEqual.js) _already_ uses templates, arrows, and lexical bindings: https://github.com/tc39/test262/blob/5ae7de994287c93a75e474252ab70adfadc81b04/harness/deepEqual.js#L59-L63

A somewhat ergonomic fix printf-style fix is possible, but really seems more appropriate for a followup:
```js
var tag = Symbol.toStringTag && Symbol.toStringTag in value
  ? value[Symbol.toStringTag]
  : Object.getPrototypeOf(value) === null ? '[Object: null prototype]' : 'Object';
let keys = Reflect.ownKeys(value).filter(function(key) {
  return getOwnPropertyDescriptor(value, key).enumerable;
});
let contents = keys.map(function(key) {
  // return lazyString`${escapeKey(key)}: ${format(value[key], seen)}`;
  return callTag(lazyString, '%v: %v', escapeKey(key), format(value[key], seen));
});
// return lazyResult`${tag ? `${tag} ` : ''}{${contents}}`(String, join);
return callTag(lazyResult, '%v{%v}', tag ? String(tag) + ' ' : '', contents)(String, join);
```

ljharb:
Indeed if the syntax is already in the file, my comment doesn't apply.

Issue #4096, opened 2024-05-27
Testing plan for JSON-parse-with-source


ioannad:
[Proposal README](https://github.com/tc39/proposal-json-parse-with-source/)
[Proposal rendered spec](https://tc39.es/proposal-json-parse-with-source/)

For testing the tests, the proposal is shipped in v8.

--------------

- [ ] Tests for new function `JSON.isRawJSON`:
  - [ ] Boilerplate:
    - [ ] Can be invoked as a function.
    - [ ] Function length.
    - [ ] Function name.
    - [ ] Property descriptor.
    - [ ] Function is not a constructor.
    - [ ] Brand check for this method.
  - [ ] Success cases:
    - [ ] An instance created by a call to `JSON.rawJSON`.
    - [ ] Some objects without an `[[IsRawJSON]]` internal slot: `{}`, `[]`.
    - [ ] Some non-Objects: `null`, `undefinded`, a Boolean, a symbol, a number.
- [ ] Tests for new function `JSON.rawJSON`:
  - [ ] Boilerplate:
    - [ ] Can be invoked as a function.
    - [ ] Function length.
    - [ ] Function name.
    - [ ] Function is not a constructor.
    - [ ] Property descriptor.
    - [ ] Brand check for this method.
  - [ ] Success cases, a test for each branch:
    - [ ] Number, negative number, string, `true`, `false`, `null`
  - [ ] Failure cases, throwing:
    - [ ] `SyntaxError` in case of:
      - [ ] The empty string.
      - [ ] A string starting or ending with a tab, a linefeed, a carriage return, or a space.
      - [ ] A non-Array object literal: `{}`.
      - [ ] An Array literal: `[1]`.
    - [ ] `TypeError` when trying to extend a rawJSON result object.
- [ ] Tests for the new abstract operations in the algorithm of `JSON.parse`, each test should have a callable `reviver` argument:
  - [ ] For paths in `CreateJSONParseRecord`, only success cases (normal output), a test for each branch:
    - [ ] an Array literal
    - [ ] an Object literal
    - [ ] an Object literal with a repeated key
    - [ ] NullLiteral, BooleanLiteral, NumericLiteral, StringLiteral, ArrayLiteral, ObjectLiteral, UnaryExpression
    - [ ] a combination of all the above success cases nested
[...]

PR #4255, opened 2024-10-09
Boilerplate tests for rawJSON and isRawJSON

ioannad:
from the proposal json-parse-with-source.

Partially addresses testing plan issue #4096


ljharb (approved):

test/built-ins/JSON/isRawJSON/length.js
```diff
@@ -0,0 +1,25 @@
...
+esid: sec-json.rawjson
+description: >
+  JSON.isRawJSON.length value and descriptor.
+info: |
+  JSON.isRawJSON ( text )
+
+  18 ECMAScript Standard Built-in Objects
+  Every other data property described in clauses 19 through 28 and in Annex B.2
+  has the attributes { [[Writable]]: true, [[Enumerable]]: false,
+  [[Configurable]]: true } unless otherwise specified.
```
ptomato, line 14: It would be fine with me to delete this instead of changing it, but it is not the relevant section of ECMA-262 for this test. If you decide to keep it, I'd use this frontmatter:

```
esid: sec-built-in-function-objects
info: |
    Every built-in function object, including constructors, has a *"length"*
    property whose value is a non-negative integral Number. Unless otherwise
    specified, this value is the number of required parameters shown in the
    subclause heading for the function description. Optional parameters and rest
    parameters are not included in the parameter count.

    Unless otherwise specified, the *"length"* property of a built-in function
    object has the attributes { [[Writable]]: *false*, [[Enumerable]]: *false*,
    [[Configurable]]: *true* }.
```

test/built-ins/JSON/isRawJSON/name.js
```diff
@@ -0,0 +1,29 @@
...
+esid: sec-json.israwjson
+description: >
+  JSON.isRawJSON.name value and descriptor.
+info: |
+  JSON.isRawJSON ( text )
+
+  18 ECMAScript Standard Built-in Objects
+  Every other data property described in clauses 19 through 28 and in Annex B.2
+  has the attributes { [[Writable]]: true, [[Enumerable]]: false,
+  [[Configurable]]: true } unless otherwise specified.
```
ptomato, line 14: Same, either delete or replace with this section:

```
esid: sec-built-in-function-objects
info: |
    Every built-in function object, including constructors, has a *"name"*
    property whose value is a String. Unless otherwise specified, this value is
    the name that is given to the function in this specification. Functions that
    are identified as anonymous functions use the empty String as the value of
    the *"name"* property. For functions that are specified as properties of
    objects, the name value is the property name string used to access the
    function. Functions that are specified as get or set accessor functions of
    built-in properties have *"get"* or *"set"* (respectively) passed to the
    prefix parameter when calling CreateBuiltinFunction.

    The value of the *"name"* property is explicitly specified for each built-in
    function whose property key is a Symbol value. If such an explicitly
    specified value starts with the prefix *"get "* or *"set "* and the function
    for which it is specified is a get or set accessor function of a built-in
    property, the value without the prefix is passed to the name parameter, and
    the value *"get"* or *"set"* (respectively) is passed to the prefix
    parameter when calling CreateBuiltinFunction.

    Unless otherwise specified, the *"name"* property of a built-in function
    object has the attributes { [[Writable]]: *false*, [[Enumerable]]: *false*,
    [[Configurable]]: *true* }.
```

test/built-ins/JSON/rawJSON/length.js
```diff
@@ -0,0 +1,25 @@
...
+esid: sec-json.rawjson
+description: >
+  JSON.rawJSON.length value and descriptor.
+info: |
+  JSON.rawJSON ( text )
+
+  18 ECMAScript Standard Built-in Objects
+  Every other data property described in clauses 19 through 28 and in Annex B.2
+  has the attributes { [[Writable]]: true, [[Enumerable]]: false,
+  [[Configurable]]: true } unless otherwise specified.
```
ptomato, line 14: Same comment about section of ECMA-262.

test/built-ins/JSON/rawJSON/name.js
```diff
@@ -0,0 +1,29 @@
...
+description: >
+  JSON.rawJSON.name value and descriptor.
+info: |
+  JSON.rawJSON ( text )
+
+  18 ECMAScript Standard Built-in Objects
+  Every other data property described in clauses 19 through 28 and in Annex B.2
+  has the attributes { [[Writable]]: true, [[Enumerable]]: false,
+  [[Configurable]]: true } unless otherwise specified.
+
```
ptomato, line 15: Same.

test/built-ins/JSON/rawJSON/name.js
```diff
@@ -0,0 +1,29 @@
...
+includes: [propertyHelper.js]
+features: [json-parse-with-source]
+---*/
+
+assert.sameValue(
+  JSON.rawJSON.name, 'rawJSON',
+  'The value of JSON.rawJSON.name is "rawJSON"'
+);
```
ptomato, line 23: Same comment about moving this into the `verifyProperty` call.

ptomato (approved):
Thanks, looks great, but I'd either update or delete the frontmatter in the length and name tests so that it points to the relevant section.

ptomato (approved):
:+1:

PR #4256, opened 2024-10-10
Delete .hgignore

linusg:


ljharb (approved):

Ms2ger:
Any particular reason?

linusg:
Leftover from more than a decade ago, serving no purpose and cluttering up the file view.

ptomato:
I thought this might still be present for convenience because test262 is imported into the Firefox source tree, which uses Mercurial for version control. But looking there, I see it's not even present because they only import the `test/` folder: https://searchfox.org/mozilla-central/source/js/src/tests/test262
So I'm going with the assumption that nothing uses this.

PR #4257, opened 2024-10-10
Update built-ins/Date to use verifyProperty

anba:


Ms2ger (approved):
This is way too much to usefully review. I'll rubber-stamp, but wait until next week to see if anyone has comments.

ljharb:
hm, it looks like a bunch of the files were renamed but git didn't track the rename. @anba, would it be possible to use revert the names, and then use `git mv` to rename them? it's also possible the file renaming needs to be done in a different commit than the one that makes changes to the files.

anba:
@ljharb I'm a bit lost here. I didn't move/rename any files.

ljharb:
oh, that's my confusion then. why are some files deleted?

anba:
Because `verifyProperty` subsumes the individual tests from the deleted files.

ptomato:
I guess these should use `verifyPrimordialProperty` now; #4234 already ported one of them as an example.

anba:
I'd prefer to bulk replace `verifyProperty` with `verifyPrimordialProperty` in a separate PR.

anba:
Rebased to apply cleanly on main.

Ms2ger:
I'm afraid you'll need another one after #4263

anba:
Rebase done.

PR #4262, opened 2024-10-13
test(top-level-await): move out of staging module graphs hanging

### LINKED ISSUE #4250: Tests for top-level await with circular module import
Repo: https://github.com/tc39/proposal-top-level-await/
Spec text: https://tc39.github.io/proposal-top-level-await/
ECMA-262 PR: https://github.com/tc39/ecma262/pull/2408

In the Google Chrome v129 release, the JavaScript module import is broken with circular + top-level-await (discovered by @antfu).
Related:
- https://github.com/antfu/repro-chrome-v129
- https://x.com/antfu7/status/1836846357489012816

To avoid other engines or implementers to also break accidentally this feature, or even for Google Chrome to avoid breaking it again, it would be great to add tests for this in Test262 (as suggested here: https://x.com/robpalmer2/status/1836846820418601384).

## Demonstration

[Online demo](https://repro-chrome-129-circular-tla.netlify.app/)

When the entry file is `foo.js`, with the following content:

```js
// bar.js
import { define } from './foo.js'

export default define('It works!')
```

```js
// foo.js
export function define(foo) {
  return foo.toUpperCase()
}

const getter = () => import('./bar.js')

await 1

getter()
  .then(r => {
    // this will never be resolved
  })
  .catch(e => {
    // this either
  })
```

Note that if remove the `await 1`, it will work again.

theoludwig:
Fixes #4250 

This PR moves the tests added in the staging directory in #4116, to the main test262 test suite.

(I'm new here, I don't know if this is the way to do it, I renamed the test files so that they can be close to each other)


nicolo-ribaudo (approved):

theoludwig:
I don't think the CI failure is related to this PR, right?

nicolo-ribaudo:
Yeah, it was due to GitHub upgrading their Ubuntu images.

ptomato (approved):
Thanks!

PR #4263, opened 2024-10-14
chore: convert many tests to use verifyProperty

nektro:
edit: i see `verifyPrimordialProperty` got added recently, if any of these should use that instead lmk

### DIFF
--- a/test/built-ins/Array/prototype/Symbol.unscopables/array-find-from-last.js
+++ b/test/built-ins/Array/prototype/Symbol.unscopables/array-find-from-last.js
@@ -31,6 +31,6 @@ verifyProperty(unscopables, "findLast", {
 assert.sameValue(unscopables.findLastIndex, true, '`findLastIndex` property value');
 verifyProperty(unscopables, "findLastIndex", {
   writable: true,
-  enumerable: false,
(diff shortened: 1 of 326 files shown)


test/built-ins/Array/prototype/Symbol.unscopables/array-find-from-last.js
```diff
@@ -21,11 +21,16 @@ var unscopables = Array.prototype[Symbol.unscopables];
...
+});
+
 
 assert.sameValue(unscopables.findLastIndex, true, '`findLastIndex` property value');
-verifyEnumerable(unscopables, 'findLastIndex');
-verifyWritable(unscopables, 'findLastIndex');
-verifyConfigurable(unscopables, 'findLastIndex');
+verifyProperty(unscopables, "findLastIndex", {
+  writable: true,
+  enumerable: false,
```
anba, line 34:
```suggestion
  enumerable: true,
```

anba (approved):
There's a single typo, the rest looks correct to me.

Ms2ger (approved):
Probably most if not all of those should use verifyPrimordialProperty, but I think it's worth merging these changes anyway.

PR #4265, opened 2024-10-15
Migrate staging tests for JSON-parse-with-source

ioannad:
Partially addresses testing plan issue #4096


test/built-ins/JSON/isRawJSON/invoked-as-a-fn.js
```diff
@@ -0,0 +1,42 @@
...
+
+includes: [deepEqual.js]
+features: [json-parse-with-source]
+---*/
+
+function assertIsRawJson(rawJson, expectedRawJsonValue) {
+  assert.sameValue(null, Object.getPrototypeOf(rawJson));
+  assert(Object.hasOwn(rawJson, 'rawJSON'));
+  assert.deepEqual(['rawJSON'], Object.getOwnPropertyNames(rawJson));
+  assert.deepEqual([], Object.getOwnPropertySymbols(rawJson));
```
ptomato, line 22: We are trying to phase out `deepEqual` (#3476), but these work equally well with compareArray:
```suggestion
  assert.compareArray(Object.getOwnPropertyNames(rawJson), ['rawJSON']);
  assert.compareArray(Object.getOwnPropertySymbols(rawJson), []);
```
(note you will have to replace the includes with `[compareArray.js]`, also note that test262's assert functions place the actual value first then the expected value. V8's assert functions do it the other way around, so we have to switch them otherwise implementations will get confusing debug messages.)

test/built-ins/JSON/isRawJSON/invoked-as-a-fn.js
ptomato, line 1: I'd suggest naming this test `basic.js` and changing the description to "Basic functionality of JSON.isRawJSON". That's really what it's testing, not just that the `isRawJSON` function object is callable.

test/built-ins/JSON/isRawJSON/invoked-as-a-fn.js
```diff
@@ -0,0 +1,42 @@
...
+  assert.sameValue(expectedRawJsonValue, rawJson.rawJSON);
+}
+
+assertIsRawJson(JSON.rawJSON(1), '1');
+assertIsRawJson(JSON.rawJSON(null), 'null');
+assertIsRawJson(JSON.rawJSON(true), 'true');
+assertIsRawJson(JSON.rawJSON(false), 'false');
+assertIsRawJson(JSON.rawJSON('"foo"'), '"foo"');
```
ptomato, line 30: These lines probably belong more in a `rawJSON()` test, e.g. `JSON/rawJSON/returns-expected-object.js`

test/built-ins/JSON/rawJSON/invoked-as-a-fn-BigInt.js
ptomato, line 1: I'd change the description to "Result of JSON.rawJSON on a BigInt can be stringified" and the name to something like `bigint-raw-json-can-be-stringified.js`

test/built-ins/JSON/rawJSON/invoked-as-a-fn-BigInt.js
```diff
@@ -0,0 +1,31 @@
...
+  8. Return obj.
+
+features: [json-parse-with-source]
+---*/
+
+const tooBigForNumber = BigInt(Number.MAX_SAFE_INTEGER) + 2n;
+const intToBigInt = (key, val, { source }) =>
+  typeof val === 'number' && val % 1 === 0 ? BigInt(source) : val;
+const roundTripped = JSON.parse(String(tooBigForNumber), intToBigInt);
+assert.sameValue(tooBigForNumber, roundTripped);
```
ptomato, line 26: Here and throughout, reverse the order of (expected, actual) to match (actual, expected)

test/built-ins/JSON/rawJSON/illegal-empty-and-start-end-chars.js
```diff
@@ -0,0 +1,32 @@
+// Copyright (C) 2023 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-json.rawjson
+description: >
+  JSON.rawJSON can be invoked as a function
```
ptomato, line 7:
```suggestion
  Throws SyntaxError on empty string, or string with illegal start and end characters
```

ptomato (changes requested):
Thanks!

ptomato (approved):
Ioanna isn't available in the short term to work further on this, so I went ahead and made the modifications that I requested. Anyone else feel like giving it another look?

Ms2ger (approved):

PR #4266, opened 2024-10-15
add `Error.isError` tests

ljharb:
See https://github.com/tc39/proposal-is-error/issues/7


test/built-ins/Error/isError/is-a-constructor.js
```diff
@@ -0,0 +1,24 @@
...
+
+  If Type(argument) is not Object, return false.
+  If argument has a [[Construct]] internal method, return true.
+  Return false.
+includes: [isConstructor.js]
+features: [Reflect.construct]
+---*/
+
+assert.sameValue(isConstructor(Error.isError), false, 'isConstructor(Error.isError) must return false');
+assert.throws(function () { new Error.isError(); });
```
ptomato, line 23:
```suggestion
assert.throws(TypeError, function () { new Error.isError(); });
```

ptomato (approved):
LGTM but with `s/assert.equal/assert.sameValue/g` (or in the case of `assert.equal(..., boolean)` consider replacing with `assert()`)

ljharb:
Thanks, updated (kept the explicit comparisons because i prefer them over implicit ones)

ptomato (approved):
Thanks!

iamstolis:
I think that these tests are missing a feature flag (for example, `Error.isError`).

ptomato:
> I think that these tests are missing a feature flag (for example, `Error.isError`).

Thanks, but would you mind opening an issue (or a new PR if you have the time?) As a comment on a merged PR, it is likely to be missed.

ljharb:
whoops, thanks. i'll add one in a PR right now.

ljharb:
→ #4271

PR #4268, opened 2024-10-15
DurationFormat: Add tests for durations with large hour/minute/second unit values while using digital style

### LINKED ISSUE #4190: Need test for intl-duration-format/pull/198
@ben-allen 
Need unit test to test the changes inside
https://github.com/tc39/proposal-intl-duration-format/pull/198 

@sffc @anba

ben-allen:
Grouping separators suppressed in this situation. See tc39/proposal-intl-duration-format/pull/198

fix #4190


ptomato (approved):
Thanks!

PR #4269, opened 2024-10-15
A few minor fixes for the test generation tool

ptomato:
Found these while working on other stuff.

### DIFF
--- a/tools/generation/lib/template.py
+++ b/tools/generation/lib/template.py
@@ -182,7 +182,9 @@ class Template:
         flags += case_values['meta'].get('flags', [])
         flags += self.attribs['meta'].get('flags', [])
         flags = without_duplicates(flags)
-        if 'async' in flags and negative and negative.get('phase') == 'parse' and negative.get('type') == 'SyntaxError':
+        if (negative and negative.get('phase') == 'parse'
+                and negative.get('type') == 'SyntaxError'
+                and 'async' in flags):
             flags.remove('async')
         lines += ['flags: ' + yaml.dump(flags, default_flow_style=True,
                   width=float('inf')).strip()]
--- a/tools/generation/lib/template.py
+++ b/tools/generation/lib/template.py
@@ -49,6 +49,11 @@ def indent(text, prefix = '    ', js_value = False):
 
     return '\n'.join(indented)
 
+
+def without_duplicates(iterable):
+    return list(OrderedDict.fromkeys(iterable))
+
+
 class Template:
     def __init__(self, filename, encoding):
         self.filename = filename
@@ -160,7 +165,7 @@ class Template:
         features = []
         features += case_values['meta'].get('features', [])
         features += self.attribs['meta'].get('features', [])
-        features = list(OrderedDict.fromkeys(features))
+        features = without_duplicates(features)
         if len(features):
             lines += ['features: ' + yaml.dump(features,
                       default_flow_style=True, width=float('inf')).strip()]
@@ -176,7 +181,7 @@ class Template:
         flags = ['generated']
         flags += case_values['meta'].get('flags', [])
         flags += self.attribs['meta'].get('flags', [])
-        flags = list(OrderedDict.fromkeys(flags))
+        flags = without_duplicates(flags)
         if 'async' in flags and negative and negative.get('phase') == 'parse' and negative.get('type') == 'SyntaxError':
             flags.remove('async')
         lines += ['flags: ' + yaml.dump(flags, default_flow_style=True,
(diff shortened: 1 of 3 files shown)


tools/generation/lib/template.py
```diff
@@ -178,14 +179,14 @@ def _frontmatter(self, case_filename, case_values):
         flags = list(OrderedDict.fromkeys(flags))
         if 'async' in flags and negative and negative.get('phase') == 'parse' and negative.get('type') == 'SyntaxError':
             flags.remove('async')
-        lines += ['flags: ' + re.sub('\n\s*', ' ', yaml.dump(flags, default_flow_style=True).strip())]
+        lines += ['flags: ' + re.sub(initial_space_pattern, ' ', yaml.dump(flags, default_flow_style=True).strip())]
```
gibson042, line 182:
```suggestion
        lines += ['flags: ' + yaml.dump(flags, default_flow_style=True).strip()]
```

tools/generation/lib/template.py
```diff
@@ -161,8 +161,9 @@ def _frontmatter(self, case_filename, case_values):
         features += case_values['meta'].get('features', [])
         features += self.attribs['meta'].get('features', [])
         features = list(OrderedDict.fromkeys(features))
```
gibson042, line 163: It took me a while to figure out the purpose of this line.
```suggestion
        # remove duplicates while preserving order
        features = list(OrderedDict.fromkeys(features))
```
It's also a good candidate for abstraction à la `indent`:
```python
def without_duplicates(iterable):
    return list(OrderedDict.fromkeys(iterable))
```

tools/generation/lib/template.py
```diff
@@ -178,14 +179,14 @@ def _frontmatter(self, case_filename, case_values):
         flags = list(OrderedDict.fromkeys(flags))
         if 'async' in flags and negative and negative.get('phase') == 'parse' and negative.get('type') == 'SyntaxError':
             flags.remove('async')
```
gibson042, line 181: More micro-optimizations:
```suggestion
        if negative and negative.get('phase') == 'parse' and negative.get('type') == 'SyntaxError' and 'async' in flags:
            flags.remove('async')
```

tools/generation/lib/template.py
```diff
@@ -178,14 +179,14 @@ def _frontmatter(self, case_filename, case_values):
...
-        lines += ['flags: ' + re.sub('\n\s*', ' ', yaml.dump(flags, default_flow_style=True).strip())]
+        lines += ['flags: ' + re.sub(initial_space_pattern, ' ', yaml.dump(flags, default_flow_style=True).strip())]
 
         includes = []
         includes += case_values['meta'].get('includes', [])
         includes += self.attribs['meta'].get('includes', [])
         includes = list(OrderedDict.fromkeys(includes))
         if len(includes):
-            lines += ['includes: ' + re.sub('\n\s*', ' ', yaml.dump(includes, default_flow_style=True).strip())]
+            lines += ['includes: ' + re.sub(initial_space_pattern, ' ', yaml.dump(includes, default_flow_style=True).strip())]
```
gibson042, line 189:
```suggestion
            lines += ['includes: ' + yaml.dump(includes, default_flow_style=True).strip()]
```

gibson042 (approved):
:+1:

tools/generation/lib/template.py
```diff
@@ -161,8 +161,9 @@ def _frontmatter(self, case_filename, case_values):
         features += case_values['meta'].get('features', [])
         features += self.attribs['meta'].get('features', [])
         features = list(OrderedDict.fromkeys(features))
+        initial_space_pattern = re.compile(r'\n\s*')
         if len(features):
-            lines += ['features: ' + re.sub('\n\s*', ' ', yaml.dump(features, default_flow_style=True).strip())]
+            lines += ['features: ' + re.sub(initial_space_pattern, ' ', yaml.dump(features, default_flow_style=True).strip())]
```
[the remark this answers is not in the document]
ptomato: Good point!

The proof is in the pudding, so I deleted all the generated files and regenerated them with this modification; turns out `yaml.dump` wraps long lines at 80 by default, so that's what this regexp was undoing. I added `width=float('inf')` and that seems sufficient to reproduce the original generated files.

tools/generation/lib/template.py
ptomato: Yeah, good idea, this makes it less puzzling.

ptomato:
Checked that all existing generated tests are regenerated with no changes.

Issue #4267, opened 2024-10-15
ESMeta CI job is failing


ptomato:
Since a couple of days ago, the ESMeta CI job is failing because it cannot locate the `sbt` binary.

ptomato:
cc @doehyunbaek @jhnaldo - any idea what is happening?

ptomato:
See https://github.com/actions/runner-images/issues/10788 and https://github.com/actions/runner-images/issues/10636

jhnaldo:
Thanks for letting me know the problem.

> See [actions/runner-images#10788](https://github.com/actions/runner-images/issues/10788) and [actions/runner-images#10636](https://github.com/actions/runner-images/issues/10636)

This issue seems the main reason for the problem.

Should we wait until this issue is resolved in `actions`, or should we try to resolve this issue now?

ljharb:
What would probably be wisest is to add a step to https://github.com/tc39/test262/blob/main/.github/workflows/esmeta-test262.yml (and https://github.com/tc39/ecma262/blob/main/.github/workflows/esmeta-typecheck.yml) that installs `sbt`, that's skipped if it's already present. That way, if github fixes it, the job will be faster, but it'll work regardless.

Additionally, it'd probably be good to create a single esmeta action that both of these workflows can use, so that there's only one place needed to change for stuff like this.

ptomato:
https://github.com/tc39/test262/pull/4270 should fix it. It doesn't skip installing SBT if already present, but I think it's pretty unlikely that GitHub would re-add it, given that people have been complaining for three weeks already.

PR #4270, opened 2024-10-16
CI: Setup SBT manually

ptomato:
One of the core developers of SBT has kindly created an action to replace the missing SBT in newer Ubuntu images.
https://github.com/marketplace/actions/setup-sbt-installer

Closes: #4267

### DIFF
--- a/.github/workflows/esmeta-test262.yml
+++ b/.github/workflows/esmeta-test262.yml
@@ -14,6 +14,8 @@ jobs:
         with:
           distribution: temurin
           java-version: 17
+      - name: Setup SBT
+        uses: sbt/setup-sbt@v1
       - name: Add to PATH
         shell: bash
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/esmeta-test262.yml
```diff
@@ -14,6 +14,8 @@ jobs:
         with:
           distribution: temurin
           java-version: 17
+      - name: Setup SBT
+        uses: sbt/setup-sbt@v1
```
ljharb, line 18: can we skip this step when sbt is already present?

ptomato: I'm not so familiar with GitHub actions that I know how to do that off the top of my head. I can figure it out but I have some other stuff to do first.

ljharb: `if: ` and then the condition; actually checking the condition might be tricky tho. doesn't have to be a blocker.

ljharb (approved):
skipping not a blocker

.github/workflows/esmeta-test262.yml
ptomato: Found something that seems like it should work.

ptomato:
Merging as per maintainers meeting

PR #4273, opened 2024-10-21
fix: heritage-arrow-function.js

KermanX:
Previously this test was the exact same as https://github.com/tc39/test262/tree/main/test/language/expressions/class/heritage-async-arrow-function.js, which I think is wrong.


Ms2ger (commented):
Please also update "async arrow" on line 7

Issue #4259, opened 2024-10-11
Need to add test for ECMA402 PR925


FrankYFTang:
https://github.com/tc39/ecma402/pull/925 merged after TC39 2024 Oct meeting. We need to add test to verify the implementation.

@ben-allen @sffc @anba

PR #4274, opened 2024-10-21
Add test:`Intl.NumberFormat` using "currency" style and "compact", "engineering", or "scientific" notations.

ben-allen:
Related PR: https://github.com/tc39/ecma402/pull/925

fix https://github.com/tc39/test262/issues/4259

### DIFF
--- a/test/intl402/NumberFormat/currency-digits-nonstandard-notation.js
+++ b/test/intl402/NumberFormat/currency-digits-nonstandard-notation.js
@@ -1,0 +1,34 @@
+// Copyright 2024 Igalia S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-intl.numberformat
+description: >
+  Checks that the special handling for number of fractional digits when formatting currency values in "standard" notation is ignored when using "compact", "engineering", or "scientific" notation.
+info: |
+  Intl.DateTimeFormat ( [ locales [ , options ] ] )
+
+  ...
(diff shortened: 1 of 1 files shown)


ptomato (approved):
Thanks!

trflynn89:
Hi @ben-allen - question about this test when the notation is "compact". In that case, wouldn't `resolvedOptions.maximumFractionDigits` be 0?

In the constructor, we have `mnfdDefault=0` and `mxfdDefault=3`. But when we go into `SetNumberFormatDigitOptions`, we end up setting `needFd` to false here:
```
21. If roundingPriority is "auto", then
    a. Set needSd to hasSd.
    b. If needSd is true, or hasFd is false and notation is "compact", then
       i. Set needFd to false.
```
`hasFd` is false (because the test doesn't specify `minimumFractionDigits` or `maximumFractionDigits`), and notation is "compact", so `needFd` becomes false.

`needSd` is similarly false, so we end up setting both `intlObj.[[MinimumFractionDigits]]` and `intlObj.[[MaximumFractionDigits]]` to 0:
```
24. If needSd is false and needFd is false, then
    a. Set intlObj.[[MinimumFractionDigits]] to 0.
    b. Set intlObj.[[MaximumFractionDigits]] to 0.
```

iamstolis:
> Hi @ben-allen - question about this test when the notation is "compact". In that case, wouldn't `resolvedOptions.maximumFractionDigits` be 0?

I agree that `maximumFractionDigits` should be `0` when `notation` is `"compact"`. So, I have created the corresponding issue (#4349).

PR #4276, opened 2024-10-22
Add tests to `Intl.DateTimeFormat` and `Intl.RelativeTimeFormat` for various numbering systems

### LINKED ISSUE #4260: Need to add test for ECMA402 PR919
https://github.com/tc39/ecma402/pull/919 reach TC39 consense in 2024 Oct meeting and merged. We need to add test to verify the implementation
@sffc @ben-allen @anba

ben-allen:
Previously there was a spec bug: the `Intl.NumberFormat`s created within `Intl.DateTimeFormat.prototype.format` and `Intl.RelativeTimeFormat` constructor were not passed the numbering system to be used. 

See https://github.com/tc39/ecma402/pull/919/
Fix #4260

### DIFF
--- a/test/intl402/DateTimeFormat/prototype/format/numbering-system.js
+++ b/test/intl402/DateTimeFormat/prototype/format/numbering-system.js
@@ -1,0 +1,59 @@
+// Copyright 2024 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-formatdatetimepattern
+description: |
+  Checks that numberingSystem property used correctly.
+info: >
+  11.5.5 FormatDateTimePattern ( dateTimeFormat, format, pattern, epochNanoseconds )
+    ...
+    3. Perform ! CreateDataPropertyOrThrow(nfOptions, "numberingSystem", dateTimeFormat.[[NumberingSystem]]).
+    ...
+    8. Perform ! CreateDataPropertyOrThrow(nf2Options, "numberingSystem", dateTimeFormat.[[NumberingSystem]]).
+    ...
+    11. If format has a field [[fractionalSecondDigits]], then
+    ...
+      d. Perform ! CreateDataPropertyOrThrow(nf3Options, "numberingSystem", dateTimeFormat.[[NumberingSystem]]).
+
+locale: [en-US]
+---*/
+
+const localesAndResults = [
+  [ "en-US", "2:35:06", "2:35:06.789", "02:35:06", "6" ],
+  [ "en-US-u-nu-arab", "٢:٣٥:٠٦", "٢:٣٥:٠٦٫٧٨٩", "٠٢:٣٥:٠٦", "٦" ],
+  [ "en-US-u-nu-deva", "२:३५:०६", "२:३५:०६.७८९", "०२:३५:०६", "६" ],
+  [ "en-US-u-nu-hanidec", "二:三五:〇六", "二:三五:〇六.七八九", "〇二:三五:〇六 AM", "六" ],
+]
+const time = new Date(2024, 0, 1, 2, 35, 6, 789);
+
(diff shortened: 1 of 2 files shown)


ben-allen:
Updated tests to be slightly less "golden string" -- instead of checking for exact match between formatted string and expected string, now expected is a substring without whitespace, and test is for whether formatted string includes the expected substring.

test/intl402/DateTimeFormat/prototype/format/numbering-system.js
```diff
@@ -0,0 +1,59 @@
...
+    ...
+    3. Perform ! CreateDataPropertyOrThrow(nfOptions, "numberingSystem", dateTimeFormat.[[NumberingSystem]]).
+    ...
+    8. Perform ! CreateDataPropertyOrThrow(nf2Options, "numberingSystem", dateTimeFormat.[[NumberingSystem]]).
+    ...
+    11. If format has a field [[fractionalSecondDigits]], then
+    ...
+      d. Perform ! CreateDataPropertyOrThrow(nf3Options, "numberingSystem", dateTimeFormat.[[NumberingSystem]]).
+
+locale: [en-US]
```
ptomato, line 19: I'm not sure about this, but I think you need to include `en-US-u-nu-arab` and the others in this list as well.

test/intl402/RelativeTimeFormat/prototype/format/en-us-numbering-systems.js
```diff
@@ -0,0 +1,26 @@
...
+
+/*---
+esid: sec-Intl.RelativeTimeFormat.prototype.format
+description: Checks that numberingSystem option used correctly.
+info: |
+    17.5.2 PartitionRelativeTimePattern ( relativeTimeFormat, value, unit )
+
+    11. Let fv be PartitionNumberPattern(relativeTimeFormat.[[NumberFormat]], ℝ(value)).
+
+locale: [en-US]
```
ptomato, line 12: Ditto.

test/intl402/RelativeTimeFormat/prototype/format/en-us-numbering-systems.js
```diff
@@ -0,0 +1,26 @@
...
+  ["en-US-u-nu-arab", "١٬٢٣٤٬٥٦٧٬٨٩٠"],
+  ["en-US-u-nu-deva", "१,२३४,५६७,८९०"],
+  ["en-US-u-nu-hanidec", "一,二三四,五六七,八九〇"],
+];
+const seconds = 1234567890;
+
+for (const [locale, expected] of localesAndResults){
+  const formatted = new Intl.RelativeTimeFormat(locale, {style: "short"}).format(seconds, "seconds");
+  assert.sameValue(formatted.includes(expected), true, `locale: ${locale}`);
+}
```
ptomato, line 26: Unlike the previous test, you might actually be able to make this one even less "golden", with something like
```js
const expected = new Intl.NumberFormat(locale).format(seconds);
```

ptomato (changes requested):
Thanks! Your effort in using fewer "golden strings" is appreciated.

test/intl402/RelativeTimeFormat/prototype/format/en-us-numbering-systems.js
ben-allen: updated!

ptomato (approved):

iamstolis:
I find it problematic/incorrect that the `expectedTwoDigit` result of `en-US-u-nu-hanidec` ends with ` AM`. Not only this is inconsistent with other expected results (that do not contain this suffix despite the actual format contains it most likely) but the space before `AM` in the expected string is `U+0020` Space while ICU/CLDR suggests `U+202F` Narrow No-Break Space. While the suggestion of `U+202F` is a bit controversial and some engines decided not to follow it (see, for example, [these lines](https://github.com/v8/v8/blob/524b202bf1c281f09ddf70582ea53be63716f780/src/objects/js-date-time-format.cc#L1451-L1454) in V8), this test should not prefer the engines that do so.

Issue #4215, opened 2024-09-02
`import defer` testing plan


nicolo-ribaudo:
https://github.com/tc39/proposal-defer-import-eval/

### Valid cases

```js
import defer * as ns from "mod";
```
```js
import defer from "mod";
```
```js
import defer * as ns from "mod";
export { ns };
```

### Invalid cases (one per file)

```js
import defer x from "mod";
import defer { x } from "mod";
import defer y, * as ns from "mod";
import defer as ns from "mod";
export defer * as ns from "mod";
```

- [x] All the existing dynamic import syntax tests should be duplicated to `import.defer` (both valid and invalid) (https://github.com/tc39/test262/pull/4374)

## Deferred namespace object

- [x] It is an object
- [x] It is not the same object as the non-deferred one
- [x] Importing twice gives the same object
    - [x] static/static
    - [x] static/dynamic
    - [x] dynamic/dynamic
    - [x] static from different files
- [x] import and import defer + exec only execute the module once
- [x] `@@toStringTag` (https://github.com/tc39/test262/pull/4360)

### Sync

- [x] `import defer` does not trigger evaluation
- [x] `import.defer` does not trigger evaluation (https://github.com/tc39/test262/pull/4375)
- [x] `[[Get]]` triggers evaluation
    - [x] with a known export name
    - [x] with a non-exported name
    - [x] does not trigger evaluation with a symbol
- [x] What operations trigger evaluation (https://github.com/tc39/test262/pull/4341)

### Errors

- [x] Property access re-throws the evaluation error
    - [x] On first access for not evaluated modules
    - [x] On second access for not evaluated modules
    - [x] On first access for already evaluated modules
- [x] Import defer of an errored module doesn't throw
- [x] `ReadyForSyncExecution`
    - [x] Cannot evaluate a module in its _evaluating_ state
    - [x] Cannot evaluate itself
    - [x] Cannot evaluate a module with a dependency in an _evaluating_ state
    - [x] The error when evaluating a module with two dependencies, where the first one is _linked_ and the second one is _evaluating_, happens before evaluating the first one
    - [x] A deferred module evaluates itself while being deferred-evaluated
    - [x] Cannot evaluate a module in its _evaluating-async_ state
    - [x] Cannot evaluate itself after a TLA
    - [x] Cannot evaluate a module with a dependency in an _evaluating-async_ state.
- [x] Syntax error are not deferred
- [x] Linking errors are not deferred

### Async

- [x] Async modules are not deferred
[...]

PR #4278, opened 2024-10-25
Add partial tests for `import defer`

nicolo-ribaudo:
Plan: https://github.com/tc39/test262/issues/4215 (I checked the checkbox for tests added by this PR)

This PR includes tests for:
- syntax
- Deferred namespace objects
- Behavior of synchronous and asynchronous deferred modules
- Errors for synchronous and asynchronous cases

I will open another PR (or push another commit here) for tests about async modules, since I still have to finish polishing it up.

There are a few details of the proposal that are still in flux because of some bugs, so this PR does not include tests for them:
- `Symbol.toStringTag` of deferred namespace objects
- `test/language/import/import-defer/evaluation-sync/evaluation-*` for the other object operations
- `import.defer`, which currently doesn't actually defer due to a problem with `.then`


takikawa:
I tried running these in my WebKit implementation and found a few minor issues in the tests, such as missing module flags. This commit has the fixes I made: https://github.com/takikawa/WebKit/commit/d536aa2a38a5f2e9be5a36fa111a269a6b9fbdce

After the fixes I had about 8 failures but at least some of these are due to bugs in my implementation, I think.

nicolo-ribaudo:
Thanks! I imported your patch.

> After the fixes I had about 8 failures but at least some of these are due to bugs in my implementation, I think.

If you could share the list, I can double-check that the tests are correct.

test/language/import/import-defer/syntax/invalid-defer-default-and-namespace copy.js
ptomato, line 1: This seems to be a duplicate of `invalid-default-and-defer-namespace.js`, was it added by mistake?

test/language/import/import-defer/syntax/valid-default-binding-named-defer.js
```diff
@@ -0,0 +1,28 @@
...
+    ImportedBinding
+
+  ImportedBinding :
+    BindingIdentifier[~Yield, +Await]
+
+flags: [module]
+features: [import-defer]
+---*/
+
+import defer from "./dep_FIXTURE.js";
```
ptomato, line 28: Do we have a test elsewhere (not in syntax) that ensures this syntax is treated as a sync default binding named `defer` and not as a deferred import? e.g. the imported module performs a side effect that is tested for.

test/language/import/import-defer/deferred-namespace-object/identity.js
```diff
@@ -0,0 +1,43 @@
...
+
+import defer * as nsDeferred1 from "./dep_FIXTURE.js";
+import defer * as nsDeferred2 from "./dep_FIXTURE.js";
+import { depDeferredNamespace as nsDeferred3 } from "./dep-defer-ns_FIXTURE.js";
+const nsDeferred4 = await import.defer("./dep_FIXTURE.js");
+
+assert(nsDeferred1 === nsDeferred2, "Deferred import of the same module twice gives the same object");
+assert(nsDeferred1 === nsDeferred3, "Deferred import of the same module twice from different files gives the same object");
+assert(nsDeferred1 === nsDeferred4, "Static and dynamic deferred import of the same module gives the same object");
+assert(nsDeferred1 !== nsEager, "Deferred namespaces are distinct from eager namespaces");
```
ptomato, line 43: Prefer `assert.sameValue` and `assert.notSameValue`, for better assertion messages for implementors. (This applies throughout the PR.)

test/language/import/import-defer/deferred-namespace-object/object-properties.js
```diff
@@ -0,0 +1,52 @@
+// Copyright (C) 2024 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-modulenamespacecreate
+description: >
+  `defer` is a valid name for default imports
```
ptomato, line 7: Wrong description?

test/language/import/import-defer/deferred-namespace-object/object-properties.js
```diff
@@ -0,0 +1,52 @@
...
+flags: [module]
+features: [import-defer]
+includes: [propertyHelper.js, deepEqual.js]
+---*/
+
+import defer * as ns from "./dep_FIXTURE.js";
+
+assert(typeof ns === "object", "Deferred namespaces are objects");
+
+assert(Reflect.isExtensible(ns) === false, "Deferred namespaces are not extensible");
```
ptomato, line 30: Either
```js
assert(!Reflect.isExtensible(ns), "Deferred namespaces are not extensible");
```
or
```js
assert.sameValue(Reflect.isExtensible(ns), false, "Deferred namespaces are not extensible");
```
is preferred. (This also applies throughout the PR.)

test/language/import/import-defer/deferred-namespace-object/object-properties.js
```diff
@@ -0,0 +1,52 @@
...
+assert(Reflect.setPrototypeOf(ns, null) === true, "Deferred namespaces' prototype can be 'set' to null");
+
+assert.throws(TypeError, () => Reflect.apply(ns, null, []), "Deferred namespaces are not callable");
+assert.throws(TypeError, () => Reflect.construct(ns, null, []), "Deferred namespaces are not constructable");
+
+assert.deepEqual(
+  Reflect.ownKeys(ns),
+  ["bar", "foo", Symbol.toStringTag],
+  "Deferred namespaces' keys are the exports sorted alphabetically, followed by @@toStringTag"
+);
```
ptomato, line 44: We are trying to phase out usage of deepEqual (see https://github.com/tc39/test262/issues/3476)
How about something like this?
```suggestion
assert.compareArray(
  Reflect.ownKeys(ns),
  ["bar", "foo", Symbol.toStringTag],
  "Deferred namespaces' keys are the exports sorted alphabetically, followed by @@toStringTag"
);
```
(You will need to add `includes: [compareArray.js]` to the front matter instead of `deepEqual.js`)

test/language/import/import-defer/errors/get-other-while-evaluating/main.js
```diff
@@ -0,0 +1,38 @@
+// Copyright (C) 2024 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-module-namespace-exotic-objects-get-p-receiver-EnsureDeferredNamespaceEvaluation
+description: >
+  Modules cannot try to trigger their own evaluation
```
ptomato, line 7: Wrong description?

test/language/import/import-defer/errors/module-throws/third-party-evaluation-after-defer-import.js
```diff
@@ -0,0 +1,35 @@
...
+      1. Return ThrowCompletion(_promise_.[[PromiseResult]]).
+    1. ...
+
+flags: [module]
+features: [import-defer, top-level-await]
+---*/
+
+import defer * as ns from "./throws_FIXTURE.js";
+
+await import("./throws_FIXTURE.js").catch(() => {});
```
ptomato, line 33: For my own understanding, the URIError should be thrown regardless of whether this line is present, right?

Or maybe not, since the testing plan says "import and import defer + exec only execute the module once"

In any case, you should probably have both a `.then` and a `.catch` callback, and assert that only one is executed.

ptomato (commented):
I haven't gotten most of the way through this, but here are some initial comments.

test/language/import/import-defer/errors/module-throws/third-party-evaluation-after-defer-import.js
nicolo-ribaudo: This test for the following scenario:
- you get a deferred namespace of a module that has not been evaluated yet
- somebody else (in this case the dynamic import, but in the real world it will probably be a static import somewhere in another file) evaluates it. The test doesn't care about the details of this evaluation, just that it happens
- you try to access the deferred namespace you originally got a hold of, and it re-throws the error cached by the other evaluation

In the plan, it's "Property access re-throws the evaluation error > On first access for already evaluated modules"

test/language/import/import-defer/syntax/invalid-defer-default-and-namespace copy.js
nicolo-ribaudo: It has `defer` in a different position. I just accidentally had `copy` in the filename, but the test is unique.

test/language/import/import-defer/syntax/valid-default-binding-named-defer.js
nicolo-ribaudo: No. Should I just add it in this file, given that we don't really have the concept of "syntax tests"?

nicolo-ribaudo:
I noticed that I cannot use `assert` in `_FIXTURE` files (https://github.com/tc39/test262/blob/main/CONTRIBUTING.md#rules-for-module-_fixture-files). It's a bit weird because they are running in the same global scope as the non-fixture files, but I pushed a commit to fix it anyway.

test/language/import/import-defer/syntax/valid-default-binding-named-defer.js
ptomato: Sure, that would be fine too. By "in syntax" I meant in this `syntax/` folder where the test is currently placed; if the test no longer purely tests syntax, you could move it, but I don't think it matters that much.

test/language/import/import-defer/evaluation-top-level-await/import-defer-transitive-async-module/main.js
```diff
@@ -0,0 +1,55 @@
...
+features: [import-defer, top-level-await]
+includes: [compareArray.js]
+---*/
+
+import "../setup_FIXTURE.js";
+import defer * as ns "./imports-tla_FIXTURE.js";
+
+assert.compareArray(globalThis.evaluations, ["tla start", "tla end"]);
+ns.x;
+assert.compareArray(globalThis.evaluations, ["imports-dla"]);
```
ptomato, line 53:
```suggestion
assert.compareArray(globalThis.evaluations, ["imports-tla"]);
```
Additionally, how do the previous contents of the array get erased? I don't see where that happens.

test/language/import/import-defer/evaluation-top-level-await/sync-dependency-of-deferred-async-module/main.js
```diff
@@ -0,0 +1,61 @@
...
+features: [import-defer, top-level-await]
+includes: [compareArray.js]
+---*/
+
+import "../setup_FIXTURE.js";
+import defer * as ns from "./imports-tla-with-dep_FIXTURE.js";
+
+assert.compareArray(
+  globalThis.evaluations,
+  ["dep", "tla-with-dep start", "tla-with-dep end"]
```
ptomato, line 53:
```suggestion
  ["dep start", "tla-with-dep start", "tla-with-dep end"]
```
?

test/language/import/import-defer/evaluation-top-level-await/sync-dependency-of-deferred-async-module/main.js
```diff
@@ -0,0 +1,61 @@
...
+);
+ns.x;
+assert.compareArray(
+  globalThis.evaluations,
+  ["dep", "tla-with-dep start", "tla-with-dep end", "imports-tla-with-dep"]
```
ptomato, line 58:
```suggestion
  ["dep start", "tla-with-dep start", "tla-with-dep end", "imports-tla-with-dep"]
```

test/language/import/import-defer/evaluation-top-level-await/sync-dependency-of-deferred-async-module/main.js
ptomato: I'm not sure I understand how we end up with this result. `dep start` and `tla-with-dep start` I get, because they are sync dependencies of `imports-tla-with-dep_FIXTURE.js`. But isn't `tla-with-dep end` only pushed after microtasks execute?

test/language/import/import-defer/evaluation-sync/evaluation-ignore-isExtensible.js
```diff
@@ -0,0 +1,24 @@
...
+
+import "./setup_FIXTURE.js";
+
+import defer * as ns1 from "./dep-1_FIXTURE.js";
+
+assert.sameValue(globalThis.evaluations.length, 0, "import defer does not trigger evaluation");
+
+Object.isExtensible(ns1);
+
+assert.sameValue(globalThis.evaluations.length, 0, "[[GetPrototypeOf]] does not trigger evaluation");
```
ptomato, line 24:
```suggestion
assert.sameValue(globalThis.evaluations.length, 0, "[[IsExtensible]] does not trigger evaluation");
```

test/language/import/import-defer/evaluation-sync/evaluation-ignore-preventExtensions.js
```diff
@@ -0,0 +1,24 @@
...
+
+import "./setup_FIXTURE.js";
+
+import defer * as ns1 from "./dep-1_FIXTURE.js";
+
+assert.sameValue(globalThis.evaluations.length, 0, "import defer does not trigger evaluation");
+
+Object.preventExtensions(ns1);
+
+assert.sameValue(globalThis.evaluations.length, 0, "[[GetPrototypeOf]] does not trigger evaluation");
```
ptomato, line 24:
```suggestion
assert.sameValue(globalThis.evaluations.length, 0, "[[PreventExtensions]] does not trigger evaluation");
```

test/language/import/import-defer/evaluation-sync/evaluation-ignore-setPrototypeOf.js
```diff
@@ -0,0 +1,25 @@
...
+import "./setup_FIXTURE.js";
+
+import defer * as ns1 from "./dep-1_FIXTURE.js";
+
+assert.sameValue(globalThis.evaluations.length, 0, "import defer does not trigger evaluation");
+
+Reflect.setPrototypeOf(ns1, null);
+Reflect.setPrototypeOf(ns1, {});
+
+assert.sameValue(globalThis.evaluations.length, 0, "[[GetPrototypeOf]] does not trigger evaluation");
```
ptomato, line 25:
```suggestion
assert.sameValue(globalThis.evaluations.length, 0, "[[SetPrototypeOf]] does not trigger evaluation");
```

ptomato (changes requested):
Thanks! In general these tests are pretty easy to read and understand.

There were some things that I think are errors in assertions, like the tags pushed to `globalThis.evaluations` in some cases, so if your implementation is passing these tests in their current state, you might want to check if that's correct.

test/language/import/import-defer/evaluation-top-level-await/import-defer-transitive-async-module/main.js
```diff
@@ -0,0 +1,55 @@
...
+        1. If _result_ does not contain _m_, append _m_ to _result_.
+    1. Return _result_.
+
+flags: [module, async]
+features: [import-defer, top-level-await]
+includes: [compareArray.js]
+---*/
+
+import "../setup_FIXTURE.js";
+import defer * as ns "./imports-tla_FIXTURE.js";
```
takikawa, line 49: This needs a `from` token I think

test/language/import/import-defer/evaluation-top-level-await/flattening-order/main.js
```diff
@@ -0,0 +1,77 @@
...
+    1. Return _result_.
+
+flags: [module, async]
+features: [import-defer, top-level-await]
+includes: [compareArray.js]
+---*/
+
+import "../setup_FIXTURE.js";
+
+import "./dep-1.js";
```
takikawa, line 50: I think these imports here need `"_FIXTURE"` inserted into the specifier

test/language/import/import-defer/errors/get-self-while-evaluating-async/main.js
```diff
@@ -0,0 +1,38 @@
...
+flags: [module, async]
+features: [import-defer, top-level-await]
+---*/
+
+import defer * as ns from "./dep_FIXTURE.js";
+
+assert(globalThis["error on ns.foo"] instanceof TypeError, "ns.foo while evaluating throws a TypeError");
+
+ns.foo;
+
```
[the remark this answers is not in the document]
nicolo-ribaudo: What is happening here is that `./dep_FIXTURE.js` is _not_ deferred, because it contains top-level await. So what is happening in this test is that:
- the deferred module is not actually deferred, so `dep_FIXTURE.js` starts executing and goes in its `evaluating` state
- it has access to a deferred namespace of itself
- once it reaches the `await`, the state changes to `evaluating-async`
- the test tries then to access a property from the deferred namespace while it's `evaluating-async` (which is what this test it testing). It should throw.
- `dep_FIXTURE.js` is done, and becomes `evaluated`
- `main.js` starts evaluating, and the error is already there
- `ns.foo` now works, because `ns` is `evaluated` and not `evaluating-async`
-

test/language/import/import-defer/errors/get-other-while-dep-evaluating/main.js
```diff
@@ -0,0 +1,41 @@
...
+    1. Return *true*.
+
+flags: [module]
+features: [import-defer]
+---*/
+
+import "./dep-1_FIXTURE.js";
+
+assert(globalThis["evaluating dep2.foo error"] instanceof TypeError, "evaluating dep2.foo throws a TypeError");
+assert(!globalThis["evaluating dep2.foo evaluates dep2"], "evaluating dep2.foo does not evaluate dep2");
```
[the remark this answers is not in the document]
nicolo-ribaudo: No, but I'll add the opposite assertion. `dep3` never gets evaluated in this test, because the `dep2.foo` access aborts before doing any evaluation, since `ReadyForSyncExecution(dep2)` is false (because it has a dependency that is being evaluated).

test/language/import/import-defer/errors/module-throws/trigger-evaluation.js
```diff
@@ -0,0 +1,34 @@
...
+    1. ...
+
+flags: [module]
+features: [import-defer]
+---*/
+
+import defer * as ns from "./throws_FIXTURE.js";
+
+assert.throws(URIError, () => ns.foo, "Evaluation errors are thrown when evaluating");
+assert.throws(URIError, () => ns.foo, "Evaluation errors are thrown for already evaluated modules");
```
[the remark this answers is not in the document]
nicolo-ribaudo: I'll throw a specific object rather than a URIError (that I picked just because it was an uncommon error).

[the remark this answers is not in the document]
nicolo-ribaudo: I'll throw a specific object rather than a built-in error, to avoid any possible confusion.

test/language/import/import-defer/evaluation-top-level-await/import-defer-transitive-async-module/main.js
nicolo-ribaudo: I'll add the original content to the beginning of this array.

test/language/import/import-defer/evaluation-top-level-await/sync-dependency-of-deferred-async-module/main.js
nicolo-ribaudo: Changed `dep start` to `dep` in `dep_FIXTURE`. As a rule, I'm using start/end for async modules (where the distinction is meaningful), and just the filename for sync modules.

> I'm not sure I understand how we end up with this result. `dep start` and `tla-with-dep start` I get, because they are sync dependencies of `imports-tla-with-dep_FIXTURE.js`. But isn't `tla-with-dep end` only pushed after microtasks execute?

Due to the TLA, this file is equivalent to
```js
import "./setup_FIXTURE.js";
import "./tla-with-dep_FIXTURE.js";
import defer * as ns from "./imports-tla-with-dep_FIXTURE.js";

assert.compareArray(
  globalThis.evaluations,
  ["dep", "tla-with-dep start", "tla-with-dep end"]
);
...
```

so `main.js` doesn't run until when `tla-with-dep_FIXTURE.js` is done running.

nicolo-ribaudo:
Thanks both for the reviews :) Comments addressed.

test/language/import/import-defer/evaluation-top-level-await/flattening-order/main.js
```diff
@@ -0,0 +1,77 @@
...
+
+flags: [module, async]
+features: [import-defer, top-level-await]
+includes: [compareArray.js]
+---*/
+
+import "./setup_FIXTURE.js";
+
+import "./dep-1_FIXTURE.js";
+import defer * as ns2 from "./dep-2_FIXTURE.js.js";
```
takikawa, line 51: Looks like there's a typo here and below in the file extension `.js.js`

test/language/import/import-defer/errors/module-throws/trigger-evaluation.js
```diff
@@ -0,0 +1,38 @@
...
+
+flags: [module]
+features: [import-defer]
+---*/
+
+import defer * as ns from "./throws_FIXTURE.js";
+
+let err1, err2;
+try { ns.foo } catch (e) { err1 = e };
+assert.deepEqual(err1, { someError: "the error from throws_FIXTURE" }, "Evaluation errors are thrown when evaluating");
```
takikawa, line 35: Looks like this needs the deepEqual include or be written to avoid it

test/language/import/import-defer/errors/module-throws/defer-import-after-evaluation.js
```diff
@@ -0,0 +1,46 @@
...
+    1. If _promise_.[[PromiseState]] is ~rejected~, then
+      1. Return ThrowCompletion(_promise_.[[PromiseResult]]).
+    1. ...
+
+flags: [module, async]
+features: [import-defer]
+includes: [asyncHelpers.js]
+---*/
+
+asyncText(async () => {
```
takikawa, line 32: Typo here `asyncTest` (also `deepEqual` below may need an include)

test/language/import/import-defer/errors/module-throws/third-party-evaluation-after-defer-import.js
```diff
@@ -0,0 +1,46 @@
...
+    1. ...
+
+flags: [module, async]
+features: [import-defer]
+includes: [asyncHelpers.js]
+---*/
+
+import defer * as ns from "./throws_FIXTURE.js";
+
+asyncText(async () => {
```
takikawa, line 34: Ditto typo on `asyncTest` (& `deepEqual`)

nicolo-ribaudo:
All fixed, thanks @takikawa!

test/language/import/import-defer/evaluation-top-level-await/flattening-order/dep-2.2.1_FIXTURE.js
```diff
@@ -0,0 +1,8 @@
+// Copyright (C) 2024 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+globalThis.evaluations.push("2.2 start");
```
takikawa, line 4: Should this file have non-TLA push of "2.2.1"?

test/language/import/import-defer/evaluation-top-level-await/flattening-order/dep-2.1.1-tla_FIXTURE.js
```diff
@@ -0,0 +1,4 @@
+// Copyright (C) 2024 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+globalThis.evaluations.push("2.1.1");
```
takikawa, line 4: IIUC this file is intended to have TLA and "start" and "end" items pushed.

test/language/import/import-defer/errors/get-other-while-evaluating-async/dep-1-tla_FIXTURE.js
```diff
@@ -0,0 +1,15 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+
+import { resolveFirst, resolveThird } from "./promises_FIXTURE.js";
+import "./dep-1.1_FIXTURE.js"
+
+await Promise.resolve();
+
+resolveFirst();
+
+await second;
```
takikawa, line 11: I think this also needs to be added to the specifiers in the named import above.

test/language/import/import-defer/errors/get-other-while-dep-evaluating-async/dep-1-tla_FIXTURE.js
```diff
@@ -0,0 +1,13 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+
+import { resolveFirst, resolveThird } from "./promises_FIXTURE.js";
+import "./dep-1.1_FIXTURE.js"
+
+await Promise.resolve();
+
+resolveFirst();
+
+await second;
```
takikawa, line 11: Also needs to be added to specifiers list above.

takikawa:
Aside from the `flattening-order` test which might need more adjustment, I ran the tests in JSC and saw 7 failures. I think these are likely all due to three existing issues with JSC's module loader (doesn't re-throw after exception during eval, module namespace object properties aren't writable, and certain cyclic module graphs cause debug assertions to fail) that aren't related to import defer.

test/language/import/import-defer/errors/module-throws/trigger-evaluation.js
ptomato: Could we instead write the test to avoid `deepEqual`? We are trying to avoid it in new tests, see #3476.
The thrown value could be a primitive, like `"the error from throws_FIXTURE"`, since we are not checking for object identity anyway.

takikawa:
> @takikawa You mentioned that the `flattening-order` test needs more adjustments, can you elaborate? I'll approve when that's resolved.

I think Nic made some fixes in https://github.com/tc39/test262/pull/4278/commits/2e0982e6ec3659f6afd64cf9e4c12d1d19c32f78 which was what I was referring to. But I haven't run it since that modification, so I can take another look at it

test/language/import/import-defer/errors/module-throws/trigger-evaluation.js
nicolo-ribaudo: I want to throw an object to test that it's the same, but I'll remove deepEqual

nicolo-ribaudo:
@ptomato handled all of your remaining comments

test/language/import/import-defer/evaluation-top-level-await/flattening-order/main.js
```diff
@@ -0,0 +1,77 @@
...
+  "1",
+  "2.1.1 start",
+  "2.2.1",
+  "2.2 start",
+  "3",
+  "5",
+  "2.1.1 end",
+  "2.2 end",
+  "4.1 start",
+  "4.1 end"
```
takikawa, line 66: IIUC I think this should also have `"5"` at the end, since that's a non-deferred import of a normal module.

takikawa:
I ran the flattening order test again, and these are the results I got

(formatted to make it easier to read)

```
[294! NEW FAIL test/language/import/import-defer/evaluation-top-level-await/flattening-order/main.js (module)
    Exception: Test262Error: 
      Actual [1, 2.1.1 start, 2.1.1 end, 2.2.1, 2.2 start, 2.2 end, 3, 4.1 start, 4.1 end, 5] and expected
             [1, 2.1.1 start, 2.2.1, 2.2 start, 3, 5, 2.1.1 end, 2.2 end, 4.1 start, 4.1 end] should have the same contents.
```

As commented above, I think `5` is genuinely missing in the test. I suspect the rest of the differences may be due to how JSC's module loader is implemented (in particular, how it runs async dependencies [here](https://github.com/WebKit/WebKit/blob/main/Source/JavaScriptCore/builtins/ModuleLoader.js#L511)); basically the TLA doesn't seem to yield to the other imports. Not 100% sure though.

Ms2ger (approved):

PR #4285, opened 2024-10-29
402: Replace currency-digits.js Intl.NumberFormat test

ben-allen:
This test treats ISO 4217 data on the number of minor unit digits to use when formatting currencies. ISO 4217 is no longer normative as of https://github.com/tc39/ecma402/pull/922

Additionally, web reality before that PR is that some implementations already used CLDR data rather than ISO 4217.

I could write a test that only checks the digits on a few currencies for which CLDR and ISO 4217 agree, and which aren't likely to change any time soon. (i.e. USD and EUR using 2 minor digits, JPY using 0). Not sure if that's the right path, though -- any guidance welcome.

### DIFF
--- a/test/intl402/NumberFormat/currency-digits.js
+++ b/test/intl402/NumberFormat/currency-digits.js
@@ -17,7 +17,7 @@ author: Ben Allen
 ---*/
 
 const nf = Intl.NumberFormat([], {style: "currency", currency: "USD"});
-const max = nf.resolvedOptions.maximumFractionDigits;
-const min = nf.resolvedOptions.minimumFractionDigits;
+const max = nf.resolvedOptions().maximumFractionDigits;
+const min = nf.resolvedOptions().minimumFractionDigits;
 
-assert.sameValue(min === max, true, "Currency data not used; maximumFractionDigits does not match minimumFractionDigits");
+assert.sameValue(min, max, "Currency data not used; maximumFractionDigits should match minimumFractionDigits");
--- a/test/intl402/NumberFormat/currency-digits.js
+++ b/test/intl402/NumberFormat/currency-digits.js
@@ -1,189 +1,23 @@
-// Copyright 2011-2012 Norbert Lindenberg. All rights reserved.
-// Copyright 2012 Mozilla Corporation. All rights reserved.
+// Copyright 2024 Igalia S.L. All rights reserved.
 // This code is governed by the BSD license found in the LICENSE file.
 
 /*---
-es5id: 11.1.1_20_c
+esid: sec-intl.numberformat
 description: >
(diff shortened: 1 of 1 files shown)


Ms2ger:
Hmm, there's no other way to test those code paths, is there? I guess strictly speaking this wouldn't be in scope for test262, but I might be convinced to keep it.

ben-allen:
Sudden thought: If `CurrencyDigits` is being used to set `mininumFractionDigits` and `maximumFractionDigits`, then those two values will be identical, which isn't the case if we're using the defaults for these properties. As such it's possible to test for whether currency data is being used to set these properties without testing the specific values. Going to put together a test for this in a second

test/intl402/NumberFormat/currency-digits.js
```diff
@@ -1,189 +1,23 @@
...
-    XOF: 0,
-    XPF: 0,
-    YER: 2,
-    ZAR: 2,
-    ZMW: 2,
-    ZWL: 2,
-};
+const nf = Intl.NumberFormat([], {style: "currency", currency: "USD"});
+const max = nf.resolvedOptions.maximumFractionDigits;
+const min = nf.resolvedOptions.minimumFractionDigits;
```
ptomato, line 21:
```suggestion
const max = nf.resolvedOptions().maximumFractionDigits;
const min = nf.resolvedOptions().minimumFractionDigits;
```

test/intl402/NumberFormat/currency-digits.js
```diff
@@ -1,189 +1,23 @@
...
 
-Object.getOwnPropertyNames(currencyDigits).forEach(function (currency) {
-    var digits = currencyDigits[currency];
-    var format = Intl.NumberFormat([], {style: "currency", currency: currency});
-    var min = format.resolvedOptions().minimumFractionDigits;
-    var max = format.resolvedOptions().maximumFractionDigits;
-    assert.sameValue(min, digits, "Didn't get correct minimumFractionDigits for currency " + currency + ".");
-    assert.sameValue(max, digits, "Didn't get correct maximumFractionDigits for currency " + currency + ".");
-});
+assert.sameValue(min === max, true, "Currency data not used; maximumFractionDigits does not match minimumFractionDigits");
```
ptomato, line 23:
```suggestion
assert.sameValue(min, max, "Currency data not used; maximumFractionDigits does not match minimumFractionDigits");
```
(and while we're editing this line anyway; prefer to phrase assertion messages using "should", to avoid the ambiguity of whether it is describing the situation where the assertion holds or fails)

ptomato (approved):
LGTM modulo a few minor comments. This is a nice improvement in that test.

ptomato (approved):
Thanks!

PR #4286, opened 2024-10-29
Temporal: Move tests out of staging/Temporal/Regex/old and into test/…

catamorphism:
…built-ins


ptomato (commented):
I think I want to put this one on hold until I've worked through all the other open PRs. Looking over it now, I noticed some weirdness with strings that seem like they shouldn't parse correctly, so I'm not certain these tests are correct in the first place. That probably needs some digging into.
We also might want to expand it to cover all entry points that parse strings, not just `from`.

PR #4289, opened 2024-10-29
Temporal: Move more ZonedDateTime tests out of staging

catamorphism:


test/built-ins/Temporal/ZonedDateTime/prototype/add/overflow.js
ptomato, line 1: This one should be split into `add/overflow.js` and `subtract/overflow.js`. The `add` and `subtract` parts of the test are pretty much independent from each other.

test/built-ins/Temporal/ZonedDateTime/prototype/equals/casts-argument.js
```diff
@@ -0,0 +1,19 @@
+// Copyright (C) 2021 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.zoneddatetime.prototype.equals
+description: Different time zones not equal.
```
ptomato, line 6: Wrong description

ptomato (changes requested):
Thanks, one minor comment and one further split

test/built-ins/Temporal/ZonedDateTime/prototype/add/overflow.js
catamorphism: Done

test/built-ins/Temporal/ZonedDateTime/prototype/equals/casts-argument.js
catamorphism: Fixed

ptomato (approved):
Thanks!

PR #4294, opened 2024-10-30
Temporal: Split up ZonedDateTime/old/string-parsing.js and move out of staging

catamorphism:


test/built-ins/Temporal/ZonedDateTime/from/reject-throws-if-offset-does-not-match-iana-time-zone.js
ptomato, line 1: This one should probably be combined with `offset-does-not-match-iana-time-zone.js` which you ported in one of your earlier PRs.

ptomato (changes requested):
Naming nitpick: Let's prefix the ones specifically about string parsing with `argument-string-`, to match the existing string parsing tests in this folder. e.g. `decimal-places.js` → `argument-string-decimal-places.js`.

We should also expand these to cover all API entry points that parse a string, but I'll add that to the list of followups.

test/built-ins/Temporal/ZonedDateTime/from/reject-throws-if-offset-does-not-match-iana-time-zone.js
catamorphism: Done.

ptomato (approved):
LGTM, thanks!

PR #4295, opened 2024-10-30
Temporal: Move more ZonedDateTime tests out of staging into separate files under test/built-ins

catamorphism:
Move tests for subtract, toInstant, toPlainDate, toPlainTime, and toString


catamorphism:
I kept the calls to `Temporal.ZonedDateTime.from` and `Temporal.Instant.from` here for readability, but lmk if you'd rather I change them to explicitly construct the times/instants.

ptomato:
> I kept the calls to `Temporal.ZonedDateTime.from` and `Temporal.Instant.from` here for readability, but lmk if you'd rather I change them to explicitly construct the times/instants.

If they are good for readability it's fine. For me that justifies bringing in the additional code paths.

ptomato (approved):
LGTM, thanks!

PR #4300, opened 2024-10-31
Temporal: Add test for setting minimum valid year-month with non-ISO8601 calendar

catamorphism:
See https://github.com/tc39/proposal-temporal/issues/2998

### DIFF
--- a/test/intl402/Temporal/PlainYearMonth/prototype/with/minimum-valid-year-month.js
+++ b/test/intl402/Temporal/PlainYearMonth/prototype/with/minimum-valid-year-month.js
@@ -1,0 +1,14 @@
+// Copyright (C) 2024 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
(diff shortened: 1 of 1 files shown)


test/built-ins/Temporal/PlainYearMonth/prototype/with/minimum-valid-year-month.js
ptomato, line 1: Non-ISO calendar tests need to go in `test/intl402/Temporal/...` for the time being. (Technically, implementations without ECMA-402 can still support non-ISO calendars, but we haven't figured out a way to express that in the front matter yet.)

ptomato (approved):
LGTM, just needs to live in a different folder.

test/built-ins/Temporal/PlainYearMonth/prototype/with/minimum-valid-year-month.js
catamorphism: Done

PR #4301, opened 2024-10-31
Temporal: Add test for date with offset transition

catamorphism:
See https://github.com/tc39/proposal-temporal/issues/2938


test/built-ins/Temporal/ZonedDateTime/from/same-date-starts-twice.js
```diff
@@ -0,0 +1,27 @@
+// Copyright (C) 2021 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.zoneddatetime.from
+description: Handles dates with offset transitions.
```
ptomato, line 6:
```suggestion
description: Handles dates with offset transitions where midnight occurs twice.
```

test/built-ins/Temporal/ZonedDateTime/from/same-date-starts-twice.js
```diff
@@ -0,0 +1,27 @@
+// Copyright (C) 2021 Igalia, S.L. All rights reserved.
```
ptomato, line 1:
```suggestion
// Copyright (C) 2024 Igalia, S.L. All rights reserved.
```

test/built-ins/Temporal/ZonedDateTime/from/same-date-starts-twice.js
ptomato, line 1: I think this one should probably be split into two tests, in `startOfDay/` and `hoursInDay/` respectively. Since those are really the methods being tested, not `from`.

Additionally, tests that rely on named time zone definitions need to live in `test/intl402/Temporal/...` for the time being. Similar to non-ISO calendars, although it's possible for implementations without ECMA-402 to support named time zones, it's not required, and we don't have a way to indicate that in the front matter yet.

ptomato (changes requested):
Thanks!

test/built-ins/Temporal/ZonedDateTime/from/same-date-starts-twice.js
catamorphism: Done

test/intl402/Temporal/ZonedDateTime/prototype/hoursInDay/same-date-starts-twice.js
```diff
@@ -0,0 +1,20 @@
+// Copyright (C) 2024 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.zoneddatetime.from
```
ptomato, line 5:
```suggestion
esid: sec-temporal.zoneddatetime.prototype.hoursinday
```

test/intl402/Temporal/ZonedDateTime/prototype/startOfDay/same-date-starts-twice.js
```diff
@@ -0,0 +1,23 @@
+// Copyright (C) 2024 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.zoneddatetime.from
```
ptomato, line 5:
```suggestion
esid: sec-temporal.zoneddatetime.prototype.startofday
```

ptomato (approved):
ESIDs need updating to match the new locations of the tests, otherise LGTM. Thanks!

PR #4302, opened 2024-11-01
Add tests for adding negative durations to/subtracting durations from the last representable year/month of the ISO calendar

catamorphism:
See https://github.com/tc39/proposal-temporal/issues/3029


ptomato (approved):
Thanks!

PR #4303, opened 2024-11-02
Import and modernize bocoup/test262-regexp-generator

ptomato:
https://github.com/bocoup/test262-regexp-generator was used to generate the tests in `test/built-ins/RegExp/CharacterClassEscapes/`. It makes sense for this script to live in the same repo as the tests it generates. Especially since in the meantime, the tests themselves have gotten out of sync with the upstream.

This imports the script, modernizes and cleans up the code, implements the changes necessary to match the changes made in the meantime to the tests, and integrates the script into the build process.

### DIFF
--- a/tools/regexp-generator/index.mjs
+++ b/tools/regexp-generator/index.mjs
@@ -1,7 +1,8 @@
 import filenamify from 'filenamify';
 import fs from 'node:fs';
-import jsesc from 'jsesc';
+import regenerate from 'regenerate';
 import rewritePattern from 'regexpu-core';
+import ESCAPE_SETS from 'regexpu-core/data/character-class-escape-sets.js';
 import slugify from 'slugify';
 
 import header from './header.mjs';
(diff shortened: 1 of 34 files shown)


ptomato:
Unfortunately, `ci_build.sh` fails due to my changes to `make.py` but that does not result in the CI job failing. I'll sort that out later.

tools/regexp-generator/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "test262-regexp-class-escapes",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.mjs",
```
mathiasbynens, line 5: Since this is not meant to be published, we could add

```json
"private": true
```

and remove unnecessary fields:

```suggestion
  "private": true,
  "main": "index.mjs",
```

mathiasbynens (approved):
Thanks for your work on this!

tools/regexp-generator/package.json
ljharb: i believe name and version are always required, even for private packages

mathiasbynens: Nope, they’re only required when you want to publish the package, which isn’t the case here

tools/regexp-generator/index.mjs
```diff
@@ -15,30 +16,85 @@ const patterns = {
...
+	const rest = escapeData.clone().remove(LOW_SURROGATES);
+	const [ lowLoneCodePoints, lowRanges ] = toTestData(lowSurrogates);
+	const [ loneCodePoints, ranges ] = toTestData(rest);
+	loneCodePoints.unshift(...lowLoneCodePoints);
+	ranges.unshift(...lowRanges);
+	return prettyPrint([ loneCodePoints, ranges ]);
+}
+
 function buildContent(desc, pattern, range, max, flags, skip180e) {
+	let string = buildString(pattern[1], flags);
```
Ms2ger, line 81: Indentation looks off in this commit (Implement downstream changes). Tabs v spaces?

Ms2ger (approved):
lgtm modulo other comments

ptomato:
Rather than mess with the Circle-CI images to get one that has both Python and Node installed, I'll migrate the linter and test generation jobs to GitHub Actions where they can run on the `ubuntu-latest` image which has everything already. I'll do that in a separate PR, and hopefully make some improvements along the way. Let's consider this PR on hold for the moment.

ptomato:
With CI set up to handle npm dependencies, I've rebased this and will land it now.

PR #4305, opened 2024-11-04
Temporal: Add test for balancing up to weeks when year/month are present

catamorphism:
See https://github.com/tc39/proposal-temporal/issues/2813

### DIFF
--- a/test/built-ins/Temporal/Duration/prototype/round/balances-up-to-weeks.js
+++ b/test/built-ins/Temporal/Duration/prototype/round/balances-up-to-weeks.js
@@ -3,7 +3,7 @@
 
 /*---
 esid: sec-temporal.duration.prototype.round
-description: Balances up to weeks correctly when years and months are present.
+description: Rounds up to weeks correctly when years and months are present.
 includes: [temporalHelpers.js]
 features: [Temporal]
 ---*/
--- a/test/built-ins/Temporal/Duration/prototype/round/balances-up-to-weeks.js
+++ b/test/built-ins/Temporal/Duration/prototype/round/balances-up-to-weeks.js
@@ -1,0 +1,52 @@
+// Copyright (C) 2018 Bloomberg LP. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.duration.prototype.round
+description: Balances up to weeks correctly when years and months are present.
+includes: [temporalHelpers.js]
+features: [Temporal]
+---*/
+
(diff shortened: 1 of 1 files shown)


test/built-ins/Temporal/Duration/prototype/round/balances-up-to-weeks.js
```diff
@@ -0,0 +1,52 @@
+// Copyright (C) 2018 Bloomberg LP. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.duration.prototype.round
+description: Balances up to weeks correctly when years and months are present.
```
ptomato, line 6: Nitpick about the terminology:
```suggestion
description: Rounds up to weeks correctly when years and months are present.
```
I'd suggest renaming the test file as well.

(Rounding is when the duration is modified, balancing is when the duration remains the same length but is expressed in different units)

test/built-ins/Temporal/Duration/prototype/round/balances-up-to-weeks.js
```diff
@@ -0,0 +1,52 @@
...
+includes: [temporalHelpers.js]
+features: [Temporal]
+---*/
+
+const oneMonthOneDay = new Temporal.Duration(0, 1, 0, 1, 0, 0, 0, 0, 0, 0);
+const oneYearOneMonthOneDay = new Temporal.Duration(1, 1, 0, 1, 0, 0, 0, 0, 0, 0);
```
ptomato, line 12: I'd suggest something like
```js
const relativeTo = new Temporal.PlainDate(2024, 1, 1);
```
and using that in the options bags throughout.

ptomato: Also, I've been looking at the existing tests in `round/` and I think we might not have coverage of rounding a duration with only days up to weeks. So maybe add something like
```js
const severalWeeksInDays = new Temporal.Duration(0, 0, 0, 29, 0, 0, 0, 0, 0, 0);
```
(I might have missed it if we already had coverage of this, though! There are a lot of tests in this folder)

ptomato (changes requested):
Thanks!

test/built-ins/Temporal/Duration/prototype/round/balances-up-to-weeks.js
catamorphism: Done and done.

ptomato (approved):
Thanks!

PR #4306, opened 2024-11-05
Temporal: Add test case for rounding Duration relative to a ZonedDateTime

catamorphism:
This is the test for the change in https://github.com/tc39/proposal-temporal/pull/3036 and is a variation of the test from https://github.com/tc39/proposal-temporal/issues/2814 .

### DIFF
--- a/test/intl402/Temporal/Duration/prototype/round/rounding-with-largestunit.js
+++ b/test/intl402/Temporal/Duration/prototype/round/rounding-with-largestunit.js
@@ -9,8 +9,8 @@ features: [Temporal]
 
 // Based on a test case by Adam Shaw
 
-const dur = Temporal.Duration.from({"hours": 13});
-const zdt = Temporal.ZonedDateTime.from('2024-03-10T00:00:00[America/New_York]');
+const dur = new Temporal.Duration.from(0, 0, 0, 0, /* hours = */ 13, 0, 0, 0, 0, 0);
+const zdt = new Temporal.ZonedDateTime(0n, "UTC");
 
 TemporalHelpers.assertDuration(
     dur.round({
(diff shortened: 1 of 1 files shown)


test/intl402/Temporal/Duration/prototype/round/rounding-with-largestunit.js
```diff
@@ -0,0 +1,22 @@
...
+esid: sec-temporal.duration.prototype.round
+description: Rounding mode is taken into account when largestUnit is present.
+includes: [temporalHelpers.js]
+features: [Temporal]
+---*/
+
+// Based on a test case by Adam Shaw
+
+const dur = Temporal.Duration.from({"hours": 13});
+const zdt = Temporal.ZonedDateTime.from('2024-03-10T00:00:00[America/New_York]');
```
ptomato, line 13:
```suggestion
const dur = new Temporal.Duration.from(0, 0, 0, 0, /* hours = */ 13, 0, 0, 0, 0, 0);
const zdt = new Temporal.ZonedDateTime(0n, "UTC");
```

(I think the actual contents of the ZonedDateTime don't matter here, so let's just replace it with epoch time 0 and UTC. We should avoid requiring named time zones in non-Intl tests for the time being.)

ptomato (approved):
Thanks!

test/intl402/Temporal/Duration/prototype/round/rounding-with-largestunit.js
ptomato: I verified that `0n, "UTC"` still exercises the same code path that had the implementation bug, so I'll go ahead and merge it with this modification.

PR #4309, opened 2024-11-05
Allow alternative era for Chinese calendar

anba:
Year 1 in the Chinese calendar corresponds to 2637 BCE in ICU4X.

See <https://docs.rs/icu/latest/icu/calendar/chinese/struct.Chinese.html#year-and-era-codes>.

### DIFF
--- a/test/intl402/Temporal/PlainDate/prototype/until/until-across-lunisolar-leap-months.js
+++ b/test/intl402/Temporal/PlainDate/prototype/until/until-across-lunisolar-leap-months.js
@@ -9,8 +9,18 @@ features: [Temporal]
 
 // 2001 is a leap year in the Chinese calendar with a M04L leap month.
 // Therefore, month: 6 is M05 in 2001 but M06 in 2000 which is not a leap year.
-const one = Temporal.PlainDate.from({ year: 2000, month: 6, day: 1, calendar: 'chinese' });
-const two = Temporal.PlainDate.from({ year: 2001, month: 6, day: 1, calendar: 'chinese' });
+
+const year2000 = new Temporal.PlainDate(2000, 3, 1).withCalendar("chinese").year;
+const year2001 = new Temporal.PlainDate(2001, 3, 1).withCalendar("chinese").year;
+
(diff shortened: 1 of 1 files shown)


ptomato:
Similar to #4101, I'm hesitant to modify the tests to allow both results here. Either the ICU4C result is correct, or the ICU4X result is. That's something that ICU4X needs to sort out.

anba:
This is different from #4101, because the different epoch year isn't a ICU4C vs. ICU4X difference, but a design decision from the Temporal polyfill. The Temporal polyfill formats dates using `Intl.DateTimeFormat`, which outputs in many cases the [related Gregorian year](https://unicode.org/reports/tr35/tr35-dates.html#dfst-year), cf. `new Intl.DateTimeFormat("en-u-ca-chinese").formatToParts(new Date)`. 

Reading <https://github.com/unicode-org/icu/blob/0357501948d2f0ab43c891f446e68f19b07b442d/icu4c/source/i18n/chnsecal.h#L52-L60>, it seems like ICU4C and ICU4X actually agree on 2637 BCE as the initial epoch year. 

And a similar modification was already allowed in #4079.

ptomato:
OK, thanks, got it. If ICU4C and ICU4X both agree on the result, then my comment doesn't apply.

ptomato:
Coming back to this. If there is one correct answer that ICU4C and ICU4X agree on, could we make the test unambiguous so that it requires the correct answer?

anba:
I don't think there's one clear answer for this. For example see:
- https://web.archive.org/web/20160821170722if_/http://www.math.nus.edu.sg/~mathelmr/calendar/cal.pdf, §4.11 What year is it in the Chinese calendar?
- https://en.wikipedia.org/wiki/Chinese_calendar#Epochs
- https://en.wikipedia.org/wiki/Adoption_of_the_Gregorian_calendar#China_and_Taiwan
- https://en.wikipedia.org/wiki/Sexagenary_cycle

ptomato (approved):
Hmmm, all right then. Better to have this than not have it.

PR #4312, opened 2024-11-05
Temporal: Add test case for rounding up to a non-24-hour ZonedDateTime

catamorphism:
See https://github.com/tc39/proposal-temporal/issues/2814


ptomato (approved):
Thanks!

PR #4318, opened 2024-11-07
Disable grouping for numeric units in Intl.DurationFormat

anba:
Cf. `FormatNumeric{Hours,Minutes,Seconds}` in <https://tc39.es/proposal-intl-duration-format/>.


ptomato (approved):
Thanks!

PR #4320, opened 2024-11-07
Temporal: Fix name of constructor

ptomato:
I mistakenly committed something slightly different in my code review suggestion for #4306 than what I had modified locally.


ctcpip (approved):
you need a 262 maintainer to review, but LGTM

Ms2ger (approved):

PR #4321, opened 2024-11-10
Add additional test cases for RegExp modifiers

graphemecluster:
This is motivated by a bug discovered in both [V8](https://issues.chromium.org/issues/377820802) and [SpiderMonkey](https://bugzilla.mozilla.org/show_bug.cgi?id=1929939).
Follow up of #3960.


graphemecluster:
CC @rbuckton and @JLHwung (who pushed me to report this bug to the engines).

test/built-ins/RegExp/regexp-modifiers/nesting-remove-ignoreCase-within-add-ignoreCase.js
graphemecluster, line 1: This file is originally identical to `nesting-add-ignoreCase-within-remove-ignoreCase.js`, which is probably unintended.

JLHwung (approved):
LGTM, thank you for providing new tests.

Ms2ger (approved):
Thanks

PR #4322, opened 2024-11-12
Error.isError: Add test for DOMException

lionel-rowe:
Given that `DOMException`s diverge in certain ways from other types of error (notably not having an `[[ErrorData]]` internal slot as of current Web IDL spec, pending https://github.com/whatwg/webidl/pull/1421), it seems prudent to include a test for this case.

### DIFF
--- a/test/built-ins/Error/isError/dom-exception.js
+++ b/test/built-ins/Error/isError/dom-exception.js
@@ -1,0 +1,13 @@
+// Copyright 2024 Lionel Rowe. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-error.iserror
+description: >
+  Returns true on DOMException instances
+features: [Error.isError]
+---*/
+
+if (typeof DOMException === 'function') {
+  assert.sameValue(Error.isError(new DOMException()), true);
+}


ljharb (approved):

linusg:
I feel like this belongs in WPT, no? It's testing a part of the WebIDL spec.

lionel-rowe:
> I feel like this belongs in WPT, no? It's testing a part of the WebIDL spec.

Possibly, I'm new to this process. I guess it depends whether you consider the subject-under-test to be the `Error.isError` method or its `DOMException` argument.

Ms2ger:
I agree that wpt makes more sense here

ljharb:
yeah, that's a fair point, i approved too quickly. this definitely belongs in WPT.

nicolo-ribaudo:
> Possibly, I'm new to this process. I guess it depends whether you consider the subject-under-test to be the `Error.isError` method or its `DOMException` argument.

@lionel-rowe To clarify why they are asking to move it to WPT, the rule for these two test suites is that they should only be using features present on the target platforms that they are testing.

WPT is tests for the web platform, while test262 is tests for ECMA-262-compliant platforms. The feature set of the web platform is a superset of the feature set of ECMA-262-compliant platforms: for example, DOMException is only available on the web platform and not in all ECMA-262 platforms.

test262 tests can only use ECMA-262 features, and WPT tests can only use web platform features (which include all ECMA-262 features).

Since this test is testing the interaction of an ECMA-262 feature with a web-platform-but-not-ECMA-262 feature, it should go in WPT.

lionel-rowe:
Thanks for the explanation @nicolo-ribaudo ! I'll close this and probably open a corresponding PR against WPT sometime (can't promise when though).

PR #4323, opened 2024-11-12
Fix the not-a-constructor test for Math.f16round()

rmg-x:
Previously, this was using Math.fround() instead of the f16 version.


Ms2ger (approved):
Thanks!

PR #4326, opened 2024-11-14
Add initial tests for the "iterator-sequencing" proposal

anba:
Tests are based on existing tests in "test/built-ins/Iterator".

### DIFF
--- a/test/built-ins/Iterator/concat/fresh-iterator-result.js
+++ b/test/built-ins/Iterator/concat/fresh-iterator-result.js
@@ -1,0 +1,49 @@
+// Copyright (C) 2024 André Bargull. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-iterator.concat
+description: >
+  Returns a fresh iterator result object
+info: |
+  Iterator.concat ( ...items )
+
+  ...
+  3. Let closure be a new Abstract Closure with no parameters that captures iterables and performs the following steps when called:
+    a. For each Record iterable of iterables, do
+      ...
+      v. Repeat, while innerAlive is true,
+        1. Let innerValue be ? IteratorStepValue(iteratorRecord).
+        2. If innerValue is done, then
+          ...
+        3. Else,
+          a. Let completion be Completion(Yield(innerValue)).
+    ...
+features: [iterator-sequencing]
+---*/
+
+let oldIterResult = {
+  done: false,
+  value: 123,
+};
+
+let testIterator = {
+  next() {
+    return oldIterResult;
+  }
+};
+
+let iterable = {
+  [Symbol.iterator]() {
+    return testIterator;
+  }
+};
+
+let iterator = Iterator.concat(iterable);
+
+let iterResult = iterator.next();
+
+assert.sameValue(iterResult.done, false);
+assert.sameValue(iterResult.value, 123);
+
+assert.notSameValue(iterResult, oldIterResult);
--- a/test/built-ins/Iterator/concat/fresh-iterator-result.js
+++ b/test/built-ins/Iterator/concat/fresh-iterator-result.js
@@ -46,4 +46,4 @@ let iterResult = iterator.next();
 assert.sameValue(iterResult.done, false);
 assert.sameValue(iterResult.value, 123);
 
-assert.notSameValue(iterResult, oldIterResult);
+assert.sameValue(iterResult, oldIterResult);
--- a/test/built-ins/Iterator/concat/many-arguments.js
+++ b/test/built-ins/Iterator/concat/many-arguments.js
@@ -1,0 +1,45 @@
+// Copyright (C) 2024 André Bargull. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-iterator.concat
+description: >
+  Iterator.concat when called with many arguments.
+info: |
+  Iterator.concat ( ...items )
+
+  1. Let iterables be a new empty List.
+  2. For each element item of items, do
+    ...
+  3. Let closure be a new Abstract Closure with no parameters that captures iterables and performs the following steps when called:
+    a. For each Record iterable of iterables, do
+      ...
+    b. Return ReturnCompletion(undefined).
+  ...
+  6. Return gen.
+features: [iterator-sequencing]
+---*/
+
+let iterables = [
+  [/* empty */],
+  [1],
+  [2, 3],
+  [4, 5, 6],
+  [7, 8, 9, 10],
+];
+
+let iterator = Iterator.concat(...iterables);
+
+let array = [].concat(...iterables);
+
+for (let i = 0; i < array.length; i++) {
+  let iterResult = iterator.next();
+
+  assert.sameValue(iterResult.done, false);
+  assert.sameValue(iterResult.value, array[i]);
+}
+
+let iterResult = iterator.next();
+
+assert.sameValue(iterResult.done, true);
+assert.sameValue(iterResult.value, undefined);
--- a/test/built-ins/Iterator/concat/next-method-returns-non-object.js
+++ b/test/built-ins/Iterator/concat/next-method-returns-non-object.js
@@ -1,0 +1,37 @@
+// Copyright (C) 2024 André Bargull. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-iterator.concat
+description: >
+  Underlying iterator next returns non-object
+info: |
+  Iterator.concat ( ...items )
+
+  ...
+  3. Let closure be a new Abstract Closure with no parameters that captures iterables and performs the following steps when called:
+    a. For each Record iterable of iterables, do
+      ...
+      v. Repeat, while innerAlive is true,
+        1. Let innerValue be ? IteratorStepValue(iteratorRecord).
+        ...
+features: [iterator-sequencing]
+---*/
+
+let nonObjectIterator = {
+  next() {
+    return null;
+  }
+};
+
+let iterable = {
+  [Symbol.iterator]() {
+    return nonObjectIterator;
+  }
+};
+
+let iterator = Iterator.concat(iterable);
+
+assert.throws(TypeError, function() {
+  iterator.next();
+});
--- a/test/built-ins/Iterator/concat/next-method-returns-throwing-done.js
+++ b/test/built-ins/Iterator/concat/next-method-returns-throwing-done.js
@@ -1,0 +1,45 @@
+// Copyright (C) 2024 André Bargull. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-iterator.concat
+description: >
+  Underlying iterator next returns object with throwing done getter
+info: |
+  Iterator.concat ( ...items )
+
+  ...
+  3. Let closure be a new Abstract Closure with no parameters that captures iterables and performs the following steps when called:
+    a. For each Record iterable of iterables, do
+      ...
+      v. Repeat, while innerAlive is true,
+        1. Let innerValue be ? IteratorStepValue(iteratorRecord).
+        ...
+features: [iterator-sequencing]
+---*/
+
+let throwingIterator = {
+  next() {
+    return {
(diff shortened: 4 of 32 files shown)


test/built-ins/Iterator/concat/next-method-returns-non-object.js
```diff
@@ -0,0 +1,37 @@
...
+let iterable = {
+  [Symbol.iterator]() {
+    return nonObjectIterator;
+  }
+};
+
+let iterator = Iterator.concat(iterable);
+
+assert.throws(TypeError, function() {
+  iterator.next();
```
ljharb, line 36: in my polyfill, `.next().next()` throws instead. this is a reminder for myself

test/built-ins/Iterator/concat/next-method-returns-throwing-done.js
```diff
@@ -0,0 +1,45 @@
...
+let iterable = {
+  [Symbol.iterator]() {
+    return throwingIterator;
+  }
+};
+
+let iterator = Iterator.concat(iterable);
+
+assert.throws(Test262Error, function() {
+  iterator.next();
```
ljharb, line 44: here too

test/built-ins/Iterator/concat/next-method-returns-throwing-value.js
```diff
@@ -0,0 +1,45 @@
...
+let iterable = {
+  [Symbol.iterator]() {
+    return throwingIterator;
+  }
+};
+
+let iterator = Iterator.concat(iterable);
+
+assert.throws(Test262Error, function() {
+  iterator.next();
```
ljharb, line 44: and here

test/built-ins/Iterator/concat/next-method-throws.js
```diff
@@ -0,0 +1,37 @@
...
+let iterable = {
+  [Symbol.iterator]() {
+    return throwingIterator;
+  }
+};
+
+let iterator = Iterator.concat(iterable);
+
+assert.throws(Test262Error, function() {
+  iterator.next();
```
ljharb, line 36: and here

test/built-ins/Iterator/concat/return-is-forwarded.js
```diff
@@ -0,0 +1,58 @@
...
+let iterResult = iterator.next();
+assert.sameValue(returnCount, 0);
+assert.sameValue(iterResult.done, false);
+assert.sameValue(iterResult.value, 1);
+
+iterator.return();
+assert.sameValue(returnCount, 1);
+
+iterator.return();
+assert.sameValue(returnCount, 1);
```
ljharb, line 58: my polyfill reports 2 here

test/built-ins/Iterator/concat/throws-typeerror-when-generator-is-running-next.js
```diff
@@ -0,0 +1,47 @@
...
+    return testIterator;
+  }
+};
+
+let iterator = Iterator.concat(iterable);
+
+assert.sameValue(enterCount, 0);
+
+assert.throws(TypeError, function() {
+  iterator.next();
```
ljharb, line 44: also here

test/built-ins/Iterator/concat/throws-typeerror-when-generator-is-running-return.js
```diff
@@ -0,0 +1,53 @@
...
+
+iterator.next();
+
+assert.sameValue(enterCount, 0);
+
+assert.throws(TypeError, function() {
+  iterator.return();
+});
+
+assert.sameValue(enterCount, 1);
```
ljharb, line 53: lol this one blew the stack of my polyfill at 864, clearly i have a bug in this one

ljharb (commented):
Almost all of these tests pass verbatim on my polyfill, which is great!

I've commented the ones that don't - I assume they represent bugs in my polyfill, but I'll do a deep dive on the spec text to make sure.

test/built-ins/Iterator/concat/throws-typeerror-when-generator-is-running-return.js
```diff
@@ -0,0 +1,53 @@
...
+};
+
+let iterator = Iterator.concat(iterable);
+
+iterator.next();
+
+assert.sameValue(enterCount, 0);
+
+assert.throws(TypeError, function() {
+  iterator.return();
```
ljharb, line 50: fixed the other two issues, but now mine doesn't throw here.

michaelficarra: 1. this line calls `%IteratorHelperPrototype%.return`
1. `%IteratorHelperPrototype%.return` step 6 calls `GeneratorResumeAbrupt`
1. `GeneratorResumeAbrupt` step 8 sets `[[GeneratorState]]` to executing
1. `GeneratorResumeAbrupt` step 10 resumes the underlying generator
1. `Iterator.concat` step 3.a.v.3.a, Yield produces the return completion from `%IteratorHelperPrototype%.return` step 5
1. `Iterator.concat` step 3.a.v.3.b.i calls `IteratorClose`
1. `IteratorClose` step 4.c calls `testIterator.return`
1. line 32 calls `%IteratorHelperPrototype%.return`
1. `%IteratorHelperPrototype%.return` step 6 calls `GeneratorResumeAbrupt`
1. `GeneratorResumeAbrupt` step 1 calls `GeneratorValidate`
1. `GeneratorValidate` step 6 throws

anba: https://github.com/es-shims/iterator-helpers/blob/dd80b2f1718092219e94e24afdc1737f062e3df4/aos/GeneratorResumeAbrupt.js#L34-L37 seems to be missing to implement GeneratorResumeAbrupt, step 8 where `generator.[[GeneratorState]]` is set to `executing`.

ljharb: I've pushed up some local changes I had, so now: if i comment out/remove this (incorrect) block https://github.com/es-shims/iterator-helpers/blob/temp/aos/GeneratorResumeAbrupt.js#L39-L42

then i go from 1 failing test to 9 (from `node test`). I'm pretty convinced this is a polyfill problem, though, because I'm not using generators and I've made my own quasi-state machine that surely hits some edge cases.

test/built-ins/Iterator/concat/throws-typeerror-when-iterable-not-an-object.js
```diff
@@ -0,0 +1,44 @@
...
+/*---
+esid: sec-iterator.concat
+description: >
+  Throws a TypeError when the iterable is not an object.
+info: |
+  Iterator.concat ( ...items )
+
+  1. Let iterables be a new empty List.
+  2. For each element item of items, do
+    a. If item is not an Object, throw a TypeError exception.
```
michaelficarra, line 13: Relatedly, the primitive object wrappers (`new Boolean(true)`, etc) should pass this step. We should have a test for that.

test/built-ins/Iterator/concat/result-is-iterator.js
```diff
@@ -0,0 +1,13 @@
...
+/*---
+esid: sec-iterator.concat
+description: >
+  The value of the [[Prototype]] internal slot of the return value of Iterator.concat
+  is the intrinsic object %IteratorHelperPrototype%.
+features: [iterator-sequencing]
+---*/
+
+var iter = Iterator.concat();
+assert(iter instanceof Iterator, "Iterator.concat() must return an Iterator");
```
michaelficarra, line 13: We should make this same assertion when passed 1 iterator that doesn't inherit from Iterator. I could imagine an implementation trying to make an optimisation for the 1-iterator case where they pass the argument right through.

test/built-ins/Iterator/concat/many-arguments.js
```diff
@@ -0,0 +1,45 @@
...
+  [/* empty */],
+  [1],
+  [2, 3],
+  [4, 5, 6],
+  [7, 8, 9, 10],
+];
+
+let iterator = Iterator.concat(...iterables);
+
+let array = [].concat(...iterables);
```
michaelficarra, line 33: As the champion of `Array.prototype.flat`, it breaks my heart to see this.

```suggestion
let array = iterables.flat();
```

test/built-ins/Iterator/concat/fresh-iterator-result.js
```diff
@@ -0,0 +1,49 @@
...
+};
+
+let iterator = Iterator.concat(iterable);
+
+let iterResult = iterator.next();
+
+assert.sameValue(iterResult.done, false);
+assert.sameValue(iterResult.value, 123);
+
+assert.notSameValue(iterResult, oldIterResult);
```
michaelficarra, line 49: This seems unnecessarily wasteful. I've opened an issue to reconsider this behaviour: https://github.com/tc39/proposal-iterator-sequencing/issues/17

michaelficarra (approved):
LGTM otherwise! Thanks for writing these, @anba.

test/built-ins/Iterator/concat/many-arguments.js
ptomato: We try to avoid newer ES language features in tests if they're not the feature being tested - it's not a prohibition and many tests do it anyway, but we generally don't go out of our way to add them.

michaelficarra:
I pushed up a couple of commits. This LGTM now, modulo the normative question which we can resolve at the upcoming plenary.

test/built-ins/Iterator/concat/fresh-iterator-result.js
michaelficarra: See https://github.com/tc39/proposal-iterator-sequencing/pull/18#issuecomment-2512825730. I've updated the test.

ptomato (approved):
As discussed in the maintainers chat, merging this on the strength of Jordan's and Michael's reviews.

test/built-ins/Iterator/concat/return-is-not-forwarded-before-initial-start.js
```diff
@@ -0,0 +1,32 @@
...
+let iterable = {
+  [Symbol.iterator]() {
+    return testIterator;
+  }
+};
+
+let iterator = Iterator.concat(iterable);
+iterator.return();
+iterator.next();
+iterator.return();
```
bnoordhuis, line 32: Implementer here. I'm confused by this test, can someone explain?

I understand why the first return() should not forward, but the second one should, right? And that makes the test fail because the return method throws.

Is return-before-next a terminal state? But then next() should not be invoked either, it seems like? But if that's the case, I would expect its implementation in this file to throw.

anba: 1. `Iterator.concat` calls `CreateIteratorFromClosure` which in turn calls `GeneratorStart`. `[[GeneratorState]]` is still "suspended-start" at this point. This finishes initialising the iterator and control returns to the caller.
2. When now calling `%IteratorHelperPrototype%.return()`, `[[GeneratorState]]` is still "suspended-start" and we transition `[[GeneratorState]]` to "completed". `IteratorClose` is called for all opened underlying iterators, but at this point no iterator has yet been opened.
3. When then calling `%IteratorHelperPrototype%.next()`, nothing should happen, because `GeneratorResume` directly returns when `[[GeneratorState]]` is "completed".
4. And when calling `%IteratorHelperPrototype%.return()` after `next()`, still no user-code will be called, because `GeneratorResumeAbrupt` also directly returns when `[[GeneratorState]]` is "completed".

bnoordhuis: Thanks! Does that mean the next() method here could/should throw Test262Error, like the return() method does?

anba: Yes, `testIterator.next` can be changed to throw a `Test262Error` without changing the test semantics.

bnoordhuis: Thanks, I've opened https://github.com/tc39/test262/pull/4585 to make that more explicit.

PR #4333, opened 2024-11-25
Fix `single-line-html-close-first-line-*` failing with `Test262Error is not defined`

### LINKED ISSUE #4020: Tests for `-->` in the first line should not have `raw` flag.
The following tests added in https://github.com/tc39/test262/pull/4002 have the `raw` flag but are expected to throw a `Test262Error`:
```
test/annexB/language/comments/single-line-html-close-first-line-1.js
test/annexB/language/comments/single-line-html-close-first-line-2.js
test/annexB/language/comments/single-line-html-close-first-line-3.js
```
If executed without `harness/sta.js`, these tests will fail with `ReferenceError: Test262Error is not defined` instead. Therefore, the `raw` flag should be removed from these tests.

Boshen:
closes #4020

Updated the tests per https://github.com/tc39/test262/issues/4020#issuecomment-2009885892

### DIFF
--- a/test/annexB/language/comments/single-line-html-close-first-line-1.js
+++ b/test/annexB/language/comments/single-line-html-close-first-line-1.js
@@ -21,7 +21,7 @@ info: |
       WhiteSpaceSequence[opt] SingleLineDelimitedCommentSequence[opt] --> SingleLineCommentChars[opt]
 negative:
   phase: runtime
-  type: Error
+  type: EvalError
 ---*/
 
 // Because this test concerns the interpretation of non-executable character
@@ -31,4 +31,4 @@ negative:
 // Express the intended behavior by intentionally throwing an error; this
 // guarantees that test runners will only consider the test "passing" if
 // executable sequences are correctly interpreted as such.
-throw new Error("This is not in a comment");
+throw new EvalError("This is not in a comment");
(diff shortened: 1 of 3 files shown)


Boshen:
Discovered this test case while setting up runtime execution tests in Oxc.

ptomato (changes requested):
Thanks for the PR!

I think these will still fail because the `negative` key in the front matter says to expect a Test262Error. (see https://github.com/tc39/test262/blob/HEAD/CONTRIBUTING.md#negative)

I'd suggest throwing a built-in error type, preferably not SyntaxError since that might be accidentally thrown otherwise in the parse phase. Probably Error is fine.

Boshen:
Learned how to verify with test262-harness:

```
test262-harness  master ❯ ./bin/run.js --host-type node --host-path `which node` ~/github/test262/test/annexB/language/comments/*.js
Ran 13 tests
13 passed
0 failed
```

ljharb:
(fwiw in tests I love using EvalError for a builtin error type that isn't likely to come from anything but the test)

Boshen:
I changed to `EvalError`. Verified with `test262-harness`.

ptomato:
Thanks! EvalError works too.

ptomato (approved):

PR #4334, opened 2024-11-26
Fix AbstractModuleSource.prototype descriptor

legendecas:
The descriptor should be `{ [[Writable]]: false, [[Enumerable]]: false, [[Configurable]]: false }`.


Ms2ger (approved):
Thanks!

PR #4328, opened 2024-11-15
Test that Intl.DateTimeFormat canonicalizes the timezones "Etc/GMT", "Etc/UTC", and "GMT" to "UTC"

ben-allen:
fix #4327

PR #4336, opened 2024-12-04
Update UTC time zone canonicalization to match proposal-canonical-tz

ptomato:
The test for https://github.com/tc39/ecma402/pull/724 (added in https://github.com/tc39/test262/pull/4328) didn't take the Time Zone Canonicalization proposal into account; but it should, because that proposal is stage 3.

As of that proposal, the [[TimeZone]] slot of DateTimeFormat gets the case-regularized original identifier, no longer the primary identifier. So the resolvedOptions().timeZone property also no longer returns the primary identifier.

### DIFF
--- a/test/intl402/DateTimeFormat/canonicalize-utc-timezone.js
+++ b/test/intl402/DateTimeFormat/canonicalize-utc-timezone.js
@@ -2,25 +2,33 @@
 // This code is governed by the BSD license found in the LICENSE file.
 /*---
 esid: sec-createdatetimeformat
-description: Tests that the time zone names "Etc/UTC", "Etc/GMT", and "GMT" all resolve to "UTC".
+description: >
+  Tests that the time zone names "Etc/UTC", "Etc/GMT", and "GMT" are not
+  canonicalized to "UTC" in "resolvedOptions".
 info: |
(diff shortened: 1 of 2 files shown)


ptomato:
cc @ben-allen @justingrant

anba:
Related open issues:
- https://github.com/tc39/proposal-canonical-tz/issues/21
- https://github.com/tc39/proposal-canonical-tz/issues/2

justingrant (approved):
This seems fine, but should this test be relocated out of the `test/intl402/DateTimeFormat/` folder into a Temporal-specific folder?

Alternatively, maybe this test should both be copied into a Temporal folder, AND have a test remain here that instead of validating that DataTimeFormat canonicalizes zone IDs (current behavior) the test should validate that it's *not* canonicalized (new behavior)?  Note that V8 is not planning to change the canonicalization behavior of DateTimeFormat until Temporal Stage 4, so it may be premature to land a DTF non-canonical change now.

ljharb:
I’m in favor of it, but our policy is to not put undue burden on PR authors by requesting those changes, and instead to change them in future PRs.

ptomato:
I've gone ahead and made the ES5 changes as well, it's a tiny PR so why not

PR #4337, opened 2024-12-05
Temporal: Improve coverage of relativeto-string-limits tests

ptomato:
These tests did not fully cover Temporal.Duration.prototype.round and Temporal.Duration.prototype.total because they called those methods on a blank duration (all components zero), for which there is an early return in round() and total().

This meant that we missed an assertion that would be hit after the early return.

This makes sure to test both blank and non-blank Durations with these relativeTo strings, and expect some strings to fail at different steps with the two cases.

See: tc39/proposal-temporal#3015


ptomato:
cc @anba

Ms2ger (approved):
Thanks!

PR #4341, opened 2024-12-12
Update import defer tests for evaluation triggers

nicolo-ribaudo:
Ref https://github.com/tc39/test262/issues/4215.

This updates the tests that check what operations trigger evaluation of deferred modules, based on consensus in the last TC39 meeting. It's effectively what is included in https://github.com/tc39/proposal-defer-import-eval/pull/53.

This PR doesn't include tests about what the _results_ of those operations are (which will need to be tested for `.then`).

I suggest reviewing commit-by-commit:
- first the first one
- then the last one
- then the second one
- then the third one

All the tests are auto-generated from templates.

There is a bug with templates btw: their `desc` cannot start with `[`, because the generator will not escape/prefix it and so it becomes a Yaml array. I had to prefix all the descs with `_` for this reason.


takikawa:
I tried running this in my webkit prototype that's been updated to the new semantics, and as far as I can tell the test cases all run and agree with the implementation.

ptomato:
> There is a bug with templates btw: their `desc` cannot start with `[`, because the generator will not escape/prefix it and so it becomes a Yaml array. I had to prefix all the descs with `_` for this reason.

I think this is probably working as intended — the `.case` frontmatter is supposed to be valid YAML, so it wouldn't be escaped, rather parsed and merged into the `.template` frontmatter. Would it work if you did something like `desc: "[[DefineOwnProperty]]"`?

takikawa:
@nicolo-ribaudo asked me to take another look at this to check the coverage, and so I looked through the test cases and as far as I can tell it covers the evaluation trigger cases for module namespaces in the spec well. So LGTM.

Ms2ger (approved):

Ms2ger:
> No, because the generator parses that code, finds `[[DefineOwnProperty]]` as the value of `desc`, and then creates the generated files without the quotes around it.

Please file an issue for this

PR #4343, opened 2024-12-12
Remove invalid test that offset time zone are unsupported

### LINKED ISSUE #4339: intl402/Temporal/ZonedDateTime/prototype/toLocaleString/offset-time-zone-not-supported.js is no longer valid
Offset time zones are supported since <https://github.com/tc39/ecma402/pull/788>.

anba:
Fixes #4339


ptomato (approved):
Thanks! This is correct.

PR #4353, opened 2024-12-17
Update parseTestRecord and monkeyYaml to handle new tests

anba:
And make it easier to directly load existing tests using `yaml.safe_load`.


Ms2ger:
Haven't looked at the changes in detail, but wouldn't it be nice if we had CI

Ms2ger (approved):
Thanks!

PR #4354, opened 2024-12-17
Update staging/sm

Ms2ger:
Closes https://github.com/tc39/test262/pull/4346
Closes https://github.com/tc39/test262/pull/4347
Closes #4356


ptomato:
This is the result of running the import again after incorporating downstream feedback in the upstream SM copy of these tests? I think we should just merge it; ideally we'd automate this process in the future just like the V8 bot. Any objections if I go ahead and do that?

Maybe the `isError` test can be removed from SM if it is a duplicate of the existing non-staging `isError` tests.

anba:
Likely needs to be rebased after #4368.

ptomato (approved):

PR #4356, opened 2024-12-18
Add feature tag for async-iterator-helpers

anba:
The tests in "staging/sm/AsyncIterator" incorrectly use "async-iteration" (which is the feature tag for `async function*`). A new feature tag is needed for <https://github.com/tc39/proposal-async-iterator-helpers>.

### DIFF
--- a/features.txt
+++ b/features.txt
@@ -92,6 +92,10 @@ import-defer
 # https://github.com/tc39/proposal-iterator-sequencing
 iterator-sequencing
 
+# Async Iterator Helpers
+# https://github.com/tc39/proposal-async-iterator-helpers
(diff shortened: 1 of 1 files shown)


Ms2ger:
Thanks for flagging this. What stage is this proposal at? If it's too early, I'd need to skip them instead.

anba:
It's a stage 2 proposal.

ljharb:
should this PR also be updating the feature flag used in the relevant staging tests?

anba:
> should this PR also be updating the feature flag used in the relevant staging tests?

#4354 should cover updating the tests.

ljharb (approved):

PR #4358, opened 2024-12-18
Don't call assert.deepEqual.format when the assertion will succeed

anba:


anba:
"staging/sm/TypedArray/forEach" needs multiple seconds without this change. With this change applied, it completes in under a second.

Ms2ger (approved):
Makes sense to me.

PR #4360, opened 2024-12-18
Add test for @@toStringTag on deferred module namespaces

nicolo-ribaudo:
Test for https://github.com/tc39/proposal-defer-import-eval/pull/58.

Spec: https://tc39.es/proposal-defer-import-eval/#sec-modulenamespacecreate

### DIFF
--- a/test/language/import/import-defer/deferred-namespace-object/to-string-tag.js
+++ b/test/language/import/import-defer/deferred-namespace-object/to-string-tag.js
@@ -25,7 +25,7 @@ features: [import-defer]
 includes: [propertyHelper.js]
 ---*/
 
-import * as ns from "./dep_FIXTURE.js";
+import defer * as ns from "./dep_FIXTURE.js";
 
 verifyProperty(ns, Symbol.toStringTag, {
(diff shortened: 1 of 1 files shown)


test/language/import/import-defer/deferred-namespace-object/to-string-tag.js
```diff
@@ -0,0 +1,35 @@
...
+       is _toStringTag_ and whose [[Writable]], [[Enumerable]], and [[Configurable]]
+       attributes are false.
+    1. Return _M_.
+
+flags: [module]
+features: [import-defer]
+includes: [propertyHelper.js]
+---*/
+
+import * as ns from "./dep_FIXTURE.js";
```
linusg, line 28:
```suggestion
import defer * as ns from "./dep_FIXTURE.js";
```

nicolo-ribaudo: 🤦  Thanks

linusg (approved):

Ms2ger (approved):
Thanks

PR #4361, opened 2024-12-18
Ensure ToDateTimeFormattable is called for both arguments

anba:
Note: The Temporal polyfill doesn't pass this test.


Ms2ger (approved):
Thanks!

PR #4362, opened 2024-12-19
Fix two minor errors for import-defer tests

takikawa:
Fixes two minor issues in import-defer tests that can cause test failures.


Ms2ger (approved):
Thanks!

PR #4371, opened 2025-01-10
Add coverage when searching undefined after shrinking buffer

anba:
The "search-undefined-after-shrinking-buffer-index-is-oob.js" test fails in JSC and V8.


test/built-ins/TypedArray/prototype/includes/search-undefined-after-shrinking-buffer.js
```diff
@@ -0,0 +1,44 @@
...
+      rab.resize(0);
+
+      // Index is in-bounds when comparing to the original length.
+      return 0;
+    }
+  };
+
+  var result = ta.includes(undefined, index);
+
+  assert(result);
```
gibson042, line 42:
```suggestion
  assert.sameValue(result, true);
```

gibson042 (approved):

ljharb (approved):

Ms2ger (approved):

PR #3464, opened 2022-04-04
Add "Change Array by Copy" tests (stage 3)

nicolo-ribaudo:
This PR adds tests for https://github.com/tc39/proposal-change-array-by-copy, which reached consensus for Stage 3 at the March 2022 meeting.

I'm marking it as draft because I wrote these tests a few months ago, so I need to check that they are up-to-date with the current specification.

cc @acutmore @catamorphism

PR #4372, opened 2025-01-10
Move files from metadata

anba:
This was mentioned in https://github.com/tc39/test262/pull/3464#issuecomment-1088832153, but it looks like this clean-up never happened.


Ms2ger (approved):
Thanks

PR #4374, opened 2025-01-10
Add syntax tests for `import.defer`

nicolo-ribaudo:
These tests are all based on the existing dynamic import syntax tests, just updated to use the new syntax.

This PR does not contain any manually written tests -- they are all generated.

Ref https://github.com/tc39/test262/issues/4215

### DIFF
--- a/src/dynamic-import/import-defer-no-rest-param.case
+++ b/src/dynamic-import/import-defer-no-rest-param.case
@@ -1,0 +1,18 @@
+// Copyright (C) 2025 Igalia, S.L. All rights reserved.
+// Copyright (C) 2018 Leo Balter. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+desc: ImportCall is not extensible - no rest parameter
+template: syntax/invalid
+info: |
+  ImportCall :
+      import . source ( AssignmentExpression[+In, ?Yield] )
+
+  This production doesn't allow the following production from ArgumentsList:
+
+  ... AssignmentExpression
+features: [import-defer]
+---*/
+
+//- import
+import.defer(...['./empty_FIXTURE.js'])
--- a/src/dynamic-import/import-defer-no-rest-param.case
+++ b/src/dynamic-import/import-defer-no-rest-param.case
@@ -6,7 +6,7 @@ desc: ImportCall is not extensible - no rest parameter
 template: syntax/invalid
 info: |
   ImportCall :
-      import . source ( AssignmentExpression[+In, ?Yield] )
+      import . defer ( AssignmentExpression[+In, ?Yield] )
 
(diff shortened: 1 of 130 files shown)


src/dynamic-import/import-defer-no-rest-param.case
```diff
@@ -0,0 +1,18 @@
+// Copyright (C) 2025 Igalia, S.L. All rights reserved.
+// Copyright (C) 2018 Leo Balter. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+desc: ImportCall is not extensible - no rest parameter
+template: syntax/invalid
+info: |
+  ImportCall :
+      import . source ( AssignmentExpression[+In, ?Yield] )
```
ptomato, line 9:
```suggestion
      import . defer ( AssignmentExpression[+In, ?Yield, ?Await] )
```

src/dynamic-import/import-defer-script-code-valid.case
```diff
@@ -0,0 +1,16 @@
...
+desc: import.defer() can be used in script code
+template: syntax/valid
+features: [import-defer]
+---*/
+
+//- setup
+// This is still valid in script code, and should not be valid for module code
+// https://tc39.github.io/ecma262/#sec-scripts-static-semantics-lexicallydeclarednames
+var smoosh; function smoosh() {}
+
```
ptomato, line 14: I did not realize that we had automatic `setup` and `teardown` substitution variables in the test generator! (They are documented; I just never noticed.)

Anyway, not having `flags: [module]` in the test seems like it's sufficient for ensuring it's executed as script code, so is this really needed?

ptomato (approved):
Couple of minor comments, otherwise looks good.

(I reviewed only the `.case` files, and only glanced at the generated tests.)

takikawa:
I tried running this on the WebKit prototype implementation and the implementation and tests agreed, so looks good from that perspective.

src/dynamic-import/import-defer-script-code-valid.case
nicolo-ribaudo: I copied this from the `import.source` tests (these `.case` files are basically a copy-paste of those). It is also use in the `import` tests. I'm going to remove it from the `import.defer` tests, but probably it should then also be removed from those, since it's indeed useless (unless we want to catch broken test262 runners).

ptomato: Sure, would you like to make a follow-up PR for that?

nicolo-ribaudo: I'll open an issue assigned to myself for now :P

ptomato (approved):
Thank you!

PR #4375, opened 2025-01-10
Add evaluation tests for `import.defer` with sync/async modules

nicolo-ribaudo:
This is the last part of https://github.com/tc39/test262/issues/4215 :)

Except for the test in the `sync` folder, the others are adaptations of the ones in https://github.com/tc39/test262/blob/main/test/language/import/import-defer/evaluation-top-level-await

### DIFF
--- a/test/language/expressions/dynamic-import/import-defer/sync-dependency-of-deferred-async-module/main.js
+++ b/test/language/expressions/dynamic-import/import-defer/sync-dependency-of-deferred-async-module/main.js
@@ -1,0 +1,79 @@
+// Copyright (C) 2025 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-ContinueDynamicImport
+description: >
+  `import.defer` causes eager evaluation of synchronous dependencies of async dependencies
+info: |
+  13.3.10.1.1 ContinueDynamicImport (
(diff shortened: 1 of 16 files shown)


test/language/expressions/dynamic-import/import-defer/sync-dependency-of-deferred-async-module/main.js
```diff
@@ -0,0 +1,79 @@
...
+        1. If _result_ does not contain _m_, append _m_ to _result_.
+    1. Return _result_.
+
+flags: [module, async]
+features: [import-defer, top-level-await]
+includes: [compareArray.js]
+---*/
+
+import "./setup_FIXTURE.js";
+import defer * as ns from "./imports-tla-with-dep_FIXTURE.js";
```
ptomato, line 67: This line seems like it shouldn't be here, as it'll pre-evaluate the async parts of the graph before we get to the dynamic `import.defer` below?

ptomato (commented):
Just one comment, though I might be misunderstanding why that's there.

takikawa:
I tested these in the WebKit prototype and two of the tests found a bug in the implementation related to the async evaluation behavior. After fixing that bug, the tests pass on the prototype. LGTM modulo ptomato's comment above.

nicolo-ribaudo:
Fixed!

ptomato (approved):
Thank you!

PR #4385, opened 2025-01-30
Explicit resource management basic constructor and prototype tests

debadree25:
Hello test262 maintainers!

Based on the instructions in https://github.com/tc39/test262/pull/3866#issuecomment-2369085177 have split out this pr it includes some of the simple tests on the (Async)DisposableStack prototypes and constructor

Do let me know if this looks ok I shall follow this up with more split out prs sorry for the delay

Thank you!


ptomato (approved):
Thanks very much. This is about the right size to get done quickly and it's very helpful that the tests you've split out are similar to each other.

PR #4386, opened 2025-01-31
Update missed parens in module-self-import-async-resolution-ticks.js

Boshen:


ljharb (approved):
kind of hilarious how this was actually valid syntax beforehand :-p

ptomato (approved):
Much appreciated!

PR #4387, opened 2025-01-31
Explicit resource management iterator return

debadree25:
Hello!

This pr splits out tests for whether (Async)Iterator Symbol.dispose calls the return method or not from https://github.com/tc39/test262/pull/3866

Thank you!


ptomato (approved):
Thanks!

PR #4389, opened 2025-02-05
Temporal: Test coverage for rejecting too large dates in PlainMonthDay strings

ptomato:
This adds coverage for the normative change in https://github.com/tc39/proposal-temporal/pull/3054, which specifies that implementations must throw on out-of-range dates in RFC 9557 strings for non-ISO calendars.


ptomato:
Draft until the normative change is approved in TC39.

ptomato:
The corresponding normative change which this PR tests was approved on 2025-02-18 in the TC39 plenary.

Ms2ger (approved):

PR #4390, opened 2025-02-10
Add tests for SuppressedError

debadree25:
Hello!

This pr adds the test of SuppressedError from https://github.com/tc39/test262/pull/3866

Thank you!


ptomato (approved):
Thanks!

test/built-ins/NativeErrors/SuppressedError/order-of-args-evaluation.js
```diff
@@ -0,0 +1,37 @@
...
+    return '';
+  }
+};
+const error = {};
+const suppressed = {};
+
+const e = new SuppressedError(error, suppressed, message);
+
+assert.sameValue(messageStringified, true);
+const keys = Object.getOwnPropertyNames(e);
```
ljharb, line 33: i think this might not work as expected due to the own `stack` property in v8?

PR #4391, opened 2025-02-10
Add tests for symbols added in Explicit Resource Management

debadree25:
Hello!

This pr just adds the simple symbol tests from https://github.com/tc39/test262/pull/3866

Thank you!

### DIFF
--- a/test/built-ins/Symbol/asyncDispose/cross-realm.js
+++ b/test/built-ins/Symbol/asyncDispose/cross-realm.js
@@ -1,0 +1,14 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: pending
+description: Value shared by all realms
+info: |
+  Unless otherwise specified, well-known symbols values are shared by all
+  realms.
+features: [cross-realm, explicit-resource-management]
+---*/
+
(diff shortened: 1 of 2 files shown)


test/built-ins/Symbol/asyncDispose/cross-realm.js
```diff
@@ -0,0 +1,14 @@
+// Copyright (C) 2016 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: pending
```
ptomato, line 4: We should not add new tests with `esid: pending`. I guess the relevant sections haven't been added to the spec yet (https://github.com/tc39/proposal-explicit-resource-management/issues/185), but let's just assume they'll be called `sec-symbol.asyncdispose` and `sec-symbol.dispose`.

ptomato (approved):
Other than the `esid: pending`, looks good. Thanks!

test/built-ins/Symbol/asyncDispose/cross-realm.js
debadree25: Understood i guess the sections are there now, lemme check!

ljharb:
Can we also add some tests that `Symbol.keyFor()` of both of these is `undefined`?

ptomato:
Addressed the review comments, merging now, as this has been open for a long time

PR #4392, opened 2025-02-12
Some improvements to well-known intrinsic objects harness file

ptomato:
We have a `harness/wellKnownIntrinsicObjects.js` helper to aid with testing well-known intrinsic objects. There were some issues with it:
- The list of well-known intrinsic objects was out of date
- Some of the obtained objects were not the correct one
- We also had a `harness/hidden-constructors.js` file that duplicated some of the functionality

Here's an attempt at fixing this up a bit.

It doesn't convert any existing tests to use `getWellKnownIntrinsicObject` other than the ones that previously used `harness/hidden-constructors.js` which is now removed. If we like this API, I'll go on and update more existing tests to use it.


ptomato:
cc @p-bakker @linusg @anba since you recently touched `hidden-constructors.js`

linusg (approved):
Looks like a nice improvement!

harness/wellKnownIntrinsicObjects.js
```diff
@@ -140,8 +150,12 @@ const WellKnownIntrinsicObjects = [
     source: 'isNaN',
   },
   {
-    name: '%IteratorPrototype%',
```
ljharb, line 143: this intrinsic name will still be in the spec, so we should probably still have it in this list?

ljharb (approved):

harness/wellKnownIntrinsicObjects.js
ptomato: It is no longer there as far as I can tell. It seems to have been replaced by %Iterator.prototype% which means you can get it in this case with `getWellKnownIntrinsic('%Iterator%').prototype`

Ms2ger (approved):
Thanks

harness/wellKnownIntrinsicObjects.js
ljharb: hmm, that change was my PR, and the intention was to leave intact any pre-existing intrinsics, so that might be a spec bug I’ll have to fix.

Issue #4393, opened 2025-02-12
Add `Intl.Locale` test for `language` getter where language subtag is "und"


lionel-rowe:
Currently, both Chromium ([bug](https://issues.chromium.org/issues/334251989)) and Safari ([bug](https://bugs.webkit.org/show_bug.cgi?id=282504)) fail the following:

```js
assert(new Intl.Locale('und').language === 'und')
assert(new Intl.Locale('und-US-u-co-emoji').language === 'und')
```

Firefox has the correct behavior.

Could be useful to have a test for this specific case.

gibson042:
> I'm not sure if Firefox's is the correct behaviour. I'm looking at [Canonical Unicode Locale Identifiers](https://unicode.org/reports/tr35/#Canonical_Unicode_Locale_Identifiers) and I don't understand all of the terms there, but I do note that it has a special case to exclude `und` from one of the steps. Maybe someone with more expertise (like [@gibson042](https://github.com/gibson042) or [@sffc](https://github.com/sffc)) would be able to say what the correct behaviour is.
> 
> However, it does seem to me that Chrome and Safari's behaviour is incorrect. The _loc_.[[Locale]] slot being undefined fails several assertions in the spec, so it should not be possible to return `undefined` from this property getter.

The [[Locale]] slot of a Locale instance must always be a Unicode canonicalized locale identifier string, which for input "und" is "und" and for input "und-US-u-co-emoji" is "und-US-u-co-emoji" (i.e., both are already in canonical form). The `baseName`, `language`, `script`, and `region` getters extract and return relevant substrings. In short, the `assert`s in this report must pass in any conforming implementation of ECMA-402.

PR #4394, opened 2025-02-12
Add `Intl.Locale` test for `language` getter where language subtag is "und"

lionel-rowe:
Closes https://github.com/tc39/test262/issues/4393


ptomato:
As per https://github.com/tc39/test262/issues/4393#issuecomment-2654534692 I'm not sure whether this is the correct behaviour, so I'd like to settle that first before merging this.

gibson042 (approved):
Thanks!

ptomato (approved):

Issue #4310, opened 2024-11-05
Temporal: Move generic era and eraYear tests from intl402/Temporal to built-ins/Temporal


anba:
For example `get Temporal.PlainDate.prototype.era` is tested in <https://github.com/tc39/test262/tree/main/test/intl402/Temporal/PlainDate/prototype/era>, but all tests are generic and `get Temporal.PlainDate.prototype.era` is not ECMA-402 specific, so the tests should be moved to built-ins/Temporal.

ptomato:
Closed by #4397.

PR #4395, opened 2025-02-12
Moved generic era and eraYear tests from intl402/Temporal to built-ins/Temporal

brageh01:
Moved tests for era and eraYear due to all tests being generic and ```get Temporal.PlainDate.prototype.era``` not being ECMA-402 specific.

Issue: #4310


Ms2ger:
Why only for PlainDate?

brageh01:
> Why only for PlainDate?

I see now PlainDate was only used as an example. I will correct it for the other granularities of time as well.

PR #4396, opened 2025-02-12
Update String.fromCodePoint info to align with latest ECMAScript spec

yossydev:
It might be a minor detail, but it seems that the specification for fromCodePoint was changed in [tc39/ecma262#2018](https://github.com/tc39/ecma262/pull/2018) and [tc39/ecma262#2152](https://github.com/tc39/ecma262/pull/2152), while the info section had not been updated. So, I updated it accordingly.

### DIFF
--- a/test/built-ins/String/fromCodePoint/argument-is-Symbol.js
+++ b/test/built-ins/String/fromCodePoint/argument-is-Symbol.js
@@ -7,14 +7,12 @@ description: >
 info: |
   String.fromCodePoint ( ...codePoints )
 
-  1. Let codePoints be a List containing the arguments passed to this function.
-  2. Let length be the number of elements in codePoints.
-  3. Let elements be a new List.
-  4. Let nextIndex be 0.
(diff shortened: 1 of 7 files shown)


linusg (commented):
No strong feelings for or against this, but there are thousands of tests that have outdated metadata due to editorial changes - unless there's a way to automate these updates it won't be feasible to keep tests in sync with the spec.

yossydev:
> No strong feelings for or against this, but there are thousands of tests that have outdated metadata due to editorial changes - unless there's a way to automate these updates it won't be feasible to keep tests in sync with the spec.

You’re right. I noticed it this time, so I went ahead and fixed it, and I think that makes sense.

ptomato (approved):
Thanks!
Like Linus says, I don't want to suggest in any way that we commit to keeping these comment blocks up to date. But I'm happy to merge PRs to update them on a best-effort basis, like this one.

PR #4397, opened 2025-02-12
Moved generic era and eraYear tests from intl402/Temporal to built-ins/Temporal

brageh01:
Moved tests for era and eraYear due to all tests being generic and not being ECMA-402 specific.

Issue: #4310


ptomato (approved):
Thanks!

PR #4404, opened 2025-02-18
Add tests for not calling well-known symbols on primitives for string methods

lucacasonato:
See https://github.com/tc39/ecma262/pull/3009

### DIFF
--- a/test/built-ins/String/prototype/match/cstm-matcher-on-bigint-primitive.js
+++ b/test/built-ins/String/prototype/match/cstm-matcher-on-bigint-primitive.js
@@ -28,5 +28,4 @@ var separator = 1n;
 const matched = "a1b1c".match(separator);
 assert.sameValue(matched.index, 1);
 assert.sameValue(matched.input, "a1b1c");
-assert.sameValue(matched[0], "1");
-assert.sameValue(matched.length, 1);
+assert.compareArray(matched, ["1"]);
--- a/test/built-ins/String/prototype/match/cstm-matcher-on-bigint-primitive.js
+++ b/test/built-ins/String/prototype/match/cstm-matcher-on-bigint-primitive.js
@@ -4,12 +4,12 @@
 /*---
 esid: sec-string.prototype.match
 description: >
-  If a separator is a bigint primitive, its Symbol.match property is not accessed.
+  If a regexp property is a bigint primitive, its Symbol.match property is not accessed.
 info: |
   String.prototype.match ( regexp )
 
   [...]
-  2. If separator is not Object, then
+  2. If regexp is not Object, then
     [...]
   [...]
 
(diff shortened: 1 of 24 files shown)


test/built-ins/String/prototype/match/cstm-matcher-on-bigint-primitive.js
```diff
@@ -0,0 +1,32 @@
...
+  If a separator is a bigint primitive, its Symbol.match property is not accessed.
+info: |
+  String.prototype.match ( regexp )
+
+  [...]
+  2. If separator is not Object, then
+    [...]
+  [...]
+
+includes: [compareArray.js]
```
ljharb, line 16:
```suggestion
```

test/built-ins/String/prototype/match/cstm-matcher-on-boolean-primitive.js
```diff
@@ -0,0 +1,32 @@
...
+  If a separator is a boolean primitive, its Symbol.match property is not accessed.
+info: |
+  String.prototype.match ( regexp )
+
+  [...]
+  2. If separator is not Object, then
+    [...]
+  [...]
+
+includes: [compareArray.js]
```
ljharb, line 16:
```suggestion
```

test/built-ins/String/prototype/match/cstm-matcher-on-number-primitive.js
```diff
@@ -0,0 +1,32 @@
...
+  },
+});
+
+var separator = 1;
+
+const matched = "a1b1c".match(separator);
+assert.sameValue(matched.index, 1);
+assert.sameValue(matched.input, "a1b1c");
+assert.sameValue(matched[0], "1");
+assert.sameValue(matched.length, 1);
```
ljharb, line 32: i think comparing to an array of `['1']` might be simpler here, assuming it ignores expandos

ljharb (commented):
generally seems good overall

ljharb (commented):
for Symbol.match, what about `new RegExp`?

lucacasonato:
The normative PR does not change the behaviour of the regexp constructor. It already only gets `%Symbol.match%` for objects (see https://tc39.es/ecma262/#sec-isregexp).

ljharb:
aha, then the normative PR has an even stronger case in favor of it :-)

ljharb (approved):

ptomato (approved):
Seems good. Any value for implementations in being exhaustive and adding tests for Symbol primitives as well? I'm leaning towards no.

lucacasonato:
Agree

ptomato:
The normative change has consensus and there were no further comments from implementations. I think we should merge this now.

PR #4410, opened 2025-02-25
Update sm tests for iterator helpers

rmahdav:
After a recent normative change on iterator helpers (https://github.com/tc39/ecma262/pull/3467), the expectations of these tests should be updated.


ptomato (approved):
Thank you!

PR #4411, opened 2025-02-26
Rewrite import defer resolution test to not expect specific error

nicolo-ribaudo:
The error type is host-defined, so we cannot use a negative-style test.

### DIFF
--- a/test/language/import/import-defer/errors/resolution-error/import-defer-of-missing-module-fails.js
+++ b/test/language/import/import-defer/errors/resolution-error/import-defer-of-missing-module-fails.js
@@ -9,14 +9,18 @@ info: |
   LoadRequestedModules ([ _hostDefined_ ])
     - just notice that it does not check if the module is deferred
 
-flags: [module]
+flags: [async, module]
+includes: [asyncHelpers.js]
 features: [import-defer]
-
-negative:
-  phase: resolution
-  type: SyntaxError
 ---*/
 
-$DONOTEVALUATE();
-
-import defer * as ns from "./resolution-error_FIXTURE.js";
+asyncTest(async () => {
+  globalThis.evaluated = false;
+  try {
+    await import("./main_FIXTURE.js");
+  } catch {
+    assert.sameValue(globalThis.evaluated, false, "The module should not be evaluated");
+    return;
+  }
+  throw new Error("The module should throw");
+})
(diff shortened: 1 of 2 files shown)


test/language/import/import-defer/errors/resolution-error/main_FIXTURE.js.js
linusg, line 1: I assume the duplicate file extension is not intentional?

nicolo-ribaudo: yep

Ms2ger (approved):
Ugh

linusg:
Unlike missing or ambiguous exports, module _resolution_ failures cannot be easily categorized as "syntax errors". Why should a module fetch timeout throw a SyntaxError for example?

In practice runtimes commonly throw plain errors for module resolution failures. JSC also throws a TypeError for certain invalid specifiers.

- https://github.com/nodejs/node/blob/304743655d5236c2edc39094336ee2667600b684/src/node_errors.h#L102
- https://github.com/v8/v8/blob/3c8ed98716870fdeed8e5acea2b99ace054af464/src/d8/d8.cc#L1188-L1195
- https://github.com/WebKit/WebKit/blob/9c07d7ffebff370be207c8f184fd6d176dbb98e4/Source/JavaScriptCore/jsc.cpp#L1145-L1159

nicolo-ribaudo:
The spec for this is at https://tc39.es/ecma262/#sec-HostLoadImportedModule, first bullet point. It requires the host to return a throw completion, but has no requirements on what that throw completion must contain. The host could even through a primitive.

ljharb:
and test262 negative tests can't omit the `type`?

linusg:
https://github.com/tc39/test262/blob/HEAD/CONTRIBUTING.md#negative doesn't mention anything so I assume test runners currently expect it.

nicolo-ribaudo:
> and test262 negative tests can't omit the type?

I'd love to, but it already caused problems with some runners for the `import source` tests.

I think we can change the requirements, but I'd do it as a follow-up to this PR, and it would take some time for runners to update.

ljharb:
fair point. we should pursue ensuring type can be omitted, but this seems fine as-is until then

takikawa (approved):
I tested this in the WebKit prototype for import-defer and the test passes now.

Issue #3896, opened 2023-08-14
Test gap for invalid string calendar IDs


justingrant:
Codecov found that there's no test coverage for invalid calendar string IDs, e.g. `"notacal"` in the following AOs:
* ToRelativeTemporalObject
* ToTemporalMonthDay
* ToTemporalYearMonth
* ToTemporalZonedDateTime

These AOs are used in a lot of places, so I'm not sure which test files need to be changed or added (likely 100s of them), but here's a few code samples as a starting point:

```js
Temporal.Duration.from('P1M').add({months: 2}, {relativeTo: "2020-01-01[u-ca=notacal]"} );
Temporal.PlainMonthDay.from("2020-01-01[u-ca=notacal]");
Temporal.PlainYearMonth.from("2020-01-01[u-ca=notacal]");
Temporal.ZonedDateTime.from("2020-01-01[UTC][u-ca=notacal]");
```

catamorphism:
This is still an issue; the examples above correctly throw exceptions now, but this, for example, doesn't throw:

```
Temporal.ZonedDateTime.from("2020-01-01[UTC]").round({smallestUnit: "nanosecond", relativeTo: "2020-01-01[UTC][u-ca=notacal]"})
```

ptomato:
What does the spec say about the new example? It could be that `relativeTo` intentionally isn't validated because rounding to the nearest 1 nanosecond is a no-op.

catamorphism:
This also succeeds:

```
Temporal.ZonedDateTime.from("2020-01-01T00:17:00[UTC]").round({smallestUnit: "hour", roundingIncrement: 12, relativeTo: "2020-01-01[UTC][u-ca=notacal]"})
```

and the rounding is not a no-op in this case.

catamorphism:
Oops, that wasn't a good example since `ZonedDateTime.prototype.round` doesn't actually use the `relativeTo` option. I don't have a test case right now that succeeds when it should fail, but the test coverage should be added if it's not there already (I haven't fully checked).

ioannad:
The PR adding the test coverage is now landed, closing this.

PR #4415, opened 2025-03-04
Temporal: Reorganize calendar and wrong-type tests

brageh01:
Changes made to:
- PlainDate
- PlainYearMonth
- ZonedDateTime
- PlainDateTime
- PlainMonthDay

Merged ```calendar-number.js``` and ```calendar-wrong-type.js```, due to no special behaviour of number to be tested.

Extracted string tests to ```calendar-iso-string.js``` and renamed the file to ```calendar-invalid-iso-string.js```, due to the contents of the file initially testing invalid ISO strings.

**Questions:**

Are change to be made for the ```argument-propertybag-calendar-number.js``` and ```argument-propertybag-calendar-wrong-type.js``` files as well? 
However, I notice that the ```argument-propertybag-calendar-iso-string.js``` actually tests for valid and not invalid ISO strings here. Perhaps create a ```invalid-iso-string.js```? 

Is the "merge-and-extract» also relevant for ```relativeto-propertybag-calendar-number.js``` and ```relativeto-propertybag-calendar-wrong-type.js```?


sffc:
About the questions: my sense is that the `propertybag` tests are probably also in scope to be refactored here in a similar way.

test/built-ins/Temporal/ZonedDateTime/from/argument-propertybag-calendar-wrong-type.js
```diff
@@ -9,20 +9,23 @@ description: >
 features: [BigInt, Symbol, Temporal]
 ---*/
 
+const timeZone = "UTC";
```
ptomato, line 12: I see you found a pre-existing error in this test, we forgot the time zone. Thanks!

(Well, technically it's not an error; in step 4.b-c of [ToTemporalZonedDateTime](https://tc39.es/proposal-temporal/#sec-temporal-totemporalzoneddatetime) the check for a faulty calendar occurs before the check for a missing time zone. So the test was still correct, but it was testing something different than I expected.)

Normally I would suggest that we should make a separate test file to ensure that the "faulty calendar" exception is thrown before the "missing time zone" exception, but in this case I think no further action is needed. We generally don't test that sort of thing anyway currently. (See also #2746 for a draft of how we might test it in the future.)

ptomato (commented):
Thanks! It looks like this is on the right track. There may be some tests still left to do. For example, there should be `relativeto-propertybag-...` tests in `Duration/prototype/round/` and `Duration/compare/` as well, and I think there are a lot more `argument-propertybag-...` tests; there should be some for every API entry point that takes a Temporal object, for example `withPlainTime` methods.

Feel free to use a script to automate a first draft of changes if there are a lot of files to do.

To answer your questions:

> Are change to be made for the argument-propertybag-calendar-number.js and argument-propertybag-calendar-wrong-type.js files as well?

Yes please.

> However, I notice that the argument-propertybag-calendar-iso-string.js actually tests for valid and not invalid ISO strings here. Perhaps create a invalid-iso-string.js?

That sounds correct.

> Is the "merge-and-extract» also relevant for relativeto-propertybag-calendar-number.js and relativeto-propertybag-calendar-wrong-type.js?

Yes.

ptomato:
@brageh01 Looks like there's some new stuff on here, thanks! When it's ready for another review, please let me know either on Matrix chat or by clicking the "re-request review" button.

test/built-ins/Temporal/Duration/prototype/total/relativeto-propertybag-invalid-iso-string.js
```diff
@@ -1,26 +1,23 @@
-// Copyright (C) 2022 Igalia, S.L. All rights reserved.
+// Copyright (C) 2025 Igalia, S.L. All rights reserved.
```
ptomato, line 1: No need to update copyright years in existing files (unless you're really rewriting a substantial portion of the file.) This comment applies to several files.

test/built-ins/Temporal/Duration/prototype/total/relativeto-propertybag-invalid-iso-string.js
```diff
@@ -1,26 +1,23 @@
...
 
 const instance = new Temporal.Duration(1, 0, 0, 0, 24);
 
-const numbers = [
-  1,
-  19970327,
-  -19970327,
-  1234567890,
+const invalidStrings = [
+  ["", "empty string"],
```
ptomato, line 13: Optional: you could consider adding more cases in tests such as this one. For example: strings that are neither ISO strings nor valid calendar identifiers. (If you decide this is out of scope for this PR, that would be fine, but I'd appreciate if you could open a follow up issue on GitHub in that case.)

test/built-ins/Temporal/PlainDate/calendar-invalid-iso-string.js
```diff
@@ -0,0 +1,21 @@
+// Copyright (C) 2025 Igalia, S.L. All rights reserved.
```
ptomato, line 1: I think it's OK to put your own copyright here (yourself, or University of Bergen, whatever applies in this case.) Although the test is similar to one that Igalia wrote, but it's a new test.

test/built-ins/Temporal/PlainDate/compare/argument-propertybag-calendar-wrong-type.js
ptomato, line 1: For this test and other files like it, we should now be able to merge `primitiveTests` and `typeErrorTests` into one array of tests that we can just call `typeErrorTests` or `wrongTypeTests` or even just `tests`, and merge the two loops into one. We no longer have to deal with some entries giving a RangeError.

test/built-ins/Temporal/PlainDateTime/prototype/withCalendar/calendar-invalid-iso-string.js
```diff
@@ -0,0 +1,23 @@
...
+esid: sec-temporal.plaindatetime.prototype.withcalendar
+description: Invalid ISO 8601 string is not accepted as calendar
+features: [Temporal]
+---*/
+
+const instance = new Temporal.PlainDateTime(1976, 11, 18, 15, 23, 30, 123, 456, 789, "iso8601");
+
+const invalidStrings = [
+  ["", "empty string"],
+  ["1997-12-04[u-ca=iso8601]", "ISO string with calendar annotation"],
```
ptomato, line 14: I think this line should not be here. According to my reading of [the proposal](https://tc39.es/proposal-temporal/#sec-temporal.plaindate.prototype.withcalendar), it's valid to pass an ISO string to this method.

ptomato (changes requested):
@brageh01 Thank you for all the work! It looks nearly complete now. Just some minor comments.

test/built-ins/Temporal/Duration/prototype/total/relativeto-propertybag-invalid-iso-string.js
brageh01: I believe updating the copyright year for this file is correct. Instead of deleting the previous ```calendar-number.js``` I renamed the file to an ```invalid-iso-string.js```, but it is initially a new file with different functionality.

Looks like I’ve forgotten the «calendar» tag as well. I will fix it.

test/built-ins/Temporal/Duration/prototype/total/relativeto-propertybag-invalid-iso-string.js
brageh01: I now see that I have not done the merge and extract for ```round``` in the same file location which I will do. In regards of your suggestion of adding more cases in tests "such as this one", do you imply only for the relativeto-propertybag files, or all of the calendar-invalid-iso-string files in general?

I believe it will be more transparent to create a new issue for this, but I can both create a new issue and start a PR after knowing more specificly which files that are relevant.

ptomato: I meant for the calendar-invalid-iso-string files in general. Every file where the only test case is `["", "empty string"]`. I agree, it's fine to create a new issue.

ptomato (approved):
Thank you for all the work on this pull request! This looks all correct.

test/built-ins/Temporal/Duration/prototype/total/relativeto-propertybag-invalid-iso-string.js
ptomato: I found an already open issue that covers this topic, I think: https://github.com/tc39/test262/issues/3896

PR #4417, opened 2025-03-04
Sync test to https://github.com/tc39/proposal-intl-locale-info/pull/99

FrankYFTang:


FrankYFTang:
@sffc

test/intl402/Locale/prototype/getWeekInfo/output-object-keys.js
```diff
@@ -53,12 +52,8 @@ let original = new Intl.Locale('en').getWeekInfo().weekend;
...
-  enumerable: true,
-  configurable: true
-});
-assert(
-  isIntegerBetweenOneAndSeven(new Intl.Locale('en').getWeekInfo().minimalDays),
-  '`minimalDays` must be an integer between one and seven (inclusive)'
+assert.sameValue(
+  new Intl.Locale('en').getWeekInfo().minimalDays,
+  undefined,
+  '`minimalDays` is not defined.'
```
ptomato, line 58: This can probably just be deleted, as the `ownKeys` test above already ensures there isn't a `minimalDays` own property. (An implementation would still be technically, but uselessly, conformant if it added a `Object.prototype.minimalDays` property.)

ptomato (approved):
Looks good, thanks. I would suggest just deleting the check that `.minimalDays` evaluates to undefined.

test/intl402/Locale/prototype/getWeekInfo/output-object-keys.js
FrankYFTang: PTAL

PR #4418, opened 2025-03-10
Add testcases for setUTC* functions return value

DevSDK:
I found some missing testcase for setUTC* functions unlike set* functions do. While I am working on [this CL](https://chromium-review.googlesource.com/c/v8/v8/+/6332619), I found that I was missing the `std::optional::value_or` of setUTC* functions by mistake but test262  succeeded. (But set* functions failed.)

So, I think this is a good point to add the cases. The TCs are from set* cases. (e.g., [setHours](https://github.com/tc39/test262/blob/main/test/built-ins/Date/prototype/setHours/this-value-valid-date-hour.js))


ptomato (approved):
Thanks, looks good. This seems like a straightforward port of the coverage we have for the non-UTC methods.

Issue #4416, opened 2025-03-04
import defer test bugs


woess:
While implementing support for `import defer`, I *think* I've found the following bugs in the import-defer tests:

1. `language/import/import-defer/evaluation-top-level-await/flattening-order/main.js`:
The test expects `"5"` twice. I don't see how that could happen, the second (last) `"5"` should be removed:
https://github.com/tc39/test262/blob/9d8efae5c2c1848b9c3362c4aa81dd360ff8d3a7/test/language/import/import-defer/evaluation-top-level-await/flattening-order/main.js#L67
2. `language/import/import-defer/errors/get-self-while-defer-evaluating/main.js`:
I don't see how the assertion can pass before the evaluation trigger (`ns.foo`). The lines probably just need to be swapped?
https://github.com/tc39/test262/blob/9d8efae5c2c1848b9c3362c4aa81dd360ff8d3a7/test/language/import/import-defer/errors/get-self-while-defer-evaluating/main.js#L38-L42
3. [`language/import/import-defer/errors/module-throws/defer-import-after-evaluation.js`](https://github.com/tc39/test262/blob/9d8efae5c2c1848b9c3362c4aa81dd360ff8d3a7/test/language/import/import-defer/errors/module-throws/defer-import-after-evaluation.js):
In order for `assert.sameValue(err1, err2)` to be pass, both imported modules need to produce the same error (they don't). So I guess the import in `import-defer-throws_FIXTURE.js` was supposed to import `throws_FIXTURE.js` rather than itself, i.e.:
```diff
-import defer * as ns from "./import-defer-throws_FIXTURE.js";
+import defer * as ns from "./throws_FIXTURE.js";
```
4. `language/import/import-defer/deferred-namespace-object/exotic-object-behavior.js`:
https://github.com/tc39/test262/blob/9d8efae5c2c1848b9c3362c4aa81dd360ff8d3a7/test/language/import/import-defer/deferred-namespace-object/exotic-object-behavior.js#L46-L51
This one's tricky: While the test is technically verifying the correct thing, I believe `verifyProperty` cannot be used in this case because it not only checks the property descriptor but also tries to verify actual writability by setting a new value, but module namespace exotic object properties aren't truly writable in the sense that writes (aka `[[Set]]`, `[[DefineOwnProperty]]`) don't have any effect/don't succeed.

PR #4420, opened 2025-03-12
Fix import defer test bugs

nicolo-ribaudo:
All the suggestions in https://github.com/tc39/test262/issues/4416 are correct. @woess Can you confirm that this makes the tests pass? :)

Fixes https://github.com/tc39/test262/issues/4416


woess:
@nicolo-ribaudo I can confirm that this fixes the tests, and that _all_ import-defer tests pass in my branch of GraalJS now. :)

takikawa (approved):
LGTM, this change makes two more tests pass on WebKit (two still fail).

PR #4423, opened 2025-03-13

Aurele-Barriere:
This PR is the follow-up of https://github.com/tc39/test262/pull/4364 and https://github.com/tc39/test262/pull/4195

For each character class escape (\d, \D, \s, \S, \w, \W), check positive cases (the escape matches all characters it's supposed to match) and negative cases (the escape doesn't match any of the characters it should not match).  Each of these checks is also done in Unicode mode and with the v flag.

This uses regenerate.js from the unicode-property-escapes-tests repo to generate strings that contain exactly the characters that are supposed to be matched or not matched for each escape.

Comparison is done with regex test instead of regex replace to optimize the tests.

This is part of my work at the SYSTEMF lab at EPFL.


Aurele-Barriere:
Replying to a comment from @ptomato in #4364 here:

> For future reference — when I cleaned up the original test generator I removed this file that modified the regenerate library's object prototype, and instead I turned it into regular functions in index.mjs. I would like to try to minimize the diff in this PR so that it's easier to see what changed, so I've taken the liberty of editing these functions so they don't modify the regenerate prototype and moving them back to index.mjs.

I can't see the changes you've made but I've tried replicating them in the most recent commit, and removed the file `regenerate.mjs`. Let me know if this is what you had in mind.

ptomato (approved):
Actually never mind, I answered my own questions by reading the description of the original PR you opened (#4195). I think this is ready to merge. Thank you very much for sticking with this through the long process of fixing and moving over the test generator script.

Aurele-Barriere:
Perfect, thanks for the review and edits! Happy to have helped!

PR #4425, opened 2025-03-14
Update non-iso-calendars.js and addition-across-lunisolar-leap-months.js

anba:
1. non-iso-calendars.js
- Split into separate files for each calendar.
- Fix various issues in the test data. For example adjusting Hijri dates to work with ICU4C and ICU4X based implementations. Or adjusting `islamic-umalqura` to use dates which don't fallback to `islamic-civil` in ICU4C.
- Remove workarounds for old V8 versions when formatting ISO-8601 and Gregorian calendars.
- Remove test data which are specific to ICU4C and don't necessarily reflect reality. For example Japanese eras before Meiji, which are based on questionable CLDR data. 
- It's also worth noting that most tests in "non-iso-calendars.js" were skipped, because the skip condition was wrong: <https://github.com/tc39/test262/blob/7708262a21785d8ed188b89272e3341c6453909e/test/staging/Intl402/Temporal/old/non-iso-calendars.js#L548>
- The if-condition means the test is skipped unless the current calendar is either `chinese` or `dangi` **and** the underlying Temporal implementation uses out-dated Chinese calendar data...

2. addition-across-lunisolar-leap-months.js
- Split into separate files for each calendar.
- Remove unnecessary check for out-dated Chinese calendar data.


test/staging/Intl402/Temporal/old/addition-across-lunisolar-leap-months-chinese.js
```diff
@@ -0,0 +1,49 @@
...
+
+/*---
+esid: sec-temporal-intl
+description: Addition across lunisolar leap months
+features: [Temporal]
+---*/
+
+const chineseYearOffset = new Temporal.PlainDate(1, 1, 1, "chinese").year;
+
+// Adding years across Chinese leap month"
```
Ms2ger, line 12:
```suggestion
// Adding years across Chinese leap month
```

Ms2ger (approved):
Thanks. I don't have time to do a full review right now, but I'll approve and wait until early next week in case anyone else has time or objections

test/staging/Intl402/Temporal/old/addition-across-lunisolar-leap-months-chinese.js
anba: Fixed.

PR #4430, opened 2025-03-17
Fix comment

catamorphism:


catamorphism:
(This seems to be a copy/paste error; in the spec, `ToNumber` is crossed out.)

PR #4431, opened 2025-03-18
Temporal: Add test for different combinations of DateTimeFormat options

catamorphism:
This test file adds tests for all types of Temporal objects and various combinations of style options for Intl.DateTimeFormat.

### DIFF
--- a/test/intl402/DateTimeFormat/prototype/format/temporal-objects-formatting.js
+++ b/test/intl402/DateTimeFormat/prototype/format/temporal-objects-formatting.js
@@ -4,7 +4,7 @@
 /*---
 esid: sec-datetime-format-functions
 description: Different combinations of style options and Temporal types format correctly.
-locale: [en]
+locale: [en-US]
 features: [Temporal]
 ---*/
 
--- a/test/intl402/DateTimeFormat/prototype/format/temporal-objects-formatting.js
+++ b/test/intl402/DateTimeFormat/prototype/format/temporal-objects-formatting.js
@@ -33,156 +33,345 @@ const dateTimeFormatterFullMedium = new Intl.DateTimeFormat("en-US", { dateStyle
 const dateTimeFormatterFullShort = new Intl.DateTimeFormat("en-US", { dateStyle: "full", timeStyle: "short", timeZone: "Pacific/Apia" });
 const dateTimeFormatterFullLong = new Intl.DateTimeFormat("en-US", { dateStyle: "full", timeStyle: "long", timeZone: "Pacific/Apia" });
 
-const date = new Temporal.PlainDate(2021, 8, 4);
-const datetime = new Temporal.PlainDateTime(2021, 8, 4, 0, 30, 45, 123, 456, 789);
-const monthday = new Temporal.PlainMonthDay(8, 4, "gregory");
-const yearmonth = new Temporal.PlainYearMonth(2021, 8, "gregory");
-const time = new Temporal.PlainTime(0, 30, 45, 123, 456, 789);
-
 // PlainDate
 
-var dateResult = dateFormatterShort.format(date);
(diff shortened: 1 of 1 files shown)


test/intl402/DateTimeFormat/prototype/format/temporal-objects-formatting.js
```diff
@@ -0,0 +1,188 @@
+// Copyright (C) 2025 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-datetime-format-functions
+description: Different combinations of style options and Temporal types format correctly.
+locale: [en]
```
ptomato, line 7: I believe this'd need to be
```suggestion
locale: [en-US]
```

ptomato (commented):
I have mixed feelings about this one. We are trying to reduce the number of tests that use "golden output" to test locale-dependent functions, because they tend to break anytime a browser updates its CLDR version. But, we do not yet have a policy of not accepting golden output tests.

So this could be merged as is, but it'd be really helpful if the assertions could be changed so that they assert more general properties of the output instead of the exact contents of the string. Would that be possible?

A few suggestions off the top of my head:
- the date-only types could assert that the output of the dateStyle+timeStyle formatters was identical to the corresponding dateStyle formatter and that timeStyle was ignored.
- For PlainDateTime you could use 2222-03-04T05:06:07.888999111 and assert that, for example, "2", "3", and "4" do not occur in any string output by a timeStyle formatter
- For PlainYearMonth and PlainMonthDay you could create it with an unusual reference day or reference year respectively, and assert that it does not occur in any string output

(And above all, thanks for doing the work!)

More info:
https://github.com/tc39/test262/issues/3786
http://ptomato.name/talks/tc39tg2-2024-10

catamorphism:
@ptomato I've tried to do what you suggested; I did keep the original tests as commented-out code, so it's easier to read the test and figure out what the expected output should (roughly) look like.

test/intl402/DateTimeFormat/prototype/format/temporal-objects-formatting.js
```diff
@@ -0,0 +1,377 @@
...
+
+const timeResultMedium = timeFormatterMedium.format(time);
+// assert.sameValue(timeResultMedium, "12:34:56 AM", "plain time, dateStyle=medium");
+assert.notSameValue(timeResultMedium.search("56"), -1, "plainTime, dateStyle=medium: seconds should appear");
+assert.sameValue(timeResultMedium.search("777"), -1, "plainTime, dateStyle=medium: milliseconds should not appear");
+
+const timeResultLong = timeFormatterLong.format(time);
+// assert.sameValue(timeResultLong, "12:34:56 AM", "plain time, dateStyle=long");
+assert.notSameValue(timeResultMedium.search("34"), -1, "plainTime, dateStyle=long: minutes should appear");
+assert.sameValue(timeResultMedium.search("888"), -1, "plainTime, dateStyle=long: microseconds should not appear");
```
ptomato, line 360: s/dateStyle/timeStyle/ in these

test/intl402/DateTimeFormat/prototype/format/temporal-objects-formatting.js
ptomato: I think this one got missed in the rewrite.

ptomato (approved):
Thanks, this looks exactly like I was hoping it'd look. Just some minor fixups

test/intl402/DateTimeFormat/prototype/format/temporal-objects-formatting.js
catamorphism: Fixed

PR #4433, opened 2025-03-21
Allow for implementation-defined Error properties

anba:
Various implementations (JSC, V8, SM) define additional properties on Error instances. Adjust the test case to account for that.


Ms2ger (approved):
Please add a note to the test why it's written this way

anba:
> Please add a note to the test why it's written this way

Done. Added the following comment:
```js
// Allow implementation-defined properties before "message" and after "suppressed".
```

ljharb (approved):
it might be cleaner to slice keys at messageIndex for 3 total items, and compareArray, but either way is fine

PR #4435, opened 2025-03-21
Split "Temporal/old/islamic-calendars.js" into different files for each calendar variant

anba:
This allows implementations to more easily skip some tests.


Ms2ger (approved):
Thanks

PR #4438, opened 2025-03-25
Remove chakra

ryzokuken:
While still a relatively active project, ChakraCore is no longer maintained by Microsoft. It's not supported by esvu anymore and it was already disabled on CI.

Also realized while making this PR that my editor automatically trimmed some unnecessary whitespace from CONTRIBUTING.md, wasn't sure if I should go back and get rid of those changes or if it's actually okay. Let me know either way.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -68,15 +68,6 @@ jobs:
       hostPath: spidermonkey
       hostName: spidermonkey
     <<: [*execution_steps]
-  # "ChakraCore: New or modified tests execution":
-  #   docker:
-  #     - image: *node_image
-  #   working_directory: ~/test262
-  #   environment:
-  #     hostType: ch
-  #     hostPath: chakra
-  #     hostName: chakra
-  #   <<: [*execution_steps]
   "JSC: New or modified tests execution":
     docker:
       - image: *node_image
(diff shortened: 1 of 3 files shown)


linusg:
> While still a relatively active project

By what metric?

image

linusg:
It had _one_ commit in February, the last commit before that was in August 2024. I'm not arguing against the removal (and in fact I wish we'd remove it from https://test262.fyi already), but I think it's important to be clear about the project status here - if it was active we should encourage the team behind it to restore esvu support and re-enable CI. Since it's completely dead and that's not happening we should not pretend otherwise.

ryzokuken:
I see your point that it's effectively unmaintained but not sure what to do about it apart from maybe changing the commit text.

ptomato (changes requested):
Hmm... +1 on removing `npm run test:diff:chakra`. -0 on removing the commented-out CI job. -1 on removing the link to the Chakracore bug tracker. But you could replace it with the up-to-date link, https://github.com/chakra-core/ChakraCore/issues

ryzokuken:
Honestly I just wanted to remove it from `npm run test:diff` because I couldn't run it without chakra, everything else is optional. I'll make the changes you suggested and break the whole thing up into multiple commits so you could take what we manage to get agreement on, but as I mentioned the only thing I actually feel strongly about is the one you're positive about.

ptomato (approved):

PR #4446, opened 2025-04-03
Add more tests for `Object.prototype.[[SetPrototypeOf]]`

ExE-Boss:
Turns out, this case fails in **engine262**, which I’m fixing in:
- <https://github.com/engine262/engine262/pull/267>

### DIFF
--- a/test/built-ins/Object/prototype/setPrototypeOf-with-non-circular-values-__proto__.js
+++ b/test/built-ins/Object/prototype/setPrototypeOf-with-non-circular-values-__proto__.js
@@ -1,0 +1,33 @@
+// Copyright (C) 2025 ExE Boss. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-immutable-prototype-exotic-objects-setprototypeof-v
+description: >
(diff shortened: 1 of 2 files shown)


test/built-ins/Object/prototype/setPrototypeOf-with-non-circular-values-__proto__.js
```diff
@@ -0,0 +1,33 @@
...
+  ...
+  2. Let current be the value of the [[Prototype]] internal slot of O.
+  3. If SameValue(V, current), return true.
+  4. Return false.
+
+  19.1.3 Properties of the Object Prototype Object
+
+  The value of the [[Prototype]] internal slot of the Object prototype object is
+  null and the initial value of the [[Extensible]] internal slot is true.
+features: [Reflect.setPrototypeOf]
```
ljharb, line 20: it would be nice to split the Reflect tests into their own files (4 total) so we have maximal coverage across feature matrices, but not a requirement

ljharb (approved):

ptomato (approved):

ptomato:
Been approved for 2 weeks, so merging now. Do feel free to submit a follow up with Jordan's suggestion if you're still planning to.

PR #4453, opened 2025-04-12
Fix time zone canonicalization behaviour in DateTimeFormat test

ptomato:
As of the Time Zone Canonicalization proposal which is stage 3, the original time zone name should be preserved in Intl.DateTimeFormat .prototype.resolvedOptions.

Add a separate test that uses Temporal.ZonedDateTime.prototype.equals to test the canonicalization behaviour.


ptomato:
cc @justingrant

justingrant (approved):
👍

ryzokuken (approved):

test/intl402/DateTimeFormat/canonicalize-timezone.js
```diff
@@ -96,7 +97,7 @@ const timeZones = [
 for (const [timeZone, expected] of timeZones) {
```
Ms2ger, line 97: Please rename "expected"

Ms2ger (approved):

PR #4454, opened 2025-04-15
Add upsert tests

dminor:
These tests are currently being executed in SpiderMonkey as `non262` tests, and were exported using the `test262-export.py` script. Linting errors were then corrected.


test/staging/upsert/Map/getOrInsert/append-new-values-normalizes-zero-key.js
```diff
@@ -0,0 +1,28 @@
...
+  3. Set key to CanonicalizeKeyedCollectionKey(key).
+  4. For each Record { [[Key]], [[Value]] } p of M.[[MapData]], do
+    a. If p.[[Key]] is not empty and SameValue(p.[[Key]], key) is true, return p.[[Value]].
+  5. Let p be the Record { [[Key]]: key, [[Value]]: value }.
+  6. Append p to M.[[MapData]].
+  ...
+features: [Symbol]
+flags: [noStrict]
+---*/
+// Copyright (C) 2015 the V8 project authors. All rights reserved.
```
bakkot, line 20: possibly not? (ditto 50+ times below)

dminor: Yep, that looks like an artifact of the export process that's not present in the Firefox copies of the tests.

bakkot: Why are there two lines? I'm confused by these having 2015 authorship.

dminor: A lot of the tests were based on existing Map/WeakMap tests, so they kept the original copyright statement.

ptomato (approved):
Thanks!

This is staging, so you should feel free to (and already have the appropriate permissions to) commit these as long as they are executable by other engines' test262 harnesses.

That said, I suspect they will not pass in any other engines unless you add `includes: [sm/non262.js, sm/non262-shell.js]` to each one.

Optionally, there's also some low-hanging fruit that could make them slightly more ready to move out of staging when the time comes:
- Add a feature for the upsert proposal to `features.txt` and add it to `features` in the frontmatter of each test
- Use `verifyProperty` instead of `assert.deepEqual(Object.getOwnPropertyDescriptor(...etc))`
- Use `assert.throws(ExceptionType, function () { ... })` instead of `assertThrowsInstanceOf(function () { ... }, ExceptionType)`

dminor:
Thanks for the suggestions on how to fix these up, I'll look into that next week prior to committing.

nicolo-ribaudo:
> Remove cross-compartment tests

I haven't seen it tested before in test262, but if you want you could restore those tests using `$262.createRealm().evalScript`.

dminor:
@ptomato Hi Philip, I've addressed your suggestions and checked that the tests pass in engine262 (with my upsert PR) and SpiderMonkey. I don't have permissions to commit this, could you please have another look and if looks ok, commit it?

gibson042 (approved):

ptomato:
Sure thing. I didn't review these for completeness, just gave them a quick glance and executed them.

I found it puzzling that you didn't have permission to land this but then I realized it's because of the change to `features.txt`. My bad; I should have done that in a separate PR, then you would have been able to merge this yourself whenever it was ready. Sorry about that.

dminor:
No problem, thank you for the help :)

PR #4455, opened 2025-04-16
Add 'notation' option to PluralRules tests and update related assertions

ryzokuken:
Refs: https://github.com/tc39/ecma402/pull/989

### DIFF
--- a/test/intl402/PluralRules/prototype/constructor/notation.js
+++ b/test/intl402/PluralRules/prototype/constructor/notation.js
@@ -1,0 +1,25 @@
+// Copyright 2025 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-initializepluralrules
+description: Checks that the notation option is picked up correctly.
+info: |
+  Intl.PluralRules ( [ _locales_ [ , _options_ ] ] )
+    ...
+    1. Let notation be ? GetOption(options, "notation", "string", « "standard", "compact", "scientific", "engineering" », "standard").
+    ...
+---*/
+
+const validValues = ["standard", "compact", "scientific", "engineering", new String("standard"), new String("compact"), new String("scientific"), new String("engineering")];
+const invalidValues = ["COMPACT", "ståndard", 123, false, Symbol("foo"), null, {}, [], ""];
+
+for (const value of validValues) {
+  const pr = new Intl.PluralRules("en", { notation: value });
+  assert(pr.resolvedOptions().notation === value, `Resolved options should have notation ${value}`);
+}
+for (const value of invalidValues) {
+  assert.throws(RangeError, () => {
+    new Intl.PluralRules("en", { notation: value });
+  }, `Exception should be thrown for ${value}`);
+}
--- a/test/intl402/PluralRules/prototype/select/notation.js
+++ b/test/intl402/PluralRules/prototype/select/notation.js
@@ -1,0 +1,20 @@
+// Copyright 2025 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-pluralruleselect
+description: Checks that the notation option is used appropriately in the select method.
+info: |
+  PluralRuleSelect (
+    _locale_: a language tag,
+    _type_: *"cardinal"* or *"ordinal"*,
+    _notation_: a String,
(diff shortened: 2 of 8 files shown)


ptomato (commented):
Looks good, but on looking at the normative PR I think there should at least be a test that tries to construct a PluralRules with each valid value of `notation`, and verifies that the constructor throws on an invalid value such as `COMPACT` or `ståndard`.

ryzokuken:
@ptomato thanks for the review. I replaced the constructor test with what you suggested, moved the old test to `select` since it was testing that function's behavior anyway and edited another existing test that I thought could be further tested. I hope things look better now.

test/intl402/PluralRules/prototype/constructor/notation.js
```diff
@@ -0,0 +1,26 @@
...
+    ...
+    1. Let notation be ? GetOption(options, "notation", "string", « "standard", "compact", "scientific", "engineering" », "standard").
+    ...
+---*/
+
+const validValues = ["standard", "compact", "scientific", "engineering", new String("standard"), new String("compact"), new String("scientific"), new String("engineering")];
+const invalidValues = ["COMPACT", "ståndard", 123, false, Symbol("foo"), null, {}, [], ""];
+
+for (const value of validValues) {
+  assert.doesNotThrow(() => {
```
ptomato, line 18: `assert.doesNotThrow` is not a thing :smile: Either just `new Intl.PluralRules()` and do nothing with the created instance, or assert that it is of the correct type.

ptomato (approved):
Thanks, this looks like what I expected. Please do fix the assert function though.

test/intl402/PluralRules/prototype/select/notation.js
```diff
@@ -0,0 +1,20 @@
...
+    _type_: *"cardinal"* or *"ordinal"*,
+    _notation_: a String,
+    _s_: a decimal String,
+  ): *"zero"*, *"one"*, *"two"*, *"few"*, *"many"*, or *"other"*
+    ...
+    The returned String characterizes the plural category of _s_ according to _locale_, _type_ and _notation_.
+    ...
+---*/
+
+assert.sameValue(new Intl.PluralRules('fr-FR', { notation: 'compact' }).select(1.00000020e6), 'one', 'compact notation');
```
eemeli, line 19: Shouldn't this be `'many'`, given that the value formats the same as `1c6`?

eemeli: Actually, the [CLDR plural rule](https://github.com/unicode-org/cldr/blob/faa2818490a34a2288f307720ab0abdab8c91ff7/common/supplemental/plurals.xml#L123-L127) examples for `many` even include `1.0000001c6`.

eemeli (changes requested):

test/intl402/PluralRules/prototype/select/notation.js
ryzokuken: good point, I guess this applies to all related romance languages with a similar definition of `"many"` so the original example must've been unsound. Replacing it with a different language without that category like Slovene. Does that sound like a good idea?

eemeli (approved):

test/intl402/PluralRules/prototype/select/notation.js
```diff
@@ -0,0 +1,20 @@
...
+    _notation_: a String,
+    _s_: a decimal String,
+  ): *"zero"*, *"one"*, *"two"*, *"few"*, *"many"*, or *"other"*
+    ...
+    The returned String characterizes the plural category of _s_ according to _locale_, _type_ and _notation_.
+    ...
+---*/
+
+assert.sameValue(new Intl.PluralRules('sl', { notation: 'compact' }).select(1.00000020e6), 'one', 'compact notation');
+assert.sameValue(new Intl.PluralRules('sl', { notation: 'standard' }).select(1.00000020e6), 'other', 'standard notation');
```
trflynn89, line 20: Curious about these values as well - trying to validate our implementation in Ladybird.

For `notation=standard`, since that's effectively the default value, wouldn't the result be "few"? Chrome, Firefox, Safari and Ladybird all agree on:
```js
>> new Intl.PluralRules('sl').select(1.00000020e6)
"few" 
```

and ICU continues to give us "few" after setting the specific notation.

ICU is also giving us "other" for `notation=compact`, regardless of whether I use a compact display of short or long.

PR #4457, opened 2025-04-18
Test that string comparison is per-code-unit

andreubotella:
The Ecma-262 specification describes string comparison as comparing lexicographically a string's code units. This is different from comparing by code points (see the readme to the [compare strings by codepoint proposal](https://github.com/tc39/proposal-compare-strings-by-codepoint)).

However, the existing string comparison tests describe the spec's behavior as comparing by code points (which is wrong), and do not include any tests covering lone surrogates or characters greater than U+FFFF, which is when the difference appears. This patch fixes those descriptions, and adds some such missing tests.


Ms2ger (approved):
Those files could use a 21st century rewrite...

PR #4459, opened 2025-04-21
Add tests for "Runtime Errors for Function Call Assignment Targets"

rkirsling:
Add tests for normative PR https://github.com/tc39/ecma262/pull/3568, which adds Annex B section, "Runtime Errors for Function Call Assignment Targets".

Specifically:
1. Correct the name of `src/assignment-target-type/callexpression-arguments.case` to `callexpression.case` and mark it `onlyStrict`.
2. Add `onlyStrict` cases to `language/expressions/assignmenttargettype/` for `+=`, `++`, and `for`-`in`/`of`. Also add case for `&&=` which applies regardless of strict or sloppy.
3. Add `noStrict` cases to `annexB/language/expressions/assignment` for `=`, `+=`, `++`, and `for`-`in`/`of` which check that:
    - a ReferenceError is thrown
    - the function is called, but its return value is not resolved to a primitive
    - (`=` and `+=` only) the RHS is not evaluated
    
---
 
Note: All hosts have some work to do.
- V8: `onlyStrict` cases fail
- SM: "its return value is not resolved to a primitive" fails for `++` (but `+=` is OK)
- JSC: ~~"the function is called" is never true~~ (WebKit/WebKit#44364)


michaelficarra:
@rkirsling You should have tests where the function is named `async` to cover the case where the cover grammar is used.

ptomato:
Ready to merge whenever is convenient to synchronize with https://github.com/tc39/ecma262/pull/3568

rkirsling:
https://github.com/tc39/ecma262/pull/3568 has now been merged.

PR #4460, opened 2025-04-24
Add a testcase for consuming length argument first on TypedArray construction

DevSDK:
Implement missing test case to test for throwing an error when evaluating arguments first to make sure not to execute newTarget getter.

FYI: https://issues.chromium.org/u/2/issues/42209961


Ms2ger (approved):
Thanks!

Issue #4402, opened 2025-02-18
Deal with slow staging/sm tests


ptomato:
Some of the `staging/sm` tests take a long time to run. For this reason the whole folder is currently omitted from test262.fyi. We should find a way to either mark particular tests as slow so that consumers can skip them if necessary, or make them less slow.

See also #3328, #1231.

ptomato:
cc @CanadaHonk @Ms2ger

PR #4463, opened 2025-04-30
Start cleaning up staging/sm tests

anba:
Changes:
- Replace harness functions from harness/sm with standard test262 harness functions.
- Remove now unused harness functions from harness/sm.
- Both changes should help a bit for #4402, because there's now less harness code to execute.
- Avoid testing error messages, because error messages aren't standardised. 
- Remove unused files.
- Remove `noStrict` flag where possible. 
- And many more clean-ups and fixes.

I've also identified a couple of tests which should be removed from test262, because they test non-standard behaviour:
- Tests for `Function.caller`.
- `Date.parse` with non-standard inputs.
- Tests how accurate `Math` functions are.
- If the ToString output for bound functions contains the function name.
- Large RegExp tests.
- Cycle detection tests.
- Tests using SpiderMonkey shell functions like `disassemble` or `timeout`.

In a follow-up PR I can remove these tests, too.


ptomato (commented):
Reviewed up to "Remove non262-extensions-shell.js"

test/staging/sm/extensions/sps-generators.js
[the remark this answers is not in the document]
anba: The test boils down to:
```js
function* gen() {
  var x = yield 'hi';
  yield x;
  yield 'bye';
}

for (var unused of [ true, false ]) {
  g = gen();
  assert.sameValue(g.next().value, 'hi');
  assert.sameValue(g.next('gurgitating...').value, 'gurgitating...');
  for (var x of g)
    assert.sameValue(x, 'bye');
}
```

which is covered through various tests in "built-ins/GeneratorPrototype/next" and "language/statements/for-of".

test/staging/sm/strict/15.4.5.1.js
```diff
@@ -33,55 +33,12 @@ function strict1(out)
...
-  throw "no error";
-}
-catch (e)
-{
-  assert.sameValue(e instanceof TypeError, true, "expected TypeError, got " + e);
-}
-assert.sameValue(deepEqual(out.array, addx([1, 2, 3])), true);
+assert.throws(TypeError, function() {
+  strict1(out);
+});
```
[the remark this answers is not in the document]
anba: The file tests that modifying `length` fails for an array with a non-configurable array element. If we want to have a test to cover that arrays support non-index properties, we shouldn't use this file for that. (And we probably already have coverage that arrays support non-index properties.) 

The test itself was added in <https://bugzilla.mozilla.org/show_bug.cgi?id=537873>. And the "slow array" part became obsolete twelve years ago in <https://bugzilla.mozilla.org/show_bug.cgi?id=827490>.

test/staging/sm/generators/objects.js
```diff
@@ -22,10 +22,10 @@ function TestGeneratorObject() {
 
   var iter = g();
   assert.sameValue(Object.getPrototypeOf(iter), g.prototype);
-  assertTrue(iter instanceof g);
+  assert.sameValue(iter instanceof g, true);
```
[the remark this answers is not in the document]
ptomato: Oh, too bad. Well, let's keep it as is then, I don't want to conflict with others' comments.

test/staging/sm/Array/sort_small.js
```diff
@@ -2,23 +2,32 @@
...
-        let sorted = (Array.from(permutation)).sort(comparefn);
-        for (let i in sorted) {
-            assert.sameValue(sorted[i], data[i],
-            [`[${permutation}].sort(${comparefn})`,
-            `returned ${sorted}, expected ${data}`].join(' '));
-        }
+
+// Yield every permutation of the elements in some array.
+function* Permutations(items) {
+  if (items.length == 0) {
```
ptomato: Good practice to replace this with `===` while we're here

test/staging/sm/Array/sort_small.js
```diff
@@ -2,23 +2,32 @@
...
+        yield head.concat(e);
+      }
     }
+  }
+}
+
+function sortAllPermutations(data, comparefn) {
+  for (let permutation of Permutations(data)) {
+    let sorted = permutation.sort(comparefn);
+    assert.compareArray(sorted, data);
```
ptomato: `sorted` is in fact now the same array as `permutation`, better to just use `permutation` to avoid confusion

test/staging/sm/TypedArray/sort_small.js
```diff
@@ -2,50 +2,66 @@
...
 description: |
-  pending
+  Sort every possible permutation of some TypedArrays.
 esid: pending
+features: [TypedArray]
 ---*/
+
+// Yield every permutation of the elements in some array.
+function* Permutations(items) {
+  if (items.length == 0) {
```
ptomato: Ditto, replace with `===`

test/staging/sm/String/thisv-error.js
```diff
@@ -30,7 +29,7 @@ function testName(thisv) {
       assertThrowsInstanceOf(() => String.prototype[key].call(thisv), TypeError, key);
     } else {
       var expected = `String.prototype.${key} called on incompatible ${thisv}`;
-      assertThrowsInstanceOfWithMessage(() => String.prototype[key].call(thisv), TypeError, expected, key)
+      assert.throws(TypeError, () => String.prototype[key].call(thisv), TypeError, expected)
```
ptomato: extra `TypeError` in argument position 3

ptomato (commented):
Reviewed up to "Remove assertThrownErrorContains"

ptomato:
We discussed this in the test262 maintainers meeting today. We'll just go ahead and merge it; these are staging tests and so should not be held up for this long by review, and the improvements have been good so far. I'll address the existing comments, and there's one more thing I'd like to do — I'd like to convert the `BUGNUMBER` assignments to comments with links to Mozilla's bug tracker, rather than just removing them, as I think the history can be relevant to future consumers of these tests. After I do that, I'll merge it.

iamstolis:
The following tests started to fail after this PR:
* `staging/sm/Math/acosh-exact.js`
* `staging/sm/Math/atanh-exact.js`
* `staging/sm/Math/log1p-exact.js`
* `staging/sm/Math/trunc.js`
These tests use `ONE_MINUS_EPSILON` and/or `ONE_PLUS_EPSILON` but they do not declared dependency on `sm/non262-Math-shell.js` (where these constants are defined).

I have created #4588 to track this problem.

PR #4464, opened 2025-04-30
update tests after disallowing using and await using in switch cases

rmahdav:
The following PR got consensus in the latest TC39 meeting. After it got merge, these tests should be updated:
https://github.com/rbuckton/ecma262/pull/14


ptomato:
Thanks for making the PR! Apologies that it fell through the cracks.

ptomato:
(FWIW, anyone from one of the implementation teams should have permission to land changes to staging tests — if that's not the case please let us know)

rmahdav:
Thanks Philip!

PR #4465, opened 2025-05-02
Test a module depending on a non-root node of an asyn cycle

nicolo-ribaudo:
See https://github.com/tc39/ecma262/pull/3584#issuecomment-2846649818.

This case was not previously covered by any test. I tested it by updating engine262's implementation to not "Set _requiredModule_ to _requiredModule_.[[CycleRoot]]", and all existing test262 tests were still passing.


ptomato (approved):
I'm not sure if I completely understood this test, so a review from someone else would be helpful. But it looks correct.

test/language/module-code/top-level-await/pending-async-dep-from-cycle.js
```diff
@@ -0,0 +1,70 @@
...
+           ▼    │               │
+    ┌──────────────────┐        │
+    │ cycle leaf (TLA) │ ◄──────┘
+    └──────────────────┘
+
+  This test exercises step 11.c.iv.1 of the following algorithm when _module_ is
+  "importer of cycle leaf", _requiredModule_ is "cycle leaf (TLA)", and
+  _requiredModule_.[[CycleRoot]] is "cycle root (TLA)".
+  The [[Status]] of "cycle leaf (TLA)" and of "cycle root (TLA)" is ~evaluating-async~,
+  because they have alredy been traversed and they are blocked on the TLA in "cycle leaf (TLA)".
```
gibson042:
```suggestion
  because they have already been traversed and they are blocked on the TLA in "cycle leaf (TLA)".
```

PR #4468, opened 2025-05-05
harness: Add a verifyCallableProperty helper

gibson042:
[ECMAScript Standard Built-in Objects](https://tc39.es/ecma262/multipage/ecmascript-standard-built-in-objects.html#sec-ecmascript-standard-built-in-objects) specifies non-writable non-enumerable "length" and "name" properties for all built-in functions, but test262 coverage of this is a) incomplete, and b) noisy [in the form of `length.js` and `name.js` files for covered functions].

This PR addresses both issues by introducing a helper to be used in prop-desc.js files for relevant functions, covering their callability, property descriptor attributes, and length/name property descriptors (actually two helpers, with the one for primordial properties intended to be overridden in hardened environments with a version that requires non-configurability).

It is currently only used for global is{Finite,NaN} and parse{Int,Float} functions as a proof of concept (and to limit review burden), but should be capable of covering every function in the spec.


harness/propertyHelper.js
```diff
@@ -202,6 +208,56 @@ function isWritable(obj, name, verifyProp, value) {
...
+ * @param {object} [options]
+ * @param {boolean} [options.restore] revert mutations from verifying writable/configurable
+ */
+function verifyCallableProperty(obj, name, functionName, functionLength, desc, options) {
+  var value = obj[name];
+
+  assert.sameValue(typeof value, "function",
+    "obj['" + String(name) + "'] descriptor should be a function");
+
+  if (!__hasOwnProperty(desc, value)) desc.value = value;
```
Ms2ger, line 226: Did you mean
```suggestion
  if (!__hasOwnProperty(desc, "value")) desc.value = value;
```
?

test/built-ins/isFinite/prop-desc.js
```diff
@@ -8,7 +8,8 @@ description: >
 includes: [propertyHelper.js]
 ---*/
 
-verifyProperty(this, "isFinite", {
+verifyPrimordialCallableProperty(this, "isFinite", "isFinite", 1, {
+  value: this.isFinite,
```
Ms2ger, line 12: Let's not manually add these `value` properties

gibson042: I'm not 100% married to them, but this is not an original pattern and even appears in [INTERPRETING.md](https://github.com/tc39/test262/blob/main/INTERPRETING.md#includes).

ptomato (approved):
No opinion on redundant `value` so I'll defer to Ms2ger on that, but overall I love this helper

anba:
> test262 coverage of this is a) incomplete

Did we regress coverage that much since #457 and #505?

test/built-ins/isFinite/prop-desc.js
gibson042: Removed.

PR #4472, opened 2025-05-07
Move Map.prototype.getOrInsert tests from staging

dminor:


test/built-ins/Map/prototype/getOrInsert/append-new-values-normalizes-zero-key.js
ptomato, line 1: Here it'd be good to add `assert.sameValue(map.get(-0), 43);` at the end as well.

ptomato (approved):
Thanks, these are indeed already in good shape to move out of staging.

Before merging, in each file:
- Replace `esid: proposal-upsert` with `esid: sec-map.prototype.getorinsert`
- Remove the `noStrict` flag

ljharb (approved):
confirmed these tests pass on my polyfill as well

PR #4473, opened 2025-05-08
RegExp: Add test cases for Unicode Script_Extensions=Unknown

gibson042:
Ref https://github.com/tc39/ecma262/issues/3590

regexp-generator is unfortunately unaware of the value: https://github.com/mathiasbynens/regenerate-unicode-properties/issues/7

### DIFF
--- a/test/built-ins/RegExp/property-escapes/special-property-value-Script_Extensions-Unknown.js
+++ b/test/built-ins/RegExp/property-escapes/special-property-value-Script_Extensions-Unknown.js
@@ -1,0 +1,17 @@
+// Copyright 2025 Richard Gibson. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+author: Richard Gibson
+description: >
+  Unicode property "Script_Extensions" and alias "scx" must support
+  special value "Unknown" and alias "Zzzz" (cf.
+  https://www.unicode.org/reports/tr24/#Script_Extensions_Def ).
+esid: sec-compiletocharset
+features: [regexp-unicode-property-escapes]
+---*/
+
+/\p{Script_Extensions=Unknown}/u;
+/\p{Script_Extensions=Zzzz}/u;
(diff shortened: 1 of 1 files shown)


test/built-ins/RegExp/property-escapes/special-property-value-Script_Extensions-Unknown.js
```diff
@@ -0,0 +1,17 @@
...
+author: Richard Gibson
+description: >
+  Unicode property "Script_Extensions" and alias "scx" must support
+  special value "Unknown" and alias "Zzzz" (cf.
+  https://www.unicode.org/reports/tr24/#Script_Extensions_Def ).
+esid: sec-compiletocharset
+features: [regexp-unicode-property-escapes]
+---*/
+
+/\p{Script_Extensions=Unknown}/u;
```
Jack-Works, line 14: Does it mean to match anything that isn't in any Script?

gibson042: Yes, see https://www.unicode.org/reports/tr24/#Script_Extensions
> The empty set is not allowed; the scx value for unassigned, private use, and non-character code points is the set { **Unknown** }.

That aspect will be covered once a fix for https://github.com/mathiasbynens/regenerate-unicode-properties/issues/7 is published.

Jack-Works, line 14: Do you need to call `buildStrings` to test it like other sc/scx tests?

gibson042: No, that aspect will be covered once a fix for https://github.com/mathiasbynens/regenerate-unicode-properties/issues/7 is published. This is just supplemental regression coverage of validity.

ptomato (approved):
Thanks Richard for the live explanation of this during the maintainers meeting.

PR #4474, opened 2025-05-08
Add tests for Intl.Locale.prototype.variants

gibson042:
Ref tc39/ecma402#960

### DIFF
--- a/test/intl402/Locale/constructor-options-variants-invalid.js
+++ b/test/intl402/Locale/constructor-options-variants-invalid.js
@@ -1,0 +1,44 @@
+// Copyright 2025 Richard Gibson. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-intl.locale
+description: >
+    Checks error cases for the options argument to the Locale
+    constructor.
+info: |
+    Intl.Locale( tag [, options] )
+    12. Set _tag_ to ? UpdateLanguageId(_tag_, _options_).
+
+    UpdateLanguageId ( tag, options )
+    8. Let _variants_ be ? GetOption(_options_, *"variants"*, ~string~, ~empty~, GetLocaleVariants(_baseName_)).
+    9. If _variants_ is not *undefined*, then
+      a. If _variants_ cannot be matched by the <code>unicode_variant_subtag</code> Unicode locale nonterminal, throw a *RangeError* exception.
+
+features: [Intl.Locale]
+---*/
+
+/*
+ unicode_variant_subtag = (alphanum{5,8} | digit alphanum{3})
+*/
+const invalidVariantsOptions = [
+  "",
+  "a",
+  "1",
+  "ab",
+  "2x",
+  "abc",
+  "3xy",
+  "abcd",
+  "12345",
+  "5wxyz",
+  "abcdefghi",
+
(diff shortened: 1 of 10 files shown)


gibson042:
https://github.com/tc39/ecma402/pull/960 reached consensus in TC39 plenary, and has been merged into ECMA-402. This can be merged as soon as it has a review (which should be relatively low effort, especially if going by commit-by-commit).

test/intl402/Locale/constructor-options-variants-valid.js
```diff
@@ -0,0 +1,50 @@
...
+features: [Intl.Locale]
+---*/
+
+const validVariantsOptions = [
+  ['en', undefined, undefined],
+  ['en', 'spanglis', 'en-spanglis'],
+];
+for (const [lang, variants, baseName] of validVariantsOptions) {
+  let options = { variants };
+  let optionsRepr = `{variants: ${variants == null ? variants : `"${variants}"`}}`;
```
ptomato, line 27: Nitpick: the `==` operator here is very subtle but actually significant and needed. I overlooked it in several reads, and thought the code looked wrong. Would there be a more obvious way to write it?

ptomato (approved):

test/intl402/Locale/constructor-options-variants-invalid.js
```diff
@@ -0,0 +1,44 @@
...
+  "",
+  "a",
+  "1",
+  "ab",
+  "2x",
+  "abc",
+  "3xy",
+  "abcd",
+  "12345",
+  "5wxyz",
```
anba, line 34: `"12345"` and `"5wxyz"` are both valid variant subtags.

test/intl402/Locale/constructor-options-variants-invalid.js
```diff
@@ -0,0 +1,44 @@
...
+  "abcdefghi",
+
+  // Value contains more than just the 'variants' production.
+  "GB-scouse",
+];
```
anba, line 39: Can we also test for leading, trailing, and consecutive `"-"`?

```js
const alsoInvalid = [
  "-",
  "-spanglis",
  "spanglis-",
  "-spanglis-oxendict",
  "spanglis-oxendict-",
  "spanglis--oxendict",
];
```

anba (changes requested):
And maybe also test canonicalisation works correctly when the `variants` option is used:

1. Variant subtags are correctly sorted:
```js
js> new Intl.Locale("en", {variants: "spanglis-oxendict"}).toString()
"en-oxendict-spanglis"
```

2. Deprecated language tags are canonicalised to their modern form:
```js
js> new Intl.Locale("cel", {variants: "gaulish"}).toString()           
"xtg"
```

test/intl402/Locale/constructor-options-variants-invalid.js
anba: And test that duplicate variants throw:
```js
js> new Intl.Locale("en", {variants: "spanglis-spanglis"})
typein:1:1 RangeError: duplicate variant subtag
```

anba:
LGTM! I've confirmed all tests pass with the patches from <https://bugzilla.mozilla.org/show_bug.cgi?id=1970161> applied.

PR #4476, opened 2025-05-09
Move WeakMap.prototype.getOrInsert tests from staging to built-ins

dminor:


ljharb (approved):
verified with my polyfill

ljharb:
yup, i'll land it after a second test262 maintainer stamps it :-)

test/built-ins/WeakMap/prototype/getOrInsert/length.js
ptomato, line 1: Could you remove `noStrict` from this one as well?

test/built-ins/WeakMap/prototype/getOrInsert/returns-value-if-key-is-not-present-symbol-key.js
ptomato, line 1: Nitpick: the assertion messages "Regular symbol as key, added in constructor" and "Regular symbol as key, added with set()" don't apply to this test. (I might have missed that in the review of the Map tests...)

ptomato (approved):
Two minor comments, then let's merge as far as I'm concerned. Thanks!

PR #4478, opened 2025-05-12
[explicit-resource-management] Add remaining tests specific to `AsyncDisposableStack`

rbuckton:
In an effort to make review more manageable, this extracts the remaining tests specific to `AsyncDisposableStack` from #3866

### DIFF
--- a/test/built-ins/AsyncDisposableStack/prototype/constructor.js
+++ b/test/built-ins/AsyncDisposableStack/prototype/constructor.js
@@ -1,0 +1,31 @@
+// Copyright (C) 2023 Ron Buckton. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-properties-of-the-asyncdisposablestack-prototype-object
+description: AsyncDisposableStack.prototype.constructor
+info: |
+  AsyncDisposableStack.prototype.constructor
+
+  Normative Optional
+
+  The initial value of AsyncDisposableStack.prototype.constructor is the intrinsic object %AsyncDisposableStack%.
+
+  This property has the attributes { [[Writable]]: false, [[Enumerable]]: false, [[Configurable]]: true }.
+
+  This section is to be treated identically to the "Annex B" of ECMA-262, but to be written in-line with the main specification.
+includes: [propertyHelper.js]
+features: [explicit-resource-management]
+---*/
+
+var actual = AsyncDisposableStack.prototype.hasOwnProperty('constructor');
+
+// If implemented, it should conform to the spec text
+if (actual) {
+  verifyProperty(AsyncDisposableStack.prototype, 'constructor', {
+    value: AsyncDisposableStack,
+    writable: true,
+    enumerable: false,
+    configurable: true
+  });
+}
--- a/test/built-ins/AsyncDisposableStack/prototype/disposeAsync/disposes-resources-in-reverse-order.js
+++ b/test/built-ins/AsyncDisposableStack/prototype/disposeAsync/disposes-resources-in-reverse-order.js
@@ -1,0 +1,77 @@
+// Copyright (C) 2023 Ron Buckton. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-asyncdisposablestack.prototype.disposeAsync
+description: Added resources are disposed in reverse order
+info: |
+  AsyncDisposableStack.prototype.disposeAsync ( )
+
+  1. Let asyncDisposableStack be the this value.
+  2. Let promiseCapability be ! NewPromiseCapability(%Promise%).
+  3. If asyncDisposableStack does not have an [[AsyncDisposableState]] internal slot, then
+    a. Perform ! Call(promiseCapability.[[Reject]], undefined, « a newly created TypeError object »).
+    b. Return promiseCapability.[[Promise]].
+  4. If asyncDisposableStack.[[AsyncDisposableState]] is disposed, then
+    a. Perform ! Call(promiseCapability.[[Resolve]], undefined, « undefined »).
+    b. Return promiseCapability.[[Promise]].
+  5. Set asyncDisposableStack.[[AsyncDisposableState]] to disposed.
+  6. Let result be DisposeResources(asyncDisposableStack.[[DisposeCapability]], NormalCompletion(undefined)).
+  7. IfAbruptRejectPromise(result, promiseCapability).
+  8. Perform ! Call(promiseCapability.[[Resolve]], undefined, « result »).
+  9. Return promiseCapability.[[Promise]].
+
+  DisposeResources ( disposeCapability, completion )
+
+  1. For each resource of disposeCapability.[[DisposableResourceStack]], in reverse list order, do
+    a. Let result be Dispose(resource.[[ResourceValue]], resource.[[Hint]], resource.[[DisposeMethod]]).
+    b. If result.[[Type]] is throw, then
+      i. If completion.[[Type]] is throw, then
+        1. Set result to result.[[Value]].
+        2. Let suppressed be completion.[[Value]].
+        3. Let error be a newly created SuppressedError object.
+        4. Perform ! CreateNonEnumerableDataPropertyOrThrow(error, "error", result).
+        5. Perform ! CreateNonEnumerableDataPropertyOrThrow(error, "suppressed", suppressed).
+        6. Set completion to ThrowCompletion(error).
+      ii. Else,
+        1. Set completion to result.
+  2. Return completion.
+
+  Dispose ( V, hint, method )
+
+  1. If method is undefined, let result be undefined.
+  2. Else, let result be ? Call(method, V).
+  3. If hint is async-dispose, then
+    a. Perform ? Await(result).
+  4. Return undefined.
+
+flags: [async]
+includes: [asyncHelpers.js]
+features: [explicit-resource-management]
+---*/
+
+asyncTest(async function () {
+  var stack = new AsyncDisposableStack();
+  var disposed = [];
+  var resource1 = { async [Symbol.asyncDispose]() { disposed.push(this); } };
+  var resource2 = { [Symbol.dispose]() { disposed.push(this); } };
+  var resource3 = {};
+  async function dispose3(res) { disposed.push(res); }
+  var resource4 = {};
+  function dispose4(res) { disposed.push(res); }
+  async function dispose5() { disposed.push(dispose5); }
+  function dispose6() { disposed.push(dispose6); }
+  stack.use(resource1);
+  stack.use(resource2);
+  stack.adopt(resource3, dispose3);
+  stack.adopt(resource4, dispose4);
+  stack.defer(dispose5);
+  stack.defer(dispose6);
+  await stack.disposeAsync();
+  assert.sameValue(disposed[0], dispose6);
+  assert.sameValue(disposed[1], dispose5);
+  assert.sameValue(disposed[2], resource4);
+  assert.sameValue(disposed[3], resource3);
+  assert.sameValue(disposed[4], resource2);
+  assert.sameValue(disposed[5], resource1);
+});
--- a/test/built-ins/AsyncDisposableStack/prototype/disposeAsync/does-not-reinvoke-disposers-if-already-disposed.js
+++ b/test/built-ins/AsyncDisposableStack/prototype/disposeAsync/does-not-reinvoke-disposers-if-already-disposed.js
@@ -1,0 +1,58 @@
+// Copyright (C) 2023 Ron Buckton. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-asyncdisposablestack.prototype.disposeAsync
+description: Does not re-invoke disposal on resources after stack has already been disposed.
+info: |
+  AsyncDisposableStack.prototype.disposeAsync ( )
+
+  1. Let asyncDisposableStack be the this value.
+  2. Perform ? RequireInternalSlot(asyncDisposableStack, [[AsyncDisposableState]]).
+  3. If asyncDisposableStack.[[AsyncDisposableState]] is disposed, return undefined.
+  4. Set asyncDisposableStack.[[AsyncDisposableState]] to disposed.
+  5. Return DisposeResources(asyncDisposableStack.[[DisposeCapability]], NormalCompletion(undefined)).
+
+  DisposeResources ( disposeCapability, completion )
+
+  1. For each resource of disposeCapability.[[DisposableResourceStack]], in reverse list order, do
+    a. Let result be Dispose(resource.[[ResourceValue]], resource.[[Hint]], resource.[[DisposeMethod]]).
+    b. If result.[[Type]] is throw, then
+      i. If completion.[[Type]] is throw, then
+        1. Set result to result.[[Value]].
+        2. Let suppressed be completion.[[Value]].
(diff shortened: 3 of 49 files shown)


ptomato:
We've tried to divide this up between the folks present in the maintainers meeting, in order to land it:
- [x] `disposeAsync` and `disposed`: @jugglinmike 
- [x] `use`: @gibson042 
- [x] `defer` and `move`: @ioannad 
- [x] `adopt`, `constructor.js`, `toStringTag.js`: @ptomato

test/built-ins/AsyncDisposableStack/prototype/constructor.js
```diff
@@ -0,0 +1,31 @@
...
+info: |
+  AsyncDisposableStack.prototype.constructor
+
+  Normative Optional
+
+  The initial value of AsyncDisposableStack.prototype.constructor is the intrinsic object %AsyncDisposableStack%.
+
+  This property has the attributes { [[Writable]]: false, [[Enumerable]]: false, [[Configurable]]: true }.
+
+  This section is to be treated identically to the "Annex B" of ECMA-262, but to be written in-line with the main specification.
```
ptomato: This seems to contradict https://github.com/rbuckton/ecma262/pull/13 - I don't see this marked as normative optional in the current spec PR. Should this test just be the `verifyProperty` call by itself?

(And it looks like I missed this when reviewing `DisposableStack/prototype/constructor.js`)

ptomato (commented):
I reviewed `adopt/`, `constructor.js`, and `toStringTag.js`. One comment on `constructor.js`, but I may just be confused about that.

I'd also suggest adding a test that exercises the following:
```js
const stack = new AsyncDisposableStack();
await stack.disposeAsync();
stack.adopt({}, async () => {});
  // Should throw a ReferenceError
```
I see this exists already for move() in `move/throws-if-disposed.js`, but could be added for adopt() and defer(), as well as for DisposableStack. If you want to do that in a separate PR I could review and merge that independently of this PR.

test/built-ins/AsyncDisposableStack/prototype/constructor.js
rbuckton: Likely a copy/paste transcription error. I'll fix both.

rbuckton:
> If you want to do that in a separate PR I could review and merge that independently of this PR.

Added here, hope that's not too much noise. If it is I can revert and pull that out to a separate PR.

test/built-ins/AsyncDisposableStack/prototype/disposeAsync/does-not-reinvoke-disposers-if-already-disposed.js
```diff
@@ -0,0 +1,58 @@
...
+
+asyncTest(async function () {
+  var stack = new AsyncDisposableStack();
+  var useCount = 0;
+  var adoptCount = 0;
+  var deferCount = 0;
+  stack.use({ async [Symbol.asyncDispose]() { useCount++; } });
+  stack.adopt({}, _ => { adoptCount++; });
+  stack.defer(() => { deferCount++; });
+  await stack.disposeAsync();
```
jugglinmike, line 53:
```suggestion
  await stack.disposeAsync();
  assert.sameValue(useCount, 1);
  assert.sameValue(adoptCount, 1);
  assert.sameValue(deferCount, 1);
```

test/built-ins/AsyncDisposableStack/prototype/disposeAsync/does-not-reject-if-already-disposed.js
```diff
@@ -0,0 +1,51 @@
...
+
+flags: [async]
+includes: [asyncHelpers.js]
+features: [explicit-resource-management]
+---*/
+
+asyncTest(async function () {
+  var stack = new AsyncDisposableStack();
+  var p = stack.disposeAsync();
+  await stack.disposeAsync();
```
jugglinmike: We should verify the fulfillment value since it's specified by a distinct algorithm step.

test/built-ins/AsyncDisposableStack/prototype/disposeAsync/rejects-with-suppressederror-if-multiple-errors-during-disposal.js
```diff
@@ -0,0 +1,65 @@
...
+  class MyError extends Error {}
+  var error1 = new MyError();
+  var error2 = new MyError();
+  var error3 = new MyError();
+  var stack = new AsyncDisposableStack();
+  stack.defer(function () { throw error1; });
+  stack.defer(function () { throw error2; });
+  stack.defer(function () { throw error3; });
+  try {
+    await stack.disposeAsync();
```
jugglinmike: To make sure the test doesn't spuriously pass on synchronous exceptions:

```suggestion
  var promise = stack.disposeAsync();
  try {
    await promise;
```

test/built-ins/AsyncDisposableStack/prototype/disposeAsync/rejects-with-suppressederror-if-multiple-errors-during-disposal.js
```diff
@@ -0,0 +1,65 @@
...
+    assert(false, 'Expected await stack.disposeAsync() to have thrown an error.');
+  }
+  catch (e) {
+    assert(e instanceof SuppressedError, "Expected await stack.disposeAsync() to have thrown a SuppressedError");
```
jugglinmike: `instanceof` is a little imprecise for our needs. Let's check the prototype so that the test rejects subclasses of `SuppressedError`.

test/built-ins/AsyncDisposableStack/prototype/disposeAsync/returns-promise.js
```diff
@@ -0,0 +1,44 @@
...
+  2. Else, let result be ? Call(method, V).
+  3. If hint is async-dispose, then
+    a. Perform ? Await(result).
+  4. Return undefined.
+
+features: [explicit-resource-management]
+---*/
+
+var stack = new AsyncDisposableStack();
+assert(stack.disposeAsync() instanceof Promise);
```
jugglinmike: Same thing regarding `instanceof` versus prototype comparison.

test/built-ins/AsyncDisposableStack/prototype/disposeAsync/disposes-resources-in-reverse-order.js
```diff
@@ -0,0 +1,77 @@
...
+  stack.adopt(resource4, dispose4);
+  stack.defer(dispose5);
+  stack.defer(dispose6);
+  await stack.disposeAsync();
+  assert.sameValue(disposed[0], dispose6);
+  assert.sameValue(disposed[1], dispose5);
+  assert.sameValue(disposed[2], resource4);
+  assert.sameValue(disposed[3], resource3);
+  assert.sameValue(disposed[4], resource2);
+  assert.sameValue(disposed[5], resource1);
```
jugglinmike:
```suggestion
  assert.sameValue(disposed[5], resource1);
  assert.sameValue(disposed.length, 6);
```

Or use `assert.compareArray`

test/built-ins/AsyncDisposableStack/prototype/disposeAsync/rejects-with-suppressederror-if-multiple-errors-during-disposal.js
rbuckton: Is this necessary? We use `e instanceof SomeError` with `assert` in hundreds of places throughout the tests.

rbuckton: I see we use both, apparently. I see ~1100 cases of `instanceof` and just under 1400 cases of `Object.getPrototypeOf(...)` throughout all of the tests, so it seems we're fairly inconsistent.

jugglinmike:
I've suggested a few tests for the "Await(undefined)" in the DisposeResources algorithm: https://github.com/rbuckton/test262/pull/1

rbuckton:
@gibson042, @ioannad any chance on final reviews this week? It's very possible I will be able to wrap up the ecma262 PR before plenary, which would leave this review as the final blocker to Stage 4.

test/built-ins/AsyncDisposableStack/prototype/move/returns-new-asyncdisposablestack-that-contains-moved-resources.js
```diff
@@ -0,0 +1,41 @@
...
+
+  var stack2 = stack1.move();
+
+  var wasDisposed = disposed.slice();
+  await stack2.disposeAsync();
+  var isDisposed = disposed.slice();
+
+  assert.deepEqual(wasDisposed, []);
+  assert.deepEqual(isDisposed, [2, 1]);
+
```
ioannad:
```suggestion
```

nit

ioannad:
TLDR: `defer` and `move` LGTM with a couple of optional additions that are covered in staging, and an extra new line nit.


------

For step 3 of `defer` and `move` I think we also need async tests, to verify that we don't have a race condition, where [[AsyncDisposableState]] hasn't changed yet. There are such tests in staging [this one for defer](https://github.com/tc39/test262/blob/main/test/staging/explicit-resource-management/async-disposable-stack-defer-on-disposed-stack.js) and [this one for move](https://github.com/tc39/test262/blob/main/test/staging/explicit-resource-management/async-disposable-stack-move-on-disposed-stack.js)) that test exactly that. So I think these could be moved from staging in a different PR, along with the similarly named staging test files for `adopt` and `dispose`.

Otherwise the tests seem to completely cover the spec steps for `defer` and `move`, thanks!

It would be nice to optionally also have some basic funcionality tests (these are also in staging [for defer](https://github.com/tc39/test262/blob/main/test/staging/explicit-resource-management/async-disposable-stack-defer.js) and [for move](https://github.com/tc39/test262/blob/main/test/staging/explicit-resource-management/async-disposable-stack-move.js)). 

I was also going to suggest a couple of exception propagation tests, but I see these are covered in `disposeAsync`. :)

rbuckton:
Is there a process in place for moving tests out of staging in general? Is there a reason we don't move all of the resource management tests out of staging?

test/built-ins/AsyncDisposableStack/prototype/use/allows-null-value.js
```diff
@@ -0,0 +1,28 @@
...
+    a. If V is either null or undefined and hint is sync-dispose, then
+      i. Return unused
+    ...
+  ...
+
+features: [explicit-resource-management]
+---*/
+
+var stack = new AsyncDisposableStack();
+stack.use(null);
```
gibson042: I'm wondering if this should have `flags: [async]` and `includes: [asyncHelpers.js]` so it can `await stack.disposeAsync()` to verify that `stack.use(null)` does not cause it to throw an exception or return a rejected promise.

test/built-ins/AsyncDisposableStack/prototype/use/allows-undefined-value.js
```diff
@@ -0,0 +1,28 @@
...
+    a. If V is either null or undefined and hint is sync-dispose, then
+      i. Return unused
+    ...
+  ...
+
+features: [explicit-resource-management]
+---*/
+
+var stack = new AsyncDisposableStack();
+stack.use(undefined);
```
gibson042: I'm wondering if this should have `flags: [async]` and `includes: [asyncHelpers.js]` so it can `await stack.disposeAsync()` to verify that `stack.use(undefined)` does not cause it to throw an exception or return a rejected promise.

test/built-ins/AsyncDisposableStack/prototype/use/gets-value-Symbol.dispose-property-after-trying-Symbol.asyncDispose.js
```diff
@@ -0,0 +1,65 @@
...
+        order.push('Symbol.asyncDispose');
+        return undefined;
+    },
+    get [Symbol.dispose]() {
+        order.push('Symbol.dispose');
+        return function() {};
+    },
+};
+stack.use(resource);
+assert.deepEqual(order, ['Symbol.asyncDispose', 'Symbol.dispose']);
```
gibson042: I think this file _should_ have `flags: [async]` and `includes: [asyncHelpers.js]` so it can `await stack.disposeAsync()` to verify that `order` is not affected.

test/built-ins/AsyncDisposableStack/prototype/use/adds-async-disposable-value.js
gibson042, line 1: This test seems redundant with test/built-ins/AsyncDisposableStack/prototype/use/Symbol.asyncDispose-getter.js , which AFAICT covers a strict superset. I don't strongly object, but wanted to call it out regardless.

And similarly for test/built-ins/AsyncDisposableStack/prototype/use/adds-sync-disposable-value.js w.r.t. test/built-ins/AsyncDisposableStack/prototype/use/Symbol.dispose-getter.js .

test/built-ins/AsyncDisposableStack/prototype/use/throws-if-value-Symbol.asyncDispose-property-is-null-or-undefined.js
gibson042, line 1: I'd prefer `undefined` and `null` to be asserted in the same file, and wouldn't object to that file also including the not-callable cases (currently test/built-ins/AsyncDisposableStack/prototype/use/throws-if-value-Symbol.asyncDispose-property-not-callable.js ), but don't object to keeping this set of files as-is either.

And likewise for the `Symbol.dispose` analogs.

test/built-ins/AsyncDisposableStack/prototype/use/adds-async-disposable-value.js
rbuckton: While there is apparent overlap, one tests that it works as a standard method/data property while the other tests that it works as a function-returning getter. I'd prefer to keep them both in despite the overlap.

rbuckton:
> For step 3 of `defer` and `move` I think we also need async tests, to verify that we don't have a race condition, where [[AsyncDisposableState]] hasn't changed yet. There are such tests in staging [this one for defer](https://github.com/tc39/test262/blob/main/test/staging/explicit-resource-management/async-disposable-stack-defer-on-disposed-stack.js) and [this one for move](https://github.com/tc39/test262/blob/main/test/staging/explicit-resource-management/async-disposable-stack-move-on-disposed-stack.js)) that test exactly that. So I think these could be moved from staging in a different PR, along with the similarly named staging test files for `adopt` and `dispose`.
> 
> Otherwise the tests seem to completely cover the spec steps for `defer` and `move`, thanks!
> 
> It would be nice to optionally also have some basic funcionality tests (these are also in staging [for defer](https://github.com/tc39/test262/blob/main/test/staging/explicit-resource-management/async-disposable-stack-defer.js) and [for move](https://github.com/tc39/test262/blob/main/test/staging/explicit-resource-management/async-disposable-stack-move.js)).

Done. See #4991.

ptomato:
> Is there a process in place for moving tests out of staging in general? Is there a reason we don't move all of the resource management tests out of staging?

Sorry, didn't notice this comment before. The reason we don't move them automatically is because tests in staging are considered best-effort for correctness, and are committed without review. So they need to be reviewed for correctness, and if incomplete should be completed at the same time.

The process is just to open a PR as you've done in #4991.

ptomato:
Looks like all the reviews are complete and all the review comments have been addressed, I'll merge this after lunch unless I hear otherwise.

PR #4479, opened 2025-05-12
[explicit-resource-management] Add remaining tests specific to `DisposableStack`

rbuckton:
In an effort to make review more manageable, this extracts the remaining tests specific to `DisposableStack` from #3866


ljharb (approved):
Verified with my polyfill

PR #4482, opened 2025-05-12
[explicit-resource-management] Add remaining tests specific to using statement semantics

rbuckton:
In an effort to make review more manageable, this extracts the remaining tests specific to `using` statement semantics from #3866

### DIFF
--- a/test/language/statements/using/initializer-disposed-at-end-of-switchstatement.js
+++ b/test/language/statements/using/initializer-disposed-at-end-of-switchstatement.js
@@ -1,0 +1,57 @@
+// Copyright (C) 2023 Ron Buckton. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-switch-statement-runtime-semantics-evaluation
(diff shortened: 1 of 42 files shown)


test/language/statements/using/initializer-disposed-at-end-of-switchstatement.js
```diff
@@ -0,0 +1,57 @@
...
+var resource = {
+    disposed: false,
+    [Symbol.dispose]() {
+        this.disposed = true;
+    }
+};
+
+switch (0) {
+  default:
+    using _ = resource;
```
arai-a: This should now be disallowed, by https://github.com/rbuckton/ecma262/pull/14

arai-a (commented):
tested on SpiderMonkey, and one test fails

rbuckton:
Thanks @arai-a, should be fixed now.

ptomato:
Since this has been reviewed by someone involved in the implementation, I would be comfortable to merge it now, to avoid holding it up any further. I personally won't have the bandwidth to get around to it in the near future. @gibson042 @ljharb @Ms2ger What do you think?

PR #4483, opened 2025-05-12
[explicit-resource-management] Add remaining tests specific to using statement syntax

rbuckton:
In an effort to make review more manageable, this extracts the remaining tests specific to `using` statement syntax from #3866

### DIFF
--- a/test/language/statements/using/syntax/using-invalid-assignment-next-expression-for.js
+++ b/test/language/statements/using/syntax/using-invalid-assignment-next-expression-for.js
@@ -1,0 +1,13 @@
+// Copyright (C) 2023 Ron Buckton. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: pending
+description: >
+    using: invalid assignment in next expression
+features: [explicit-resource-management]
+---*/
+
+assert.throws(TypeError, function() {
+  for (using i = 0; i < 1; i++) {}
+});
--- a/test/language/statements/using/syntax/using-invalid-assignment-statement-body-for-of.js
+++ b/test/language/statements/using/syntax/using-invalid-assignment-statement-body-for-of.js
@@ -1,0 +1,13 @@
+// Copyright (C) 2023 Ron Buckton. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: pending
+description: >
+    using: invalid assignment in Statement body
+features: [explicit-resource-management]
+---*/
+
+assert.throws(TypeError, function() {
+  for (using x of [1, 2, 3]) { x++ }
+});
--- a/test/language/statements/using/syntax/using-invalid-for-using-of-of.js
+++ b/test/language/statements/using/syntax/using-invalid-for-using-of-of.js
@@ -1,0 +1,14 @@
+// Copyright (C) 2011 the V8 project authors. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: pending
+description: >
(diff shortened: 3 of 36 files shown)


test/language/statements/using/syntax/using-invalid-for-using-of-of.js
```diff
@@ -0,0 +1,14 @@
...
+description: >
+    using: 'for (using of' is always interpreted as identifier 
+negative:
+  phase: parse
+  type: SyntaxError
+features: [explicit-resource-management]
+---*/
+
+$DONOTEVALUATE();
+for (using of of [1, 2, 3]) { }
```
JLHwung, line 14: Could you add a new test case when `for (using of` starts a for-variable-declaration, e.g.
```js
for(using of = null;;;);
```

This should be valid according to current spec, but a few implementations reject it as if it always start a for-of statement.

c.f. https://github.com/tc39/proposal-explicit-resource-management/issues/248

rbuckton: Fixed in [fc3079a](https://github.com/tc39/test262/pull/4483/commits/fc3079a585d827fdd094f323674e1e2d66f531d4)

ptomato (commented):
Since I just reviewed the `await using` PR, this went fairly quickly. Please apply all the comments there to this PR as well.

test/language/statements/using/syntax/using-invalid-assignment-next-expression-for.js
```diff
@@ -2,12 +2,13 @@
...
+esid: sec-declarative-environment-records-setmutablebinding-n-v-s
 description: >
-    using: invalid assignment in next expression
+    using: invalid assignment in next expression.  Since a `using` declaration introduces an immutable
+    binding, any attempt to change it results in a ReferenceError.
 features: [explicit-resource-management]
 ---*/
 
-assert.throws(TypeError, function() {
+assert.throws(ReferenceError, function() {
```
ptomato: Same question about ReferenceError vs TypeError as in the `await using` PR.

test/language/statements/using/syntax/using-invalid-assignment-statement-body-for-of.js
```diff
@@ -2,12 +2,13 @@
...
+esid: sec-declarative-environment-records-setmutablebinding-n-v-s
 description: >
-    using: invalid assignment in Statement body
+    using: invalid assignment in Statement body. Since a `using` declaration introduces an immutable
+    binding, any attempt to change it results in a ReferenceError.
 features: [explicit-resource-management]
 ---*/
 
-assert.throws(TypeError, function() {
+assert.throws(ReferenceError, function() {
```
ptomato: Ditto.

test/language/statements/using/syntax/using-invalid-assignment-next-expression-for.js
rbuckton: The test implementation is the problem as the initial value isn't disposable so I've pushed an update. What this should be testing is an attempt to reassign a constant, which should produce a `ReferenceError`.

test/language/statements/using/syntax/using-invalid-assignment-statement-body-for-of.js
rbuckton: I've fixed the test implementation as I did above.

test/language/statements/using/syntax/using-invalid-assignment-next-expression-for.js
ptomato: Discussed in chat that it should be a TypeError, which the PR now reflects. Merging, thanks!

PR #4484, opened 2025-05-12
[explicit-resource-management] Add remaining tests specific to using in for-of and for-await-of syntax

rbuckton:
In an effort to make review more manageable, this extracts the remaining tests specific to `using`/`await using` in `for-of` and `for-await-of` statements from #3866

### DIFF
--- a/test/language/statements/for-of/head-await-using-bound-names-fordecl-tdz.js
+++ b/test/language/statements/for-of/head-await-using-bound-names-fordecl-tdz.js
@@ -1,0 +1,17 @@
+// Copyright (C) 2023 Ron Buckton. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 13.6.4.12_S2
+description: >
+    ForIn/Of: Bound names of ForDeclaration are in TDZ (for-of)
+flags: [async]
+includes: [asyncHelpers.js]
+features: [explicit-resource-management]
+---*/
+
+asyncTest(async function () {
+  await assert.throwsAsync(ReferenceError, async function() {
+    let x = { async [Symbol.asyncDispose]() { } };
+    for (await using x of [x]) {}
+  });
+});
--- a/test/language/statements/for-of/head-await-using-bound-names-fordecl-tdz.js
+++ b/test/language/statements/for-of/head-await-using-bound-names-fordecl-tdz.js
@@ -1,7 +1,7 @@
 // Copyright (C) 2023 Ron Buckton. All rights reserved.
 // This code is governed by the BSD license found in the LICENSE file.
 /*---
-es6id: 13.6.4.12_S2
+esid: sec-for-in-and-for-of-statements
 description: >
     ForIn/Of: Bound names of ForDeclaration are in TDZ (for-of)
 flags: [async]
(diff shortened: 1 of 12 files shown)


rbuckton:
@syg do I recall correctly that you found a bug in one of the `for-of`/`for-await-of` tests? I recall seeing something a few weeks ago but I cannot seem to find the message.

syg:
Sorry, I don't remember either.

test/language/statements/for-of/head-await-using-bound-names-fordecl-tdz.js
```diff
@@ -0,0 +1,17 @@
+// Copyright (C) 2023 Ron Buckton. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+es6id: 13.6.4.12_S2
```
ptomato, line 4: We're not using `es6id` anymore so this should be replaced by `esid` with the section ID anchor. (Please check in several other tests as well)

test/language/statements/for-of/head-await-using-bound-names-fordecl-tdz.js
```diff
@@ -0,0 +1,17 @@
...
+flags: [async]
+includes: [asyncHelpers.js]
+features: [explicit-resource-management]
+---*/
+
+asyncTest(async function () {
+  await assert.throwsAsync(ReferenceError, async function() {
+    let x = { async [Symbol.asyncDispose]() { } };
+    for (await using x of [x]) {}
+  });
```
ptomato, line 16: I'm probably misunderstanding this due to my unfamiliarity with this part of the spec, but where is this ReferenceError thrown? I'm seeing in step 9.h.iii of ForIn/OfBodyEvaluation, we create a new declarative environment that already inherits the outer environment's `let x` binding in it from line 14. Then in ForDeclarationBindingInstantiation we do CreateImmutableBinding with `x` on the environment, which asserts that the environment does not already contain `x`.

ptomato (changes requested):
Thanks! Just needs the `es6id` entries replaced with `esid` section names.

I also had one question about `head-{await-,}using-bound-names-fordecl-tdz.js`, but that is probably due to me not understanding what exactly is happening in the spec.

test/language/statements/for-of/head-await-using-bound-names-fordecl-tdz.js
rbuckton: Due to TDZ, the `x` is looked up in the same scope as the `await using x`. Since `x` has been bound but not initialized, this results in a ReferenceError.

Binding occurs in [14.7.5.4 Runtime Semantics: ForDeclarationBindingInstantiation](https://arai-a.github.io/ecma262-compare/?pr=3000&id=sec-runtime-semantics-fordeclarationbindinginstantiation). The `ReferenceError` is thrown in [9.1.1.1.6 GetBindingValue (N, S)](https://tc39.es/ecma262/#sec-declarative-environment-records-getbindingvalue-n-s), step 2.

ptomato (approved):

rbuckton:
> Sorry, I don't remember either.

Ah, I found it. This is the test that is incorrect: [test/language/statements/using/syntax/using-invalid-for-using-of-of.js](https://github.com/tc39/test262/pull/4483/files/57bc079697f6cb7009da8ef029992c83855c120b#diff-f012caf0a9ec769e44e2f4d6fd302a7f2de9e1a6f53e94f87382cf6942f8717e)

PR #4485, opened 2025-05-13
add test for tc39/proposal-iterator-sequencing#19

michaelficarra:
See tc39/proposal-iterator-sequencing#19 for more info.

### DIFF
--- a/test/built-ins/Iterator/concat/get-value-after-done.js
+++ b/test/built-ins/Iterator/concat/get-value-after-done.js
@@ -1,0 +1,51 @@
+// Copyright (C) 2025 Michael Ficarra. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-iterator.concat
+description: >
+  Iterator.concat accesses the value of a done IteratorResult, matching the behaviour of yield*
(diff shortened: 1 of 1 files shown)


test/built-ins/Iterator/concat/get-value-after-done.js
```diff
@@ -0,0 +1,51 @@
...
+          done: true,
+        };
+      },
+    };
+  }
+};
+
+Array.from(function*() { yield* iter; }());
+
+assert.sameValue(valueAccesses, 1, 'yield* accesses value getter after done');
```
bakkot, line 47: This doesn't really belong in this test; it's an assertion about something unrelated.

michaelficarra: Fine. I thought it was cute and expository.

ptomato (approved):
Makes sense and seems to match the linked change.

bakkot (approved):

michaelficarra:
lmao @ptomato we decided at plenary yesterday to back out these changes, so we'll need to revert this now

ptomato:
:facepalm: You're right. I forgot.

PR #4486, opened 2025-05-13
Fixup incorrect assertion in AsyncGenerator.p.return with broken promises

legendecas:
Refs: https://github.com/tc39/test262/pull/3472/files#r2085898854

### DIFF
--- a/test/built-ins/AsyncGeneratorPrototype/return/return-state-completed-broken-promise.js
+++ b/test/built-ins/AsyncGeneratorPrototype/return/return-state-completed-broken-promise.js
@@ -46,17 +46,21 @@ Object.defineProperty(brokenPromise, 'constructor', {
   }
 });
 
-it.next();
-it.return(brokenPromise)
-  .then(
-    () => {
-      throw new Test262Error("Expected rejection");
-    },
-    err => {
-      assert(unblocked, false, 'return should be rejected before generator is resumed');
-      assert.sameValue(err.message, 'broken promise');
-    }
-  )
-  .then($DONE, $DONE);
+it.next().then(function(result) {
+  assert.sameValue(result.value, undefined);
+  assert.sameValue(result.done, true);
+
+  it.return(brokenPromise)
+    .then(
+      () => {
+        throw new Test262Error("Expected rejection");
+      },
+      err => {
+        assert(unblocked, 'return should be rejected when the generator is completed');
+        assert.sameValue(err.message, 'broken promise');
+      }
+    )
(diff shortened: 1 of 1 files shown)


test/built-ins/AsyncGeneratorPrototype/return/return-state-completed-broken-promise.js
```diff
@@ -53,7 +53,7 @@ it.return(brokenPromise)
       throw new Test262Error("Expected rejection");
     },
     err => {
-      assert(unblocked, false, 'return should be rejected before generator is resumed');
+      assert(unblocked, 'return should be rejected when the generator is completed');
```
Jack-Works, line 56:
```suggestion
      assert(unblocked, 'return should be rejected after the generator is resume');
```

https://github.com/tc39/ecma262/issues/3596#issuecomment-2875186161

bakkot: @legendecas In this test, the state of the generator when `.return` is called is `~executing~`, not `~completed~`, because at the time `.return` is called the generator is paused at `await blocking`. So the quoted step doesn't apply.

legendecas: @bakkot thanks, yeah, it should be awaited. Updated the test to reflect the expected state should be `~completed~`.

test/built-ins/AsyncGeneratorPrototype/return/return-state-completed-broken-promise.js
```diff
@@ -46,17 +46,21 @@ Object.defineProperty(brokenPromise, 'constructor', {
...
-    () => {
-      throw new Test262Error("Expected rejection");
-    },
-    err => {
-      assert(unblocked, false, 'return should be rejected before generator is resumed');
-      assert.sameValue(err.message, 'broken promise');
-    }
-  )
-  .then($DONE, $DONE);
+iter.next().then(function(result) {
```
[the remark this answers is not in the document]
legendecas: Ah, yes, it should be `it`.

ptomato (approved):
Fails on SpiderMonkey. Is that expected?

anba:
> Fails on SpiderMonkey. Is that expected?

Yes. That's expected &rarr; https://searchfox.org/mozilla-central/rev/043d59a948b9428685118b2d6f756658456577ec/js/src/tests/jstests.list#650

PR #4500, opened 2025-05-22
Add test for importing a succesfully evaluated module of an async errored SCC

nicolo-ribaudo:
This is a test for https://github.com/tc39/ecma262/pull/3584#issuecomment-2900042350, for which we currently do not have any test262 coverage.


nicolo-ribaudo (commented):
The test is failing on CI with this error:
```
FAIL test/language/module-code/top-level-await/evaluated-dep-from-async-errored-scc.js (default)
  asyncTest called without async flag

FAIL test/language/module-code/top-level-await/evaluated-dep-from-async-errored-scc.js (strict mode)
  asyncTest called without async flag
```

What am I doing wrong?

PR #4503, opened 2025-05-26
Fix tests for Intl.PluralRules notation

anba:
See also the review notes from @trflynn89 in #4455.


test/intl402/PluralRules/prototype/select/notation.js
```diff
@@ -12,9 +12,45 @@ info: |
...
+var prengineering = new Intl.PluralRules("fr", { notation: "engineering" });
+var prscientific = new Intl.PluralRules("fr", { notation: "scientific" });
+var prcompact = new Intl.PluralRules("fr", { notation: "compact" });
+
+for (var {value, standard, engineering, scientific, compact} of values) {
+  assert.sameValue(prstandard.select(value), standard, `standard notation: ${value}`);
+  assert.sameValue(prengineering.select(value), engineering, `engineering notation: ${value}`);
+  assert.sameValue(prscientific.select(value), scientific, `scientific notation: ${value}`);
+  assert.sameValue(prcompact.select(value), compact, `compact notation: ${value}`);
+}
```
Ms2ger, line 56: I slightly prefer looping over the "notation" values as well, but am willing to land either way. Let me know

anba: `compactDisplay` will likely be added shortly, cf. <https://github.com/tc39/ecma402/pull/989#issuecomment-2906752480>. Then we can no longer just iterate over the four possible `notation` values, but also need to take `compactDisplay` into account. Therefore I think I prefer the current PR.

PR #4504, opened 2025-05-26
Add coverage for resolved locale when Unicode extensions and options are present

anba:
These tests currently fail in V8, but pass in JSC and SM.

### DIFF
--- a/test/intl402/DateTimeFormat/prototype/resolvedOptions/resolved-numbering-system-unicode-extensions-and-options.js
+++ b/test/intl402/DateTimeFormat/prototype/resolvedOptions/resolved-numbering-system-unicode-extensions-and-options.js
@@ -1,0 +1,74 @@
+// Copyright 2025 André Bargull. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-intl.datetimeformat.prototype.resolvedoptions
+description: >
+  Resolved numbering system when using Unicode extension values and options.
+locale: [en]
+---*/
+
+var tests = [
+  // Unicode extension value is present and supported. Options value present,
+  // but unsupported. Unicode extension value is used and reflected in the
+  // resolved locale.
+  {
+    locale: "en-u-nu-arab",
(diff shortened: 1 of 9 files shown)


test/intl402/DateTimeFormat/prototype/resolvedOptions/resolved-numbering-system-unicode-extensions-and-options.js
```diff
@@ -0,0 +1,74 @@
...
+locale: [en]
+---*/
+
+var tests = [
+  // Unicode extension value is present and supported. Options value present,
+  // but unsupported. Unicode extension value is used and reflected in the
+  // resolved locale.
+  {
+    locale: "en-u-nu-arab",
+    numberingSystem: "invalid",
```
FrankYFTang, line 17: which line in the spec prevent "invalid" or "invalid2" to be used for numberingSystem?

anba: The relevant spec operation is [ResolveLocale](https://tc39.es/ecma402/#sec-resolvelocale), step 13.k:
> k. If SameValue(optionsValue, value) is false and keyLocaleData contains optionsValue, then

In the test case, `optionsValue` is `"invalid"` and `value` is `"arab"`. `SameValue("invalid", "arab")` returns `false`, but because `keyLocaleData` doesn't contain a numbering system named `"invalid"`, `optionsValue` is ignored and the numbering system defined in the Unicode extension sequence `u-nu-arab` is used.

This assumes that no valid numbering system is named `"invalid"`. If it helps, we can adjust the test case to also include:
```js
assert.sameValue(
  Intl.supportedValuesOf("numberingSystem").includes("invalid"),
  false,
  "'invalid' is not a valid numbering system"
);
```

FrankYFTang: I see. Thanks

PR #4511, opened 2025-05-30
Change `deepEqual.js` as Rhino doesn't support Array spreading.

aardvark179:
Small change to allow `deepEqual.js` to parse and work on Rhino again. Weirdly all the template syntax is supported just fine, but array spreading is the thing that isn't actually supported.

Checked with engine 262 that no tests have regressed with this change.


ptomato:
Seems fine. Although we are trying to phase out `deepEqual.js`, it's still desirable to ensure that harness files work with all test262 consumers.

Please ping when it's no longer draft, and ready for review.

aardvark179:
@ptomato ready for review. I only put it in draft because I realised it wasn't up to date with main.

PR #4513, opened 2025-05-30
Test that Intl.Locale language "root" is rejected

gibson042:
cf. https://unicode.org/reports/tr35/#unicode_bcp47_locale_id


Ms2ger:
I'm missing the connection between your link and the spec; could you update the `info` block (which is outdated anyway)? Maybe that'd help me grasp the issue.

gibson042:
@Ms2ger Clarified in context. I do plan to update the info block, but not yet because I want to cover this entire neighborhood of test files.

ptomato (approved):
Discussed in maintainers meeting, makes sense

PR #4515, opened 2025-06-04
Add missing tests for 'length' value of Intl.Locale.prototype methods

anba:
I've noticed this when reviewing the tests for `Intl.Locale.prototype.variants`.

### DIFF
--- a/test/intl402/DisplayNames/prototype/of/length.js
+++ b/test/intl402/DisplayNames/prototype/of/length.js
@@ -1,0 +1,29 @@
+// Copyright (C) 2025 André Bargull. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-Intl.DisplayNames.prototype.of
+description: >
+  Intl.DisplayNames.prototype.of.length is 1.
+info: |
+  ECMAScript Standard Built-in Objects
+
+  Every built-in function object, including constructors, has a "length" property
+  whose value is a non-negative integral Number. Unless otherwise specified,
+  this value is the number of required parameters shown in the subclause heading
+  for the function description. Optional parameters and rest parameters are not
+  included in the parameter count.
+
+  Unless otherwise specified, the "length" property of a built-in function object
+  has the attributes { [[Writable]]: false, [[Enumerable]]: false,
+  [[Configurable]]: true }.
+includes: [propertyHelper.js]
+features: [Intl.DisplayNames]
+---*/
+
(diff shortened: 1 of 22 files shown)


test/intl402/DisplayNames/prototype/of/length.js
```diff
@@ -0,0 +1,29 @@
...
+  included in the parameter count.
+
+  Unless otherwise specified, the "length" property of a built-in function object
+  has the attributes { [[Writable]]: false, [[Enumerable]]: false,
+  [[Configurable]]: true }.
+includes: [propertyHelper.js]
+features: [Intl.DisplayNames]
+---*/
+
+verifyProperty(Intl.DisplayNames.prototype.of, "length", {
```
Ms2ger, line 24: Can you instead update the other tests to use verifyPrimordialCallableProperty?

anba: `verifyPrimordialCallableProperty` doesn't support accessor methods, therefore I've instead used plain `verifyProperty` throughout this PR.

gibson042: > `verifyPrimordialCallableProperty` doesn't support accessor methods, therefore I've instead used plain `verifyProperty` throughout this PR.

Yes it does, and I'm specifically planning for uses like **test/intl402/Locale/prototype/baseName/descriptor.js**:
```js
var desc = Object.getOwnPropertyDescriptor(Intl.Locale.prototype, "baseName");
var isHardened = Object.isFrozen(Object);
if (isHardened && desc.get) {
  Object.defineProperty(desc, "get", { configurable: false });
}

assert.sameValue(desc.value, undefined);
assert.sameValue(desc.set, undefined);
verifyCallableProperty(desc, "get", "get baseName", 0, { configurable: !isHardened });

verifyPrimordialProperty(Intl.Locale.prototype, "baseName", {
  enumerable: false,
  configurable: true,
});
```

I'm also open to making even better accommodations, e.g.

```js
var desc = Object.getOwnPropertyDescriptor(Intl.Locale.prototype, "baseName");

assert.sameValue(desc.value, undefined);
assert.sameValue(desc.set, undefined);
verifyPrimordialCallable(desc.get, "get baseName", 0);

verifyPrimordialProperty(Intl.Locale.prototype, "baseName", {
  enumerable: false,
  configurable: true,
});
```

gibson042 (changes requested):
See https://github.com/tc39/test262/pull/4515#discussion_r2127050283

PR #4516, opened 2025-06-05
Update number of tests in README to 50k+

foolip:
Since 2017 the test suite has grown significantly and 50k is a significant number!

The command `git grep -lF '/*---' test | wc -l` found 50568 tests.

A count in May (commit 5c9ff876dcefd19a27ef735224bd69ac1d1ee8e9) is used so that the past tense still works.


foolip:
I'm not certain that `git grep -lF '/*---' test | wc -l` is accurate, but I went to look for when the count surpassed 50k using that metric. It was all the way back in c8cd136888a804772ef774c3b2fe73299e3c7ef3 in early 2024. I'm late to the celebration!

PR #4521, opened 2025-06-18
Temporal: Add coverage for BubbleRelativeDuration called with time unit

ptomato:
See: tc39/proposal-temporal#3121

A type assertion in the spec was incorrect, but the assertion wasn't hit in the Firefox implementation in any test262 test. These three tests add coverage for all the paths that would have hit that assertion.


ptomato:
cc @anba

justingrant (approved):
Looks good!

Ms2ger (approved):
Thanks!

PR #4523, opened 2025-06-20
update Iterator Sequencing tests to reflect May 2025 plenary updates

michaelficarra:
These tests should match the spec changes in https://github.com/tc39/proposal-iterator-sequencing/pull/26.

### DIFF
--- a/test/built-ins/Iterator/concat/fresh-iterator-result.js
+++ b/test/built-ins/Iterator/concat/fresh-iterator-result.js
@@ -46,4 +46,4 @@ let iterResult = iterator.next();
 assert.sameValue(iterResult.done, false);
 assert.sameValue(iterResult.value, 123);
 
-assert.sameValue(iterResult, oldIterResult);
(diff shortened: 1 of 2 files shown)


Ms2ger (approved):

iamstolis:
Shouldn't [built-ins/Iterator/concat/next-method-returns-throwing-value.js](https://github.com/tc39/test262/blob/main/test/built-ins/Iterator/concat/next-method-returns-throwing-value.js) be updated as well? [iterator.next()](https://github.com/tc39/test262/blob/main/test/built-ins/Iterator/concat/next-method-returns-throwing-value.js#L49) did not throw before because `iteratorResult` was reused (i.e. its value was not read). Now, `IteratorStepValue()` always reads the value of the inner iterator (which throws in this case). So, `iterator.next()` should throw now, right?

In other words, I think that https://github.com/tc39/test262/commit/71794b138844536a36b341700f88ee91c5754431 should be reverted now.

Issue #4112, opened 2024-06-19
 Create testing plan for the Joint Iteration proposal.


ioannad:
This is to plan and track the tests for the Joint Iteration proposal.

Proposal repo: https://github.com/tc39/proposal-joint-iteration

ioannad:
I made a PR with a testing plan guide, still needs review but may contain useful information: https://github.com/tc39/test262/pull/4113/files

bakkot:
A partial list of things that need testing, written with @michaelficarra . Stopped when we ran out of time, not when we ran out of things that need testing.

## `zip`

- empty iterable as argument produces iterator which is finished
- cases for the iterable returned by `padding`:
  - it returns `{ done: true }` in fewer steps than there are things to be zipped
    - in this case `next` should not be called any more times
    - and `.return` should not be looked at or called
  - it does not return `{ done: true }` within the number of steps as there are things to be zipped
    - in this case its `.return` should be called
- result objects are arrays

## `zipKeyed`

- object with no own-enumerable keys as argument produces iterator which is finished
- confirm inherited keys are not used
- confirm non-enumerable keys are not used
- make sure that output objects have correct attributes for the properties
  - output properties are writable/configurable, even if they were not on the input
- output objects inherit from `null`
- order of properties reported by `Object.keys()` is the same as they were on the first argument
- tests with Symbol-named properties

## both

- throws if the first argument is not an object (including if it is a string)
- throws if the second argument is not an object and not null/undefined
- `mode` is read from second argument
- `mode` is not coerced to a string
- `padding` is read if and only if `mode` is `"longest"`
- when `mode` is `"longest"`, throws if `padding` is not `undefined` and not an object (including if it is a string)
- `mode` is not read if the first argument is not an object
- throws if any iterables are strings (after reading the `padding`)
  - currently the proposal allows non-string primitives (e.g. with `Symbol.iterator` on `Number.prototype`) but see https://github.com/tc39/proposal-joint-iteration/issues/34)
- order of side effects in general (there are lots of these, just make something that makes every single side-effect observable and assert on the totality)
  - should assert immediately after calling `.zip`, and also after calling `.next` on the result
- can zip exactly one thing
- can zip exactly two things with the same number of elements
- can zip two iterable things of different length with no options object (which behaves like `shortest`)
  - need tests for first thing is longer, and second thing is longer
- as above but with each of the three modes explicitly
[...]

PR #4528, opened 2025-07-03
Add tests for Joint Iteration proposal

anba:
See <https://github.com/tc39/test262/issues/4112#issuecomment-2415235086> for the testing plan.

### DIFF
--- a/test/built-ins/Iterator/zip/basic.js
+++ b/test/built-ins/Iterator/zip/basic.js
@@ -1,0 +1,236 @@
+// Copyright (C) 2025 André Bargull. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-iterator.zip
+description: >
+  Basic Iterator.zip test using all three possible modes.
+includes: [compareArray.js, propertyHelper.js]
+features: [joint-iteration]
+---*/
+
+// Assert |result| is an object created by CreateIteratorResultObject.
+function assertIteratorResult(result, value, done) {
+  assert.sameValue(
+    Object.getPrototypeOf(result),
+    Object.prototype,
+    "[[Prototype]] of iterator result is Object.prototype"
+  );
+
+  assert(Object.isExtensible(result), "iterator result is extensible");
+
+  var ownKeys = Reflect.ownKeys(result);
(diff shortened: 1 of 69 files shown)


michaelficarra:
Tests have been validated with the polyfill in https://github.com/tc39/proposal-joint-iteration/pull/44. I'll do a review for completeness soon and then ask for Stage 3. Thanks for implementing these, @anba.

test/built-ins/Iterator/zip/basic.js
ptomato, line 1: I would prefer we don't compute a large list of expected results inline in a test using what is essentially a mini-polyfill. When the test fails, it's unclear whether the implementation or the mini-polyfill is wrong.

There was some more discussion on this re. locale-sensitive testing in the talk I presented in the April TC39 meeting. [slides](https://ptomato.name/talks/tc39-2025-04/#14), [discussion](https://github.com/tc39/notes/blob/6e3e0aaad700a212cb92da6be5a5b409c8807805/meetings/2025-04/april-16.md?plain=1#L764)

Same for `zipKeyed/basic.js`.

ptomato (commented):
I didn't do a real review, just skimmed most of it. As a reviewer, it'd be very helpful if you could cross-reference these with the testing plan. I hate asking for extra work, but anything that makes it easier for us to do the review in small chunks without having to swap in extra context is appreciated.

One test that occurred to me that I didn't see in this PR or the testing plan (although as I said, I only skimmed it so I may have missed it) is verifying the actual behaviour when `mode` is missing. In other words I see we test thoroughly that absent or undefined `mode` is handled, but not what it actually does.

test/built-ins/Iterator/zip/basic.js
bakkot: I don't think this is really a mini-polyfill, just verifying several properties at once. I've split it into two tests + a new harness file, if you think that's an improvement: https://github.com/anba/test262/pull/1

bakkot:
I opened a PR including both the commits from this branch and my changes from https://github.com/anba/test262/pull/1, in case that's simpler: https://github.com/tc39/test262/pull/4609

bakkot:
Landed in https://github.com/tc39/test262/pull/4609.

PR #4531, opened 2025-07-07
Fix and extend well-known intrinsics harness file

anba:
The `Intl` intrinsics can probably be shortened to just include `%Intl%`, `%IntlSegmentIteratorPrototype%`, and `%IntlSegmentsPrototype%`, because the other constructors are reachable through `%Intl%`. But because <https://tc39.es/ecma402/#sec-402-well-known-intrinsic-objects> includes all `Intl` constructors, I've used the same approach here.


Ms2ger (approved):
Thanks!

PR #4533, opened 2025-07-07
add tests for invalid Uint8Array.fromBase64 chunks

ChALkeR:
https://tc39.es/proposal-arraybuffer-base64/spec/#sec-frombase64
* 10.3, step 10.b.i
* 10.3, step 10.e

Passes on SpiderMonkey, fails on JSC
Current tests don't catch it


ljharb (approved):

bakkot (approved):
Thanks, sorry for the lack of coverage here.

PR #4534, opened 2025-07-08
Add missing include to Uint8Array/fromBase64/last-chunk-invalid.js

trflynn89:
We were seeing an error about `assert.compareArray` being undefined over at Ladybird.


Ms2ger (approved):
Thanks

PR #4535, opened 2025-07-10
Add script to generate standard property tests

anba:
0166442283bbcd242d4cd23ee3a51d3a42a84484
- Add `verifyBuiltinFunction` and `verifyBuiltinConstructor`. Both functions are used in the generated tests.
- The new harness functions use `Reflect` and `Proxy` to identify constructors without invoking any extra MOP methods. The existing `isConstructor` harness function performs an extra [[Get]] access from [\[\[Construct\]\]](https://tc39.es/ecma262/#sec-ecmascript-function-objects-construct-argumentslist-newtarget) &rarr; `OrdinaryCreateFromConstructor` &rarr; `GetPrototypeFromConstructor` &rarr; `Get`. The new approach avoids this extra [[Get]]. 

9ad1b2b6a04bf73db50d061c38b7342331e14571
- Change the existing generator to not delete generated tests from other generators.
- This change is necessary because otherwise CI complains when regenerating tests from "make.py".

c5351e479498d8b88832d2a4b69b290d103e6067
- Python scripts to generate the standard property tests.
- The tests are generated from the actual spec files. The spec "parsing" is mostly regular expression matching against known spec phrases and some additional heuristics to properly identify partial spec files from proposals. 
- The property test generator was tested against ECMA-262, ECMA-402, and all stage 3 and stage 2.7 proposals.

cfe8c2ad15ac00ee4bbffa135eb12050c1f692ee
- Example commit showing generated property tests for the `Number` built-in. 

b012d8ed3e99cd70733348682c087c6ad57270b9
- Example commit showing generated property tests for ECMA-402 overrides of the `Number` built-in. 

933d746746b14191deb717749b1ac16c49062b40
3b224c4aa8fd5c40e2e668fad9df88b9b81f3e8e
- Remove obsolete tests which are now covered by the generated tests.

8b5cdbc6a00e143d59715de15db599060df486dd
- Example commit showing generated property tests for a spec proposal.


harness/builtin.js
```diff
@@ -0,0 +1,158 @@
...
+  // Unless otherwise specified, the "length" property of a built-in function
+  // object has the attributes { [[Writable]]: false, [[Enumerable]]: false,
+  // [[Configurable]]: true }.
+  verifyPrimordialProperty(fun, "length", {
+    value: length,
+    writable: false,
+    enumerable: false,
+    configurable: true,
+  });
+
```
gibson042, line 69: Assertion ordering suggestion:
```suggestion
  // Unless otherwise specified, the "name" property of a built-in function
  // object has the attributes { [[Writable]]: false, [[Enumerable]]: false,
  // [[Configurable]]: true }.
  verifyPrimordialProperty(fun, "name", {
    value: name,
    writable: false,
    enumerable: false,
    configurable: true,
  });

  // Unless otherwise specified, the "length" property of a built-in function
  // object has the attributes { [[Writable]]: false, [[Enumerable]]: false,
  // [[Configurable]]: true }.
  verifyPrimordialProperty(fun, "length", {
    value: length,
    writable: false,
    enumerable: false,
    configurable: true,
  });

  // Built-in function objects that are not constructors do not have a
  // "prototype" property unless otherwise specified in the description of a
  // particular function.
  //
  // NOTE: |verifyBuiltinFunction| is more strict and allows only exactly two
  // own properties: "name" and "length".
  assert.sameValue(__Reflect_ownKeys(fun).length, 2);
```

harness/builtin.js
```diff
@@ -0,0 +1,158 @@
...
+    // ```
+    // class C extends null { constructor() { return {}; } };
+    // assert.throws(TypeError, function() {
+    //   __Reflect_construct(C, [], fun);
+    // });
+    // ```
+    __Reflect_construct(__Proxy, [{}, {}], fun);
+  });
+  assert.throws(TypeError, function() {
+    // Test with `new` expression in addition to Reflect.construct.
```
gibson042, line 96:
```suggestion
  // Test with `new` expression in addition to Reflect.construct.
  assert.throws(TypeError, function() {
```

harness/builtin.js
```diff
@@ -0,0 +1,158 @@
...
+/**
+ * Verify `fun` is a normal built-in constructor function object with the
+ * default properties as defined in
+ * <https://tc39.es/ecma262/#sec-ecmascript-standard-built-in-objects>.
+ */
+function verifyBuiltinConstructor(fun, name, length, prototype) {
+  // Unless specified otherwise, a built-in object that is callable as a
+  // function is a built-in function object with the characteristics described
+  // in 10.3.
+  assert.sameValue(typeof fun, "function");
```
gibson042, line 110: I wonder if the common assertions between `verifyBuiltinFunction` and `verifyBuiltinConstructor` should be extracted.

<details>

```js
/**
 * Verify `fun` is a normal built-in function object or constructor
 * function object with the default properties as defined in
 * <https://tc39.es/ecma262/#sec-ecmascript-standard-built-in-objects>.
 */
function verifyBuiltinCallable(fun, name, length, prototype) {
  // Unless specified otherwise, a built-in object that is callable as a
  // function is a built-in function object with the characteristics described
  // in 10.3.
  assert.sameValue(typeof fun, "function");

  // Unless otherwise specified every built-in function and every built-in
  // constructor has the Function prototype object, which is the initial value
  // of the expression Function.prototype (20.2.3), as the value of its
  // [[Prototype]] internal slot.
  if (prototype === undefined) {
    prototype = __Function.prototype;
  }
  assert.sameValue(__Reflect_getPrototypeOf(fun), prototype);

  // Unless specified otherwise, the [[Extensible]] internal slot of a built-in
  // object initially has the value true.
  assert.sameValue(__Reflect_isExtensible(fun), true);

  // Unless otherwise specified, the "name" property of a built-in function
  // object has the attributes { [[Writable]]: false, [[Enumerable]]: false,
  // [[Configurable]]: true }.
  verifyPrimordialProperty(fun, "name", {
    value: name,
    writable: false,
    enumerable: false,
    configurable: true,
  });

  // Unless otherwise specified, the "length" property of a built-in function
  // object has the attributes { [[Writable]]: false, [[Enumerable]]: false,
  // [[Configurable]]: true }.
  verifyPrimordialProperty(fun, "length", {
    value: length,
    writable: false,
    enumerable: false,
    configurable: true,
  });
}

/**
 * Verify `fun` is a normal built-in function object with the default
 * properties as defined in
 * <https://tc39.es/ecma262/#sec-ecmascript-standard-built-in-objects>.
 */
function verifyBuiltinFunction(fun, name, length) {
  verifyBuiltinCallable(fun, name, length);

  // Built-in function objects that are not constructors do not have a
  // "prototype" property unless otherwise specified in the description of a
  // particular function.
  // |verifyBuiltinFunction| is more strict, allowing only  "name" and "length".
  assert.sameValue(__Reflect_ownKeys(fun).length, 2);

  // Built-in function objects that are not identified as constructors do not
  // implement the [[Construct]] internal method unless otherwise specified in
  // the description of a particular function.
    // Reflect.construct throws a TypeError if `fun` is not a constructor. Use
    // the Proxy constructor because it ignores `NewTarget`.
    //
    // Two alternatives which also ensure no additional operations are called:
    //
    // 1. Create a Proxy for `fun` with a "construct" trap. Then call `new` on
    // the newly created Proxy object.
    // ```
    // var p = new Proxy(fun, {construct(){ return {}; }});
    // assert.throws(TypeError, function() { new p; });
    // ```
    //
    // 2. Use a derived class object.
    // ```
    // class C extends null { constructor() { return {}; } };
    // assert.throws(TypeError, function() {
    //   __Reflect_construct(C, [], fun);
    // });
    // ```
    __Reflect_construct(__Proxy, [{}, {}], fun);
  });

  // Test with `new` expression in addition to Reflect.construct.
    new fun();
  });
}

/**
 * Verify `fun` is a normal built-in constructor function object with the
 * default properties as defined in
 * <https://tc39.es/ecma262/#sec-ecmascript-standard-built-in-objects>.
 */
function verifyBuiltinConstructor(fun, name, length, prototype) {
  verifyBuiltinCallable(fun, name, length, prototype);

  // Built-in function objects that are not identified as constructors do not
  // implement the [[Construct]] internal method unless otherwise specified in
  // the description of a particular function.

  // Reflect.construct throws a TypeError if `fun` is not a constructor.
  //
  // See verifyBuiltinFunction for why Proxy is used here.
    __Reflect_construct(__Proxy, [{}, {}], fun);

    // Throw the expected error.
    throw new Test262Error();
  });
}
```

</details>

harness/propertyHelper.js
```diff
@@ -101,7 +101,7 @@ function verifyProperty(obj, name, desc, options) {
     }
   }
 
-  if (__hasOwnProperty(desc, 'enumerable') && desc.enumerable !== undefined) {
+  if (__hasOwnProperty(desc, 'enumerable')) {
```
gibson042, line 104: :+1:

harness/builtin.js
gibson042: We could also follow the `__push(failures, msg);` approach of harness/propertyHelper.js.

tools/property-test-generator/README.md
```diff
@@ -0,0 +1,43 @@
+# Property Test Generator
+
+This tool generates standard property tests for built-in functions and objects.
+
+Supported options:
+
+- `--dry-run`: Don't write any output files.
+- `--include`: Comma separated list of globs to filter which built-ins are included.
+- `--exclude`: Comma separated list of globs to filter which built-ins are excluded.
+- `--features`: Comma separated list of features from "features.txt". If not specified features are guessed based on files in the same directory.
```
gibson042, line 10:
```suggestion
- `--features`: Comma separated list of features from "features.txt". If not specified, features are guessed based on files in the same directory.
```

tools/property-test-generator/README.md
```diff
@@ -0,0 +1,43 @@
...
+Examples:
+
+- Generate property tests for all built-ins defined in "$ECMA262/spec.html":
+```sh
+$ python tools/property-test-generator/main.py $ECMA262/spec.html
+```
+
+- Generate property tests for the `Number` constructor, but not the `Number` prototype object:
+```sh
+$ python tools/property-test-generator/main.py --include=Number* --exclude=Number.prototype* $ECMA262/spec.html
```
gibson042, line 28: For better shell and environment portability:
```suggestion
$ python tools/property-test-generator/main.py --include='Number*' --exclude='Number.prototype*' $ECMA262/spec.html
```

tools/property-test-generator/README.md
```diff
@@ -0,0 +1,43 @@
...
+```
+
+- Generate property tests for all built-ins defined in "$ECMA402/spec":
+```sh
+python tools/property-test-generator/main.py --global=$ECMA262/spec.html $ECMA402/spec/
+```
+
+The `--global` argument is necessary to correctly pick up global definitions from ECMA-262.
```
gibson042, line 36: Internal consistency:
`````suggestion
- Generate property tests for all built-ins defined in "$ECMA402/spec":
```sh
python tools/property-test-generator/main.py --globals=$ECMA262/spec.html $ECMA402/spec/
```

The `--globals` argument is necessary to correctly pick up global definitions from ECMA-262.
`````

**Review by gibson042 — COMMENTED** [review]
I had held off on helpers like this because of dependency upon Reflect and especially Proxy, but if we're ready to raise that bar then I think this is great—although I do have concerns about the maintenance burden of the assumptions and parsing aspects of property-test-generator, which seem underdocumented. Maybe some of it should be pulled out of test262 for broader use and distinct maintenance?

I also **love** seeing removal of superseded files like [test/built-ins/Number/isFinite/length.js](https://github.com/tc39/test262/blob/main/test/built-ins/Number/isFinite/length.js), [test/built-ins/Number/isFinite/name.js](https://github.com/tc39/test262/blob/main/test/built-ins/Number/isFinite/name.js), [test/built-ins/Number/isFinite/not-a-constructor.js](https://github.com/tc39/test262/blob/main/test/built-ins/Number/isFinite/not-a-constructor.js), etc.

File: harness/builtin.js (hunk 1 above)
Comment (anba) [reply harness/builtin.js:69]: `Reflect.ownKeys` is intentionally called first, because `verifyPrimordialProperty` deletes the properties. And adding `{restore: true}` just to move `Reflect.ownKeys` after `verifyPrimordialProperty` didn't seem useful to me.

Comment (gibson042) [reply harness/builtin.js:69]: Hmm, noted.

**Comment by anba** [comment]
> I had held off on helpers like this because of dependency upon Reflect and especially Proxy, [...]

I don't actually know if there are any test262 consumers which pull the latest test262 commits, but don't support `Reflect` or `Proxy`. But I guess we'll find out if there are any complaints. :smile:  

> [...] although I do have concerns about the maintenance burden of the assumptions and parsing aspects of property-test-generator, which seem underdocumented. Maybe some of it should be pulled out of test262 for broader use and distinct maintenance?

Hmm, I understand your proposal, but I'm not sure it's useful right now to invest time making the parser more reusable for others, if there aren't yet any other projects which could benefit from it. Should this ever happen, I'm open to move the parser into a separate project, so it's easier to coordinate and plan how to extend it to match the project needs of others. 

I don't think we need to spend much time on maintenance, because the specification seems to be in a fairly stable shape w.r.t. how properties and functions are defined. When trying the parser on past ECMA-262 editions, I didn't observe any issues which can't be easily fixed through small code updates:

| Edition | Parse Success? | Notes |
| :---- | :----: | :--- | 
| 2025  | :x:  | Fails for `Iterator.prototype.constructor`: <https://github.com/tc39/ecma262/pull/3558> |
| 2024 | :ok:  | Prints a warning for unsupported old `@@symbol` notation. |
| 2021-2023 | :ok:  | Prints a warning for unsupported value definitions. |
| 2016-2020 | :x:  | Fails for old host-defined global prototype definition: https://github.com/tc39/ecma262/pull/1951 |

File: harness/builtin.js (hunk 1 above)
Comment (gibson042) [reply harness/builtin.js:69]: So alternatively:
```patch
   // NOTE: |verifyBuiltinFunction| is more strict and allows only exactly two
   // own properties: "name" and "length".
-  assert.sameValue(__Reflect_ownKeys(fun).length, 2);
+  assert.compareArray(__Reflect_ownKeys(fun).sort(), ["length", "name"],
+    "own property keys");
 
   // Unless otherwise specified, the "name" property of a built-in function
```

<details><summary>GitHub PR suggestion</summary>

```suggestion
  // Built-in function objects that are not constructors do not have a
  // "prototype" property unless otherwise specified in the description of a
  // particular function.
  //
  // NOTE: |verifyBuiltinFunction| is more strict and allows only exactly two
  // own properties: "name" and "length".
  assert.compareArray(__Reflect_ownKeys(fun).sort(), ["length", "name"],
    "own property keys");

  // Unless otherwise specified, the "name" property of a built-in function
  // object has the attributes { [[Writable]]: false, [[Enumerable]]: false,
  // [[Configurable]]: true }.
  verifyPrimordialProperty(fun, "name", {
    value: name,
    writable: false,
    enumerable: false,
    configurable: true,
  });

  // Unless otherwise specified, the "length" property of a built-in function
  // object has the attributes { [[Writable]]: false, [[Enumerable]]: false,
  // [[Configurable]]: true }.
  verifyPrimordialProperty(fun, "length", {
    value: length,
    writable: false,
    enumerable: false,
    configurable: true,
  });
```

</details>

anba: Calling `assert.compareArray` requires including "harness/compareArray.js" in each test file including "harness/builtin.js". This kind of dependency is a bit annoying, so I think I'd prefer this alternative:
```js
var ownKeys = __Reflect_ownKeys(fun);
assert.sameValue(ownKeys.length, 2);
assert(
  (ownKeys[0] === "length" && ownKeys[1] === "name") ||
  (ownKeys[0] === "name" && ownKeys[1] === "length")
);
```

Strictly speaking `(ownKeys[0] === "name" && ownKeys[1] === "length")` isn't allowed, because all built-in functions are created through [CreateBuiltinFunction](https://tc39.es/ecma262/#sec-createbuiltinfunction), which first defines `"length"` and then `"name"`. But not all implementations get this right:
- <https://bugs.webkit.org/show_bug.cgi?id=295485> (recently fixed)
- <https://bugzilla.mozilla.org/show_bug.cgi?id=1629803> (still open)

gibson042: That makes sense to me, but I'm specifically trying to ensure that assertion-failure output is useful. So:
```js
  // NOTE: |verifyBuiltinFunction| is more strict and allows only exactly two
  // own properties: "name" and "length".
  var ownKeys = __Reflect_ownKeys(fun);
  if (
    ownKeys.length !== 2 ||
    (ownKeys[0] !== "name" && ownKeys[1] !== "name") ||
    (ownKeys[0] !== "length" && ownKeys[1] !== "length")
  ) {
    var actualKeysStr = "[" + ownKeys.map(String).join(", ") + "]";
    var msg = "Actual own keys " + actualKeysStr + " should have been [length, name]";
    assert(false, msg);
  }

  // Unless otherwise specified, the "name" property of a built-in function
```

ptomato:
I didn't review it as closely as Richard. The first few commits seem fine to go ahead with, maybe in a separate PR so they're not blocked while resolving the rest of the comments.

I have to say that I'm somewhat skeptical of running regular expressions on the ECMA-262 prose and generating tests from that directly. Personally, I'd prefer a config file with all the entries for which to generate property tests. The config file could still be generated/updated with a fancy script that reads ECMA-262 if you like, but at least that way there'd be a human in the loop and a clear overview of what changed whenever we update it.

harness/builtin.js
```diff
@@ -0,0 +1,158 @@
...
+---*/
+
+// Capture primordial functions that are used in verification but might be
+// destroyed *by* that process itself.
+var __Function = Function;
+var __Proxy = Proxy;
+var __Reflect_construct = Reflect.construct;
+var __Reflect_getPrototypeOf = Reflect.getPrototypeOf;
+var __Reflect_isExtensible = Reflect.isExtensible;
+var __Reflect_ownKeys = Reflect.ownKeys;
```
Ms2ger, line 19: Should we use an IIFE rather than this naming convention?

harness/builtin.js
```diff
@@ -0,0 +1,158 @@
...
+  // Unless specified otherwise, a built-in object that is callable as a
+  // function is a built-in function object with the characteristics described
+  // in 10.3.
+  assert.sameValue(typeof fun, "function");
+
+  // Unless otherwise specified every built-in function and every built-in
+  // constructor has the Function prototype object, which is the initial value
+  // of the expression Function.prototype (20.2.3), as the value of its
+  // [[Prototype]] internal slot.
+  assert.sameValue(__Reflect_getPrototypeOf(fun), __Function.prototype);
```
Ms2ger, line 36: Should this capture `Function.prototype` rather than pulling it off `Function` here?

harness/builtin.js
Ms2ger: I agree with the goal, but the resulting code is pretty dense. Are we worried about the impact of doing the formatting if the test passes? Otherwise it could be something like
```js
  var actualKeysStr = "[" + ownKeys.map(String).join(", ") + "]";
  var msg = "Actual own keys " + actualKeysStr + " should have been [length, name]";
  var ownKeys = __Reflect_ownKeys(fun);
  assert.sameValue(ownKeys.length, 2, msg);
  assert(ownKeys[0] === "name" || ownKeys[1] === "name", msg);
  assert(ownKeys[0] === "length" || ownKeys[1] === "length", msg);
```
but note that the formatting code is not robust against changes in the environment either (`.map()` in particular)

tools/property-test-generator/clause.py
```diff
@@ -0,0 +1,722 @@
...
+
+    def is_operation(self):
+        """Return true if this clause is an abstract operation, sdo, etc."""
+        # "abstract operation" in type or
+        # "sdo" in type or
+        # "concrete method" in type or
+        # "implementation-defined" in type or
+        # "host-defined" in type or
+        # "internal method" in type or
+        # "numeric method" in type
```
Ms2ger, line 53: Please remove (or explain why it's here)

tools/property-test-generator/main.py
```diff
@@ -0,0 +1,330 @@
...
+    with open(file_path, encoding="utf-8") as f:
+        text = f.read()
+
+    # Read the spec html file.
+    parser = SpecParser()
+    parser.feed(text)
+    parser.close()
+
+    # Assert all tags are properly closed.
+    assert parser.current_element == parser.root_element
```
Ms2ger, line 47: Please provide a helper function for this near the implementation of `SpecParser`

tools/property-test-generator/main.py
```diff
@@ -0,0 +1,330 @@
...
+
+        actual_path = os.path.commonpath((test262_test, os.path.realpath(out_path)))
+        if actual_path != test262_test:
+            logger.error(
+                f"Can't write outside the test262/test directory to '{cfg.out}'!"
+            )
+            return
+
+        if os.path.exists(out_path) and not os.path.isdir(out_path):
+            logger.error(f"Can't overwrite existent non-directoy file '{cfg.out}'!")
```
Ms2ger, line 223:
```suggestion
            logger.error(f"Can't overwrite existing non-directory file '{cfg.out}'!")
```

tools/property-test-generator/parser.py
```diff
@@ -0,0 +1,50 @@
...
+
+class SpecParser(HTMLParser):
+    def __init__(self):
+        super().__init__()
+        self.root_element = Element(None, None, [], None)
+        self.current_element = self.root_element
+
+    @staticmethod
+    def ignorable_tag(tag):
+        """Tuple of tags which are ignored."""
```
Ms2ger, line 20: "Returns whether the tag should be ignored"? The tuple is an implementation detail.

test/built-ins/Number/MAX_SAFE_INTEGER/prop-desc.js
```diff
@@ -0,0 +1,26 @@
+// Test generated by: property-test-generator
+
+/*---
+esid: sec-number.max_safe_integer
+description: Property test for Number.MAX_SAFE_INTEGER
+info: |
+  Number.MAX_SAFE_INTEGER
+  - The value of `Number.MAX_SAFE_INTEGER` is 9007199254740991𝔽 (𝔽(253 - 1)).
```
Ms2ger, line 8: Hm, that's quite a confusing way to render `2**53`.

harness/builtin.js
anba: No, the `prototype` property of `Function` is non-writable and non-configurable.

test/built-ins/Number/MAX_SAFE_INTEGER/prop-desc.js
anba: Hmm, fixing this will require rewriting `<sup>` tags to some other representation. 

This the source code from spec.html:
```html
<p>The value of `Number.MAX_SAFE_INTEGER` is *9007199254740991*<sub>𝔽</sub> (𝔽(2<sup>53</sup> - 1)).</p>
```

ptomato:
```suggestion
  - The value of `Number.MAX_SAFE_INTEGER` is 9007199254740991𝔽 (𝔽(2⁵³ - 1)).
```

gibson042: > Hmm, fixing this will require rewriting `<sup>` tags to some other representation.

We're already doing this as of https://github.com/tc39/ecmarkup/pull/517 . For example, the text at https://tc39.es/ecma262/multipage/numbers-and-dates.html#sec-number.max_safe_integer now copies as
```suggestion
  - The value of Number.MAX_SAFE_INTEGER is 9007199254740991𝔽 (𝔽(2**53 - 1)).
```

PR #4537, opened 2025-07-16
Add test case when countBytes in TA.p.slice is set to zero

anba:
`countBytes` can be set to zero in step 14.d when resizable array buffers are used.

### DIFF
--- a/test/built-ins/TypedArray/prototype/slice/resize-count-bytes-to-zero.js
+++ b/test/built-ins/TypedArray/prototype/slice/resize-count-bytes-to-zero.js
@@ -1,0 +1,97 @@
+// Copyright 2025 André Bargull. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-%typedarray%.prototype.slice
+description: >
+  Count bytes is set to zero when underlying buffer is resized to zero.
+info: |
+  %TypedArray%.prototype.slice ( start, end )
+
+  ...
+  2. Let taRecord be ? ValidateTypedArray(O, seq-cst).
+  3. Let srcArrayLength be TypedArrayLength(taRecord).
+  4. Let relativeStart be ? ToIntegerOrInfinity(start).
+  ...
+  14. If countBytes > 0, then
+    a. Set taRecord to MakeTypedArrayWithBufferWitnessRecord(O, seq-cst).
+    b. If IsTypedArrayOutOfBounds(taRecord) is true, throw a TypeError exception.
+    c. Set endIndex to min(endIndex, TypedArrayLength(taRecord)).
+    d. Set countBytes to max(endIndex - startIndex, 0).
+    ...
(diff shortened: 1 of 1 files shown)


anba:
This case isn't covered by existing tests.

test/built-ins/TypedArray/prototype/slice/resize-count-bytes-to-zero.js
```diff
@@ -0,0 +1,81 @@
...
+      valueOf() {
+        ta.buffer.resize(0);
+        return startIndex;
+      }
+    };
+
+    var sliced = ta.slice(start);
+
+    // Result typed array has the correct length with all elements set to zero.
+    assert.compareArray(sliced, new TA(N - startIndex));
```
gibson042:
```suggestion
    assert.compareArray(sliced, new TA(N - startIndex), TA.name + ", start index " + startIndex);
```

test/built-ins/TypedArray/prototype/slice/resize-count-bytes-to-zero.js
```diff
@@ -0,0 +1,81 @@
...
+
+// Repeat test with different target typed array constructors.
+for (var SourceTA of typedArrayConstructors) {
+  for (var TargetTA of typedArrayConstructors) {
+    var byteLength = N * SourceTA.BYTES_PER_ELEMENT;
+    var buffer = new ArrayBuffer(byteLength, {maxByteLength: byteLength});
+    var ta = new SourceTA(buffer);
+
+    // Create own "constructor" property to create a different result typed array.
+    ta.constructor = TargetTA;
```
gibson042: For not breaking in a hardened environment:
```suggestion
    Object.defineProperty(ta, "constructor", { value: TargetTA });
```

test/built-ins/TypedArray/prototype/slice/resize-count-bytes-to-zero.js
```diff
@@ -0,0 +1,81 @@
...
+        valueOf() {
+          ta.buffer.resize(0);
+          return startIndex;
+        }
+      };
+
+      var sliced = ta.slice(start);
+
+      // Result typed array has the correct length with all elements set to zero.
+      assert.compareArray(sliced, new TargetTA(N - startIndex));
```
gibson042:
```suggestion
      // Result typed array has the correct type and length with all elements set to zero.
      var label = SourceTA.name + " to " + TargetTA.name;
      assert.sameValue(sliced.constructor, TargetTA, label);
      assert.compareArray(sliced, new TargetTA(N - startIndex),
        label + ", start index " + startIndex);
```

gibson042 (changes requested):
Thanks; just a few suggestions.

Ms2ger:
@anba would you mind addressing the comments above?

Ms2ger (approved):
Thanks!

Issue #4509, opened 2025-05-29
Testing plan for Immutable ArrayBuffers


gibson042:
Proposal repo: https://github.com/tc39/proposal-immutable-arraybuffer

# Existing ArrayBuffer.prototype properties

- [ ] `ArrayBuffer.prototype.resize`
  - [ ] throws TypeError for an immutable receiver, before touching any argument

- [ ] `ArrayBuffer.prototype.slice`
  - [ ] throws a TypeError if the receiver constructs an immutable ArrayBuffer

- [ ] `ArrayBuffer.prototype.transfer`
  - [ ] throws TypeError for an immutable receiver, after processing _newLength_ (_**note the unusual receiver-after-argument order**_)

- [ ] `ArrayBuffer.prototype.transferToFixedLength`
  - [ ] throws TypeError for an immutable receiver, after processing _newLength_ (_**note the unusual receiver-after-argument order**_)

# New ArrayBuffer.prototype properties

- [ ] `ArrayBuffer.prototype.immutable`
  - [ ] getter-only accessor, enumerable false and configurable true
  - [ ] name "get immutable" and length 0
  - [ ] brand-checking (throws TypeError unless receiver is an ArrayBuffer and not a SharedArrayBuffer)
  - [ ] returns `true` for an immutable receiver and false for a non-immutable one
  - [ ] does not throw for a detached receiver

- [ ] `ArrayBuffer.prototype.sliceToImmutable`
  - [ ] `verifyPrimordialCallableProperty(ArrayBuffer.prototype, "sliceToImmutable", undefined, 2, { writable: true, enumerable: false, configurable: true })`
  - [ ] brand-checking (throws TypeError unless receiver is an ArrayBuffer and not a SharedArrayBuffer), before resolving bounds
  - [ ] throws TypeError if receiver is detached, before resolving bounds
  - [ ] resolves bounds after passing the preceding checks—first _start_, clamping negative values to [0, _len_ - 1] and positive values to [0, _len_], then _end_, defaulting to _len_ and clamping negative values to [0, _len_ - 1] and positive values to [0, _len_] (and capturing _len_ before coercing either _start_ or _end_)
  - [ ] throws TypeError if receiver was detached by bounds resolution, after bounds resolution
  - [ ] throws RangeError if length < specified upper bound, after passing the second detachment check and even if length was ≥ specified upper bound until bounds resolution
[...]

PR #4541, opened 2025-07-18
[immutable-arraybuffer] ArrayBuffer.prototype.immutable

gibson042:
Ref #4509

> 
> - [x] `ArrayBuffer.prototype.immutable`
>   - [x] getter-only accessor, enumerable false and configurable true
>   - [x] name "get immutable" and length 0
>   - [x] brand-checking (throws TypeError unless receiver is an ArrayBuffer and not a SharedArrayBuffer)
>   - [x] returns `true` for an immutable receiver and false for a non-immutable one
>   - [x] does not throw for a detached receiver

### DIFF
--- a/test/built-ins/ArrayBuffer/prototype/immutable/prop-desc.js
+++ b/test/built-ins/ArrayBuffer/prototype/immutable/prop-desc.js
@@ -1,0 +1,37 @@
+// Copyright (C) 2025 Richard Gibson. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-get-arraybuffer.prototype.immutable
+description: Checks the "immutable" property of ArrayBuffer.prototype.
+info: |
+  ArrayBuffer.prototype.immutable is an accessor property whose set accessor
+  function is undefined.
+
+  ECMAScript Standard Built-in Objects
+  ...
+  For functions that are specified as properties of objects, the name value is
+  the property name string used to access the function. Functions that are
+  specified as get or set accessor functions of built-in properties have "get"
+  or "set" (respectively) passed to the prefix parameter when calling
+  CreateBuiltinFunction.
+  ...
+  Every accessor property described in clauses 19 through 28 and in Annex B.2
+  has the attributes { [[Enumerable]]: false, [[Configurable]]: true } unless
+  otherwise specified. If only a get accessor function is described, the set
+  accessor function is the default value, undefined.
+features: [immutable-arraybuffer]
+includes: [propertyHelper.js]
+---*/
+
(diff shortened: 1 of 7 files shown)


test/built-ins/ArrayBuffer/prototype/immutable/prop-desc.js
```diff
@@ -0,0 +1,40 @@
...
+  accessor function is the default value, undefined.
+features: [immutable-arraybuffer]
+includes: [propertyHelper.js]
+---*/
+
+var desc = Object.getOwnPropertyDescriptor(ArrayBuffer.prototype, "immutable");
+var isHardened = Object.isFrozen(Object);
+if (isHardened && desc.get) {
+  Object.defineProperty(desc, "get", { configurable: false });
+}
```
Ms2ger: I don't think this should be in individual tests, if we support it at all.

Ms2ger (dismissed):
Thanks!

test/built-ins/ArrayBuffer/prototype/immutable/prop-desc.js
gibson042: Agreed on `defineProperty` (and removed), but varying the `configurable` assumption is pretty much necessary to avoid excessive work in a hardened environment.

bakkot:
@Ms2ger looks like your comment was addressed; should this now be "needs review" instead of "awaiting author"?

test/built-ins/ArrayBuffer/prototype/immutable/prop-desc.js
```diff
@@ -0,0 +1,37 @@
...
+features: [immutable-arraybuffer]
+includes: [propertyHelper.js]
+---*/
+
+var desc = Object.getOwnPropertyDescriptor(ArrayBuffer.prototype, "immutable");
+var isHardened = Object.isFrozen(Object);
+
+assert.sameValue(desc.value, undefined);
+assert.sameValue(desc.set, undefined);
+verifyCallableProperty(desc, "get", "get immutable", 0, { configurable: !isHardened });
```
ptomato: I think we wanted to isolate the accommodations for hardened JS environments (i.e. nonstandard) in the harness files as much as possible. Although `verifyPrimordialCallableProperty` doesn't seem quite right here, as `desc.get` isn't primordial.

That said, do we really need to check the configurable flag of `desc.get` here? If it's set incorrectly that seems more like an implementation issue with `Object.getOwnPropertyDescriptor` that should be covered in the tests for that API. So maybe `assert.sameValue(typeof desc.get, "function")` would be sufficient here.

WDYT?

ptomato (approved):
I'd prefer not to have the `isHardened` thing, and I have a suggestion for avoiding it. Otherwise all looks good.

PR #4542, opened 2025-07-18
[immutable-arraybuffer] ArrayBuffer.prototype.transferToImmutable

gibson042:
Ref #4509

> 
> - [x] `ArrayBuffer.prototype.transferToImmutable`
>   - [x] `verifyPrimordialCallableProperty(ArrayBuffer.prototype, "transferToImmutable", "transferToImmutable", 0);`
>   - [x] brand-checking (throws TypeError unless receiver is an ArrayBuffer and not a SharedArrayBuffer), before coercing _newLength_
>   - [x] defaults _newLength_ to [[ArrayBufferByteLength]]
>   - [x] if _newLength_ is not undefined, casts it to an index (integral number from 0 to 2**53 - 1 inclusive) or throws a RangeError for out-of-range values, after brand-checking
>   - [x] throws TypeError if receiver is detached or immutable, after processing _newLength_ (_**note the unusual receiver-after-argument order**_)
>   - [x] detaches the receiver
>   - [x] returns an immutable ArrayBuffer with the correct length
>   - [x] if _newLength_ ≤ [[ArrayBufferByteLength]], returns an ArrayBuffer with contents matching the appropriate prefix of the receiver up to exclusive index _newLength_
>   - [x] if _newLength_ > [[ArrayBufferByteLength]], returns an ArrayBuffer with contents matching those of the receiver, followed by the appropriate count of zeros

### DIFF
--- a/test/built-ins/ArrayBuffer/prototype/sliceToImmutable/argument-coercion.js
+++ b/test/built-ins/ArrayBuffer/prototype/sliceToImmutable/argument-coercion.js
@@ -118,7 +118,7 @@ for (var i = 0; i < goodInputs.length; i++) {
     var source = make32ByteArrayBuffer();
     var expectContents = Array.from(new Uint8Array(source)).slice(rawStart, rawEnd);
     var dest = source.sliceToImmutable(rawStart, rawEnd);
-    var label = "sliceToImmutable(" + assert._toString(rawStart) + ", " + assert._toString(rawEnd) + ")";
+    var label = "sliceToImmutable(" + formatSimpleValue(rawStart) + ", " + formatSimpleValue(rawEnd) + ")";
     assert.sameValue(dest.byteLength, intLength, label + ".byteLength");
     assert.compareArray(new Uint8Array(dest), expectContents, label + " contents");
     assert.sameValue(dest.immutable, true, label + ".immutable");
@@ -154,7 +154,7 @@ for (var i = 0; i < goodInputs.length; i++) {
     var source = make32ByteArrayBuffer();
     var expectContents = Array.from(new Uint8Array(source)).slice(rawStart, rawEnd);
     var dest = source.sliceToImmutable(paddedStart.string, paddedEnd.string);
-    var label = "sliceToImmutable(" + assert._toString(paddedStart.string) + ", " + assert._toString(paddedEnd.string) + ")";
+    var label = "sliceToImmutable(" + formatSimpleValue(paddedStart.string) + ", " + formatSimpleValue(paddedEnd.string) + ")";
     assert.sameValue(dest.byteLength, intLength, label + ".byteLength");
     assert.compareArray(new Uint8Array(dest), expectContents, label + " contents");
     assert.sameValue(dest.immutable, true, label + ".immutable");
@@ -199,8 +199,8 @@ for (var i = 0; i < goodInputs.length; i++) {
       }
     };
     function reprArgs(startMethodName, endMethodName) {
-      var startRepr = "{ " + startMethodName + ": () => " + assert._toString(rawStart) + " }";
-      var endRepr = "{ " + endMethodName + ": () => " + assert._toString(rawEnd) + " }";
+      var startRepr = "{ " + startMethodName + ": () => " + formatSimpleValue(rawStart) + " }";
+      var endRepr = "{ " + endMethodName + ": () => " + formatSimpleValue(rawEnd) + " }";
       return "(" + startRepr + ", " + endRepr + ")";
     }
 
@@ -324,10 +324,10 @@ for (var i = 0; i < badInputs.length; i++) {
     function() {
       ab.sliceToImmutable(rawBad, objGood);
     },
-    "sliceToImmutable(" + assert._toString(rawBad) + ", { valueOf: () => " + assert._toString(rawGood) + " })"
+    "sliceToImmutable(" + formatSimpleValue(rawBad) + ", { valueOf: () => " + formatSimpleValue(rawGood) + " })"
   );
   assert.compareArray(calls, [],
-    "sliceToImmutable(" + assert._toString(rawBad) + ", { valueOf: () => " + assert._toString(rawGood) + " }) calls");
+    "sliceToImmutable(" + formatSimpleValue(rawBad) + ", { valueOf: () => " + formatSimpleValue(rawGood) + " }) calls");
 
   calls = [];
   assert.throws(
@@ -335,10 +335,10 @@ for (var i = 0; i < badInputs.length; i++) {
     function() {
       ab.sliceToImmutable(objGood, rawBad);
     },
-    "sliceToImmutable({ valueOf: () => " + assert._toString(rawGood) + " }, " + assert._toString(rawBad) + ")"
+    "sliceToImmutable({ valueOf: () => " + formatSimpleValue(rawGood) + " }, " + formatSimpleValue(rawBad) + ")"
   );
   assert.compareArray(calls, ["good.valueOf"],
-    "sliceToImmutable({ valueOf: () => " + assert._toString(rawGood) + " }, " + assert._toString(rawBad) + ") calls");
+    "sliceToImmutable({ valueOf: () => " + formatSimpleValue(rawGood) + " }, " + formatSimpleValue(rawBad) + ") calls");
 }
 
 var calls = [];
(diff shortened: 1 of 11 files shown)


test/built-ins/ArrayBuffer/prototype/transferToImmutable/new-length-coercion.js
```diff
@@ -0,0 +1,262 @@
...
+ab = new ArrayBuffer(8);
+assert.sameValue(ab.transferToImmutable(undefined).byteLength, 8,
+  "Must default undefined to receiver byteLength.");
+
+function repr(value) {
+  if (typeof value === "string") return JSON.stringify(value);
+  if (typeof value === "bigint") return String(value) + "n";
+  if (!value && 1/value === -Infinity) return "-0";
+  return String(value);
+}
```
Ms2ger: I don't think this belongs here

test/built-ins/ArrayBuffer/prototype/transferToImmutable/new-length-coercion.js
```diff
@@ -0,0 +1,262 @@
...
+
+var goodLengths = [
```
Ms2ger, line 42: I'm willing to accept this test, but I think it would be nicer if we could have a generic helper function that runs through a set of interesting ToIndex arguments; this isn't the only API that could be tested in this way.

test/built-ins/ArrayBuffer/prototype/transferToImmutable/to-larger.js
```diff
@@ -0,0 +1,75 @@
...
+  },
+  ArrayBuffer.prototype.resize && function grown() {
+    var ab = new ArrayBuffer(byteLength / 2, { maxByteLength: byteLength });
+    ab.resize(byteLength);
+    return ab;
+  }
+];
+
+for (var i = 0; i < sourceMakers.length; i++) {
+  var name = sourceMakers[i].name;
```
Ms2ger, line 57: This doesn't take into account that `sourceMakers[i]` might be `undefined`.

test/built-ins/ArrayBuffer/prototype/transferToImmutable/to-larger.js
```diff
@@ -0,0 +1,75 @@
...
+  var newLength = byteLength + i + 1;
```
Ms2ger: Calculating the argument based on the index in the array seems a recipe for confusion

test/built-ins/ArrayBuffer/prototype/transferToImmutable/to-same-or-smaller.js
```diff
@@ -0,0 +1,78 @@
...
+  },
+  ArrayBuffer.prototype.resize && function grown() {
+    var ab = new ArrayBuffer(byteLength / 2, { maxByteLength: byteLength });
+    ab.resize(byteLength);
+    return ab;
+  }
+];
+
+for (var i = 0; i < sourceMakers.length; i++) {
+  var name = sourceMakers[i].name;
```
Ms2ger, line 56: Same here

test/built-ins/ArrayBuffer/prototype/transferToImmutable/to-same-or-smaller.js
```diff
@@ -0,0 +1,78 @@
...
+  for (
+    var newLength = undefined;
+    (newLength || 0) < byteLength;
+    newLength = newLength === undefined ? 0 : newLength + 1
+  ) {
```
Ms2ger: This is very convoluted, probably too much so

Ms2ger (dismissed):

test/built-ins/ArrayBuffer/prototype/sliceToImmutable/argument-coercion.js
```diff
@@ -154,7 +150,7 @@ for (var i = 0; i < goodInputs.length; i++) {
     var source = make32ByteArrayBuffer();
     var expectContents = Array.from(new Uint8Array(source)).slice(rawStart, rawEnd);
     var dest = source.sliceToImmutable(paddedStart.string, paddedEnd.string);
-    var label = "sliceToImmutable(" + repr(paddedStart.string) + ", " + repr(paddedEnd.string) + ")";
+    var label = "sliceToImmutable(" + assert._toString(paddedStart.string) + ", " + assert._toString(paddedEnd.string) + ")";
```
ptomato: I'm guessing the underscore in `assert._toString()` means "this isn't intended to be public". Maybe it's best to keep `repr`?

test/built-ins/ArrayBuffer/prototype/transferToImmutable/new-length-coercion.js
```diff
@@ -0,0 +1,255 @@
...
+  assert.sameValue(ab.transferToImmutable(objLength).byteLength, intLength,
+    "transferToImmutable({ [Symbol.toPrimitive]: () => " + assert._toString(rawLength) + " }).byteLength");
+  assert.compareArray(calls, ["Symbol.toPrimitive(number)"],
+    "transferToImmutable({ [Symbol.toPrimitive]: () => " + assert._toString(rawLength) + " }) calls");
+}
+
+var badLengths = [
+  // Out of range numbers
+  [-1, RangeError],
+  [9007199254740992, RangeError], // Math.pow(2, 53) = 9007199254740992
```
ptomato, line 145: Might be more readable as `Number.MAX_SAFE_INTEGER + 1`? (or maybe you're trying to avoid any builtins in favour of number literals?)

test/built-ins/ArrayBuffer/prototype/transferToImmutable/new-length-coercion.js
```diff
@@ -0,0 +1,255 @@
...
+    ab.transferToImmutable(rawLength);
+  }, "transferToImmutable(" + assert._toString(rawLength) + ")");
+}
+
+for (var i = 0; i < badLengths.length; i++) {
+  if (!badLengths[i]) continue;
+  var rawLength = badLengths[i][0];
+  var expectedErr = badLengths[i][1];
+  if (typeof rawLength !== "number") continue;
+  rawLength = isNegativeZero(rawLength) ? "-0" : String(rawLength);
```
ptomato: The -0 test is not needed here, I think? -0 is a "good" length

ptomato (commented):
Didn't get all the way through this, but here are my comments so far.

I think the thing that most needs attention is `assert._toString` - not sure we want to be encouraging tests to use that.

test/built-ins/ArrayBuffer/prototype/sliceToImmutable/argument-coercion.js
gibson042: This was in response to https://github.com/tc39/test262/pull/4542#discussion_r2293646039 , which I consider a fair criticism... many tests need a way to format values for readable labels, basically a rudimentary [`util.inspect`](https://nodejs.org/docs/latest/api/util.html#utilinspectobject-options). But you're also right that using an underscore method on `assert` isn't the best way to do so. Should we document harness/assert.js as defining something like `prettyFormat`?

test/built-ins/ArrayBuffer/prototype/transferToImmutable/new-length-coercion.js
gibson042: > maybe you're trying to avoid any builtins in favour of number literals?

Yes, exactly. I think there's precedent as well: https://github.com/search?q=repo%3Atc39%2Ftest262%209007199254740992&type=code

test/built-ins/ArrayBuffer/prototype/transferToImmutable/new-length-coercion.js
gibson042: Good point; removed.

test/built-ins/ArrayBuffer/prototype/sliceToImmutable/argument-coercion.js
ptomato: I don't love adding more complexity just for assertion messages, but I don't hate the idea either :smile:

gibson042: OK, done. I ended up just adding the functions backing `assert._formatIdentityFreeValue` and `assert._toString` to harness/assert.js `defines` as `formatIdentityFreeValue` and `formatSimpleValue` (respectively), leaving future space for formatting non-simple values using e.g. object literal syntax.

ptomato (approved):

PR #4544, opened 2025-07-22
Add tests for WebKit Bug 242740: simultaneous imports with top-level await

dmvjs:
# Add tests for WebKit Bug 242740: simultaneous imports with top-level await

## Summary

Adds test cases that reproduce WebKit Bug 242740, where simultaneous imports of a module with top-level await can fail with ReferenceError or resolve prematurely.

## Bug Reference

- **WebKit Bug 242740**: https://bugs.webkit.org/show_bug.cgi?id=242740
- **Description**: ReferenceError when multiple modules are simultaneously importing a module containing a top-level await

## Tests Added

1. **`simultaneous-imports-race-condition.js`** - Tests that simultaneous imports wait for top-level await completion
2. **`simultaneous-imports-reference-error.js`** - Tests that no ReferenceError is thrown during simultaneous imports

## Expected Behavior

- All simultaneous imports should wait for the module's top-level await to complete
- No ReferenceError should be thrown during the import process
- All imports should return the same module namespace
- Module exports should be correctly defined

## Affected Browsers

- Safari (WebKit-based browsers) - will fail these tests
- Chrome, Firefox, Edge - should pass these tests


nicolo-ribaudo:
Is this behaviour already covered by https://github.com/tc39/test262/blob/main/test/language/module-code/top-level-await/dynamic-import-of-waiting-module.js? Maybe just add the namespace check to that test?

ptomato (changes requested):
I'm not sure this test is correct as is. The `importCount` assertion looks iffy to me — shouldn't the module only get executed once, no matter how many times it's imported? FWIW, SM and V8 fail that assertion despite not being subject to the WebKit bug.

PR #4545, opened 2025-07-23
[immutable-arraybuffer] Include immutable ArrayBuffers in TypedArray testing

gibson042:
Extends #4921 ([diff](https://github.com/gibson042/test262/compare/2026-02-generalize-testwithtypedarrayconstructors...gibson042:test262:2025-07-generalize-testwithtypedarrayconstructors))

This is foundational for many remaining tests of #4509, particularly coverage of existing %TypedArray%.prototype properties.

Best reviewed commit-by-commit, because much of the work is mechanical text replacement.

### DIFF
--- a/harness/testTypedArray.js
+++ b/harness/testTypedArray.js
@@ -333,9 +333,19 @@ var nonAtomicsFriendlyTypedArrayConstructors = floatArrayConstructors.concat([Ui
  *
  * @param {typedArrayConstructorCallback} f - the function to call for each typed array constructor.
  * @param {Array} selected - An optional Array with filtered typed arrays
+ * @param {typedArrayArgFactoryFeature[]} [includeArgFactories] - for selecting
+ *   initial constructor argument factory functions, rather than starting with
+ *   all argument factories
+ * @param {typedArrayArgFactoryFeature[]} [excludeArgFactories] - for excluding
+ *   constructor argument factory functions, after an initial selection
  */
-function testWithNonAtomicsFriendlyTypedArrayConstructors(f) {
-  testWithTypedArrayConstructors(f, nonAtomicsFriendlyTypedArrayConstructors);
+function testWithNonAtomicsFriendlyTypedArrayConstructors(f, includeArgFactories, excludeArgFactories) {
+  testWithAllTypedArrayConstructors(
+    f,
+    nonAtomicsFriendlyTypedArrayConstructors,
+    includeArgFactories,
+    excludeArgFactories
+  );
 }
 
 /**
@@ -343,16 +353,26 @@ function testWithNonAtomicsFriendlyTypedArrayConstructors(f) {
  *
  * @param {typedArrayConstructorCallback} f - the function to call for each typed array constructor.
  * @param {Array} selected - An optional Array with filtered typed arrays
+ * @param {typedArrayArgFactoryFeature[]} [includeArgFactories] - for selecting
+ *   initial constructor argument factory functions, rather than starting with
+ *   all argument factories
+ * @param {typedArrayArgFactoryFeature[]} [excludeArgFactories] - for excluding
+ *   constructor argument factory functions, after an initial selection
  */
-function testWithAtomicsFriendlyTypedArrayConstructors(f) {
-  testWithTypedArrayConstructors(f, [
-    Int32Array,
-    Int16Array,
-    Int8Array,
-    Uint32Array,
-    Uint16Array,
-    Uint8Array,
-  ]);
+function testWithAtomicsFriendlyTypedArrayConstructors(f, includeArgFactories, excludeArgFactories) {
+  testWithAllTypedArrayConstructors(
+    f,
+    [
+      Int32Array,
+      Int16Array,
+      Int8Array,
+      Uint32Array,
+      Uint16Array,
+      Uint8Array,
+    ],
+    includeArgFactories,
+    excludeArgFactories
+  );
 }
 
 /**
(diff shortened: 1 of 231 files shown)


harness/testTypedArray.js
```diff
@@ -42,39 +42,214 @@ if (typeof Float16Array !== 'undefined') {
...
+function makePassthrough(TA, primitiveOrIterable) {
+  return primitiveOrIterable;
+}
+
+function makeArray(TA, primitiveOrIterable) {
+  if (isPrimitive(primitiveOrIterable)) {
+    var n = Number(primitiveOrIterable);
+    // Only values between 0 and 2**53 - 1 inclusive can get mapped into TA contents.
+    if (!(n >= 0 && n < 9007199254740992)) return primitiveOrIterable;
+    return Array.from({ length: n }, function(){ return "0"; });
```
Ms2ger:
```suggestion
    return Array.from({ length: n }, function() { return "0"; });
```

harness/testTypedArray.js
```diff
@@ -75,6 +79,17 @@ function testWithTypedArrayConstructors(f, selected) {
...
 }
 
+/**
+ * Calls the provided function for every BigInt typed array constructor.
+ *
+ * @param {typedArrayConstructorCallback} f - the function to call
+ * @param {Array} constructors - An optional Array with filtered typed arrays
+ */
+function testWithBigIntTypedArrayConstructors(f, constructors) {
+  if (!constructors) constructors = [BigInt64Array, BigUint64Array];
```
ptomato: Re. the merge of testBigIntTypedArray.js into testTypedArray.js, fair enough! I guess they were separated originally so that implementations without BigInt could still run the rest of the TypedArray tests, but this seems like a good solution that preserves that capability, as there won't be a ReferenceError unless you actually call `testWithBigIntTypedArrayConstructors`.

harness/testTypedArray.js
```diff
@@ -42,39 +42,214 @@ if (typeof Float16Array !== 'undefined') {
...
+        return argFactory === makeArray || argFactory === makeArrayLike;
+      case "iterable":
+        return argFactory === makeIterable;
+      case "arraybuffer":
+        return (
+          argFactory === makeArrayBuffer ||
+          argFactory === makeResizableArrayBuffer ||
+          argFactory === makeGrownArrayBuffer ||
+          argFactory === makeShrunkArrayBuffer ||
+          argFactory === makeImmutableArrayBuffer
```
ptomato, line 205: Nitpick, `makeImmutableArrayBuffer` isn't yet available at this point in the commit stack.

harness/testTypedArray.js
```diff
@@ -42,39 +42,214 @@ if (typeof Float16Array !== 'undefined') {
...
+        );
+      case "resizable":
+        return (
+          argFactory === makeResizableArrayBuffer ||
+          argFactory === makeGrownArrayBuffer ||
+          argFactory === makeShrunkArrayBuffer
+        );
+      case "immutable":
+        return argFactory === makeImmutableArrayBuffer;
```
ptomato: Ditto

harness/testTypedArray.js
```diff
@@ -42,39 +42,214 @@ if (typeof Float16Array !== 'undefined') {
...
+}
+
 /**
  * Callback for testing a typed array constructor.
  *
  * @callback typedArrayConstructorCallback
- * @param {Function} Constructor the constructor object to test with.
+ * @param {Function} TypedArrayConstructor the constructor object to test with
+ * @param {Function} [TypedArrayConstructorArgFactory] a function for making
+ *   a TypedArrayConstructor arguments from a primitive (usually a number) or
```
ptomato: Typo: arguments → argument

harness/testTypedArray.js
```diff
@@ -42,39 +42,214 @@ if (typeof Float16Array !== 'undefined') {
...
+ * number) or iterable (usually an array) into a value suitable as the first
+ * argument of typedArrayCtor (an Array, arraylike, iterable, or ArrayBuffer).
  *
  * typedArrayCtor will not be BigInt64Array or BigUint64Array unless one or both
  * of those are explicitly provided.
  *
- * @param {typedArrayConstructorCallback} f - the function to call for each typed array constructor.
- * @param {Array} selected - An optional Array with filtered typed arrays
+ * @param {typedArrayConstructorCallback} f - the function to call
+ * @param {Array} [constructors] - an explict list of TypedArray constructors
```
ptomato: Typo: explict → explicit

ptomato (commented):
The new testTypedArrayConstructors helper could use some more explanation in
doc comments. For example, do I understand correctly that the overall purpose is
to expand the test matrix to not only test each piece of functionality with as
many TypedArray constructors, as possible, but also test with as many different
argument types as possible to each TypedArray constructor?

I don't understand the purpose of passing `, null, ["passthrough"]` in tests
that don't even use the factory argument. (It seems like some but not all
instances of that are removed in 53a9ecb2?)

Particularly in 5855df79 (but also b7e4f20f8) I think a prose description of
what I'm looking at would be more helpful than the contents of the sed script,
although it's probably good to include the script for reference. I can read the
sed scripts eventually, but given limited time to spend on review that's not the
most effective place for me to spend it on.

It might be helpful to split this PR into two, where one is just a refactor of
the helper and increase of coverage in the existing tests without immutable ABs,
and the second adds the facilities for immutable ABs to the helper.

ptomato (approved):
OK, I've looked this over again and all the issues seem to have been resolved.

PR #4546, opened 2025-07-23
Try to avoid "Argument list too long" when many files are changed

gibson042:
As seen at https://github.com/tc39/test262/actions/runs/16462995254/job/46533977407?pr=4545 for the 1577 changed files of #4545


Ms2ger:
I don't feel confident reviewing the shell code here; can we move this into python somehow?

ljharb (approved):
shell code lgtm; i wouldn't be able to review it if moved to python, but don't let that stop you if you feel it best :-)

PR #4547, opened 2025-07-23
[immutable-arraybuffer] Coverage for read-only TypedArray methods

gibson042:
Ref #4509

> # %TypedArray%.prototype properties
> …
> - [x] read-only `%TypedArray%.prototype` properties
>   - [x] work as expected if the receiver is backed by an immutable ArrayBuffer
>   - [x] `%TypedArray%.prototype.filter` throws a TypeError if the receiver constructs an instance backed by an immutable ArrayBuffer, after determining contents
>   - [x] `%TypedArray%.prototype.map` throws a TypeError if the receiver constructs an instance backed by an immutable ArrayBuffer, after validating that _callback_ is callable but before invoking it
>   - [x] `%TypedArray%.prototype.slice` throws a TypeError if the receiver constructs an instance backed by an immutable ArrayBuffer, after coercing bounds

Note that the "Cover existing read-only TypedArray methods" commit is large but mostly mechanical, the others are small and actually specific to immutable ArrayBuffer.

### DIFF
--- a/test/built-ins/TypedArray/prototype/filter/speciesctor-destination-backed-by-immutable-buffer.js
+++ b/test/built-ins/TypedArray/prototype/filter/speciesctor-destination-backed-by-immutable-buffer.js
@@ -1,0 +1,82 @@
+// Copyright (C) 2025 Richard Gibson. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-%typedarray%.prototype.filter
+description: >
+  Throws a TypeError exception when the receiver constructs an instance backed
+  by an immutable buffer
+info: |
+  %TypedArray%.prototype.filter ( callback [ , thisArg ] )
+  ...
+  8. Repeat, while k < len,
+     a. Let Pk be ! ToString(𝔽(k)).
+     b. Let kValue be ! Get(O, Pk).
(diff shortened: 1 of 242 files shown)


test/built-ins/TypedArray/prototype/filter/speciesctor-destination-backed-by-immutable-buffer.js
```diff
@@ -0,0 +1,78 @@
...
+      return constructor;
+    }
+  });
+  Object.defineProperty(constructor, Symbol.species, {
+    get: function() {
+      calls.push("get ta.constructor[Symbol.species]");
+      var iab = (new TA(["3", "4"])).buffer.transferToImmutable();
+      var result = new TA(iab);
+      calls.push("construct result");
+      return result;
```
iamstolis: I wondered why this test passes even without implementing the relevant changes in `%TypedArray%.prototype.filter`. The root of the problem is here. This `Symbol.species` getter doesn't return constructor (it returns a typed array). Not returning a constructor results in the expected `TypeError` (before we have any chance to get to the `IsImmutableBuffer()` that's supposed to be checked by this test).

In other words, the getter should not return `result` but a function that returns `result`.

The same problem is in the `slice` and `map` versions of this test.

ptomato (approved):
I did only skim the large commit, but the newly added tests look correct to me. Don't forget to autosquash before merging.

PR #4548, opened 2025-07-24
[immutable-arraybuffer] Coverage for mutating TypedArray methods

gibson042:
Ref #4509

> # %TypedArray%.prototype properties
> 
> - [x] `%TypedArray%.prototype.{copyWithin,fill,reverse,set}`
>   - [x] throws a TypeError if the receiver is backed by an immutable ArrayBuffer, before touching any argument
>   - [x] ...even if the targeted range is of length 0
> 
> - [x] `%TypedArray%.prototype.sort`
>   - [x] throws a TypeError if the receiver is backed by an immutable ArrayBuffer, after validating that _comparator_ is callable (_**note the unusual receiver-after-argument order**_)
>   - [x] ...even if the receiver is of length 0
> 
> - [x] `Uint8Array.prototype.{setFromBase64,setFromHex}`
>   - [x] throws a TypeError if the receiver is backed by an immutable ArrayBuffer, before touching any argument

### DIFF
--- a/test/built-ins/TypedArray/prototype/reverse/immutable-buffer.js
+++ b/test/built-ins/TypedArray/prototype/reverse/immutable-buffer.js
@@ -1,0 +1,28 @@
+// Copyright (C) 2025 Richard Gibson. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-%typedarray%.prototype.reverse
+description: Throws a TypeError exception when the backing buffer is immutable
+info: |
+  %TypedArray%.prototype.reverse ( )
+  1. Let O be the this value.
+  2. Let taRecord be ? ValidateTypedArray(O, ~seq-cst~, ~write~).
+
+  ValidateTypedArray ( O, order [ , accessMode ] )
+  1. If accessMode is not present, set accessMode to read.
+  2. Perform ? RequireInternalSlot(O, [[TypedArrayName]]).
+  3. Assert: O has a [[ViewedArrayBuffer]] internal slot.
+  4. If accessMode is ~write~ and IsImmutableBuffer(O.[[ViewedArrayBuffer]]) is true, throw a TypeError exception.
+features: [TypedArray, immutable-arraybuffer]
+includes: [testTypedArray.js, compareArray.js]
+---*/
+
+testWithAllTypedArrayConstructors((TA, makeCtorArg) => {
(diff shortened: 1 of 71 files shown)


test/built-ins/TypedArray/prototype/reverse/immutable-buffer.js
```diff
@@ -0,0 +1,31 @@
...
+
+testWithAllTypedArrayConstructors((TA, makeCtorArg) => {
+  var calls = [];
+
+  var ta = new TA(makeCtorArg(["1", "2", "3", "4"]));
+
+  assert.throws(TypeError, function() {
+    ta.reverse();
+  });
+  assert.compareArray(calls, [], "Must verify mutability before reading arguments.");
```
ptomato: There aren't any arguments to read, or any reentrancy into user code at all for that matter, so this is a bit overkill. Maybe delete `calls` and its assertion?

test/built-ins/Uint8Array/prototype/setFromBase64/throws-when-target-is-backed-by-immutable-arraybuffer.js
```diff
@@ -1,19 +1,41 @@
...
+      return undefined;
+    },
+    get lastChunkHandling() {
+      calls.push("get options.lastChunkHandling");
+      return undefined;
+    },
+  };
 
-assert.throws(TypeError, function () {
-  target.setFromBase64('');
```
ptomato, line 18: Looks like we're losing the coverage for `setFromBase64('')`, is it significant?

test/built-ins/Uint8Array/prototype/setFromHex/throws-when-target-is-backed-by-immutable-arraybuffer.js
```diff
@@ -1,19 +1,25 @@
...
+testWithAllTypedArrayConstructors((TA, makeCtorArg) => {
+  var calls = [];
 
-assert.throws(TypeError, function () {
-  target.setFromHex('aa');
-});
+  var ta = new TA(makeCtorArg(["1", "2", "3", "4"]));
 
-assert.throws(TypeError, function () {
-  target.setFromHex('');
```
ptomato, line 18: Likewise here, for `''`

ptomato (approved):
Minor comments, feel free to merge when addressed.

PR #4549, opened 2025-07-24
[immutable-arraybuffer] TypedArray static functions

gibson042:
Ref #4509

> # %TypedArray% static properties
> 
> - [x] `%TypedArray%.from`
>   - [x] throws a TypeError if the receiver constructs an immutable ArrayBuffer, after determining the length from _source_ but before checking if the constructed ArrayBuffer is too short or (more testably) calling _mapper_
> 
> - [x] `%TypedArray%.of`
>   - [x] throws a TypeError if the receiver constructs an immutable ArrayBuffer, after determining the length from _source_ but before checking if the constructed ArrayBuffer is too short or (more testably) coercing any value to bigint/number


ptomato (approved):
Comprehensive, thanks!

PR #4551, opened 2025-07-29
[immutable-arraybuffer] TypedArray internal operations

gibson042:
Ref #4509

> # %TypedArray% instance operations
> 
> - [x] [[GetOwnProperty]]
>   - [x] property descriptors for in-bounds indexes of a typed array backed by an immutable ArrayBuffer are reported as writable false, enumerable true, configurable false
> 
> - [x] [[DefineOwnProperty]]
>   - [x] attempts to redefine own properties for in-bounds indexes of a typed array backed by an immutable ArrayBuffer are validated against descriptors for immutable properties
>     - [x] presence of `writable: true`, `enumerable: false`, `configurable: true`, `get`, and/or `set` results in rejection
>     - [x] presence of `value` with any value that is not treated as identical to the current value by `Object.is` results in rejection (e.g., `value: -0` and `value: "0"` are rejected when the current value is `0`)
>     - [x] all other property descriptors (e.g., any subset of `{ writable: false, enumerable: true, configurable: false, value: $currentValue }`) are accepted but do not result in observable changes
> 
> - [x] [[Set]]
>   - [x] any attempt to set the value for a property whose name is a canonical numeric index string (including non-index values such as `"-0"`, `"1.5"`, `"590295810358705700000"`, `"5e-324"`, `"1.7976931348623157e+308"`, `"Infinity"`, and `"NaN"`) on a typed array backed by an immutable ArrayBuffer is rejected, even if the new value is identical to the current value
>   - [x] likewise for attempts on a receiver having such a typed array in its prototype chain when the [[Set]] makes it to the typed array

Most of the changes just relax filters applied to `testWithTypedArrayConstructors` and similar harness/testTypedArray.js exports such that immutable ArrayBuffers are no longer excluded.

### DIFF
--- a/harness/propertyHelper.js
+++ b/harness/propertyHelper.js
@@ -25,6 +25,7 @@ defines:
 // Capture primordial functions and receiver-uncurried primordial methods that
 // are used in verification but might be destroyed *by* that process itself.
 var __isArray = Array.isArray;
+var __max = Math.max;
 var __defineProperty = Object.defineProperty;
 var __getOwnPropertyDescriptor = Object.getOwnPropertyDescriptor;
 var __getOwnPropertyNames = Object.getOwnPropertyNames;
@@ -32,7 +33,12 @@ var __join = Function.prototype.call.bind(Array.prototype.join);
 var __push = Function.prototype.call.bind(Array.prototype.push);
 var __hasOwnProperty = Function.prototype.call.bind(Object.prototype.hasOwnProperty);
 var __propertyIsEnumerable = Function.prototype.call.bind(Object.prototype.propertyIsEnumerable);
-var nonIndexNumericPropertyName = Math.pow(2, 32) - 1;
+var __getTypedArrayByteLength =
+  Object.getPrototypeOf &&
+  typeof Uint8Array !== "undefined" &&
+  Function.prototype.call.bind(
+    __getOwnPropertyDescriptor(Object.getPrototypeOf(Uint8Array.prototype), "byteLength").get
+  );
 
 /**
  * @param {object} obj
@@ -119,7 +125,7 @@ function verifyProperty(obj, name, desc, options) {
 
   if (__hasOwnProperty(desc, 'configurable') && desc.configurable !== undefined) {
     if (desc.configurable !== originalDesc.configurable ||
-        desc.configurable !== isConfigurable(obj, name)) {
+        desc.configurable !== isConfigurable(obj, name, originalDesc)) {
       __push(failures, label + " descriptor should " + (desc.configurable ? '' : 'not ') + "be configurable");
     }
   }
@@ -135,7 +141,30 @@ function verifyProperty(obj, name, desc, options) {
   return true;
 }
 
-function isConfigurable(obj, name) {
+function _toNumber(value) {
+  // https://tc39.es/ecma262/multipage/numbers-and-dates.html#sec-math.max
+  return __max(value);
+}
+
+// https://tc39.es/ecma262/multipage/abstract-operations.html#sec-canonicalnumericindexstring
+function _isCanonicalNumericIndexString(value) {
+  if (typeof value !== "string") return false;
+  if (value === "-0") return true;
+  var n = _toNumber(value);
+  return String(n) === value;
+}
+
+function _isTypedArray(value) {
+  try {
+    // https://tc39.es/ecma262/multipage/indexed-collections.html#sec-get-%typedarray%.prototype.bytelength
+    __getTypedArrayByteLength(value);
+    return true;
+  } catch (_err) {
+    return false;
+  }
+}
+
+function isConfigurable(obj, name, originalDesc) {
   try {
     delete obj[name];
   } catch (e) {
@@ -143,7 +172,17 @@ function isConfigurable(obj, name) {
       throw new Test262Error("Expected TypeError, got " + e);
     }
   }
-  return !__hasOwnProperty(obj, name);
+  var deleted = !__hasOwnProperty(obj, name);
+
+  // TypedArray canonical numeric string properties are never *actually* deleted,
+  // so we skip post-hoc verification for them
+  // https://tc39.es/ecma262/multipage/ordinary-and-exotic-objects-behaviours.html#sec-typedarray-delete
+  if (_isCanonicalNumericIndexString(name) && _isTypedArray(obj) && originalDesc) {
+    if (deleted) throw new Test262Error("Expected TypedArray index deletion to be ignored");
+    return originalDesc.configurable;
+  }
+
+  return deleted;
 }
 
 function isEnumerable(obj, name) {
@@ -172,17 +211,32 @@ function isSameValue(a, b) {
 }
 
 function isWritable(obj, name, verifyProp, value) {
-  var unlikelyValue = __isArray(obj) && name === "length" ?
-    nonIndexNumericPropertyName :
-    "unlikelyValue";
-  var newValue = value || unlikelyValue;
   var hadValue = __hasOwnProperty(obj, name);
   var oldValue = obj[name];
-  var writeSucceeded;
-
-  if (arguments.length < 4 && newValue === oldValue) {
-    newValue = newValue + "2";
+  var newValue = value;
+  if (arguments.length < 4) {
+    // Initialize newValue, preserving a numeric type from oldValue with a
+    // replacement that moves towards zero (to support the special behaviour of
+    // Array and TypedArray "length" and index properties).
+    switch (typeof oldValue) {
+      case "number":
+        // As a special case, never decrease array length (which would delete
+        // properties not covered by the new length).
+        if (oldValue <= 0 || __isArray(obj) && name === "length") {
+          newValue = oldValue === -Infinity ? 0 : oldValue + 1;
+        } else {
+          newValue = oldValue === Infinity ? 1 : oldValue - 1;
+        }
+        break;
+      case "bigint":
+        newValue = oldValue <= BigInt(0) ? oldValue + BigInt(1) : oldValue - BigInt(1);
+        break;
+      default:
+        newValue = oldValue !== "unlikelyValue" ? "unlikelyValue" : "unlikelyValue2";
+        break;
+    }
   }
+  var writeSucceeded;
 
   try {
     obj[name] = newValue;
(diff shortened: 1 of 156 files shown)


test/built-ins/TypedArrayConstructors/internals/GetOwnProperty/immutable-buffer.js
```diff
@@ -0,0 +1,43 @@
...
+---*/
+
+testWithAllTypedArrayConstructors(function(TA, makeCtorArg) {
+  var sample = new TA(makeCtorArg(["42", "43"]));
+
+  var value0 = sample[0];
+  var value1 = sample[1];
+  assert.compareArray([String(value0), String(value1)], ["42", "43"]);
+
+  verifyProperty(sample, "0", {
```
iamstolis, line 31: Note that `verifyProperty()` is based on various assumptions that don't work for typed arrays. For example, `isWritable(new Uint8Array(8), "0")` is `false` (incorrectly) and `isWritable(new BigInt64Array(8), "0")` even results in `SyntaxError: Cannot convert unlikelyValue to a BigInt.` 

The current version of the test seems to be lucky to pass despite the usage of `verifyProperty()`, but it is probably better to avoid it (for typed arrays). Note that, for example, `index-prop-desc.js` that is kind of mutable version of this test avoids `verifyProperty()` for the mentioned reasons.

test/built-ins/TypedArrayConstructors/internals/GetOwnProperty/immutable-buffer.js
```diff
@@ -0,0 +1,43 @@
...
+        ii. If value is undefined, return undefined.
+        iii. Let mutable be true.
+        iv. If IsImmutableBuffer(O.[[ViewedArrayBuffer]]) is true, set mutable to false.
+        v. Return the PropertyDescriptor { [[Value]]: value, [[Writable]]: true mutable,
+           [[Enumerable]]: true, [[Configurable]]: true mutable }.
+features: [TypedArray, immutable-arraybuffer]
+includes: [testTypedArray.js, compareArray.js, propertyHelper.js]
+---*/
+
+testWithAllTypedArrayConstructors(function(TA, makeCtorArg) {
```
iamstolis, line 24: Other tests in the same directory do not test `BigInt` typed arrays i.e. use `testWithTypedArrayConstructors` instead of `testWithAllTypedArrayConstructors` (and contain a variant with `testWithBigIntTypedArrayConstructors` in `BigInt` sub-directory).

test/built-ins/TypedArrayConstructors/internals/GetOwnProperty/immutable-buffer.js
gibson042: I just fixed `verifyProperty` to work with TypedArray input.

test/built-ins/TypedArrayConstructors/internals/GetOwnProperty/immutable-buffer.js
gibson042: Right, in the long term I'll be removing the "bigint" sub-directories to make the testing more consolidated.

ptomato:
Needs a non-trivial rebase.

gibson042:
> Needs a non-trivial rebase.

Done.

harness/propertyHelper.js
```diff
@@ -135,15 +141,48 @@ function verifyProperty(obj, name, desc, options) {
   return true;
 }
 
-function isConfigurable(obj, name) {
+function toNumber(value) {
```
ptomato: Should be `__toNumber` unless we want to expose it more broadly?

harness/propertyHelper.js
```diff
@@ -135,15 +141,48 @@ function verifyProperty(obj, name, desc, options) {
...
+  // https://tc39.es/ecma262/multipage/numbers-and-dates.html#sec-math.max
+  return __max(value);
+}
+
+// https://tc39.es/ecma262/multipage/abstract-operations.html#sec-canonicalnumericindexstring
+function isCanonicalNumericIndexString(value) {
```
ptomato: Likewise, `__isCanonicalNumericIndexString`?

harness/propertyHelper.js
```diff
@@ -135,15 +141,48 @@ function verifyProperty(obj, name, desc, options) {
...
+  if (typeof value !== "string") return false;
+  if (value === "-0") return true;
+  var n = toNumber(value);
+  return String(n) === value;
+}
+
+function isTypedArray(value) {
```
ptomato: Likewise, `__isTypedArray`?

harness/propertyHelper.js
```diff
@@ -172,17 +211,28 @@ function isSameValue(a, b) {
 }
 
 function isWritable(obj, name, verifyProp, value) {
-  var unlikelyValue = __isArray(obj) && name === "length" ?
-    nonIndexNumericPropertyName :
```
ptomato, line 176: `nonIndexNumericPropertyName` is no longer used, I think - is that intentional? Seems like it still could/should be used for array `length` properties.

ptomato (commented):
I haven't gotten most of the way through this yet, but here are some comments. I also don't understand what the purpose is of the first three commits - the first one ("Make better use of testTypedArray.js") seems like it's actually removing coverage in some cases?

harness/propertyHelper.js
gibson042: Yes, that is intentional. It's only use was for `isWritable`, which this PR is improving by preserving numeric values and generally mapping them towards zero:

| `obj`        | `name`   | `oldValue`        | `newValue`<br>(current) | `newValue`<br>(this PR) | Notes             |
|--------------|----------|-------------------|-------------------------|-------------------------|-------------------|
| _array_      | "length" | `4294967295`      | `42949672952`           | `4294967296`            | fails in both[^1] |
| _array_      | "length" | *                 | `4294967295`            | `oldValue + 1`          | works in both     |
| _typedArray_ | _index_  | `-Infinity`       | `"unlikelyValue"`       | `0`                     | fails → works     |
| _typedArray_ | _index_  | _number_ ≤ 0      | `"unlikelyValue"`       | `oldValue + 1`          | fails → works     |
| _typedArray_ | _index_  | `Infinity`        | `"unlikelyValue"`       | `1`                     | fails → works     |
| _typedArray_ | _index_  | _number_ \> 0     | `"unlikelyValue"`       | `oldValue - 1`          | fails → works     |
| _typedArray_ | _index_  | _bigint_ ≤ 0n     | `"unlikelyValue"`       | `oldValue + 1n`         | fails → works     |
| _typedArray_ | _index_  | _bigint_ \> 0n    | `"unlikelyValue"`       | `oldValue - 1n`         | fails → works     |
| *            | *        | `"unlikelyValue"` | `"unlikelyValue2"`      | `"unlikelyValue2"`      | _(no change)_     |
| *            | *        | *                 | `"unlikelyValue"`       | `"unlikelyValue"`       | _(no change)_     |

I've just pushed a fixup commit with clarifying comments.

[^1]: but setting to any smaller value would be destructive

test/built-ins/TypedArrayConstructors/internals/Set/immutable-buffer-key-is-canonical-non-strict.js
```diff
@@ -0,0 +1,94 @@
...
+    toString() {
+      calls.push(label + " toString");
+      return value;
+    },
+    valueOf() {
+      calls.push(label + " valueOf");
+      return value;
+    }
+  };
+}
```
ptomato, line 32: optional: You could use `TemporalHelpers.toPrimitiveObserver` for this. Although, we really need to move that into its own helper file. (Same comment for files in the same commit)

test/built-ins/TypedArrayConstructors/internals/Set/immutable-buffer-key-is-canonical-non-strict.js
```diff
@@ -0,0 +1,94 @@
...
+
+function unloggedHasOwn(obj, key, calls) {
+  var length = calls.length;
+  var result = obj.hasOwnProperty(key);
+  calls.length = length;
+  return result;
+}
```
ptomato, line 39: If I'm reading correctly, this won't log anything, regardless? In that case it seems better to just use `hasOwnProperty` and only add this in the future if it's necessary. (Same comment for files in the same commit)

ptomato:
The linter failure seems to be a side effect of my recent fix of harness file `defines`. I believe rebasing on https://github.com/tc39/test262/pull/5094 should fix it.

ptomato:
(Removing `compareArray.js` from the frontmatter in these tests would also fix it, but that's another pass over an already long PR so it's fine if you don't want to do that right now.)

test/built-ins/TypedArrayConstructors/internals/DefineOwnProperty/immutable-buffer.js
```diff
@@ -0,0 +1,112 @@
...
+          if (v) newDesc.value = descriptor.value;
+          if (c) newDesc.configurable = descriptor.configurable;
+          if (e) newDesc.enumerable = descriptor.enumerable;
+          if (w) newDesc.writable = descriptor.writable;
+
+          if (typeof Reflect !== "undefined" && Reflect.defineProperty) {
+            assert.sameValue(Reflect.defineProperty(sample, "0", newDesc), true,
+              "Reflect.defineProperty " + reprDesc(newDesc));
+            assert.deepEqual(Object.getOwnPropertyDescriptor(sample, "0"), descriptor,
+              "results of Reflect.defineProperty " + reprDesc(newDesc));
```
ptomato, line 54: optional: I'd like to avoid `assert.deepEqual` in new tests (https://github.com/tc39/test262/issues/3476) but this PR has been waiting long enough so I wouldn't hold things up on that.

If you do feel like fixing it, I _think_ at first glance `verifyProperty(sample, "0", descriptor, { label: ..., restore: true })` ought to work? But I haven 't looked at that too closely. If not, then a custom comparison function seems like the way to go.

harness/propertyHelper.js
```diff
@@ -172,17 +211,32 @@ function isSameValue(a, b) {
...
+    // Array and TypedArray "length" and index properties).
+    switch (typeof oldValue) {
+      case "number":
+        // As a special case, never decrease array length (which would delete
+        // properties not covered by the new length).
+        if (oldValue <= 0 || __isArray(obj) && name === "length") {
+          newValue = oldValue === -Infinity ? 0 : oldValue + 1;
+        } else {
+          newValue = oldValue === Infinity ? 1 : oldValue - 1;
+        }
```
ptomato, line 229: I think this produces a false positive for readonly NaN properties, because `newValue` is also NaN, whereas the old code would've written `"unlikelyValue"` and correctly determined that the write failed. Verified with a spot check:
```js
var obj = {};
Object.defineProperty(obj, "a", { value: NaN, writable: false });
assert(verifyProperty(obj, 'a', { value: NaN, writable: false }));
```

ptomato (approved):
I've convinced myself that the new tests cover the remaining unchecked items in https://github.com/tc39/test262/issues/4509, so this works for me. Feel free to merge when the comments are addressed, as far as I'm concerned.

I only skimmed the files where the TypedArray arg factories were edited.

It occurred to me that it'd be nice to add harness tests for the bugs you fixed in `verifyProperty`, but I'd say that's optional too.

PR #4553, opened 2025-08-01
[immutable-arraybuffer] ArrayBuffer.prototype.sliceToImmutable

gibson042:
Ref #4509

> - [x] `ArrayBuffer.prototype.sliceToImmutable`
>   - [x] `verifyPrimordialCallableProperty(ArrayBuffer.prototype, "sliceToImmutable", "sliceToImmutable", 2);`
>   - [x] brand-checking (throws TypeError unless receiver is an ArrayBuffer and not a SharedArrayBuffer), before resolving bounds
>   - [x] throws TypeError if receiver is detached, before resolving bounds
>   - [x] resolves bounds after passing the preceding checks—first _start_, clamping negative values to [0, _len_ - 1] and positive values to [0, _len_], then _end_, defaulting to _len_ and clamping negative values to [0, _len_ - 1] and positive values to [0, _len_] (and capturing _len_ before coercing either _start_ or _end_)
>   - [x] throws TypeError if receiver was detached by bounds resolution, after bounds resolution
>   - [x] throws RangeError if length < specified upper bound, after passing the second detachment check and even if length was ≥ specified upper bound until bounds resolution
>   - [x] returns an immutable ArrayBuffer with the correct length and contents (both unaffected by subsequent changes to the original receiver, including detachment)

### DIFF
--- a/test/built-ins/ArrayBuffer/prototype/sliceToImmutable/argument-coercion.js
+++ b/test/built-ins/ArrayBuffer/prototype/sliceToImmutable/argument-coercion.js
@@ -1,0 +1,363 @@
+// Copyright (C) 2025 Richard Gibson. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-arraybuffer.prototype.slicetoimmutable
+description: >
+  Requested start and end are coerced to integers and checked for validity, then
+  (if valid) an immutable ArrayBuffer with the correct length and contents is
+  returned
+info: |
+  ArrayBuffer.prototype.sliceToImmutable ( start, end )
+  ...
+  5. Let len be O.[[ArrayBufferByteLength]].
+  6. Let bounds be ? ResolveBounds(len, start, end).
+  7. Let first be bounds.[[From]].
+  8. Let final be bounds.[[To]].
+  9. Let newLen be max(final - first, 0).
+  ...
+  12. Let fromBuf be O.[[ArrayBufferData]].
+  13. Let currentLen be O.[[ArrayBufferByteLength]].
+  14. If currentLen < final, throw a RangeError exception.
+  15. Let newBuffer be ? AllocateImmutableArrayBuffer(%ArrayBuffer%, newLen, fromBuf, first, newLen).
+  16. Return newBuffer.
+
+  ResolveBounds ( len, start, end )
+  1. Let relativeStart be ? ToIntegerOrInfinity(start).
+  2. If relativeStart = -∞, let from be 0.
+  3. Else if relativeStart < 0, let from be max(len + relativeStart, 0).
+  4. Else, let from be min(relativeStart, len).
+  5. If end is undefined, let relativeEnd be len; else let relativeEnd be ? ToIntegerOrInfinity(end).
+  6. If relativeEnd = -∞, let to be 0.
+  7. Else if relativeEnd < 0, let to be max(len + relativeEnd, 0).
+  8. Else, let to be min(relativeEnd, len).
+  9. Return the Record { [[From]]: from, [[To]]: to }.
+
+  ToIntegerOrInfinity ( argument )
+  1. Let number be ? ToNumber(argument).
+  2. If number is one of NaN, +0𝔽, or -0𝔽, return 0.
+  3. If number is +∞𝔽, return +∞.
+  4. If number is -∞𝔽, return -∞.
+  5. Return truncate(ℝ(number)).
+features: [immutable-arraybuffer]
+includes: [compareArray.js]
+---*/
+
+var ab = new ArrayBuffer(8);
+assert.sameValue(ab.sliceToImmutable().byteLength, 8,
+  "Must default to receiver byteLength.");
+ab = new ArrayBuffer(8);
+assert.sameValue(ab.sliceToImmutable(undefined, undefined).byteLength, 8,
+  "Must default (undefined, undefined) to receiver byteLength.");
+
+function repr(value) {
+  if (typeof value === "string") return JSON.stringify(value);
+  if (typeof value === "bigint") return String(value) + "n";
+  if (!value && 1/value === -Infinity) return "-0";
+  return String(value);
+}
+
+function make32ByteArrayBuffer() {
+  var ab = new ArrayBuffer(32);
+  var view = new Uint8Array(ab);
+  for (var i = 0; i < 8; i++) view[i] = i + 1;
+  return ab;
+}
+
+var goodInputs = [
+  // Unmodified non-negative integral numbers
+  [0, 0],
+  [1, 1],
+  [10, 10],
+  // Truncated non-negative integral numbers
+  [0.9, 0],
+  [1.9, 1],
+  [-0.9, 0],
+  // Negative integral numbers
+  [-1, -1],
+  [-1.9, -1],
+  [-2.9, -2],
+  // Coerced to integral numbers
+  [-0, 0],
+  [null, 0],
+  [false, 0],
+  [true, 1],
+  ["", 0],
+  ["8", 8],
+  ["+9", 9],
+  ["-9", -9],
+  ["10e0", 10],
+  ["+1.1E+1", 11],
+  ["+.12e2", 12],
+  ["130e-1", 13],
+  ["0b1110", 14],
+  ["0XF", 15],
+  ["0xf", 15],
+  ["0o20", 16],
+  // Coerced to NaN and mapped to 0
+  [NaN, 0],
+  ["7up", 0],
+  ["1_0", 0],
+  ["0x00_ff", 0],
+  // Clamped
+  [-32, 0],
+  ["-32", 0],
+  [-Infinity, 0],
+  ["-Infinity", 0],
+  [33, 32],
+  ["33", 32],
+  [9007199254740992, 32], // Math.pow(2, 53) = 9007199254740992
+  ["9007199254740992", 32], // Math.pow(2, 53) = 9007199254740992
+  [Infinity, 32],
+  ["Infinity", 32]
+];
+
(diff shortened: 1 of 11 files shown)


test/built-ins/ArrayBuffer/prototype/sliceToImmutable/this-is-sharedarraybuffer.js
```diff
@@ -0,0 +1,38 @@
...
+  ArrayBuffer.prototype.sliceToImmutable ( start, end )
+  1. Let O be the this value.
+  2. Perform ? RequireInternalSlot(O, [[ArrayBufferData]]).
+  3. If IsSharedArrayBuffer(O) is true, throw a TypeError exception.
+features: [SharedArrayBuffer, ArrayBuffer, immutable-arraybuffer]
+includes: [compareArray.js]
+---*/
+
+var fn = ArrayBuffer.prototype.sliceToImmutable;
+assert.sameValue(typeof fn, "function", "Method must exist.");
```
ptomato, line 17:
```suggestion
```
I think this is sufficiently implied by the `immutable-arraybuffer` feature flag.

test/built-ins/ArrayBuffer/prototype/sliceToImmutable/this-is-not-object.js
```diff
@@ -0,0 +1,53 @@
...
+info: |
+  ArrayBuffer.prototype.sliceToImmutable ( start, end )
+  1. Let O be the this value.
+  2. Perform ? RequireInternalSlot(O, [[ArrayBufferData]]).
+features: [ArrayBuffer, immutable-arraybuffer]
+includes: [compareArray.js]
+---*/
+
+var fn = ArrayBuffer.prototype.sliceToImmutable;
+assert.sameValue(typeof fn, "function", "Method must exist.");
```
ptomato, line 16: Ditto

test/built-ins/ArrayBuffer/prototype/sliceToImmutable/this-is-not-object.js
```diff
@@ -0,0 +1,53 @@
...
+  ["true", true],
+  ["false", false],
+  typeof Symbol === "undefined" ? undefined : ["unique symbol", Symbol("1")],
+  typeof Symbol === "undefined" || !Symbol.for ? undefined : ["registered symbol", Symbol.for("1")],
+  typeof BigInt === "undefined" ? undefined : ["bigint", BigInt(1)]
+];
+
+for (var i = 0; i < badReceivers.length; i++) {
+  var label = badReceivers[i][0];
+  var value = badReceivers[i][1];
```
ptomato, line 46: This will break on implementations without Symbol or BigInt, I think you probably want `if (!badReceivers[i]) continue;`

test/built-ins/ArrayBuffer/prototype/sliceToImmutable/this-has-no-arraybufferdata-internal.js
```diff
@@ -0,0 +1,66 @@
...
+info: |
+  ArrayBuffer.prototype.sliceToImmutable ( start, end )
+  1. Let O be the this value.
+  2. Perform ? RequireInternalSlot(O, [[ArrayBufferData]]).
+features: [DataView, Int8Array, ArrayBuffer, immutable-arraybuffer]
+includes: [compareArray.js]
+---*/
+
+var fn = ArrayBuffer.prototype.sliceToImmutable;
+assert.sameValue(typeof fn, "function", "Method must exist.");
```
ptomato, line 18: Already implied by feature flag.

test/built-ins/ArrayBuffer/prototype/sliceToImmutable/argument-coercion.js
```diff
@@ -0,0 +1,369 @@
...
+    var intLength = intEnd - intStart;
+    var source = make32ByteArrayBuffer();
+    var expectContents = Array.from(new Uint8Array(source)).slice(rawStart, rawEnd);
+    var dest = source.sliceToImmutable(rawStart, rawEnd);
+    assert.sameValue(dest.byteLength, intLength,
+      "sliceToImmutable(" + repr(rawStart) + ", " + repr(rawEnd) + ")");
+    assert.compareArray(new Uint8Array(dest), expectContents,
+      "sliceToImmutable(" + repr(rawStart) + ", " + repr(rawEnd) + ")");
+    assert.sameValue(dest.immutable, true,
+      "sliceToImmutable(" + repr(rawStart) + ", " + repr(rawEnd) + ")");
```
ptomato: Minor suggestion, add something to these assertion messages to distinguish which assertion is failing (byte length, contents, or immutability) (Same for below)

ptomato (approved):
Minor comments, feel free to merge when resolved

test/built-ins/ArrayBuffer/prototype/sliceToImmutable/modify-source-after-return.js
```diff
@@ -0,0 +1,37 @@
...
+  16. Return newBuffer.
+features: [immutable-arraybuffer]
+includes: [compareArray.js, detachArrayBuffer.js]
+---*/
+
+var source = new ArrayBuffer(8, { maxByteLength: 8 });
+var view = new Uint8Array(source);
+for (var i = 0; i < 8; i++) view[i] = i + 1;
+
+var dest = new ArrayBuffer.sliceToImmutable();
```
iamstolis: `new ArrayBuffer.sliceToImmutable()` makes no sense. I guess it should be something like `source.sliceToImmutable()`.

test/built-ins/ArrayBuffer/prototype/sliceToImmutable/argument-coercion.js
```diff
@@ -0,0 +1,369 @@
...
+  "Must default (undefined, undefined) to receiver byteLength.");
+
+function repr(value) {
+  if (typeof value === "string") return JSON.stringify(value);
+  if (typeof value === "bigint") return String(value) + "n";
+  if (!value && 1/value === -Infinity) return "-0";
+  return String(value);
+}
+
+var make32ByteArrayBuffer() {
```
iamstolis: Typo? `var make32ByteArrayBuffer()` -> `function make32ByteArrayBuffer()`

test/built-ins/ArrayBuffer/prototype/sliceToImmutable/argument-coercion.js
```diff
@@ -0,0 +1,369 @@
...
+  ["Infinity", 32]
+];
+
+for (var i = 0; i < goodInputs.length; i++) {
+  var rawStart = goodInputs[i][0];
+  var intStart = goodInputs[i][1];
+  for (var j = 0; j < goodInputs.length; j++) {
+    var rawEnd = goodInputs[j][0];
+    var intEnd = goodInputs[j][1];
+    var intLength = intEnd - intStart;
```
iamstolis: This is not a correct calculation of the length of the slice. It does not work when `intStart` or `intEnd` is negative or when `intStart > intEnd`. Note that there is the same problem on lines 151 and 170.

test/built-ins/ArrayBuffer/prototype/sliceToImmutable/argument-coercion.js
```diff
@@ -0,0 +1,369 @@
...
+      "sliceToImmutable" + reprArgs("[Symbol.toPrimitive]", "[Symbol.toPrimitive]"));
+    assert.compareArray(new Uint8Array(dest), expectContents,
+      "sliceToImmutable" + reprArgs("[Symbol.toPrimitive]", "[Symbol.toPrimitive]"));
+    assert.sameValue(dest.immutable, true,
+      "sliceToImmutable" + reprArgs("[Symbol.toPrimitive]", "[Symbol.toPrimitive]"));
+    assert.compareArray(
+      calls,
+      ["start[Symbol.toPrimitive](number)", "end[Symbol.toPrimitive](number)"],
+      "sliceToImmutable" + reprArgs("[Symbol.toPrimitive]", "[Symbol.toPrimitive]"));
+    );
```
iamstolis: SyntaxError, duplicate `);` (here and on the previous line). There is the same problem on lines 316 and 324.

test/built-ins/ArrayBuffer/prototype/sliceToImmutable/argument-coercion.js
```diff
@@ -0,0 +1,369 @@
...
+  }
+}
+
+var badInputs = [
+  // Out of range numbers
+  [-1, RangeError],
+  [9007199254740992, RangeError], // Math.pow(2, 53) = 9007199254740992
+  [Infinity, RangeError],
+  [-Infinity, RangeError],
```
iamstolis, line 290: I don't see why would these inputs caused `RangeError`s.

iamstolis: I don't see why would these out of range numbers caused `RangeError`s.

test/built-ins/ArrayBuffer/prototype/sliceToImmutable/argument-coercion.js
iamstolis: This comment is marked as resolved, but it seems that just the negative `intStart` and `intEnd` are fixed. The current version is using
```
var intLength =
      (intEnd >= 0 ? intEnd : 32 + intEnd) - (intStart >= 0 ? intStart : 32 + intStart);
```
which is still incorrect when `intStart > intEnd`. `intLength` should be maximum of 0 and the current value, I guess.

gibson042: Ah shoot, you're right. Followup PR: #5049.

PR #4554, opened 2025-08-01
[immutable-arraybuffer] Atomics

gibson042:
Ref #4509

> # Atomics
> 
> - [x] `Atomics.{add,and,compareExchange,exchange,or,store,sub,xor}`
>   - [x] throws a TypeError if _typedArray_ is an immutable ArrayBuffer, before coercing _index_
> 
> - [x] `Atomics.load`
>   - [x] works as expected if _typedArray_ is an immutable ArrayBuffer
> 
> - [x] `Atomics.notify`
>   - [x] works as expected if _typedArray_ is an immutable ArrayBuffer, returning 0 when _index_ is in range and coercing _count_ is successful


test/built-ins/Atomics/exchange/nonshared-int-views.js
ptomato, line 1: What is actually the difference between this file and `non-shared-int-views-throws.js` in what's being tested? They seem identical except for the value being exchanged. Could one of them just be deleted?

test/built-ins/Atomics/and/non-shared-int-views-throws.js
```diff
@@ -15,4 +15,4 @@ testWithNonAtomicsFriendlyTypedArrayConstructors((TA, makeCtorArg) => {
   assert.throws(TypeError, function() {
     Atomics.and(view, 0, 1);
   }, `Atomics.and(new ${TA.name}(buffer), 0, 1) throws TypeError`);
-}, ["arraybuffer"], ["immutable"]);
```
ptomato, line 18: If I'm reading this correctly, "immutable" is being removed from the exclude list, so this test will now run on immutable buffers as well. But all new coverage in this file is also already covered in `and/immutable-buffer.js`, because that runs on all typed array constructors. I don't know if that's actually a problem; probably not.

ptomato (approved):
All comments/questions are nonblocking.

test/built-ins/Atomics/exchange/nonshared-int-views.js
gibson042: Excellent observation! Done.

test/built-ins/Atomics/and/non-shared-int-views-throws.js
gibson042: I think there probably is room for such consolidation, but out of scope here, and the focus is sufficiently different that I would want to see it in a dedicated sweep.

PR #4557, opened 2025-08-15
Temporal: Add tests for observable order of options get and cast

ptomato:
These tests cover the normative change approved in the July 2025 TC39 plenary. See https://github.com/tc39/proposal-temporal/pull/3130. All properties of user-passed options objects should be read and cast before any "algorithmic validation" is done.

This is a lot of tests, that cover every entry point where an options object is passed in, where subsequently an exception can be thrown or user code can be called.

However, the normative change only affects 10 of these tests:

- Instant.p.since,until,toString
- PlainDate.p.since,until
- PlainTime.p.since,until
- PlainYearMonth.p.since,until
- ZonedDateTime.p.toString

The other 35 tests simply close a gap in coverage.

### DIFF
--- a/test/built-ins/Temporal/PlainDateTime/prototype/toString/options-read-before-algorithmic-validation.js
+++ b/test/built-ins/Temporal/PlainDateTime/prototype/toString/options-read-before-algorithmic-validation.js
@@ -1,0 +1,40 @@
+// Copyright (C) 2025 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.plaindatetime.prototype.tostring
+description: >
+  All options properties are read and cast before any algorithmic validation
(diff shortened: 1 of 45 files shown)


test/built-ins/Temporal/PlainDateTime/prototype/toString/options-read-before-algorithmic-validation.js
```diff
@@ -0,0 +1,40 @@
...
+  smallestUnit: "month",
+  fractionalSecondDigits: "auto",
+  roundingMode: "expand",
+}, "options");
+
+const instance = new Temporal.PlainDateTime(2025, 8, 14, 12);
+
+assert.throws(RangeError, function () {
+  instance.toString(options);
+}, "exception thrown when unit is a date unit");
```
Ms2ger:
```suggestion
}, "exception thrown when smallestUnit is a date unit");
```

Ms2ger (approved):

PR #4560, opened 2025-08-19
Add 8 numberingSystems to test for Unicode 16

FrankYFTang:
see also https://github.com/tc39/ecma402/pull/929


harness/testIntl.js
```diff
@@ -2119,11 +2119,13 @@ function allNumberingSystems() {
     "ethi",
     "finance",
     "fullwide",
+    "gara",
     "geor",
     "gong",
     "gonm",
     "grek",
     "greklow",
+    "gukh",
```
anba: Nit: Below "gujr" to keep the list sorted.

FrankYFTang:
PTAL

FrankYFTang:
PTAL

FrankYFTang:
> Approved if Anba's comment is addressed.
> 
> We talked in the maintainers meeting and think the right time frame for merging this is 1 month after [CLDR 48](https://cldr.unicode.org/downloads/cldr-48) is released.

@ptomato  These numbering systems were added into CLDR 46 not CLDR 48. CLDR 46 was released in Oct 2024 (one year ago) already

https://cldr.unicode.org/downloads/cldr-46#dtd-changes

What CLDR48 added are another two other ( "chis" and "tols") which are not in the scope of THIS test. ( see https://github.com/unicode-org/cldr/pull/4726/files ) 

CLDR 47 didn't add new numbering Systems. 
CLDR 46 (released in Oct 2024 - more than one year ago) added these 8 ( see  https://github.com/unicode-org/cldr/pull/3658 )

ptomato:
OK, got it. Then no need to wait.

PR #4561, opened 2025-08-21
Further updates for Intl Era Monthcode

ptomato:
Follow up to #4558. I don't think this is exhaustive yet, but a draft polyfill implementation passes it.


ptomato:
Cc @ben-allen @Manishearth

Manishearth (approved):

PR #4565, opened 2025-08-26
CI: run harness tests

### LINKED ISSUE #4497: Harness tests should be checked in CI
#4468 broke test/harness/verifyProperty-value-error.js (see #4493), but CI was all green. We should fix that.

Ms2ger:
Fixes #4497.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -91,3 +91,16 @@ jobs:
             New changes were found after re-generating the tests.
             Please read the documentation on procedurally generated tests
             https://github.com/tc39/test262/blob/HEAD/CONTRIBUTING.md#procedurally-generated-tests
+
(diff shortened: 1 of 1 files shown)


.github/workflows/checks.yml
```diff
@@ -91,3 +91,16 @@ jobs:
...
+
+  self-test:
+    name: Run the harness tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install test262-harness
+        run: npm install
```
ljharb, line 103: is node already available?

Ms2ger: Seems like it!

ptomato (approved):

PR #2062, opened 2019-02-02
Add tests for %TypedArray%.prototype.sort stability

mathiasbynens:
https://github.com/tc39/ecma262/pull/1433

PR #4569, opened 2025-09-01
Classify tests for the stability of array sorting

jugglinmike:
These tests were originally introduced via https://github.com/tc39/test262/pull/1977 & https://github.com/tc39/test262/pull/2062 and subsequently modified via https://github.com/tc39/test262/pull/2169.

### DIFF
--- a/features.txt
+++ b/features.txt
@@ -214,6 +214,7 @@ rest-parameters
 Set
 set-methods
 SharedArrayBuffer
+stable-array-sort  # https://github.com/tc39/ecma262/pull/1340 & https://github.com/tc39/ecma262/pull/1433
 string-trimming
 String.fromCodePoint
 String.prototype.at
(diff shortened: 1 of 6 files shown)


jugglinmike:
Because the committee reached consensus for each method separately, it might be more appropriate to track this with two distinct feature flags instead of one. I've made that change in a dedicated commit that may be squashed or discarded.

jugglinmike:
Nope, the webfeatures classification won't impose any change in Test262's existing practices. The absence of a feature flag here just seemed like a good opportunity for synergy between the two projects: Test262 occasionally does use feature flags for normative PRs, and *these* normative PRs happen to have a corresponding web-feature.

We could certainly classify that web feature without the new frontmatter flags I've proposed here, but doing so would involve hard-coding paths in the [nascent](https://github.com/tc39/test262/issues/4571) classification file. Feature flags generally seem like a more resilient mechanism.

PR #4572, opened 2025-09-10
Add tests for Unicode 17 property escapes

mathiasbynens:
Issue: https://github.com/tc39/ecma262/issues/3688


mathiasbynens:
The test failures are expected, at least until engines update their Unicode data to v17.

ptomato:
We talked in the maintainers meeting and think the right time frame for merging this is 1 month after [CLDR 48](https://cldr.unicode.org/downloads/cldr-48) is released.

linusg:
They didn't do a timely update of the [release notes](https://cldr.unicode.org/downloads/cldr-48) which still say _beta draft_ but the final version of CLDR 48 was released on October 23: https://github.com/unicode-org/cldr/releases/tag/release-48

ptomato:
The failures are all timeouts (no output for 10 minutes)

PR #4573, opened 2025-09-10
Add tests for Unicode 17 identifiers and unicodeSets

mathiasbynens:
Issue: https://github.com/tc39/ecma262/issues/3688


mathiasbynens:
The test failures are expected, at least until engines update their Unicode data to v17.

ptomato:
We talked in the maintainers meeting and think the right time frame for merging this is 1 month after [CLDR 48](https://cldr.unicode.org/downloads/cldr-48) is released.

linusg:
They didn't do a timely update of the [release notes](https://cldr.unicode.org/downloads/cldr-48) which still say _beta draft_ but the final version of CLDR 48 was released on October 23: https://github.com/unicode-org/cldr/releases/tag/release-48

linusg:
A month has passed so this and #4572 can be merged.

Additionally I'd like to point out that there's a circular dependency here - engines might not want to update their version of CLDR before relevant tests have landed, e.g. https://github.com/quickjs-ng/quickjs/issues/1168. Perhaps it makes sense to treat this similar to stage 2.7 and merge tests *before* asking implementations to update in the future.

mathiasbynens:
> Additionally I'd like to point out that there's a circular dependency here - engines might not want to update their version of CLDR before relevant tests have landed, e.g. quickjs-ng/quickjs#1168. Perhaps it makes sense to treat this similar to stage 2.7 and merge tests before asking implementations to update in the future.

+1 to this. In the past, we landed these updated tests as soon as they became available, and IMHO we should get back to that.

ptomato:
I'm assuming the ESMeta test failures are because of ESMeta not having updated Unicode data.

PR #4574, opened 2025-09-10
Remove unnecessary call to compareArray in this-value-boolean.js

trevorade:
This assertion throws before `compareArray` is called. In addition, `compareArray` takes two parameters, not one.

It's unclear to me why it was added in the first place.


ptomato (approved):
Thanks. Probably a search-and-replace error.

trevorade:
> Thanks. Probably a search-and-replace error.

Thx for the fast approval! Looks like it's ready to merge.

PR #4563, opened 2025-08-22
[nonextensible-applies-to-private] Adding private symbol throws

erights:
Following
https://github.com/tc39/proposal-nonextensible-applies-to-private adding private symbols to non-nonextensible objects throws a TypeError.

PR #4577, opened 2025-09-10
[nonextensible-applies-to-private] Adding private field to non-extensible object throws

erights:
Alternative to https://github.com/tc39/test262/pull/4563 

To be the test262 tests for the https://github.com/tc39/proposal-nonextensible-applies-to-private 

See https://github.com/tc39/proposal-nonextensible-applies-to-private/issues/5

Initial PR content by @o-


ptomato (commented):
Looks correct so far.

For completeness, would we need to test stamping a private accessor and private method?

(I would guess that the case of stamping a private static method, or private static field cannot occur, since those go on the constructor, which cannot be frozen prior to the class declaration.)

Finally, this needs a feature flag for the proposal to be added in [`features.txt`](https://github.com/tc39/test262/blob/main/features.txt) and in the frontmatter of the two modified tests.

erights:
> Looks correct so far.
> 
> For completeness, would we need to test stamping a private accessor and private method?

For inheritance, done.

For a base class, AFAICT, impossible. There are no observable states where the class, instance, or prototype are observable before methods are accessors are added.

> (I would guess that the case of stamping a private static method, or private static field cannot occur, since those go on the constructor, which cannot be frozen prior to the class declaration.)

For a private static method, AFAICT, it is indeed impossible.

For a private static field, it can be done in the initializer expression. Done.

> 
> Finally, this needs a feature flag for the proposal to be added in [`features.txt`](https://github.com/tc39/test262/blob/main/features.txt) and in the frontmatter of the two modified tests.

Done.

(with much help from @gibson042 )

PTAL

ptomato:
I left this open for a few days in case any other maintainers had any feedback, but now that it's been presented in the TC39 meeting I think it's time to merge it. I'll do that now.

test/language/statements/class/elements/private-class-field-on-nonextensible-objects.js
```diff
@@ -0,0 +1,117 @@
...
+  };
+  // public accessors are on the prototype, so are ok.
+  get publicAccessor() {
+    return this.#privateAccessor;
+  }
+}
+
+const a = new ClassWithPrivateAccessor(false);
+// extensible objects can be extended
+assert.sameValue(m.publicAccessor, 42);
```
iamstolis, line 95: This is a copy/paste bug most likely. It should be `a.publicAccessor` (not `m.publicAccessor`), I guess.

test/language/statements/class/subclass/private-class-field-on-nonextensible-return-override.js
```diff
@@ -0,0 +1,100 @@
...
+    this.#val = 42;
+  }
+  val() {
+    return this.#val;
+  }
+}
+
+const t = new ClassWithPrivateField({});
+// extensible objects can be extended
+assert.sameValue(t.val(), 42);
```
iamstolis, line 45: This looks incorrect. You expect `t` to have `val` method, but `t` is `{}` i.e. does not have methods from `ClassWithPrivateField.prototype`.

test/language/statements/class/subclass/private-class-field-on-nonextensible-return-override.js
```diff
@@ -0,0 +1,100 @@
...
+  };
+  // public methods are on the prototype, so are ok.
+  publicMethod() {
+    return this.#privateMethod();
+  }
+}
+
+const m = new ClassWithPrivateMethod({});
+// extensible objects can be extended
+assert.sameValue(m.publicMethod(), 42);
```
iamstolis, line 70: There is no `publicMethod` on `m`: `m` is `{}` while `publicMethod` is on `ClassWithPrivateMethod.prototype`.

test/language/statements/class/subclass/private-class-field-on-nonextensible-return-override.js
```diff
@@ -0,0 +1,100 @@
...
+  };
+  // public accessors are on the prototype, so are ok.
+  get publicAccessor() {
+    return this.#privateAccessor;
+  }
+}
+
+const a = new ClassWithPrivateAccessor({});
+// extensible objects can be extended
+assert.sameValue(m.publicAccessor, 42);
```
iamstolis, line 95: Copy/paste bug (`m` should be `a`) + the same problem as above (no `publicAccessor` on `a`).

iamstolis: I have created #4587 to track these bugs.

test/language/statements/class/elements/private-class-field-on-nonextensible-objects.js
erights: In progress at https://github.com/tc39/test262/pull/4590

test/language/statements/class/subclass/private-class-field-on-nonextensible-return-override.js
erights: In progress at https://github.com/tc39/test262/pull/4590

test/language/statements/class/subclass/private-class-field-on-nonextensible-return-override.js
erights: In progress at https://github.com/tc39/test262/pull/4590

test/language/statements/class/subclass/private-class-field-on-nonextensible-return-override.js
erights: In progress at https://github.com/tc39/test262/pull/4590

PR #4578, opened 2025-09-14
Check that callbackfn is not called if the key cannot be held weakly

johauke:
There is no validation to check that the callback function is not called if the key cannot be held weakly.
This pull request modifies the throw-if-key-cannot-be-held-weakly test to also check for this.

Discovered here:
https://bugzilla.mozilla.org/show_bug.cgi?id=1988369


Ms2ger (approved):
Thanks!

Issue #4581, opened 2025-09-15
Follow up from May 2025 iterator sequencing update


ptomato:
Per @iamstolis in https://github.com/tc39/test262/pull/4523#issuecomment-3244199586

> Shouldn't [built-ins/Iterator/concat/next-method-returns-throwing-value.js](https://github.com/tc39/test262/blob/main/test/built-ins/Iterator/concat/next-method-returns-throwing-value.js) be updated as well? [iterator.next()](https://github.com/tc39/test262/blob/main/test/built-ins/Iterator/concat/next-method-returns-throwing-value.js#L49) did not throw before because iteratorResult was reused (i.e. its value was not read). Now, IteratorStepValue() always reads the value of the inner iterator (which throws in this case). So, iterator.next() should throw now, right?
> 
> In other words, I think that https://github.com/tc39/test262/commit/71794b138844536a36b341700f88ee91c5754431 should be reverted now.

ptomato:
cc @michaelficarra - can you please check if this matches your understanding?

michaelficarra:
Yes, this sounds correct. It looks like I missed reverting that commit.

PR #4583, opened 2025-09-16
Revert "fix another test now that we don't access the IteratorResult …

ptomato:
…value"

This reverts commit 71794b138844536a36b341700f88ee91c5754431.

Follow up from #4523, missed reverting this commit.

Closes: #4581


ptomato:
cc @michaelficarra @iamstolis

michaelficarra (approved):

PR #4586, opened 2025-09-25
String.prototype.toLocale{Lower,Upper}Case validates all locale identifiers

anba:
Both tests pass in JSC and SM, but fail in V8.


anba:
V8 also doesn't implement [TransformCase](https://tc39.es/ecma402/#sec-transform-case), step 3 correctly when no argument is passed to `toLocale{Lower,Upper}Case`. But I don't think this case can be tested through test262, because it requires changing the default locale. For example when running d8  with `LANG=tr`:
```js
d8> "i".toLocaleUpperCase().charCodeAt(0).toString(16)
"49"
d8> "i".toLocaleUpperCase([]).charCodeAt(0).toString(16)
"130"
```

ptomato:
I opened https://issues.chromium.org/issues/447346221 for the case that can't be tested with test262, so the information doesn't get lost.

PR #4592, opened 2025-10-01
Move compareArray into assert.js

gibson042:
I think it's sufficiently fundamental to justify omnipresence. This also allows use by _other_ helpers without having to solve the problem of a non-shallow `includes` graph.


harness/compareArray.js
```diff
@@ -2,49 +2,6 @@
 // This code is governed by the BSD license found in the LICENSE file.
 /*---
 description: |
-    Compare the contents of two arrays
-defines: [compareArray]
+    Deprecated now that compareArray is defined in assert.js.
```
ljharb, line 5: why not just delete the entire file?

harness/assert.js
```diff
@@ -101,6 +101,42 @@ assert.throws = function (expectedErrorConstructor, func, message) {
...
+
+  for (var i = 0; i < a.length; i++) {
+    if (!assert._isSameValue(b[i], a[i])) {
+      return false;
+    }
+  }
+  return true;
+}
+
+compareArray.format = function(arrayLike) {
```
ljharb:
```suggestion
compareArray.format = function (arrayLike) {
```

harness/assert.js
```diff
@@ -101,6 +101,42 @@ assert.throws = function (expectedErrorConstructor, func, message) {
...
+  return `[${[].map.call(arrayLike, String).join(', ')}]`;
```
ljharb:
```suggestion
  return `[${Array.prototype.join.call(arrayLike, ', ')}]`;
```
since join also calls toString?

harness/assert.js
```diff
@@ -101,6 +101,42 @@ assert.throws = function (expectedErrorConstructor, func, message) {
   throw new Test262Error(message);
 };
 
+assert.compareArray = function(actual, expected, message) {
```
ljharb:
```suggestion
assert.compareArray = function (actual, expected, message) {
```

harness/assert.js
bakkot: `[].map.call([Symbol()], String).join(', ') // 'Symbol()'`
`Array.prototype.join.call([Symbol()], ', ') // throws`

harness/compareArray.js
bakkot: That would break a bunch of my [one-off harnesses](https://github.com/tc39/proposal-joint-iteration/blob/898a420edb8cfd63897ef09a13b9555e04da6823/run-test262.mjs#L15).

Yeah, if you're doing the full suite the right way then you or your tooling will be parsing the header so it can include exactly the files it needs, but that's overkill for a lot of cases.

harness/assert.js
ljharb: ahhh right, because of symbols. fair point.

harness/compareArray.js
ljharb: Should the file be emitting a warning?

gibson042: Saving that for a followup; _lots_ of tests mention this file in their `includes`.

harness/assert.js
gibson042: That would break on symbols.
```js
Array.prototype.join.call(["foo", "bar"], ', ');
// => "foo, bar"
Array.prototype.join.call([Symbol("foo"), Symbol("bar")], ', ');
// => TypeError
```

harness/compareArray.js
bakkot: I had claude write me a script to update the existing tests, if you want to use it; the resulting diff looks fine to me (but I didn't read the script).

```js
#!/usr/bin/env node

const fs = require('fs');
const path = require('path');

function processFile(filePath) {
  const content = fs.readFileSync(filePath, 'utf8');

  // Find the YAML header block
  const headerMatch = content.match(/^([\s\S]*?\/\*---\n)([\s\S]*?)(\n---\*\/[\s\S]*)$/);

  if (!headerMatch) {
    return; // No header found
  }

  const [, before, header, after] = headerMatch;
  let modified = false;
  let lines = header.split('\n');
  let newLines = [];

  for (let i = 0; i < lines.length; i++) {
    const line = lines[i];

    // Check for inline array format: includes: [...]
    const inlineMatch = line.match(/^(\s*)includes:\s*\[([^\]]+)\]$/);
    if (inlineMatch) {
      const [, indent, items] = inlineMatch;
      const itemList = items.split(',').map(s => s.trim()).filter(s => s !== 'compareArray.js');

      if (itemList.length === 0) {
        modified = true;
        // Skip this line entirely
        continue;
      } else if (items.split(',').map(s => s.trim()).includes('compareArray.js')) {
        modified = true;
        newLines.push(`${indent}includes: [${itemList.join(', ')}]`);
        continue;
      }
    }

    // Check for multi-line array start: includes:
    const multiLineMatch = line.match(/^(\s*)includes:\s*$/);
    if (multiLineMatch) {
      const [, indent] = multiLineMatch;
      let arrayLines = [line];
      let j = i + 1;

      // Collect all array items
      while (j < lines.length && lines[j].match(/^\s+- /)) {
        arrayLines.push(lines[j]);
        j++;
      }

      // Filter out compareArray.js
      const itemLines = arrayLines.slice(1).filter(l => !l.includes('compareArray.js'));

      if (itemLines.length === 0) {
        modified = true;
        // Skip all these lines
        i = j - 1;
        continue;
      } else if (itemLines.length !== arrayLines.length - 1) {
        modified = true;
        newLines.push(arrayLines[0]);
        newLines.push(...itemLines);
        i = j - 1;
        continue;
      } else {
        // No changes needed, keep all lines
        newLines.push(...arrayLines);
        i = j - 1;
        continue;
      }
    }

    newLines.push(line);
  }

  if (modified) {
    const newHeader = newLines.join('\n');
    const newContent = before + newHeader + after;
    fs.writeFileSync(filePath, newContent, 'utf8');
    console.log(`Updated: ${filePath}`);
  }
}

function walkDirectory(dir) {
  const entries = fs.readdirSync(dir, { withFileTypes: true });

  for (const entry of entries) {
    const fullPath = path.join(dir, entry.name);

    if (entry.isDirectory()) {
      walkDirectory(fullPath);
    } else if (entry.isFile() && (entry.name.endsWith('.js') || entry.name.endsWith('.case'))) {
      try {
        processFile(fullPath);
      } catch (err) {
        console.error(`Error processing ${fullPath}:`, err.message);
      }
    }
  }
}

walkDirectory('.');
console.log('Done!');
```

harness/assert.js
gibson042: I'd withheld any implementation changes, but this is sufficiently minor that I pushed it and also a tighter validity check (non-primitive rather than non-nullish). But if any of that is too far, I'm happy to defer to a followup.

harness/assert.js
```diff
@@ -101,6 +101,42 @@ assert.throws = function (expectedErrorConstructor, func, message) {
...
+}
+
+assert.compareArray = function (actual, expected, message) {
+  message = message === undefined ? '' : message;
+
+  if (typeof message === 'symbol') {
+    message = message.toString();
+  }
+
+  assert(!isPrimitive(actual), `Actual argument shouldn't be primitive. ${message}`);
```
bakkot: How confident are you that none of the tests are passing strings to `compareArray`? Certainly they _shouldn't_ but it does work fine with the old implementation.

gibson042: I would have said "sufficiently confident to proceed with this", but I just checked and actually found three cases of exactly that (all in staging and now fixed). Good thinking!

PR #4597, opened 2025-10-10
Temporal: Add tests for PlainMonthDay.from with month and invalid monthCode

gibson042:
Prompted by discussion in Matrix.

### DIFF
--- a/test/built-ins/Temporal/PlainMonthDay/from/monthcode-invalid.js
+++ b/test/built-ins/Temporal/PlainMonthDay/from/monthcode-invalid.js
@@ -9,7 +9,9 @@ features: [Temporal]
 
 ["m1", "M1", "m01"].forEach((monthCode) => {
   assert.throws(RangeError, () => Temporal.PlainMonthDay.from({ monthCode, day: 17 }),
-    `monthCode '${monthCode}' is not well-formed`);
+    `monthCode '${monthCode}' is not well-formed (without numeric month)`);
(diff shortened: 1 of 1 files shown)


test/built-ins/Temporal/PlainMonthDay/from/monthcode-invalid.js
```diff
@@ -9,15 +9,23 @@ features: [Temporal]
...
 });
 
 assert.throws(RangeError, () => Temporal.PlainMonthDay.from({ year: 2021, month: 12, monthCode: "M11", day: 17 }),
   "monthCode and month conflict");
 
 ["M00", "M19", "M99", "M13", "M00L", "M05L", "M13L"].forEach((monthCode) => {
   assert.throws(RangeError, () => Temporal.PlainMonthDay.from({ monthCode, day: 17 }),
-    `monthCode '${monthCode}' is not valid for ISO 8601 calendar`);
+    `monthCode '${monthCode}' is not valid for ISO 8601 calendar (without numeric month)`);
+  var plausibleMonth = Number(monthCode.slice(1, 3)) + (monthCode.length - 3);
```
ptomato: I think I see what the intention of this line is, but 0, 19, 99, 13, and 14 are definitely not valid numeric months in the ISO calendar, so it wouldn't catch the case where the numeric month is valid but the month code isn't.

Maybe that's OK as is, or maybe a possible solution would be to clamp `plausibleMonth` between 1 and 12?

Issue #4571, opened 2025-09-09
Adding web-features metadata


jugglinmike:
In gh-4567, we discussed the potential of adding a new layer of test classification information to Test262. After receiving a positive signal there, I'm opening this issue to facilitate a technical design process. I'll start with the design goals that I've identified based on the needs of the initial consumers of this data, and I'll follow that with a concrete proposal which I believe satisfies those goals.

### Design Goals

There are two high-level design goals: generalizabilty and power.

The design should be **generalizable** so that it is recognizable to similar projects that support web-features (e.g. [the web-platform-tests project](https://github.com/web-platform-tests/wpt)--[RCF here](https://github.com/web-platform-tests/rfcs/blob/main/rfcs/web_features.md)). That will allow folks with shared interests (e.g. maintainers of web-status data and the audience of the projects which consume that data) to recognize the pattern across projects. This should promote voluntary contributions, and it will also allow folks external to Test262 to build shared tooling and write common documentation. Practically speaking, this means that (1) the solution should have parity with syntax and semantics established within WPT, and (2) any extensions to that syntax should conceivably find application beyond Test262.

The design should be **powerful** so that the classifications are both correct and maintainable. This will further encourage volunteer contributions and limit the amount of maintenance churn as Test262 evolves and grows. Practically speaking, this means that the design should be (1) adaptable enough to leverage Test262's existing classification system, (2) granular enough to match subsets of directories, and (3) dynamic enough to limit brittleness (e.g. avoiding static lists of filenames).

### Proposal

We'll place classification information in YAML-formatted files named `WEB_FEATURES.yml`. Those files will satisfy the following [CDDL](https://datatracker.ietf.org/doc/html/rfc8610) schema:

```cddl
WebFeatures = {
  features: [*WebFeature],
}

WebFeature = {
  name: text,
  WebFeaturePattern,
}

WebFeaturePattern = (
  WebFeatureFilePattern /
  WebFeatureTagPattern
)

WebFeatureFilePattern = (
  files: [*TextPattern] / "**"
)

WebFeatureTagPattern = (
  tags: [*TextPattern]
)

TextPattern = text .regexp "!?[A-Za-z0-9_.-]+"
```

The `WebFeatureFilePattern` would be used to match the names of files which are present in the same directory of the YAML file. The string value `”**”` would match all files in the current directory and all subdirectories.

[...]

Issue #4567, opened 2025-08-27
Mapping tests to web features


jugglinmike:
Hello! I'm writing today to explore opportunities for making Test262 results more legible to web developers.

Essentially, this would mean adding metadata which maps Test262 tests to "features" as perceived by web developers and as defined in the [web-features](https://github.com/web-platform-dx/web-features) project by the [WebDX Community Group](https://www.w3.org/community/webdx/). That project powers the "Baseline" statistics on [MDN](https://developer.mozilla.org/), [Can I use...](https://caniuse.com/), and [web.dev](https://web.dev); along with the conformance statistics displayed on [webstatus.dev](https://webstatus.dev).

While it's true that Test262 already has machine-readable test-classification mechanism ([the `features` frontmatter](https://github.com/tc39/test262/blob/c185485452edcca9de3e2ebf26a990934c09e194/CONTRIBUTING.md#features)), that mechanism is insufficient for the web-features use case for a couple reasons:

- It's only applied for new features
- It would be both tedious and noisy to modify every file that covers many of these features (e.g. "logical assignment")

The solution (whatever its final form) can and should leverage this existing metadata, but further classification (and perhaps another classification technique) will be necessary.

Of course, the benefits aren't without cost. I'm available to perform the initial classification work, but maintaining metadata in the long term will unavoidably require additional effort. There's evidence that the work is worthwhile since [the Web Platform Tests project](https://web-platform-tests.org/) took up the task [a couple years ago](https://github.com/web-platform-tests/rfcs/pull/163). Test262's audience is wider than the web platform, but [the features defined for ECMAScript](https://webstatus.dev/?q=group%3Ajavascript&sort=name_asc) aren't intrinsically web-centric. [test262.fyi](https://test262.fyi) could take a cue from [wpt.fyi](https://wpt.fyi) and offer these classifiers as a filtering mechanism (e.g. [this view of test results for the CSS pseudo-selector `:dir()`](https://wpt.fyi/results/?label=master&label=experimental&aligned&q=feature%3Adir-pseudo)).

The design space for web-features support is large, and I'll be excited to collaborate on nailing down all the implementation details. At this stage, though, I'm only wondering whether the maintainers fundamentally object to introducing any form of metadata for web-feature classification. Do you folks believe the benefit outweighs the cost?

ljharb:
what are the tests in test262 that would benefit from this categorization?

jugglinmike:
[...]

PR #4599, opened 2025-10-14
Add WEB_FEATURES.yml

jugglinmike:
As proposed in gh-4567 and designed in gh-4571.

This patch includes a rudimentary linting tool for local use. Would you folks support a follow-up patch to introduce a new GitHub Workflow which uses it to validate pull requests?


jugglinmike:
I recognized a few opportunities for improvements to this changeset. Since no one has had an opportunity to review it, yet, I went ahead and made those improvements directly.

ptomato:
We discussed this in the maintainers meeting. Makes sense for @jugglinmike to make changes to these files directly while the workflow is in development.

jugglinmike:
Thanks!!

PR #4600, opened 2025-10-15
Test that resetting [[ModuleAsyncEvaluationCount]] is unobservable

nicolo-ribaudo:
Both V8 and SpiderMonkey were/are affected by a bug in how they reset the global [[ModuleAsyncEvaluationCount]]. This test case ensures that it's not reset too eagerly.


nicolo-ribaudo:
CI results match my expectation:
- V8 passes (it was fixed a few months ago)
- SpiderMonkey fails (expected, I already reported the bug)
- JSC fails (I didn't test it before)

PR #4604, opened 2025-10-22
Fix and tighten up cross realm `%ThrowTypeError%` test.

aardvark179:
The cross realm test for `%ThrowTypeError%` had a bug in that the function created in the other realm was not strict and so didn't have `%ThrowTypeError%` as the `callee` getter. The test passed because a regular getter was distinct from `ThrowTypeError` in the base realm.

I've expanded the test to check two functions on the other realm have identical callee getters, and that this function throws a TypeError.

### DIFF
--- a/test/built-ins/ThrowTypeError/distinct-cross-realm.js
+++ b/test/built-ins/ThrowTypeError/distinct-cross-realm.js
@@ -17,8 +17,15 @@ var localArgs = function() {
   "use strict";
   return arguments;
 }();
-var otherArgs = (new other.Function('return arguments;'))();
+var otherArgs = (new other.Function('"use strict"; return arguments;'))();
+var otherArgs2 = (new other.Function('"use strict"; return arguments;'))();
 var localThrowTypeError = Object.getOwnPropertyDescriptor(localArgs, "callee").get;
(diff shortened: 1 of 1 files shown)


ljharb (approved):

ptomato:
The ESMeta failure is just that the `$262` object is not supported.

linusg (commented):
I don't think this change is correct? The thrown type error is from the current realm, and the current realm is set to the builtin function's `[[Realm]]` when creating the execution context. In other words:

```diff
--- a/test/built-ins/ThrowTypeError/distinct-cross-realm.js
+++ b/test/built-ins/ThrowTypeError/distinct-cross-realm.js
@@ -23,7 +23,7 @@ var localThrowTypeError = Object.getOwnPropertyDescriptor(localArgs, "callee").g
 var otherThrowTypeError = Object.getOwnPropertyDescriptor(otherArgs, "callee").get;
 var otherThrowTypeError2 = Object.getOwnPropertyDescriptor(otherArgs, "callee").get;

-assert.throws(TypeError, function() {
+assert.throws(other.TypeError, function() {
   otherThrowTypeError();
 });
```

PR #4605, opened 2025-10-22
Add test for trailing garbage handling in setFromBase64

anba:
- "setFromBase64/trailing-garbage.js" fails in both JSC and V8
- "setFromBase64/trailing-garbage-empty.js" fails only in JSC
- SM passes both tests.

CC @bakkot


anba:
It's probably an issue with simdutf, given that the test fails in both JSC and V8.

PR #4607, opened 2025-10-29
Add tests for Import Text proposal

eemeli:
Filing initially as a draft, as these tests are for the [Import Text proposal](https://github.com/eemeli/proposal-import-text), which is currently at Stage 0.

### DIFF
--- a/test/built-ins/AsyncDisposableStack/prototype/Symbol.toStringTag.js
+++ b/test/built-ins/AsyncDisposableStack/prototype/Symbol.toStringTag.js
@@ -1,0 +1,23 @@
+// Copyright (C) 2023 Ron Buckton. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-asyncdisposablestack.prototype-@@toStringTag
+description: >
+    `Symbol.toStringTag` property descriptor
+info: |
+    The initial value of the @@toStringTag property is the String value
+    'AsyncDisposableStack'.
+
+    This property has the attributes { [[Writable]]: false, [[Enumerable]]:
+    false, [[Configurable]]: true }.
+includes: [propertyHelper.js]
+features: [explicit-resource-management, Symbol, Symbol.toStringTag]
+---*/
+
+verifyProperty(AsyncDisposableStack.prototype, Symbol.toStringTag, {
+  value: 'AsyncDisposableStack',
+  writable: false,
+  enumerable: false,
+  configurable: true
+});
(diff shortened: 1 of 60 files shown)


eemeli:
The linter failure seems like it's maybe a bug in its exclusion rules?
```
/home/runner/work/test262/test262/test/language/import/import-attributes/text-javascript_FIXTURE.js: LICENSE - Invalid Copyright header
```

Do `_FIXTURE` files really need a copyright header, if they happen to use `.js` as their extension?

ptomato:
> Do `_FIXTURE` files really need a copyright header, if they happen to use `.js` as their extension?

They may or may not have a bunch of code in them, so yes.

eemeli:
The proposal has now advanced to Stage 3.

Ms2ger:
@jugglinmike can you look at the lint failure?

test/language/expressions/dynamic-import/import-attributes/2nd-param-with-type-text.js
```diff
@@ -0,0 +1,14 @@
...
+/*---
+esid: sec-create-text-module
+description: Dynamic import of text files
+flags: [async]
+features: [dynamic-import, import-attributes, import-text]
+---*/
+
+import('./2nd-param_FIXTURE.json', { with: { type: 'text' } })
+  .then((module) => {
+    assert.sameValue(module.default, '262\n');
```
nicolo-ribaudo: I think this should only test that `typeof module.default` is a string, which is what our spec requires.

test/language/import/import-attributes/text-empty.js
```diff
@@ -0,0 +1,20 @@
...
+  It performs the following steps when called:
+
+    1. Return CreateDefaultExportSyntheticModule(source).
+flags: [module]
+features: [import-attributes, import-text]
+---*/
+
+import value from './text-valid_FIXTURE' with { type: 'text' };
+
+assert.sameValue(value, '');
```
nicolo-ribaudo: Also here

test/language/import/import-attributes/text-javascript_FIXTURE.js
```diff
@@ -0,0 +1,4 @@
+// Copyright (C) 2025 Mozilla Foundation. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+throw "This should not be evaluated."
```
nicolo-ribaudo: It would be even better putting a syntax error in here, so that we catch that the file not only is not executed, but it's also not _linked_ as a JS module.

test/language/import/import-attributes/text-empty.js
```diff
@@ -0,0 +1,20 @@
...
+  The abstract operation CreateTextModule takes argument source (a String) and
+  returns a normal completion containing a Synthetic Module Record.
+  It performs the following steps when called:
+
+    1. Return CreateDefaultExportSyntheticModule(source).
+flags: [module]
+features: [import-attributes, import-text]
+---*/
+
+import value from './text-valid_FIXTURE' with { type: 'text' };
```
trflynn89, line 18: Was this meant to be `./text-empty_FIXTURE`?

eemeli: D'oh, yes. Thanks for catching that. I'll file a PR for the fix.

PR #4608, opened 2025-10-29
Add tests for keep-trailing-zeros proposal

eemeli:
Proposal repo: https://github.com/tc39/proposal-intl-keep-trailing-zeros

~The currently proposed spec text won't satisfy all of these tests; I'll be filing a PR incorporating the required changes.~ Edit: This PR matches the proposal's behaviour, once tc39/proposal-intl-keep-trailing-zeros#19 is merged.

I've created a patched fork of the FormatJS Intl.NumberFormat polyfill to run these tests: https://github.com/eemeli/formatjs-for-keep-trailing-zeros


test/intl402/NumberFormat/prototype/format/value-decimal-string-trailing-zeros.js
```diff
@@ -0,0 +1,78 @@
...
+assert.sameValue(nf.format('.10'), '0.10');
+assert.sameValue(nf.format('.0'), '0.0');
+assert.sameValue(nf.format('00.0'), '0.0');
+assert.sameValue(nf.format('001'), '1');
+assert.sameValue(nf.format('-0.00'), '-0.00');
+assert.sameValue(nf.format('-.00'), '-0.00');
+assert.sameValue(nf.format('3.100'), '3.100');
+assert.sameValue(nf.format('6.000000'), '6.000000');
+assert.sameValue(nf.format('1.230e1'), '12.30');
+assert.sameValue(nf.format('1.230e-2'), '0.0123');
```
gibson042: This assertion seems to be incorrect; wouldn't preservation of trailing zeros look like this?
```suggestion
assert.sameValue(nf.format('1.230e-2'), '0.01230');
```

test/intl402/NumberFormat/prototype/format/value-decimal-string-trailing-zeros.js
```diff
@@ -0,0 +1,78 @@
...
+
+// Default maximumfractionDigits is 3
```
gibson042:
```suggestion
// maximumFractionDigits defaults to 3
```

test/intl402/NumberFormat/prototype/format/value-decimal-string-trailing-zeros.js
eemeli: Yes. I'll also add a couple more tests to ensure this works as intended.

gibson042 (commented):
LGTM after filling in pending front matter.

test/intl402/NumberFormat/prototype/format/value-decimal-string-trailing-zeros.js
```diff
@@ -0,0 +1,102 @@
...
+assert.sameValue(nf.format('3.100'), '3.100');
+assert.sameValue(nf.format('6.000000'), '6.000000');
+assert.sameValue(nf.format('1.230e1'), '12.30');
+assert.sameValue(nf.format('1.230e0'), '1.230');
+assert.sameValue(nf.format('1.230e-1'), '0.1230');
+assert.sameValue(nf.format('1.230e-2'), '0.01230');
+assert.sameValue(nf.format('12.30e-1'), '1.230');
+assert.sameValue(nf.format('12.30e-2'), '0.1230');
+assert.sameValue(nf.format('12.30e-3'), '0.01230');
+assert.sameValue(nf.format('1.2345e-1000'), '0.0000');
```
gibson042: https://github.com/tc39/proposal-intl-keep-trailing-zeros/issues/15
```suggestion
assert.sameValue(nf.format('1.2345e-1000'), '0.00000000000000000000');
```
(1.2345e-1000 formats exactly as "0.0…0…12345", which maximumFractionDigits rounds to "0.0…0". In the absence of trailing zero preservation it would be further transformed to "0", but that no longer happens here.)

test/intl402/NumberFormat/prototype/format/value-decimal-string-trailing-zeros.js
```diff
@@ -0,0 +1,113 @@
...
+assert.sameValue(nf3.format('0.000000'), '0.000');
+assert.sameValue(nf3.format('1.000000'), '1.000');
+assert.sameValue(nf3.format('1.99999'), '2.000');
+assert.sameValue(nf3.format('1.2'), '1.2');
+assert.sameValue(nf3.format('0.12'), '0.12');
+assert.sameValue(nf3.format('0.012'), '0.012');
+assert.sameValue(nf3.format('1.2345e-21'), '0.000000000000000000001235');
+assert.sameValue(nf3.format('1.2345e-1000'), '0.000');
+
+const nf4 = new NumberFormat('en-US', {
```
iamstolis: `NumberFormat` should be `Intl.NumberFormat` here

eemeli: Thank you for spotting that! I've been developing these tests by running them through the [polyfill](https://github.com/eemeli/formatjs-for-keep-trailing-zeros), and clearly forgot to make this adjustment when applying the changes here.

```suggestion
const nf4 = new Intl.NumberFormat('en-US', {
```

iamstolis:
The proposal is not backward compatible and there are several existing tests that get broken by the implementation of the proposal. They should be updated as well:
```
intl402/DurationFormat/prototype/format/numeric-hour-with-zero-minutes-and-non-zero-seconds-with-auto-display.js
intl402/DurationFormat/prototype/format/numeric-hour-with-zero-minutes-and-non-zero-seconds.js
intl402/DurationFormat/prototype/format/precision-exact-mathematical-values.js
```
While `DurationFormat` is not affected by the proposal, the tests produce the expected result using several calls to `NumberFormat` (that is affected by the proposal). This can be fixed, for example, by the explicit removal of trailing zeros in `partitionDurationFormatPattern` in `harness/testIntl.js`
```
-    r = String(r).padStart(exponent, "0");
+    r = String(r).padStart(exponent, "0").replace(/0+$/, "");
```
On the other hand, it shows that this proposal is not a harmless bug-fix and that there may be use-cases on the web that will get broken.

The following two tests are broken by the implementation of the proposal as well
```
intl402/NumberFormat/prototype/format/format-fraction-digits.js
intl402/NumberFormat/test-option-roundingPriority-mixed-options.js
```

sffc:
The proposal is Stage 3. Could we merge the test262 for it? Thanks.

ptomato:
> The proposal is Stage 3. Could we merge the test262 for it? Thanks.

https://github.com/tc39/test262/pull/4608#issuecomment-4554807350 is still unresolved, so it'd help move things along if you could respond to it @eemeli.

PR #4609, opened 2025-10-29
Add tests for Joint Iteration proposal + split tests

bakkot:
This is all of the commits from https://github.com/tc39/test262/pull/4528 plus one additional one to split tests per maintainer request there.

### DIFF
--- a/harness/iteratorZipUtils.js
+++ b/harness/iteratorZipUtils.js
@@ -90,8 +90,8 @@ function _assertIsNullProtoMutableObject(object, label) {
   }
 }
 
-// Assert that the `zipped` iterator results as for Iterator.zip for the first `count` elements.
-// Assumes inputs is an array of arrays of length >= count.
+// Assert that the `zipped` iterator yields the first `count` outputs of Iterator.zip.
+// Assumes `inputs` is an array of arrays, each with length >= `count`.
 // Advances `zipped` by `count` steps.
 function assertZipped(zipped, inputs, count, label) {
   // Last returned elements array.
@@ -117,14 +117,12 @@ function assertZipped(zipped, inputs, count, label) {
     assert.compareArray(value, expected, itemLabel + ": values");
 
     // Ensure value is a packed array with default data properties.
-    //
-    // This operation is destructive, so it has to happen last.
     assertIsPackedArray(value, itemLabel);
   }
 }
 
-// Assert that the `zipped` iterator results as for Iterator.zipKeyed for the first `count` elements.
-// Assumes inputs is an object whose values are arrays of length >= count.
+// Assert that the `zipped` iterator yields the first `count` outputs of Iterator.zipKeyed.
+// Assumes `inputs` is an object whose values are arrays, each with length >= `count`.
 // Advances `zipped` by `count` steps.
 function assertZippedKeyed(zipped, inputs, count, label) {
   // Last returned elements array.
@@ -154,8 +152,6 @@ function assertZippedKeyed(zipped, inputs, count, label) {
     assert.compareArray(Object.values(value), expectedValues, itemLabel + ": result object values");
 
     // Ensure resulting object is a null-prototype mutable object with default data properties.
-    //
-    // This operation is destructive, so it has to happen last.
     _assertIsNullProtoMutableObject(value, itemLabel);
   }
 }
--- a/test/built-ins/Iterator/zip/iterables-iteration-after-reading-options.js
+++ b/test/built-ins/Iterator/zip/iterables-iteration-after-reading-options.js
@@ -49,3 +49,25 @@ assert.compareArray(log, [
   "get padding",
   "get iterator",
 ]);
+
+for (var mode of [undefined, "shortest", "strict"]) {
+  log.length = 0;
+
+  options = {
+    get mode() {
+      log.push("get mode");
+      return mode;
+    },
+    get padding() {
+      log.push("unexpected get padding");
+      return [];
+    }
+  };
+
+  Iterator.zip(iterables, options);
+
+  assert.compareArray(log, [
+    "get mode",
+    "get iterator",
+  ]);
+}
--- a/test/built-ins/Iterator/zip/iterables-iteration.js
+++ b/test/built-ins/Iterator/zip/iterables-iteration.js
@@ -70,7 +70,7 @@ var iterableReturningThrowingIterator = {
 
 // "iterables" argument must be an iterable.
 assert.throws(TypeError, function() {
-  Iterator.zip({});
+  Iterator.zip(Object.create(null));
 });
 
 // GetIteratorFlattenable accepts both iterables and iterators.
@@ -116,7 +116,7 @@ var elements = [
   makeProxyWithGetHandler("second", iterableReturningThrowingIterator),
 
   // An object without any iteration methods.
-  makeProxyWithGetHandler("third", {}),
+  makeProxyWithGetHandler("third", Object.create(null)),
 ];
 
 var elementsIter = elements.values();
--- a/test/built-ins/Iterator/zip/iterator-non-iterable.js
+++ b/test/built-ins/Iterator/zip/iterator-non-iterable.js
@@ -9,9 +9,11 @@ features: [joint-iteration]
 ---*/
 
 var invalidIterables = [
-  {},
-  { next() {}, return() {} },
-  function() {}
+  Object.create(null),
+  Object.create(null, {
+    next: { value: function(){} },
+    return: { value: function(){} },
+  }),
 ];
 
 // Throws a TypeError for invalid iterables values.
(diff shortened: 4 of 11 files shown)


harness/iteratorZipUtils.js
```diff
@@ -0,0 +1,223 @@
...
+
+    // Ensure all array elements have the expected value.
+    var expected = inputs.map(function (array) {
+      return array[i];
+    });
+    assert.compareArray(value, expected, itemLabel + ": values");
+
+    // Ensure value is a packed array with default data properties.
+    //
+    // This operation is destructive, so it has to happen last.
```
michaelficarra: What does it mean for this to be destructive? It doesn't look like it mutates its input.

harness/iteratorZipUtils.js
```diff
@@ -0,0 +1,223 @@
...
+    assertIsPackedArray(value, itemLabel);
+  }
+}
+
+// Assert that the `zipped` iterator results as for Iterator.zipKeyed for the first `count` elements.
```
michaelficarra: This sentence has some kind of grammatical error that is preventing me from reading it. Same for `assertZipped`.

harness/iteratorZipUtils.js
```diff
@@ -0,0 +1,223 @@
...
+// Assumes inputs is an object whose values are arrays of length >= count.
```
michaelficarra:
```suggestion
// Assumes `inputs` is an object whose values are arrays, each with length >= `count`.
```

Same for `assertZipped`.

test/built-ins/Iterator/zip/iterables-iteration-after-reading-options.js
```diff
@@ -0,0 +1,51 @@
...
+  },
+  get padding() {
+    log.push("get padding");
+    return [];
+  }
+};
+
+Iterator.zip(iterables, options);
+
+assert.compareArray(log, [
```
michaelficarra, line 47: I'd like to see another similar assertion that `log` is just `["get mode", "get iterator"]` whenever the mode getter returns any of the other acceptable modes. In other words, `padding` is only read when `mode` is `"longest"`.

test/built-ins/Iterator/zip/options-padding.js
```diff
@@ -0,0 +1,54 @@
...
+}
+
+// Throws a TypeError for invalid padding options.
+for (var padding of invalidPadding) {
+  assert.throws(TypeError, function() {
+    Iterator.zip([], {mode: "longest", padding});
+  });
+}
+
+// Invalid padding options are ignored when mode is not "longest".
```
michaelficarra: This is a bit misleading. They're not even read.

test/built-ins/Iterator/zip/options-padding.js
```diff
@@ -0,0 +1,54 @@
...
+    7. If mode is "longest", then
+      a. Set paddingOption to ? Get(options, "padding").
+      b. If paddingOption is not undefined and paddingOption is not an Object, throw a TypeError exception.
+    ...
+features: [joint-iteration]
+---*/
+
+var validPadding = [
+  undefined,
+  [],
```
michaelficarra, line 21:
```suggestion
  [],
  Object("string"),
```

test/built-ins/Iterator/zip/iterables-iteration.js
```diff
@@ -0,0 +1,164 @@
...
+
+var iterableReturningThrowingIterator = {
+  [Symbol.iterator]() {
+    return throwingIterator;
+  }
+};
+
+// "iterables" argument must be an iterable.
+assert.throws(TypeError, function() {
+  Iterator.zip({});
```
michaelficarra:
```suggestion
  Iterator.zip(Object.create(null));
```

This test requires the reader to know that `Object.prototype` doesn't have a `Symbol.iterator`. It's more self-contained to just show it to them.

test/built-ins/Iterator/zip/iterables-iteration.js
```diff
@@ -0,0 +1,164 @@
...
+
+var elements = [
+  // An iterator.
+  makeProxyWithGetHandler("first", throwingIterator),
+
+  // An iterable.
+  makeProxyWithGetHandler("second", iterableReturningThrowingIterator),
+
+  // An object without any iteration methods.
+  makeProxyWithGetHandler("third", {}),
```
michaelficarra:
```suggestion
  makeProxyWithGetHandler("third", Object.create(null)),
```

test/built-ins/Iterator/zip/iterator-non-iterable.js
```diff
@@ -0,0 +1,22 @@
...
+esid: sec-iterator.zip
+description: >
+  Throws a TypeError when the "iterables" argument is not iterable.
+features: [joint-iteration]
+---*/
+
+var invalidIterables = [
+  {},
+  { next() {}, return() {} },
+  function() {}
```
michaelficarra: I don't think these tests add anything over just testing `Object.create(null)`.

test/built-ins/Iterator/zip/non-constructible.js
```diff
@@ -0,0 +1,21 @@
...
+  Iterator.zip is not constructible.
+
+  Built-in function objects that are not identified as constructors do not
+  implement the [[Construct]] internal method unless otherwise specified in the
+  description of a particular function.
+features: [joint-iteration]
+---*/
+
+assert.throws(TypeError, () => {
+  new Iterator.zip();
```
michaelficarra: This assertion is misleading and unnecessary, since the same error would be thrown even if `new` was not used. And we already have a test for a missing first parameter in `test/built-ins/Iterator/zip/iterables-primitive.js`.

test/built-ins/Iterator/zip/iterables-iteration-after-reading-options.js
michaelficarra: I guess the assertions in `test/built-ins/Iterator/zip/options-mode.js` are good enough.

test/built-ins/Iterator/zipKeyed/options-padding.js
```diff
@@ -0,0 +1,54 @@
...
+    7. If mode is "longest", then
+      a. Set paddingOption to ? Get(options, "padding").
+      b. If paddingOption is not undefined and paddingOption is not an Object, throw a TypeError exception.
+    ...
+features: [joint-iteration]
+---*/
+
+var validPadding = [
+  undefined,
+  [],
```
michaelficarra: Not a big deal, but this should probably be `{}` since `zipKeyed` doesn't rely on iterability.

test/built-ins/Iterator/zipKeyed/results-object-has-default-attributes.js
```diff
@@ -0,0 +1,106 @@
...
+  for (var key of expectedKeys) {
+    verifyProperty(actual, key, {
+      writable: true,
+      enumerable: true,
+      configurable: true,
+    });
+  }
+}
+
+var iterables = Object.create(Array.prototype, {
```
michaelficarra:
```suggestion
var iterables = Object.create(null, {
```

Why are we inheriting from `Array.prototype` here?

test/built-ins/Iterator/zipKeyed/iterables-iteration-get-abrupt-completion.js
```diff
@@ -0,0 +1,96 @@
...
+
+    // IteratorClose ignores new exceptions when called with a Throw completion.
+    throw new Test262Error();
+  },
+};
+
+var iterables = {
+  first,
+  second,
+  get third() {
```
michaelficarra, line 110: Do we need a separate test that asserts the full object property enumeration order, including symbols and numeric keys?

michaelficarra (approved):
LGTM otherwise! Looks like this fully covers the test plan and then some. Thanks again @anba for writing these tests.

test/built-ins/Iterator/zipKeyed/results-object-has-default-attributes.js
michaelficarra: I'm assuming it's due to this line from the testing plan:

> output object is not an array (even if the input iterables object is an array)

Though inheriting from Array.prototype doesn't make it an array. I'd just leave that aspect out, there's no reason to believe that the input object being an array would change the kind of output object in any reasonable implementation.

michaelficarra:
FYI I've confirmed the latest diff with the polyfill from the proposal repo and it still passes.

ptomato:
This was written by someone involved in the implementation and reviewed by two proposal champions. I would be comfortable to merge it now, to avoid holding it up any further. I would prefer that in the future we are more strict with the "mini-polyfill" stuff but I personally won't have the bandwidth to propose an alternative, nor review the rest of the PR, in the near future. @gibson042 @ljharb @Ms2ger What do you think?

Ms2ger:
I would be okay with that.

ptomato (approved):
Merging as discussed

PR #4612, opened 2025-10-29
Added test for Islamic calendar fallbacks [intl-era-monthcode]

ben-allen:
The Era Monthcode spec indicates that the fallback when "islamic" or "islamic-rgsa" calendars used must be  "an implementation- and locale-defined [calendar type](https://tc39.es/proposal-intl-era-monthcode/#sec-ecma402-calendar-types) that is one of the values returned from [AvailableCalendars](https://tc39.es/proposal-intl-era-monthcode/#sup-availablecalendars)."

This tests for that.


ben-allen:
closed, replaced with #4627

ben-allen:
closed, replaced by #4627

PR #4618, opened 2025-10-29
Temporal: Add an overflow test for date addition with maximum year

catamorphism:


ptomato (changes requested):
We should try to check for this bug in all of the places where it could occur:

- PlainDate.prototype.subtract
- PlainDateTime.prototype.{add, subtract}
- PlainYearMonth.prototype.{add, subtract}
- ZonedDateTime.prototype.{add, subtract}

test/built-ins/Temporal/PlainDate/prototype/subtract/overflow-subtracting-months-from-min-year.js
```diff
@@ -0,0 +1,13 @@
+// Copyright (C) 2025 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.plaindate.prototype.subtract
+description: Subtracting months from minimum year should throw
+features: [Temporal]
+---*/
+
+const minYear = new Temporal.PlainDate(-271821, 1, 1);
```
ptomato:
```suggestion
const minYear = new Temporal.PlainDate(-271820, 1, 1);
```
(Otherwise these won't construct in the first place; same comment for all the `overflow-subtracting-months-from-min-year.js)

catamorphism: Fixed in [563ed2d](https://github.com/tc39/test262/pull/4618/commits/563ed2d73734deb834ef1d79f133ef1126002fd3)

catamorphism:
> I'd also suggest adding `minYear.add(new Temporal.Duration(0, -5432, -5432))` and `maxYear.subtract(new Temporal.Duration(0, -5432, -5432)` etc.

@ptomato Done in [8a6ad69](https://github.com/tc39/test262/pull/4618/commits/8a6ad69083a38100067ab7045abd5eacf1e12fde).

PR #4621, opened 2025-10-30
[import-defer] adding new test cases to validate evaluation trigger

caiolima:
This PR is adding cases when namespace object is in the prototype chain go another object for `[[Get]]` and `[[HasProperty]]`. It also adds a case that validates that Private Names installation, set and get don't trigger deferred modules evaluation.


test/language/import/import-defer/evaluation-triggers/ignore-private-name-access.js
```diff
@@ -0,0 +1,45 @@
...
+  static mark(obj) {
+    new Marker(obj);
+  }
+}
+
+assert.throws(TypeError, function () {
+  Marker.mark(ns);
+});
+
+assert.sameValue(globalThis.evaluations.length, 0, "It does not trigger evaluation");
```
nicolo-ribaudo, line 51: Can we also check that `#mark in ns` does not trigger evaluation?

caiolima: Done.

caiolima:
Thanks for the review!

PR #4622, opened 2025-10-30
Temporal: Add tests for correct rounding of epoch milliseconds in toString()

catamorphism:
This tests for a bug observed in JSC where `toString()` was off by one second for Instants due to rounding seconds up instead of down when extracting the time parts from the instant.


Ms2ger (approved):

ptomato:
I'm surprised that this wasn't covered by https://github.com/tc39/test262/blob/main/test/built-ins/Temporal/Instant/prototype/toString/negative-epochnanoseconds.js - was JSC passing that test?

catamorphism:
> I'm surprised that this wasn't covered by https://github.com/tc39/test262/blob/main/test/built-ins/Temporal/Instant/prototype/toString/negative-epochnanoseconds.js - was JSC passing that test?

@ptomato Yes, JSC was passing that test. I haven't tried to analyze why the two cases are different.

ptomato:
Huh, interesting. Thanks!

PR #4623, opened 2025-10-30
Fix %ThrowTypeError% cross-realm error check

linusg:
See https://github.com/tc39/test262/pull/4604#pullrequestreview-3402010945. Currently this test throws `Expected a TypeError but got a different error constructor with the same name`.


linusg:
test262.fyi data backs this up:

image

linusg:
CI doesn't work for this (https://github.com/tc39/test262/pull/4604#issuecomment-3462327626) so it won't auto-merge :slightly_smiling_face:

PR #4624, opened 2025-10-30
Temporal: Add a test for rounding negative instants

catamorphism:
This test catches a bug found in JSC's implementation of `RoundNumberToIncrementAsIfPositive`.

### DIFF
--- a/test/built-ins/Temporal/Instant/prototype/toString/negative-instant-rounding.js
+++ b/test/built-ins/Temporal/Instant/prototype/toString/negative-instant-rounding.js
@@ -14,55 +14,55 @@ const roundedDown = "1938-04-24T22:13:20.000Z";
 const roundedUp = "1938-04-24T22:13:20.001Z";
 
 assert.sameValue(
-    instance.toString({smallestUnit: 'millisecond', roundingMode: 'halfCeil'}),
-    roundedDown,
-    "Rounding with halfCeil rounds to the closest millisecond"
+  instance.toString({smallestUnit: 'millisecond', roundingMode: 'halfCeil'}),
+  roundedDown,
+  "Rounding with halfCeil rounds to the closest millisecond"
 );
 
 assert.sameValue(
-    instance.toString({smallestUnit: 'millisecond', roundingMode: 'halfFloor'}),
-    roundedDown,
-    "Rounding with halfFloor rounds to the closest millisecond"
+  instance.toString({smallestUnit: 'millisecond', roundingMode: 'halfFloor'}),
+  roundedDown,
+  "Rounding with halfFloor rounds to the closest millisecond"
 );
 
 assert.sameValue(
-    instance.toString({smallestUnit: 'millisecond', roundingMode: 'halfExpand'}),
-    roundedDown,
-    "Rounding with halfExpand rounds to the closest millisecond"
+  instance.toString({smallestUnit: 'millisecond', roundingMode: 'halfExpand'}),
+  roundedDown,
+  "Rounding with halfExpand rounds to the closest millisecond"
 );
 
 assert.sameValue(
-    instance.toString({smallestUnit: 'millisecond', roundingMode: 'halfTrunc'}),
-    roundedDown,
-    "Rounding with halfTrunc rounds to the closest millisecond"
+  instance.toString({smallestUnit: 'millisecond', roundingMode: 'halfTrunc'}),
+  roundedDown,
+  "Rounding with halfTrunc rounds to the closest millisecond"
 );
 
 assert.sameValue(
-    instance.toString({smallestUnit: 'millisecond', roundingMode: 'halfEven'}),
-    roundedDown,
-    "Rounding with halfEven rounds to the closest millisecond"
+  instance.toString({smallestUnit: 'millisecond', roundingMode: 'halfEven'}),
+  roundedDown,
+  "Rounding with halfEven rounds to the closest millisecond"
 );
 
 assert.sameValue(
-    instance.toString({smallestUnit: 'millisecond', roundingMode: 'ceil'}),
-    roundedUp,
-    "Rounding with ceil rounds to the next millisecond"
+  instance.toString({smallestUnit: 'millisecond', roundingMode: 'ceil'}),
+  roundedUp,
+  "Rounding with ceil rounds to the next millisecond"
 );
 
 assert.sameValue(
-    instance.toString({smallestUnit: 'millisecond', roundingMode: 'floor'}),
-    roundedDown,
-    "Rounding with floor rounds to the closest millisecond"
+  instance.toString({smallestUnit: 'millisecond', roundingMode: 'floor'}),
+  roundedDown,
+  "Rounding with floor rounds to the closest millisecond"
 );
 
 assert.sameValue(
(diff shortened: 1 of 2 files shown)


test/built-ins/Temporal/Instant/prototype/round/negative-instant.js
```diff
@@ -0,0 +1,67 @@
+// Copyright (C) 2025 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.instant.prototype.round
+description: RoundNumberToIncrementAsIfPositive should multiply the remainder by an extra sign
+             before comparing it
```
ptomato: To fix the linter error:
```suggestion
description: >
  RoundNumberToIncrementAsIfPositive should multiply the remainder by an extra sign
  before comparing it
```
(same below)

catamorphism: Fixed separately

catamorphism:
@ptomato 
> A few suggestions for these tests:
> 
>  RoundNumberToIncrementAsIfPositive is also called from Temporal.Instant.prototype.toString() and Temporal.ZonedDateTime.prototype.toString(), is it possible to trigger the bug from there?
> 

Yes, it is -- I added tests for that.

>  `rounding-direction.js` seems to cover the same topic but doesn't hit this bug (JSC passes it, except for the ZonedDateTime one) so maybe these cases could be folded into that file?

Since it tests a different code path, I would rather keep it separate.

> 
> Would it make sense to have `instance` be exactly halfway between `roundedDown` and `roundedUp` so that the rounding modes really get exercised, or does that not trigger the bug?

That doesn't trigger the bug.

test/built-ins/Temporal/Instant/prototype/toString/negative-instant-rounding.js
```diff
@@ -0,0 +1,68 @@
...
+
+/*---
+esid: sec-temporal.instant.prototype.tostring
+description: >
+  RoundNumberToIncrementAsIfPositive should multiply the remainder by an extra sign
+  before comparing it.
+features: [Temporal]
+---*/
+
+const instance = new Temporal.Instant(-999999999999999990n); // 1938-04-24T22:13:20.001Z
```
Ms2ger:
```suggestion
const instance = new Temporal.Instant(-999999999999999990n); // 1938-04-24T22:13:20.00000001Z
```

test/built-ins/Temporal/Instant/prototype/round/negative-instant.js
```diff
@@ -0,0 +1,68 @@
...
+esid: sec-temporal.instant.prototype.round
+description: >
+  RoundNumberToIncrementAsIfPositive should multiply the remainder by an extra sign
+  before comparing it
+features: [Temporal]
+---*/
+
+const instance = new Temporal.Instant(-1000000000000000000n);  // 1938-04-24T22:13:20Z
+const roundedDown = new Temporal.Instant(-1000000800000000000n); // 1938-04-24T22:00:00Z
+const roundedUp = new Temporal.Instant(-999997200000000000n); // 1938-04-24T23:00:00Z
```
Ms2ger: You're creating these `Instant` objects only to repeatedly read out the nanoseconds value you passed in. Up to you to consider whether the indirection is helpful.

test/built-ins/Temporal/Instant/prototype/toString/negative-instant-rounding.js
```diff
@@ -0,0 +1,68 @@
...
+const roundedDown = "1938-04-24T22:13:20.000Z";
+const roundedUp = "1938-04-24T22:13:20.001Z";
+
+assert.sameValue(
+    instance.toString({smallestUnit: 'millisecond', roundingMode: 'halfCeil'}),
```
Ms2ger: I don't think we have a policy, but you're using two-space indents elsewhere.

test/built-ins/Temporal/ZonedDateTime/prototype/round/negative-time.js
```diff
@@ -0,0 +1,68 @@
...
+esid: sec-temporal.zoneddatetime.prototype.round
+description: >
+  RoundNumberToIncrementAsIfPositive should multiply the remainder by an extra sign
+  before comparing it
+features: [Temporal]
+---*/
+
+const instance = new Temporal.ZonedDateTime(-1000000000000000000n, "UTC");  // 1938-04-24T22:13:20Z
+const roundedDown = new Temporal.ZonedDateTime(-1000000800000000000n, "UTC"); // 1938-04-24T22:00:00Z
+const roundedUp = new Temporal.ZonedDateTime(-999997200000000000n, "UTC"); // 1938-04-24T23:00:00Z
```
Ms2ger: Same here about the objects

test/built-ins/Temporal/ZonedDateTime/prototype/toString/negative-zoneddatetime-rounding.js
```diff
@@ -0,0 +1,68 @@
...
+  before comparing it.
+features: [Temporal]
+---*/
+
+const instance = new Temporal.ZonedDateTime(-999999999999999990n, "UTC"); // '1938-04-24T22:13:20.00000001+00:00[UTC]'
+const roundedDown = "1938-04-24T22:13:20.000+00:00[UTC]";
+const roundedUp = "1938-04-24T22:13:20.001+00:00[UTC]";
+
+assert.sameValue(
+    instance.toString({smallestUnit: 'millisecond', roundingMode: 'halfCeil'}),
```
Ms2ger: Same about indentation

test/built-ins/Temporal/Instant/prototype/round/negative-instant.js
catamorphism: Good point, changed in [64153f3](https://github.com/tc39/test262/pull/4624/commits/64153f328a84f715ec7ced7d2af9f527f47c6172)

test/built-ins/Temporal/ZonedDateTime/prototype/round/negative-time.js
catamorphism: Fixed in [64153f3](https://github.com/tc39/test262/pull/4624/commits/64153f328a84f715ec7ced7d2af9f527f47c6172)

test/built-ins/Temporal/Instant/prototype/toString/negative-instant-rounding.js
catamorphism: Fixed in [d5d1f2a](https://github.com/tc39/test262/pull/4624/commits/d5d1f2a32df8092c26db0d4fc358d5c76129f928)

test/built-ins/Temporal/ZonedDateTime/prototype/toString/negative-zoneddatetime-rounding.js
catamorphism: Fixed in [d5d1f2a](https://github.com/tc39/test262/pull/4624/commits/d5d1f2a32df8092c26db0d4fc358d5c76129f928)

PR #4625, opened 2025-10-31
Temporal: Add coverage for adding durations with large subsecond values to a PlainTime

catamorphism:


test/built-ins/Temporal/PlainTime/prototype/add/add-large-milliseconds.js
```diff
@@ -0,0 +1,26 @@
...
+
+/*---
+esid: sec-temporal.plaintime.prototype.add
+description: Adding unbalanced durations with large subsecond values to a time
+includes: [temporalHelpers.js]
+features: [Temporal]
+---*/
+
+const t = new Temporal.PlainTime(0);
+const bigNumber = 9007199254740990976;
```
Ms2ger, line 12: It looks like this is actually rounded to 9007199254740991000; please adjust.

ptomato: Depends on [how you print it](https://float.exposed/0x43df3fffffffffff).

```js
> 9007199254740990976
9007199254740991000
> BigInt(9007199254740991000).toString()
'9007199254740990976'
```

catamorphism: Despite how it's printed, I think the internal representation is correct?

```
> Temporal.Duration.from({milliseconds: 9007199254740990976})
Temporal.Duration <PT9007199254740990.976S> {
  milliseconds: 9007199254740991000
}
```
The seconds and nanoseconds values are correct in the string representation of the duration.

catamorphism:
> We should probably have similar tests for PlainDateTime and ZonedDateTime.

@ptomato Done in [f96a250](https://github.com/tc39/test262/pull/4625/commits/f96a25047c58c99134b8cca30b8e934572f04011)

test/built-ins/Temporal/PlainTime/prototype/add/add-large-milliseconds.js
ptomato: I think probably it's better to write it as is, since that's closest to what Temporal.Duration picks up.

PR #4626, opened 2025-10-31
Added Intl.Era-monthcode test verifying all calendars supported

ben-allen:
Intl Era Monthcode requires support for the calendars given in [this table](https://tc39.es/proposal-intl-era-monthcode/#table-calendar-types). This tests for that.

### DIFF
--- a/test/intl402/Intl/supportedValuesOf/calendars-required-by-intl-era-monthcode.js
+++ b/test/intl402/Intl/supportedValuesOf/calendars-required-by-intl-era-monthcode.js
@@ -46,11 +46,6 @@ const requiredCalendars = [
 ]
 
 const supportedCalendars = Intl.supportedValuesOf("calendar");
-let currentCalendar = "";
-
-function containedInCalendar(requiredCalendar){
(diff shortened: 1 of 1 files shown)


test/intl402/Intl/supportedValuesOf/calendars-required-by-intl-era-monthcode.js
```diff
@@ -0,0 +1,56 @@
...
+
+const supportedCalendars = Intl.supportedValuesOf("calendar");
+let currentCalendar = "";
+
+function containedInCalendar(requiredCalendar){
+  currentCalendar = requiredCalendar;
+  return supportedCalendars.includes(requiredCalendar);
+}
+
+assert.sameValue(requiredCalendars.every(containedInCalendar), true, "Required calendar: " + currentCalendar + " not supported");
```
ptomato: This is a bit more complicated than it needs to be, could I suggest something like this?
```suggestion
for (const calendar of requiredCalendars) {
  assert(supportedCalendars.includes(calendar), "Required calendar: " + calendar + " must be supported");
}
```

PR #4632, opened 2025-11-03
Intl.Era-monthcode: tests to ensure `weekofYear` only set when using `"iso8601"` calendar 

ben-allen:
Added tests to Temporal.PlainDate, Temporal.PlainDateTime, and Temporal.ZonedDateTime ensuring weekOfYear only defined for "iso8601" calendar

### DIFF
--- a/test/intl402/Temporal/PlainDate/prototype/weekOfYear/non-iso-week-of-year.js
+++ b/test/intl402/Temporal/PlainDate/prototype/weekOfYear/non-iso-week-of-year.js
@@ -9,14 +9,29 @@ description: >
 features: [Temporal, Intl.Era-monthcode]
 ---*/
 
-assert.sameValue(
-  new Temporal.PlainDate(2024, 1, 1, "gregory").weekOfYear,
-  undefined,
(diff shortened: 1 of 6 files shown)


test/intl402/Temporal/PlainDate/prototype/weekOfYear/non-iso-week-of-year.js
```diff
@@ -9,14 +9,31 @@ description: >
...
+  "dangi",
+  "ethioaa",
+  "ethiopic",
+  "gregory",
+  "hebrew",
+  "indian",
+  "islamic-civil",
+  "islamic-tbla",
+  "islamic-umalqura",
+  "islamicc",
```
ptomato:
```suggestion
```

ptomato: Needs to be applied to the list in each file, not just the first file

ptomato:
Fixed it up since that's the only thing blocking.

PR #4633, opened 2025-11-04
Temporal: Add test for CalendarYearMonthFromFields with Japanese dates

catamorphism:
Japanese eras can begin in the middle of a month. This adds tests for the existing behavior, which is that PlainYearMonth.from returns a PlainYearMonth with an era matching the first day of the month, even when that conflicts with the era given by the argument.

See https://github.com/tc39/proposal-temporal/issues/3150 for more details.

### DIFF
--- a/test/intl402/Temporal/PlainYearMonth/from/reference-day-japanese.js
+++ b/test/intl402/Temporal/PlainYearMonth/from/reference-day-japanese.js
@@ -3,7 +3,9 @@
 
 /*---
 esid: sec-temporal.calendar.prototype.calendaryearmonthfromfields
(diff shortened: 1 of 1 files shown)


Ms2ger:
Can you add a pointer to the discussion inside the test?

catamorphism:
> Can you add a pointer to the discussion inside the test?

@Ms2ger done in [c953a03](https://github.com/tc39/test262/pull/4633/commits/c953a03b6899fdd64fbae81789d686224ca235e9)

ptomato (approved):
Thanks!

PR #4634, opened 2025-11-04
Temporal: Add test for formatting Temporal objects with era formatter

catamorphism:
See https://github.com/tc39/proposal-temporal/issues/3049


catamorphism:
Awaiting https://github.com/tc39/proposal-temporal/pull/3175

test/intl402/DateTimeFormat/prototype/format/temporal-objects-format-with-era.js
```diff
@@ -0,0 +1,19 @@
...
+  If era option and no other options are provided to DateTimeFormat constructor,
+  objects should be formatted with default options
+features: [Temporal]
+---*/
+
+const formatter = new Intl.DateTimeFormat([], { era: "narrow" });
+
+assert(formatter.format(new Temporal.PlainDate(2025, 11, 4)).startsWith("11"), "formatting a PlainDate should work");
+assert(formatter.format(new Temporal.PlainYearMonth(2025, 11, "gregory")).startsWith("11"), "formatting a PlainYearMonth should work");
+assert(formatter.format(new Temporal.PlainMonthDay(11, 4, "gregory")).startsWith("11"), "formatting a PlainMonthDay should work");
```
ptomato, line 17: Using `"gregory"` may fail depending on the calendar that `formatter` gets from the host's locale. Probably the best way to work around this would be to hardcode the formatter to `"en"`, and include `locales: [en]` in the frontmatter.

test/intl402/DateTimeFormat/prototype/format/temporal-objects-format-with-era.js
```diff
@@ -0,0 +1,19 @@
...
+assert(formatter.format(new Temporal.PlainTime(14, 46)).startsWith("A"), "formatting a PlainTime should work");
```
ptomato: I'm a bit surprised by this expectation... It does match what comes out of `new Date().toLocaleTimeString('en', { era: 'narrow' })`, but my reading of the spec ([GetDateTimeFormat](https://tc39.es/proposal-temporal/#sec-getdatetimeformat)) is that `era` isn't copied to _formatOptions_ when constructing [[TemporalPlainTimeFormat]].

test/intl402/DateTimeFormat/prototype/format/temporal-objects-format-with-era.js
```diff
@@ -0,0 +1,19 @@
...
+assert(formatter.format(new Temporal.PlainDateTime(2025, 11, 4, 14, 46)).startsWith("11"), "formatting a PlainDateTime should work");
+assert(formatter.format(new Temporal.Instant(0n)).startsWith("12"), "formatting an Instant should work");
```
ptomato: This may fail depending on the host's time zone offset. Probably the best way to work around that is pass `timeZone: 'UTC'` in the formatter options.

ptomato (changes requested):
Thanks!

We should have similar tests for `formatRange`, `formatToParts`, `formatRangeToParts`, and the `toLocaleString` methods of PlainDate, PlainYearMonth, PlainMonthDay, PlainTime, PlainDateTime, Instant, and ZonedDateTime.

test/intl402/DateTimeFormat/prototype/format/temporal-objects-format-with-era.js
catamorphism: Done in [8ae6c62](https://github.com/tc39/test262/pull/4634/commits/8ae6c6275539d863398cc89d86fc478ea78db52b)

test/intl402/DateTimeFormat/prototype/format/temporal-objects-format-with-era.js
catamorphism: Changed https://github.com/tc39/proposal-temporal/pull/3175#pullrequestreview-3425041423 accordingly and updated this test.

test/intl402/DateTimeFormat/prototype/format/temporal-objects-format-with-era.js
ptomato: s/locales/locale apparently. Sorry!

test/intl402/DateTimeFormat/prototype/format/temporal-objects-format-with-era.js
catamorphism: Fixed in [28e653d](https://github.com/tc39/test262/pull/4634/commits/28e653d28d985bea3622b15f9ea8d96ea6ae79f6)

test/intl402/DateTimeFormat/prototype/format/temporal-objects-format-with-era.js
catamorphism: Actually, it looks like the linter complains about including `locales: [en]` in the frontmatter? Or did I do that incorrectly?

test/intl402/DateTimeFormat/prototype/format/temporal-objects-format-with-era.js
ptomato: I guess I gave you bad advice here, since the test is supposed to check what happens when you pass `era` with no other options.

Maybe a better approach would be to check that the output of `formatter.format(new Temporal.Instant(0n))` is identical to that of `new Date(0).toLocaleString(["en"], { era: "narrow" })`?

test/intl402/DateTimeFormat/prototype/format/temporal-objects-format-with-era.js
ptomato: I got it wrong, it's `locale` https://github.com/tc39/test262/blob/main/CONTRIBUTING.md#locale

catamorphism: Fixed in [bd540ad](https://github.com/tc39/test262/pull/4634/commits/bd540ad97673d99d15728af767fe7e3f940bd445)

test/intl402/DateTimeFormat/prototype/format/temporal-objects-format-with-era.js
catamorphism: Done in [eb7cd3f](https://github.com/tc39/test262/pull/4634/commits/eb7cd3f84fc20fccc37bbfd906dff942ea734541)

catamorphism:
> These look correct to me. One nitpick: all the files in `test/built-ins/` should move to `test/intl402/` since they are only valid for implementations with Intl.
> 

@ptomato Fixed.

> (Not merging yet because it's marked draft)

Ready now!

PR #4636, opened 2025-11-04
Temporal: Add tests for a non-ISO-calendar edge case with leap years and differencing

catamorphism:


catamorphism:
Depends on spec PR https://github.com/tc39/proposal-temporal/pull/3176

catamorphism:
@ptomato Added some more tests and extended to PlainDateTime/ZonedDateTime. I don't think it's possible to trigger the bug with PlainYearMonth, since it has to do with the days field of the duration having the wrong sign.

PR #4644, opened 2025-11-06
Fix filename linter in Windows

lpardosixtosMs:
When parsing the file names in Windows, `tools/lint/lib/collect_files.py` produces strings contain backslashes that are later flagged as invalid characters in the file name. This change normalizes the strings to always use forward slashes.


ptomato:
Seems like we should use the `posixpath` module, or else amend the `tools/lint/lib/checks/filename.py` check to handle Windows paths.

Issue #4647, opened 2025-11-09
Test plan for import bytes


styfle:
Proposal repo: https://github.com/tc39/proposal-import-bytes

- [x] import empty bytes
- [x] import txt as bytes
- [x] import png as bytes
- [x] import js as bytes
- [x] import json as bytes
- [x] dynamic import bytes
- [x] import twice idempotency?
- [x] in all cases above, check for immutability

ljharb:
- import JS as bytes
 - import JSON as bytes

nicolo-ribaudo:
For the two suggested by @ljharb, it'd be great to also have wpt tests that serve those files with the corresponding mime type.

Also, test262 tests for a file imported both as json/js _and_ as bytes.

ptomato:
Some more suggestions:
- files with potentially troublesome `\0` characters, like a UTF-8 text file followed by a `\0` followed by more text
- files with potentially troublesome encoding, like invalid UTF-8
- results are always backed by an immutable ArrayBuffer

nicolo-ribaudo:
Also, note that the proposal doesn't actually say anything about encoding or about the content of the Uint8Array. It would be valid for a file to contain a random list of bytes and for the Uint8Array to contain a different list of bytes.

ptomato:
> It would be valid for a file to contain a random list of bytes and for the Uint8Array to contain a different list of bytes.

What? Surely that's an omission in the proposal, then.

nicolo-ribaudo:
The proposal doesn't deal with I/O, it only says "when you use the `type: "bytes"` attribute, either error or give me an Uint8Array around an Immutable array buffer". The process of getting those bytes to JavaScript is entirely host-defined, for example Node.js will deal with a filesystem while the web will fetch.

A lot of the tests for this proposal will probably need to be in WPT.

ptomato:
I think we have to make at least _some_ concession to practicality here...

bakkot:
Ecma262 doesn't say anything about encoding of ES modules either. I think we can make at least the same assumptions that we use for testing imports of ES modules which are represented as files on disk. But I do think any trickier encoding-related stuff (byte-order marks or whatever) should be left for WPT.

styfle:
Thanks all, I updated my PR with a few more tests:

- https://github.com/tc39/test262/pull/4648

As for "wpt", is that something in this repo or is that another repo? Please let me know how to proceed, thanks!

bakkot:
[...]

PR #4648, opened 2025-11-09
[import-bytes] add initial tests for import bytes proposal

styfle:
- Closes https://github.com/tc39/test262/issues/4647

### DIFF
--- a/test/language/import/import-bytes/bytes-from-js.js
+++ b/test/language/import/import-bytes/bytes-from-js.js
@@ -13,8 +13,8 @@ import value from './bytes-from-js_FIXTURE.js' with { type: 'bytes' };
 assert(value instanceof Uint8Array);
 assert(value.buffer instanceof ArrayBuffer);
 
-assert.sameValue(value.length, 16);
-assert.sameValue(value.buffer.byteLength, 16);
+assert.sameValue(value.length, 139);
+assert.sameValue(value.buffer.byteLength, 139);
 assert.sameValue(value.buffer.immutable, true);
 
 assert.compareArray(


test/language/import/import-bytes/bytes-from-empty.js
```diff
@@ -0,0 +1,27 @@
...
+description: Creates bytes module from txt file
+flags: [module]
+features: [import-attributes, immutable-arraybuffer, import-bytes]
+includes: [compareArray.js]
+---*/
+
+import value from './bytes-from-empty-FIXTURE.bin' with { type: 'bytes' };
+
+assert(value instanceof Uint8Array);
+assert(value.buffer instanceof ArrayBuffer);
```
ljharb, line 14: there should be some kind of assertion that it's immutable (not sure what that'd look like tho)

same throughout

ptomato:
Note, the empty file is failing because fixture files need to contain the string `_FIXTURE` and these have a dash where they should have an underscore: https://github.com/tc39/test262/blob/main/INTERPRETING.md#modules

test/language/import/import-bytes/bytes-from-empty.js
styfle: The lines below are checking for immutable (asserts that an error is thrown on resize and also on transfer)

```ts
  value.buffer.resize(0);
});

  value.buffer.transfer();
});
```

Let me know if there is a better way to check for immutability.

anba: Directly call the `immutable` accessor property <https://tc39.es/proposal-immutable-arraybuffer/#sec-get-arraybuffer.prototype.immutable> ?

styfle:
@ptomato Thanks! I updated the fixtures to use underscores and that solved the `Test262File: missing "contents"` error.

However, I ran into another error.

```
  File "/home/runner/work/test262/test262/tools/generation/lib/test.py", line 29, in load
    self.source = handle.read()
                  ~~~~~~~~~~~^^
  File "<frozen codecs>", line 325, in decode
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x89 in position 0: invalid start byte
```

So I pushed some changes to the python code to ignore binary files.

I see there are still a few engines on circleci that are failing. Please let me know how to proceed, thanks!

test/language/import/import-bytes/bytes-from-empty.js
styfle: Interesting. I don't see any existing tests for `ArrayBuffer.prototype.immutable`.

I was basing this new test on the existing tests here:

https://github.com/tc39/test262/blob/3fd4ec27f1798ebecafc73b354a45dcdda9bde29/test/built-ins/ArrayBuffer/prototype/transfer/this-is-immutable-arraybuffer.js#L27-L30

https://github.com/tc39/test262/blob/3fd4ec27f1798ebecafc73b354a45dcdda9bde29/test/built-ins/ArrayBuffer/prototype/resize/this-is-immutable-arraybuffer-object.js#L20-L23

styfle: I went ahead and added the new assertion in [cafcddc](https://github.com/tc39/test262/pull/4648/commits/cafcddc7c48df7404435567faf599c03001550ea)

bakkot: > Interesting. I don't see any existing tests for `ArrayBuffer.prototype.immutable`.

Still [in PR](https://github.com/tc39/test262/pull/4541), alas.

ptomato:
@styfle instead of hardcoding binary extensions I think it'd be better to just skip cleaning any files with `_FIXTURE` in the name? Would that work for you?

The CI jobs are timing out because they are detecting hundreds of tests as different in this PR than from main. Probably if you rebase on the latest state of the main branch, that'll go away.

styfle:
@ptomato Thanks! I rebased on main and force pushed. That seemed to fix CI ✅ 

I also applied your suggestion to skip files with `_FIXTURE` in [7d42a35](https://github.com/tc39/test262/pull/4648/commits/7d42a3584376fafaafe6e19e13da8b136dd414b8).

Let me know how that looks, thanks!

test/language/import/import-bytes/bytes-from-js.js
```diff
@@ -0,0 +1,48 @@
...
+import value from './bytes-from-js_FIXTURE.js' with { type: 'bytes' };
+
+assert(value instanceof Uint8Array);
+assert(value.buffer instanceof ArrayBuffer);
+
+assert.sameValue(value.length, 16);
+assert.sameValue(value.buffer.byteLength, 16);
+assert.sameValue(value.buffer.immutable, true);
+
+assert.compareArray(
```
ptomato, line 20: This will fail, no? Comments won't be stripped from the JS file?

ptomato (changes requested):
:+1: on the modifications to the test generator, everything else generally looks good.

test/language/import/import-bytes/bytes-from-js.js
styfle: The comments are only here for the human reader to make it clear that each byte represents a character in the js file that was imported, in this case:

```js
var foo = 'bar'
```

It should work just fine. Am I misunderstanding you?

ptomato: By my reading, the imported byte array should contain the UTF-8 encoding of
```
// Copyright (C) 2025 @styfle. All rights reserved.
var foo = 'bar'
```
and not
```
var foo = 'bar'
```
but this test expects the latter.

styfle: I understand now, thanks!

You're right, I pushed commit [fdc288c](https://github.com/tc39/test262/pull/4648/commits/fdc288ccd065fd52e4dd7356107836fe4bdabe7d) to remove it but then lint fails.

Is there a recommended way to disable lint on the FIXTURE files?

styfle: I think I figured it out. I pushed [ac5338b](https://github.com/tc39/test262/pull/4648/commits/ac5338b5168eaadc24d368a144630199801926f4) to disable license checks on files containing `_FIXTURE` in the name.

bakkot: Might be nicer to go the other way and include the bytes of the comment here instead of messing with the lint rules.

styfle: I considered that but I figured I would fix it for everyone since fixture files are not source code. 

It seems like this was the first fixture file to be js perhaps?

ptomato: On the contrary, fixture files are source code and this is not the first JS fixture file. JS fixture files are used extensively in other importing tests.

I think the best course of action is to keep the license header and include the bytes of the comment in the test, as annoying as that may be.

(You could say that it's ridiculous to license `var foo = 'bar'` and you would be right, but we don't want to get into the weeds of "is this JS file big enough to merit a license header". All JS files just need to have either the license header or the public domain declaration.)

styfle: Ok I reverted those commits and pushed a change to read all the bytes 👍 

Please take another look.

styfle:
@ptomato Sounds good. I added a note in [c7ec727](https://github.com/tc39/test262/pull/4648/commits/c7ec727c1c01d8f51a01669a1c0a05c533f25325).

Is there anyone else that needs to approve before this gets merged?

ptomato:
Not really. The rule of thumb I've been trying to stick to is: if it's small, or has 2 reviews already, or is urgent for plenary, I merge it; otherwise, I leave a couple of days to see if anyone else comments.

juice928:
👋 Hi, I'm an automated AI code review bot. I ran some checks on this PR and found **2 points that might be worth attention** (could be false positives, please use your judgment):

1. <strong>Some referenced fixture files appear to be missing from the commit</strong>
   - [test/language/import/import-bytes/bytes-from-empty.js:L11](https://github.com/tc39/test262/blob/c7ec727c1c01d8f51a01669a1c0a05c533f25325/test/language/import/import-bytes/bytes-from-empty.js#L11)
   - Impact: This might lead to module resolution errors as the tests expect specific `.bin`, `.json`, `.png`, and `.txt` files that aren't currently in the diff.
   - Suggestion: Please double-check if all necessary fixture files have been added to the commit.

2. <strong>The length assertion in `bytes-from-js.js` might need a quick adjustment</strong>
   - [test/language/import/import-bytes/bytes-from-js.js:L16-L17](https://github.com/tc39/test262/blob/c7ec727c1c01d8f51a01669a1c0a05c533f25325/test/language/import/import-bytes/bytes-from-js.js#L16-L17)
   - Impact: The current assertion for a length of 16 may cause the test to fail since the fixture content appears to be approximately 139 bytes.
   - Suggestion: Consider updating the assertion to match the actual size of the fixture file.

If you find these suggestions disruptive, you can reply "stop" , and I'll automatically skip this repository in the future.

ptomato:
> Some referenced fixture files appear to be missing from the commit

lies

> The length assertion in bytes-from-js.js might need a quick adjustment

Seems to be legit, @styfle PTAL?

ptomato:
> > Some referenced fixture files appear to be missing from the commit
> 
> lies
> 
> > The length assertion in bytes-from-js.js might need a quick adjustment
> 
> Seems to be legit, @styfle PTAL?

What is it with these LLM review bots that randomly pop up, leave review comments, and then after you reply to them the comment gets deleted?? It looks like I'm talking to myself.

In any case, the above was my reply to a review bot that made 2 suggestions, one of them nonsense, the other one was a legit comment about the length assertion in `bytes-from-js.js` not being updated to match the new length of the fixture file.

styfle:
@ptomato Good call! Updated in [73d12de](https://github.com/tc39/test262/pull/4648/commits/73d12de26e0f075da6eb44eb3f556d4d90d6c546)

styfle:
@ljharb @bakkot Any other feedback?

ptomato:
I think it's fine to merge it now, it's been up for several weeks since I approved it.

PR #4651, opened 2025-11-10
Fix invalid async test for await-using syntax

### LINKED ISSUE #4638: block-scope-syntax-await-using-declarations-mixed-with-without-initializer.js has conflicting metadata
"block-scope-syntax-await-using-declarations-mixed-with-without-initializer.js" has:
https://github.com/tc39/test262/blob/5c07b5224261aae30e54f69727a730956ff4ec41/test/language/statements/await-using/syntax/block-scope-syntax-await-using-declarations-mixed-with-without-initializer.js#L21-L24

An [async test](https://github.com/tc39/test262/blob/main/INTERPRETING.md#flags) needs to print "Test262:AsyncTestComplete" to be considered passing. This cannot happen when the test also expects a `SyntaxError`.

cc @rbuckton

rbuckton:
Fixes #4638


ptomato:
Thanks!

PR #4654, opened 2025-11-11
Intl Era Monthcode: DateTimeFormat eras and leap months

ptomato:
Tests that cover the areas where the Intl Era Monthcode proposal intersects with Intl.DateTimeFormat.

### DIFF
--- a/test/intl402/DateTimeFormat/constructor-options-calendar-islamic-fallback.js
+++ b/test/intl402/DateTimeFormat/constructor-options-calendar-islamic-fallback.js
@@ -15,7 +15,7 @@ info: |
     b. Set _resolvedCalendar_ to CanonicalizeUValue(*"ca"*, _fallbackCalendar_).
     c. If the ECMAScript implementation has a mechanism for reporting diagnostic warning messages, a warning should be issued.
   10. Set _dateTimeFormat_.[[Calendar]] to _resolvedCalendar_.
-locale: [en, en-u-ca-islamic]
+locale: [en]
 features: [Intl.Era-monthcode]
 ---*/
 


test/intl402/DateTimeFormat/prototype/formatToParts/era.js
```diff
@@ -0,0 +1,105 @@
...
+    assert.notSameValue(yearPart, undefined, `Format of ${calendar} ISO ${isoYear} should have year part`);
+  }
+
+  assert.sameValue(eras[0], eras[1], `${calendar} era does not change between negative to positive years`);
+}
+
+const noEraTests = [
+  "chinese",
+  "dangi",
+  "iso8601",
```
ptomato, line 90: Note, this test (for `iso8601` specifically) fails in all implementations, it produces "Anno Domini 2025" or "2025 Anno Domini". But I am pretty sure that as per the proposal, it should not include the era.

test/intl402/DateTimeFormat/prototype/formatToParts/era.js
```diff
@@ -0,0 +1,105 @@
...
+    eras.push(eraPart.value);
+
+    const format = formatter.format(date);
+    assert(format.includes(eraPart.value), `${format} should include ${eraPart.value} era`);
+
+    const yearPart = parts.find(({ type }) => type === "year");
+    assert.notSameValue(yearPart, undefined, `Format of ${calendar} ISO ${isoYear} should have year part`);
+  }
+
+  assert.sameValue(new Set(eras).size, eras.length, `${calendar} eras are unique`);
```
Ms2ger: Please add the value of `eras` to the error message

test/intl402/DateTimeFormat/prototype/formatToParts/era.js
FrankYFTang: The spec text https://tc39.es/proposal-intl-era-monthcode/ does not include any specification about the calendar "iso8601"

https://tc39.es/proposal-intl-era-monthcode/#sec-ecma402-calendar-types
mentioned 
"[ECMA-262](https://tc39.es/ecma402/#ecma262) describes calendar types, of which "iso8601" is required to be supported. This specification additionally requires ECMAScript implementations to support calendar types corresponding with those of the [Unicode Common Locale Data Repository (CLDR)](https://cldr.unicode.org/)."

https://tc39.es/proposal-intl-era-monthcode/#table-calendar-types


"iso8601" | ISO 8601 calendar. Fully specified in ECMA-262.
-- | --

https://tc39.es/proposal-intl-era-monthcode/#sec-temporal-canonicalizeeraincalendar
The abstract operation CanonicalizeEraInCalendar takes arguments calendar (a [calendar type](https://tc39.es/proposal-intl-era-monthcode/#sec-ecma402-calendar-types) that is not "iso8601") and era (a String) and returns a String or undefined.

https://tc39.es/proposal-intl-era-monthcode/#sec-temporal-calendarhasmidyeareras
The abstract operation CalendarHasMidYearEras takes argument calendar (a [calendar type](https://tc39.es/proposal-intl-era-monthcode/#sec-ecma402-calendar-types) that is not "iso8601") and returns a Boolean.

and many more. Notice all the AO explicitly specify these AO are for calendar type which  is not "iso8601"

I do not think the specification is related to "iso8601". I think we should remove the verification of "iso8601" out of this test.

PR #4660, opened 2025-11-13
Intl Era Monthcode: Era code updates for CLDR 48

ptomato:
Coptic, Ethiopic, and Hijri calendars got era code changes in CLDR 48.

As per the maintainers policy on the other CLDR 48 changes, this should be merged about a month after CLDR 48 released, so in late November, to give implementations a chance to catch up.


ptomato:
It's more than 1 month after the CLDR 48 release, so this is free to go in when it's reviewed.

PR #4673, opened 2025-11-17
Remove non-canonical values from Intl.supportedValuesOf("calendar") test

### LINKED ISSUE #4639: calendars-required-by-intl-era-monthcode.js requires non-canonical "ethiopic-amete-alem" calendar to be present
The test requires both "ethiopic-amete-alem" and "ethioaa" to be present, but "ethiopic-amete-alem" is just a (non-canonical) alias for "ethioaa".

jessealama:
Fixes #4639.


ptomato:
> Should islamicc (non-canonical alias of islamic-civil) be removed as well?

Yes :smile: 

Was there anywhere else that the aliases are included in a list and shouldn't be? Probably a quick `git grep ethiopic-amete-alem` and `git grep islamicc` will show, but depending on the test, there may be cases where they are correctly included.

jessealama:
OK, I'll remove `islamicc`. That's the only other one I can see.

PR #4675, opened 2025-11-17
Update chinese calendar reference day

ben-allen:
Updated reference year for `"M11L"` in Chinese calendar to 2033 (which is in our supported range) rather than a date in the 1500s (before supported range).

Additionally removed/updated comments related to what leap months exist.


ptomato (approved):
Thanks!

PR #4678, opened 2025-11-18
Intl Era Monthcode: Basic arithmetic (add, subtract, since, until) for Gregorian-like calendars

ptomato:
Basic addition and difference tests for the four non-ISO calendars that use ISO-like months and years (`buddhist`, `gregory`, `japanese`, `roc`).

Many of the tests are similar, so I recommend reviewing only the `PlainDate/prototype/add/` ones in detail.

### DIFF
--- a/test/intl402/Temporal/PlainDate/prototype/since/basic-buddhist.js
+++ b/test/intl402/Temporal/PlainDate/prototype/since/basic-buddhist.js
@@ -1,0 +1,342 @@
+// Copyright (C) 2025 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.plaindate.prototype.since
+description: >
+  Check various basic calculations not involving leap years or constraining
+  (buddhist calendar)
+includes: [temporalHelpers.js]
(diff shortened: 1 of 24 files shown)


test/intl402/Temporal/PlainDate/prototype/add/basic-buddhist.js
```diff
@@ -0,0 +1,302 @@
...
+TemporalHelpers.assertPlainDate(
+  date25640716.add(months5),
+  2564, 12, "M12", 16, "add 5mo with result in the same year",
+  "be", 2564);
+TemporalHelpers.assertPlainDate(
+  Temporal.PlainDate.from({ year: 2564, monthCode: "M08", day: 16, calendar }).add(months5),
+  2565, 1, "M01", 16, "add 5mo with result in the next year",
+  "be", 2565);
+TemporalHelpers.assertPlainDate(
+  Temporal.PlainDate.from({ year: 2562, monthCode: "M10", day: 01, calendar }).add(months5),
```
Ms2ger: I think we avoid octal `01`

ptomato: Glad you caught that.

PR #4681, opened 2025-11-18
Fix constructor-options-calendar-islamic-fallback.js

FrankYFTang:
The defintion of AvailableCalendars in https://tc39.es/proposal-intl-era-monthcode/#sup-availablecalendars is "The List must include the Calendar Type value of every row of Table 1, except the header row." It does said it ONLY include those listed in that Table. so the return list may also include "islamic" and "islamic-rgsa" and therefore the test is incorrect.

so we should get availableCalendars from calling Intl.supportedValuesOf("calendar") instead

@ben-allen @sffc


ptomato:
I need to think about this, if it's still possible to write a meaningful test for this feature at all.

ptomato:
As per the discussion in yesterday's TG2, it's my understanding that this test is correct as-is.

PR #4694, opened 2025-11-20
Add tests for two-digit date when day is over 31

Gumichocopengin8:
The `Date` object allows two-digit years, but Chrome and Firefox currently return `NaN` when the year is between `50` and `99` and the day exceeds `31`. WebKit also has an open [PR](https://github.com/WebKit/WebKit/pull/47652)  to adopt the same behavior.
This patch adds test cases to confirm that two-digit dates return NaN for the compatibility across browsers.

### DIFF
--- a/test/staging/sm/Date/two-digit-years.js
+++ b/test/staging/sm/Date/two-digit-years.js
@@ -64,3 +64,16 @@ for (let year of Array(1000).keys()) {
 
 assert.sameValue(new Date("may 1999 1999").getTime(), new Date(NaN).getTime());
 assert.sameValue(new Date("may 0 0").getTime(), new Date(NaN).getTime());
+
+for (let year = 50; year <= 99; ++year) {
+  for (let month = 0; month < 12; ++month) {
+    for (let day = 1; day <= 100; ++day) {
+        const date = `${year}/${month + 1}/${day}`;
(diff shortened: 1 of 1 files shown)


Ms2ger:
Can you point to the specification for this behavior?

Gumichocopengin8:
> It's not specified anywhere, support for `Y/M/D` formats is implementation defined. This whole test is a poor fit for test262 since it assumes SpiderMonkey-specific implementation details.
> 
> > For the sake of cross compatibility with other implementations we implement date parsing heuristics which support single and double digit years.

I was also wondering if the test file was applicable in Test262 because the behavior is not specified in the spec. I made a PR in case the test file is legit for Test262.
So we should remove the file instead of adding more tests, right?

ptomato:
> So we should remove the file instead of adding more tests, right?

That depends on whether the current contents of the file is correctly specified. If not, then maybe we should try to re-contribute this test back to SpiderMonkey...

Gumichocopengin8:
As far as I see in https://tc39.es/ecma262/#sec-date.parse, it is not specified for two-digit years. I'll close this PR.

ptomato:
Even though the contribution didn't land, thank you for the time you were able to spend on it.

PR #4701, opened 2025-11-22
Validate WEB_FEATURES.yml

jugglinmike:
For every pull request, verify that the `WEB_FEATURES.yml` file is internally consistent and that the pull request does not reduce the number of tests associated with a given classifier.

Failing for any reduction in the number of matched tests is intended to alert contributors during file renaming operations. This failure will need to be ignored when tests are simply deleted, and this could potentially become a distraction for maintainers. However, test deletion has historically been so rare that spurious failures are not expected to be common enough to substantively interfere with routine maintenance.


ptomato:
> However, test deletion has historically been so rare that spurious failures are not expected to be common enough to substantively interfere with routine maintenance.

I think the main place where this would come up is when moving tests out of staging, into the main tree. They might get merged into existing tests. I don't actually know if the classification applies to the staging folder in the first place, so maybe this isn't a problem.

jugglinmike:
> I think the main place where this would come up is when moving tests out of staging, into the main tree. They might get merged into existing tests. I don't actually know if the classification applies to the staging folder in the first place, so maybe this isn't a problem.

That's a good call-out! We haven't classified tests in staging, so fortunately those kinds of moves won't be an issue.

tools/web-features/README.md
```diff
@@ -84,8 +84,22 @@ For details on the design considerations, please see [GitHub issue
...
+
+### `lint.py`
+
+This script validates basic expectations about the contents of the
+`WEB_FEATURES.yml` (e.g. that the file conforms to the schema and that every
+pattern impacts the set of matched test files). If invoked with the argument
+`--manifest`, this utility will write a "manifest" of all web-features and
+their associated test file names.
+
+### `check-coverage.py'
```
gibson042:
```suggestion
### `check-coverage.py`
```

tools/web-features/test/run.py
```diff
@@ -0,0 +1,236 @@
+#!/usr/bin/env python
+# Copyright (C) 2025 the V8 project authors. All rights reserved.
+# This code is governed by the BSD license found in the LICENSE file.
+
+import shutil, subprocess, sys, os, unittest, tempfile
+import json
+
+testDir = os.path.dirname(os.path.abspath(__file__))
+OUT_DIR = os.path.join(testDir, 'out')
```
gibson042, line 9: Is this directory ignored by git?

.github/workflows/checks.yml
```diff
@@ -54,6 +54,58 @@ jobs:
...
+          clean: false
+
+      - name: Generate manifest on base branch
+        run: |
+          ./tools/web-features-stable/lint.py --manifest WEB_FEATURES-manifest-base.json
+
+      - name: debug
+        run: ls -lah
+
+      # Checking out the original branch ensure that if the patch under review
```
gibson042:
```suggestion
      # Checking out the original branch ensures that if the patch under review
```

jugglinmike:
Thanks for the review, @gibson042! In addition to adding the file to the project's `.gitignore` exclusions as per your recommendation, I also inserted a second defensive directory removal so that the automated tests can gracefully recover from that state.

PR #4702, opened 2025-11-22
Publish web-features manifest files

jugglinmike:
For every push to the `main` branch, generate a web-features manifest
file and publish it in compressed form using the GitHub "Release"
feature.

---

As discussed during a recent maintainers' call. This patch depends on gh-4701


jugglinmike:
Thanks, @ptomato! After incorporating suggestions from @gibson042 in gh-4701, the base commit here was out-of-date. I never know what action is going to invalidate an approving review on GitHub, so I avoided rebasing to address this. Instead, I just reverted the old base commit since this patch only introduces a new file on top of it.

foolip:
Nice, I see the releases showing up in https://github.com/tc39/test262/releases, thanks @jugglinmike!

@jcscottiii FYI, I expect this will be relevant for webstatus.dev.

PR #4703, opened 2025-11-24
Temporal: Add order-of-operations tests for toPlainDate

catamorphism:
Add order-of-operations tests for toPlainDate for PlainMonthDay and PlainYearMonth


ptomato (approved):
Thanks!

From what I can tell we are still missing a few such tests, I've opened  https://github.com/tc39/proposal-temporal/issues/3189 for a future follow up

PR #4709, opened 2025-11-25
Add PlainMonthDay month code tests for non-ISO calendars

jessealama:
Tests for `Temporal.PlainMonthDay.from()` verifying that all valid month codes are accepted for various calendar systems.

### DIFF
--- a/test/intl402/Temporal/PlainMonthDay/from/buddhist-month-codes.js
+++ b/test/intl402/Temporal/PlainMonthDay/from/buddhist-month-codes.js
@@ -1,0 +1,36 @@
+// Copyright (C) 2025 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.plainmonthday.from
+description: PlainMonthDay can be created for all month codes (M01-M12) in Buddhist calendar
+features: [Temporal, Intl.Era-monthcode]
+includes: [temporalHelpers.js]
+---*/
+
+// Test that all month codes M01-M12 are valid for the Buddhist calendar
+// The Buddhist calendar follows the Gregorian calendar structure
+
+const calendar = "buddhist";
+
+for (const { monthCode, daysInMonth } of TemporalHelpers.ISOMonths) {
+  // Test creation with monthCode and day 1
+  const pmd = Temporal.PlainMonthDay.from({ calendar, monthCode, day: 1 });
+  TemporalHelpers.assertPlainMonthDay(pmd, monthCode, 1, `monthCode ${monthCode} should be preserved`);
+
+  // Test with maximum day value for this month (minimum for PlainMonthDay)
+  const pmdMax = Temporal.PlainMonthDay.from({ calendar, monthCode, day: daysInMonth });
+  TemporalHelpers.assertPlainMonthDay(pmdMax, monthCode, daysInMonth, `${monthCode} with day ${daysInMonth} should be valid`);
+
+  // Test constrain overflow
+  const constrained = Temporal.PlainMonthDay.from(
+    { calendar, monthCode, day: daysInMonth + 1 },
+    { overflow: "constrain" }
+  );
+  TemporalHelpers.assertPlainMonthDay(constrained, monthCode, daysInMonth, `day ${daysInMonth + 1} should be constrained to ${daysInMonth} for ${monthCode}`);
+
+  // Test reject overflow
+  assert.throws(RangeError, () => {
+    Temporal.PlainMonthDay.from({ calendar, monthCode, day: daysInMonth + 1 }, { overflow: "reject" });
+  }, `${monthCode} with day ${daysInMonth + 1} should throw with reject overflow`);
+}
--- a/test/intl402/Temporal/PlainMonthDay/from/chinese-leap-month-codes-common.js
+++ b/test/intl402/Temporal/PlainMonthDay/from/chinese-leap-month-codes-common.js
@@ -1,0 +1,53 @@
+// Copyright (C) 2025 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.plainmonthday.from
+description: PlainMonthDay can be created for common leap month codes in Chinese calendar
+features: [Temporal, Intl.Era-monthcode]
+---*/
+
+// Test common leap months in the Chinese calendar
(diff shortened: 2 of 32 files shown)


test/intl402/Temporal/PlainMonthDay/from/chinese-leap-month-codes-common.js
```diff
@@ -0,0 +1,51 @@
...
+// The Chinese calendar is a lunisolar calendar where leap months are inserted
+// to keep the lunar year synchronized with the solar year.
+// Common leap months (occurring more frequently in the astronomical cycle):
+// M02L, M03L, M04L, M05L, M06L, M07L, M08L
+//
+// The distribution of leap months follows astronomical calculations.
+//
+// Note: M02L and M08L are temporarily omitted from this test because while these
+// leap months can have 30 days, there isn't a year in the supported range where
+// they do, and it's not yet well-defined what reference ISO year should be used.
```
ptomato: Future-proof the comment a bit:
```suggestion
// Note: M01L, M02L and M08L through M12L are temporarily omitted from this
// test because while any leap month can have 30 days, there isn't a year in the
// supported range where these months do, and it's not yet well-defined what
// reference ISO year should be used.
// See https://github.com/tc39/proposal-intl-era-monthcode/issues/60
```
(also for `dangi-leap-month-codes-common.js`)

test/intl402/Temporal/PlainMonthDay/from/chinese-month-codes.js
```diff
@@ -0,0 +1,43 @@
...
+for (const monthCode of monthCodes) {
+  // Test creation with monthCode
+  const pmd = Temporal.PlainMonthDay.from({ calendar, monthCode, day: 1 });
+  assert.sameValue(pmd.monthCode, monthCode, `monthCode ${monthCode} should be preserved`);
+  assert.sameValue(pmd.day, 1, "day should be 1");
+
+  // Test with day 29 (minimum for PlainMonthDay in lunisolar calendars)
+  const pmd29 = Temporal.PlainMonthDay.from({ calendar, monthCode, day: 29 });
+  assert.sameValue(pmd29.monthCode, monthCode, `${monthCode} with day 29 should be valid`);
+  assert.sameValue(pmd29.day, 29, `day should be 29 for ${monthCode}`);
```
ptomato: Chinese calendar months can have 30 days although not in every year, so this should be tested with day 30

test/intl402/Temporal/PlainMonthDay/from/chinese-month-codes.js
```diff
@@ -0,0 +1,43 @@
...
+
+  // Test constrain overflow - Chinese months vary from 29-30 days
+  const constrained = Temporal.PlainMonthDay.from(
+    { calendar, monthCode, day: 31 },
+    { overflow: "constrain" }
+  );
+  assert.sameValue(constrained.monthCode, monthCode, `${monthCode} should be preserved with constrain`);
+  assert(constrained.day <= 30, `day 31 should be constrained to 30 or less for ${monthCode}`);
```
ptomato:
```suggestion
  assert.sameValue(constrained.day, 30, `day 31 should be constrained to 30 for ${monthCode}`);
```

(Same in `dangi-month-codes.js`)

test/intl402/Temporal/PlainMonthDay/from/buddhist-month-codes.js
```diff
@@ -0,0 +1,37 @@
...
+  const pmdMax = Temporal.PlainMonthDay.from({ calendar, monthCode, day: daysInMonth });
+  TemporalHelpers.assertPlainMonthDay(pmdMax, monthCode, daysInMonth, `${monthCode} with day ${daysInMonth} should be valid`);
+
+  // Test constrain overflow
+  const constrained = Temporal.PlainMonthDay.from(
+    { calendar, monthCode, day: daysInMonth + 1 },
+    { overflow: "constrain" }
+  );
+  assert.sameValue(constrained.monthCode, monthCode, `${monthCode} should be preserved with constrain`);
+  assert.sameValue(constrained.day, daysInMonth, `day ${daysInMonth + 1} should be constrained to ${daysInMonth} for ${monthCode}`);
```
ptomato:
```suggestion
  TemporalHelpers.assertPlainMonthDay(constrained, monthCode, daysInMonth, `day ${daysInMonth + 1} should be constrained to ${daysInMonth} for ${monthCode}`);
```

Please also look for other places where `TemporalHelpers.assertPlainMonthDay()` can be used instead of individual assertions for `monthCode` and `day`. (At least in the other ISO-like calendars. However, I'd avoid it at least in `chinese` and `dangi` until the issue with reference years is solved.)

ptomato (changes requested):
Thanks for this. A few minor revisions and then it should be ready to go.

test/intl402/Temporal/PlainMonthDay/from/buddhist-month-codes.js
jessealama: I tried applying this to `gregory`, `japanese`, and `roc` but get `referenceISOYear` errors (1971 vs 1972) when testing locally with the polyfill. Any guidance?

ptomato: May be a bug with the reference year detection in the polyfill. I'm pretty sure the spec says those calendars should just always use 1972 by definition. I can investigate this.

ptomato: Yes, this is a polyfill bug. The year gets bumped to 1971 but only in the constrain case. This equality should hold: `Temporal.PlainMonthDay.from({ calendar: 'gregory', monthCode: 'M12', day: 31 }).toString() === Temporal.PlainMonthDay.from({ calendar: 'gregory', monthCode: 'M12', day: 32 }).toString()` (but it doesn't because the `day: 32` instance gets the correct day of 31 but the wrong reference year)

ptomato:
https://github.com/tc39/test262/pull/4709#discussion_r2562017793 still needs to be applied, otherwise good to go

ptomato:
I can fix that up since it's the only thing blocking.

ptomato (approved):

test/intl402/Temporal/PlainMonthDay/from/buddhist-month-codes.js
ptomato: See https://github.com/tc39/proposal-temporal/pull/3190

PR #4711, opened 2025-11-26
Temporal: Basic date difference tests for most non-ISO calendars

catamorphism:
Covers Chinese, Dangi, Orthodox, Hebrew, Indian, Hijri, and Persian calendars.

### DIFF
--- a/test/intl402/Temporal/PlainDate/prototype/since/basic-islamic-umalqura.js
+++ b/test/intl402/Temporal/PlainDate/prototype/since/basic-islamic-umalqura.js
@@ -5,7 +5,7 @@
 esid: sec-temporal.plaindate.prototype.since
 description: >
   Check various basic calculations not involving leap years or constraining
-  (islamic-umalqura calendar)
+  (Islamic umalqura calendar)
 includes: [temporalHelpers.js]
 features: [Temporal, Intl.Era-monthcode]
 ---*/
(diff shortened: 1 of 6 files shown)


test/intl402/Temporal/PlainDate/prototype/since/basic-ethioaa.js
ptomato, line 1: These dates are all 300 years ago, on the face of it there's nothing wrong with that since the calendar is proleptic, but I think it makes more sense to test recent or near-future dates.

test/intl402/Temporal/PlainDate/prototype/since/basic-islamic-civil.js
ptomato, line 1: Likewise, the dates in the Hijri calendar files are all ~75 years ago.

test/intl402/Temporal/PlainDate/prototype/since/basic-persian.js
ptomato, line 1: And these are centered around 40 years ago

catamorphism:
> These make sense as a representative sample of diff arithmetic but I'd prefer that the dates are centered around years in the very recent past, to align with the Gregorian-like ones which test ranges from 1960-2031.

@ptomato Everything should be recent now, as of [92db90b](https://github.com/tc39/test262/pull/4711/commits/92db90b11aba2c5e96879579688e8b26789f9e21)!

PR #4712, opened 2025-11-26
Temporal: Add tests for order of operations on property bag reads

catamorphism:
Add tests for order of operations on property bag reads for the following methods:

* PlainDate.prototype.toPlainDateTime
* PlainDateTime.prototype.withPlainTime
* ZonedDateTime.prototype.getTimeZoneTransition

Closes https://github.com/tc39/proposal-temporal/issues/3189


ptomato (approved):
Thanks!

PR #4721, opened 2025-11-27
Update old Temporal tests for CLDR 48

anba:
CLDR 48 changed some era names, update the tests accordingly.


test/staging/Intl402/Temporal/old/non-iso-calendars-coptic.js
```diff
@@ -26,14 +26,14 @@ function compareFormatToPartsSnapshot(isoString, expectedComponents) {
...
+  era: "AM",
   month: 4,
   day: 22,
 });
 
 compareFormatToPartsSnapshot("0001-01-01T00:00Z", {
-  year: 284,
-  era: "ERA0",
+  year: -283,
+  era: "AM",
```
anba, line 36: NB: This requires a patched ICU to work around <https://unicode-org.atlassian.net/browse/ICU-23277>.

PR #4722, opened 2025-11-27
Don't expect era part when formatting MonthDay

anba:
`era` options are ignored when formatting `Temporal.PlainMonthDay` objects. See GetDateTimeFormat, step 12.b.


anba:
The spec polyfill returns `"11/4 A"`, but this doesn't match my reading of the spec.

PR #4708, opened 2025-11-25
Add CalendarResolveFields error ordering tests

jessealama:
Verify that CalendarResolveFields throws TypeError for missing required fields before throwing RangeError for invalid field values.

Covers:
- `from()` methods for PlainDate, PlainDateTime, PlainMonthDay, PlainYearMonth, ZonedDateTime (ISO and non-ISO calendars)
- `with()` methods for Japanese calendar (PlainDate, PlainDateTime, PlainYearMonth, ZonedDateTime)

PR #4723, opened 2025-11-27
Test that resolved MonthDay can have a different ordinal month

anba:
PR #4708 covered the second issue reported in
<https://bugzilla.mozilla.org/show_bug.cgi?id=1999918>. Extend the test cases to also cover the first issue reported there.


ptomato (approved):
I see, thanks for adding this!

PR #4727, opened 2025-11-28
Intl Era Monthcode: Move tests for extreme dates out of staging

ptomato:
These tests cover creating dates at the edges of the range, and at the edges of the non-approximated range for calendars where that applies.


Ms2ger (approved):

ptomato:
I think I might have made a mistake in the `ethiopic` dates ... I had a comment there reminding myself to check them, and I forgot to do that before submitting the PR.

ptomato:
→ #4728

PR #4730, opened 2025-11-28
Temporal: Move test for Indian calendar out of staging

catamorphism:
Also split into separate tests for `from` and `with`, and duplicate for PlainDateTime and ZonedDateTime as well as PlainDate.


ptomato (changes requested):
:+1: on the `with/leap-year-indian.js` changes

The roundtripping in `from/basic-indian.js` isn't really covered in the staging test, so I'd prefer to save that for another PR. I'm intending to build a calendar-agnostic roundtripping test based on some of the other staging tests, so let's sync about that offline.

catamorphism:
> 👍 on the `with/leap-year-indian.js` changes
> 
> The roundtripping in `from/basic-indian.js` isn't really covered in the staging test, so I'd prefer to save that for another PR. I'm intending to build a calendar-agnostic roundtripping test based on some of the other staging tests, so let's sync about that offline.

@ptomato Removed in [2823777](https://github.com/tc39/test262/pull/4730/commits/2823777453fbd628f48cd7b0487a957b23523f25)

ptomato:
I guess I don't get what the remaining code in `from/basic-indian.js` is testing? It looks more like a test for `with()` but as far as I can tell the `with()` call is a no-op?

catamorphism:
> I guess I don't get what the remaining code in `from/basic-indian.js` is testing? It looks more like a test for `with()` but as far as I can tell the `with()` call is a no-op?

@ptomato No, the `with()` was trying to work around something in the test generator. I was able to get rid of it in [805c504](https://github.com/tc39/test262/pull/4730/commits/805c504d322a12994eb0d9fd2983b66df8c22d0e). All that's being tested is the string parsing in `from`, but maybe that doesn't need to be covered in a calendar-specific way?

test/intl402/Temporal/PlainDate/from/basic-indian.js
```diff
@@ -0,0 +1,42 @@
...
+};
+const dates = {
+  leapYear: Temporal.PlainDate.from("2004-03-21[+00:00][u-ca=indian]"),
+  nonLeapYear: Temporal.PlainDate.from("2005-03-22[+00:00][u-ca=indian]")
+};
+for (var [name, result] of Object.entries(cases)) {
+  const date = dates[name];
+
+  TemporalHelpers.assertPlainDate(date, result.year, result.month, result.monthCode, result.day, name, result.era, result.eraYear);
+}
```
ptomato: OK, I get what this is doing now.
I'd rewrite it to be a bit more straightforward though:
```js
const cases = [
  ["2004-03-21[u-ca=indian]", 1926, 1, "M01", 1, "first day of leap year"],
  ["2005-03-22[u-ca=indian]", 1927, 1, "M01", 1, "first day of common year"],
];
for (const [isoString, year, month, monthCode, day, descr] of cases) {
  const date = Temporal.PlainDate.from(isoString);
  TemporalHelpers.assertPlainDate(date, year, month, monthCode, day, descr, "shaka", year);
}
```

ptomato: (and rename the test to `from-iso-string-indian.js` or something)

catamorphism: Done in [a842ee8](https://github.com/tc39/test262/pull/4730/commits/a842ee802e7749574d2a465e55517e4ff701f92c)

ptomato (approved):
Thanks!

PR #4733, opened 2025-11-29
Temporal: Test that "islamic" and "islamic-rgsa" calendars are unsupported

catamorphism:
And remove the equivalent staging tests.


ptomato (approved):
:+1: (when the linter is satisfied)

ptomato:
(I don't get what the linter's problem is though!)

catamorphism:
> (I don't get what the linter's problem is though!)

@ptomato It didn't like the quotation marks in the description field (there's probably a way to escape them, but they're not really necessary in this case so I just removed them)

PR #4740, opened 2025-12-01
Split out tests from test/staging/Intl402/Temporal/old/non-iso-calendars-indian.js

catamorphism:
* Add more tests for add/subtract/since/until
* Add tests for inLeapYear, monthsInYear, daysInMonth properties
* Add tests checking that leap months are disallowed
* Add tests for wrapping at the end of the month in with


test/intl402/Temporal/PlainDate/prototype/with/wrapping-at-end-of-month-indian.js
ptomato, line 1: I believe this and similar files are already covered in `constrain-day-indian.js`.

catamorphism: Removed and force-pushed

PR #4742, opened 2025-12-01
Split out tests from test/staging/Intl402/Temporal/old/non-iso-calendars-islamic*.js

catamorphism:
* Add more tests for add/subtract/since/until
* Add tests for inLeapYear, monthsInYear, daysInMonth properties
* Add tests checking that leap months are disallowed
* Add tests for wrapping at the end of the month in with

Remove staging tests for islamicc (alias) and islamic and islamic-rgsa (unsupported) calendars.


test/intl402/Temporal/PlainDate/prototype/with/wrapping-at-end-of-month-islamic-civil.js
ptomato, line 1: Similar to the other PRs, I believe the wrapping-at-end-of-month files are already covered in the existing constrain-day tests.

ptomato (approved):
Same, just remove the already-covered files

test/intl402/Temporal/PlainDate/prototype/with/wrapping-at-end-of-month-islamic-civil.js
catamorphism: Removed files and force-pushed

PR #4744, opened 2025-12-02
Split out tests from test/staging/Intl402/Temporal/old/non-iso-calendars-iso8601.js

catamorphism:
* Duplicate basic-arithmetic tests across types, then add new cases
* Add `with` tests for different types
* Duplicate since/until tests across types, then add new cases
* Test addition/subtraction at month boundary
* Add a test checking that there are no leap months in iso8601 calendar
* Add tests for `with` wrapping at end of month


Ms2ger:
There seems to be an overlap with #4745

catamorphism:
> There seems to be an overlap with #4745

@Ms2ger Fixed in #4745

test/built-ins/Temporal/PlainDate/prototype/with/wrapping-at-end-of-month.js
ptomato, line 1: We should name this `constrain-day.js` to match the other tests that test a similar feature. (Might be better to copy `test/intl402/Temporal/PlainDate/prototype/with/constrain-day-gregory.js` which is more comprehensive, keeping the same skeleton but removing the calendar and era stuff.)

test/built-ins/Temporal/PlainDate/prototype/with/basic-year-month-day.js
ptomato, line 1: Let's rename this back to `basic.js`, it's still the basic functionality of `with()`

catamorphism: I gave it a different name because PlainDateTime and ZonedDateTime have existing tests named `basic.js`, and I didn't want to combine these tests with those files.

PR #4747, opened 2025-12-02
[Temporal] Fix strict mode errors in recently-added tests

catamorphism:
Remove accidental octal literals and add variable declarations.


Ms2ger:
Fixed in #4746

PR #4748, opened 2025-12-02
Intl Era Monthcode: Move date roundtripping tests out of staging

ptomato:
Several of the staging tests involved roundtripping the calendar equivalents of the dates 2000-01-01 and 0001-01-01, as well as a few incidental dates alongside those.

Add tests with a comprehensive loop through all calendars and remove the staging tests. With this format, it's easy to continue to add cases in the future if any further bugs or regressions are found.


ptomato:
May need rebasing after #4744, #4749, #4750 are merged - I've removed a couple of stanzas from those files, but in those PRs the files are removed altogether.

PR #4754, opened 2025-12-02
Temporal: Consolidate non-ISO month boundary tests

catamorphism:


test/intl402/Temporal/PlainDate/prototype/add/month-boundary-non-iso.js
```diff
@@ -0,0 +1,137 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.plaindate.prototype.add
+description: Addition around end of month in the Buddhist calendar
+features: [Temporal, Intl.Era-monthcode]
+includes: [temporalHelpers.js]
+---*/
+
+const calendar = "buddhist";
```
ptomato:
```suggestion
```

test/built-ins/Temporal/PlainDate/prototype/add/month-boundary.js
ptomato, line 1: We still need the ISO ones separately, no?

test/intl402/Temporal/PlainDate/prototype/add/month-boundary-non-iso.js
```diff
@@ -0,0 +1,137 @@
...
+  Temporal.PlainDate.from({ year: 2000, monthCode: "M12", day: 1, calendar: "japanese" }),
+];
+
+const datesPersian = [
+  Temporal.PlainDate.from({ year: 1378, monthCode: "M01", day: 1, calendar: "persian" }),
+];
+
+const datesRoc = [
+  Temporal.PlainDate.from({ year: 90, monthCode: "M01", day: 1, calendar: "roc" }),
+];
```
ptomato, line 94: Somehow I'd gotten the impression that all the existing tests were all starting on the calendar equivalent of ISO date 2000-01-01, but looking at this, I guess I was mistaken. Is there anything that relates all the dates? (except for the extra `chinese` and `dangi` dates of course)

ptomato (commented):
Looks like a good improvement. Didn't have time to go through thoroughly, but here's what jumped out at me.

test/intl402/Temporal/PlainDate/prototype/add/month-boundary-non-iso.js
catamorphism: Fixed separately

test/built-ins/Temporal/PlainDate/prototype/add/month-boundary.js
catamorphism: Right. Restored those.

test/intl402/Temporal/PlainDate/prototype/add/month-boundary-non-iso.js
catamorphism: No, it was arbitrary. I changed the dates to be equivalent to ISO 2000-01-01 (except for the extra dates).

ptomato: Sorry for the back-and-forth. I should have started out by asking the second part of my question, which is, if the dates are all related, can we avoid writing them all out? Like in `intl402/Temporal/PlainDate/from/roundtrip-from-iso.js`, we just loop through all the calendars and create
```js
Temporal.PlainDate.from(`2000-01-01[u-ca=${calendar}]`)
```
and then add a few additional cases down below (we could do the same for the extra `chinese` and `dangi` dates here.)

Since the test is just verifying an invariant for addition and subtraction, it shouldn't matter how exactly we are creating the instances, and we already verify that those dates can be created from a property bag in `roundtrip-from-propertybag.js`.

ptomato: I guess that won't work. The testing loop always assumes that the input date is the first of a month, otherwise the loop that advances the "yesterday" dates past the month boundary is incorrect. So maybe we should just leave the tests as they are, and close this PR.

PR #4757, opened 2025-12-03
Temporal: Refactor tests for absence of leap months

catamorphism:
Incorporate all non-lunisolar calendars into a single test.


test/built-ins/Temporal/PlainDate/prototype/monthCode/no-leap-months.js
```diff
@@ -1,20 +0,0 @@
-// Copyright (C) 2025 Igalia, S.L. All rights reserved.
-// This code is governed by the BSD license found in the LICENSE file.
-
-/*---
-esid: sec-temporal.plaindate.prototype.monthcode
-description: iso8601 calendar does not have leap months
```
Ms2ger, line 6: I don't think the new common test covers iso8601 (and probably that shouldn't be tested in intl402/ anyway)

catamorphism: Oops, restored those files.

PR #4759, opened 2025-12-03
Intl Era Monthcode: Fix Persian leap year discrepancy

ptomato:
I moved this test in from staging, but after doing more of a deep dive on it I believe it is incorrect. There are two popular algorithms for calculating leap years in the Persian calendar: a 33-year cycle and a 2820-year cycle. The 2820-year cycle happens to be incorrect in the spring equinox of ISO year 2025. (ref: https://www.tondering.dk/claus/cal/persian.php#leap)

The correct dates are provided by the Iranian calendar authority: https://calendar.ut.ac.ir/documents/2139738/7092644/Kabise+Shamsi+1206-1498.pdf
Unofficial English translation: https://github.com/roozbehp/persiancalendar/blob/main/kabise.txt
AP 1403 is a leap year, ending on 2025-03-20, so 2025-04-19 falls on 30 Farvardin of AP 1404, not 31 Farvardin.

At the time Anba wrote the staging test, ICU4X used the 2820-year cycle and has now switched to the 33-year cycle, which ICU4C already used. (See discussion in https://github.com/unicode-org/icu4x/issues/4713).


ptomato:
cc @Ms2ger @catamorphism

PR #4760, opened 2025-12-03
Temporal: Update duration tests to cover the largest possible ms, µs, and ns values

catamorphism:
Closes https://github.com/tc39/proposal-temporal/issues/3106


catamorphism:
Re: @ptomato's [comment](https://github.com/tc39/proposal-temporal/issues/3106#issuecomment-2887470048) on the issue, "We do need to update tests such as https://github.com/tc39/test262/blob/main/test/built-ins/Temporal/Duration/prototype/round/out-of-range-when-converting-from-normalized-duration.js " -- I searched for tests mentioning `MAX_SAFE_INTEGER` and winnowed down from there, but it's possible I missed some relevant tests.

PR #4761, opened 2025-12-04
Intl era monthcode cleanups

ptomato:
Various fixups and cleanups of the Intl Era Monthcode tests that have been merged recently. Best reviewed commit by commit.


harness/temporalHelpers.js
```diff
@@ -881,18 +881,18 @@ var TemporalHelpers = {
...
+        },
+      });
+      Object.defineProperty(temporalObject, "calendarId", {
+        get() {
+          throw new Test262Error("should not get 'calendarId' property");
         },
       });
 
       func(temporalObject);
-      assert.compareArray(actual, expected, "calendar getter not called");
```
Ms2ger, line 895: Pedantically, this now relies on `func` propagating the exception where it didn't before. I'm not sure if that's a blocker.

Ms2ger (approved):

harness/temporalHelpers.js
ptomato: Eh ... this gets called with `func`s such as
```js
(temporalObject) => {
  const result = Temporal.PlainDate.from({ year: 2000, month: 5, day: 2, calendar: temporalObject });
}
```
I think it's fine. (At some point I'd like to inline these, but that point is not now.)

ptomato:
Thanks for the reviews!

PR #4768, opened 2025-12-08
tests for Iterator.prototype.join

### LINKED ISSUE #4697: Test plan for Iterator.prototype.join
The [iterator join proposal](https://github.com/bakkot/proposal-iterator-join) is at stage 2.7 (not to be confused with the [joint iteration proposal](https://github.com/tc39/proposal-joint-iteration), newly at stage 3).

I came up with this list of tests by thinking about it, and then [looking at the tests for Array.prototype.join](https://github.com/tc39/test262/tree/3aa75310772673da078ca1d96fe3b75da0e21bd0/test/built-ins/Array/prototype/join).

- [ ] The usual boilerplate around name / length / enumerability / etc
- [ ] Expected results when joining iterator length 0, 1, 2, 3 with no provided separator
- [ ] As above but with a nonempty string as separator
- [ ] As above but with empty string as separator
- [ ] Throws if the receiver is not an object
- [ ] Coerces the separator argument to string, including `null` or `undefined`
- [ ] Coerces the iterator contents to string, except `null` and `undefined`
- [ ] `null` and `undefined` from the iterator produce empty strings in the result
- [ ] If ToString'ing the separator throws, the iterator gets closed
- [ ] If ToString'ing a value from the iterator throws, the iterator gets closed
- [ ] If the iterator throws, the iterator does not get closed
- [ ] If the iterator otherwise violates the iterator protocol, e.g. by returning `null` from its `next`, the iterator does not get closed
- [ ] If the iterator is exhausted, the iterator does not get closed
- [ ] `.next` is looked up only after ToString'ing the separator

bakkot:
Fixes https://github.com/tc39/test262/issues/4697.

Tested against [a polyfill](https://github.com/tc39/proposal-iterator-join/blob/c1498da49d963972b5f6bd286e17f129193485e4/polyfill.js).

### DIFF
--- a/test/built-ins/Iterator/prototype/join/closes-on-separator-coercion-exception.js
+++ b/test/built-ins/Iterator/prototype/join/closes-on-separator-coercion-exception.js
@@ -1,0 +1,33 @@
+// Copyright (C) 2025 Kevin Gibbons. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-iterator.prototype.join
+description: Iterator.prototype.join closes its receiver if coercing the separator throws.
+features: [Iterator.prototype.join]
+---*/
+
+var throwy = {
+  toString: function () {
+    throw new Test262Error();
+  },
+};
+
+var gotNext = false;
+var calledReturn = false;
+var it = {
+  get next() {
+    // we use a variable instead of simply throwing because throwing is expected in this test
+    gotNext = true;
+  },
+  return: function () {
+    calledReturn = true;
+  },
+};
+
+assert.throws(Test262Error, function () {
+  Iterator.prototype.join.call(it, throwy);
+});
+
+assert.sameValue(gotNext, false);
+assert(calledReturn);
--- a/test/built-ins/Iterator/prototype/join/contents-nullish.js
+++ b/test/built-ins/Iterator/prototype/join/contents-nullish.js
@@ -1,0 +1,13 @@
+// Copyright (C) 2025 Kevin Gibbons. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-iterator.prototype.join
+description: Iterator.prototype.join formats nullish iterator contents as an empty string.
+features: [Iterator.prototype.join]
+---*/
+
+assert.sameValue(
+  ['one', null, 'two', undefined, 'three'].values().join(),
+  'one,,two,,three'
+);
--- a/test/built-ins/Iterator/prototype/join/contents-nullish.js
+++ b/test/built-ins/Iterator/prototype/join/contents-nullish.js
@@ -8,6 +8,11 @@ features: [Iterator.prototype.join]
 ---*/
 
 assert.sameValue(
+  ['one', null, 'two', undefined].values().join(),
+  'one,,two,'
+);
+
+assert.sameValue(
   ['one', null, 'two', undefined, 'three'].values().join(),
   'one,,two,,three'
 );
--- a/test/built-ins/Iterator/prototype/join/contents-tostring.js
+++ b/test/built-ins/Iterator/prototype/join/contents-tostring.js
@@ -1,0 +1,22 @@
+// Copyright (C) 2025 Kevin Gibbons. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-iterator.prototype.join
+description: Iterator.prototype.join coerces non-nullish iterator contents to string.
+features: [Iterator.prototype.join]
+---*/
+
+var called = false;
+var coercible = {
+  toString: function () {
+    if (called) {
+      throw new Test262Error('toString should be called exactly once');
+    }
+    called = true;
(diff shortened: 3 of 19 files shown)


test/built-ins/Iterator/prototype/join/separator-tostring.js
linusg, line 1: I'd like a response to https://github.com/tc39/proposal-iterator-join/issues/1 before merging this please.

linusg: (resolved)

test/built-ins/Iterator/prototype/join/closes-on-separator-coercion-exception.js
```diff
@@ -0,0 +1,33 @@
...
+  return: function () {
+    calledReturn = true;
+  },
+};
+
+assert.throws(Test262Error, function () {
+  Iterator.prototype.join.call(it, throwy);
+});
+
+assert.sameValue(gotNext, false);
```
michaelficarra, line 32:
```suggestion
assert(!gotNext);
```

?

test/built-ins/Iterator/prototype/join/contents-tostring.js
```diff
@@ -0,0 +1,22 @@
...
+esid: sec-iterator.prototype.join
+description: Iterator.prototype.join coerces non-nullish iterator contents to string.
+features: [Iterator.prototype.join]
+---*/
+
+var called = false;
+var coercible = {
+  toString: function () {
+    if (called) {
+      throw new Test262Error('toString should be called exactly once');
```
michaelficarra, line 14: To assert exactly-once behaviour, I would just use a counter and assert on the counter at the end.

test/built-ins/Iterator/prototype/join/closes-on-separator-coercion-exception.js
bakkot: Arguably this is more readable in terms of the failures you get. Certainly it is not worse. In any case I don't think it's worth changing.

test/built-ins/Iterator/prototype/join/contents-tostring.js
bakkot: That is also a valid way to do it but I'm happy with this as-is.

michaelficarra (approved):

test/built-ins/Iterator/prototype/join/contents-nullish.js
```diff
@@ -0,0 +1,13 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-iterator.prototype.join
+description: Iterator.prototype.join formats nullish iterator contents as an empty string.
+features: [Iterator.prototype.join]
+---*/
+
+assert.sameValue(
+  ['one', null, 'two', undefined, 'three'].values().join(),
```
ptomato, line 16: optional: Might be good to test a nullish value as the last value of the iterator, as well.

test/built-ins/Iterator/prototype/join/does-not-close-on-iterator-protocol-violation.js
```diff
@@ -0,0 +1,24 @@
+// Copyright (C) 2025 Kevin Gibbons. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-iterator.prototype.join
+description: Iterator.prototype.join does not close its receiver if the iterator itself throws.
```
ptomato: Description is for another file

ptomato (approved):
These look good, let's merge them in any case.

There's one remaining item from the testing plan https://github.com/tc39/test262/issues/4697 that I don't think is covered in this PR: "`.next` is looked up only after toString'ing the separator".

I hadn't read the testing plan too carefully until now, but it seems fairly complete to me. It occurred to me that you could test the following two things:
- iterator with `get next() { throw new Test262Error() }`
- iterator with `get return() { throw new Test262Error() }`

...these cases would exercise code paths in implementations that these tests don't yet exercise, but I don't think they'd produce observably different results than the test cases you've already got, so it's debatable whether these two cases are meaningful.

bakkot:
~~Unfortunately I can no longer push to this branch.~~ fixed by https://github.com/tc39/test262/issues/5085

test/built-ins/Iterator/prototype/join/next-lookup-after-separator-tostring.js
```diff
@@ -0,0 +1,40 @@
+// Copyright (C) 2026 Kevin Gibbons. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-iterator.prototype.join
+description: Iterator.prototype.join looks up `next` on its receiver only after coercing the separator.
+includes: [compareArray.js]
```
ptomato:
```suggestion
```

bakkot: Oh, I forgot about https://github.com/tc39/test262/pull/4592, nice.

PR #4771, opened 2025-12-09
[Temporal] Make DateTimeFormat tests for Chinese and Dangi dates more lenient

catamorphism:
These tests were failing in JSC because it formats the '1' in one of the dates with a leading 0 ( https://bugs.webkit.org/show_bug.cgi?id=303776 ). Fixed to check that the actual string contains the expected string, rather than an exact match.


Ms2ger:
Hm, if the issue is leading zeroes, I'd rather just check for that. Now you're allowing formatting `1` as `11`, for example.

catamorphism:
> Hm, if the issue is leading zeroes, I'd rather just check for that. Now you're allowing formatting `1` as `11`, for example.

@Ms2ger OK, I changed it to check for leading zeroes.

Ms2ger (approved):
Thanks!

PR #4773, opened 2025-12-09
Fix locale metadata value on `Intl.DateTimeFormat` tests

nekevss:
Based off the documentation in [CONTRIBUTING.md](https://github.com/tc39/test262/blob/main/CONTRIBUTING.md?plain=1#L233), the locale key's value is expected to be an array of one or more language tags or subtags.

This broke Boa's test runner downstream when trying to pin our test262 commit to a more recent commit.


ptomato (approved):
Thanks for the catch.

PR #4775, opened 2025-12-10
Fix format parts test

anba:
Two issues:
1. `typeof part.value` is always a String value.
2. `expectedValue` is a Number value, but the test expected a String.


ptomato (approved):
Thanks!

PR #4776, opened 2025-12-10
Document the CLDR grace period in a new MAINTAINERS.md

gibson042:
To be discussed with implementers in plenary, in case they prefer a different grace period (or none at all—cf. https://github.com/tc39/test262/pull/4573#issuecomment-3572483974 : "<em>engines might not want to update their version of CLDR before relevant tests have landed, e.g. https://github.com/quickjs-ng/quickjs/issues/1168</em>").

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -1,0 +1,6 @@
+# Test262 Maintainer Guidelines
+
+## CLDR
+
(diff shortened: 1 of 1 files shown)


ptomato:
Nice! I've tried to label this appropriately :smile: As discussed in the maintainers meeting, we should include this in our status update in the January TC39 plenary.

ptomato:
Forgot to include it in the status update in the January plenary, so we should do this in March.

PR #4780, opened 2025-12-11
Add test to ensure RoundMVResult in ToIntlMathematicalValue applies 𝔽

anba:
Use "1e10000" from <https://github.com/tc39/ecma402/pull/1022> as the limit so we don't have to modify this test after the PR has been merged.

Note: "1e309" is the smallest decimal string number outside the limits before that PR.


anba:
This test fails in JSC and V8 for the non-decimal number string values.

PR #4781, opened 2025-12-11
Add all possible inputs options to resolved caseFirst/numeric tests

anba:


anba:
The original tests added in #4504 now pass in V8, but when testing all possible input values for `caseFirst` and `numeric`, I still see test failures in V8.

anba:
cc @FrankYFTang

FrankYFTang:
LGTM It also surface two bugs in v8 . I fix them in https://chromium-review.googlesource.com/c/v8/v8/+/7253662

FrankYFTang:
> cc @FrankYFTang

thank you so much for adding these tests and inform me.

PR #4782, opened 2025-12-11
Temporal: Further adjustments to formatToParts with calendars

ptomato:
Follow up to #4771. The expectedValue here is a number, while part.value is a string. Before the previous change, we'd convert expectedValue to a string, so do that again here.

Add an assertion message for debugging.


ptomato:
Never mind, duplicate of #4775.

PR #4783, opened 2025-12-11
Temporal: Change subtract-from-last-representable-month tests

catamorphism:
With the normative change proposed in https://github.com/tc39/proposal-temporal/issues/3197 , these tests are no longer valid. Changed them so they require subtracting a duration from the last representable year-month to succeed, rather than throwing an exception.


ptomato (approved):
LGTM assuming the normative change gets consensus.

catamorphism:
Just pushed a patch to update two more tests that are similar, which I didn't notice the first time.

ptomato:
The additional tests may change based on comments in https://github.com/tc39/proposal-temporal/issues/3197.

ptomato:
It was a different normative change that ended up getting the consensus, so I'll close this.

PR #4796, opened 2025-12-19
Test: Add proposal-error-stack-accessor feature flag to regress-645160

### LINKED ISSUE #4789: test/staging/sm/extensions/regress-645160.js requires non-standard Error().stack
Should be removed or given the feature flag for https://github.com/tc39/proposal-error-stack-accessor.

polar51:
Fixes #4789

As discussed in the issue, this test file requires the `proposal-error-stack-accessor` feature.
I have added the missing feature flag to the frontmatter.


test/staging/sm/extensions/regress-645160.js
```diff
@@ -6,6 +6,8 @@
 /*---
 flags:
   - noStrict
+features:
+  - proposal-error-stack-accessor
```
linusg: there is no such feature in https://github.com/tc39/test262/blob/main/features.txt

polar51: Sorry, I missed checking features.txt before submitting. I'll fix it right away. As you initially suggested, I will remove this file instead of adding a flag since it tests non-standard behavior.

ptomato: Makes sense. The Error stack proposal is stage 2 currently, so we should not add a feature yet.

PR #4799, opened 2025-12-19
[Temporal] Add coverage for negative months and days in `from`

catamorphism:
In the `from` method for the five types that can accept a month or day property, test that negative months or days throw an exception.

### DIFF
--- a/test/built-ins/Temporal/PlainDate/from/negative-month-or-day.js
+++ b/test/built-ins/Temporal/PlainDate/from/negative-month-or-day.js
@@ -1,0 +1,11 @@
+// Copyright (C) 2025 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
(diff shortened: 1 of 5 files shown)


test/built-ins/Temporal/PlainDate/from/negative-month-or-day.js
```diff
@@ -0,0 +1,11 @@
+// Copyright (C) 2025 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.plaindate.prototype.from
```
Ms2ger: Please remove `.prototype` from all

Ms2ger (approved):

test/built-ins/Temporal/PlainDate/from/negative-month-or-day.js
catamorphism: Fixed.

PR #4803, opened 2025-12-29
Drop tests for plural categories of scientific & engineering notation

eemeli:
PR #4503 added tests that include the expected plural categorization of values that are formatted as "1E6", "1,5E6", and "1E-6" in French. These tests should be dropped, because no data exists on the actually appropriate plural categorization of values with exponential notation; see for instance the [CLDR instructions](https://cldr.unicode.org/index/cldr-spec/plural-rules) for determining a language's plural rules, which make no mention of exponential notation.

The values currently in the test are likely whatever ICU4C produces for exponential notation, but those are effectively random; see for instance how the plural category of "1E-6" is expected to be `many`, which makes no sense.

CC @anba, @ryzokuken

### DIFF
--- a/test/intl402/PluralRules/prototype/select/notation.js
+++ b/test/intl402/PluralRules/prototype/select/notation.js
@@ -21,36 +21,26 @@ var values = [
   {
     value: 1e6,
     standard: "many",
-    engineering: "many",
-    scientific: "many",
     compact: "many",
   },
 
   {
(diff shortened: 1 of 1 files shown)


anba (approved):
That makes me wonder if `engineering` and `scientific` shouldn't have been added to `Intl.PluralRules`, given that CLDR doesn't define how to handle these cases.

PR #4804, opened 2026-01-01
Add coverage for abrupt completion from ToNumber via ToUint16 in String.fromCharCode

d01c2:
This PR adds coverage for the case where [ToUint16](https://tc39.es/ecma262/#sec-touint16) produces an abrupt completion because its initial [ToNumber](https://tc39.es/ecma262/#sec-tonumber) step throws. (ToUint16 step 1)

[String.fromCharCode](https://tc39.es/ecma262/#sec-string.fromcharcode) converts each argument with ToUint16, and ToUint16 begins by applying ToNumber. So if the argument triggers an abrupt completion during ToNumber, that abrupt completion must be observable at the String.fromCharCode call site.

String.fromCharCode is a representative builtin that exercises ToUint16, so String.fromCharCode(0n) is a straightforward way to cover this path: ToNumber(0n) throws, and the builtin must propagate that error.

More generally, should test262 aim to add similar coverage for other observable abrupt-completion paths that currently aren’t covered? Or is there a consensus in test262 to keep this kind of coverage limited to a few representative cases?

### DIFF
--- a/test/built-ins/String/fromCharCode/touint16-tonumber-throws-valueof.js
+++ b/test/built-ins/String/fromCharCode/touint16-tonumber-throws-valueof.js
@@ -4,11 +4,8 @@
 /*---
 esid: sec-string.fromcharcode
 description: >
-    String.fromCharCode propagates abrupt completion from ToNumber via ToUint16.
+    String.fromCharCode propagates abrupt completion from ToNumber via ToUint16. (ToPrimitive/valueOf throws)
 info: |
-    String.fromCharCode applies ToUint16 to each argument, and ToUint16 begins
-    with ToNumber, so abrupt completion from ToNumber is propagated.
-
     String.fromCharCode ( ..._codeUnits_ )
 
     2. For each element _next_ of _codeUnits_, do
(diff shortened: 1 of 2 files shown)


d01c2:
I agree with your suggestion. I’ll add a test for that case right away.

ptomato:
> More generally, should test262 aim to add similar coverage for other observable abrupt-completion paths that currently aren’t covered? Or is there a consensus in test262 to keep this kind of coverage limited to a few representative cases?

Ideally, we'd have a similar test for every place ToUint16 is called. JS engines are not under any obligation to structure their code the same way the specification is structured, so they might not use the same ToUint16 code path everywhere the spec has a ToUint16 operation. But 100% coverage of every operation would not be realistic, and in practice engines _probably_ all have a ToUint16 operation so after a certain point we get diminishing returns. It's just a matter of priorities and what people are willing to contribute. Certainly we wouldn't _reject_ a PR that added more similar tests in additional places where ToUint16 is called.

test/built-ins/String/fromCharCode/touint16-tonumber-throws-valueof.js
```diff
@@ -0,0 +1,31 @@
...
+
+assert.throws(TypeError, function () {
+  String.fromCharCode(0n);
+}, "ToNumber throws when argument is BigInt, and String.fromCharCode must propagate it");
+
+assert.throws(Test262Error, function () {
+  String.fromCharCode({
+    valueOf: function () { throw new Test262Error(); }
+  });
+}, "ToNumber throws when its argument's ToPrimitive step calls a throwing valueOf, and String.fromCharCode must propagate it.");
```
ptomato, line 27: I'm assuming @ljharb meant to put this Test262Error test in a separate file, so that it would be run even on engines without BigInt support.

ljharb: yes, precisely that :-)

d01c2: I’ve just split it into two files.

ptomato (approved):
Thanks!

PR #4806, opened 2026-01-05
[Temporal] Coverage for formatting a PlainDateTime with invalid formatter options

catamorphism:
A PlainDateTime cannot be formatted if the formatter was created with a defined `timeZoneName` option. Add coverage for this case.


Ms2ger (approved):

anba:
These tests should be invalid, because `timeZoneName` can't be used as a standalone option:
- [GetDateTimeFormat](https://tc39.es/proposal-temporal/#sec-getdatetimeformat), step 14 doesn't include `"timeZoneName"`.
- `anyPresent` is therefore always `false`, which means step 17.a never returns `null`.

PR #4807, opened 2026-01-05
[Temporal] Add coverage for non-string month codes in `from` methods

catamorphism:
Add coverage for primitive non-string values passed in for the `monthCode` option in the `from` method for PlainDate, PlainDateTime, PlainYearMonth, and ZonedDateTime.

### DIFF
--- a/test/built-ins/Temporal/PlainDate/from/month-code-wrong-type.js
+++ b/test/built-ins/Temporal/PlainDate/from/month-code-wrong-type.js
@@ -1,0 +1,24 @@
+// Copyright (C) 2026 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.plaindate.from
+description: Month code must be a string
+features: [Temporal]
+---*/
+
+const monthCodeValues = [
+  5, 5n, false, Symbol(), null, { toString: () => 5 }
+];
+
+const year = 2026;
+
+for (const monthCode of monthCodeValues) {
+  assert.throws(TypeError, () => Temporal.PlainDate.from({
+    year,
+    monthCode,
+    day: 1
+  }), typeof monthCode === 'symbol' ?
+      "Symbol should be rejected as month code" :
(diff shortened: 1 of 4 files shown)


test/intl402/Temporal/PlainDate/prototype/monthCode/month-code-wrong-type.js
```diff
@@ -0,0 +1,24 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.plaindate.prototype.monthcode
+description: Month code must be a string
+features: [Temporal]
+---*/
+
+const monthCodeValues = [
+  5, 5n, false, Symbol(), null
```
Ms2ger: Could consider adding an object with a `toString()` method that returns a non-string

test/intl402/Temporal/PlainDate/prototype/monthCode/month-code-wrong-type.js
```diff
@@ -0,0 +1,24 @@
+// Copyright (C) 2026 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.plaindate.prototype.monthcode
```
Ms2ger: I don't think this is in the right place. It's a test for `.from()`, not `date.monthCode`, and I think it doesn't even need to be under `intl402/`

Ms2ger (changes requested):

test/intl402/Temporal/PlainDate/prototype/monthCode/month-code-wrong-type.js
catamorphism: Done.

test/intl402/Temporal/PlainDate/prototype/monthCode/month-code-wrong-type.js
catamorphism: Moved tests.

PR #4808, opened 2026-01-05
[Temporal] Add coverage for offset time zones in toLocaleString

catamorphism:
toLocaleString should be able to format a ZonedDateTime with an offset time zone; add coverage for this case.


catamorphism:
The accompanying Temporal polyfill change is https://github.com/tc39/proposal-temporal/pull/3241

PR #4809, opened 2026-01-05
[Temporal] Add coverage for month before earliest supported month

catamorphism:
For PlainYearMonth.from with non-ISO calendars, add test coverage for the cases where the month is one month before the earliest supported month, and the month is one month after the latest supported month (an exception should be thrown).

Closes https://github.com/tc39/proposal-temporal/issues/3082

### DIFF
--- a/test/intl402/Temporal/PlainYearMonth/from/extreme-unsupported-dates.js
+++ b/test/intl402/Temporal/PlainYearMonth/from/extreme-unsupported-dates.js
@@ -1,0 +1,52 @@
+// Copyright (C) 2026 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-temporal.plaindate.from
+description: from() throws for month before earlier supported month, and month after latest supported month
+features: [Temporal, Intl.Era-monthcode]
+---*/
+
+// Lunisolar/lunar calendars can't accurately predict celestial orbits for dates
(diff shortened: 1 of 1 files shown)


Manishearth:
I don't think this test is correct for Coptic, and I think the polyfill is wrong here.

`Temporal.PlainDate.from({ calendar: "coptic", year: 275471, era: "am", eraYear: 275471, month: 5, day: 1);` gives `275760-08-23[u-ca=coptic]` in the polyfill. 275760-09-13 is the max ISO date. 

The coptic test tests `Temporal.PlainYearMonth.from({ calendar: "coptic", year: 275471, era: "am", eraYear: 275471, month: 6, monthCode: "M06"});`, which will of course produce a value after `275760-09-13`. The corresponding date for `day: 1` is  `275760-09-22`

But: [`CalendarYearMonthFromFields`](https://tc39.es/proposal-temporal/#sec-temporal-calendaryearmonthfromfields) doesn't check for the ISO date being out of range, it only checks `ISOYearMonthWithinLimits`, which checks the month not the day. As far as I can understand it, this test does not match the spec.

cc @ptomato who might catch something I missed.

ptomato:
Thanks for checking, I'll take a look tomorrow.

ptomato:
@Manishearth I concur, looks like I made a wrong assumption when I wrote the issue description in https://github.com/tc39/proposal-temporal/issues/3082

catamorphism:
Created https://github.com/tc39/proposal-temporal/issues/3251 to track the bug in the polyfill.

PR #4810, opened 2026-01-05
[Temporal] Remove staging/sm/PlainMonthDay tests

catamorphism:
For from-chinese.js, from-coptic.js, and result-not-after-1972-dec-31.js, the functionality is already in
intl402/Temporal/PlainMonthDay/from/*-month-codes.js.

from-chinese-leap-month-uncommon.js tests dates outside the supported range.


anba:
> from-chinese-leap-month-uncommon.js tests dates outside the supported range.

The tests didn't require any specific reference year, but instead just checked that implementations can handle uncommon leap months. Does either spec ("Temporal" or "Intl Era and MonthCode Proposal") allow to reject these inputs? Otherwise the test should probably be restored.

ptomato:
We'll need a slightly different approach for testing if https://github.com/tc39/proposal-intl-era-monthcode/pull/108#discussion_r2683345353 is approved in the upcoming plenary meeting. I plan to submit a new PR for that.

PR #4838, opened 2026-01-08
[import-defer] adding cases when namespace is being accessed as super alias

caiolima:
This is adding cases when namespace is being accessed aliased by super property, or when property definition is called because of class fields. The motivation here is that on V8 implementation, `super` [[Set]] operation have dedicated code path that differs from ordinary property [[Set]], and a bug was found during review[1].

[1] - https://chromium-review.googlesource.com/c/v8/v8/+/7247865/comment/59cf8f4e_b3185151/.

### DIFF
--- a/src/import-defer/super-get.case
+++ b/src/import-defer/super-get.case
@@ -1,0 +1,24 @@
+// Copyright (C) 2026 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-module-namespace-exotic-objects-get-p-receiver
+desc: _ [[Get]] when namespace is aliased by super
+info: |
+  [[Get]] ( _P_, _Receiver_ )
+    1. If IsSymbolLikeNamespaceKey(_P_, _O_), return ! OrdinaryGet(_O_, _P_, _Receiver_).
+    1. Let _exports_ be ? GetModuleExportsList(_O_).
+    1. ...
+
+template: trigger-on-possible-export
+---*/
+
+//- body
+let obj = {
+  superGet(key) {
+    return super[key];
+  }
+}
+
+Object.setPrototypeOf(obj, ns);
+obj.superGet(key);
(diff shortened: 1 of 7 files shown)


src/import-defer/super-property-set-exported.case
```diff
@@ -0,0 +1,25 @@
...
+
+template: ignore
+---*/
+
+//- body
+class A { constructor() { return ns; } };
+class B extends A {
+  constructor() {
+    super();
+    super.exported = 14;
```
nicolo-ribaudo, line 19: Can the file be renamed to explicitly mention that this is a _set_ access? We do have tests for get already, right?

caiolima: Renamed.

Regarding super get, I couldn't find cases for it, but they are kind of similar to cases where we add `ns` as prototype. Anyways, I added a case for it as well, because even if V8 doesn't treat [[Get]] on super different from ordinary get, other implementation might have a different path.

caiolima: The [[Get]] test is different from set, because `super.prop = ...` will actually try to perform a set on `this.prop` because of the logic in [PutValue](https://tc39.es/ecma262/#sec-putvalue), and `super[expression]` will perform `GetPrototypeOf` home object [[1]](https://tc39.es/ecma262/#sec-super-keyword-runtime-semantics-evaluation).

caiolima:
Thanks a lot for reviews!

iamstolis:
I believe that `language/import/import-defer/evaluation-triggers/ignore-super-property-set-exported.js` and `language/import/import-defer/evaluation-triggers/ignore-super-property-set-not-exported.js` added by this PR are incorrect. They expect deferred module evaluation not to be triggered because of namespace module object's `[[Set]]` override.

`super.x = ...` leads to `PutValue(superPropertyReference, ...)` and yes `PutValue(V, ...)` invokes `[[Set]]` but on `V.[[Base]]` and `superPropertyReference.[[Base]]` is not `this` but `[[HomeObject]].[[GetPrototypeOf]]()`. In other words, it is an ordinary `[[Set]]` (not the one of module namespace objects). The namespace object is passed as the `receiver`. So, we go through `OrdinarySet()` and `OrdinarySetWithOwnDescriptor()` that reads `existingDescriptor` (using `Receiver.[[GetOwnProperty]](P)`) which triggers the module evaluation.

I filled #4980 about it.

PR #4860, opened 2026-01-14
Intl Era Monthcode: Test both leap and common year in invalid-month-codes-hebrew

catamorphism:
The existing test checked that M13 is not a valid month code in 5779 (a leap year). Add a test to check the same property in 5781 (a common year).


catamorphism:
Updated the other tests except for PlainMonthDay; this change doesn't apply when there's no year.

PR #4873, opened 2026-01-16
Fix running Temporal tests in strict mode

trflynn89:
Previously, a ReferenceError would be thrown for the options object in these loops.


Ms2ger (approved):
Thanks!

PR #4874, opened 2026-01-19
Delete `two-digit-years.js` and `non-iso.js`

### LINKED ISSUE #4788: test/staging/sm/Date/non-iso.js should be deleted
Tests implementation-defined behavior of SM.

Gumichocopengin8:
> If the String does not conform to that format the function may fall back to any implementation-specific heuristics or implementation-specific date formats.
>
> https://tc39.es/ecma262/#sec-date.parse

Based on https://tc39.es/ecma262/#sec-date.parse, these test files do not conform to date time string format (https://tc39.es/ecma262/#sec-date-time-string-format). Therefore, remove the files.

close #4788


Ms2ger:
Ref #4788 as well

Gumichocopengin8:
@Ms2ger Deleted `non-iso.js` as well.

Gumichocopengin8:
Is there anything else I should do to get a review?

ptomato:
The PR itself looks fine. Although I'm not sure if everything in these files is just invalid, or we need to discuss having coverage for web reality in some other way that isn't disruptive for implementations executing the test suite?

PR #4876, opened 2026-01-19
Added the case where cause field does not exist in options in NativeError constructor

G-Lee1031:
Hello.
This PR adds the case where _options_ is an object but does not have a cause field in [NativeError(_message_, [_options_])](https://tc39.es/ecma262/#sec-nativeerror).

In NativeError step 4, [InstallErrorCause(O, options)](https://tc39.es/ecma262/#sec-installerrorcause) is called. But the test where options is an object but not having a cause field does not exist.

In order to test that, _options_ should be an object  but should not have a cause field.

So `EvalError(0, {})` is a straightforward example.

For clarity, I included every types of NativeErrors in the same file.


ljharb (approved):

linusg (changes requested):
This test most definitely doesn't belong in `NativeErrors/SyntaxError/`.

That aside, this is already somewhat covered by https://github.com/tc39/test262/blob/e85173bc211eb110477de3279ec1dcd36d9a8a58/test/built-ins/NativeErrors/cause_property_native_error.js, and instead of this new test one line could be added:

```js
verifyProperty(new nativeError(message, {}), "cause", undefined);
```

G-Lee1031:
Apologies for placing the file in the wrong directory. Thank you for pointing that out. And, it was my oversight to create a new file instead of using the existing one. I have updated the PR to merge the test case into the correct file as you suggested.

linusg (approved):

PR #4877, opened 2026-01-20
Intl Era Monthcode: Update Meiji era start date

ptomato:
See https://github.com/tc39/proposal-intl-era-monthcode/pull/102

The historical regnal period starts on 1868-10-23 but at that point the lunisolar calendar was still in use. On 1873-01-01 we start using the era code "meiji", before that we use eras and era years identical to the "gregory" calendar.


ptomato:
The normative change gained consensus in the TC39 plenary 2026-01-20

PR #4881, opened 2026-01-21
Added coverage for abrupt completion returning cases in ToNumber called by math

G-Lee1031:
Hello.
This PR covers 26 cases of `Math` method where [ToNumber](https://tc39.es/ecma262/#sec-tonumber) is returned with an abrupt completion.

I have added a total of **77 files**. Aside from two specific cases (explained below), the remaining 75 files follow a consistent pattern to test the error-handling path of the `ToNumber` operation for each `Math` method.

#### 1. Common Test Patterns (75 files)
* **Case 01 (`*-tonumber01.js`)**: when an argument is a BigInt, ToNumber throws a TypeError.
* **Case 02 (`*-tonumber02.js`)**: when an argument is a Symbol, ToNumber throws a TypeError.
* **Case 03 (`*-tonumber03.js`)**: when [ToPrimitive](https://tc39.es/ecma262/#sec-toprimitive) is called and returns an abrupt completion.

Specifically in this PR, I made ToPrimitive throw a TypeError by setting _argument_ as `{toString: undefined, valueOf: undefined}` so that ToPrimitive throws in its 1.d step, where
 [OrdinaryToPrimitive(O, hint)](https://tc39.es/ecma262/2025/#sec-ordinarytoprimitive) is called and throws in step 4.

#### 2. Specific Cases (2 files)
These two were named differently because a [similar example](https://github.com/tc39/test262/blob/main/test/built-ins/Math/hypot/Math.hypot_ToNumberErr.js) where ToNumber throws because of `valueOf` existed. 
* `Math.hypot_ToNumberErr02.js`: when an argument is a BigInt, ToNumber throws a TypeError.
* `Math.hypot_ToNumberErr03.js`: when an argument is a Symbol, ToNumber throws a TypeError.

### Covered Methods
The tests cover the following 25 `Math` methods (3 files each):
`abs`, `acos`, `acosh`, `atan`, `atan2`, `cbrt`, `ceil`, `cos`, `cosh`, `exp`, `f16round`, `fround`, `imul`, `log`, `log10`, `log2`, `max`, `min`, `pow`, `round`, `sin`, `sinh`, `sqrt`, `tan`, `tanh`.


ljharb:
Can you name the files descriptively, please, instead of numbering them?

(also, please be sure to not use an LLM for open source contributions)

G-Lee1031:
Thank you for the detailed feedback!

Regarding the file names, I named them *-tonumber01... to keep them consistent with my previous PRs, but I understand your point. Also, just to clarify—I wrote all the code manually without using any LLMs. :)

I realize that 77 files can be overwhelming to review, and I apologize for the heavy load. I grouped them together because they all share the exact same logic and format.

Would it be better if I renamed them to something more descriptive like abs-tonumber-bigint? Please let me know if this works for you, or if you have a preferred naming convention!

ptomato:
Yes, names like `abs-tonumber-bigint.js` would be strongly preferred.

G-Lee1031:
I changed the file names as suggested! :)

PR #4975, opened 2026-03-06
Add tests for RegExp Buffer Boundaries

rbuckton:
Adds tests for [RegExp Buffer Boundaries](https://github.com/tc39/proposal-regexp-buffer-boundaries)


gibson042 (approved):
LGTM, pending proposal advancement.

rbuckton:
Updated per the consensus to add `\Z` using the `(?=(?:\r\n|\n|\r|\u2028|\u2029)?(?-m:$))` semantics.

test/built-ins/RegExp/buffer-boundaries/syntax/u-mode.js
```diff
@@ -0,0 +1,63 @@
...
+features: [regexp-buffer-boundaries]
+---*/
+
+/\A/u;
+/\A/um;
+/\A/umi;
+/\A/us;
+/\A/usi;
+/\A/usm;
+/\A/usmi;
```
gibson042, line 29: Is the omission of `/\A/ui` intentional? And likewise for `/\z/ui` and `/\Z/ui`, and analogously in sibling file v-mode.js.

test/built-ins/RegExp/buffer-boundaries/slash-lower-case-z-matches-end-of-buffer.js
```diff
@@ -0,0 +1,59 @@
...
+assert(/x\z/u.test("x"), "Expected \\z to match the end of the buffer outside of multiline mode");
+assert(/x\z/um.test("x"), "Expected \\z to match the end of the buffer inside of multiline mode");
+assert(!/x\z/u.test("xy"), "Expected \\z to not match when not at the end of the buffer outside of multiline mode");
+assert(!/x\z/um.test("xy"), "Expected \\z to not match when not at the end of the buffer inside of multiline mode");
+assert(!/x\z/u.test("x\ny"), "Expected \\z to not match when only at the end of a line outside of multiline mode");
+assert(!/x\z/um.test("x\ny"), "Expected \\z to not match when only at the end of a line inside of multiline mode");
+assert(!/x\z/u.test("x\n"), "Expected \\z to not match when only at the end of a line outside of multiline mode");
+assert(!/x\z/um.test("x\n"), "Expected \\z to not match when only at the end of a line inside of multiline mode");
+
+assert(/x\z/v.test("x"), "Expected \\z to match the end of the buffer outside of multiline mode");
```
gibson042: It would be preferable for each assertion message in this file to be unique. I'd recommend a pattern like "_Expected \\z to {`match end-of-buffer`,`not match when not at end-of-buffer`,`not match when only at end-of-line`,`not match when preceding a final line feed`} {`inside`,`outside`} of multiline mode ({`/u`,`/v`}, {`literal`,`constructed`})_". But also, separating `/u` and `/v` into distinct files would be better, because the latter depends upon feature "regexp-v-flag" but the former does not (in which case the `/u`/`/v` is implied by the file itself and therefore excluded).

test/built-ins/RegExp/buffer-boundaries/slash-upper-case-a-matches-start-of-buffer.js
```diff
@@ -0,0 +1,49 @@
...
+---*/
+
+assert(/\Ax/u.test("x"), "Expected \\A to match the start of the buffer outside of multiline mode");
+assert(/\Ax/um.test("x"), "Expected \\A to match the start of the buffer inside of multiline mode");
+assert(!/\Ax/u.test("yx"), "Expected \\A to not match when not at the start of the buffer outside of multiline mode");
+assert(!/\Ax/um.test("yx"), "Expected \\A to not match when not at the start of the buffer inside of multiline mode");
+assert(!/\Ax/u.test("y\nx"), "Expected \\A to not match when at the start of a new line outside of multiline mode");
+assert(!/\Ax/um.test("y\nx"), "Expected \\A to not match when at the start of a new line inside of multiline mode");
+
+assert(/\Ax/v.test("x"), "Expected \\A to match the start of the buffer outside of multiline mode");
```
gibson042: Similarly, "_Expected \\A to {`match the start of the buffer`,`not match when not at the start of the buffer`,`not match when at the start of a new line`} {`inside`,`outside`} of multiline mode ({`/u`,`/v`}, {`literal`,`constructed`})_".

test/built-ins/RegExp/buffer-boundaries/slash-upper-case-z-matches-lineterminatorsequence-end-of-buffer.js
```diff
@@ -0,0 +1,127 @@
...
+          i. Return continue(matchState).
+      h. If e = inputLength - 2, the character input[e] is matched by <CR>; and the character input[e + 1] is matched by <LF>;, then
+          i. Return continue(matchState).
+      i. Return failure.
+
+esid: sec-patterns
+features: [regexp-buffer-boundaries]
+---*/
+
+assert(/x\Z/u.test("x"), "Expected \\Z to match the end of the buffer outside of multiline mode");
```
gibson042: This file, on the other hand, has so many cases that I'd switch over to a literal loop rather than unrolling it into a static sequence of assertions.

```js
assert(/x\Z/u.test("x"),
  "Expected \\Z to match before end of buffer outside of multiline mode");
assert(/x\Z/um.test("x"),
  "Expected \\Z to match before end of buffer inside of multiline mode");
assert(!/x\Z/u.test("xy"),
  "Expected \\Z to not match when not at end of buffer outside of multiline mode");
assert(!/x\Z/um.test("xy"),
  "Expected \\Z to not match when not at end of buffer inside of multiline mode");

var terminators = [
  ["\n", "line feed"],
  ["\r", "carriage return"],
  ["\u2028", "line separator"],
  ["\u2029", "paragraph separator"],
  ["\r\n", "CRLF"]
];

for (var i = 0; i < terminators.length; i++) {
  var where = terminators[i][1];

  assert(/x\Z/u.test("x" + terminators[i][0]),
    "Expected \\Z to match before a final " + where + " outside of multiline mode (literal)");
  assert(/x\Z/um.test("x" + terminators[i][0]),
    "Expected \\Z to match before a final " + where + " inside of multiline mode (literal)");

  assert(!/x\Z/u.test("x" + terminators[i][0] + terminators[i][0],
    "Expected \\Z to not match before two final " + where + "s outside of multiline mode (literal)");
  assert(!/x\Z/um.test("x" + terminators[i][0] + terminators[i][0],
    "Expected \\Z to not match before two final " + where + "s inside of multiline mode (literal)");

  assert(!/x\Z/u.test("x" + terminators[i][0] + "y"),
    "Expected \\Z to not match when only before " + where + " outside of multiline mode (literal)");
  assert(!/x\Z/um.test("x" + terminators[i][0] + "y"),
    "Expected \\Z to not match when only before " + where + " inside of multiline mode (literal)");
}

for (var i = 0; i < terminators.length; i++) {
  var where = terminators[i][1];

  assert(new RegExp("x\\Z", "u").test("x" + terminators[i][0]),
    "Expected \\Z to match before a final " + where + " outside of multiline mode (constructed)");
  assert(new RegExp("x\\Z", "um").test("x" + terminators[i][0]),
    "Expected \\Z to match before a final " + where + " inside of multiline mode (constructed)");

  assert(!new RegExp("x\\Z", "u").test("x" + terminators[i][0] + terminators[i][0],
    "Expected \\Z to not match before two final " + where + "s outside of multiline mode (constructed)");
  assert(!new RegExp("x\\Z", "um").test("x" + terminators[i][0] + terminators[i][0],
    "Expected \\Z to not match before two final " + where + "s inside of multiline mode (constructed)");

  assert(!new RegExp("x\\Z", "u").test("x" + terminators[i][0] + "y"),
    "Expected \\Z to not match when only before " + where + " outside of multiline mode (constructed)");
  assert(!new RegExp("x\\Z", "um").test("x" + terminators[i][0] + "y"),
    "Expected \\Z to not match when only before " + where + " inside of multiline mode (constructed)");
}
```

test/built-ins/RegExp/buffer-boundaries/slash-upper-case-z-matches-lineterminatorsequence-end-of-buffer.js
```diff
@@ -0,0 +1,127 @@
+// Copyright 2026 Ron Buckton. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+author: Ron Buckton
+description: >
+  `\Z` (lower-case z) matches end of buffer in any unicode mode.
```
gibson042:
```suggestion
  `\Z` (upper-case z) matches end of buffer after optional final line terminator in any unicode mode.
```

test/built-ins/RegExp/buffer-boundaries/syntax/v-mode.js
```diff
@@ -0,0 +1,63 @@
...
+  
+  Assertion[UnicodeMode, UnicodeSetsMode, NamedCaptureGroups] :
+    ...
+    [+UnicodeMode] `\` `A`
+    [+UnicodeMode] `\` `z`
+    [+UnicodeMode] `\` `Z`
+    ...
+
+esid: sec-patterns
+features: [regexp-buffer-boundaries]
```
gibson042:
```suggestion
features: [regexp-v-flag, regexp-buffer-boundaries]
```

gibson042 (changes requested):
Please split the files to separate assertions that depend upon feature "regexp-v-flag" from those that do not. See also other suggestions below.

test/built-ins/RegExp/buffer-boundaries/syntax/u-mode.js
rbuckton: Not really, the `i` shouldn't be relevant, to be honest, but I can add them for coverage.

rbuckton:
Also updated the descriptions for the annex-b identity escape tests

test/built-ins/RegExp/buffer-boundaries/slash-lower-case-z-matches-end-of-buffer-u-mode.js
```diff
@@ -0,0 +1,37 @@
...
+esid: sec-patterns
+features: [regexp-buffer-boundaries]
+---*/
+
+assert(/x\z/u.test("x"), "Expected \\z to match end-of-buffer outside of multiline mode (literal)");
+assert(/x\z/um.test("x"), "Expected \\z to match end-of-buffer inside of multiline mode (literal)");
+assert(!/x\z/u.test("xy"), "Expected \\z to not match when not at end-of-buffer outside of multiline mode (literal)");
+assert(!/x\z/um.test("xy"), "Expected \\z to not match when not at end-of-buffer inside of multiline mode (literal)");
+assert(!/x\z/u.test("x\n"), "Expected \\z to not match when only at end-of-line outside of multiline mode (literal)");
+assert(!/x\z/um.test("x\n"), "Expected \\z to not match when only at end-of-line inside of multiline mode (literal)");
```
gibson042: On second thought, this file also should check against all line terminators in case some implementation allows `/x\z/` to incorrectly match e.g. "x\u2028".

gibson042 (approved):
Looks good, just one final suggestion.

gibson042 (approved):

rbuckton:
Missed one of the comments

sno2:
Hello, I believe this test should be updated to allow these new escapes as well, right?

https://github.com/tc39/test262/blob/d86b2294eb0a17eaa281ff12c73c473ec864c72f/test/built-ins/RegExp/unicode_restricted_identity_escape_alpha.js#L17-L40

test/annexB/built-ins/RegExp/buffer-boundaries/not-supported-outside-unicode-modes.js
```diff
@@ -0,0 +1,60 @@
...
+  ["\r", "<CR>"],
+  ["\u2028", "<LS>"],
+  ["\u2029", "<PS>"],
+  ["\r\n", "<CRLF>"],
+];
+
+for (var i = 0; i < terminators.length; i++) {
+  var lts = terminators[i][0], desc = terminators[i][1];
+  assert(!/^x\Z$/.test("x" + lts), "Expected \\Z to not match at " + desc + " before end-of-buffer outside of unicode mode. (literal)");
+  assert(!new RegExp("^x\\Z$").test("x" + lts), "Expected \\Z to not match at " + desc + " before end-of-buffer outside of unicode mode. (constructed)");
```
bakkot, line 59:
```suggestion
  assert(!new RegExp("^x\\Z").test("x" + lts), "Expected \\Z to not match at " + desc + " before end-of-buffer outside of unicode mode. (constructed)");
```
presumably? Unless you intended these to be multiline mode.

test/built-ins/RegExp/buffer-boundaries/syntax/u-mode.js
```diff
@@ -0,0 +1,69 @@
+// Copyright 2026 Ron Buckton. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+author: Ron Buckton
+description: >
+  `\A`, `\z`, and `\Z` are parsed successfully in u-mode
```
bakkot, line 7: We should have negative tests (for both modes) asserting you can't have one of these followed by a quantifier (i.e. `*` `+` `?` `{1, 2}`).

test/built-ins/RegExp/buffer-boundaries/slash-upper-case-z-matches-lineterminatorsequence-end-of-buffer-u-mode.js
```diff
@@ -0,0 +1,61 @@
+// Copyright 2026 Ron Buckton. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+author: Ron Buckton
+description: >
+  `\Z` (upper-case z) matches end of buffer after optional LineTerminatorSequence in u-mode.
```
bakkot, line 7: We also need to assert that it is zero width; unless I'm missing something that's not asserted anywhere. For completeness I think we want both `/x\Z/u.exec("x\n")[0] === "x"` and `/x\Z\n/u.test("x\n") === true` (for all the line terminators).

test/built-ins/RegExp/buffer-boundaries/slash-upper-case-z-matches-lineterminatorsequence-end-of-buffer-u-mode.js
```diff
@@ -0,0 +1,61 @@
...
+assert(/x\Z/um.test("x"), "Expected \\Z to match end-of-buffer inside of multiline mode (literal)");
+assert(!/x\Z/u.test("xy"), "Expected \\Z to not match when not at end-of-buffer outside of multiline mode (literal)");
+assert(!/x\Z/um.test("xy"), "Expected \\Z to not match when not at end-of-buffer inside of multiline mode (literal)");
+
+for (var i = 0; i < terminators.length; i++) {
+  var lts = terminators[i][0], desc = terminators[i][1];
+  assert(/x\Z/u.test("x" + lts), "Expected \\Z to match when at " + desc + " before end-of-buffer outside of multiline mode (literal)");
+  assert(/x\Z/um.test("x" + lts), "Expected \\Z to match when at " + desc + " before end-of-buffer inside of multiline mode (literal)");
+  assert(!/x\Z/u.test("x" + lts + "y"), "Expected \\Z to not match when at " + desc + " but not before end-of-buffer outside of multiline mode (literal)");
+  assert(!/x\Z/um.test("x" + lts + "y"), "Expected \\Z to not match when at " + desc + " but not before end-of-buffer inside of multiline mode (literal)");
```
bakkot, line 47: Can we also test when there are multiple line terminators? I.e. `!/x\Z/u.test("x\n\n")`.

And as an edge case, `/x\r\Z\n/u.test("x\r\n")` should match.

bakkot (commented):
I agree with the above comment about unicode_restricted_identity_escape_alpha.js.

Also some coverage gaps. I left a few comments above; in addition I would like to see:
- all assertions should match an empty input
- behavior in lookarounds
- behavior of `\A` when matching with sticky + nonzero `lastIndex` (specifically that it does not match, except that it can match with a lookbehind)

Tests look good otherwise.

PR #5005, opened 2026-03-24
Module source object host invariant test

guybedford:
This adds a test for source phase imports of the host invariant:

> The Module Source Object corresponding to this source Module Record's source phase (<emu-xref href="#sec-module-source-objects"></emu-xref>), or ~empty~ if it is not available for this module kind. When not ~empty~, it must be an object whose initial [[Prototype]] is an object whose initial [[Prototype]] is %AbstractModuleSource%.prototype.

which was recently updated in https://github.com/tc39/proposal-source-phase-imports/pull/77.


test/staging/source-phase-imports/module-source-prototype-chain.js
```diff
@@ -0,0 +1,35 @@
...
+  When not ~empty~, it must be an object whose initial [[Prototype]] is
+  an object whose initial [[Prototype]] is %AbstractModuleSource%.prototype.
+
+features: [source-phase-imports, source-phase-imports-module-source]
+flags: [async]
+includes: [asyncHelpers.js]
+---*/
+
+asyncTest(async function () {
+  const moduleSource = await import.source('<module source>');
```
bakkot, line 23: Hosts are generally permitted to not provide a source for a module, right? So presumably this needs a try-catch, to only assert anything in the case that there is something available?

Separately, this is kind of a weird module specifier to be assuming is available.

nicolo-ribaudo: It's defined in https://github.com/tc39/test262/blob/main/INTERPRETING.md#modules.

bakkot: Aha, nice.

bakkot (approved):

PR #5012, opened 2026-04-08
web-feature classification for additional Iterator and Map features

ChrisC:
This PR maps tests for the following newer web features:
* [Iterator.concat()](https://github.com/web-platform-dx/web-features/blob/main/features/iterator-concat.yml)
* [getOrInsert](https://github.com/web-platform-dx/web-features/blob/main/features/getorinsert.yml) (which covers both Map and WeakMap)

cc @jugglinmike for review

### DIFF
--- a/WEB_FEATURES.yml
+++ b/WEB_FEATURES.yml
@@ -345,6 +345,12 @@ features:
       - test/built-ins/global
     tags:
       - globalThis
+  
+  - name: getorinsert
+    files:
+      - test/built-ins/*Map/prototype/getOrInsert*
+    tags: 
+      - upsert
 
   - name: hashbang-comments
     files:
@@ -423,6 +429,12 @@ features:
   - name: is-error
     files:
       - test/built-ins/Error/isError
+
+  - name: iterator-concat
+    files:
+      - test/built-ins/Iterator/concat
+    tags:
+      - iterator-sequencing
 
   - name: iterator-methods
     files:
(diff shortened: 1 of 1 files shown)


WEB_FEATURES.yml
```diff
@@ -1034,6 +1045,7 @@ features:
   - name: weakmap
     files:
       - test/built-ins/WeakMap
+      - "!test/built-ins/WeakMap/prototype/getOrInsert*"
```
ChrisC: Noting that we'll have less tests mapped to `weakmap` as of this change to avoid overlap, so it's expected that the coverage checker will throw a warning. Theses tests are now mapped to the `getorinsert` feature, so there's no overall reduction in test-to-feature coverage.

jugglinmike:
I think using tags (`upsert` and `iterator-sequencing`) might be better here. Not so much for concision (in fact, there's a good argument to be made for the readability of an explicit list of directories) but for flexibility. The two approaches are equivalent today, but if someone adds a cross-cutting test to another directory later, then a tag-based classifier has a better chance of automatically including it. What do you think?

ChrisC:
> I think using tags (`upsert` and `iterator-sequencing`) might be better here...

@jugglinmike addressed in [688838b](https://github.com/tc39/test262/pull/5012/commits/688838b0c52626cbc50f93052687108f0931b2ee). How does that look?

jugglinmike (approved):

jugglinmike:
@ptomato Can you admin-merge this?

(Context: the linter is rejecting this patch because one of matched sets is shrinking. That's intentional. [We designed the linter to behave this way](https://github.com/tc39/test262/pull/4701) in the interest of avoiding unintentional regression and also to limit its complexity. That design assumes that admins can merge despite failure, though, and Test262 isn't currently configured that way--not for me, anyway. I'm also unable to push the commit to `main` from my local repository.)

jugglinmike:
Thanks!

PR #5014, opened 2026-04-09
Extend coverage for Intl.Locale firstDayOfWeek

anba:
In "constructor-options-firstDayOfWeek-invalid.js":
- Test strings containing non-ASCII alphanum characters
- Test numbers and bigints smaller than 0 and larger than 7.
- Test fractional numbers and negative Infinity.

In "constructor-options-firstDayOfWeek-valid.js":
- Test more numbers and also bigints.
- Test NaN and positive Infinity.
- Test the string `"yes"` as a special-case.
- Test upper case strings.

In "firstDayOfWeek/valid-id.js":
- Test not just the known weekdays, but also additional strings.
- Include special cases: empty string, `"true"`, and `"yes"`.
- Test number-only strings.
- Test upper case strings.

In "firstDayOfWeek/valid-options.js":
- Test options added to "constructor-options-firstDayOfWeek-valid.js".
- Test upper case strings.

Drive-by change:
- Use `"` consistently for string values.


anba:
Test failures in JSC and V8 are expected, because of ICU4C bugs.

jugglinmike (approved):
No conflicts and no overlap with more recent additions; this all looks good to me!

PR #5018, opened 2026-04-10
Remove substr in Temporal tests

Jack-Works:
Currently, engine262 does not implement Annex B; therefore, those tests failed, but they pass after replacing substr with slice.


ljharb (approved):
slice > > > substr, always :-)

ptomato (approved):

PR #5021, opened 2026-04-13
Add more tests for Intl.Locale.prototype.getCollations

anba:
Add coverage for:
- Return collations are supported by `Intl.Collator`.
- Language subtag is `und`.
- Language subtag has more than three letters.
- Output array is sorted.

Also update existing test meta data to match current spec text and simplify checks in "output-array-values.js"


anba:
- `output-array-is-sorted.js` fails in JSC, because JSC doesn't sort the output array.
- `preferred-from-unicode-extension-true-empty.js` fails in JSC and V8 because of an ICU4C-related issue.

PR #5022, opened 2026-04-13
Add more tests for Intl.Locale.prototype.getNumberingSystems

anba:
Add coverage for:
- Language subtag has more than three letters.
- Lookup locale uses prefix matching.
- Unicode extension is present.
- Unicode extension is empty or the string "true".
- "latn" is returned for an unsupported locale.

Also update existing tests and test meta data to match current spec text.


anba:
- `preferred-from-unicode-extension-true-empty.js` fails in JSC and V8 because of an ICU4C-related issue.

PR #5024, opened 2026-04-13
update esvu

devsnek:
i bumped to a new minor in the last release


ptomato (approved):
Thanks for the maintenance work on esvu!

PR #5025, opened 2026-04-14
Add test for Number.MIN_VALUE's value

Sharktheone:
There where no tests in place to actually test what `Number.MIN_VALUE` is. The testsuite would happily accept any value, negative or high positive ones.

I dividing the `Number.MIN_VALUE` by 2 and check if it is 0 as the spec mentions denormalized values which we need to account for here.

I checked that the test passes in Yavashark (my own JS Engine) and from looking at all other Engines they should pass this new test too.

### DIFF
--- a/test/built-ins/Number/MIN_VALUE/value.js
+++ b/test/built-ins/Number/MIN_VALUE/value.js
@@ -1,0 +1,20 @@
+// Copyright (C) 2026 Luna Pfeiffer. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-number.min_value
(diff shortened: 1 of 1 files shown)


ljharb (approved):

Sharktheone:
What is missing here to get this PR merged? (Mainly so I can tick it off in my todo-list)

Same for my 2 other PRs (#5026 and #5027)

ptomato (approved):

ptomato:
For #5027, someone will need to find time to review it, as it isn't simple enough to be done in a short time like the `MAX_VALUE` and `MIN_VALUE` tests.

PR #5027, opened 2026-04-14
Add tests for return in `for..in` statements

Sharktheone:
The ability to return from `for..in` loops was not tested. I'm adding them here. I am testing the same cases as for the `for..of` loop.

The test is passing in Yavashark from which I also assume they will pass in any other engine.


bakkot (approved):
I am fine having these tests given that you already wrote them and presumably actually needed them, but it's not practical to have coverage testing all combinations of arbitrary statements; I don't think we generally want to try for that.

(When I was implementing try/finally semantics in an engine I ended up writing a fuzzer, which found something like a dozen distinct and occasionally quite complicated ways that my initial implementation was wrong. We wouldn't want to add tests for all of those; they were pretty specific to my implementation.)

Sharktheone:
The main reason was since I got this wrong in my engine and returns did not work in for in. The catch. finally and try are mainly just copied from how it is tested in `for..of`

PR #5028, opened 2026-04-15
Fix check in break tests

lebe-g:
Change "&&" to "||" in the test condition of some break tests.

The reason is that violating one of the individual conditions would already indicate a bug.


ptomato (commented):
Could we try to modernize these while we're here? e.g.

```js
assert.sameValue(x, 2, '#1 - x');
assert.sameValue(y, 0, '#1 - y');
```
(obviously, if there are better assertion messages to be written that describe what failed, use those 😄)

PR #5030, opened 2026-04-15
Add deferred-reexports to features.txt

caiolima:
This PR adds a new flag for tests coming to [deferred re-exports](https://github.com/tc39/proposal-deferred-reexports) proposal. This flag is going to be used by tests that will be included by https://github.com/tc39/test262/issues/5010.

### DIFF
--- a/features.txt
+++ b/features.txt
@@ -63,7 +63,7 @@ import-defer
 
 # Deferred re-exports
 # https://tc39.es/proposal-deferred-reexports/
-deferred-reexports
+export-defer
 
 # Import Text
 # https://github.com/tc39/proposal-import-text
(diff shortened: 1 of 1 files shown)


ljharb (approved):

nicolo-ribaudo (approved):

nicolo-ribaudo:
Actually, what do you think about using `export-defer`, like `import-defer`?

caiolima:
> Actually, what do you think about using `export-defer`, like `import-defer`?

Agree. I've changed it.

nicolo-ribaudo (approved):

caiolima:
Thank you very much for reviews!

PR #5011, opened 2026-04-02
iterator chunking tests

michaelficarra:
Adds tests for https://github.com/tc39/proposal-iterator-chunking. Includes a second commit that assumes https://github.com/tc39/proposal-iterator-chunking/pull/30 will be merged. We can back it out or modify it if that PR is rejected or modified.

Disclaimer: I used AI in the process of creating this PR.

PR #5031, opened 2026-04-15
iterator includes tests

michaelficarra:
Adds tests for https://github.com/michaelficarra/proposal-iterator-includes. 

Disclaimer: I used AI in the process of creating this PR (using the same process as in #5011, including a full review and minor tweaks).


test/built-ins/Iterator/prototype/includes/argument-effect-order.js
```diff
@@ -0,0 +1,67 @@
...
+features: [iterator-includes]
+---*/
+
+let effects = [];
+
+assert.throws(TypeError, function() {
+  Iterator.prototype.includes.call(null, 0, NaN);
+});
+
+assert.compareArray(effects, []);
```
bakkot, line 29: what could this possibly be asserting

test/built-ins/Iterator/prototype/includes/closes-on-match.js
```diff
@@ -0,0 +1,25 @@
...
+    return { done: false, value: i };
+  },
+  return() {
+    closed = true;
+    return {};
+  },
+};
+
+assert.sameValue(iter.includes(5), true);
+assert.sameValue(closed, true);
```
bakkot, line 25:
```suggestion
assert.sameValue(iter.includes(5), true);
assert.sameValue(closed, true);
assert.sameValue(i, 5);
```

test/built-ins/Iterator/prototype/includes/exhaustion-does-not-call-return.js
```diff
@@ -0,0 +1,29 @@
...
+    ++index;
+    if (index <= 3) {
+      return { done: false, value: index };
+    }
+    return { done: true, value: undefined };
+  },
+  return() {
+    ++returnCalls;
+    return {};
+  },
```
bakkot, line 25:
```suggestion
  get return() {
    throw new Test262Error('return should not be read');
  },
```
matches the style above and strengthens the assertion

test/built-ins/Iterator/prototype/includes/get-next-method-only-once.js
```diff
@@ -0,0 +1,29 @@
...
+/*---
+esid: sec-iterator.prototype.includes
+description: >
+  Gets the next method from the iterator only once
+features: [iterator-includes]
+---*/
+
+let nextGets = 0;
+
+class TestIterator extends Iterator {
```
bakkot, line 12: It's weird to switch to using class instead of `__proto__` for some of these, when that's not the thing under test. It made me think this was going to be testing something else. It would be better to stick to one style.

test/built-ins/Iterator/prototype/includes/get-return-method-throws.js
```diff
@@ -0,0 +1,27 @@
...
+
+  get return() {
+    throw new Test262Error();
+  }
+}
+
+let iterator = new IteratorThrows();
+
+assert.throws(Test262Error, function() {
+  iterator.includes(0);
```
bakkot, line 26: It is worth clarifying that this only happens because the search finishes before the iterator is exhausted (e.g. by first testing that `iterator.includes(1)` does not throw).

That said this is pretty redundant with the effects tests above - no one is going to be triggering the getter but not throwing the error.

test/built-ins/Iterator/prototype/includes/get-next-method-throws.js
```diff
@@ -0,0 +1,20 @@
...
+
+class IteratorThrows extends Iterator {
+  get next() {
+    throw new Test262Error();
+  }
+}
+
+let iterator = new IteratorThrows();
+
+assert.throws(Test262Error, function() {
```
bakkot, line 18: It would be good to also test that this does not get the `return` property.

test/built-ins/Iterator/prototype/includes/iterator-has-no-return.js
```diff
@@ -0,0 +1,21 @@
+// Copyright (C) 2026 Michael Ficarra. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-iterator.prototype.includes
+description: >
+  The underlying iterator may be unable to be closed (has no return method)
+features: [iterator-includes]
+---*/
+
+let iterator = [1, 2, 3, 4, 5][Symbol.iterator]();
```
bakkot, line 10:
```suggestion
let iterator = [1, 2, 3, 4, 5].values();
```
like you do every other time you call this method

test/built-ins/Iterator/prototype/includes/iterator-return-method-throws.js
```diff
@@ -0,0 +1,27 @@
...
+
+  return() {
+    throw new Test262Error();
+  }
+}
+
+let iterator = new IteratorThrows();
+
+assert.throws(Test262Error, function() {
+  iterator.includes(0);
```
bakkot, line 26: As above, it is worth clarifying that this only happens because the search finishes before the iterator is exhausted (e.g. by first testing that `iterator.includes(1)` does not throw).

test/built-ins/Iterator/prototype/includes/next-method-returns-non-object.js
```diff
@@ -0,0 +1,20 @@
...
+class NonObjectIterator extends Iterator {
+  next() {
+    return null;
+  }
+}
+
+let iterator = new NonObjectIterator();
+
+assert.throws(TypeError, function() {
+  iterator.includes(0);
```
bakkot, line 19: It would be good to also test that this does not get the `return` property.

test/built-ins/Iterator/prototype/includes/next-method-returns-throwing-done.js
```diff
@@ -0,0 +1,25 @@
...
+      },
+      value: 1,
+    };
+  }
+}
+
+let iterator = new ThrowingIterator();
+
+assert.throws(Test262Error, function() {
+  iterator.includes(0);
```
bakkot, line 24: It would be good to also test that this does not get the `return` property.

test/built-ins/Iterator/prototype/includes/next-method-returns-throwing-value.js
```diff
@@ -0,0 +1,25 @@
...
+        throw new Test262Error();
+      },
+    };
+  }
+}
+
+let iterator = new ThrowingIterator();
+
+assert.throws(Test262Error, function() {
+  iterator.includes(0);
```
bakkot, line 24: It would be good to also test that this does not get the `return` property.

test/built-ins/Iterator/prototype/includes/next-method-throws.js
```diff
@@ -0,0 +1,20 @@
...
+class ThrowingIterator extends Iterator {
+  next() {
+    throw new Test262Error();
+  }
+}
+
+let iterator = new ThrowingIterator();
+
+assert.throws(Test262Error, function() {
+  iterator.includes(0);
```
bakkot, line 19: It would be good to also test that this does not get the `return` property.

test/built-ins/Iterator/prototype/includes/object-identity.js
```diff
@@ -0,0 +1,15 @@
+// Copyright (C) 2026 Michael Ficarra. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-iterator.prototype.includes
+description: >
+  Includes compares objects by identity
```
bakkot, line 6: The only other way this could plausibly be comparing is by `==`, which this does not definitively rule out. For that you'd need a needle which is a string and an object with a `toString` (which you would assert not to be read).

bakkot (commented):
some nits but looks reasonably complete

test/built-ins/Iterator/prototype/includes/get-return-method-throws.js
michaelficarra: I don't know what you're looking for here. Doing `iterator.includes(1)` with this iterator would never finish.

bakkot: Make the iterator return `{ done: true }` on the second call to `next`.

michaelficarra:
@bakkot Feedback addressed.

test/built-ins/Iterator/prototype/includes/get-return-method-throws.js
```diff
@@ -0,0 +1,30 @@
...
+    } else {
+      return { done: true, value: undefined };
+    }
+  },
+  get return() {
+    throw new Test262Error();
+  }
+};
+
+assert.sameValue(iterator.includes(1), false);
```
bakkot, line 26:
```suggestion
assert.sameValue(iterator.includes(1), false);
counter = 0;
```

bakkot (approved):

michaelficarra:
Added a new commit for https://github.com/tc39/proposal-iterator-includes/pull/15.

kmiller68:
Would be good to merge this as it's been stage 3 for a few months.

PR #5041, opened 2026-04-29
[await-dictionary] add failure case and edge case tests

danialasaria:
## Summary

This PR adds the remaining failure-case and edge-case coverage for the await-dictionary proposal's Promise keyed combinators:

- `Promise.allKeyed`
- `Promise.allSettledKeyed`

It follows the runtime/behavioral tests PR (#4932) and completes the testing plan from #4886.

### NewPromiseCapability failures (sync throws)

- Constructor body throws (`ctx-ctor-throws.js`)
- Executor does not provide callable resolve/reject in various combinations (`capability-executor-not-callable.js`)

### GetPromiseResolve failures (async rejections)

- `Get(constructor, "resolve")` accessor throws (`invoke-resolve-get-error-reject.js`)
- Calling the resolved `promiseResolve` function throws (`invoke-resolve-error-reject.js`)

### Invoke(then) failures (async rejections)

- `Get(nextPromise, "then")` accessor throws (`invoke-then-get-error-reject.js`)
- Calling `nextPromise.then(...)` throws (`invoke-then-error-reject.js`)

### allKeyed rejection ordering

- Second promise to settle is rejected (`reject-second.js`)
- Last promise to settle is rejected (`reject-last.js`)

### Synchronous thenables (step 8 edge case)

- Thenables that synchronously invoke `onFulfilled` during the loop, causing `remainingElementsCount` to reach zero before the loop exits (`resolve-before-loop-exit.js`)

### Exotic object abrupt completions (Proxy)

- `[[OwnPropertyKeys]]` throws (`ownkeys-throws.js`)
- `[[GetOwnProperty]]` throws (`getownproperty-throws.js`)
- `[[GetOwnProperty]]` returns `undefined` -- key is skipped (`getownproperty-returns-undefined.js`)

### Property descriptor filtering

- Non-enumerable keys are skipped (`getownproperty-not-enumerable.js`)

## Notes

- All async tests use `asyncTest` from `asyncHelpers.js` per reviewer feedback on #4932
- Sync tests use `assert.throws(...)`
- All tests are feature-gated with `features: [await-dictionary]`; Proxy tests additionally gate on `Proxy`
- Tests mirror existing patterns from `Promise.all` / `Promise.allSettled` test suites
- `capability-executor-not-callable` includes poisoned `resolve` getters (matching `Promise.all` counterpart) to detect wrong evaluation order

## What this PR does not cover

This completes all items from the testing plan (#4886).

## References

- Testing plan issue: #4886
- Prior runtime tests PR: #4932
- Prior boilerplate PR: #4927
- Proposal text: [Await Dictionary](https://tc39.es/proposal-await-dictionary/)


test/built-ins/Promise/allSettledKeyed/resolve-before-loop-exit.js
```diff
@@ -0,0 +1,79 @@
...
+      onFulfilled("c-fulfill");
+    }
+  }
+};
+
+assert.sameValue(callCount, 0, "callCount before call to allSettledKeyed()");
+
+Promise.allSettledKeyed.call(Constructor, input);
+
+assert.sameValue(callCount, 1, "callCount after call to allSettledKeyed()");
```
acutmore: Could we either move the callCount to after the assertions, or have a separate count for the start and end of the callback. Otherwise technically the assertions could reject and they are swallowed by the function caller yet the test would pass (unless the harness has some tracking logic to guarantee a test fails if a harness assertion errors, rather than relying on the exception bubbling up)

acutmore (commented):
Could we also have a test that a custom constructor is being 'constructed' rather than 'called'.

If we can use `new.target` we can check like:

```js
var error = new Test262Error();
function Constructor(executor) {
  if (new.target !== Constructor) {
    throw error
  }
}
```

We can also have assertions that validate the arguments passed to a custom constructor are correct: should only be one argument: the executor. And it should be a function with length 2.

test/built-ins/Promise/allSettledKeyed/getownproperty-throws.js
```diff
@@ -0,0 +1,42 @@
...
+includes: [asyncHelpers.js]
+flags: [async]
+features: [await-dictionary, Proxy]
+---*/
+
+var error = new Test262Error();
+var input = new Proxy({ key: 1 }, {
+  ownKeys: function() {
+    return ['key'];
+  },
```
acutmore: not harmful but also feels superfluous to the test as `ownKeys` will already be returning this.

test/built-ins/Promise/allSettledKeyed/getownproperty-returns-undefined.js
```diff
@@ -13,26 +13,23 @@ info: |
...
-  visible: Promise.resolve(2)
-};
-
-Object.defineProperty(input, 'hidden', {
-  enumerable: false,
-  value: Promise.resolve(1)
+var input = new Proxy({ key: Promise.resolve(1) }, {
+  ownKeys: function() {
+    return ['key'];
+  },
```
acutmore: Same comment as: https://github.com/tc39/test262/pull/5041/changes#r3241633686

danialasaria:
> Could we also have a test that a custom constructor is being 'constructed' rather than 'called'.
> 
> If we can use `new.target` we can check like:
> 
> ```js
> var error = new Test262Error();
> function Constructor(executor) {
>   if (new.target !== Constructor) {
>     throw error
>   }
> }
> ```
> 
> We can also have assertions that validate the arguments passed to a custom constructor are correct: should only be one argument: the executor. And it should be a function with length 2.

Added ctx-ctor-constructed.js coverage for both Promise.allKeyed and Promise.allSettledKeyed. The custom constructor checks new.target, verifies it receives exactly one argument, confirms the executor is callable, and asserts executor.length === 2.

test/built-ins/Promise/allSettledKeyed/resolve-before-loop-exit.js
danialasaria: Fixed by replacing the single counter with resolveStartCount and resolveEndCount. The end counter is incremented only after the resolve callback assertions complete, so an assertion failure inside the callback won’t be hidden by the final count check.

test/built-ins/Promise/allSettledKeyed/getownproperty-throws.js
danialasaria: Removed the explicit ownKeys trap. The Proxy target already has an own enumerable key property, so default [[OwnPropertyKeys]] still exposes the key while the test remains focused on the custom getOwnPropertyDescriptor behavior.

test/built-ins/Promise/allSettledKeyed/getownproperty-returns-undefined.js
danialasaria: Fixed here as well--removed the redundant ownKeys trap

test/built-ins/Promise/allKeyed/ctx-ctor-constructed.js
```diff
@@ -0,0 +1,35 @@
...
+
+  ...
+  6. Let promise be Construct(C, « executor »).
+features: [await-dictionary]
+---*/
+
+var error = new Test262Error();
+
+function Constructor(executor) {
+  if (new.target !== Constructor) {
```
Ms2ger, line 25: Probably you should add `new.target` to the `features` in the preamble.

test/built-ins/Promise/allKeyed/getownproperty-not-enumerable.js
```diff
@@ -0,0 +1,48 @@
...
+Object.defineProperty(input, "c", {
+  value: Promise.resolve(3),
+  enumerable: true,
+  configurable: true
+});
+
+asyncTest(function() {
+  return Promise.allKeyed(input).then(function(result) {
+    assert.sameValue(Object.getPrototypeOf(result), null, "result is null-prototype");
+    var keys = Reflect.ownKeys(result);
```
Ms2ger, line 41: Add the `Reflect` feature

test/built-ins/Promise/allKeyed/getownproperty-returns-undefined.js
```diff
@@ -0,0 +1,35 @@
...
+  },
+  getOwnPropertyDescriptor: function() {
+    return undefined;
+  }
+});
+
+asyncTest(function() {
+  return Promise.allKeyed(input).then(function(result) {
+    assert.sameValue(Object.getPrototypeOf(result), null, "result is null-prototype");
+    assert.sameValue(Reflect.ownKeys(result).length, 0);
```
Ms2ger: Ditto

test/built-ins/Promise/allKeyed/invoke-then-error-reject.js
```diff
@@ -0,0 +1,39 @@
...
+  PerformPromiseAllKeyed ( variant, promises, constructor, resultCapability, promiseResolve )
+
+  ...
+  6. For each element key of allKeys, do
+    ...
+    b. If desc is not undefined and desc.[[Enumerable]] is true, then
+      ...
+      iv. Let nextPromise be ? Call(promiseResolve, constructor, « nextValue »).
+      ...
+      11. Perform ? Invoke(nextPromise, "then", « onFulfilled, onRejected »).
```
Ms2ger:
```suggestion
      xiii. Perform ? Invoke(nextPromise, "then", « onFulfilled, onRejected »).
```
as far as I can tell

test/built-ins/Promise/allKeyed/invoke-then-get-error-reject.js
```diff
@@ -0,0 +1,39 @@
...
+  PerformPromiseAllKeyed ( variant, promises, constructor, resultCapability, promiseResolve )
+
+  ...
+  6. For each element key of allKeys, do
+    ...
+    b. If desc is not undefined and desc.[[Enumerable]] is true, then
+      ...
+      iv. Let nextPromise be ? Call(promiseResolve, constructor, « nextValue »).
+      ...
+      11. Perform ? Invoke(nextPromise, "then", « onFulfilled, onRejected »).
```
Ms2ger:
```suggestion
      xiii. Perform ? Invoke(nextPromise, "then", « onFulfilled, onRejected »).
```

Ms2ger (commented):
Didn't get through it yet, but here are some initial comments.

test/built-ins/Promise/allSettledKeyed/getownproperty-throws.js
Ms2ger: I'm still seeing the ownKeys trap. Did you forget to push?

Ms2ger (approved):
Please address the previous comments in the analogous `allSettledKeyed` tests as well.

danialasaria:
> Please address the previous comments in the analogous `allSettledKeyed` tests as well.

Should be all updated now!

PR #5044, opened 2026-05-06
Temporal: Add tests for duration rounding edge case near +24h UTC shift

ptomato:
This edge case was discovered thanks to Firefox fuzzing: https://bugzilla.mozilla.org/show_bug.cgi?id=2029455

See https://github.com/tc39/proposal-temporal/issues/3310

LLM disclosure: I used a bot to make a first draft of these, then hand-edited them until I was satisfied.


ptomato:
Don't merge this yet. These tests currently hit an assertion failure in the spec. We need to decide what to do with it.

ptomato:
Just updated this with more test cases.

PR #5046, opened 2026-05-07
Fix a harmless bug in test/built-ins/TypedArray/prototype/includes/searchelement-not-integer.js

gibson042:
`throw new Test262Error()` does what it appears to do, but `throw Test262Error()` throws `undefined`.

### DIFF
--- a/test/built-ins/TypedArray/prototype/includes/searchelement-not-integer.js
+++ b/test/built-ins/TypedArray/prototype/includes/searchelement-not-integer.js
@@ -18,13 +18,7 @@ includes: [testTypedArray.js]
 features: [TypedArray]
 ---*/
 
-function CustomError() {}
-
-function throwError() {
-  throw new CustomError();
-}
-
 testWithTypedArrayConstructors(function(TA, makeCtorArg) {
   var sample = new TA(makeCtorArg(10));
-  assert.sameValue(sample.includes({ valueOf: throwError }), false);
+  assert.sameValue(sample.includes({ valueOf: Test262Error.thrower }), false);
 });
--- a/test/built-ins/TypedArray/prototype/includes/searchelement-not-integer.js
+++ b/test/built-ins/TypedArray/prototype/includes/searchelement-not-integer.js
@@ -18,12 +18,13 @@ includes: [testTypedArray.js]
 features: [TypedArray]
 ---*/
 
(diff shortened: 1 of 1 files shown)


ljharb:
I'm confused, why? we use Test262Error all over the place, don't we?

gibson042:
Clarified the description—there's a difference between `throw new Test262Error()` and `throw Test262Error()`, and the former is much more useful.

ljharb:
since for all builtin error types, they're identical, why don't we just fix Test262Error to work properly?

gibson042:
I'm fine with that as well. And separately, I'd also like to move away from test code manually creating Test262Error instances, which should be reserved for our `assert` framework rather than used in arbitrary spying.

test/built-ins/TypedArray/prototype/includes/searchelement-not-integer.js
```diff
@@ -18,12 +18,13 @@ includes: [testTypedArray.js]
...
+
 testWithTypedArrayConstructors(function(TA, makeCtorArg) {
   var sample = new TA(makeCtorArg(10));
-  function throwFunc(){
-    throw Test262Error()
-    return 0;
-  }
-
-    assert.sameValue(sample.includes({valueOf : throwFunc}), false);
+  assert.sameValue(sample.includes({ valueOf: throwError }), false);
```
gibson042: Since we expect the `valueOf` function to _not_ be called, this could also be
```suggestion
  assert.sameValue(sample.includes({ valueOf: Test262Error.thrower }), false);
```
or
```suggestion
  const nonCoercible = {
    valueOf() {
      Test262Error.thrower("unexpected valueOf()");
    }
  };
  assert.sameValue(sample.includes(nonCoercible), false);
```
or if we introduce another level of indirection,
```suggestion
  const nonCoercible = {
    valueOf: Test262Error.makeThrower("unexpected valueOf()")
  };
  assert.sameValue(sample.includes(nonCoercible), false);
```

ljharb: I like `.thrower`, that's nice and clean

gibson042: Updated.

ljharb (approved):

PR #5047, opened 2026-05-08
Temporal: Add ZonedDateTime tests for repeated midnight

ptomato:
In an Antarctica time zone, in 2010, there was a time zone shift that went from 02:00 to 23:00, repeating midnight twice. This was handled in ZonedDateTime.prototype.startOfDay() and .hoursInDay, but not in .round({ smallestUnit: 'days' }), where the spec text had an assertion. We'll remove that assertion as it is faulty. This PR adds test coverage for the situation.

See: https://github.com/tc39/proposal-temporal/issues/3312


ptomato:
This is an in-progress draft until https://github.com/tc39/proposal-temporal/issues/3312 is fully solved.

ptomato:
The needs-consensus PR reached consensus in the TC39 meeting of 2026-05-19.

Ms2ger (approved):

PR #5048, opened 2026-05-10
Add decorator tests

bennypowers:
Add behavioral tests for the TC39 decorators proposal, covering:

- Decorator context object (`kind`, `name`, `access`, `static`, `private`)
- Return value handling for class, method, getter, setter, field, and accessor decorators
- Decorator ordering (application order, initialization phases)
- Class decorator rebinding and proxy/subclass returns
- Field and accessor initializer `this` binding
- Invalid return value error cases
- Computed member expression syntax errors (`@decorators[0]`)
- Constructor decoration syntax errors
- `addInitializer` error cases (non-callable argument, call after decoration)
- Async generator method decorator support

Tests use test262's template/case generation system to cover all
combinations of class declarations/expressions, public/private,
instance/static, and standard/with-init variants.

Based on #4103 by @pzuraq, with fixes for:
- Class expression error template decorating a method instead of the class
- Generator method template path collisions with standard method templates
- Getter template spurious `error/` path segment
- Copy-paste description errors across multiple case files
- Inverted `context-private-true/false` filenames
- Wrong `esid` in class expression template
- "acessor" typo in accessor template names
- Missing generator/async qualifiers in template name fields
- Unused variable in proxy test case

Validated test behavior against [ecma262 PR #2417](https://github.com/tc39/ecma262/pull/2417)
and [V8's decorator implementation](https://chromium-review.googlesource.com/c/v8/v8/+/7571048).

Co-Authored-By: Kristen Maevyn <kris@bitski.com>


bennypowers:
Would maintainers prefer this be split into smaller PRs?

ptomato:
@bennypowers Thanks very much for this. Smaller PRs would be better for reviews. But note that decorators are on the agenda in next week's TC39 meeting (https://github.com/tc39/agendas/blob/main/2026/05.md) so please wait and see what comes out of that discussion before doing any substantial further work on this PR, in case it would be wasted.

ayden94:
Following up on the earlier “please wait for the May TC39 meeting” note: I checked the May agenda and current public proposal status again. I do not see public meeting notes/outcome linked from the proposals table yet, and Decorators still appears to be listed as Stage 2.7.

I prepared a very small local split candidate from this PR, but have not opened a PR because I want to coordinate first and avoid duplicating or disrupting this work. The candidate contains four non-generated tests:

- class declaration/expression cases for `context.addInitializer` throwing when called after decoration has completed;
- class declaration/expression negative syntax cases for computed decorator member expressions like `@decorators[0]`.

Local validation I ran:

- applied the patch to current `tc39/test262` `main` with `git apply --check`;
- ran `python3 tools/lint/lint.py` on the four files: `0 error(s) found`;
- ran the two runtime `addInitializer` tests with engine262 using `--features=decorators --test262-harness`.

This was prepared with AI assistance from the existing #5048/#4103 work, so I would preserve the original copyright/authorship/co-author metadata and would mention the assistance in the PR if opened.

Would a small split PR like this be useful now, or should I keep waiting until the May meeting outcome is publicly documented / maintainers give a different direction?

sandstrom:
Great with tests for decorators! 😄 

I've been waiting for browsers and tooling to add proper decorator support, and this seems like a great step in that direction.

I guess it could also be used to verify deno and bun support for decorators, to make sure they match the spec.

PR #5049, opened 2026-05-10
[immutable-arraybuffer] Correct expected length for ArrayBuffer.prototype.sliceToImmutable

gibson042:
Ref https://github.com/tc39/test262/pull/4553#discussion_r3215248072


iamstolis:
Looks good to me. Thank you for a quick fix.

ptomato (approved):

PR #5050, opened 2026-05-11
Verify argument counts when forwarding dispose/asyncDispose to return

rbuckton:
There was no validation for argument count in the tests for when either `Iterator.prototype[Symbol.dispose]()` or `AsyncIterator.prototype[Symbol.asyncDispose]()` invoke `this.return();`. 

This currently expects both to be `0`, to match the Needs Consensus PR https://github.com/rbuckton/ecma262/pull/17. If this PR does not advance, this test can be adjusted to expect `0` for `Iterator.prototype[Symbol.dispose]()` and `1` for `AsyncIterator.prototype[Symbol.asyncDispose]()`.

Related: rbuckton/ecma262#17
Related: tc39/proposal-explicit-resource-management#273

### DIFF
--- a/test/built-ins/AsyncIteratorPrototype/Symbol.asyncDispose/invokes-return.js
+++ b/test/built-ins/AsyncIteratorPrototype/Symbol.asyncDispose/invokes-return.js
@@ -13,7 +13,7 @@ info: |
   5. If return is undefined, then
     a. Perform ! Call(promiseCapability.[[Resolve]], undefined, « undefined »).
   6. Else,
-    a. Let result be Call(return, O, « undefined »).
+    a. Let result be Call(return, O, « »).
     b. IfAbruptRejectPromise(result, promiseCapability).
     c. Let resultWrapper be Completion(PromiseResolve(%Promise%, result)).
     d. IfAbruptRejectPromise(resultWrapper, promiseCapability).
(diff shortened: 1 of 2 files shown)


ljharb (approved):

test/built-ins/AsyncIteratorPrototype/Symbol.asyncDispose/invokes-return.js
```diff
@@ -34,11 +34,14 @@ asyncTest(async function () {
...
+  var argumentCount = 0;
   iter.return = async function () {
+    argumentCount = arguments.length;
     returnCalled = true;
     return { done: true };
   };
 
   await iter[Symbol.asyncDispose]();
   assert.sameValue(returnCalled, true);
+  assert.sameValue(argumentCount, 0);
```
ptomato, line 46: Optional: I think you can just assert `arguments.length` inside the function and the test will still fail. `returnCalled` will guarantee that the assertion was executed. But if you like it better this way, that's fine too.

ptomato (approved):

test/built-ins/AsyncIteratorPrototype/Symbol.asyncDispose/invokes-return.js
rbuckton: IMO, I generally prefer to keep assertions grouped together in the Arrange-Act-Assert style as it makes the test implications and expectations clearer.

rbuckton:
FYI, the associated spec text reached consensus at the May 2026 plenary and has been merged into the main spec PR.

PR #5051, opened 2026-05-13
Intl Era Month Code: Test not-yet-supported calendars in Intl.DateTimeFormat and Temporal

ptomato:
As per https://github.com/tc39/ecma402/blob/main/meetings/notes-2025-12-04.md#datetimeformatconstructor-options-calendar-islamic-fallbackjs-should-allow-other-fallback-values-4677 the supported calendars in Intl and Temporal must be a closed set for implementations conforming to ECMA-402. Unlike most other Intl entries such as number formats, implementations may not support extra calendars until they've gone through the process to define month codes and era codes interoperably.

Apple's fork of ICU contains code for some South Asian calendars not in other versions of ICU. These calendars therefore form good test data for the above, since they technically could be supported in implementations using that version of ICU in the future, but must not (yet) be.

See: https://github.com/tc39/proposal-intl-era-monthcode/issues/128

### DIFF
--- a/harness/temporalHelpers.js
+++ b/harness/temporalHelpers.js
@@ -113,6 +113,23 @@ var TemporalHelpers = {
     ],
   },
 
+
+  /**
+   * Apple's fork of ICU contains code for these calendars, but they are not yet
+   * allowed to be supported in AvailableCalendars. See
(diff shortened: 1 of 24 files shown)


gibson042 (approved):
LGTM. You should consider defining `notYetSupportedCalendars` once, in a harness/ helper.

Ms2ger (approved):

PR #5061, opened 2026-05-25
Test that %Intl%.[[FallbackSymbol]] is per-realm

ptomato:
Adds a test that ensures the Intl fallback symbol is not shared between realms. Also while we are here adds verification of the property descriptor for legacy-constructed DateTimeFormat and NumberFormat instances.

### DIFF
--- a/test/intl402/DateTimeFormat/intl-legacy-constructed-symbol-on-unwrap.js
+++ b/test/intl402/DateTimeFormat/intl-legacy-constructed-symbol-on-unwrap.js
@@ -12,21 +12,14 @@ features: [intl-normative-optional]
 let object = new Intl.DateTimeFormat();
 let newObject = Intl.DateTimeFormat.call(object);
 let symbol = null;
-let error = null;
-try {
-    let proxy = new Proxy(newObject, {
-        get(target, property) {
-            symbol = property;
-            return target[property];
-        }
-    });
-    Intl.DateTimeFormat.prototype.resolvedOptions.call(proxy);
-} catch (e) {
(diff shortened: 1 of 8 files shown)


linusg (commented):
all other `intl-normative-optional` tests pass when the optional behavior is **not** implemented, they should either be updated to fail or these new tests pass gracefully too. It's confusing if half of them pass and the other half don't in absence of the feature.

ptomato:
Fair enough. It seems to me they should all fail, then. That's how the other features work. If an engine doesn't implement that feature they can filter it out.

linusg:
Thanks!

test/intl402/DateTimeFormat/intl-legacy-constructed-symbol-on-unwrap.js
```diff
@@ -13,20 +13,14 @@ let object = new Intl.DateTimeFormat();
 let newObject = Intl.DateTimeFormat.call(object);
 let symbol = null;
 let error = null;
```
Ms2ger: Remove this line

test/intl402/NumberFormat/intl-legacy-constructed-symbol-on-unwrap.js
```diff
@@ -13,20 +13,14 @@ let object = new Intl.NumberFormat();
 let newObject = Intl.NumberFormat.call(object);
 let symbol = null;
 let error = null;
```
Ms2ger: Remove this line

Ms2ger (approved):
Thanks!

PR #5074, opened 2026-06-25
[await-dictionary] add remaining keyed Promise tests

danialasaria:
Adds explicit coverage for the remaining unchecked await-dictionary keyed Promise test plan cases and proposal-specific algorithm edges:

- objects with own keys but no enumerable properties
- property value access failures and skipped non-enumerable getters
- missing `Promise.resolve`, `promiseResolve` call counts, custom receiver use, and returned thenable use
- first-of-multiple `Promise.allKeyed` rejection ordering
- `promiseResolve` returning an object without callable `then`
- result resolve edge cases
- resolve/reject element callback behavior and callback function properties
- keyed result and allSettled entry property descriptors

Refs: #4886


test/built-ins/Promise/allKeyed/resolve-from-same-thenable.js
```diff
@@ -0,0 +1,49 @@
...
+  Cannot tamper remainingElementsCount when the same resolve element function is
+  called multiple times.
+info: |
+  PerformPromiseAllKeyed ( variant, promises, constructor, resultCapability, promiseResolve )
+
+  ...
+  The onFulfilled closure:
+    ...
+    2. If alreadyCalled.[[Value]] is true, return undefined.
+    3. Set alreadyCalled.[[Value]] to true.
```
Ms2ger:
```suggestion
    2. If activeFunc.[[AlreadyCalled]].[[Value]] is true, return undefined.
    3. Set activeFunc.[[AlreadyCalled]].[[Value]] to true.
```

test/built-ins/Promise/allSettledKeyed/reject-from-same-thenable.js
```diff
@@ -0,0 +1,51 @@
...
+  Cannot tamper remainingElementsCount when an allSettledKeyed reject element
+  function is called before later calls.
+info: |
+  PerformPromiseAllKeyed ( variant, promises, constructor, resultCapability, promiseResolve )
+
+  ...
+  The onRejected closure:
+    ...
+    2. If alreadyCalled.[[Value]] is true, return undefined.
+    3. Set alreadyCalled.[[Value]] to true.
```
Ms2ger: Ditto

Ms2ger (approved):
Thanks!

Ms2ger (approved):

PR #5075, opened 2026-06-26
Do not test processing of the removed parameter of Atomics.pause().

iamstolis:
`Atomics.pause()` had an integer parameter in some older versions of the specification. This parameter was removed later, but the corresponding test remained in `test262`. This PR just removes the test that is no longer valid.


ljharb (approved):

linusg (changes requested):
https://github.com/tc39/test262/blob/de8e621/test/built-ins/Atomics/pause/returns-undefined.js should be updated too.

iamstolis:
> https://github.com/tc39/test262/blob/de8e621/test/built-ins/Atomics/pause/returns-undefined.js should be updated too.

Sure, I can remove the part that invokes `Atomics.pause()` with some arguments, but strictly speaking `returns-undefined.js` is correct (in the sense that it does not assert anything not requested by the current version of the specification).

linusg (approved):
Thanks :)

gibson042 (approved):

PR #5078, opened 2026-07-02
Add tests for abrupt completion of the then call in NewPromiseResolveThenableJob

alii:
There's no coverage for NewPromiseResolveThenableJob step 1.c (reject the promise when the then call throws). The existing thenable tests only poison the then getter, use a custom then that resolves, or throw after resolving.

The two species tests also catch a real engine bug: JSC never settles the promise when constructor[Symbol.species] throws inside the job (https://bugs.webkit.org/show_bug.cgi?id=318399, fix in https://github.com/WebKit/WebKit/pull/68487). They fail on every current JSC and pass on V8 and on JSC trunk with that fix. The other two pass everywhere and just fill the coverage gap.


ptomato:
Thanks! You may want to consider using the `asyncHelpers` harness (https://github.com/tc39/test262/blob/main/rfcs/async-helpers.md) to see if that makes these tests any simpler.

alii:
Thank you! Updated to use them.

PR #5079, opened 2026-07-02
Fix unit related hardcoded patterns for AppleICU

Constellation:
ECMA-402 allows customized formatting for i18n output, and these tests are relying on a particular format in ICU, which is changed in AppleICU to align the output to Apple HIG. This patch makes them tolerant against the difference between ICU and AppleICU output.

### DIFF
--- a/test/intl402/NumberFormat/prototype/format/unit-ja-JP.js
+++ b/test/intl402/NumberFormat/prototype/format/unit-ja-JP.js
@@ -6,56 +6,63 @@ esid: sec-intl.numberformat.prototype.format
 description: Checks handling of the unit style.
 locale: [ja-JP]
 features: [Intl.NumberFormat-unified]
+includes: [testIntlNumberFormat.js]
 ---*/
 
+// The unit/number separator is CLDR data, not specified by ECMA-402, so
+// read it from the implementation instead of hardcoding it.
+// (e.g. "時速 987 キロメートル" vs. "時速987キロメートル")
+const shortSep = getUnitSeparators("ja-JP", "short");
+const narrowSep = getUnitSeparators("ja-JP", "narrow");
+const longSep = getUnitSeparators("ja-JP", "long");
 
 const tests = [
   [
     -987,
     {
-      "short": "-987 km/h",
-      "narrow": "-987km/h",
-      "long": "時速 -987 キロメートル",
+      "short": `-987${shortSep.suffix}km/h`,
+      "narrow": `-987${narrowSep.suffix}km/h`,
+      "long": `時速${longSep.prefix}-987${longSep.suffix}キロメートル`,
     }
   ],
   [
     -0.001,
     {
-      "short": "-0.001 km/h",
-      "narrow": "-0.001km/h",
-      "long": "時速 -0.001 キロメートル",
+      "short": `-0.001${shortSep.suffix}km/h`,
+      "narrow": `-0.001${narrowSep.suffix}km/h`,
+      "long": `時速${longSep.prefix}-0.001${longSep.suffix}キロメートル`,
     }
   ],
   [
     -0,
     {
-      "short": "-0 km/h",
-      "narrow": "-0km/h",
-      "long": "時速 -0 キロメートル",
+      "short": `-0${shortSep.suffix}km/h`,
+      "narrow": `-0${narrowSep.suffix}km/h`,
+      "long": `時速${longSep.prefix}-0${longSep.suffix}キロメートル`,
     }
   ],
   [
     0,
     {
-      "short": "0 km/h",
-      "narrow": "0km/h",
-      "long": "時速 0 キロメートル",
+      "short": `0${shortSep.suffix}km/h`,
+      "narrow": `0${narrowSep.suffix}km/h`,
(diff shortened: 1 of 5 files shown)


test/intl402/NumberFormat/prototype/format/unit-ja-JP.js
```diff
@@ -8,54 +8,73 @@ locale: [ja-JP]
 features: [Intl.NumberFormat-unified]
 ---*/
 
+function unitSeparators(unitDisplay) {
```
Ms2ger: I'm finding these helper functions quite hard to follow. Could you add some documentation on what they're (supposed to be) doing?

Constellation: It is extracting the prefix and suffix around the number generated with the unit. Some ICU variants (e.g. AppleICU, or some version of ICU) puts prefix / suffix like special spaces etc., thus this function is extracting that to construct an expectation.
It is typically a space, sometimes a bit special space character, but this function is handling it in a tolerant way.

Constellation (commented):
Added comments and examples.

test/intl402/NumberFormat/prototype/format/unit-zh-TW.js
```diff
@@ -8,54 +8,76 @@ locale: [zh-TW]
 features: [Intl.NumberFormat-unified]
 ---*/
 
+// The unit/number separator is CLDR data, not specified by ECMA-402, so
+// read it from the implementation instead of hardcoding it.
+// (e.g. "-987 公里/小時" vs. "-987公里/小時")
+function unitSeparators(unitDisplay) {
```
gibson042: I'm seeing many additions of this kind of function, and would prefer it be generalized into `function getUnitSeparators(locale, unitDisplay)` into a harness file (suggestion: harness/testIntlNumberFormat.js) as e.g.
```js
/**
 * @description Determine the unit/number separator(s) used by an
 *   Intl.NumberFormat implementation, which are not specified by ECMA-402 and
 *   generally come from CLDR (and observable in e.g. Japanese long display
 *   `時速 2 キロメートル` vs. `時速2キロメートル`).
 * @param {string} locale
 * @param {'narrow' | 'short' | 'long'} unitDisplay
 * @return {{ prefix: string, suffix: string }}
 */
function getUnitSeparators(locale, unitDisplay) {
  var nf = new Intl.NumberFormat(locale, { style: "unit", unit: "kilometer-per-hour", unitDisplay: unitDisplay });
  var parts = nf.formatToParts(1);
  var unitIndices = [];
  for (var i = 0; i < parts.length; i++) {
    if (parts[i].type === "unit") unitIndices.push(i);
  }

  var prefix = unitIndices[0] === 0 && parts[1] && parts[1].type === "literal"
    ? parts[1].value
    : "";

  const lastUnitIndex = unitIndices[unitIndices.length - 1];
  const suffix = lastUnitIndex === parts.length - 1 && parts[lastUnitIndex - 1] && parts[lastUnitIndex - 1].type === "literal"
    ? parts[lastUnitIndex - 1].value
    : "";

  return { prefix: prefix, suffix: suffix };
}
```

test/intl402/NumberFormat/prototype/formatToParts/unit-ja-JP.js
```diff
@@ -19,79 +19,81 @@ function verifyFormatParts(actual, expected, message) {
...
   ],
 ];
 
-for (const [number, expectedData] of tests) {
+for (const [number, numberParts] of tests) {
+  const expectedData = {
+    "short": suffixUnitParts(numberParts, "km/h", shortSep),
+    "narrow": suffixUnitParts(numberParts, "km/h", narrowSep),
+    "long": prefixSuffixUnitParts(numberParts, "時速", "キロメートル", longSep),
+  };
```
gibson042, line 75: Suggestion: consolidate `suffixUnitParts` and `prefixSuffixUnitParts`.
```suggestion
  const expectedData = {
    "short": addUnitParts(numberParts, shortSep, "", "km/h"),
    "narrow": addUnitParts(numberParts, narrowSep, "", "km/h"),
    "long": addUnitParts(numberParts, longSep, "時速", "キロメートル"),
  };
```
```js
function makePart(type, value) {
  return value ? [{ type: type, value: value }] : [];
}

function addUnitParts(numberParts, separators, prefixText, suffixText) {
  return [].concat(
    makePart("unit", prefixText),
    makePart("literal", separators.prefix),
    numberParts,
    makePart("literal", separators.suffix),
    makePart("unit", suffixText),
  );
}
```
(and likewise in test/intl402/NumberFormat/prototype/formatToParts/unit-zh-TW.js)

test/intl402/NumberFormat/prototype/format/unit-zh-TW.js
Constellation: That sounds good!

test/intl402/NumberFormat/prototype/formatToParts/unit-ja-JP.js
Constellation: Nice, added that.

Constellation:
@gibson042 @Ms2ger Thanks! Updated.

gibson042 (approved):

PR #5080, opened 2026-07-06
Test fallbacks in RegionPreference

ptomato:
Intl.Locale.p.getCalendars and getHourCycles call RegionPreference which has several untested steps. The "sd" extension must be taken into account, the likely region must be looked up, and the "rg" extension must override the given region if locale data exists for it.

LLM disclosure: I used a bot to draft these tests, then hand-edited them until I was satisfied.

Note: I found it a bit hard to understand the observable implications of RegionPreference, so I left more of the LLM overexplanation intact in the comments than I usually would. I did edit the prose to be less chirpy and verified that it matches my understanding. If you'd prefer to have less of it, I'm happy to edit the comments down further or remove them.

### DIFF
--- a/test/intl402/Locale/prototype/getCalendars/likely-subtags-region.js
+++ b/test/intl402/Locale/prototype/getCalendars/likely-subtags-region.js
@@ -1,0 +1,66 @@
+// Copyright 2026 Igalia, S.L. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-Intl.Locale.prototype.getCalendars
+description: >
+  When the locale has no region, the Add Likely Subtags algorithm is used to
+  derive the region whose calendar preference is returned.
+info: |
+  RegionPreference ( locale )
+  ...
+  2. If _region_ is *undefined*, then
+    a. Set _region_ to CanonicalUnicodeSubdivision(_locale_, *"sd"*).
+    b. If _region_ is *undefined*, then
+      i. Let _maximal_ be the result of the Add Likely Subtags algorithm applied
+         to _locale_. If an error is signaled, set _maximal_ to _locale_.
+      ii. Set _maximal_ to CanonicalizeUnicodeLocaleId(_maximal_).
+      iii. Set _region_ to GetLocaleRegion(_maximal_).
+      iv. If _region_ is *undefined*, then
+        1. Set _region_ to *"001"*.
+  ...
+features: [Intl.Locale, Intl.Locale-info]
+---*/
+
+// In order not to rely on a particular implementation's locale data, this test
+// searches for a suitable locale without a region subtag, but where the likely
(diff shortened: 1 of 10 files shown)


ptomato:
The tests show that JSC and V8 both fail to take the `-u-sd` extensions into account for hour cycles, but correctly apply them for calendars.

test/intl402/Locale/prototype/getCalendars/likely-subtags-region.js
```diff
@@ -0,0 +1,69 @@
...
+      iii. Set _region_ to GetLocaleRegion(_maximal_).
+      iv. If _region_ is *undefined*, then
+        1. Set _region_ to *"001"*.
+  ...
+features: [Intl.Locale, Intl.Locale-info]
+---*/
+
+function arrayEqual(a, b) {
+  return a.length === b.length && a.every((value, i) => value === b[i]);
+}
```
Ms2ger: This is `compareArray` (not `assert.compareArray`) from `assert.js`, so I think you can use that?

test/intl402/Locale/prototype/getCalendars/likely-subtags-region.js
```diff
@@ -0,0 +1,69 @@
...
+  const candidates = ["th-TH", "fa-IR", "ja-JP", "sa-IN", "ps-AF"];
+
+  for (const candidate of candidates) {
+    const locale = new Intl.Locale(candidate);
+    const calendarsWithLikelyRegion = locale.getCalendars();
+
+    const bareLanguage = locale.language;
+    const fallbackLocale = new Intl.Locale(`${bareLanguage}-001`);
+    if (!arrayEqual(fallbackLocale.getCalendars(), calendarsWithLikelyRegion))
+      return [bareLanguage, calendarsWithLikelyRegion];
```
Ms2ger: My personal preference is bracing single-line ifs; take that into account as much or as little as you like

Ms2ger (approved):

test/intl402/Locale/prototype/getCalendars/likely-subtags-region.js
ptomato: I did not realize that was exposed! I'll check.

ptomato: `compareArray` isn't technically exposed, as it's not in the harness file's `defines` frontmatter.

ptomato: (But it seems to me that's probably a mistake, since it is still in the `defines` of the now-empty `harness/compareArray.js`. I'll add a commit fixing that here.)

ptomato:
Normally I'd just merge this after resolving the comments, but since I made a small change to a harness file it's worth another review.

Ms2ger (approved):
Thanks!

PR #5081, opened 2026-07-07
[import-defer] Checking SCC sync readiness of a cycle with TLA

caiolima:
This is the case for the changes on [1]. In this dependency graph, we have a situation where the cycle root of {A (TLA), B} cycle is A (TLA) and B gets to EVALUATED state before its cycle root. In such scneario, an `import defer` of B (in this case it gets there through D -> B) needs to wait until B's SCC gets to `EVALUATED` state.

[1] - https://github.com/tc39/proposal-defer-import-eval/pull/85

### DIFF
--- a/test/language/import/import-defer/evaluation-top-level-await/async-cycle-dependency-of-deferred-module/main.js
+++ b/test/language/import/import-defer/evaluation-top-level-await/async-cycle-dependency-of-deferred-module/main.js
@@ -8,13 +8,16 @@ description: >
 info: |
   In this test, the module A contains top-level await and forms a cycle with
   the module B, and the module Middle defers the module D, whose only
-  dependency is B. The first dynamic import starts evaluating the graph from
-  A, so A becomes the [[CycleRoot]] of the strongly connected component
-  {A, B}. When the concurrent dynamic import of Middle then starts
-  evaluating, B is already EVALUATED, but its cycle root A is still
-  EVALUATING-ASYNC. Middle's body must therefore only execute after the cycle
-  {A, B} has been fully evaluated, and accessing the deferred namespace of D
-  must then only evaluate D itself.
+  dependency is B. This test first dynamically imports A and waits until A's
+  evaluation has started and is suspended on the "blocker" promise, so A is
+  the [[CycleRoot]] of the strongly connected component {A, B} and is
+  guaranteed to still be EVALUATING-ASYNC. It then dynamically imports C,
+  which imports Middle followed by ResolveBlocker. When Middle's evaluation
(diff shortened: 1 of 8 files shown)


test/language/import/import-defer/evaluation-top-level-await/async-cycle-dependency-of-deferred-module/main.js
```diff
@@ -0,0 +1,78 @@
...
+    1. Return true.
+flags: [module, async]
+features: [import-defer, top-level-await, dynamic-import]
+includes: [compareArray.js]
+---*/
+
+import "./setup_FIXTURE.js";
+
+const pA = import("./a-tla_FIXTURE.js");
+const pMiddle = import("./middle_FIXTURE.js");
```
nicolo-ribaudo: Nothing guarantees whether we call `a-tla_FIXTURE .Evaluate()` or `middle_FIXTURE .Evaluate()` first.

We can force an order with something like this test: https://github.com/tc39/test262/blob/7a096c205fd422ecba49a407d5ac4d1b3f842296/test/language/module-code/top-level-await/rejection-order.js

caiolima: I missed that. Thanks for pointing to other test. I think in this new version I can guarantee the ordering now.

test/language/import/import-defer/evaluation-top-level-await/async-cycle-dependency-of-deferred-module/main.js
```diff
@@ -0,0 +1,86 @@
...
+
+// Wait until A's evaluation has started and is suspended on the blocker, so
+// that A is guaranteed to be EVALUATING-ASYNC when Middle starts evaluating.
+await aStarted.promise;
+
+const pC = import("./c_FIXTURE.js");
+
+await Promise.all([pA, pC]);
+
+assert.compareArray(globalThis.evaluations, [
```
nicolo-ribaudo, line 76: Could you also add to this list C and resolve-blocker?

nicolo-ribaudo (approved):
Minor suggestion, but the test looks good

Ms2ger (approved):

nicolo-ribaudo:
This got consensus in the July 2026 meeting; it can be merged once https://github.com/tc39/test262/pull/5081#discussion_r3595369169 is resolved.

caiolima:
> This got consensus in the July 2026 meeting; it can be merged once [#5081 (comment)](https://github.com/tc39/test262/pull/5081#discussion_r3595369169) is resolved.

Done.

PR #5087, opened 2026-07-13
add tests for `using` and `await using` statements at the top level of modules

### LINKED ISSUE #5073: top-level using
Our XS engine has supported Explicit Resource Management for years behind a flag. We are preparing to enable that by default, now that the proposal has achieved Stage 4. 🎉 Despite passing 100% of Explicit Resource Management tests in test262, we found that XS was not invoking `[Symbol.dispose]` when exiting the top-level of a module (see example below). Perhaps it is worth extending test262 coverage to ensure this is handled correctly and consistently.

```javascript
class Foo {
	close() {
		trace(`Dispose!\n`);
	}
	toString() {
		return "I'm foo";
	}
	static {
		this.prototype[Symbol.dispose] = this.prototype.close;
	}
}

using x = new Foo
trace(x, "\n");
```

bakkot:
Fixes https://github.com/tc39/test262/issues/5073.

### DIFF
--- a/INTERPRETING.md
+++ b/INTERPRETING.md
@@ -142,7 +142,7 @@ This must precede any additional text modifications described by test metadata.
 ### Modules
 
 Test262 includes tests for ECMAScript 2015 module code, denoted by the "module"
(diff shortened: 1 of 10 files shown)


gibson042 (approved):

phoddie:
Thank you!

PR #5092, opened 2026-07-23
Test removal of default locale fallback in Intl.Locale.p.getCollations

ptomato:
This PR adds baseline tests for the return value of Intl.Locale.p.getCollations() reflecting the situation before normative PR https://github.com/tc39/ecma402/pull/1072 in the first commit, and in the second commit makes the necessary changes for the normative PR.

Note that because the default locale is out-of-band, `und-locale.js` might pass or fail if the normative change is not implemented, depending on what the environment's default locale is. It should always pass regardless of the environment's default locale if the normative change is implemented correctly. Try running it with `LC_ALL=de` in the environment, for example.

PR #5093, opened 2026-07-23
Test removal of default locale fallback in Intl.Locale.p.getCollations

gibson042:
Replaces #5092, which I accidentally closed.

This PR adds baseline tests for the return value of Intl.Locale.p.getCollations() reflecting the situation before normative PR https://github.com/tc39/ecma402/pull/1072 in the first commit, and in the second commit makes the necessary changes for the normative PR.

Note that because the default locale is out-of-band, `und-locale.js` might pass or fail if the normative change is not implemented, depending on what the environment's default locale is. It should always pass regardless of the environment's default locale if the normative change is implemented correctly. Try running it with `LC_ALL=de` in the environment, for example.


ptomato (approved):
Rubber-stamp (I actually opened this PR originally and Richard reviewed it)

PR #5095, opened 2026-07-24
Add tests for iterator close semantics of Array.fromAsync

littledivy:
The existing `Array.fromAsync` tests cover that an abrupt mapper completion closes the iterator (`mapfn-*-throws-close-*-iterator.js`), but not the surrounding close semantics:

- **`does-not-close-exhausted-async-iterator.js` / `does-not-close-exhausted-sync-iterator.js`**: a normal completion (`done: true`) must not call `return()`; iteration ends with a return completion, never AsyncIteratorClose.
- **`mapfn-throws-close-async-iterator-awaited.js`**: the close is awaited, so the promise from `return()` settles before the promise returned by `Array.fromAsync` is rejected (AsyncIteratorClose step 5.d).
- **`mapfn-throws-close-rejection-not-masked.js`**: when both the mapper and `return()` fail, the promise rejects with the mapper's error (AsyncIteratorClose step 6 returns the original completion).

These distinguish implementations today: quickjs-ng closed exhausted iterators without awaiting the result (fix in quickjs-ng/quickjs#1596, which prompted these tests), and V8 (Node 24.14) never settles the returned promise at all when `return()` throws or rejects during an abrupt close, so the last test currently hangs there.

### DIFF
--- a/test/built-ins/Array/fromAsync/does-not-close-exhausted-async-iterator.js
+++ b/test/built-ins/Array/fromAsync/does-not-close-exhausted-async-iterator.js
@@ -1,0 +1,47 @@
+// Copyright (C) 2026 Divy Srivastava. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-array.fromasync
+description: >
+  The iterator of an asynchronous iterable is not closed when the iterator
+  reports it is done.
+info: |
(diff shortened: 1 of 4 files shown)


littledivy:
V8 CL: https://chromium-review.googlesource.com/c/v8/v8/+/8140683

PR #5097, opened 2026-07-27
Add Temporal duration and DateTimeFormat regression coverage

arshaw:
Adds Test262 coverage for two previously uncovered behaviors:

- `Temporal.Duration.from()` rejects duration strings with a date component followed by a dangling time designator, such as `P1DT`.
- `Intl.DateTimeFormat` honors inherited and non-enumerable options when formatting a `Temporal.PlainDate`.

The DateTimeFormat test compares against a reference formatter instead of hard-coding locale-specific output.

CC @ptomato

### DIFF
--- a/test/intl402/DateTimeFormat/prototype/format/temporal-objects-exotic-options.js
+++ b/test/intl402/DateTimeFormat/prototype/format/temporal-objects-exotic-options.js
@@ -1,0 +1,38 @@
+// Copyright (C) 2026 Adam Shaw. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-datetime-format-functions
+description: Inherited and non-enumerable options apply when formatting Temporal objects
+features: [Temporal]
+locale: [en]
+---*/
+
+const date = new Temporal.PlainDate(2020, 5, 15);
+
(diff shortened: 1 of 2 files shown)


Ms2ger (approved):

Ms2ger:
Please rebase

test/intl402/DateTimeFormat/prototype/format/temporal-objects-exotic-options.js
```diff
@@ -0,0 +1,38 @@
...
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-datetime-format-functions
+description: Inherited and non-enumerable options apply when formatting Temporal objects
+features: [Temporal]
+locale: [en]
+---*/
+
+const date = new Temporal.PlainDate(2020, 5, 15);
```
ptomato, line 11: This test seems like it should apply regardless of whether a Temporal object or legacy Date object is being formatted. So maybe it's worth doing this and removing the `Temporal` feature from the front matter?

```suggestion
const date = new Date(Date.UTC(2020, 4, 15));
```

ptomato (commented):
Sorry for the long response time on this PR. Thanks for submitting it!

PR #5101, opened 2026-08-03
CI: Update of CI actions

ptomato:
These new versions should eliminate all of the warnings about actions running on Node 20 rather than Node 24. I looked at the changelogs and it seems that no further changes are needed to the invocations of the actions.

**Edit:** I tacked on a few improvements to the workflow definitions, fixing my own mistake in confusing `files:` and `paths:`, and removing duplicate job runs if the PR is from the same repo. The tools tests are now skipped if no files in the `tools/` folder were changed, ~~but due to a limitation in GitHub (https://github.com/orgs/community/discussions/44490) they still show up as "passed".~~

### DIFF
--- a/.github/workflows/test-tools.yml
+++ b/.github/workflows/test-tools.yml
@@ -4,72 +4,95 @@ on:
   push:
     branches:
       - main
-    files:
-      - 'tools/**'
   pull_request:
-    files:
-      - 'tools/**'
 
 jobs:
+  changed-tools:
+    name: Check for tools changes
+    runs-on: ubuntu-latest
+    outputs:
+      any_changed: ${{ steps.changed.outputs.any_changed }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+
+      - name: Identify changed tools files
+        id: changed
+        uses: tj-actions/changed-files@v47.0.6
+        with:
+          files: tools/**
+
   lint:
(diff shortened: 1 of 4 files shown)


ptomato:
> fixing my own mistake in confusing `files:` and `paths:`

...which (sigh) never would have worked as intended anyway, since jobs that are skipped apparently cannot be required...

gibson042 (approved):

bakkot (approved):

.github/workflows/test-tools.yml
```diff
@@ -2,72 +2,97 @@ name: Test tools
...
+
+      - name: Identify changed tools files
+        id: changed
+        uses: tj-actions/changed-files@v47.0.6
+        with:
+          files: tools/**
+
   lint:
     name: Test the lint tool
+    needs: changed-tools
```
bakkot: You could gate the checkout as well, here and below. Not necessary but no reason not to afaict.

bakkot: Actually, any reason not to just gate the whole job rather than each individual step?

ptomato: OK, that works. To summarize the discussion in the maintainers chat, I had thought that that wouldn't work because it didn't work when I skipped the job with `paths:`. But apparently those are two different skipping mechanisms.

PR #5102, opened 2026-08-03
[object-rest] Add coverage for comma-after-rest-property

gibson042:
e.g., `let {...rest,} = obj;` or `({...rest,} = obj)`.

Currently failed by engine262, GraalJS, Hermes, and Moddable XS.

Also performs some minor cleanup in src/dstr-assignment:
* Remove array-rest-elision-invalid.case (redundant with array-rest-before-elision.case).
* Fix a typo in obj-rest-not-last-element-invalid.case ("AssignmenProperty" → "AssignmentProperty")


src/dstr-assignment/array-rest-elision-invalid.case
```diff
@@ -1,20 +0,0 @@
-// Copyright (C) 2015 the V8 project authors. All rights reserved.
```
bakkot, line 1: Presumably you should also delete the generated tests? Or am I not understanding how this works?

bakkot (approved):
LGTM other than comment.

Incidentally I know at least some projects had a related bug for function parameter lists (e.g. [TypeScript](https://github.com/Microsoft/TypeScript/pull/22262)), and I didn't immediately see any coverage for that either.

src/dstr-assignment/array-rest-elision-invalid.case
gibson042: They were removed; GitHub just rendered them as renames.

PR #5103, opened 2026-08-04
Add tests for out-of-order braced quantifier in regexp literals

### LINKED ISSUE #819: Missing test for step 3 of 'Term :: Atom Quantifier', eg. /a{2,1}/
We have three tests for invalid quantifiers in regexs: [1](https://github.com/tc39/test262/blob/master/test/language/literals/regexp/invalid-braced-quantifier-exact.js), [2](https://github.com/tc39/test262/blob/master/test/language/literals/regexp/invalid-braced-quantifier-lower.js), [3](https://github.com/tc39/test262/blob/master/test/language/literals/regexp/invalid-braced-quantifier-range.js). But as far as I can tell we have no test for the SyntaxError in step 3 of the ' Term :: Atom Quantifier' section [here](https://tc39.github.io/ecma262/#sec-term).

That SyntaxError is slightly odd: it appears to me to be a runtime and not Early error. So, if I read the spec right, `if (false) /a{2,1}/` is not an error. SpiderMonkey and JSC at least treat it as one. EDIT: and ChakraCore. v8 is the odd one out in that it treats this, per spec, as a runtime error and not a syntax error.

hexbinoct:
Fixes #819.

test262 currently covers the `{n,m}` case where n > m only through the RegExp
constructor, in `test/built-ins/RegExp/15.10.2.5-3-1.js` (`new
RegExp("0{2,1}")`), which is a runtime error. The regexp literal form is the
early error, and it has no coverage. There is also an incidental use in
`test/annexB/built-ins/RegExp/prototype/compile/pattern-string-invalid.js`, but
that test is about `compile`.

This adds two tests in `test/language/literals/regexp/`:

* `invalid-quantifier-out-of-order.js` for `/a{2,1}/`
* `u-invalid-quantifier-out-of-order.js` for `/a{2,1}/u`

One note on the issue text. When this was filed in 2017 the rule lived in step 3
of the runtime semantics for `Term :: Atom Quantifier`, and the observation there
was that it looked like a runtime error rather than an early error, with V8 the
only engine treating it that way. In the current spec it is an early error, in
22.2.1.1:

    QuantifierPrefix :: { DecimalDigits , DecimalDigits }
    It is a Syntax Error if the MV of the first DecimalDigits is strictly
    greater than the MV of the second DecimalDigits.

so `phase: parse` is right for both files.

Annex B does not change this. B.1.2.1 adds the `ExtendedAtom ::
InvalidBracedQuantifier` error and modifies the NonemptyClassRanges rules, but
leaves the QuantifierPrefix rule alone, and the Annex B `Term` ordering considers
`ExtendedAtom Quantifier` before bare `ExtendedAtom`, so `a{2,1}` still binds as
an atom with a quantifier. The SyntaxError is the same in Annex-B and
non-Annex-B environments.

Verified with the repo linter, and run against V8 and engine262. I also checked
that swapping `{2,1}` for `{1,2}` makes both tests fail, so they are not passing
because of an unrelated parse error.

The Test262 CLA is signed. I have used my GitHub handle on the copyright lines;
happy to change them to my legal name from the CLA if you would prefer that.

### DIFF
--- a/test/language/literals/regexp/invalid-quantifier-out-of-order.js
+++ b/test/language/literals/regexp/invalid-quantifier-out-of-order.js
@@ -1,0 +1,25 @@
+// Copyright (C) 2026 hexbinoct. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-patterns-static-semantics-early-errors
+description: >
+    Braced quantifier with a lower bound greater than its upper bound
+info: |
+    QuantifierPrefix :: { DecimalDigits , DecimalDigits }
+
+    It is a Syntax Error if the MV of the first DecimalDigits is strictly
+    greater than the MV of the second DecimalDigits.
+
+    Annex B replaces Term with an alternative that admits ExtendedAtom, but
+    B.1.2.1 does not modify this rule, and the ExtendedAtom Quantifier
+    alternative is considered before the bare ExtendedAtom alternative. The
+    SyntaxError is therefore consistent between Annex-B and non-Annex-B
+    environments.
+negative:
+  phase: parse
+  type: SyntaxError
+---*/
+
+$DONOTEVALUATE();
+
+/a{2,1}/;
--- a/test/language/literals/regexp/u-invalid-quantifier-out-of-order.js
+++ b/test/language/literals/regexp/u-invalid-quantifier-out-of-order.js
@@ -1,0 +1,22 @@
+// Copyright (C) 2026 hexbinoct. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-patterns-static-semantics-early-errors
+description: >
+    Braced quantifier with a lower bound greater than its upper bound (`u` flag)
+info: |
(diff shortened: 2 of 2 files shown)


ptomato (approved):
Makes sense, thanks.

If I'm reading the spec correctly, it is indeed the case that this is a SyntaxError because of the early error for _PrimaryExpression_: "It is a Syntax Error if IsValidRegularExpressionLiteral(_RegularExpressionLiteral_) is **false**."

hexbinoct:
This has been approved since August 14 and still merges cleanly, so I wanted to check
whether anything else is needed from me before it can go in.

I have not rebased it, on purpose. Its checks are the pre-migration CircleCI engine runs
and all of them are green, while branches updated since #5104 pick up the new Actions
engine jobs along with a `CircleCI Pipeline` status that is erroring at the moment. Happy
to rebase once that settles, or straight away if you would rather see this one on the new
setup.

ptomato:
The engine262 failure indicates a defect in test262harness, not engine262, the uploaded artifact shows it's throwing the SyntaxError but test262harness is registering a test failure.

PR #5104, opened 2026-08-10
Nifty CI table

### LINKED ISSUE #4118: Automatically add a results table for CI runs on pull requests

ptomato:
This removes the old CircleCI config (which is often broken) and replaces it with a GitHub Action that runs new and changed tests on each available engine, and posts a comment with a nifty Markdown table showing which tests passed on which engine, and which failed and why.

Example of what the result looks like (on a scratch repo, so this link will disappear at some point after this PR is merged): https://github.com/ptomato/test262temp/pull/1#issuecomment-5170202750

LLM disclosure: I used a bot to make a first draft of this, and refine that draft several times. I then made further modifications without using a bot until I was satisfied with the result and that I understood it. I also removed long-winded LLM comments and replaced them with my own words if I felt a comment was necessary. Do note that I'm not super familiar with GH Actions config so if I missed anything important it's likely to be in `engine-tests.yml` or `engine-tests-report.yml`.

Closes: #4118


tools/misc/build-results-comment.mjs
```diff
@@ -0,0 +1,162 @@
...
+    let data;
+    try {
+      data = JSON.parse(fs.readFileSync(filePath, 'utf8'));
+    } catch (err) {
+      console.warn(`Failed to parse ${file}: ${err.message}`);
+      continue;
+    }
+
+    if (!Array.isArray(data)) {
+      console.warn(`Unexpected format in ${file}: expected JSON array`);
```
Ms2ger: Should this add a note to the comment instead? I doubt it'll get noticed here.

tools/misc/build-results-comment.mjs
```diff
@@ -0,0 +1,162 @@
...
+      continue;
+    }
+
+    for (const test of data) {
+      const testPath = test.relative || test.file;
```
Ms2ger: Why do we need to check two properties here?

tools/misc/build-results-comment.mjs
```diff
@@ -0,0 +1,162 @@
...
+      const result = results[engineKey];
+      return formatCell(result);
+    });
+    lines.push('| ' + [testPath, ...cells].join(' | ') + ' |');
+  }
+
+  return lines.join('\n');
+}
+
+function formatCell(result) {
```
Ms2ger, line 120: Should this check for `|` characters that would break the table? In general, how robust should this code be against untrusted json?

.github/workflows/engine-tests.yml
```diff
@@ -0,0 +1,76 @@
...
+      all_changed_files: ${{ steps.changed.outputs.all_changed_files }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+
+      - name: Identify new or changed tests
+        id: changed
+        uses: tj-actions/changed-files@v47.0.6
+        with:
+          files: test/
```
Ms2ger, line 21: Should this not be

```suggestion
          files: test/**
```
?

Ms2ger (approved):

nicolo-ribaudo:
If the table is long, could it be inside a `<details>` element? I'm thinking of PR that change 100 tests, for which the table would get a bit in the way.

.github/workflows/engine-tests.yml
ptomato: I don't think so, we use `changed-files` in several of the other workflow files without the `**` and it's been working fine.

tools/misc/build-results-comment.mjs
ptomato: I think in general it should be robust. But I didn't realize that pipes inside the link destination or title would break the table, good catch!

Here's my current "evil exception value" that I'm testing against: https://github.com/ptomato/test262temp/pull/1/changes#diff-8a9c494bdac49e59a9d740c059d1badc8795753ab0b8c512b4f99315a9098b78R21

HTML elements are not rendered inside the link titles, but I guess there are security implications if the Markdown formatting is broken and HTML elements make it outside somehow. So I've escaped `<` as well.

ptomato:
> If the table is long, could it be inside a `<details>` element? I'm thinking of PR that change 100 tests, for which the table would get a bit in the way.

I've made a cutoff of 15 tests, more than that and the table will be collapsed inside `<details>`.

tools/misc/build-results-comment.mjs
ptomato: Good idea, I've made the error handling a bit better.

tools/misc/build-results-comment.mjs
ptomato: Agreed, I should've reviewed this part a bit harder. Sorry :grimacing: I've done so now and caught an additional bug.

tools/misc/build-results-comment.mjs
```diff
@@ -0,0 +1,176 @@
...
+  const lines = ['| ' + header.join(' | ') + ' |'];
+  lines.push('|' + header.map(() => ' --- ').join('|') + '|');
+
+  for (const testPath of tests) {
+    const results = testResults.get(testPath);
+    const cells = engines.map(engineKey => {
+      const result = results[engineKey];
+      return formatCell(result);
+    });
+    lines.push('| ' + [testPath, ...cells].join(' | ') + ' |');
```
bakkot, line 114: Strictly speaking this will also needs escaping, but I think probably our paths don't have special characters in practice.

bakkot (approved):
LGTM, one nitpick but probably not worth addressing.

tools/misc/build-results-comment.mjs
ptomato: We have the escaping function though, so why not.

PR #5105, opened 2026-08-13
Add template literal tests for supplementary-plane unicode escape sequences

### LINKED ISSUE #3001: Test missing for \u{0x10000} ... \u{0x10FFFF} in Template Literals
Bullet 3 - 5 of https://tc39.es/ecma262/#sec-utf16encodecodepoint isn't hit when when a \u{...} unicode escape sequence in the range of 0x10000 through 0x10FFFF is hit in a Template Literal

hexbinoct:
Closes #3001.

Template literals had no coverage for `\u{}` escapes above U+FFFF, so the surrogate pair branch of UTF16EncodeCodePoint (steps 3 to 5) was never exercised. The only astral escapes in these tests today are the invalid seven-digit `\u{10FFFFF}` cases.

The new file checks the cooked and raw values of tagged templates for U+10000 (the first code point that needs a surrogate pair), U+10437 (the character the spec itself uses as its UTF-16 example), U+10FFFF (the largest code point), and a leading-zeros form, plus the untagged path and concatenation with adjacent template characters. Expected values are written as explicit `\uD800\uDC00` style pairs so they do not depend on the construct under test.

Following jugglinmike's suggestion in the issue, the assertions live in a dedicated file next to `tv-utf16-escape-sequence.js`. Thanks to p-bakker for reporting the gap from their work on Rhino.

The copyright line says hexbinoct. Happy to swap in the legal name from the CLA if that is preferred.

### DIFF
--- a/test/language/expressions/template-literal/tv-utf16-escape-sequence-supplementary.js
+++ b/test/language/expressions/template-literal/tv-utf16-escape-sequence-supplementary.js
@@ -1,0 +1,95 @@
+// Copyright (C) 2026 hexbinoct. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+/*---
+esid: sec-static-semantics-tv
+description: >
+    Template values of \u{} escape sequences for code points above U+FFFF
+info: |
+    The TV of TemplateCharacter :: \ TemplateEscapeSequence is the SV of
+    TemplateEscapeSequence.
+    The SV of UnicodeEscapeSequence :: u{ CodePoint } is the result of
+    performing UTF16EncodeCodePoint on the MV of CodePoint.
+
+    UTF16EncodeCodePoint ( cp )
+
+    1. Assert: 0 <= cp <= 0x10FFFF.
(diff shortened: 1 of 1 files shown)


ptomato:
Thanks!

Totally separate from a code review, I don't know the answer to the copyright name question.

bakkot:
My understanding is that the copyright notice isn't actually the thing that creates the copyright; it's mainly to track provenance as things get copied to new places, and to ensure you get credit. So your username should be OK if that's what you'd prefer. [We already have some tests with just usernames](https://github.com/tc39/test262/blob/3655e7464de3d52643ecddd4b5f9f4f3e7f62398/test/annexB/built-ins/RegExp/legacy-accessors/lastMatch/prop-desc.js#L1).

Incidentally [US law explicitly allows](https://www.law.cornell.edu/uscode/text/17/401) using "a generally known alternative designation of the owner" in a copyright notice, though I don't know about other countries.

hexbinoct:
Thanks both, I will leave the copyright line as hexbinoct, and the prop-desc.js precedent is a useful one to have on record.

PR #5107, opened 2026-08-17
Temporal: Test Duration.prototype.{round,total} with zoned property bags -> out-of-epoch-range

MidnightDesign:
While working on increasing coverage of [my PHP port of Temporal](https://github.com/MidnightDesign/temporal-php), my agent (Claude Code) encountered a gap in this test suite: calling `round` or `total` on out-of-epoch-range `Duration`s is never tested with zoned property bags. This adds those tests. The tests were AI-generated.

V8 passes both new tests.


ptomato (commented):
Thanks!

It looks like we also don't have coverage for this for property bags without `timeZone`, either? Maybe the property bag tests could be added to `relativeto-plaindate-large-time-component-out-of-range.js` and `relativeto-zoneddatetime-large-time-component-out-of-range.js` respectively.

PR #5108, opened 2026-08-17
Temporal: Add tests for unavailable time zones to `Instant`

MidnightDesign:
Adds tests for passing unavailable time zones (both bare and within ISO strings) to `Instant.prototype.toString` and `Instant.prototype.toZonedDateTimeISO`. The tests were generated by Claude Code.


ptomato (approved):
Thanks!

PR #5117, opened 2026-08-25
Handle auto-rebase actions in CI table

ptomato:
A limitation of the engine-tests-report workflow running as a separate, trusted workflow was that if you rebased a branch in the GitHub UI using the "Update branch" button, it wouldn't have access to the PR number in github.event.workflow_run.pull_requests. This array is allowed to be empty according to the GitHub Actions docs. So the table wouldn't get created or updated.

So instead save the PR number while preparing to run the engine-tests jobs (the original, untrusted workflow always has a PR number) and upload it as an artifact, which is downloaded by the engine-tests-report workflow.

Another option might be to use the jwalton/gh-find-current-pr action, but I wasn't super confident it would be actively maintained.

LLM disclosure: I used a bot to evaluate a couple of potential solutions, generated one, then hand-edited the generated code (it initially didn't work)

### DIFF
--- a/.github/workflows/engine-tests-report.yml
+++ b/.github/workflows/engine-tests-report.yml
@@ -42,6 +42,23 @@ jobs:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           run-id: ${{ github.event.workflow_run.id }}
 
+      - name: Download PR number
+        uses: actions/download-artifact@v8
+        with:
+          name: pr_number.txt
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          run-id: ${{ github.event.workflow_run.id }}
+
+      - name: Read PR number
+        id: pr
+        run: |
+          PR_NUMBER="$(head pr_number.txt -n1)"
+          if [[ ! "$PR_NUMBER" =~ ^[1-9][0-9]*$ ]]; then
+            echo "Unexpected contents of pr_number.txt: $PR_NUMBER"
+            exit 1
+          fi
+          echo "number=$PR_NUMBER" >> "$GITHUB_OUTPUT"
+
       - name: Check for results
         id: check
         run: |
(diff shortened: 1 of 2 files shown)


ptomato:
(I tested the rebase-in-UI event on https://github.com/ptomato/test262temp/pull/2, by the way)

.github/workflows/engine-tests-report.yml
```diff
@@ -42,6 +42,17 @@ jobs:
...
+      - name: Download PR number
+        uses: actions/download-artifact@v8
+        with:
+          name: pr_number.txt
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          run-id: ${{ github.event.workflow_run.id }}
+
+      - name: Read PR number
+        id: pr
+        run: echo "number=$(cat pr_number.txt)" >> "$GITHUB_OUTPUT"
```
Ms2ger: How confident are you that this is safe with untrusted input?

ptomato: I think we can trust that it's the same artifact that the other workflow uploaded? Artifacts are [immutable](https://github.com/actions/upload-artifact#not-uploading-to-the-same-artifact) and if an attacker raced to upload a malicious `pr_number.txt` artifact before the other workflow did, the other workflow would fail and this workflow wouldn't run.

We could still sanitize that it's a number, then at worst the wrong PR would get a comment.

Ms2ger: I don't really understand the security model for these workflows, but checking that the file contains a single number would reassure me, yes

ptomato: Done, and tested in https://github.com/ptomato/test262temp/pull/3

Ms2ger (approved):
Thanks!

PR #5119, opened 2026-08-29
Fix duplicate include directive in TypedArray set test

masnagam:
Remove duplicate 'testTypedArray.js' from the includes list in src-typedarray-not-big-throws.js.

The includes directive had the same file listed twice, which is unnecessary.

### DIFF
--- a/test/built-ins/TypedArray/prototype/set/BigInt/src-typedarray-not-big-throws.js
+++ b/test/built-ins/TypedArray/prototype/set/BigInt/src-typedarray-not-big-throws.js
@@ -15,7 +15,7 @@ info: |
       other does not, throw a TypeError exception.
   ...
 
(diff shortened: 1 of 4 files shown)


masnagam:
During testing with the bee-browser/bee JavaScript runtime, duplicate includes were detected in this file and three related files. The same pattern of duplicate 'testTypedArray.js' entries was found in:

- test/built-ins/TypedArray/prototype/set/src-typedarray-big-throws.js
- test/built-ins/TypedArrayConstructors/ctors-bigint/typedarray-arg/src-typedarray-not-big-throws.js
- test/built-ins/TypedArrayConstructors/ctors/typedarray-arg/src-typedarray-big-throws.js

Additional commits have been added to this PR to fix all four files together.

gibson042 (approved):

PR #5121, opened 2026-08-30
speed up linting

bakkot:
This was generated by Claude in response to

> figure out why linting is slow and make it faster (with minimal changes - this still needs to be maintained by the same set of people, so don't completely rewrite everything).

On my machine this takes it from ~90 seconds to ~5 seconds. Half of that is from caching the checks instead of reconstructing them per file; the remainder is from using pyyaml's c parser instead of the python one. Also bumps the version of pyyaml so that it is more likely to have the c parser included in CI; [the only breaking change](https://github.com/yaml/pyyaml/blob/main/CHANGES) is dropping support for Python 2.7, which seems fine.

I am reasonably sure the change to get the events by running the parser twice is semantics-preserving, and it's actually faster on my machine than the current subclassing approach even when using pyyaml's python-based parser.


gibson042 (approved):
LGTM, and (referring to the prompt) I'm happy to continue maintaining the result.

ptomato (approved):
Nice, thanks for looking into this.

Issue #5115, opened 2026-08-24
create AI policy


bakkot:
We should have a policy on AI-authored PRs. I would like to accept them but require disclosure, and (preferably) to forbid using AI to write PR descriptions or commit messages.

ptomato:
That matches my preference for a policy. I guess the PR descriptions and commit messages are already covered by TC39's policy.

As for disclosure I'd prefer discouraging `Assisted-By:` or `Co-Authored-By:` in favour of disclosure of the author's process — as a reviewer, I don't really care about which version of which model they used, but I do care about whether the tests were generated verbatim or hand-edited or whatever, and how they were verified.

Issue #5122, opened 2026-08-31
Testing plan for Thenable-Curtailment (Safe Promise Resolve)


mgaudet:
Testing Plan for https://tc39.es/proposal-thenable-curtailment/ 

Setup: 
- [ ] Add a `safeResolvePromise` host-defined function to allow testing the proposal (which otherwise has no user access in its current shape)

Synchronous Resolve

- [ ] Primitive values resolve synchronously
- [ ] Object with no `then` property resolve synchronously
- [ ] Object with un-callable `then` resolve synchronously.
- [ ] Object with un-callable `then` on prototype resolves synchronously
- [ ] Object with `then` setter, but no getter, resolve synchronously
- [ ] Object with `then` setter, but no getter, on the prototype. resolves synchronously. 
- [ ] Object with own-property then which is not callable shadows a callable then, resolves synchronously.

Deferred Resolve & Execution of User Code

- [ ] Resolving with a `Proxy` (even with empty traps) resolves deferred.
- [ ] Resolving with a Proxy with a `get` trap is resolved.
  - [ ] The get trap runs in deferral. 
- [ ] Resolving an object which has a proxy on its prototype chain is deferred. 
- [ ] Resolving with a module namespace object is deferred
- [ ] Object with callable `then` own-property is deferred 
   - [ ] The `then` call is made, deferred. 
- [ ] Object with callable `then` prototype is deferred
  - [ ] The then call is made deferred
- [ ] A deferred `then` throwing should cause the promise to be rejected. 
- [ ] Object with `then` getter
  - [ ] The then getter is run, deferred. 
 - [ ] Object with `then` getter on prototype. 
  - [ ] The then getter is run, deferred. 
 - [ ] `then` getter which throws rejects promise, but deferred 
 - [ ] Resolving with a revoked proxy rejects deferred 
 - [ ] Callable Object.prototype.then causes object resolution to be deferred 
   - [ ] unless the object has a null prototype.  
 - [ ] Ensure deferred `then` call has correct this binding. 
 - [ ] Tick count of safe resolving a promise with another native promise (resolved, unresolved). 
 - [ ] Delete the callable `then` after a safeResolve job is enqueued but before it's processed. 
 - [ ] Ensure the `Get` of `then` is run twice when 

Latching behaviour 

* [ ] Attempting to resolve a deferred promise of any sort cannot overwrite or otherwise interfere with the deferred promise
* [ ] Late rejection also fails.
[...]

PR #5123, opened 2026-09-01
Tests for Thenable-curtailment

mgaudet:
I just want to call out the commit message on c3386c6, esp. in light of #5115 

> These tests were authored with the assistance of an LLM, however they were 
> targeted and reviewed by myself. 
> 
> The tests were validated on the prototype implemenation of SafePromiseResolve 
> in the Firefox tree; the tests highlight a current divergence between the 
> SpiderMonkey implementation and the spec-text as written (specfically 
> SpiderMonkey will currently defer on a `then` setter even with no getter, 
> which is not correct by specification).

(Of course, seeing that in the Github editor I realize how badly I typo'd some of that message 🤦 )

### DIFF
--- a/INTERPRETING.md
+++ b/INTERPRETING.md
@@ -81,6 +81,7 @@ properties of the global scope prior to test execution.
           Use this property to test that ECMAScript algorithms aren't mis-implemented to treat `document.all` as being `undefined` or of type Undefined (instead of Object).
 
           **Tests using this function must be tagged with the `IsHTMLDDA` feature so that only hosts supporting this property will run them.**
+  - **`safeResolvePromise`** - (present only in implementations that can provide it) a function which takes as its first argument a promise and as its second argument a value, and uses the SafeResolve semantics provided by the [Thenable Curtailment](https://github.com/tc39/proposal-thenable-curtailment) proposal to resolve the provided promise with the value. 
   - **`agent`** - an ordinary object with the following properties:
     - **`start`** - a function that takes a script source string and runs
       the script in a concurrent agent. Will block until that agent is
--- a/test/built-ins/Promise/safe-resolve-promise/deferred-proxy-get-trap.js
+++ b/test/built-ins/Promise/safe-resolve-promise/deferred-proxy-get-trap.js
@@ -1,0 +1,81 @@
+// Copyright (C) 2026 Mozilla Corporation. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-safe-promise-resolve
+description: >
+    A Proxy "get" trap returning a callable "then" is invoked from the deferred
+    job, not from SafePromiseResolve itself.
+info: |
+    SafePromiseResolve ( promiseCapability, resolution )
+
+    3. Let _deferredSteps_ be a new Abstract Closure that captures _promise_ and
+       _resolution_ and performs the following steps when called:
+      a. Perform ? PerformPromiseResolution(_promise_, _resolution_, ~deferred~).
+
+    PerformPromiseResolution ( promise, resolution, thenCallTiming )
+
+    4. Let _then_ be Completion(Get(_resolution_, *"then"*)).
+    ...
+    9. If _thenCallTiming_ is ~deferred~, then
+      a. Perform ! PerformPromiseResolveThenable(_promise_, _resolution_,
+         _thenAction_).
+includes: [asyncHelpers.js, compareArray.js]
+flags: [async]
+features: [thenable-curtailment, safeResolvePromise, promise-with-resolvers, Proxy]
+---*/
+
+var expected = [
+  // Being a Proxy forces deferral before any trap can run.
+  "start",
+
+  "tick 1",
+
+  // The deferred job reads "then" once and calls what the trap returned.
+  "get:then",
+  "call then",
+
+  "tick 2",
+  "settled",
+];
+
+var actual = [];
+
+// "then" is the only property read from the resolution, so any other lookup
+// shows up in the comparison below as an unexpected entry.
+var value = new Proxy({}, {
+  get: function(target, key) {
+    actual.push("get:" + String(key));
+    if (key !== "then") {
+      return undefined;
+    }
+    return function(resolve) {
+      actual.push("call then");
+      resolve("from the trap");
+    };
+  },
+});
+
+asyncTest(function() {
+  var ruler = Promise.resolve(0)
+    .then(() => actual.push("tick 1"))
+    .then(() => actual.push("tick 2"))
+    .then(() => {
+      assert.compareArray(actual, expected, "Ticks for a Proxy \"get\" trap");
+    });
+
+  var capability = Promise.withResolvers();
+  $262.safeResolvePromise(capability.promise, value);
+  actual.push("start");
+
+  var settled = capability.promise.then(function(settledValue) {
+    actual.push("settled");
+    assert.sameValue(
+      settledValue,
+      "from the trap",
+      "promise is fulfilled with the value passed to the resolving function"
+    );
+  });
+
+  return Promise.all([ruler, settled]);
+});
--- a/test/built-ins/Promise/safe-resolve-promise/deferred-proxy-on-prototype-chain.js
+++ b/test/built-ins/Promise/safe-resolve-promise/deferred-proxy-on-prototype-chain.js
@@ -1,0 +1,93 @@
+// Copyright (C) 2026 Mozilla Corporation. All rights reserved.
+// This code is governed by the BSD license found in the LICENSE file.
+
+/*---
+esid: sec-has-property-which-could-run-user-code
+description: >
+    SafePromiseResolve defers resolution when an ordinary object has a Proxy on
+    its prototype chain.
+info: |
+    PropertyAccessCouldRunUserCode ( o, propertyKey, kind )
+
+    2. If _o_ has the [[GetPrototypeOf]] and [[GetOwnProperty]] internal methods
+       as defined in Proxy Object Internal Methods and Internal Slots, return
+       *true*.
+    ...
+    7. Let _proto_ be _o_.[[GetPrototypeOf]]().
+    8. If _proto_ is *null*, return *false*.
+    9. Return PropertyAccessCouldRunUserCode(_proto_, _propertyKey_, _kind_).
+includes: [asyncHelpers.js, compareArray.js]
+flags: [async]
+features: [thenable-curtailment, safeResolvePromise, promise-with-resolvers, Proxy]
+---*/
+
+var actual = [];
+
+// The Proxy is the immediate prototype, and further up the chain behind
+// ordinary objects.
+function makeValue(label, depth) {
+  var proxyProto = new Proxy({}, {
+    get: function(target, key, receiver) {
+      actual.push("get:" + String(key) + " " + label);
+      return Reflect.get(target, key, receiver);
+    },
+  });
+
+  var value = proxyProto;
+  for (var i = 0; i < depth; i += 1) {
+    value = Object.create(value);
+  }
+  return value;
+}
+
+var expected = [
+  "start",
+
+  "tick 1",
+
+  // Each deferred job performs its own "then" lookup.
+  "get:then immediate",
+  "get:then deep",
+
+  "tick 2",
+
+  // No callable "then" is found, so each promise fulfills with the resolution.
+  "settled immediate",
+  "settled deep",
+];
+
+asyncTest(function() {
+  var ruler = Promise.resolve(0)
+    .then(() => actual.push("tick 1"))
+    .then(() => actual.push("tick 2"))
+    .then(() => {
+      assert.compareArray(
+        actual,
+        expected,
+        "Ticks for a Proxy on the prototype chain"
+      );
+    });
+
+  var checks = [];
+
+  [["immediate", 1], ["deep", 3]].forEach(function(entry) {
+    var label = entry[0];
+    var value = makeValue(label, entry[1]);
+
+    var capability = Promise.withResolvers();
+    $262.safeResolvePromise(capability.promise, value);
+
+    checks.push(capability.promise.then(function(settledValue) {
+      actual.push("settled " + label);
+      assert.sameValue(
+        settledValue,
+        value,
+        "promise is fulfilled with the resolution itself: " + label
+      );
+    }));
+  });
+
+  actual.push("start");
+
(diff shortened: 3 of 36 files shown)


mgaudet:
(See #5122 for the testing plan)

INTERPRETING.md
```diff
@@ -81,6 +81,7 @@ properties of the global scope prior to test execution.
           Use this property to test that ECMAScript algorithms aren't mis-implemented to treat `document.all` as being `undefined` or of type Undefined (instead of Object).
 
           **Tests using this function must be tagged with the `IsHTMLDDA` feature so that only hosts supporting this property will run them.**
+  - **`safeResolvePromise`** - (present only in implementations that can provide it) a function which takes as its first argument a promise and as its second argument a value, and uses the SafeResolve semantics provided by the [Thenable Curtailment](https://github.com/tc39/proposal-thenable-curtailment) proposal to resolve the provided promise with the value.
```
bakkot, line 84: This signature doesn't work.

Lots of the tests rely on the fact that a Promise's state becomes locked to follow another thenable once its original resolver has been invoked. The thing is, at least at the spec level, this locking is a property of _the resolvers that are created when the Promise was originally created_, not of the Promise itself; that's why the later resolvers that get passed to the settlement value's `.then` can settle the Promise even though the original resolvers no longer can (and those resolvers have their own lock bit in case that later resolver is passed a thenable, and so on).

So this needs to be passed the resolve function from the original capability. The most straightforward thing would be to pass it a `{ promise, resolve }` pair (probably with a note that any other properties should be ignored, so it can be passed the result of `Promise.withResolvers` directly).

test/built-ins/Promise/safe-resolve-promise/deferred-proxy-get-trap.js
```diff
@@ -0,0 +1,81 @@
...
+  "tick 2",
+  "settled",
+];
+
+var actual = [];
+
+// "then" is the only property read from the resolution, so any other lookup
+// shows up in the comparison below as an unexpected entry.
+var value = new Proxy({}, {
+  get: function(target, key) {
```
bakkot: We should test all the traps, not just `get`. No need to write them out manually, just something along the lines of

```js
var value = new Proxy({}, new Proxy({}, {
  get: function (_, trap) {
    return function (...args) {
      actual.push({ trap, args });
      if (trap === 'get' && args[1] === 'then') {
        return function (resolve) {
          actual.push("call then");
          resolve("from the trap");
        };
      }
      return Reflect[trap](...args);
    };
  }
}));
```
I don't think this ends up changing the expectations.

test/built-ins/Promise/safe-resolve-promise/deferred-proxy-on-prototype-chain.js
```diff
@@ -0,0 +1,93 @@
...
+features: [thenable-curtailment, safeResolvePromise, promise-with-resolvers, Proxy]
+---*/
+
+var actual = [];
+
+// The Proxy is the immediate prototype, and further up the chain behind
+// ordinary objects.
+function makeValue(label, depth) {
+  var proxyProto = new Proxy({}, {
+    get: function(target, key, receiver) {
```
bakkot: Ditto above re other traps.

bakkot (commented):
Looks generally good except that the hook has the wrong signature. I think it should be straightforward to fix that with an LLM though. Also a couple gaps for which I have filed https://github.com/mgaudet/test262/pull/1.

I confess I did not trace through the timing for the timing tests. I'm just going to trust you got that right.

Mapping the original test plan:

Synchronous Resolve
- [x] Primitive values resolve synchronously: [sync-non-object.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/sync-non-object.js)
- [x] Object with no `then` property resolve synchronously: [sync-no-then.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/sync-no-then.js)
- [x] Object with un-callable `then` resolve synchronously: [sync-then-not-callable.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/sync-then-not-callable.js)
- [x] Object with un-callable `then` on prototype resolves synchronously: [sync-then-not-callable-proto.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/sync-then-not-callable-proto.js)
- [x] Object with `then` setter, but no getter, resolve synchronously: [sync-then-setter-only.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/sync-then-setter-only.js)
- [x] Object with `then` setter, but no getter, on the prototype. resolves synchronously: [sync-then-setter-only-proto.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/sync-then-setter-only-proto.js)
- [x] Object with own-property then which is not callable shadows a callable then, resolves synchronously: [sync-own-non-callable-then-shadows-callable.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/sync-own-non-callable-then-shadows-callable.js)

Deferred Resolve & Execution of User Code
- [x] Resolving with a `Proxy` (even with empty traps) resolves deferred: [deferred-proxy-empty-handler.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/deferred-proxy-empty-handler.js)
- [x] Resolving with a Proxy with a `get` trap is resolved: [deferred-proxy-get-trap.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/deferred-proxy-get-trap.js)
  - [x] The get trap runs in deferral: [deferred-proxy-get-trap.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/deferred-proxy-get-trap.js)
- [x] Resolving an object which has a proxy on its prototype chain is deferred: [deferred-proxy-on-prototype-chain.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/deferred-proxy-on-prototype-chain.js)
- [x] Resolving with a module namespace object is deferred: [deferred-module-namespace.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/deferred-module-namespace.js); fixture: [deferred-module-namespace_FIXTURE.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/deferred-module-namespace_FIXTURE.js)
- [x] Object with callable `then` own-property is deferred: [deferred-then-callable-own.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/deferred-then-callable-own.js)
  - [x] The `then` call is made, deferred: [deferred-then-callable-own.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/deferred-then-callable-own.js)
- [x] Object with callable `then` prototype is deferred: [deferred-then-callable-proto.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/deferred-then-callable-proto.js)
  - [x] The then call is made deferred: [deferred-then-callable-proto.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/deferred-then-callable-proto.js)
- [x] A deferred `then` throwing should cause the promise to be rejected: [deferred-then-call-throws.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/deferred-then-call-throws.js)
- [x] Object with `then` getter: [deferred-then-getter-own.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/deferred-then-getter-own.js)
  - [x] The then getter is run, deferred: [deferred-then-getter-own.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/deferred-then-getter-own.js)
- [x] Object with `then` getter on prototype: [deferred-then-getter-proto.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/deferred-then-getter-proto.js)
  - [x] The then getter is run, deferred: [deferred-then-getter-proto.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/deferred-then-getter-proto.js)
- [x] `then` getter which throws rejects promise, but deferred: [deferred-then-getter-throws.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/deferred-then-getter-throws.js)
- [x] Resolving with a revoked proxy rejects deferred: [deferred-proxy-revoked.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/deferred-proxy-revoked.js)
- [x] Callable Object.prototype.then causes object resolution to be deferred: [deferred-object-prototype-then.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/deferred-object-prototype-then.js)
  - [x] unless the object has a null prototype: [sync-null-proto-ignores-object-prototype-then.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/sync-null-proto-ignores-object-prototype-then.js)
- [x] Ensure deferred `then` call has correct this binding: [deferred-then-this-binding.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/deferred-then-this-binding.js)
- [x] Tick count of safe resolving a promise with another native promise (resolved, unresolved): [deferred-native-promise-ticks.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/deferred-native-promise-ticks.js), [deferred-native-promise-pending-ticks.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/deferred-native-promise-pending-ticks.js), [deferred-native-promise-pending-through-assimilation-ticks.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/deferred-native-promise-pending-through-assimilation-ticks.js)
- [x] Delete the callable `then` after a safeResolve job is enqueued but before it's processed: [deferred-then-deleted-before-job.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/deferred-then-deleted-before-job.js)
- [ ] Ensure the `Get` of `then` is run twice when: **got cut off, no idea what this was**.

Latching behaviour
- [x] Attempting to resolve a deferred promise of any sort cannot overwrite or otherwise interfere with the deferred promise: [latch-after-deferred.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/latch-after-deferred.js), [latch-thenable-resolving-functions.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/latch-thenable-resolving-functions.js), [latch-thenable-resolving-functions-pending.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/latch-thenable-resolving-functions-pending.js)
- [x] Late rejection also fails: [latch-after-deferred.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/latch-after-deferred.js), [latch-after-sync.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/latch-after-sync.js), [latch-thenable-resolving-functions.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/latch-thenable-resolving-functions.js), [latch-thenable-resolving-functions-pending.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/latch-thenable-resolving-functions-pending.js)
- [x] Repeated safe resolution: Latched by first and second is no-op: [latch-double-safe-resolve.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/latch-double-safe-resolve.js)
- [x] Attempting to safe resolve an already resolved promise is a no-op: [latch-already-settled.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/latch-already-settled.js)

Edge Cases
- [ ] (If possible) test HostObjectHasImpureLookupInternalMethods: **not really testable**
- [x] Self-resolve: (note: Spec currently will defer depending on the inspection of current promise.. but I actually think maybe we should change the spec to synchronously throw here?): [sync-self-resolution.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/sync-self-resolution.js), [deferred-self-resolution.js](https://github.com/mgaudet/test262/blob/thenable-curtailment-testing/test/built-ins/Promise/safe-resolve-promise/deferred-self-resolution.js). These cover the current spec I think.

test/built-ins/Promise/safe-resolve-promise/latch-after-ordinary-pending.js
```diff
@@ -0,0 +1,83 @@
...
+  var thenCallCount = 0;
+
+  return Promise.all([
+    checkIgnoredResolution("non-thenable", function() {
+      return "must be ignored";
+    }),
+
+    checkIgnoredResolution("thenable", function() {
+      return {
+        then: function(resolve) {
```
bakkot, line 70:
```suggestion
        get then() {
```
Might as well assert it's not even touched, not just that it's not called.

bakkot (approved):
LGTM other than nit.

Issue #615, opened 2016-05-03
Collator and NumberFormats test append '-u-key-value' without checking if the default locale already has '-u-'.


jungshik:
intl402/Collator/10.1.1_19_c,  intl402/NumberFormat/prototype/format/11.3.2_TRF.js and a few other tests append '-u-key-value' to the default locale name. If the default locale happens to have '-u-' (e.g. en-US-u-va-posix) , the result would be invalid because it'd look like 'en-US-u-va-posix-u-kn-true' with 
two '-u-' extension. 

Collator/10.1.1_19_c.js has the following

```
 collator = new Intl.Collator([defaultLocale + "-u-" + option.key + "-" + value]);
```

When the test is run under 'C' locale (translated to en-US-va-posix by ICU) on Linux or Mac (in Mac terminal, the default is 'C' locale), the test fails with v8 (d8) leading to the following error:

```
 RangeError: Invalid language tag: en-US-u-va-posix-u-kn-true
```

v8 works around this by running intl402 tests under an explicitly set locale instead of C. See
https://bugs.chromium.org/p/v8/issues/detail?id=4437

zbraniecki:
`Intl.Locale` proposal would give us an API to manipulate locale string - https://github.com/zbraniecki/proposal-intl-locale

PR #5125, opened 2026-09-02
The default locale should not contain a Unicode extension sequence

jessealama:
`DefaultLocale()` isn't supposed to contain a Unicode locale extension sequence. `ResolveLocale` substitutes it with an empty extension when no requested locale matches, so the resolved locale of an Intl object, constructed without a locale argument, should never contain `"-u-"`.

It looks like we don't currently test this. The new test checks it for every Intl service constructor.

This is the requirement underlying #615. The tests there append `"-u-"` to the default locale, which is OK, but only if the default locale doesn't already have one. Under the C locale, JSC reports `"en-US-u-va-posix"` as its default, and this test makes that failure visible.

Spec: https://tc39.es/ecma402/#sec-defaultlocale


gibson042 (changes requested):
I think I'd rather see this covered by an extended [and possibly renamed] test/intl402/default-locale-is-canonicalized.js.

gibson042:
I've also opened an issue to explore allowing the currently nonconforming behavior: https://github.com/tc39/ecma402/issues/1098

ptomato (approved):
Notwithstanding Richard's proposed reorganization which I don't feel strongly about either way, the test contents LGTM.

Ms2ger (approved):

jessealama:
I've folded the check into `default-locale-is-canonicalized.js` as suggested, and dropped the separate file. I also added an `info` block and `esid` to go with it.

jessealama:
-awaiting-author

test/intl402/default-locale-is-canonicalized.js
```diff
@@ -3,14 +3,30 @@
...
+    The returned String value represents the well-formed and canonicalized
+    language tag for the host environment's current locale. It must not contain
+    a Unicode locale extension sequence.
+
+    ResolveLocale ( availableLocales, requestedLocales, options,
+                    relevantExtensionKeys, localeData )
+
+    ...
+    5. If match is undefined, set match to the Record { [[locale]]:
+       DefaultLocale(), [[extension]]: empty }.
```
gibson042, line 23: https://tc39.es/ecma402/#sec-resolvelocale
```suggestion
    ResolveLocale ( availableLocales, requestedLocales, options,
                    relevantExtensionKeys, localeData )

    ...
    4. If match is undefined, set match to the Record { [[locale]]:
       DefaultLocale(), [[extension]]: empty }.
```

gibson042 (approved):

PR #5126, opened 2026-09-02
Don't append Unicode extensions to a default locale that already has one

### LINKED ISSUE #615: Collator and NumberFormats test append '-u-key-value' without checking if the default locale already has '-u-'.

jessealama:
Fixes #615.

`testNumberFormat` and nine standalone tests build locales by appending strings like `"-u-<key>-<value>"` to the default locale coming from `resolvedOptions()`. When a host reports a default locale that already carries a Unicode extension sequence (e.g. `"en-US-u-va-posix"`), the result has two `"-u-"` substrings. This causes the constructor to throw a `RangeError` before these tests reach the behavior they're meant to check. JSC does this in CI, where 43 tests currently fail for that reason.

This PR adds a new `getLocaleBaseName` helper to `testIntl.js`. It strips extension and private-use subtags (thereby mirroring `Intl.Locale.p.baseName`). The helper is then applied at each site before any extension gets appended.

Whether the default locale may contain an extension at all is a separate question, covered by a new conformance test in #5125.

The helper uses plain string juggling rather than `Intl.Locale`. That's arguably a more proper solution, but using `Intl.Locale` would add a feature dependency to legacy Collator and NumberFormat tests. I also notice that V8's `baseName` currently returns `"en-US-u-va-posix"` unchanged for exactly this input, so the helper wouldn't do anything there. Switching to `Intl.Locale` once both concerns are resolved would be a reasonable follow-up.


ptomato (commented):
I'm not positive about what the right move is here. This seems to work fine. But on the other hand, it seems simpler (and arguably more robust) to insert the extension sequences in the right place with an `addUExtension(key, value)` helper function:
- If there's no `-u-`, append `-u-${key}-${value}`.
- If there is `-u-` and no other single-letter subtags after it, append `-${key}-${value}`.
- If there is `-u-` and any other single-letter subtag after it, insert `-${key}-${value}-` right before the first such single-letter subtag.

Ms2ger (approved):

PR #5129, opened 2026-09-03
CI table improvements

ptomato:
- Make the failed test emoji less big and red
- Drop irrelevant info from test artifacts
- Print test results in CI log

### DIFF
--- a/.github/workflows/engine-tests.yml
+++ b/.github/workflows/engine-tests.yml
@@ -74,7 +74,8 @@ jobs:
             --hostArgs="${{ matrix.engine.hostArgs }}" \
             --reporter json \
             -- ${{ needs.changed-tests.outputs.all_changed_files }} \
-            > "results-${{ matrix.engine.key }}.json"
+            | jq 'map({relative, scenario, result})' \
+            | tee "results-${{ matrix.engine.key }}.json"
 
       - name: Upload results
         if: always()
--- a/test/intl402/Locale/prototype/getWeekInfo/firstDay-by-id.js
+++ b/test/intl402/Locale/prototype/getWeekInfo/firstDay-by-id.js
@@ -13,6 +13,7 @@ info: |
 features: [Reflect,Intl.Locale,Intl.Locale-info]
 ---*/
 
+
 const validFirstDayOfWeekIds = [
   ["en-u-fw-mon", 1],
   ["en-u-fw-tue", 2],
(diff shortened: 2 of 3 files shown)


ptomato:
I guess we can't test the :heavy_minus_sign: emoji on this PR because the privileged comment-writing workflow is always going to be run from main.

Ms2ger (approved):
Consider using ➕ for passing tests for consistency

test/intl402/Locale/prototype/getWeekInfo/firstDay-by-id.js
```diff
@@ -13,6 +13,7 @@ info: |
 features: [Reflect,Intl.Locale,Intl.Locale-info]
 ---*/
 
+
```
gibson042, line 16: Why add this line?

tools/misc/build-results-comment.mjs
```diff
@@ -151,8 +151,8 @@ function combineMessages(sloppyMessage, strictMessage) {
 }
 
 function formatFail(message) {
-  if (!message) return '❌';
-  return `[❌](## "${sanitizeTableMarkdown(message)}")`;
+  if (!message) return '➖';
```
gibson042, line 154: I'd also like the emoji to be part of an obvious pair—(👍, 👎), (🙆, 🙅), (✅, ❎), etc.

.github/workflows/engine-tests.yml
```diff
@@ -74,7 +74,8 @@ jobs:
             --hostArgs="${{ matrix.engine.hostArgs }}" \
             --reporter json \
             -- ${{ needs.changed-tests.outputs.all_changed_files }} \
-            > "results-${{ matrix.engine.key }}.json"
+            | jq 'map({relative, scenario, result})' \
+            | tee "results-${{ matrix.engine.key }}.json"
```
gibson042, line 78: We can do better than JSON in this log output, e.g.
```suggestion
            | tee "results-${{ matrix.engine.key }}.json"
            | jq -r '
              map(
                "\n\(if .result.pass then "PASS" else "FAIL" end) \(.relative) [\(.scenario)]\(if .result.pass then "" else "\n\(.result.message)" end)"
              )
              | .[]
            '
```

to produce text like
```
PASS path/to/foo [default]

FAIL path/to/bar [default]
failure message
```

gibson042 (approved):

test/intl402/Locale/prototype/getWeekInfo/firstDay-by-id.js
ptomato: This is just temporary, in order to get a test to run.

PR #5137, opened 2026-09-18
Expect DataView.prototype.{byteLength,byteOffset} not to throw when out of bounds

MattiasBuelens:
Accompanies https://github.com/tc39/ecma262/pull/3972.

This comes with two behavioral changes:
* `DataView.byteLength/byteOffset` no longer throw when backed by a detached buffer.
* `DataView.byteLength/byteOffset` no longer throw when the view becomes out of bounds after resizing the backing buffer.

The updated tests are now more closely aligned with the existing tests for typed arrays, which was the goal of the spec change.

_Disclaimer: I used an LLM to help update these tests. I have manually reviewed the new tests, and verified that they pass in https://github.com/engine262/engine262/pull/365._


test/built-ins/DataView/prototype/byteOffset/resizable-array-buffer-fixed.js
```diff
@@ -26,20 +41,17 @@ try {
 assert.sameValue(dataView.byteOffset, 1, "following grow");
 
 try {
-  ab.resize(BPE * 3);
```
bakkot, line 29: This was a pre-existing bug, huh.

bakkot (approved):
LGTM except I didn't review the `info` blocks.

test/built-ins/DataView/prototype/byteOffset/resizable-array-buffer-fixed.js
MattiasBuelens: Yep. It was silently caught, and it didn't affect the test's expectations. But it errored for the wrong reason: an unreferenced variable, rather than an throwing resize operation.

Would you prefer I split this off to a separate PR?

bakkot: Nope, fine to fix here, just noting.

